PR #4, opened 2017-02-12
questions

malaffoon:
This isn't a real pull request, i just don't know how else to discuss a possible change with the rest of the team. I am questioning the need for returning immutable collections from these methods, please take this as a conversation starter. Couple thoughts:
1. The interface for the repo returns Set<X> while the impl returns an ImmutableSet<X>. That is misleading and is usually only discovered by the client during runtime testing. Very unsatisfying.
2. Whenever possible i prefer to stick with core java classes in interfaces. Using a third-party class in an interface tightly binds both the implementation **and the client** to that third-party. Ok, in this case it is guava which isn't too bad but still.
3. I'm not sure using immutable here adds value. In general i love immutable things but the call is really simple controller-service-repo. Does it matter whether the collection is immutable or not?

I do feel strongly about point 1. If we keep immutable collections, we should change the interface for the repo. But, i believe that might be problematic (it certainly is when using JPA).


service/src/main/java/rdw/reporting/services/TranslationService.java
```diff
@@ -1,23 +1,23 @@
...
 import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
 
 public interface TranslationService {
 
 	/**
 	 * @return All ranslations
 	 */
-	ImmutableSet<Translation> getTranslations();
```
malaffoon: point 2

service/src/main/java/rdw/reporting/repositories/impl/TranslationRepositoryStub.java
```diff
@@ -29,7 +29,6 @@
 		checkNotNull(locale, "Argument \"locale\" must not be null");
 		return stubTranslations.stream()
 			.filter(translation -> locale.equals(translation.getLocale()))
-			.collect(toImmutableSet());
```
malaffoon: point 1

esotrope:
Hi Mark. I think I agree with you in this case.

For the record here was my thought process for people curious about the reasoning behind the use of immutables:

1. The repository layer returns mutable collections because in a controller>service>repository layered system the service is expected to perform transformations and deliver the data in its final form the the controller layer.
2. The service layer returns immutables as a way of trying to strongly advise the controller/service consumer that the collection should not be modified from this point on in the code
3. The service method signatures are typed as Immutable* because if the service layer returns immutables we would not want to type it as not-immutable because then the client would erroneously conclude that it can modify the returned collection without causing an exception.

Overall, it is a habit I historically have adopted to force best practices for separation of concerns and layering but can be omitted under the circumstance that the developers understand these underlying spring system controller/service/repo responsibilities.

malaffoon:
Thanks Landon!

PR #5, opened 2017-02-13
Groups api

esotrope:


service/src/main/java/rdw/reporting/repository/GroupRepositoryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+		this.jdbcTemplate = checkNotNull(jdbcTemplate);
+		this.getGroupSummaries = checkNotNull(getGroupSummaries);
+		this.getGroup = checkNotNull(getGroup);
+	}
+
+	public Set<GroupSummary> getGroupSummaries(@NotNull User user) {
+		return Sets.newHashSet(
+			jdbcTemplate.query(
+				getGroupSummaries,
+				ImmutableMap.of("user_login", "dwtest@example.com"),
```
malaffoon: Can we push the hard-coding of the username up a layer or two? In other words, this line would be `ImmutableMap.of("user_login", user.getUsername()),` and the dwtest@example.com would be injected at the controller layer (and tests) as a hard-coded default.

service/src/main/java/rdw/reporting/repository/GroupRepositoryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+		);
+	}
+
+	public Optional<Group> getGroup(@NotNull User user, long id) {
+		final ImmutableGroup.Builder group = ImmutableGroup.builder().id(id);
+		final List<Student> students = jdbcTemplate.query(
+			getGroup,
+			ImmutableMap.of(
+				"id", id,
+				"user_login", "dwtest@example.com"
```
malaffoon: ditto

service/src/main/resources/application.yaml
```diff
@@ -0,0 +1,36 @@
...
+    getGroupSummaries:
+      >
+      select
+        reporting.roster.id as id,
+        reporting.roster.name as name,
+        count(*) as size
+      from reporting.student
+      join reporting.roster_membership on reporting.student.id=reporting.roster_membership.student_id
+      join reporting.roster on reporting.roster_membership.roster_id=reporting.roster.id
+      where reporting.roster.id in (select roster_id from reporting.user_roster where user_login=:user_login)
```
malaffoon: Is the column in user_roster really `user_login`? Seems like it should just be `user_id` or `user_name`?

esotrope: yes it currently is user_login

service/src/main/java/rdw/reporting/repository/GroupRepositoryImpl.java
esotrope: Actually I forgot to fix this - it should be local only

PR #12, opened 2017-02-20
Updating yaml->sql code to combine aspects of old way (dynamic discov…  … …ery and loading of yaml files) with usability of @Value(${}) notation

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/12)


src/main/java/rdw/reporting/Application.java
```diff
@@ -1,12 +1,20 @@
...
+import rdw.reporting.support.YamlApplicationContextInitializer;
 
 @SpringBootApplication
 public class Application {
 
+
     public static void main(String[] args) {
-        SpringApplication.run(Application.class, args);
+
+        new SpringApplicationBuilder(Application.class).initializers(new YamlApplicationContextInitializer()).run(args);
```
rcordeau: We are adding a new custom initializer that looks for *.sql.yml files to load as properties

src/main/java/rdw/reporting/support/YamlApplicationContextInitializer.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.springframework.boot.env.YamlPropertySourceLoader;
+import org.springframework.context.ApplicationContextInitializer;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.core.env.PropertySource;
+import org.springframework.core.io.Resource;
+
+import java.io.IOException;
+
+public class YamlApplicationContextInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
```
esotrope: this seems like something we should eventually move to rdw_common I'm thinking. Just need to parametrize the *.yml string

rcordeau: yeah eventually, but for now I just want it to work with our stuff, since we are interacting with the database the most at the moment.  Considering that all of that code is pretty much Spring code, unit testing it would be moot.

rcordeau:
Thanks - felt good about keeping requested usability with the flexibility of separate sql files.

src/main/java/rdw/reporting/support/YamlApplicationContextInitializer.java
agorina: @rcordeau - I would like to start using it to in the ingest world. I guess I can move it into the common then?

agorina: I am not sure if this will work with the Database integration tests. I do not think it initiates the full application context. I may be wrong on this. 
We should try it. We have a test set up in the RDW_Ingest, AssessmentRepositoryImplIT.

rcordeau: Sorry - meant to get to this earlier today.  Yes, it will work with AssessmentRepositoryImplIT.  There needs to be another annotation on the test class:
@ContextConfiguration(classes = ExamProcessorApplication.class, initializers = YamlFileApplicationContextInitializer.class). I can move it to common if you want to keep SQL in separate files in other projects.

agorina: @rcordeau, that may be problematic. This requires the entire application context to be loaded during the tests. It may become an issue if the project has multiple dependencies. We should talk about it at the parking lot.

PR #16, opened 2017-02-23
Moving the YamlExternalPropertiesConfiguration class to RDW_Common. A…

rcordeau:
Application.java now imports common configuration bean.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/16)


agorina:
@esotrope, it it was something that only reporting tier needs, then I would agree with you. However, we need it in the Ingest project. So this truly belongs to common.

PR #17, opened 2017-02-23
use Fairway conventions for package naming and group/artifact id

malaffoon:


src/main/webapp/.gitignore
```diff
@@ -7,6 +7,7 @@
 # dependencies
 /node_modules
 /bower_components
+/node-v*
```
malaffoon: to avoid 2418 unversioned files in my tree

PR #18, opened 2017-02-24
Feature/views and navigation

esotrope:


src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1,27 +1,112 @@
...
     "types": [
       {
         "shortName": "SUMMATIVE",
-        "name": "Summative"
+        "name": "Summative",
+        "performanceLevels": [
+          "Level 1",
+          "Level 2",
+          "Level 3",
+          "Level 4"
```
agorina: Is this temporary until we are done with the real work?
I think the back end needs to drive this data.

src/main/webapp/src/app/students/students.component.ts
```diff
@@ -14,14 +14,15 @@ export class StudentsComponent implements OnInit {
...
+  constructor(private service: DataService, private route: ActivatedRoute) {
+  }
 
   ngOnInit() {
     this.route.params
       .subscribe(params => {
         this.service.getGroup(params['groupId'])
           .subscribe((group: any) => {
-              group.students = sortOn(group.students, student => student.lastName, student => student.firstName);
+              group.students = sortAscOn(group.students, student => student.lastName, student => student.firstName);
```
agorina: Does this expect all students in a Group to be returned at once?

src/main/webapp/src/assets/i18n/en.json
esotrope: This is the local file for standalone mode. so the client can be stood up without the backend.
In "backend mode" the client hits /api/translations/{locale} for the JSON translations which can then be driven by a file + db entries

src/main/webapp/src/app/students/students.component.ts
esotrope: right now yes - will change if we expect to have large amounts of students in a group

PR #21, opened 2017-03-16
adds selection summaries on group aggregate page, introduces formatte…

esotrope:
…d messages to future-proof translations of web page

### DIFF
--- a/src/main/webapp/src/assets/i18n/en.json
+++ b/src/main/webapp/src/assets/i18n/en.json
@@ -1,6 +1,5 @@
 {
-  "HELLO": "HELLO WORLD!",
-  "welcome": "Welcome",
+  "welcome": "Welcome {{firstName}} {{lastName}}",
   "application": {
     "title": "Interim Item Reporting"
   },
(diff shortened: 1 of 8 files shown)


src/main/webapp/src/assets/i18n/en.json
```diff
@@ -64,7 +63,9 @@
       "selectedRecords": "Records Selected",
       "totalStudents": "Total # of Students",
       "averageScaleScore": "Average Scale Score",
-      "performanceCategories": "Performance Categories"
+      "performanceCategories": "Performance Categories",
+      "selectionSummary": "Selected: {{total}} assessment taken on {{start}} Grade {{assessmentGrade}} {{assessmentName}} {{assessmentAcademicYear}} {{assessmentSubject}}",
```
malaffoon: assessments?

esotrope: Good point. This way of phrasing will have to be dynamic and depend on if the total is 1 or if it is more than one

PR #24, opened 2017-03-21
enables path to student exam report, and creates placeholder for exam…

esotrope:
… reports


src/main/webapp/src/app/shared/routes.ts
```diff
@@ -33,5 +34,9 @@ export const routes = [
   {
     path: 'groups/:groupId/students/:studentId/exams/:examId/items',
     component: StudentExamItemsComponent
+  },
+  {
+    path: 'groups/:groupId/students/:studentId/exams/:examId/report',
```
malaffoon: Not suggesting a change (yet) just asking ... shouldn't students be accessible outside groups? i.e. shouldn't i be able to fetch a student's exams from `students/:studentId/exams/:examId/report`?

esotrope: Good comment later we will possibly have multiple paths to the same view.
The reason for including the group ID in the path is so that we can include it in the breadcrumbs and send the info to the backend to get the group name for the breadcrumbs.

PR #33, opened 2017-03-24
Updated build file to connect to MySQL

agorina:
I doubt that it is perfect, but this enables connection to MySQL server. Otherwise it was failing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,8 +5,8 @@ buildscript {
         maven { url 'http://repo.spring.io/plugins-release' }
     }
     dependencies {
-        classpath("io.spring.gradle:dependency-management-plugin:1.0.0.RELEASE")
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -34,10 +41,10 @@ tasks.withType(JavaCompile) {
...
 repositories {
-        
-     maven { url "http://repo.maven.apache.org/maven2" }
-     maven { url "https://repository.apache.org/content/repositories/snapshots/" }
+    mavenLocal()
+    mavenCentral()
 }
+
 dependencies {
     compile('org.springframework.boot:spring-boot-starter-aop')
```
malaffoon: As long as you're in here can you clean up the format of these to make them consistent with other projects. They should be single quotes without the parens, e.g. 
`compile 'org.springframework.boot:spring-boot-starter-aop'`

agorina: Will do. Thank you.

PR #35, opened 2017-03-24
fixing the gradle build order for the front end static resources

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/35)


bgsavage:
What was the problem?

rcordeau:
The problem was that there is a processResources step that was run first, before the front end resources were even available, and the front end's resources were getting missed on a first pass.  The second pass would process the resources again, but this time seeing the new front end resources and then they would get in to the final jar. This makes it such that the front end is built before any resources are processed (i.e. copied to the target directory).

PR #36, opened 2017-03-25
Clean up - round 2.

agorina:
@esotrope - this is still not final. Just trying to contain the scope of the changes. Please let me know if something looks wrong.

I just noticed that a lot of the changes are due to my formatting settings being different. I used the defaults. Do you have yours different?


src/main/java/org/opentestsystem/rdw/reporting/config/SAMLProperties.java
```diff
@@ -1,43 +1,70 @@
...
-	 */
-	String getIdpMetadataUrl();
-
-	/**
-	 * Service Provider entity id as registered in the IDP circle of trust.
-	 */
-	String getSpEntityId();
+public class SAMLProperties {
 
+    @Value("${saml.key-store-file}")
```
malaffoon: I believe all these @Value annotations are redundant. A properties construct will be populated by default and all the values given match the defaults.

src/main/java/org/opentestsystem/rdw/reporting/controller/GroupController.java
```diff
@@ -1,33 +1,23 @@
...
-
-	@RequestMapping(value = "/{id}/students")
-	public Group getGroup(@AuthenticationPrincipal User user, @PathVariable long id) {
-		return service.getGroup(user, id)
-			.orElseThrow(ResourceNotFoundException::new);
-	}
+    @Autowired
+    private GroupService service;
 
+    @RequestMapping
```
malaffoon: Consider using @GetMapping.

src/main/java/org/opentestsystem/rdw/reporting/model/Group.java
```diff
@@ -1,14 +1,45 @@
...
+     */
+    public static class Builder {
+        private long id;
+        private String name;
+
+        public Group build() {
+            return new Group(id, name);
+        }
+
+        public Builder withName(final String name) {
```
malaffoon: Can we discuss using `name()` instead of `withName()` in builders?

src/main/java/org/opentestsystem/rdw/reporting/controller/GroupController.java
agorina: Thank you, I will. I still have more to change in here.

src/main/java/org/opentestsystem/rdw/reporting/model/Group.java
agorina: Sure, we can discuss.

src/main/java/org/opentestsystem/rdw/reporting/config/SAMLProperties.java
agorina: Yeah, I was lazy - it was no picking it up and I was lazy to research why. Will do.

PR #38, opened 2017-03-28
Feature/cleanup

malaffoon:
The goal is to make this project more consistent with the other RDW projects wrt remote config, docker, unit tests, coverage, etc.

I didn't have much success wrt security and running with local, non-SAML auth. And i couldn't get the JKS file served up from the configuration server properly. Oh, and because of that, i haven't pushed forward on dockerization yet. But some minor tweaks ...


CONTRIBUTING.md
```diff
@@ -0,0 +1,79 @@
+## RDW_Reporting for Developers
```
malaffoon: Trying to separate true developer issues from non-developer consumers of this.

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -48,7 +49,7 @@ public boolean isAccountNonLocked() {
 
     @Override
     public boolean isCredentialsNonExpired() {
-        return false;
+        return credentialsNonExpired;
```
malaffoon: actually found a bug in all of this

src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfig.java
```diff
@@ -1,11 +1,5 @@
-package org.opentestsystem.rdw.reporting.config;
```
malaffoon: oops, most of this is unintentional cmd-alt-O reformatting.

src/main/webapp/.gitignore
```diff
@@ -5,9 +5,9 @@
 /tmp
 
 # dependencies
-/node_modules
```
malaffoon: this wasn't working on my system ... i was getting 16000+ unversioned files.

build.gradle
```diff
@@ -85,9 +92,9 @@ bootRun {
         jvmArgs project.jvmArgs.split('\\s+')
     }
 }
-bootRepackage {
-    mainClass = 'org.opentestsystem.rdw.reporting.Application'
-}
+
+// generate coverage report automatically (./build/reports/jacoco/test/html/index.html)
```
agorina: Yeah!!!!

src/main/resources/application.yml
```diff
@@ -0,0 +1,14 @@
+spring:
+  datasource:
+    url: jdbc:mysql://localhost:3306/reporting?useSSL=false
+    username: root
+    password:
+    driverClassName: com.mysql.jdbc.Driver
+    testWhileIdle: true
+    validationQuery: SELECT 1
+    initialize: false
```
agorina: Since you introduced this one, the yml in the test folder can be deleted.

malaffoon: good point, will do

PR #39, opened 2017-03-29
Feature/dwr 194 iris integration

bgsavage:


src/main/webapp/angular-cli.json
```diff
@@ -21,6 +21,9 @@
         "styles.less"
       ],
       "scripts": [
+        "vendor/iris/jquery.js",
+        "vendor/iris/util_xdm.js",
+        "vendor/iris/client.js"
```
esotrope: Eventually we should probably point to the iris "service" for these scripts I'm thinking. We should look into how to do that at some point.

src/main/webapp/src/app/app.module.ts
```diff
@@ -47,7 +48,8 @@ export function createTranslateLoader(http: Http) {
     HomeComponent,
     StudentExamReportComponent,
     IabReportComponent,
-    AdminSearchComponent
+    AdminSearchComponent,
+    IrisPocComponent
```
esotrope: should probably keep the poc stuff on the poc branch and not develop so it doesnt sneak its way into a release

src/main/webapp/src/app/iris-poc/iris-poc.component.ts
```diff
@@ -0,0 +1,428 @@
...
+
+    // set the vendor guid.
+    //Note: in the OSS IRiS case we do not care for this.
+    let vendorId = '2B3C34BF-064C-462A-93EA-41E9E3EB8333';
+    // let token = {"passage":{"autoLoad":"false"},"items":[{"response":"<p>alla</p>","id":"I-187-2700"}]};
+
+    this.irisClient.loadToken(vendorId, this.token);
+  }
+
+  // initIris() {
```
esotrope: lets remove all commented code

src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -13,48 +13,78 @@ import {AssessmentSubjectType} from "../../shared/assessment-subject-type.enum";
...
+let commonVendorId = '2B3C34BF-064C-462A-93EA-41E9E3EB8333';
+
 export const iab_items = [
   {
     claim: 'Concepts and Procedures',
     target: 'Target F',
     score: 0,
-    maximumScore: 1
+    maximumScore: 1,
+    irisInfo: {
```
esotrope: I think what we would have here is just "irisToken" and the "vendorID" will make its way to the client in a different way, not through our ws response. Also - talked to Jeff Treuting and eventually we will probably move toward the API/proxying strategy so that we can lockdown the iris endpoints with a security layer (IRIS doesnt have auth)

src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -13,48 +13,78 @@ import {AssessmentSubjectType} from "../../shared/assessment-subject-type.enum";
...
   },
   {
     claim: 'Problem Solving',
     target: 'Target A',
     score: 2,
-    maximumScore: 2
+    maximumScore: 2,
+    irisInfo: {
+      vendorId: commonVendorId,
+      token: '{"passage":{"autoLoad":"false"},"items":[{"response":"<p>test123</p>","id":"I-200-22581"}]}'
```
esotrope: instead of adding the same info many times - you can add it once in the map function at the bottom of this mock data setup that function applies changes to all the items

src/main/webapp/src/app/student-exam-items/student-exam-items.component.ts
```diff
@@ -1,55 +1,99 @@
...
+  }
+
+  irisframeOnLoad(){
+    if(this.model.items.length > 0)
+      this.selectRow(this.model.items[0]);
+
+    this.irisIsLoading = false;
+  }
+
+  hoverRow(item) {
```
esotrope: can we just use css :hover style here? not sure why we need a coded event handler

src/main/webapp/angular-cli.json
bgsavage: I think this will always be required to have these on our client side in order to set up their browser messaging system.  It would be nice to have these in some artifactory or npm so that we can download them like we do our other dependencies instead of having to check them into source.

esotrope: Yeah - that's what I'm thinking. Either have these references point to [iris host]/.../...js
or have them pulled in during the build process. so the versioning is driven by one of those as opposed to a snapshot of the script that we pulled down at some point.

src/main/webapp/src/app/app.module.ts
bgsavage: Sure I can remove this.  I wasn't sure if you wanted to keep this around for some other reason.

src/main/webapp/src/app/standalone/data/data.ts
bgsavage: Yes definitely, I would normally would agree with you.  However I took this approach for a couple of reasons:

 1.  The token is slightly varied for each one.  I wanted to test different item ids and layouts
 2.  Simplicity. I want to be able to easily swap out tokens if we want to test other tokens without having to dig and change a bunch of helper functions.  
3.  This is throw away code so just want to optimize it for what makes it easy for us today and not worry about future maintainability.

Unless maybe you're referring to something else?

PR #40, opened 2017-03-29
dockerize

malaffoon:
This adds docker-compose capability at least for local-docker (i didn't put the pushImage stuff in yet).

Everybody please note that i've tried to make the naming and location of the keystore consistent. To make it easy to reuse you should have: `/opt/rdw-reporting-ui/config/rdw-reporting-ui.jks`. The local external application.yml can be wherever you want but it should point to that exact JKS file. Otherwise you'll have to have a different copy when running from docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ store file, and the credentials to read it must be provided in a spring configur
 will contain entries similar to (passwords redacted):
 ```text
 saml:
-  key-store-file: file:/opt/rdw-reporting-ui/config/saml.jks
+  key-store-file: file:/opt/rdw-reporting-ui/config/rdw-reporting-ui.jks
   key-store-password: [redacted]
   private-key-entry-alias: rdw-reporting-ui-sp
   private-key-entry-password: [redacted]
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -117,3 +123,45 @@ task buildFe(type: NpmTask) {
 // In order to get the front end resources in to the final jar, the front end needs to be built
 // before all of the other resources are processed in order to get the static resouces in the right place
 processResources.dependsOn buildFe
+
+// task to generate build-info.properties with build.version, build.timestamp
+// using this approach instead of resource filtering to avoid modifying source
+task createBuildInfoFile(dependsOn: processResources) {
```
malaffoon: this is in anticipation of further work, build-info is not actually being used yet

README.md
```diff
@@ -16,7 +16,7 @@ store file, and the credentials to read it must be provided in a spring configur
 will contain entries similar to (passwords redacted):
 ```text
 saml:
-  key-store-file: file:/opt/rdw-reporting-ui/config/saml.jks
+  key-store-file: file:/opt/rdw-reporting-ui/config/rdw-reporting-ui.jks
```
esotrope: I think you changed /opt/rdw-reporting-ui/... to /opt/rdw-reporting in a different file. I think the latter is what you want here right? and then is the name change for the jks file to show that we want one keystore for the whole app and we will add all keys there?

malaffoon: Yeah, it was different in different places; i swept thru and tried to make it consistent everywhere. As for the actual names, if anybody feels strongly about it i'm open to suggestion.

malaffoon:
No, like the ingest project, the expectation is that MySQL will be running independently.

README.md
esotrope: I think naming it "saml.jks" is best for now since it is only a dependency of the saml components and therefore tells you what it's for.

esotrope: Also, I'm not sure I wrote my other comment very well. I see you have different documentation pointing to different config root dirs:
../rdw-reporting-ui
../rdw-reporting

Can we go in and make them all "rdw-reporting" otherwise people might go in and misconfigure paths I think

malaffoon: I have changed it to ../rdw-reporting everywhere

PR #41, opened 2017-03-29
Feature/dwr 194 iris integration

bgsavage:


src/main/webapp/package.json
```diff
@@ -32,7 +32,6 @@
     "@angular/cli": "1.0.0-beta.31",
     "@angular/compiler-cli": "^2.4.0",
     "@types/jasmine": "2.5.38",
-    "@types/jquery": "^2.0.41",
```
esotrope: good catch we wont need this because we aren't using it in .ts files - is that right?

.gitignore
```diff
@@ -23,3 +23,4 @@ nbbuild/
 dist/
 nbdist/
 .nb-gradle/
+.gradle/
```
esotrope: ask rob/mark about this one - i think this dir is necessarily put in the repo - originally I thought it didn't belong in the repo but was told it did if I remember right

src/main/webapp/src/app/student-exam-items/student-exam-items.component.html
```diff
@@ -61,8 +61,7 @@
             <tab heading="Item Viewer">
               <div style="overflow-x:hidden;overflow-y:scroll">
                 <div *ngIf="irisIsLoading" class="text-align-center">Initializing Item Viewer...  Please wait.</div>
-
-                <iframe #irisframe [src]="safeIrisUrl" width="1000" height="500" frameborder="0"></iframe>
+                <iframe #irisframe [src]="safeIrisUrl" [ngClass]="{'md': size !=0, lg: size == 0}" width="1000" frameborder="0"></iframe>
```
esotrope: you can remove this class conditionality code in favor of:
.card.md iframe{ height...
.card.lg iframe{ height...

.gitignore
bgsavage: @rcordeau  @malaffoon  - Can I add this to .gitignore so I don't have to see pending changes here if they shouldn't be committed?

bgsavage: (I was going to ask on the team slack but didn't want to diverge the current conversation)

rcordeau: The (dot) .gradle directory doesn't need to be checked in, but the gradle directory does.

src/main/webapp/src/app/student-exam-items/student-exam-items.component.html
bgsavage: Are you suggesting to move the [ngClass] to the parent div tabContainer?  There's no parent .card here.

esotrope: oh I see - good point. Yeah we should set the size classing at the container level though. so whatever contains the tabs in general should get the .md/.lg size classes by which to make css form

bgsavage: So we will still have to set the height on the iframe.  I'm not sure if we want to maintain the size of the tab container and the iframe?  Or just let the tabcontainer grow to the size of the content.

See screenshot of what it looks like moving the [ngClass] to the parent container.
image

esotrope: good point - i would consider this window sizing behavior to be up in the air. right now i think you have the right idea - make the bottom tab container resize to the iframe size and just manage the iframe size since we cant have it the size of the content in the frame yet.

PR #42, opened 2017-03-30
binary resources

malaffoon:
This tweak allows us to fetch binary files from the remote config server. So we don't need to worry about all the fiddly bits with that JKS file, it is just magically served up when running with the appropriate profile using remote config. To do this the jks is specified as something like `binary-http://localhost:8888/*/*/master/rdw-reporting.jks`.

@wtritch i included you in case you're interested or want to come over and smack me.
https://contests.projected.com/media/W1siZiIsIjIwMTMvMDkvMjUvMjBfMTRfNThfNjQ5X3RlbmFjaW91czYuanBnIl0sWyJwIiwidGh1bWIiLCI2NTV4MzkwIl1d?sha=503fa109323985f6


CONTRIBUTING.md
```diff
@@ -28,6 +28,30 @@ branch to `develop` to solicit code reviews and feedback. Once approved use `squ
 
 ### Running
 
+#### SAML
```
malaffoon: All this stuff applies only to developers running it locally now.

CONTRIBUTING.md
```diff
@@ -53,10 +77,11 @@ open http://localhost:8080
 ```
 #### Running Standalone
 The artifact is a Spring Boot executable jar so you can just run it. Just as when running from the IDE the default
-is to run without a configuration server so the configuration must be specified on the command line:
+is to run without a configuration server so the configuration must be specified on the command line. You can also
+change the port if desired, e.g.:
 ```bash
-java -jar build/libs/rdw-reporting*.jar --spring.config.location=/opt/rdw-reporting/config/application.yml
```
malaffoon: unrelated, providing an example of how to change port if you want/need to.

README.md
```diff
@@ -9,31 +9,6 @@
 npm install @angular/cli
 ```
 
-#### SAML
```
malaffoon: Non-devs don't have to worry about this anymore.

src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfig.java
```diff
@@ -187,12 +189,11 @@ public SingleLogoutProfile logoutprofile() {
 
     @Bean
     public KeyManager keyManager() {
-        DefaultResourceLoader loader = new DefaultResourceLoader();
```
malaffoon: Instantiating an independent resource loader sort of defeats the purpose of being in an application context.

bgsavage:
So we don't need the saml.jks file from our opt/rdw-reporting/config folder anymore?  But we still need application.yml, correct?

malaffoon:
@bgsavage that depends on how you are running the app. If you are running it locally (from IDE or bootRun or whatever) you will still need the saml.jks. This change only helps if you are just running it from docker.

PR #43, opened 2017-03-30
fixes model to view-model mapping of ica/summative report view

esotrope:


src/main/webapp/src/styles.less
```diff
@@ -519,3 +519,464 @@ html, body {
...
+.ica-report > .report-row > div {
+  display: table-cell;
+  float: none;
+  vertical-align: top;
+}
+
+.ica-report > .report-row > div:first-child {
+  background: #f8f8f8;
+}
+ */
```
bgsavage: Let's remove commented code

src/main/webapp/src/styles.less
```diff
@@ -519,3 +519,464 @@ html, body {
...
+.ica-report.performance-level-0 .performance-score {
+  color: #fff;
+  background-color: #bb231c;
+}
+
+.ica-report.performance-level-0 .performance-level,
+.ica-report.performance-level-0 .score {
+  color: #bb231c;
+}
+
```
bgsavage: Should we create variables for our colors so that we can start defining a color palette to not deviate too much from?

src/main/webapp/src/styles.less
```diff
@@ -519,3 +519,464 @@ html, body {
...
+.ica-report .performance-heading {
+  color: #4b4b4b;
+  font-size: 115%;
+  display: block;
+}
+
+.ica-report .performance-score {
+  font-size: 125%;
```
bgsavage: Should we favor 1.25rem instead?

src/main/webapp/src/styles.less
```diff
@@ -519,3 +519,464 @@ html, body {
...
+  background-color: #237ccb;
+}
+
+.ica-report.performance-level-3 .performance-level,
+.ica-report.performance-level-3 .score {
+  color: #237ccb;
+}
+
+.ica-report .assessment {
+  padding: 16px;
```
bgsavage: We may want to consider reusable css classes instead of specific to a certain element.  In this case, maybe a .padding-md { padding: 16px } ?  This can also help us keep consistency across the app.

src/main/webapp/src/app/standalone/standalone.service.ts
```diff
@@ -14,7 +25,35 @@ export function createStandaloneHttp(mockBackend: MockBackend, options: BaseRequ
...
+          performance: 2,
+          claimPerformance: [
+            {id: 0, performance: 0},
+            {id: 1, performance: 1},
+            {id: 2, performance: 2},
+            {id: 3, performance: 1}
+          ]
+        },
+        group: mock_group,
+        student: mock_student
```
agorina: Do you need/use on the report: complete, valid and standardAdministration?

And just FYI: for now the accommodations will not be returned.

esotrope: ok interesting

esotrope:
@bgsavage I think for now I just want to keep it all in once place so that Jessica just has one file to edit. But later we can discuss splitting it up. The other benefit to keeping it together is we can have all our variables in one place and change them in one place to affect all the stylesheets

src/main/webapp/src/app/standalone/standalone.service.ts
esotrope: BTW claimPerformance object IDs are intended as claim codes - used to associate claim names/descriptions/icons.
I think there is a set of claim types and depending on the grade different claims are present/tested.

bgsavage:
Ok sounds good, yeah I think we need to discuss an overall CSS strategy at some point.

I've used BEM( http://getbem.com/ )in the past which has worked nicely to prevent duplication and has made it easy to find what you're looking for.

PR #48, opened 2017-04-05
Feature/dwr 236

bgsavage:
Incremental PR for parsing the API.

I'm using TDD to facilitate development of this feature which is the reason I'm also checking in test data files. 

Also I'm aware that currently my UI data.ts models do not line up with my API models.  While I continue to work on my API models and have that finalized, I'll go back and fix the UI models.


src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+            return item;
+        } catch (IOException e) {
+            e.printStackTrace();
+        } catch (SAXException e) {
+            e.printStackTrace();
+        } catch (XPathExpressionException e) {
+            e.printStackTrace();
+        } catch (ParserConfigurationException e) {
+            e.printStackTrace();
+        }
```
bgsavage: Let me know if there's something different I should be doing for exception handling here.

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public String getTemplate() {
+        return template;
+    }
+
```
bgsavage: I understand why we want to use the Builder pattern for modes, (for immutable objects), but my concern is this will be hard to maintain as properties change and evolve over time.  Is there an easy way to keep the builder in sync with the model?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
esotrope: services generally will wrap native exceptions with a service-specific exception and then declare that exception in a throws clause. And then the logging will eventually be done centrally in a catch all exception handler or method advice. Actually, in enterprise Java it is frowned upon to use print stack trace at all. I can show you an example of all this.

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
esotrope: I suggest we don't use builders unless it's automated for us. Because like you said, it's hard to maintain and sometimes encourages introducing business logic into models which is not what we want in a spring app.

esotrope: Also, models that will be stored in collections require both a hashCode and equals method implementation (intellij will do this for you if you right click on the class name and say generate > hashcode/equals) and select the properties on which they should be based

src/main/java/org/opentestsystem/rdw/reporting/model/Item.java
```diff
@@ -0,0 +1,41 @@
+package org.opentestsystem.rdw.reporting.model;
+import java.util.List;
+
+public class Item {
+    private String id;
+    private List<Rubric> rubrics;
+
+    public Builder builder(){
```
malaffoon: This should be static.

Although there can be a non-static method, that typically copies the existing content. Instead of that we've been using a copy() method in the builder itself.

src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -330,3 +330,81 @@ export const mock_item_results = [
 
 mock_item.results = mock_item_results;
 
+export const mock_rubrics = [
+  {
+  itemId: 1,
+  list: [{
```
esotrope: If the content of a collection is all of type Rubric I suggest we call the collection property "rubrics"
And for rubric instances we should probably rename "val" to "html" because that is what its holding

src/main/java/org/opentestsystem/rdw/reporting/model/Item.java
```diff
@@ -0,0 +1,41 @@
...
+
+        public Item build() {
+            final Item item = new Item();
+            item.id = id;
+            item.rubrics = rubrics;
+            return item;
+        }
+
+        public Builder rubrics(final List<Rubric> rubrics) {
+            this.rubrics = rubrics;
```
malaffoon: Somewhere in the call-stack, the collection should be copied into a new instance. The way it is now, a client could get their collection into the final, immutable object ... which is bad. I think the easiest and most flexible approach would be to accept an Iterable or Collection and then copying it into a new list here. e.g.
```java
public Builder rubrics(final Collection<Rubric> rubrics) {
    this.rubrics = newArrayList(rubrics);
    return this;
}
```

src/main/webapp/src/app/standalone/data/data.ts
bgsavage: Yep I agree.  I called it Rubrics on the API.  Also, I called 'val' something different on the API as well (template).  I'm still working on the API models so I don't want to go back and fix the mock data until I've finalized the API model.

src/main/java/org/opentestsystem/rdw/reporting/model/Item.java
bgsavage: Ah that makes sense.  Thanks

src/main/webapp/src/app/standalone/data/data.ts
esotrope: gotcha - thanks

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
bgsavage: I'll look into creating a general service exception to throw here instead.  I wasn't sure what to do here initially so I just took the IDE suggestion.  Thanks!

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
malaffoon: Not having a builder leads to mutable objects which leads to problems. So, overall the goal is to not have setters or only have protected setters. 

It takes 45 seconds to add a new property vs. 15 seconds without a builder. Business logic can get in regardless. That said ...

A situation where a builder is not needed is if the objects are instantiated only using frameworks. For example, if this object exists solely to be instantiated by the database layer and marshaled by the MVC framework, then i concur that a builder isn't necessary.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
bgsavage: Should I put the native exception into an innerException (is that a thing in Java?) of the service exception I throw?

malaffoon: Concur.

Also, if the handling of the different exceptions is the same, please condense them into a single catch. This can be done using exception hierarchy or catch with multiple exceptions.

malaffoon: Don't create a new Exception class unless it adds value (i.e. stores additional data). Just throw exceptions and the higher levels will deal with them.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+        builder.setNamespaceAware(true);
+
+        InputStream fileStream = artifactRepository.getItemXml(id);
+        return builder
+            .newDocumentBuilder()
+            .parse(fileStream);
+    }
+
+    private List<Rubric> GetRubrics(XPath xPath, NodeList nodes)
+    {
```
malaffoon: nit: formatting of brace

esotrope: pick: capitalization of first character of method name X P - java standard is to start lowercase  unlike .net

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
bgsavage: So it's ok to just throw the native exceptions then?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
bgsavage: oh darnit should be same line right?  Sorry I'm trying not to let my C#isms show!

src/main/java/org/opentestsystem/rdw/reporting/service/ItemService.java
```diff
@@ -0,0 +1,6 @@
+package org.opentestsystem.rdw.reporting.service;
+import org.opentestsystem.rdw.reporting.model.Item;
+
+public interface ItemService {
```
malaffoon: `Item` and `ItemService` are very vague terms. Is there something more specific we can call these?

esotrope: Probably ExamItem* would fit the rest of the system

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+import org.opentestsystem.rdw.reporting.repository.ArtifactRepository;
+import org.opentestsystem.rdw.reporting.service.ItemService;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.xpath.*;
```
malaffoon: We haven't established code style about this, and i don't even know if i can justify myself but i'm used to not using wildcards in imports. You can configure IntelliJ to automatically import just what you need instead of doing the * (Preferences, Editor, Code Style, Java, Imports).

src/main/java/org/opentestsystem/rdw/reporting/service/ItemService.java
bgsavage: I was thinking this would eventually populate the full Item model.  Though I could breakout my specific Rubric xml parsing into a separate RubricService.

Item in this context of the business domain is a test question.  I was thinking I should stick to the domain term, yeah?  Otherwise it could be AssessmentQuestion?  Or?  I'm not as familiar with the business domain yet, is there something better you could suggest?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
esotrope: i agree - java static code analyzers like it written out - usually it makes code analysis go better

bgsavage: I've just been hitting option enter key to generate all of my imports, but I'll keep a look out for these now that I know it can generate wildcards too.  Thanks.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
malaffoon: Unchecked / runtime exceptions ... yes.

Checked exceptions (e.g. IOException) should be mapped to runtime exceptions.

src/main/java/org/opentestsystem/rdw/reporting/service/ItemService.java
malaffoon: Okay, if `Item` is a real business object, then leave it and add javadocs to explain it clearly.

bgsavage: I like @esotrope 's suggestion of ExamItem

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    private ArtifactRepository artifactRepository;
+
+    public ItemServiceImpl(ArtifactRepository artifactRepository) {
+        this.artifactRepository = artifactRepository;
+    }
+
+    @Override
+    public Item getById(String id) {
+        try {
+            Document xmlDoc = GetXmlDocument(id);
```
malaffoon: There's a lot of expensive one-time set up going on here. You should probably do that once during initialization. At least the thread-safe stuff. I think you can create and save the DocumentBuilderFactory and the XPathFactory.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ArtifactRepositoryStub.java
```diff
@@ -0,0 +1,29 @@
+package org.opentestsystem.rdw.reporting.services.impl;
+
+import org.apache.commons.lang.StringUtils;
+import org.opentestsystem.rdw.reporting.repository.ArtifactRepository;
+
+import java.io.*;
+import java.nio.charset.StandardCharsets;
+import java.text.MessageFormat;
+
+public class ArtifactRepositoryStub implements ArtifactRepository {
```
malaffoon: We should talk about all this.

At the very least, you want to use resource loading during tests, not file-based.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ItemServiceImplIT.java
```diff
@@ -0,0 +1,38 @@
...
+    private ItemService service;
+
+    @Before
+    public void before() {
+        repository = new ArtifactRepositoryStub();
+        service = new ItemServiceImpl(repository);
+    }
+
+    @Test
+    public void getRubric_returnsRubric_fromStringConstant() throws Exception {
```
malaffoon: No underscores. How about getRubricShouldReturnRubricFromStringConstant.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.xpath.*;
+import java.io.*;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class ItemServiceImpl implements ItemService {
```
malaffoon: Remove public.

In general, you want to reduce scope as much as possible.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    private ArtifactRepository artifactRepository;
+
+    public ItemServiceImpl(ArtifactRepository artifactRepository) {
```
malaffoon: I assume this will be @Autowire eventually?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
bgsavage: I can wrap these in a factory and return an existing instance if one is available.  I would need more time to research how to keep the factory I create thread safe.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ItemServiceImplIT.java
bgsavage: I was following the pattern I found in Translation test.  I'll take note of this established style and change both places.

src/main/java/org/opentestsystem/rdw/reporting/repository/ArtifactRepository.java
```diff
@@ -0,0 +1,12 @@
+package org.opentestsystem.rdw.reporting.repository;
+
+import java.io.*;
+
+/**
+ * Repository for artifact retrieval.
+ */
+public interface ArtifactRepository {
```
malaffoon: In non-test, what is this going to be doing? Pulling from the IRIS system?

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ArtifactRepositoryStub.java
bgsavage: Could I load the entire directory as a resource?  I'll be needing to make subsequent calls to the directory to get any image files I find in the template html that has an

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
bgsavage: Yep once I start using this from a controller I'll use @Autowire resolve all of my dependencies.

src/main/java/org/opentestsystem/rdw/reporting/repository/ArtifactRepository.java
bgsavage: This will be going to S3 in prod.

esotrope: Is the design to connect to the raw rubrics in the data warehouse delivered in test packages?
I'm curious why we will be talking to a file system as opposed to the data mart

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ItemServiceImpl.java
bgsavage: How will I access it from my integration test?

src/main/java/org/opentestsystem/rdw/reporting/repository/ArtifactRepository.java
bgsavage: Yeah it's coming from the item xml in the item review packages.  Will this data be available in the data mart?

esotrope: not sure, we can maybe talk with the team about what is best. If we are storing large amounts of text/html or binaries we usually want to use a file system I think.

agorina: The original plan was to not load this data into data mart. I am curious how TDS does it. @bgsavage , do you know?

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
agorina: +1 for @malaffoon comment.
Additionally, if the property changes - the code needs to be changed regardless of the Builder. IntelliJ is very helpful with that.

bgsavage: I think I saw that IntelliJ does the refactor > rename nicely on all the fields, but I don't think it changes the method names on the builder does it?  Maybe I'm missing a setting or plugin for that.

src/main/java/org/opentestsystem/rdw/reporting/repository/ArtifactRepository.java
jtreuting: TDS reads from disk when it assessing item XML (each time an item is served up to a student).  Since we are going to Docker and Kubernetes, we are moving from file system to S3.

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
jtreuting: TDS team is using this IntelliJ plugin for Builders: https://github.com/mjedynak/builder-generator-idea-plugin

agorina: TDS team - thank you for sharing ! :)

PR #49, opened 2017-04-05
Feature/dwr 236

bgsavage:
If img tags are found in the rubric template html, then I'm going back to the repo to retrieve and encode these images so that the template html can be self contained and rendered without any external dependencies.

In my mock UI data, I replaced one of my sample htmls with html generated by the below ExamItemService to see that it's working in a browser.


src/test/java/org/opentestsystem/rdw/reporting/services/impl/ItemServiceImplIT.java
```diff
@@ -25,14 +25,14 @@ public void getRubricShouldReturnRubicLoadedFromStringConstant() throws Exceptio
...
     }
 
     @Test
-    public void getRubricShouldReturnsRubricWhenLoadedFromFile() throws Exception {
+    public void getRubricShouldReturnsRubricWhenLoadedFromFileAndEncodeImages() throws Exception {
         Rubric rubric = service.getById("187-1889").getRubrics().get(0);
 
         assertThat(rubric.getName(), equalTo("\n        Rubric "));
-        assertThat(rubric.getTemplate().length(), equalTo(1600));
+        assertThat(rubric.getTemplate().length(), equalTo(7374));
```
esotrope: maybe add an assert that 'src="data:' text exists

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -20,10 +22,14 @@
...
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+// TODO: Implement authorization
 public class ExamItemServiceImpl implements ExamItemService {
     private ArtifactRepository artifactRepository;
     private static final Logger logger = LoggerFactory.getLogger(ExamItemServiceImpl.class);
```
esotrope: eventually we'll switch to slf4j - just headsup

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -63,7 +69,7 @@ public ExamItem getById(String id) {
         }
     }
 
-    private Document GetXmlDocument(String id) throws IOException, SAXException, ParserConfigurationException {
+    private Document getXmlDocument(String id) throws IOException, SAXException, ParserConfigurationException {
         DocumentBuilderFactory builder = DocumentBuilderFactory.newInstance();
```
esotrope: I think this can be a flyweight or class instance level (private final ...)

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -77,23 +83,43 @@ private Document GetXmlDocument(String id) throws IOException, SAXException, Par
...
+            .map(node -> getRubric(xPath, node, id))
             .collect(Collectors.toList());
     }
 
-    private Rubric getRubric(XPath xPath, Node node){
+    private Rubric getRubric(XPath xPath, Node node, String id){
         try {
             return new Rubric()
                 .builder()
                 .name(xPath.evaluate("./name", node))
```
esotrope: Not sure if its more performant but I think xpath expressions can be compiled ahead of time with xpath.compile("/mypath") and then used like expression.evaluate(node)

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -77,23 +83,43 @@ private Document GetXmlDocument(String id) throws IOException, SAXException, Par
...
     }
+
+    private String encodeImages(String html, String id) throws IOException {
+        String imgRegex = "<img[^>]+src\\s*=\\s*['\"]([^'\"]+)['\"][^>]*>";
+        String result = html;
+
+        Matcher matcher = Pattern.compile(imgRegex).matcher(html);
+
+        while (matcher.find()) {
+            String imgSrc = matcher.group(1);
```
esotrope: nit picky, but my brain's style checker throws a warning when it sees imgSrc right next to imageStream, in general I suggest we always spell out words even when they are well known abbreviations. makes it easier on the head to read, less abbreviation unfolding.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ItemServiceImplIT.java
bgsavage: Good idea

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: Yeah, you're right.  I guess I was just thinking I was just sticking to the html img tag and src attribute.  But this is actually just a file name so I think I'll actually just go with imageFileName.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
malaffoon: 👍

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -1,5 +1,7 @@
 package org.opentestsystem.rdw.reporting.service.impl;
 
+import com.google.common.io.ByteStreams;
+import org.apache.xerces.impl.dv.util.Base64;
```
malaffoon: There is probably a better Base64 choice than this. java.util.Base64 exists, so i'd start there.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -77,23 +83,43 @@ private Document GetXmlDocument(String id) throws IOException, SAXException, Par
...
+        String result = html;
+
+        Matcher matcher = Pattern.compile(imgRegex).matcher(html);
+
+        while (matcher.find()) {
+            String imgSrc = matcher.group(1);
+            InputStream imageStream = artifactRepository.getArtifactFromItemDir(id, imgSrc);
+            String encoding = Base64.encode(ByteStreams.toByteArray(imageStream));
+
+            String mimeType = URLConnection.guessContentTypeFromName(imgSrc);
```
malaffoon: Ouch, really? There isn't any metadata telling us what the image is?

bgsavage: I can look at the bytes or the stream, but I think the end result could have the same problems.  All I have is what's in the  tag.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: The resulting code will end up being more verbose.  Is there something wrong with this one?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
esotrope: if we decide on embedding the images do we want to preprocess these rubric html to speed this up? Not sure how it would come into play, just something to think about

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: Actually, it's not that much before if you trust java.utils.Base64 more.

```
            String encoding = Base64
                .getEncoder()
                .encodeToString(ByteStreams.toByteArray(imageStream));
```

PR #50, opened 2017-04-05
Moved the rubric xPathing code to the GetRubrics method so that the e…

bgsavage:
…xceptions can be handled better.  Added a getXmlValue method which can safely parse the xPath and return an empty string (instead of throwing exception) if a value is missing.  Changed the Base64 encoder to java.utli.Base64.  Added additional integration test asserts scenarios when there is multiple rubrics and when there is none.

### DIFF
--- a/src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
+++ b/src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
@@ -1,7 +1,6 @@
 package org.opentestsystem.rdw.reporting.service.impl;
 
 import com.google.common.io.ByteStreams;
-import org.apache.xerces.impl.dv.util.Base64;
 import org.opentestsystem.rdw.reporting.model.ExamItem;
 import org.opentestsystem.rdw.reporting.model.Rubric;
 import org.opentestsystem.rdw.reporting.repository.ArtifactRepository;
@@ -17,12 +16,13 @@ import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathConstants;
-import javax.xml.xpath.XPathException;
 import javax.xml.xpath.XPathFactory;
 import javax.xml.xpath.XPathExpressionException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URLConnection;
+import java.util.ArrayList;
+import java.util.Base64;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -83,40 +77,72 @@ private Document getXmlDocument(String id) throws IOException, SAXException, Par
...
-            .mapToObj(nodes::item)
-            .map(node -> getRubric(xPath, node, id))
-            .collect(Collectors.toList());
+    private List<Rubric> getRubrics(XPath xPath, Document xmlDoc, String id){
+        try {
+            Node rubriclistNode = (Node) xPath
+                .compile("/itemrelease/item/content/rubriclist")
+                .evaluate(xmlDoc, XPathConstants.NODE);
+
+            if(rubriclistNode == null)
```
malaffoon: nit: again, we don't have coding standards but i feel that blocks like this should either be one line or have braces:
```java
    if (rubriclistNode == null) return new ArrayList<Rubric>();
OR
    if (rubriclistNode == null) {
        return new ArrayList<Rubric>();
    }
```

The main justification is the "oops" that happens when somebody modifies code like this, e.g.:
```java
    if(rubriclistNode == null)
        log.debug("rubriclist not found!");
        return new ArrayList<Rubric>();
```

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -83,40 +77,72 @@ private Document getXmlDocument(String id) throws IOException, SAXException, Par
...
+                return new ArrayList<Rubric>();
```
malaffoon: Guava has some handy collection helpers, in this case `newArrayList()`. Don't have to use them, just letting you know about them.

malaffoon:
Per Landon's comment in an earlier review, i'd like to have a data-flow / architecture chat about this. If we're going to be doing processing of the data every time a user goes to a screen, should we be pre-processing the data?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -47,23 +47,17 @@ public ExamItem getById(String id) {
...
             ExamItem examItem = new ExamItem()
                 .builder()
                 .id(id)
-                .rubrics(getRubrics(xPath, nodes, id)                )
+                .rubrics(getRubrics(xPath, xmlDoc, id)                )
                 .build();
 
             return examItem;
         } catch (IOException e) {
             logger.warn(e.getMessage(), e);
```
malaffoon: Logging with the exception message and the exception is a bit redundant; it would be better to use the message to give context to the exception, e.g. `logger.warn("Error reading item rubric", e)`.

When instantiating the RuntimeException, you should pass the caught exception in as the cause, e.g. `throw new RuntimeException("yadda yadda", e)`

I'd like to discuss with people what they think about logging and throwing, perhaps a parking lot chat some day.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: I've always preferred to keep the if statement separate than the code being executed so that code can be nicely read going down.  Same reason why we do method as one line per method.

Also, it'd be easy to confuse what's exactly in the if block if there is a line directly below it.  
    if(rubriclistNode == null) log.debug("rubriclist not found!");
        return new ArrayList<Rubric>();
"oops, now we are always returning a new ArrayList regardless of the if condition.".

I normally don't like to add braces for one line if statements because it adds too much to the code, but would prefer that over the code being executed being on the same line as the if condition.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: Thanks!

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: Good catch, I'll fix this on the next PR.

bgsavage: Yeah now I'm wondering if we should be logging and warning.  I'm sure it'll get logged again when we throw.

PR #51, opened 2017-04-06
Added S3ArtifactRepository which retrieves the documents from AmazonS…

bgsavage:
…3.  Updated a couple of exceptions.  Renamed ItemServiceImplIT to the new name of ExamItemSErviceImplIT and added two S3 integration tests.


src/main/java/org/opentestsystem/rdw/reporting/controller/ExamItemController.java
```diff
@@ -0,0 +1,20 @@
...
+
+@RestController
+@RequestMapping("/api/examitems")
+public class ExamItemController {
+    @Autowired
+    private ExamItemService service;
+
+    @GetMapping(value = "/{id}")
+    @ResponseStatus(HttpStatus.OK)
+    public ExamItem get(@PathVariable final String id) {
```
agorina: Just trying to understand: why do we return a generic ExamItem? Do we expect to load all Item related data at once. Should it be something more specific to Rubric/Exemplar.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/S3ArtifactRepository.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.text.MessageFormat;
+
+@Repository
+public class S3ArtifactRepository implements ArtifactRepository{
+    private AmazonS3 s3Client;
+    private S3Properties properties;
+
+    public S3ArtifactRepository(AmazonS3 s3Client, S3Properties properties) {
```
agorina: How does it work without the Autowired here? 
I trust that you tested it, just curious how it works.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ArtifactRepositoryStub.java
```diff
@@ -2,11 +2,13 @@
...
 import org.apache.commons.lang.StringUtils;
 import org.opentestsystem.rdw.reporting.repository.ArtifactRepository;
+import org.springframework.beans.factory.annotation.Qualifier;
 
 import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 
+@Qualifier("stub")
 public class ArtifactRepositoryStub implements ArtifactRepository {
```
agorina: Nit: comments explaining the purpose of it would be helpful. Is it for the case when we do not want to access S3?

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplIT.java
```diff
@@ -0,0 +1,97 @@
...
+    private ExamItemServiceImpl serviceUsingS3;
+
+    private final String AmazonEndpoint = "http://s3-us-west-2.amazonaws.com";
+    private final String AccessKey= ""; // TODO: Set your access key.
+    private final String SecretKey= ""; // TODO: Set your secret key.
+
+    @Before
+    public void before() {
+        // These will be autowired by spring boot.  I'm manually constructing these here
+        // for a faster test run.
```
agorina: Could you please clarify why it is faster?

bgsavage:
I had actually forgotten I already had an open pull request when I committed.  I actually wanted to take a second pass this morning (with a refreshed mind) and clean up my code a little before submitting for review.   Thanks for reviewing Alla!

src/main/java/org/opentestsystem/rdw/reporting/controller/ExamItemController.java
bgsavage: Right now I'm only returning a List of rubrics.  But soon I will adding a list of exemplars as well.  I'm thinking any info we could want from the Item xml should be added to the ExamItem object.  If later we find that the information we want is different in different contexts, we can look at a creating different models and would have better context for naming.  For right now, I was avoiding creating an object specific to just Rubric/Exemplar because that's not never extendable (I'd like to keep to the open/closed principle on this one.)

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/S3ArtifactRepository.java
bgsavage: Is it because I annotated the class with `@Repository` ? I'm just now learning how spring dependency injection works so I can't really say in this case.  I can add `@Autowired` for consistency, but if it's already being wired, should I?

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ArtifactRepositoryStub.java
bgsavage: I was just trying to troubleshoot why my ArtifactRepository wasn't being resolved at one point.  I will be removing this Qualifier.

I can add a comment to the top of the file explaining.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplIT.java
bgsavage: Loading up all of spring boot versus not having to load up any spring boot for this test.  In this case I'm not loading spring boot at all and found the test run was much faster.  I can provide specific numbers if you're interested.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExamItemController.java
agorina: Thank you for explaining. My only concern is if we decide to return all Item related data, while a user only looks at one of them.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplIT.java
agorina: My concern is that it makes it not a real "integration test". There are other ways to achieve what you want, I think. Let's chat offline.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/S3ArtifactRepository.java
agorina: @malaffoon, reaching our for your Spring experitize. @bgsavage and I chatted and it works without the 'Autowired' annotation here. I am curious to understand how. Could you please help?

build.gradle
```diff
@@ -51,6 +52,11 @@ repositories {
     mavenCentral()
 }
 
+dependencyManagement {
+    imports {
+        mavenBom 'com.amazonaws:aws-java-sdk-bom:1.10.77'
```
malaffoon: We've been using spring starters for everything else, i think you should use it for this as well. You can see an example in RDW_Ingest.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/S3ArtifactRepository.java
malaffoon: Yeah, this was added recently (last year?) ... if you care, search for "implicit constructor injection".

Personally i dislike implicitly relying on it, but ... meh, spring seems okay with it.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/AmazonS3Configuration.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Configuration
+public class AmazonS3Configuration {
+    @Autowired
+    private S3Properties s3Properties;
+
+    @Bean
+    public AmazonS3 amazonS3Client() {
+        AmazonS3 s3Client = new AmazonS3Client(new BasicAWSCredentials(s3Properties.getAccessKey(), s3Properties.getSecretKey()));
+        s3Client.setEndpoint(s3Properties.getEndpoint());
```
malaffoon: Somehow, this shouldn't be necessary. In RDW_Ingest, using AmazonS3 with spring starter, this just happens from the properties.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -55,11 +60,15 @@ public ExamItem getById(String id) {
 
             return examItem;
         } catch (IOException e) {
-            logger.warn(e.getMessage(), e);
-            throw new RuntimeException("Error reading the item xml for id: " + id);
+            String message = "Error reading the item xml for id: " + id;
+
+            logger.warn(message, e);
```
malaffoon: log-and-throw antipattern?

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/S3Properties.java
```diff
@@ -0,0 +1,63 @@
+package org.opentestsystem.rdw.reporting.repository.impl.s3;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties(prefix = "s3")
+public class S3Properties {
```
malaffoon: Again, somehow this shouldn't be necessary, at least the S3 part of it. There is already a built-in properties class that has these properties.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/S3ArtifactRepository.java
agorina: Oh, thank you! More and more 'black magic' :)

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/s3/S3Properties.java
```diff
@@ -0,0 +1,63 @@
...
+    private String endpoint;
+    private String bucket;
+    private String accessKey;
+    private String secretKey;
+    private String itemDirectory;
```
malaffoon: It seems like these two (itemDirectory, itemXml) are being munged into the properties of the backing store. If you had a local file-based implementation of the service, would these two settings be needed separate from the S3-specific settings?

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplIT.java
```diff
@@ -0,0 +1,97 @@
...
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+public class ExamItemServiceImplIT {
+    private ExamItemServiceImpl serviceUsingLocal;
+    private ExamItemServiceImpl serviceUsingS3;
+
+    private final String AmazonEndpoint = "http://s3-us-west-2.amazonaws.com";
+    private final String AccessKey= ""; // TODO: Set your access key.
```
malaffoon: So these tests will fail if it is checked in this way?

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplIT.java
```diff
@@ -0,0 +1,97 @@
...
+import org.opentestsystem.rdw.reporting.service.impl.ExamItemServiceImpl;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+public class ExamItemServiceImplIT {
+    private ExamItemServiceImpl serviceUsingLocal;
+    private ExamItemServiceImpl serviceUsingS3;
+
+    private final String AmazonEndpoint = "http://s3-us-west-2.amazonaws.com";
```
malaffoon: I feel that the endpoint should be specified simply with the s3:// protocol prefix. The detail that this may be implemented via http, https, or ftp shouldn't be a concern of the java code. That would be a configuration detail.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: Iterative. :)

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplIT.java
bgsavage: Nope.  I'm allowing the test to pass if I find the AccessKey haven't been set.  Not great to have conditional asserts.  If you let me keep the key in the .yml, I might just pull from there instead.  :)

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplIT.java
bgsavage: Yeah your s:// really threw me off for a long time because I thought that was an Amazon requirement for the bucket name, but it turned out to just be a Mark requirement :)

Once I found that out, I wasn't really a fan of the s3:// prefix.  We can discuss in the parking lot though.

malaffoon: We can discuss but "s3://" isn't a "Mark" thing, it is an official protocol prefix support by, at least, the Spring ecosystem.

agorina: https://gpdb.docs.pivotal.io/4390/admin_guide/load/topics/g-s3-protocol.html

bgsavage: So it's a greenplumb thing?  Are we using greenplum to facilitate our s3 calls then?

The reason I thought it was a Mark thing because I saw you were manually looking for it, and then manually stripping it to get the path.  We can discuss more but I feel it's a bit of a leaky implementation in that the path has to have knowledge of the vendor of where our data is stored.

bgsavage:
Please review latest commit

bgsavage:
Let me know where I can post the S3 version of the application.yml in case others are interested.

Or should I just update the existing internal location of the yml?

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
```diff
@@ -29,25 +29,25 @@ public boolean equals(Object o) {
 
         Rubric rubric = (Rubric) o;
 
-        if (scorepoint != rubric.scorepoint) return false;
+        if (scorepoint != null ? !scorepoint.equals(rubric.scorepoint) : rubric.scorepoint != null) return false;
```
malaffoon: Please use guava's Object.equals() helpers here.
Or just let IntelliJ generate your equals() method for you. 
Either approach will make it more consistent and readable.

bgsavage: I did generate through IntelliJ, does something look off?

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import java.text.MessageFormat;
+
+@Repository
+public class ArtifactRepositoryImpl implements ArtifactRepository{
+    private ArtifactProperties properties;
+    private ResourceLoader resourceLoader;
+
+    @Autowired
+    public ArtifactRepositoryImpl(ResourceLoader resourceLoader, ArtifactProperties artifactProperties)
+    {
```
malaffoon: nit

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -1,23 +1,23 @@
...
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
-import javax.xml.xpath.XPath;
-import javax.xml.xpath.XPathConstants;
-import javax.xml.xpath.XPathFactory;
-import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.*;
```
malaffoon: nit, prefer not using wildcard import; cmd-alt-O is your friend (assuming IntelliJ is configured properly)

src/main/resources/application.yml
```diff
@@ -11,4 +11,19 @@ spring:
   jackson:
     default-property-inclusion: non_null
     serialization:
-      WRITE_DATES_AS_TIMESTAMPS: false
\ No newline at end of file
+      WRITE_DATES_AS_TIMESTAMPS: false
+
+cloud:
+  aws:
+    credentials:
```
malaffoon: i don't think you need this blank entries; unless you're just putting them there as a sort of documentation thing?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: oh wow thanks!  I've been manually removing the unused and trying to find the wildcards sneaking in.

src/main/resources/application.yml
bgsavage: I was actually torn on this one.  It's good to see the what it should all look like together (so yes for documentation) but at the same time I don't want to encourage folks to add their Access / Secret key here and risk potentially checking those secrets in.  

What do you think?

PR #52, opened 2017-04-11
Feature/dwr 236

bgsavage:


src/main/java/org/opentestsystem/rdw/reporting/service/ExamItemService.java
```diff
@@ -1,6 +1,8 @@
 package org.opentestsystem.rdw.reporting.service;
-import org.opentestsystem.rdw.reporting.model.ExamItem;
+
+import org.opentestsystem.rdw.reporting.model.Rubric;
+import java.util.List;
 
 public interface ExamItemService {
-    ExamItem getById(String id);
+    List<Rubric> getByItemNumber(String id);
```
jtreuting: Based on the name of the method I would expect this to return an `ExamItem`.  Feels like the name should be something like `getRubricsByItemNumber()`

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -93,19 +91,47 @@ private Document getXmlDocument(String id) throws IOException, SAXException, Par
...
+            return Stream.concat(rubrics.stream(), samples.stream()).collect(Collectors.toList());
         } catch (XPathExpressionException e) {
-            logger.warn("Item xml contained a rubriclist element but no rubrics for id: " + id);
-            return new ArrayList<Rubric>();
+            logger.warn("Item xml contained a rubriclist element but no rubrics for id: " + id, e);
+            return new ArrayList<>();
+        }
+    }
+
+    private List<Rubric> getSamples(XPath xPath, String id, Node rubriclistNode) {
```
jtreuting: Samples are Exemplars, correct.  Returning a `Rubric` might get confusing since this is for the Exemplars.  Maybe the object name could be more generic if it's going to be used for both and are the same object structure.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExamItemController.java
```diff
@@ -1,20 +1,185 @@
...
+    public List<Rubric> get(@PathVariable final String itemNumber) {
+        return service.getByItemNumber(itemNumber);
+    }
+
+    @GetMapping(value = "")
     @ResponseStatus(HttpStatus.OK)
-    public ExamItem get(@PathVariable final String id) {
-        return service.getById(id);
+    public GetResult get(@RequestParam final int examId, @RequestParam final int studentId) {
+        List<ExamItem> result = new ArrayList<ExamItem>();
```
malaffoon: This is a *lot* of code for a controller. The bulk of this should be in either the service or an assembler.

malaffoon: Even though this is dummy data, the bulk of this code should be in the service, not the controller.

bgsavage: Yeah, I struggled with this too.  I also don't want to mix dummy data in with service code we intend to keep (i.e. ExamItemService).  I'll create a DummyDataService return from there.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: Yeah I can rename Rubric to something more generic.  I didn't want to fine tune as I understood the UI was starting to become a moving target so I opted to just get the pipeline established and we can sort out the object model once the requirements are more settled in and we know exactly which data points we want.  Right now for my POC purposes, the exemplar and rubrics both just return a Name and a Template html.

src/main/java/org/opentestsystem/rdw/reporting/service/ExamItemService.java
bgsavage: Ha, yeah I know.  i changed the name of this method and what I was returning so many times already.  I'll fix this as well.

PR #53, opened 2017-04-11
Get IT working

malaffoon:
I moved the data files so that is most of the "Files changed".


src/main/java/org/opentestsystem/rdw/reporting/Application.java
```diff
@@ -3,9 +3,11 @@
 import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Import;
 
 @SpringBootApplication
+@EnableConfigurationProperties
```
malaffoon: I'm not sure this is necessary, it was working fine without it. But SpringBootApplication does *not* do this by default. Odd.

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
```diff
@@ -27,19 +29,16 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
 
-        Rubric rubric = (Rubric) o;
+        final Rubric that = (Rubric) o;
 
-        if (scorepoint != null ? !scorepoint.equals(rubric.scorepoint) : rubric.scorepoint != null) return false;
-        if (name != null ? !name.equals(rubric.name) : rubric.name != null) return false;
-        return template != null ? template.equals(rubric.template) : rubric.template == null;
+        return Objects.equals(scorepoint, that.scorepoint)
```
malaffoon: @bgsavage  just wanted to show you an example of using Objects: makes code easier to read.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryImpl.java
```diff
@@ -10,28 +10,31 @@
 import java.text.MessageFormat;
```
malaffoon: just reformatting and final

src/main/resources/application.yml
```diff
@@ -23,7 +23,7 @@ cloud:
       static: us-west-2
 
 artifacts:
-  root: file:./
+  root: "classpath:"
```
malaffoon: Reading the files from the classpath, instead of coding a file path. Doing it this way allows the app to work when running standalone from the jar.

src/main/resources/data/irp/Item-187-1889/Item_1889_v12.gax
```diff
@@ -1,3 +1,3 @@
...
...>false</UpdateLayout><CanvasWidth>500</CanvasWidth><CanvasHeight>410</CanvasHeight><PaletteWidth>75</PaletteWidth><CenterImage>true</CenterImage><ProperLineGeometry>false</ProperLineGeometry><ScaleImage>false</ScaleImage></Options><Text /><ObjectMenuIcons><IconSpec><FileSpec>item_1889_v12_a_png16malpha.png</FileSpec><Label>3</Label></IconSpec><IconSpec><FileSpec>item_1889_v12_b_png16malpha.png</FileSpec><Label>7</Label></IconSpec><IconSpec><FileSpec>item_1889_v12_c_png16malpha.png</FileSpec><Label>21</Label></IconSpec><IconSpec><FileSpec>item_1889_v12_-a_png16malpha.png</FileSpec><Label>-3</Label></IconSpec><IconSpec><FileSpec>item_1889_v12_-b_png16malpha.png</FileSpec><Label>-7</Label></IconSpec><IconSpec><FileSpec>item_1889_v12_-c_png16malpha.png</FileSpec><Label>-21</Label></IconSpec></ObjectMenuIcons><ImageSpec><FileSpec>item_1889_v12_background_png16malpha.png</FileSpec><Position>0,0</Position></ImageSpec></QuestionPart><PreSetAnswerPart><AnswerSet><Question id="1889"><QuestionPart id="1"><ObjectSet /><SnapPoint>80@246,80;246,150;215,230;215,330;215,370;350,150;342,75;144,52;336,350;337,243;191,149</SnapPoint></QuestionPart></Question></AnswerSet></PreSetAnswerPart></Question>
```
malaffoon: Uhmm, not sure what i did here.

src/test/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryIT.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+@RunWith(SpringRunner.class)
+@EnableConfigurationProperties
```
malaffoon: This is the line it took me an hour to figure out ... sigh.

src/test/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryIT.java
```diff
@@ -0,0 +1,50 @@
...
+@SpringBootTest(classes = {
```
malaffoon: @bgsavage i experimented with using a hand-crafted context instead of using SpringBootTest. It was faster (1.5s vs 2.0s) but that didn't seem worth it for the ongoing maintenance it might cause.

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
bgsavage: sure, is this the preferred approach over using IntelliJ to generate? @esotrope , just want to make sure we're on the same page here.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryImpl.java
```diff
@@ -10,28 +10,31 @@
...
     @Autowired
-    public ArtifactRepositoryImpl(ResourceLoader resourceLoader, ArtifactProperties artifactProperties)
-    {
-        this.properties = artifactProperties;
+    public void setProperties(final ArtifactProperties properties) {
+        this.properties = properties;
+    }
+
+    @Autowired
+    public void setResourceLoader(final ResourceLoader resourceLoader) {
```
jtreuting: @malaffoon Any reason why you are favoring autowiring the setters vs the constructor?

malaffoon: No. During my hour-long venture into why this wasn't working, i was grasping at straws. I can revert this change.

src/main/resources/application.yml
bgsavage: I think once our secret .yml configurations are coming from a config service then having a local implementation will become a moot point, at which time this can probably either go away or move just the files required for the IT back to the test folder.

src/test/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryIT.java
bgsavage: yeah I agree, thanks!

src/test/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryIT.java
```diff
@@ -0,0 +1,50 @@
...
+        ArtifactProperties.class,
+        ArtifactRepositoryImpl.class,
+        YamlPropertiesConfigurator.class
+})
+public class ArtifactRepositoryIT {
```
bgsavage: Thanks for figuring this out!

src/main/java/org/opentestsystem/rdw/reporting/Application.java
esotrope: maybe it does now and prev ver didnt?

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
esotrope: Not sure we will be doing much sorting or object comparison but in general I favor the generated methods because (though ugly) they are technically more performant

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryImpl.java
esotrope: Yeah - I would prefer the constructor injection, constructor should declare all dependencies necessary for the object to perform its function and enforce their "satisfaction"

src/main/java/org/opentestsystem/rdw/reporting/model/Rubric.java
esotrope: more specifically - are any of these fields really usable to uniquely identify a rubric? do rubrics not have some kind of rubric-level non conflicting ID?

malaffoon: Perhaps i will host a beer-infused discussion of what equals/hashCode are for. If it is logical equivalence or business equivalence, then it makes more sense to create a custom comparator. I believe that equals/hashCode is a silly thing and i wish java had never introduced the idea -- given that it exists i think it should be a simple field-comparison as done here. And i challenge that, in most cases, you don't even need to implement it.

As for the generated method being more efficient, meh. Apart from the overhead of the method calls (which may be the performance hit you're alluding to?), the implementation of the helper method is as fast as the generated code. And maybe i'm just being obtuse, but i find the generated method to be really hard to read.

PR #54, opened 2017-04-12
fix test

malaffoon:
One of the tests was failing. After stepping through it was apparent that the test was out of date. 

When i was in there, i noticed that there was a bunch of XML in the java code. It was hard to read and it seemed more appropriate to put the XML in files. If you really hate this change, please let me know and i can try to convince you before reverting it.


src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -2,72 +2,69 @@
...
         assertThat(rubric.getTemplate().contains("src=\"data:image/png;base64,"), equalTo(true));
     }
 
     @Test
-    public void getByIdSWithMultiplesShouldReturnAll() {
-        List<Rubric> rubrics = service.getByItemNumber("187-2638");
-        assertThat(rubrics.size(), equalTo(5));
+    public void getByItemNumberWithMultiplesShouldReturnAll() {
+        List<Rubric> rubrics = service.getByItemNumber(multiItemId);
+        assertThat(rubrics.size(), equalTo(10));
```
malaffoon: This is the actual fix to the broken test; @bgsavage is this change right?

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -2,72 +2,69 @@
...
-        ArtifactRepository repository = mock(ArtifactRepository.class);
-
-        when(repository.getItemXml(simpleItem.id)).thenReturn(simpleItem.toInputStream());
-        when(repository.getItemXml(imageItem.id)).thenReturn(imageItem.toInputStream());
-        when(repository.getItemXml(multiItem.id)).thenReturn(multiItem.toInputStream());
-        when(repository.getItemXml(noRubricItem.id)).thenReturn(noRubricItem.toInputStream());
-
-        when(repository.getItemXml(noRubricItem.id)).thenReturn(new ByteArrayInputStream(noRubricItem.xml.getBytes(StandardCharsets.UTF_8)));
+        final ArtifactRepository repository = mock(ArtifactRepository.class);
+        when(repository.getItemXml(anyString())).thenAnswer(new Answer<InputStream>() {
```
malaffoon: We don't really need the TestItem class, we can just load the corresponding resource on demand.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
bgsavage: This is correct, I'm actually making more changes (regards to the naming as well with the model, etc) to these tests.  I was just going to include the fixed test in my next PR, I wasn't aware anyone else was running the unit tests (do we have this as part of a CI build yet?) right now.  I'll be more careful now that I know others are running unit tests to make sure all are passing prior to checking in.

bgsavage:
> When i was in there, i noticed that there was a bunch of XML in the java code. It was hard to read and it seemed more appropriate to put the XML in files. If you really hate this change, please let me know and i can try to convince you before reverting it.

My thoughts were that reaching to the file system would no longer constitute these tests as unit tests.  The xml wasn't really meant to be readable, just blobs of data that are required to run the tests.  Otherwise we should just go to the xml location already existing in the resources folder.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
bgsavage: Not anymore if you move the data out, yes.

bgsavage:
If it's alright with you, I'd prefer to hold off on this merge for now as there will be many conflicts with the model renaming I'm doing.

malaffoon:
It is accepted practice to use test resources in unit tests. Yes, the resources reside on the file system but you are not really reaching to the file system -- those are resources in the test jar. As a matter of fact when you want to do file-based tests (e.g. some local file system service) it is conventional to copy the data from resources into temporary files as part of the test setup.

I can hold off on merging the changes.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
malaffoon: We should make sure any checkin (esp. to `develop`) builds properly.

FYI, in this case i was building the project to work on moving the binary http protocol resolver into common. I needed to be able to build and run to make sure all the silly Spring injection stuff is working in all environments.

PR #55, opened 2017-04-12
use binary http protocol resolver from common

malaffoon:


src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -67,7 +67,7 @@ public void getByIdSShouldReturnsRubricWhenLoadedFromFileAndEncodeImages() {
     @Test
     public void getByIdSWithMultiplesShouldReturnAll() {
         List<Rubric> rubrics = service.getByItemNumber("187-2638");
-        assertThat(rubrics.size(), equalTo(5));
+        assertThat(rubrics.size(), equalTo(10));
```
malaffoon: needed to put this in to get through build process

PR #56, opened 2017-04-12
Feature/dwr 236

bgsavage:
The rubrics and exemplars are no longer merged into the same collection, but rather separate collections wrapped by ExamItemSolution model.  This model also composes the AnswerKey.

We only attempt to get the AnswerKey if the item type is MC, MS, or EBSR.  Otherwise, the AnswerKey attribute only contains garbage data.

Added an Exemplar tab.  Rubric and Exemplar are now shown in a table format.

UI and API are updated per these changes.  

More changes still pending around the corner cases / exception handling.


src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -50,50 +51,50 @@ public void before() throws IOException {
...
+        assertThat(scoringCriteria.getName(), equalTo(" Rubric"));
+        assertThat(scoringCriteria.getTemplate().contains("src=\"data:image/png;base64,"), equalTo(true));
     }
 
     @Test
     public void getByIdSWithMultiplesShouldReturnAll() {
-        List<Rubric> rubrics = service.getByItemNumber("187-2638");
-        assertThat(rubrics.size(), equalTo(10));
+        List<ScoringCriteria> scoringCriteria = service.getByItemNumber("187-2638").getRubrics();
+        assertThat(scoringCriteria.size(), equalTo(5));
```
bgsavage: Yes, we are back to 5 here again, not a mistake.  This unit test is only asserting on the rubrics.  I have a TODO to come back and add unit tests for exemplars and answer key.

src/main/java/org/opentestsystem/rdw/reporting/service/DummyDataService.java
```diff
@@ -148,6 +148,23 @@
...
+                .score(2)
+                .maximumScore(2)
+                .irisInfo(
+                        new IrisInfo()
+                                .builder()
+                                .vendorId(irisVendorId)
+                                .token("{\"passage\":{\"autoLoad\":\"false\"},\"items\":[{\"response\":\"<p>test</p>\",\"id\":\"I-187-214\"}],\"layout\":\"WAI\"}")
+                                .build()
+                )
+                .build()
```
esotrope: should we go ahead and move iris vendor ID to the client config and rename the item field to irisToken?

src/main/java/org/opentestsystem/rdw/reporting/model/ExamItemSolution.java
```diff
@@ -0,0 +1,66 @@
+package org.opentestsystem.rdw.reporting.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExamItemSolution {
```
jtreuting: Are you thinking this will hold the Item information details also, or is this only going to be for Rubric and Exemplars tab?

src/main/java/org/opentestsystem/rdw/reporting/service/DummyDataService.java
bgsavage: It's under field under IrisInfo, so I think "Iris"token might be smurfing / redundant.

I'm leaving the vendor ID as is as a constant until we circle back and implement the API building of this token.

src/main/java/org/opentestsystem/rdw/reporting/model/ExamItemSolution.java
bgsavage: If Item Information contains information about how to score the item, sure.  If not, then either create another object that composes ExamItemInfo and ExamItemSolution.  

If ItemInfo is something we can grab quickly with minimum perf impact, we might be able to grab it when we are populating ExamItem grid (top half of the screen) instead of on the select row click.  In that case, the ExamItemInfo can be composed on the ExamItem model.

src/main/java/org/opentestsystem/rdw/reporting/service/DummyDataService.java
esotrope: I was thinking if we remove the vendor ID the iris info is reduced to just the token, so we can remove the IrisInfo abstraction all together and replace it with a String irisToken field

bgsavage: I see that's a good point.  So we're certain then that vendor id is environment specific not specific / different for each item we're trying to render?

esotrope: based on my testing with it it is an optional field and probably just informs iris what app is accessing it for logging or something

src/main/java/org/opentestsystem/rdw/reporting/model/ExamItemSolution.java
jtreuting: I'm asking because of the name of this object.  I'm not sure it makes sense to me.  Does something like "ExamItemScoringInfo" or something like that make sense?  This is just focus on item scoring and not the rest of the data around an item.

src/main/java/org/opentestsystem/rdw/reporting/service/DummyDataService.java
```diff
@@ -148,6 +148,23 @@
...
 
+        examItems.add(new ExamItem()
+                .builder()
+                .number("187-214")
+                .claim("Problem Solving")
+                .target("Target I")
+                .score(2)
+                .maximumScore(2)
+                .irisInfo(
+                        new IrisInfo()
```
malaffoon: This should be just `IrisInfo.builder()`

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -82,35 +81,86 @@ private Document getXmlDocument(String id) throws IOException, SAXException, Par
...
-    private List<Rubric> getRubrics(XPath xPath, Document xmlDoc, String id){
+    private ExamItemSolution getExamItemSolution(XPath xPath, Document xmlDoc, String id){
         try {
             Node rubriclistNode = (Node) xPath
                 .compile("/itemrelease/item/content/rubriclist")
                 .evaluate(xmlDoc, XPathConstants.NODE);
 
-            if(rubriclistNode == null)
-                return new ArrayList<Rubric>();
+            return new ExamItemSolution()
```
malaffoon: ditto (`ExamItemSolution.builder()`)

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -82,35 +81,86 @@ private Document getXmlDocument(String id) throws IOException, SAXException, Par
...
+
+        try{
+            if(!itemFormatIsValidToHaveAnswerKey(xPath, xmlDoc))
+                return null;
+
+            Node answerKeyNode = (Node) xPath
+                .compile("/itemrelease/item/attriblist/attrib[@attid=\"itm_att_Answer Key\"]")
+                .evaluate(xmlDoc, XPathConstants.NODE);
+
+            if(answerKeyNode == null){
```
malaffoon: lots of whitespace issues, i'll assume you'll cmd-alt-L before checkin

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -50,50 +51,50 @@ public void before() throws IOException {
 
     @Test
     public void getByIdShouldReturnRubricLoadedFromStringConstant() {
-        Rubric rubric = service.getByItemNumber(simpleItem.id).get(0);
+        ScoringCriteria ScoringCriteria = service.getByItemNumber(simpleItem.id).getRubrics().get(0);
 
-        assertThat(rubric.getName(), equalTo("Rubric"));
-        assertThat(rubric.getTemplate().length(), equalTo(1603));
+        assertThat(ScoringCriteria.getName(), equalTo("Rubric"));
+        assertThat(ScoringCriteria.getTemplate().length(), equalTo(1603));
```
malaffoon: Pretty sure there is a `.hasLength(1603)` option for assertions

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -50,50 +51,50 @@ public void before() throws IOException {
...
 
     @Test
     public void getByIdSShouldReturnsRubricWhenLoadedFromFileAndEncodeImages() {
-        Rubric rubric = service.getByItemNumber("187-1889").get(0);
+        ScoringCriteria scoringCriteria = service.getByItemNumber("187-1889").getRubrics().get(0);
 
-        assertThat(rubric.getName(), equalTo(" Rubric"));
-        assertThat(rubric.getTemplate().contains("src=\"data:image/png;base64,"), equalTo(true));
+        assertThat(scoringCriteria.getName(), equalTo(" Rubric"));
+        assertThat(scoringCriteria.getTemplate().contains("src=\"data:image/png;base64,"), equalTo(true));
```
malaffoon: There is a `.contains(...)` for assertions

.gitignore
```diff
@@ -23,4 +23,5 @@ nbbuild/
 dist/
 nbdist/
 .nb-gradle/
-.gradle/
\ No newline at end of file
+.gradle/
+classes/
```
malaffoon: what is creating a `classes/` folder off of root?

src/main/java/org/opentestsystem/rdw/reporting/controller/ExamItemController.java
```diff
@@ -23,7 +24,7 @@
 
     @GetMapping(value = "/{itemNumber}/rubrics")
```
malaffoon: I know @jtreuting is iterating on naming of the class; i think the end-point should be renamed as well: if it isn't returning a list of rubrics then it should be something else, perhaps `/{itemNumber}/scoreinfo` or `/{itemNumber}/scoring`

src/main/java/org/opentestsystem/rdw/reporting/model/ScoringCriteria.java
```diff
@@ -2,8 +2,7 @@
 
 import java.util.Objects;
 
-public class Rubric {
-
+public class ScoringCriteria {
```
malaffoon: Yeah, yeah, it is the OCD in me but should this be `ScoringCriterium`? Or perhaps `ScoringTemplate`?

src/main/java/org/opentestsystem/rdw/reporting/controller/ExamItemController.java
bgsavage: Just trying to limit my refactoring and only change the route once we have a name everyone can agree on :)

src/main/java/org/opentestsystem/rdw/reporting/model/ScoringCriteria.java
bgsavage: Scoring One-day [bicycle race](https://www.google.com/webhp?sourceid=chrome-instant&rlz=1C5CHFA_enUS724US724&ion=1&espv=2&ie=UTF-8#q=criterium)? :) 

While I was getting groceries last night, I thought about ScoringGuide.  What do you all think of that?

.gitignore
bgsavage: I thought you would know.  I had a bunch of files from here after I brought in develop.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExamItemController.java
malaffoon: fair

src/main/java/org/opentestsystem/rdw/reporting/model/ScoringCriteria.java
malaffoon: ScoringGuide sounds good.

.gitignore
malaffoon: I don't have it locally. Maybe it is leftover from a previous configuration. Perhaps delete the folder and see if it comes back.

src/main/java/org/opentestsystem/rdw/reporting/model/ScoringCriteria.java
bgsavage: Sorry let me clarify.  

ScoringCriteria -> ScoringCriterion (? I think this is what you meant?)
ExamItemSolution -> ScoringGuide 

Thoughts?
@jtreuting  ?

.gitignore
bgsavage: Deleted and it came back when I ran the unit tests.

bgsavage: Doing ./gradlew build doesn't create the folder, but this command is super slow now (1 min +) so I've been using a Junit test run configuration.  I think that's what's actually creating the folder.

image

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
bgsavage: Like this? 

assertThat(scoringCriteria.getTemplate(), containsString("src=\"data:image/png;base64,"));

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
bgsavage: I didn't see a hasLength which takes an int, is there another one you're thinking of?

There's a hasLength (which is now deprecated) which takes a string, and it's just checking the string isn't null or empty

PR #57, opened 2017-04-13
pull resource data out of java

malaffoon:
This is just resubmitting my changes after Brandon's renaming work.


src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -13,109 +15,90 @@
...
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.charset.StandardCharsets;
 import java.util.List;
 
-import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.CoreMatchers.equalTo;
-import static org.junit.Assert.assertThat;
-import static org.mockito.Matchers.any;
+import static org.assertj.core.api.Assertions.assertThat;
```
malaffoon: So this is what was causing confusion in my comments in the last review. The junit assert package is fine but the core assertj package is a bit more java-y. Since this is what we're using in other modules, i'd prefer to use it here too. You'll see the differences in the code below.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -13,109 +15,90 @@
...
-        assertThat(ScoringCriterion.getTemplate().length(), equalTo(1603));
+        assertThat(scoringCriterion.getName()).isEqualTo("Rubric");
+        assertThat(scoringCriterion.getTemplate()).hasSize(1603);
     }
 
     @Test
-    public void getByIdSShouldReturnsRubricWhenLoadedFromFileAndEncodeImages() {
-        ScoringCriterion scoringCriterion = service.getScoringGuide("187-1889").getRubrics().get(0);
+    public void getScoringGuideShouldReturnsRubricWhenLoadedFromFileAndEncodeImages() {
+        ScoringCriterion scoringCriterion = service.getScoringGuide(imageItemId).getRubrics().get(0);
```
malaffoon: The assert functionality is identical of course, but the syntax is a bit different (IMO a bit easier to type and read).

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -13,109 +15,90 @@
...
+    private final String noRubricItemId = "187-2661";
 
     @Before
     public void before() throws IOException {
-        ArtifactRepository repository = mock(ArtifactRepository.class);
+        final ArtifactRepository repository = mock(ArtifactRepository.class);
+        when(repository.getItemXml(anyString())).thenAnswer(new Answer<InputStream>() {
+            @Override
+            public InputStream answer(final InvocationOnMock invocation) throws Throwable {
+                final String id = (String) invocation.getArguments()[0];
```
bgsavage: This is how you get the argument passed in Java?  There's no lambda expression to do this?

malaffoon: Yeah, the mock Answer stuff was added pre-lambda; i'm sure it could be made more java-y at this point.

PR #58, opened 2017-04-13
Improved exceptions handling.  Now throwing specific exceptions which…

bgsavage:
… will result in a 404 response when a resource is not available.  The helper method, getXmlValue now handles the XPathExpressionException and returns a default if an xPath is bad so that the callers no longer need to handle it.  In the case of an image can’t be found, now continuing to look for other images without bombing out the entire request.  From the ArtifactRepository, no longer leaking out the AmazonS3Exception when a resource cannot be found from S3, instead throwing an IOException.


bgsavage:
I think this is my final commit for the feature (finally) plus any code review comments from this PR :)

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/ArtifactRepositoryImpl.java
```diff
@@ -31,6 +32,12 @@ public InputStream getArtifactFromItemDirectory(final String id, final String ar
     }
 
     private InputStream getArtifact(final String fullyQualifiedPath) throws IOException {
-        return this.resourceLoader.getResource(properties.getRoot() + fullyQualifiedPath).getInputStream();
+        String rootAndPath = properties.getRoot() + fullyQualifiedPath;
+        try{
```
malaffoon: nit: whitespace

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -57,12 +59,12 @@ public ExamItemScoringGuide getScoringGuide(String itemNumber) {
             String message = "Error reading the item xml for item number: " + itemNumber;
 
             logger.warn(message, e);
```
malaffoon: log-and-throw?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -184,61 +185,65 @@ private boolean itemFormatIsValidToHaveAnswerKey(XPath xPath, Document xmlDoc) {
...
+            try {
+                InputStream imageStream = artifactRepository.getArtifactFromItemDirectory(itemNumber, imageFileName);
+                String encoding = Base64
+                        .getEncoder()
+                        .encodeToString(ByteStreams.toByteArray(imageStream));
+
+                String mimeType = URLConnection.guessContentTypeFromName(imageFileName);
+                result = result.replace(imageFileName, "data:" + mimeType + ";base64," + encoding);
+            } catch (IOException e) {
+                String message = MessageFormat.format("Unable to find referenced image({0}) for item number: {1}", imageFileName, itemNumber);
```
malaffoon: It is okay to catch and eat this exception with nothing more than a log message?
I believe this is a WARN level, not ERROR. The app will continue to run without any compromise in behavior.

I suggest that, when catching an exception like this, you do *not* want to emit the entire stack trace. And you can just inline this: logger knows how to format messages. The result would be something like:
```
    logger.warn("Unable to find referenced image({}) for item number {}: {}", imageFileName, itemNumber, e.getMessage());
```

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
```diff
@@ -4,12 +4,13 @@
...
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
+import org.opentestsystem.rdw.reporting.model.AnswerKey;
 import org.opentestsystem.rdw.reporting.model.ScoringCriterion;
 import org.opentestsystem.rdw.reporting.repository.ArtifactRepository;
 import org.opentestsystem.rdw.reporting.service.impl.ExamItemServiceImpl;
 
 import javax.imageio.ImageIO;
-import java.awt.Graphics;
+import java.awt.*;
```
malaffoon: IntelliJ was doing this to me to. I think there is another setting to tweak.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: Yeah this one I want to throw so that 404 is returned.  Should I not log?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: I wanted to log it as an error to draw attention to it in the logs.  This probably means we need to go investigate our item packages and find the image that should be there.  I can remove the stack trace if you think that will just create a bunch of noise.

Also good to know that I don't have to pre format the messages.  Just curious, don't I still need the param number {0} and {1} ?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
malaffoon: This is an "expected" exception if the item doesn't exist. No need to log, the NoSuchElementException being thrown will take care of things. If anybody cares enough to log, it'll happen at a higher level.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
bgsavage: Yessss I've manually set this back like 4 times already!  I've tried to find this setting and couldn't, let me know if you can.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
```diff
@@ -57,12 +59,12 @@ public ExamItemScoringGuide getScoringGuide(String itemNumber) {
...
 
             logger.warn(message, e);
-            throw new RuntimeException(message, e);
+            throw new NoSuchElementException(message);
         } catch (SAXException | ParserConfigurationException e) {
             String message = "Error parsing the item xml for item number: " + itemNumber;
 
-            logger.warn(message, e);
-            throw new RuntimeException(message, e);
+            logger.error(message, e);
```
malaffoon: This one is a little different: it really isn't expected. Since we're converting a parse error into a NoSuchElementException i think we should be logging here.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: I'm not sure that we should expect this exception to happen.  I would think we should have access to all of the items yeah?  So wouldn't we want to know when we don't?  I'm not sure if we'd ever want to log all 404s the api returns, but I would find this one as a more interesting one because it questions the integrity of our data (we got an item number from somewhere but we weren't able to resolve it against the IRP.)

malaffoon: This is a method on the service that is called based on what the user passes into the web end-point, right? As such, any user can stick whatever id they want there. If we want to distinguish between it-doesn't-exist and it-exists-but-we-can't-read-it, then more work would have to be done. 

I'm not going to complain a whole lot if you leave it, just pointing out the log-and-throw thing.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
malaffoon: You don't need to include the param number if you are just using the params in order. 

I agree that it should be logged with a high level, and we need to investigate the bad data. But ERROR (and FATAL) should be preserved for things that need immediate ops attention. I don't think this qualifies. But it is in that fuzzy gray area.

src/test/java/org/opentestsystem/rdw/reporting/services/impl/ExamItemServiceImplTest.java
malaffoon: FTR, Preferences, Editor, Code Style, Java, Imports, remove anything in "Packages to Use Import with '*'" list.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamItemServiceImpl.java
bgsavage: I'll downgrade the severity.  Thanks for the feedback, you're right this isn't a "wake us up out of bed" scenario :)

PR #59, opened 2017-04-14
Just an idea to share

agorina:
This is just to show an approach. I am not planning on checking this in.

I am not sure I like it myself. 

One one hand I like the "small and to the point" SQL, on the other hand it would be nice if we could have a common pattern for the security part of the WHERE clause.

### DIFF
--- a/src/main/java/org/opentestsystem/rdw/reporting/repository/SchoolRepository.java
+++ b/src/main/java/org/opentestsystem/rdw/reporting/repository/SchoolRepository.java
@@ -1,0 +1,10 @@
+package org.opentestsystem.rdw.reporting.repository;
+
+import java.util.List;
+import org.opentestsystem.rdw.reporting.model.Group;
+
+
+public interface SchoolRepository {
+
+    List<Integer> findAllIds(List<Integer> schoolIds, List<Integer> districtIds, boolean statewide);
+}
--- a/src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
+++ b/src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
@@ -1,0 +1,45 @@
+package org.opentestsystem.rdw.reporting.repository.impl;
+
+import java.util.List;
+import org.opentestsystem.rdw.reporting.repository.SchoolRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Lists.newArrayList;
+
+@Repository
+public class SchoolRepositoryImpl implements SchoolRepository {
+
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Value("${sql.school.findAll}")
+    private String findAllSql;
+
+    @Autowired
+    SchoolRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
+        this.jdbcTemplate = checkNotNull(jdbcTemplate);
+    }
+
+
+    @Override
+    public List<Integer> findAllIds(List<Integer> schoolIds, List<Integer> districtIds, boolean statewide) {
+        return jdbcTemplate.queryForList(findAllSql,
+                addAuthorizationRestrictions(new MapSqlParameterSource(), schoolIds, districtIds, statewide),
+                Integer.class);
+    }
+
+    private MapSqlParameterSource addAuthorizationRestrictions(MapSqlParameterSource param, List<Integer> schoolIds, List<Integer> districtIds, boolean statewide) {
+
+        final List<Integer> emptyList = newArrayList();
+        emptyList.add(-1);
+
+        return param.addValue("schoolIds", schoolIds.isEmpty() ? emptyList : schoolIds)
+                .addValue("districtIds", districtIds.isEmpty() ? emptyList : districtIds)
+                .addValue("stateWide", statewide ? 1 : 0);
+    }
+
+}
--- a/src/test/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepoIT.java
+++ b/src/test/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepoIT.java
@@ -1,0 +1,31 @@
+package org.opentestsystem.rdw.reporting.repository.impl;
+
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.opentestsystem.rdw.reporting.repository.SchoolRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.junit4.SpringRunner;
+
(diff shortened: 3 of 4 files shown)


src/main/java/org/opentestsystem/rdw/reporting/repository/SchoolRepository.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.reporting.repository;
+
+import java.util.List;
+import org.opentestsystem.rdw.reporting.model.Group;
+
+
+public interface SchoolRepository {
+
+    List<Integer> findAllIds(List<Integer> schoolIds, List<Integer> districtIds, boolean statewide);
+}
```
agorina: These three parameters would need to be wrapped into a class that represents a user security context for SQL.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+
+    @Override
+    public List<Integer> findAllIds(List<Integer> schoolIds, List<Integer> districtIds, boolean statewide) {
+        return jdbcTemplate.queryForList(findAllSql,
+                addAuthorizationRestrictions(new MapSqlParameterSource(), schoolIds, districtIds, statewide),
+                Integer.class);
+    }
+
+    private MapSqlParameterSource addAuthorizationRestrictions(MapSqlParameterSource param, List<Integer> schoolIds, List<Integer> districtIds, boolean statewide) {
```
agorina: This will be a common method.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+        final List<Integer> emptyList = newArrayList();
+        emptyList.add(-1);
```
agorina: yeap, a magic 'empty list'. This part I really do not like...

src/test/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepoIT.java
```diff
@@ -0,0 +1,31 @@
...
+    @Autowired
+    private SchoolRepository repository;
+
+    @Test
+    public void itShouldFindAllGroups() {
+        final List<Integer> schoolIds = newArrayList();
+        final List<Integer> districtIds = newArrayList();
+
+        assertThat(repository.findAllIds(schoolIds, districtIds, true)).containsExactly(1, 2);
+        assertThat(repository.findAllIds(schoolIds, districtIds, false)).isEmpty();
```
agorina: I had two schools and two districts on the sample data set.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
bgsavage: We probably wouldn't need to add a default value of -1.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
bgsavage: I wonder if there could be an AuthorizationRepository that appends these restrictions in a common place?

bgsavage: And other repos would implement it.  AuthorizationRepository could be an abstract class maybe?

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
agorina: Unless we dynamically add parts of the WHERE clause, I do not see how it could be avoided. Without it you will end up with  
```...id in ()```

bgsavage:
I like the idea, definitely something we can build on :)

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
bgsavage: Ah I see that's why you're adding it.  In that case, should we throw an exception or not even bother to making the SQL query that we know won't yield results?

agorina: This is a  valid case, for example - no school id, but some district ids. Or a state wide user. An exception is not what we want here.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
jtreuting: Another possibility is to create a `authorizationJdbcTemplate` that wraps `jdbcTemplate` and makes methods that accept the object that has the school and district IDs in it.  And it has the logic for the `addAuthorizationRestrictions`

malaffoon: Yeah, somehow we need to have a helper that is user-context aware. It could be an abstract repo, utility methods, or perhaps a wrapper around jdbcTemplate.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/SchoolRepositoryImpl.java
malaffoon: I'd want to benchmark the "IN (-1)" performance before committing to this approach.

malaffoon: @bgsavage you need it because there may be school ids but not district ids. If the user has neither school ids nor district ids, they won't get this far. Although we should code for that and short-circuit if both lists are empty.

agorina: @malaffoon - I agree about benchmarking "IN (-1)" if we want to adopt this idea.
Or we can just construct the WHERE clause in the java code..?

malaffoon: Constructing the WHERE clause in the java code will require that every single table uses the same column names; or it is configured in the helper for each repo. Reasonable, but fragile.

src/main/java/org/opentestsystem/rdw/reporting/repository/SchoolRepository.java
esotrope: Can we broaden the signature to accept Collection<Integer> I planned on making the security context represent the collections here as Sets since they are conceptually not redundant.

agorina: Yes, we can change it anyway we want :) This was just to show an idea.

esotrope: cool thanks 👍

agorina: Just to make sure we are on the same page - I am going to abandon this request. If we decide to continue with the suggested approach, I will be happy to profile "IN (-1)" with a large data set.

PR #60, opened 2017-04-14
Removed 'hateoas' dependencies.

agorina:


src/test/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryIT.java
```diff
@@ -1,16 +1,14 @@
...
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.jdbc.Sql;
-import org.springframework.test.context.junit4.SpringRunner;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@RunWith(SpringRunner.class)
-@RepositoryIT
+@ContextConfiguration(classes = GroupRepositoryImpl.class)
```
agorina: Unrelated - I noticed that I did not set it up correctly. Fixing.

src/test/java/org/opentestsystem/rdw/reporting/repository/impl/RepositoryIT.java
```diff
@@ -1,34 +1,29 @@
...
+ * test database.
  */
-@Target(ElementType.TYPE)
-@Retention(RetentionPolicy.RUNTIME)
+@RunWith(SpringRunner.class)
 @ContextConfiguration(classes = TestConfiguration.class)
 @SpringBootTest(classes = {
         DataSourceAutoConfiguration.class,
-        GroupRepositoryImpl.class,
         DataSourceConfiguration.class,
```
agorina: This is part of the 'unrelated' fix mentioned above.

src/test/java/org/opentestsystem/rdw/reporting/repository/impl/RepositoryIT.java
```diff
@@ -1,34 +1,29 @@
...
         YamlPropertiesConfigurator.class}
 )
 @Transactional
-public @interface RepositoryIT {
```
malaffoon: Hmm, that's a big change.

agorina: So there are two changes here: 1. I removed the GroupRepoImpl.class since this does not belong here. This change was needed.  2.  I change the approach to use this as a base class to avoid ```@RunWith(SpringRunner.class)``` on every Repo IT class.   Do you see any problems with it?

PR #61, opened 2017-04-18
Breadcrumbs are now driven by the app routes.

bgsavage:
Rewrote the breadcrumbs component so that it operates on navigation changes (navigation end event) and only needs to be placed in one spot on the app.component.html. Breadcrumbs now defined with the app routing.  Introduced [resolve]= syntax for defining breadcrumbs for when dynamic data values need to be shown.  Now resolving data through route resolves in some components rather than making the service calls in the component.  Removing breadcrumb logic and resolving data through the routes cleans up a lot of the subscription code that was previously required in the components.


src/main/webapp/src/app/breadcrumbs/breadcrumb.component.html
```diff
@@ -1,13 +0,0 @@
...
-  </span>
-</span>
-<span class="breadcrumb-link-container" *ngIf="!active">
-  <a class="breadcrumb-link" routerLink="{{crumb.path}}">
-    <i *ngIf="first" class="glyphicon glyphicon-home"></i>{{crumb.name}}
-  </a>
-  <span class="breadcrumb-link-separator">
-    <i class="glyphicon glyphicon-menu-right"></i>
-  </span>
-</span>
```
bgsavage: I deleted this component because I kept getting confused between what was a "breadcrumb" and what was a "breadcrumbs" :) .  I ended up just merging this template in with the breadcrumbs template.

src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.html
```diff
@@ -1,9 +1,28 @@
...
+          <div>
+
+            <span class="breadcrumb-link-container" >
+              <a class="breadcrumb-link" routerLink="/">
+                <i class="glyphicon glyphicon-home"></i>{{ 'labels.home' | translate }}
+              </a>
+                <span class="breadcrumb-link-separator">
+                  <i class="glyphicon glyphicon-menu-right"></i>
+                </span>
+            </span>
```
bgsavage: We always prefix with Home >.

src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -1,33 +1,67 @@
...
-  @Input()
-  public featureRoot: string;
+      let crumbLabel = child.snapshot.data[BreadcrumbsKeyword]
+      let routeURL: string = child.snapshot.url.map(segment => segment.path).join("/");
+      url += `/${routeURL}`;
 
+      if(crumbLabel.startsWith(ResolveKeyword)){
+        let resolveable = crumbLabel.replace(ResolveKeyword, "");
+        crumbLabel = Utils.getPropertyValue(resolveable, child.snapshot.data);
+      }
```
bgsavage: This is how I'm resolving dynamic data.

src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -1,33 +1,67 @@
...
+        url: url.replace(/\/$/, "")
+      };
+
+      let existing = breadcrumbs.find(x => x.url == breadcrumb.url);
+
+      if(existing)
+        existing.label = breadcrumb.label;
+      else
+        breadcrumbs.push(breadcrumb);
+
```
bgsavage: This breadcrumbs collection is what we iterate through on the template.

src/main/webapp/src/app/shared/routes.ts
```diff
@@ -16,34 +20,59 @@ export const routes: Routes = [
   },
   {
     path: 'search',
-    component: AdminSearchComponent
+    component: AdminSearchComponent,
+    data: { breadcrumb: 'labels.search.title'}
   },
```
bgsavage: Here's how the breadcrumb data is set.  We only need to set the label, the route will be captured at runtime.  Notice that we can set the translation pathing here as well since I'm piping the translate service through.

src/main/webapp/src/app/shared/routes.ts
```diff
@@ -16,34 +20,59 @@ export const routes: Routes = [
...
+    data: { breadcrumb: 'labels.groups' },
     children: [
-      { path: '', redirectTo: '/students', pathMatch: 'full' },
-      { path: 'students', component: GroupStudentsComponent },
-      { path: 'students/:studentId',
+      { path: '', pathMatch: 'full', component: GroupsComponent },
+      {
+        path: ':groupId/students',
+        data: { breadcrumb: '[resolve]=groupData.group.name' },
+        resolve: { groupData: GroupResolve },
```
bgsavage: `[resolve]=` is my custom implementation.  I created this syntax to mimic other angular2 syntax, This is the cleanest approach I've gotten to this point.  Note that we can only map to a single field here, so if additional logic is required to create the breadcrumb label, it will need to be encapsulated in a field or mapped when it comes back from the API.

bgsavage: In this specific case, groupData (on line 33) refers to the groupData being resolved by GroupResolve (line 34)

src/main/webapp/src/app/admin-search/admin-search.component.ts
```diff
@@ -1,7 +1,7 @@
-import { Component, OnInit } from '@angular/core';
-import { TranslateService } from "ng2-translate";
-import { DataService } from "../shared/data.service";
-import { ActivatedRoute, Router } from "@angular/router";
+import {Component, OnInit} from "@angular/core";
```
malaffoon: i thought js folks hated double quotes

src/main/webapp/src/app/app.module.ts
```diff
@@ -17,14 +17,17 @@ import {routes} from "./shared/routes";
...
 import {StudentExamItemsComponent} from "./student-exam-items/student-exam-items.component";
-import {BreadcrumbsComponent} from "./breadcrumbs/breadcrumbs.component";
 import {SearchPipe} from "./search.pipe";
-import {BreadcrumbComponent} from "./breadcrumbs/breadcrumb.component";
 import {StudentExamsCategorizedComponent} from "./student-exams/student-exams-categorized.component";
 import {GroupExamItemComponent} from "./group-exam-item/group-exam-item.component";
 import {StudentExamReportComponent} from "./student-exam-report/student-exam-report.component";
 import {IabReportComponent} from "./student-exam-report/iab-report.component";
-import { AdminSearchComponent } from './admin-search/admin-search.component';
+import {AdminSearchComponent} from './admin-search/admin-search.component';
```
malaffoon: is there any order to these, e.g. alphabetical by module or something?

PR #64, opened 2017-04-21
adds first shot at role/permission computation logic

esotrope:
First pass at role and permission security logic

I plan on adding more java docs and a bunch of unit tests in a followup pull request.


src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
```diff
@@ -0,0 +1,195 @@
...
+					}
+				}
+			}
+		}
+		return authorities.build();
+	}
+
+	private Map<String, Collection<String>> getPermissionsByRole() throws AuthorizationServiceException {
+		try {
+			return permissionService.getPermissionsByRole();
```
esotrope: Plan on adding an in-memory cache to cache the results of looking up these permissions

src/main/java/org/opentestsystem/rdw/reporting/model/Permission.java
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * The scope or access level the permission is granted for
+	 */
+	private final PermissionScope scope;
+
+	public Permission(
+		final @NotNull String id,
+		final @NotNull PermissionScope scope) {
+		this.id = checkNotNull(id, "id must not be null");
```
malaffoon: This indent seems wrong, should be 4.

src/main/java/org/opentestsystem/rdw/reporting/model/DistrictAndInstitutionIdentities.java
```diff
@@ -0,0 +1,31 @@
...
+
+import javax.validation.constraints.NotNull;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Holds district and institution natural ID to entity ID mappings
+ */
+public class DistrictAndInstitutionIdentities {
```
bgsavage: "And" in your naming is usually a code smell of breaking SRP.

Can you give a more logical name to this that describes why we are grouping district and institution ids?

src/main/java/org/opentestsystem/rdw/reporting/repository/EntityRepository.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public interface EntityRepository {
+
+	/**
+	 * Gets the local entity IDs for districts and institutions
+	 * @param districtNaturalIds The remote system district IDs used to lookup the local district entity IDs
+	 * @param institutionNaturalIds The remote system institution IDs used to lookup the local institution entity IDs
+	 * @return district and institution natural ID to entity ID mappings
+	 */
+	DistrictAndInstitutionIdentities findAll(Collection<String> districtNaturalIds, Collection<String> institutionNaturalIds);
```
malaffoon: I don't think munging districts and institutions into one method makes sense at the repository level. I would suggest either two repositories or perhaps this single repository with two methods, e.g. `findAllDistricts` and `findAllInstitutions`. Then a service layer or a utility method can combine the results into a single construct.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/EntityRepositoryImpl.java
```diff
@@ -0,0 +1,69 @@
...
+		jdbcTemplate.query(query, ImmutableMap.of(
+			"districtType", DISTRICT_TYPE,
+			"schoolType", INSTITUTION_TYPE,
+			"districtNaturalIds", districtNaturalIds,
+			"schoolNaturalIds", institutionNaturalIds
+		), (row, index) -> {
+			final int type = row.getInt("type");
+			final String naturalId = row.getString("natural_id");
+			final Long id = row.getLong("id");
+			if (type == INSTITUTION_TYPE) {
```
malaffoon: I think the complexity of this implementation supports my assertion that districts/schools should be pulled apart.

src/main/java/org/opentestsystem/rdw/reporting/model/Permission.java
bgsavage: Oh hey, @rmitchell3 was just showing us this example in code club yesterday!  Yes, as I understand it, line 25 should be less indented than 23 and 24.

src/main/java/org/opentestsystem/rdw/reporting/model/PermissionScope.java
```diff
@@ -0,0 +1,85 @@
...
+ * Used to limit database query results
+ */
+public class PermissionScope {
+
+	/**
+	 * Statewide permission scope flyweight
+	 * Since all statewide permission scopes will be identical, we can use the same immutable object instance
+	 */
+	public static final PermissionScope STATEWIDE = new PermissionScope(true, null, null);
+
```
bgsavage: Would we eventually soft code this statewide definition?

src/main/java/org/opentestsystem/rdw/reporting/service/AuthorizationService.java
```diff
@@ -0,0 +1,42 @@
...
+public interface AuthorizationService {
+
+	/**
+	 * Computes role grants from encoded tenancy chain grants provided by ART.
+	 * This method guarantees non-repeating entries, and ignores redundant entries.
+	 * An entry is considered redundant when it is equal to another or contained within another entry.
+	 *
+	 * @param encodedGrants an array of encoded grant strings provided by ART as-is
+	 * @return immutable set of unique role-specific grants
+	 * @throws AuthorizationServiceException when grants are insufficient for application data access
```
malaffoon: this seems out of place for this method; shouldn't it just be doing the grant manipulations, leaving the decision about data access for another place?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
```diff
@@ -0,0 +1,195 @@
...
+		this.entityRepository = entityRepository;
+		this.stateId = stateId;
+	}
+
+	public Set<Grant> getGrants(String[] encodedGrants) throws AuthorizationServiceException {
+		final Set<Grant> grants = newLinkedHashSet();
+		for (String encodedGrant : encodedGrants) {
+			final Grant grant = Grants.from(encodedGrant);
+
+			/*
```
malaffoon: I guess this is a personal taste thing but the big whitespace comment looks too much like javadoc spacing to me; any chance i can convince you to make these simple line comments, e.g.
```java
            // Do not add redundant grants or grants that do not belong to othe application instances' state
            if (stateId.eq...)
```

src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
```diff
@@ -0,0 +1,195 @@
...
+			final Collection<String> permissionIds = permissionsByRole.get(role);
+
+			/*
+				User may have roles not registered to this application in the permission service.
+			 */
+			if (permissionIds != null) {
+				for (String permissionId : permissionIds) {
+					permissions.add(
+						Permission.builder()
+							.id(permissionId)
```
malaffoon: yeah, this 8-space indent is starting to get to me ...

src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
```diff
@@ -0,0 +1,195 @@
...
+				do not belong to the application instance's state
+			 */
+			if (stateId.equals(grant.getStateId()) && !Grants.contains(grants, grant)) {
+				grants.add(grant);
+			}
+		}
+
+		/*
+			Secures the application by throwing exception if the user
+			has no authority to access data for this state
```
malaffoon: as noted above, i think this logic belongs somewhere else

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -1,151 +1,200 @@
...
+		}
+
+		public Builder lastName(final String lastName) {
+			this.lastName = lastName;
+			return this;
+		}
+
+		public Builder permissions(final Collection<? extends Permission> permissions) {
+			this.permissions = permissions;
+			return this;
```
bgsavage: Just a general question here about the builder pattern...

The purpose of builders is so that our objects are immutable.  However, if we are just setting our permissions collection like this (as opposed to copying / cloning) then aren't we at risk of this permissions collection changing outside the context of this class?

src/main/java/org/opentestsystem/rdw/reporting/service/AuthorizationServiceException.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.reporting.service;
+
+/**
+ * Exception thrown when a user is not authorized to access the data
+ * of the application instance and when an error occurs during authorization
+ */
+public class AuthorizationServiceException extends Exception {
```
malaffoon: I know we sort of talked about exceptions and, given that chat, i see why you are doing this. But, i don't really like our trend to creating non-value-add checked exceptions for all our services. Although that article on exceptions defended checked exceptions in certain situations, i guess my general position agrees with "checked exceptions are a failed java experiment." So i prefer to create new exception classes only when they add value (e.g. store additional information).

src/main/java/org/opentestsystem/rdw/reporting/security/support/Authorities.java
```diff
@@ -0,0 +1,47 @@
+package org.opentestsystem.rdw.reporting.security.support;
+
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+
+/**
+ * Utility for dealing with spring {@link GrantedAuthority}
+ */
+public final class Authorities {
+
```
bgsavage: Does a class like this break our dependency injection pattern?  I think this may introduce a challenge when trying to unit test consumers of this class.

src/main/java/org/opentestsystem/rdw/reporting/service/AuthorizationServiceException.java
bgsavage: Is this different than Spring's AuthorizationServiceException?  

http://docs.spring.io/spring-security/site/docs/4.0.4.CI-SNAPSHOT/apidocs//org/springframework/security/access/AuthorizationServiceException.html

src/main/resources/application.sql.yml
```diff
@@ -13,3 +13,13 @@ sql:
             from student_group_membership sgm
               join student s on sgm.student_id = s.id
               where student_group_id=:student_group_id
+
+  districtsAndInstitutions:
+    findAll: >-
+      select :districtType as type, natural_id, id
+        from district
+        where natural_id in (:districtNaturalIds)
+      union
```
chdietz: My comment on this is similar to some above, I would prefer this as two queries.  Is the goal to save a round trip to the database?

agorina: If you want to do something like this, then use **UNION ALL**.  But I suggest breaking it into two calls. Regardless of the type of union, it  will create a temp table which is not efficient. It is better to make two calls.

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
malaffoon: The constructor that the builder uses makes copies of the collection.

bgsavage: Ah, I missed that.  Thanks!

src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
chdietz: Are you talking about when a user is authenticated and we build the spring user object that we do not have to repeat this process until they log out.

src/main/java/org/opentestsystem/rdw/reporting/model/Permission.java
esotrope: it's intellij auto formatted with tabs

src/main/java/org/opentestsystem/rdw/reporting/model/DistrictAndInstitutionIdentities.java
esotrope: Yeah i don't like it either. Didn't want to call it anything generic though because the fields or strictly district and institution ids. hopefully we can eliminate the need for it altogether

src/main/java/org/opentestsystem/rdw/reporting/model/Permission.java
chdietz: I think 4 is the default, I tend to use intellij formatting when creating new files and keep it formatted.
The exception is inline sql and we don't have that issue since our sql is in yaml.

src/main/java/org/opentestsystem/rdw/reporting/repository/EntityRepository.java
esotrope: That would be great and simplify things greatly. I wasn't sure how the team would feel about it though since it implies 2 db connections at login time. Also, when we support group of districts and institutions these will likely also need their own db connections if we follow that pattern resulting in a total of 4 connections at login time.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/EntityRepositoryImpl.java
esotrope: yeah I agree, I'll be glad to change it

src/main/java/org/opentestsystem/rdw/reporting/service/AuthorizationService.java
esotrope: Yeah, I like the idea of just grant calculation. 
Originally i had the access check in the SAMLUserDetailsServiceImpl but then both classes need to be informed of the application state code to perform their job, was trying to keep the auth service responsible for it. I might just break it out into a separate method that the user detail service calls explicitely

src/main/java/org/opentestsystem/rdw/reporting/model/Permission.java
chdietz: tabs, good question, I usually go with spaces

src/main/java/org/opentestsystem/rdw/reporting/security/support/Authorities.java
esotrope: I'm not sure I follow what you're trying to say

src/main/resources/application.sql.yml
esotrope: yeah that was the goal - when we implement groups of districts and institutions that will result in 4 queries at login time. would like to review that and get a group consensus on direction there

src/main/java/org/opentestsystem/rdw/reporting/security/support/Authorities.java
bgsavage: Because the class only has static methods, it cannot be injected.  So when unit testing consumers of this dependency, this class cannot be stubbed out.  You would probably require some runtime mocking tool in order to mock the class out if you needed to.  

Isn't it generally better practice to create object instances versus using global static methods?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
esotrope: no the idea here is to look it up once and keep it in memory until the application is recycled - these permissions are app-specific permissions defined by the perm-web app

src/main/resources/application.sql.yml
esotrope: ok sounds like you, chris and mark agree on separate calls. I think I'll go ahead and make that change : j

bgsavage: Question: Can the calls be made in parallel?

src/main/java/org/opentestsystem/rdw/reporting/security/support/Authorities.java
esotrope: Ah I see what you're saying now. Yes you are right, I think OOP/spring approves more of your approach. We can maybe talk about direction on that in a parking lot. I think it works fine as long as you fully unit test your util classes - but yes in theory you no longer have a unit test of other components "invisibly" depending on this util - you are functional testing two pieces at once

src/main/resources/application.sql.yml
esotrope: not sure how easy that would be to arrange in java. sounds like it would create more overhead than its worth if the calls come back quick. we'll have to see how it plays out in performance testing i think

src/main/java/org/opentestsystem/rdw/reporting/config/CacheConfig.java
```diff
@@ -0,0 +1,25 @@
...
+
+	@Bean
+	public CacheManager cacheManager() {
+		final GuavaCache permissionsByRoleCache = new GuavaCache("permissionsByRole", CacheBuilder.newBuilder().build());
+		final SimpleCacheManager simpleCacheManager = new SimpleCacheManager();
+		simpleCacheManager.setCaches(newArrayList(permissionsByRoleCache));
+		return simpleCacheManager;
+	}
+
+}
```
agorina: I think @malaffoon showed to me how it all could be done with a few lines in the application.yml. Does not really change the behaviour, but less code to support.

esotrope: Ok cool. I look forward to seeing that

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/AbstractEntityRepository.java
```diff
@@ -0,0 +1,39 @@
...
+			jdbcTemplate.query(
+				findAllIdsQuery,
+				new MapSqlParameterSource("naturalIds", naturalIds),
+				(row, index) -> {
+					idsByNaturalId.put(row.getString("natural_id"), row.getLong("id"));
+					return (Void) null;
+				});
+		}
+		return idsByNaturalId;
+	}
```
agorina: I like this, will use in my code. 
One question: why do you cast null to (Void)?

esotrope: its a workaround - the query function required a lamda that returns something so its like returning something to satisfy the lamda requirement while returning nothing at the same time. there might be a better jdbctemplate method i just didnt see it.
I wanted to use queryForMap but it returns Map<String, Object>

agorina: I know, I had the same struggle.  But I copied your approach and just returned Null. What is the benefit of (Void)?

esotrope: oh no probably not - its an object type where null is not typed. so you can put it in the <> generic typing declarations - actually i just found this:
[ResultSetExtractor](http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jdbc/core/ResultSetExtractor.html)
I think this is what we want for mapping results manually.

agorina: Thank you. I will look at it.

src/main/java/org/opentestsystem/rdw/reporting/config/CacheConfig.java
malaffoon: Yeah, shouldn't need this. Look in either of the processor apps in RDW_Ingest.

src/main/java/org/opentestsystem/rdw/reporting/model/PermissionScope.java
```diff
@@ -0,0 +1,85 @@
...
+	 * Holds district entity IDs to which the permission is granted
+	 */
+	private final Set<Long> districtIds;
+
+	/**
+	 * Holds institution entity IDs to which the permission is granted
+	 */
+	private final Set<Long> institutionIds;
+
+	public PermissionScope(
```
malaffoon: nit: seems like the first param is exclusive of the other two params; and you have a flyweight for the statewide; so your one public constructor should not have the first param and can infer it is false.

src/main/java/org/opentestsystem/rdw/reporting/model/DistrictAndInstitutionIdentities.java
jtreuting: Smarter Balanced refers to schools, districts, etc. as the Organization Level, so this could be `OrganizationIdentities`?

src/main/resources/application.sql.yml
jtreuting: @agorina, @chdietz  Are we sure that 2 or 4 queries is more performant than using a `UNION ALL`?  Since this example is passing in a list of IDs I can't imagine we are going to have more than 10-50 since this is from the tenancy chain.  I now that `UNION` uses temp tables but can't seem to get an obvious answer on `UNION ALL` since it just puts them together and doesn't need to sort.  And even if it does, with so few records is that more overhead than multiple connections and network latencies?

Again, I'm not seeing any definitive answer (which seems to be common when I search for answers to MySQL), but I could definitively be missing something.

@bgsavage I'm guessing your question is coming from how easy it is to do in .NET.  There is a way with Java 8 streams, there is a parallelStream() concept.  But with the overhead of it, just like .NET, it's not always more efficient to parallelize the work.  But here is an example if you are interested:

```
    Set<Server> servers = getServers();
    Map<String, String> serverData = new ConcurrentHashMap<>();

    servers.parallelStream().forEach((server) -> {
        serverData.put(server.getIdentifier(), server.fetchData());
    });
```

malaffoon: @jtreuting my original suggestion wasn't really so much about performance. Given the size of this data (and the fact that it will probably be cached in production) that isn't the concern. The issue is complexity: precisely because performance isn't an issue i was suggesting we keep the query and result handling simple. I'm sure we could write nifty single SQL queries for just about everything we want to do, including every screen in the app. But is that really a good idea?  Of course, if performance becomes an issue we will have to revisit this.

agorina: @jtreuting, with my local MySQL the query plan shows a use of a temp table with *UNION ALL*.  
union all

To answer your question if I am sure - not until we test it. But I do not see why we are afraid of the connections. These types of calls (even if not cached) will be instantaneous, and with the connection pooling I do not see a problem.

jtreuting: Interesting.  And I'm fine with all of that at this point trying to clarify and learn because I had a different understanding (maybe misunderstanding) of Union all.

Is that explain with the union or Union all?

From MySQL docs on internally temporary table use:

"The server does not use a temporary table for UNION statements that meet certain qualifications. Instead, it retains from temporary table creation only the data structures necessary to perform result column typecasting. The table is not fully instantiated and no rows are written to or read from it; rows are sent directly to the client. The result is reduced memory and disk requirements, and smaller delay before the first row is sent to the client because the server need not wait until the last query block is executed. EXPLAIN and optimizer trace output reflects this execution strategy: The UNION RESULT query block is not present because that block corresponds to the part that reads from the temporary table."

https://dev.mysql.com/doc/refman/5.7/en/internal-temporary-tables.html

I think you and Mark are right on separating them and then being instantaneous, but guess I might have a misconception on what Union all did.

agorina: I am trying to learn too :) 

Interestingly the 5.6 version of the same page does not have this paragraph: https://dev.mysql.com/doc/refman/5.6/en/internal-temporary-tables.html.  

Since Aurora is based on 5.6, this is what I have installed locally. However Aurora has a lot of modifications, so it may work differently.

The explain was from this SQL with 5.6 version of MySQL:
```
explain
select id from school where id in (42)
union all
select id from district where id in(33)
```

jtreuting: I have MySQL 5.7 locally and the same schema in RDS on MySQL 5.6.  When I do something similar locally in 5.7 there is not the 3rd row with the `UNION RESULT` and `Using temporary` while there is when I do it in MySQL 5.6.

Definitely learned something new today.

PR #66, opened 2017-04-22
removes auth exception

esotrope:
replaced auth exceptions with messaged runtime exceptions
removed DistrictAndInstitutionIdentities model


src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
```diff
@@ -31,146 +29,146 @@
...
-
-	}
+    private String stateId;
+    private PermissionService permissionService;
+    private DistrictRepository districtRepository;
+    private InstitutionRepository institutionRepository;
+
+    @Autowired
+    public AuthorizationServiceImpl(
+        @NotNull @Value("${application.state}") String stateId,
```
malaffoon: well now you're just going to be annoyed with me: the line-continuation indent should be 8; guess i'll work on that IntelliJ style thing

PR #67, opened 2017-04-24
fix User

malaffoon:


src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -12,189 +12,189 @@
...
+
+        public User build() {
+            return new User(
+                    id,
+                    email,
+                    firstName,
+                    lastName,
+                    permissions,
+                    username,
+                    password,
```
malaffoon: This is another reason i don't like methods with lots of parameters. Since java doesn't have named parameters, it is too easy to not get them right. I would recommend changing this class to either take the builder as the constructor param, or set fields directly in builder.

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -5,196 +5,175 @@
...
 
-import javax.validation.constraints.NotNull;
 import java.util.Collection;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+
 public class User implements UserDetails {
 
-	private final String id;
```
malaffoon: Unfortunately, whitespace is hiding most of the change here, i'll try to call them out

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -5,196 +5,175 @@
...
+    private final Collection<? extends Permission> permissions;
+    private final String username;
+    private final String password;
+    private final Collection<? extends GrantedAuthority> authorities;
+    private final boolean accountNonExpired;
+    private final boolean accountNonLocked;
+    private final boolean credentialsNonExpired;
+    private final boolean enabled;
+
+    User(final Builder builder) {
```
malaffoon: Change scope to module-private.
Change to passing in builder instead of fields. Why? Cause there was a bug with the order of the parameters.

jtreuting: The other option is to put this same code into the `build` method so there isn't a constructor at all (though you will need an empty private constructor if using Jackson I believe).  Then the builder is in charge of building the object and not the object itself.

malaffoon:
@jtreuting - personally, that is my favorite pattern. I don't like that User now has to know about Builder (which knows about User). But then i'd have to make all the members non-final, and move the validation code to the builder. That seemed more intrusive; this was the least impact approach.

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
bgsavage: An automapper to map any builder to any model would be nice to not always have to have this boiler plate code.

esotrope: well then we might as use Immutables or lombok X P

PR #70, opened 2017-04-25
Feature/student group security

esotrope:


src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
```diff
@@ -0,0 +1,78 @@
...
+public class GroupMembership {
+
+    private final Long id;
+    private final Long subjectId;
+    private final Long institutionId;
+
+    public GroupMembership(
+            final @NotNull Long id,
+            final @NotNull Long subjectId,
+            final @NotNull Long institutionId) {
```
agorina: My understanding of the Group is that you do not need *institution* for the permissions. It is just a supporting data that is not meant to be used in the SQL.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
```diff
@@ -1,48 +1,54 @@
...
+
 import static com.google.common.base.Preconditions.checkNotNull;
 
 @Repository
 public class GroupRepositoryImpl implements GroupRepository {
 
     private final NamedParameterJdbcTemplate jdbcTemplate;
-
-    @Value("${sql.group.findAll}")
-    private String findAllSql;
```
agorina: The Ingest project has adopted this pattern. Just curious why you do not like it?

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
```diff
@@ -1,48 +1,54 @@
...
-    private static class GroupRowMapper implements RowMapper<Group> {
-        @Override
-        public Group mapRow(final ResultSet rs, final int rowNum) throws SQLException {
-            return Group.builder()
-                    .name(rs.getString("name"))
-                    .schoolName(rs.getString("school_name"))
-                    .id(rs.getLong("id"))
-                    .build();
-        }
+    public List<GroupMembership> findAllMemberships(String username) {
```
agorina: Suggest to rename it to be explicit that it uses *username* to find: **findAllMembershipForUsername**.

While I was writing the above comment, I realized that It feels like this does not belong to a Group repository. Instead maybe it is better to have a GroupMembership Repository?

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -17,6 +18,7 @@
     private final String firstName;
     private final String lastName;
     private final Collection<? extends Permission> permissions;
+    private final Collection<? extends GroupMembership> groups;
```
agorina: Here is a use case I see: a Service needs to know if a User is allowed to access a group by ID. Base on this use case, would it make sense to store this as a Map?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
```diff
@@ -119,6 +124,10 @@ public AuthorizationServiceImpl(
         return authorities.build();
     }
 
+    public Set<GroupMembership> getGroups(String username) {
+        return ImmutableSet.copyOf(groupRepository.findAllMemberships(username));
```
agorina: I have to admit that I have not been following previous reviews close enough to understand how this is meant to be used. What confused me is that there is not check for the PII GROUP role. Could you please clarify how you see it?

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public Long getSubjectId() {
+        return subjectId;
+    }
+
+    public Long getInstitutionId() {
+        return institutionId;
+    }
+
+    public boolean equals(Object o) {
```
malaffoon: When overriding equals/hashCode to not do the expected thing (deep compare of all values), please add javadoc to that effect.

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
```diff
@@ -0,0 +1,78 @@
+package org.opentestsystem.rdw.reporting.model;
+
+import javax.validation.constraints.NotNull;
+
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class GroupMembership {
```
malaffoon: It's not clear to me what `GroupMembership` is. Why have this instead of just Group? Perhaps javadoc explaining why this is necessary / different than just a Group.

src/main/java/org/opentestsystem/rdw/reporting/repository/GroupRepository.java
```diff
@@ -1,17 +1,29 @@
...
-    List<Group> findAll(String userName);
+    List<Group> findAll(String username);
+
+    /**
+     * Find all {@GroupMembership}s for the given user name
+     *
+     * @param username the user's name
+     * @return collection of {@link GroupMembership}
+     */
+    List<GroupMembership> findAllMemberships(String username);
```
malaffoon: Again, not really understanding the distinction between Group and GroupMembership.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
```diff
@@ -1,48 +1,54 @@
...
-    @Value("${sql.group.findAll}")
-    private String findAllSql;
+    private String findAllQuery;
+    private String findAllMembershipsQuery;
 
     @Autowired
-    GroupRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
+    GroupRepositoryImpl(
+            final NamedParameterJdbcTemplate jdbcTemplate,
+            final @Value("${sql.group.findAll}") String findAllQuery,
```
malaffoon: Per Alla's comment i am strongly opposed to stuffing all the sql strings into the constructor. First suggestion is use the same pattern established elsewhere. If you really don't like that for some reason, convince the team why and we'll make everything consistent.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
```diff
@@ -1,48 +1,54 @@
...
 
-    @Override
-    public List<Group> findAll(final String userName) {
-        return jdbcTemplate.query(findAllSql,
-                new MapSqlParameterSource("user_login", userName),
-                new GroupRowMapper());
+    public List<Group> findAll(final String username) {
+        return jdbcTemplate.query(findAllQuery,
+                new MapSqlParameterSource("user_login", username),
+                (results, index) -> Group.builder()
```
malaffoon: In repo classes i like to have a separate, explicit RowMapper class. Why? Because it is very likely to be used by more than one method. And IMO it is more readable. I would go so far as to have a static instance but that isn't necessary:
```
    return jdbcTemplate.query(findAllQuery, new MapSqlParameterSource("user_login", username), new GroupRowMapper());
```

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
malaffoon: I'm still trying to understand why we have a separate object for this.

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private final Long id;
+    private final Long subjectId;
+    private final Long institutionId;
+
+    public GroupMembership(
+            final @NotNull Long id,
+            final @NotNull Long subjectId,
+            final @NotNull Long institutionId) {
+        this.id = checkNotNull(id, "id must not be null");
```
bgsavage: Should we standardize on how we construct our objects?  I've been using parameterless constructors and just assigning properties from the builder because that's what I've seen as more prevalent.

jtreuting: Yes we should.  I was going to bring it up in the parking lot today

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
bgsavage: I agree, even if we take away all the Spring functionality we get, I don't think we'd ever want the caller to have to construct the SQL that a repository uses.  If anything, it should take a configuration object.  However, per Alla and Mark's comments, we do have an established pattern that we should discuss before breaking.

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
esotrope: ok cool - i'll remove it

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
esotrope: idea behind making it a constructor param is that the repo is more portable and can be created without spring and can be more easily tested

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/GroupRepositoryImpl.java
esotrope: Different repo works
The reason for having a different object model is to hold the minimum set of data necessary to represent the membership / access permission

src/main/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImpl.java
esotrope: good point - I will add a check so we don't make an unnecessary extra call here

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
esotrope: sounds good - will add

src/main/java/org/opentestsystem/rdw/reporting/security/User.java
esotrope: good point - yeah for certain permission checks this would be the right data type - for passing to the db we could just do .values() on the map

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        final GroupMembership that = (GroupMembership) o;
+        return id.equals(that.id);
+    }
+
+    public int hashCode() {
+        return Objects.hash(id);
```
agorina: I understand that you are trying to short-circuit it, but a Group's subject could change. So I think it is dangerous.

src/main/java/org/opentestsystem/rdw/reporting/repository/GroupRepository.java
```diff
@@ -1,17 +1,23 @@
...
      * Find all {@Group}s for the given user name
      *
-     * @param userName the user name
-     * @return the collection of the {@link Group}
+     * @param username the user name
+     * @return collection of {@link Group}
      */
-    List<Group> findAll(String userName);
+    @PreAuthorize("hasAuthority('PERM_GROUP_PII_READ')")
+    List<Group> findAll(String username);
```
agorina: I am struggling to decide it this belongs to the Repository or Service layer. Would like to know what other think.

esotrope: spring advocates security annotations be put on service interfaces. It's assumed that all repo access will go through some service layer

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
esotrope: What is the danger/fail case here? I'm not sure I follow.

src/test/java/org/opentestsystem/rdw/reporting/service/impl/AuthorizationServiceImplTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.opentestsystem.rdw.reporting.repository.InstitutionRepository;
+import org.opentestsystem.rdw.security.Grant;
+import org.opentestsystem.rdw.security.service.PermissionService;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
```
agorina: This maybe one more thing to be consistent: the Ingest project uses ```import static org.assertj.core.api.Assertions.assertThat;```

src/main/java/org/opentestsystem/rdw/reporting/repository/GroupRepository.java
agorina: In this case why did you put it on the Repository?

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
agorina: I do not know how it is being used, but the use case is:  somebody has this object in memory. Then a group administrator updates the Group and changes the Subject to a different value (let's say Math to All). A db lookup is done and retrieves a new Group, compares to the one in memory and finds them identical.  

I understand that the example is very hypothetical, but still...

src/main/java/org/opentestsystem/rdw/reporting/repository/GroupRepository.java
esotrope: Oh did I? oops X l
I'll move that

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
esotrope: Ok I see.
I was under the impression that if we look up groups ahead of time at login time we would cash that until the session ends. If we want it to update when the subjects change we need to look it up on the fly before every query and not at login time I think.

src/main/java/org/opentestsystem/rdw/reporting/repository/GroupRepository.java
agorina: :) thank you!

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
agorina: Yeah,  I know we talked about caching. But the more I think about it, the less I like it.  I am afraid of a use case when a teacher finds something wrong about her Group access, contacts an administrator to make a change and does not see the change. I guess if we cache just for the duration of the session, then we can instruct the users that they need to logout to see the change? This needs to be further designed and documented I think.

src/main/java/org/opentestsystem/rdw/reporting/model/GroupMembership.java
esotrope: Idea here is to have the minimal representation of a group necessary to perform security checks. I didn't want to couple it to whatever shape the Group view model/DTO will eventually take.
Will add javadocs though

PR #71, opened 2017-05-01
Feature/pdf export

esotrope:
Looking primarily for critical eyes on the design/code organization and separation of responsibilities at this point.

To come:
1. bulk export api
2. unit / integration tests
3. java docs


src/main/java/org/opentestsystem/rdw/reporting/config/ApplicationConfig.java
```diff
@@ -21,5 +33,34 @@ public PermissionService defaultPermissionService(PermissionWebServiceClient cli
         return new ComponentPermissionService(client, component);
     }
 
+    @Bean
```
malaffoon: In general, Configuration classes should be separated based on concern, so this would warrant a new class.

src/main/java/org/opentestsystem/rdw/reporting/config/ApplicationConfig.java
```diff
@@ -21,5 +33,34 @@ public PermissionService defaultPermissionService(PermissionWebServiceClient cli
...
+    public Template examReportTemplate(TemplateEngine engine, ObjectMapper mapper) {
+        return new ThymeleafTemplate("exam-report", engine, mapper);
+    }
+
+    @Bean
+    public MessageSource messageSource(TranslationService translationService) {
+        return new TranslationServiceMessageSource(translationService);
+    }
+
+    private static final class TranslationServiceMessageSource extends AbstractMessageSource {
```
malaffoon: Perhaps in the next iteration break this out into its own module. Then you can write tests, etc. more easily.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RestController;
+
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+@RestController
+public class ExportController {
+
+    @Autowired
```
malaffoon: Isn't IntelliJ warning about field injection here? Suggest either constructor or setters instead.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RestController;
+
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+@RestController
+public class ExportController {
```
malaffoon: Name is a bit vague but i don't know the whole plan on where it is going. Personally, i would've gone with ReportController or PdfController or some combination of those with "export".

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Autowired
+    private ExamReportService examReportService;
+
+    @Autowired
+    private ExamReportExportService examReportExportService;
+
+
+    @PostMapping(value = "/api/exams/{examId}/report")
```
malaffoon: This isn't specific to this PR but i'm a bit concerned that "/api" is being scattered across all the controllers. We should make an effort to have that in one place and the individual controllers just extend from there.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamReportExportServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+        this.htmlToPdfService = htmlToPdfService;
+    }
+
+    @Override
+    public void exportPDF(
+            final @NotNull ExamReport report,
+            final @NotNull ExamReportExportOptions options,
+            final @NotNull HttpServletResponse response) throws RuntimeException {
+
+        final byte[] html = examTemplate.render(report, options.getLocale());
```
malaffoon: This feels a like too many cross-cutting concerns in one method. It feels like the creation of the PDF should be separate from the HTTP response stuff. When we start doing batch operations we will need lines 38-44 but won't want the response handling. Or, if we wanted to generate single pdf's via something other than an HTTP request/response mechanism.

src/main/java/org/opentestsystem/rdw/reporting/model/PdfOptions.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.reporting.model;
+
+import java.io.Serializable;
+
+public class PdfOptions implements Serializable {
```
malaffoon: Why Serializable?

src/main/java/org/opentestsystem/rdw/reporting/model/ExamReportExportOptions.java
```diff
@@ -0,0 +1,30 @@
+package org.opentestsystem.rdw.reporting.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.io.Serializable;
+import java.util.Locale;
+
+public class ExamReportExportOptions implements Serializable {
```
malaffoon: Why Serializable?

src/main/java/org/opentestsystem/rdw/reporting/model/Template.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.reporting.model;
+
+import javax.validation.constraints.NotNull;
+import java.util.Locale;
+
+public interface Template {
+
+    byte[] render(@NotNull Object model, @NotNull Locale locale);
```
malaffoon: Why make this so generic, isn't it for rendering ExamReport?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private File createTemporaryFile() throws IOException {
+        final String temporaryFilename = "wkhtmltopdf-" + UUID.randomUUID().toString();
+        return File.createTempFile(temporaryFilename, ".html");
+    }
+
+    private byte[] createPDFFromFile(final String command) throws IOException, InterruptedException {
+
+        final Process process = Runtime.getRuntime().exec(command);
```
malaffoon: i thought you said something about bundling a jar with the functionality; that didn't work out?

src/main/java/org/opentestsystem/rdw/reporting/model/Template.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.reporting.model;
+
+import javax.validation.constraints.NotNull;
+import java.util.Locale;
+
+public interface Template {
```
malaffoon: Name is a bit confusing as it has nothing to do with thymeleaf Template and there is a class that implements this Template while wrapping thymeleaf. Perhaps some combination of Renderer, Report, Processor, ...

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
```diff
@@ -0,0 +1,39 @@
...
+    public void getExamReportAsPDF(
+            final @PathVariable long examId,
+            final @RequestBody ExamReportExportOptions options,
+            final @AuthenticationPrincipal User user,
+            final HttpServletResponse response) throws IOException {
+
+        final ExamReport report = examReportService.getReport(user, examId);
```
malaffoon: It may be necessary and sensible in this case but whenever i see a single controller end-point calling more than one core service to do work, i get suspicious. The concept of `ExamReport` the way it is now should probably be an implementation detail hidden from the controller; however, if it has additional information about the report (e.g. name) it might make sense to leave it exposed. 

I'm really just thinking "out loud" here. It feels a bit off, but i'd have to fiddle and chat to have a suggestion. And, like i mentioned, perhaps in this case it makes perfect sense.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
chdietz: We should think about the name here I agree because we will have other 'export' such as to CSV.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
bgsavage: It seems weird to me that we have to pass in a user and a response.  Will we always have to pass these two parameters?  Also I'm curious to know if we can have the examId in the ExamReportExportOptions model.

src/main/java/org/opentestsystem/rdw/reporting/model/Template.java
bgsavage: Maybe I'm missing something, but it seems the implementation doesn't do anything specific for ExamReport.

src/main/java/org/opentestsystem/rdw/reporting/repository/impl/TranslationRepositoryStub.java
```diff
@@ -15,7 +16,9 @@
 
     private final Set<Translation> stubTranslations = Sets.newHashSet(
             Translation.builder().withCode("HELLO").withLocale(Locale.ENGLISH).withMessage("Hello World").build(),
-            Translation.builder().withCode("HELLO").withLocale(Locale.JAPANESE).withMessage("こんにちは世界").build()
+            Translation.builder().withCode("HELLO").withLocale(Locale.JAPANESE).withMessage("こんにちは世界").build(),
+            Translation.builder().withCode("examreport.title").withLocale(Locale.ENGLISH).withMessage("Exam Report").build(),
+            Translation.builder().withCode("examreport.title").withLocale(Locale.JAPANESE).withMessage("こんにちは世界").build()
     );
```
bgsavage: "Hello World" in Japanese.  Nice.  :)

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamReportExportServiceImpl.java
bgsavage: I was thinking the same thing when I saw the controller passing an HttpServletResponse to this method.  You might consider an HttpMessageConverter similar to what I am doing for the CSV report so that the transport work stays outside of the service layer.

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
chdietz: Or if every endpoint start with 'api' what is the point.  I have to admit I have done this before on some projects.
So I am saying maybe we just don't need '/api/...' Maybe we just need '/{resource}/'
Such as: /exams/{examId}/pdf-report

bgsavage: The reason for the /api/ prefix is to have an easy separation between the angular routes and the spring api routes.  Though I agree we should be able to set this prefix in one spot.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
```diff
@@ -0,0 +1,69 @@
...
+
+@Service
+public class WebkitHtmlToPdfService implements HtmlToPdfService {
+
+    private static final String COMMAND = "wkhtmltopdf";
+    private static final String OPTION_GRAYSCALE = "--grayscale";
+    private static final String OPTION_OUTPUT_TO_STDOUT = "-";
+
+    @Override
+    public byte[] createPDF(final byte[] html, final @NotNull PdfOptions options) throws RuntimeException {
```
bgsavage: nit:  Should be Pdf instead of PDF right?

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
chdietz: when you say angular routes do you mean the routes to serve the static content such as javascript and css?

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
```diff
@@ -0,0 +1,69 @@
...
+        return File.createTempFile(temporaryFilename, ".html");
+    }
+
+    private byte[] createPDFFromFile(final String command) throws IOException, InterruptedException {
+
+        final Process process = Runtime.getRuntime().exec(command);
+        final byte[] inputBytes = IOUtils.toByteArray(process.getInputStream());
+        final byte[] errorBytes = IOUtils.toByteArray(process.getErrorStream());
+
+        process.waitFor();
```
bgsavage: What does this do to the current thread?  Also, what happens should this never return?

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
bgsavage: Nope I'm referring to these routes:  https://github.com/SmarterApp/RDW_Reporting/blob/develop/src/main/webapp/src/app/shared/routes.ts

src/main/java/org/opentestsystem/rdw/reporting/model/Template.java
esotrope: This was to abstract manual template rendering so that we can swap out the template rendering engine and only change the implementation of this class

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
esotrope: yeah - i tried it on my local machine and wasn't able to use it - the docs do specifically say there are limitations on the OS it can run on. See [wkhtmltopdf-binary](https://github.com/tocco/wkhtmltopdf-binary)

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
esotrope: To your comment Mark - I do like having just one service call in the body of a controller method. The way i had responsibilities set up here though is that one service is for getting/creating the domain model for a report and the other is for turning it into a pdf - these both could be wrapped though in another service - it just becomes a problem of what to call it etc.

To your comment Brandon - I like the idea of moving the exam ID to the options model - this is because we will eventually have a bulk export and we will want to send more than one ID.

Also, for the service layer I think we will need to pass the user object anytime we are going to do a database lookup for user data - this is so the service layer can forward the user ID to the repository

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
esotrope: Yeah good point

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
esotrope: process.waitFor accepts timeout args so I think we would set some timeout here eventually to avoid that sort of thing.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
malaffoon: :(

src/main/java/org/opentestsystem/rdw/reporting/model/Template.java
malaffoon: Sorry, i wasn't clear. I understand the interface, no problem with that. Just don't know why the first argument of render() isn't ExamReport.

src/main/java/org/opentestsystem/rdw/reporting/model/Template.java
esotrope: The name makes sense in the front-end world I think. But I'm open to calling it something else. Usually "template" is the name used for logical construction of html that require a model to be rendered into a final piece of html

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
bgsavage: Is there a way we can set up a UserContext @Bean and use injection to get the current user context when we need it?

src/main/java/org/opentestsystem/rdw/reporting/model/PdfOptions.java
esotrope: old habbit - in my previous life all models had to be serializable and have hashcode/equals - I'll remove these for things we don't intend on serializing

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
malaffoon: @bgsavage we did discuss that a few weeks ago. There is a standard way of doing that in spring apps but we felt that passing the user info in was the path to take for now.

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamReportExportServiceImpl.java
esotrope: Converter would be good. I'm not sure how the converter can perform the "attachment={filename}" header effectively though. This may not be a problem though because right now the angular client will have the final say on the pdf filename

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
bgsavage: 👍

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
esotrope: I think I will rename it to ExamReportExportController - thoughts?

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
esotrope: No - no complaints - Will move to constructor though

src/main/java/org/opentestsystem/rdw/reporting/config/ApplicationConfig.java
esotrope: You mean just move it into its own file right?

src/main/java/org/opentestsystem/rdw/reporting/config/ApplicationConfig.java
esotrope: K - I'll chop it up

src/main/java/org/opentestsystem/rdw/reporting/service/impl/ExamReportExportServiceImpl.java
bgsavage: You can set the headers in the writeInternal method from the HttpOutputMessage

src/main/java/org/opentestsystem/rdw/reporting/config/ApplicationConfig.java
malaffoon: yup

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
malaffoon: sounds good

src/main/java/org/opentestsystem/rdw/reporting/model/Template.java
esotrope: Ah - the idea here is that we might be rendering other things with templates and the template should be able to be used for other models
example:
new ThymeleafTemplate("templateA").render(modelTypeA)
new ThymeleafTemplate("templateB").render(modelTypeB)

src/main/java/org/opentestsystem/rdw/reporting/controller/ExportController.java
esotrope: I rechecked actually - there were complaints I just didn't see the highlighting

src/main/java/org/opentestsystem/rdw/reporting/service/impl/WebkitHtmlToPdfService.java
chdietz: Docs say it can run on OSX?
are you saying wkhtmltopdf-binary did not run on you mac?

esotrope: Looking again it may be that I am missing environment dependencies to build it. But once built it may work.
One not so nice thing about this solution is it wants you to build it yourself so you have whatever version of wkhtmltopdf you need - there do not appear to be any distributions on the github or in common places like maven central

rcordeau: "Note: The current version only supports Linux and OS X (amd64) architecture, but this could easily be extended (pull requests are welcome!)" Isn't our target "OS" really linux anyway?

malaffoon: I think we should pursue this ... but at our leisure. Perhaps create a ticket for it.

PR #75, opened 2017-05-03
introduces modules: webapp, common, pdf-report-processor

esotrope:
1. moved reporting webapp project code into webapp module
2. created pdf-report-processor module
3. created common module for common components between the aforementioned modules
4. modified build.gradle/settings.gradle files to allow for the new build
5. modified docker-compose file to include the new processor service


common/build.gradle
```diff
@@ -0,0 +1,39 @@
+import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+
+plugins {
+    id "net.ltgt.apt" version "0.9"
+}
+
+apply plugin: 'idea'
```
rcordeau: Does common really need all of these plugins? (7-15)
If so, it seems that all of the subprojects have these plugins in common. You can remove 7-13 since they are already applied in the parent build.gradle file.

common/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+apply plugin: 'propdeps'
+apply plugin: 'propdeps-maven'
+apply plugin: 'propdeps-idea'
+apply plugin: 'java'
+apply plugin: 'maven'
+apply plugin: 'jacoco'
+apply plugin: 'net.ltgt.apt'
```
rcordeau: This one and the one below it - if they really are needed for everything, you can move them out to the subprojects closure in the parent build.gradle file.

common/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// generate coverage report automatically (./build/reports/jacoco/test/html/index.html)
+test.finalizedBy(jacocoTestReport)
+
+// trigger creation of build-info.properties
+classes.dependsOn(createBuildInfoFile)
+
+// task to build the docker image
+task buildImage(type:DockerBuildImage, dependsOn:[build, createDockerfile]) {
```
rcordeau: Is :common a separate docker image, or just a common library for the other subprojects?

webapp/build.gradle
```diff
@@ -0,0 +1,104 @@
+import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+
+plugins {
+    id "com.moowork.node" version "1.1.1"
+    id "net.ltgt.apt" version "0.9"
+    id "com.dorongold.task-tree" version "1.3"
```
rcordeau: We probably don't need to keep this anymore. It was for visualizing the build DAG when we were setting up the node build dependency.

webapp/build.gradle
```diff
@@ -0,0 +1,104 @@
...
+    testCompile 'junit:junit'
+    testCompile 'org.assertj:assertj-core:3.6+'
+    testCompile('org.springframework.boot:spring-boot-starter-test') {
+        exclude(module: 'commons-logging')
+    }
+    testCompile 'org.springframework.security:spring-security-test'
+}
+
+node {
+    version = '7.5.0'
```
rcordeau: Nodejs and NPM are at a higher release now. Do we want to move to the latest release?

common/build.gradle
esotrope: i have no idea - this is where you can come in and help me

common/build.gradle
esotrope: intended as a common library so it shouldnt need an image

webapp/build.gradle
esotrope: yeah i think at this point in dev we can still do that

build.gradle
```diff
@@ -6,176 +8,83 @@ buildscript {
...
-    npmVersion = '4.0.5'
-    download = true
-    workDir = file("${project.projectDir}/src/main/webapp")
-    nodeModulesDir = file("${project.projectDir}/src/main/webapp")
-}
+    dependencyManagement {
+        dependencies {
+            dependency 'org.opentestsystem.rdw.common:rdw-common-messaging:0.0.1-SNAPSHOT'
+            dependency 'org.opentestsystem.rdw.common:rdw-common-model:0.0.1-SNAPSHOT'
+            dependency 'org.opentestsystem.rdw.common:rdw-common-utils:0.0.1-SNAPSHOT'
```
malaffoon: Probably want to add rdw-common-security in here; and rdw-common-archive

webapp/src/main/java/org/opentestsystem/rdw/reporting/Application.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Import;
+
+@SpringBootApplication
+@EnableConfigurationProperties
+@Import({ResourceLoaderConfiguration.class, YamlPropertiesConfigurator.class})
+public class Application {
+    public static void main(String[] args) {
```
malaffoon: nit: `final`

malaffoon:
I admit it's hard to see the changes clearly without pulling down the patch. Looks good and i'm sure we can tweak it.

common/build.gradle
rcordeau: Ok will do.  Let me do some tweaking.

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/ExamReportPdfRequest.java
```diff
@@ -0,0 +1,5 @@
+package org.opentestsystem.rdw.reporting.common;
```
rcordeau: This needs to be org.opentestsystem.rdw.reporting.common.export;

webapp/src/main/java/org/opentestsystem/rdw/reporting/Application.java
```diff
@@ -0,0 +1,17 @@
...
+import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Import;
+
+@SpringBootApplication
+@EnableConfigurationProperties
+@Import({ResourceLoaderConfiguration.class, YamlPropertiesConfigurator.class})
+public class Application {
```
bgsavage: Should this now be named ReportingApplication?

rcordeau: It could be, but since it will be in its own docker image, it likely won't conflict with anything.

PR #76, opened 2017-05-04
Feature/add streams

esotrope:
1. moves webapp/*Translation to common/../i18n package
2. moves webapp/../web/ export components to common/../export package
3. adds new spring cloud stream dependencies
4. adds cloud stream configs in applicaiton.yml files
5. adds placeholder stream source and sink


common/build.gradle
```diff
@@ -1,8 +1,16 @@
 //Any subproject specific plugins go here
 
 dependencies {
+    compile 'org.apache.commons:commons-io:1.3.2'
+    compile 'com.google.guava:guava:21.0'
+    compile 'com.fasterxml.jackson.core:jackson-annotations:2.8.0'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.8.7'
+    compile 'org.springframework:spring-context:4.3.7.RELEASE'
+    compile 'org.thymeleaf:thymeleaf:2.1.5.RELEASE'
     testCompile 'junit:junit'
```
malaffoon: i'd suggest that junit, assertj, and mockito all be put in the root gradle file; and that the version of guava and apache commons be put in the dependency management section of the root gradle file. You can look at RDW_Ingest for examples.

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/processor/ExportProcessorConfiguration.java
```diff
@@ -0,0 +1,19 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.thymeleaf.TemplateEngine;
+
+@Configuration
+public class ExportProcessorConfiguration {
+
+    @Bean
+    public PdfExamReportProcessor pdfExamReportProcessor(TemplateEngine engine, ObjectMapper mapper){
```
malaffoon: nit: `final` for params

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/processor/HtmlToPdfProcessor.java
```diff
@@ -0,0 +1,9 @@
+package org.opentestsystem.rdw.reporting.common.export.processor;
+
+import org.opentestsystem.rdw.reporting.common.export.PdfOptions;
+
+public interface HtmlToPdfProcessor {
+
+    byte[] process(byte[] html, PdfOptions options) throws RuntimeException;
```
malaffoon: i don't think `throws RuntimeException` really needs to be part of the interface specification.

i think interfaces and their methods should have robust javadocs.

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/processor/ThymeleafTemplateProcessor.java
```diff
@@ -1,27 +1,23 @@
...
 import java.util.Locale;
 import java.util.Map;
 
-public class ThymeleafTemplate implements Template {
+public class ThymeleafTemplateProcessor implements TemplateProcessor {
 
     private static final String DEFAULT_ENCODING = "UTF-8";
     private String encoding = DEFAULT_ENCODING;
-    private String path;
     private TemplateEngine engine;
```
malaffoon: can these be final?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/Translation.java
```diff
@@ -0,0 +1,60 @@
...
+    public static class Builder {
+        private String code;
+        private Locale locale;
+        private String message;
+
+        public Translation build() {
+            return new Translation(code, locale, message);
+        }
+
+        public Builder withCode(final String code) {
```
malaffoon: i know the `withXxx` was used for some early builders but we've been moving away from that, preferring just the field as the method name, e.g. `public Builder code(final String code)`.

pdf-report-processor/src/main/resources/application.yml
```diff
@@ -0,0 +1,26 @@
...
+
+spring:
+  cloud:
+    stream:
+      bindings:
+        input:
+          destination: EXAM_REPORT_PDF
+          group: default
+
+  rabbitmq:
```
malaffoon: This shouldn't be in the embedded application.yml; it should be in the sbac-config-repo version for the local-docker profile only.

webapp/src/main/resources/application.yml
```diff
@@ -17,6 +17,13 @@ spring:
     type: SIMPLE
     cache-names: permissionsByRole
 
+  cloud:
+    stream:
+      bindings:
+        EXAM_REPORT_PDF:
```
malaffoon: i think you also need to define the output binding, something like:
```
  cloud:
    stream:
      bindings:
        output:
          destination: EXAM_REPORT_PDF

**Comment by malaffoon** [comment]
Big PR and i probably need to download the patch to fully grok it.

File: webapp/src/main/java/org/opentestsystem/rdw/reporting/security/SAMLUserDetailsServiceImpl.java [hunk 8]
```diff
@@ -52,6 +54,11 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
         final Set<GrantedAuthority> authorities = authorizationService.getAuthorities(grants);
         final Map<Long, GroupGrant> groupsById = authorizationService.getGroupsById(permissions, username);
 
+//
+//        final Set<Permission> permissions = ImmutableSet.of();
+//        final Set<GrantedAuthority> authorities = ImmutableSet.of();
+//        final Map<Long, GroupGrant> groupsById = ImmutableMap.of();
+
```
agorina: Do you need this?

esotrope: nope - I'll remove it

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/PdfOptions.java
```diff
@@ -1,11 +1,12 @@
...
+package org.opentestsystem.rdw.reporting.common.export;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
 
 public class PdfOptions {
 
-    private boolean grayscale;
+    private final boolean grayscale;
```
bgsavage: The naming of this field could be improved to make it more intuitive that this is a boolean (ie isGrayscale ? )

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/DefaultExamReportExportService.java
```diff
@@ -0,0 +1,56 @@
...
+import org.opentestsystem.rdw.reporting.common.export.PdfExamReportResponse;
+import org.opentestsystem.rdw.reporting.security.User;
+import org.opentestsystem.rdw.reporting.service.ExamReportService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import javax.validation.constraints.NotNull;
+
+@Service
+public class DefaultExamReportExportService implements ExamReportExportService {
```
bgsavage: I've seen "Default...." and "....Impl".  Should we standardize on this?

agorina: Yes, good point. I think we should.

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/ExamReportRequest.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+
+import static com.google.common.collect.Sets.newHashSet;
+
+public class ExamReportRequest<T> {
```
bgsavage: I'm not following the intent of this class.  It feels like maybe it should be abstract if anything.  

Also, The use of generic type here feels a little odd.  If you say it the usage out loud, ExamReportRequest of type PdfOptions doesn't really make sense.  Maybe some javadocs would help here?  I'm not really sure what my suggestion is here, just something to think about really.  :)

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/PdfOptions.java
malaffoon: The convention in java is that the variable is named like this so the accessor becomes isGrayscale(). My only issue with this variable is, is grayscale one word or two? if it is two it should be grayScale (and the accessor becomes isGrayScale()).

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/DefaultExamReportExportService.java
malaffoon: Personally i think `XxxImpl` is as silly as `IXxx` for interfaces.

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/PdfOptions.java
bgsavage: That makes sense, as long as the public call to it is asking a question which the value has an answer for.

For your other question, you could also probably ask if it's gray or grey 😄

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/DefaultExamReportExportService.java
esotrope: If we make a standard I vote for Default*. Also, when there is a more specific way to describe what the implementation strategy is we should use the name of that strategy. e.g. for HtmlToPdfProcessor implementation I use WebkitHtmlToPdfProcessor to indicate that wkhtmltopdf (webkit htmltopdf) lib is being used

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/ExamReportRequest.java
esotrope: Shape/name of this class may change but I'll add javadocs

The basic idea here is that it represents a users' request for 1+ exam reports and this can be in either PDF or some other format - if in PDF we would interpret the request options as pdf specific options ... lends itself to having CSV specific options later or something

common/src/main/java/org/opentestsystem/rdw/reporting/common/export/PdfOptions.java
esotrope: grayscale is one word.
I actually originally called it greyscale but i guess that may be British english? and intellij complained because it speaks American english?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/DefaultExamReportExportService.java
malaffoon: +1

bgsavage: I'll bring it up in parking lot.  I honestly thought XxxImpl was favored because it's more prevalent.  Also comes with that nice impl folder hiding all the details.  I have no strong opinion either way, just looking for consistency so new developers coming in (me) know which one to use.

agorina: I have no preference between Default or Impl. 
Agree with @esotrope comment about the specific one.

esotrope: At some point we should type all these standards up and put them in the contributing file

bgsavage: nose goes 👃

esotrope: 👃 👈

common/src/test/java/org/opentestsystem/rdw/reporting/common/export/processor/WebkitHtmlToPdfProcessorIT.java
```diff
@@ -0,0 +1,36 @@
...
+    private WebkitHtmlToPdfProcessor processor = new WebkitHtmlToPdfProcessor();
+    private PdfOptions defaultPdfOptions = new PdfOptions(false);
+    private String emptyHtmlDocument = "";
+    private String invalidHtmlDocument = "</div>";
+    private String validHtmlDocument = "<!DOCTYPE html><html><head><title>TITLE</title></head><body>CONTENT</body></html>";
+
+    @Test
+    public void itShouldProcessEmptyHTML() throws Exception {
+        final byte[] pdf = processor.process(emptyHtmlDocument.getBytes("UTF-8"), defaultPdfOptions);
+        assertThat(pdf.length, not(0));
```
malaffoon: This is fine but it isn't consistent with other tests, i would expect one of these:
```java
    assertThat(pdf).isNotEmpty();
    assertThat(pdf.length).isGreaterThan(0);
    assertThat(pdf.length).isPositive();
```

PR #77, opened 2017-05-09
Feature/groups api

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Subjects.java
```diff
@@ -0,0 +1,18 @@
+package org.opentestsystem.rdw.reporting.assessment;
+
+public final class Subjects {
+
+    private Subjects() {
```
malaffoon: nit: i tend to put the private ctor for utility classes at the bottom; the reason is that i want the user to see the useful stuff first (at the top). Also, i don't think utility classes are quite obvious enough without javadocs; i don't think documenting the ctor is necessary but it doesn't hurt to throw on a `hide utility class constructor`.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Subjects.java
```diff
@@ -0,0 +1,18 @@
+package org.opentestsystem.rdw.reporting.assessment;
+
+public final class Subjects {
```
agorina: Just curious: why does it need to be ```final```?

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Subjects.java
```diff
@@ -0,0 +1,18 @@
...
+    private Subjects() {
+    }
+
+    /**
+     * This method will normalize subject IDs into a standard non null format
+     *
+     * @param entityId the subject ID to normalize
+     * @return the value of subject ID or <code>0L</code> if the given subject ID is null
+     */
+    public static long entityIdToCode(final Long entityId) {
```
malaffoon: A convention i've seen is to use `safe` in the method to indicate it is eating nulls, e.g. `safeEntityId(entityId)`. Another alternative is `orValue`, so in this case `entityIdOrZero(entityId)` but that is kind of hokey. Not gonna lose sleep if you ignore me, just thinking "out loud".

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Subjects.java
esotrope: This is just a part of the utility class pattern - its to show that the class isn't intended for instance-level usage and extension/inheritance.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Subjects.java
esotrope: I like safeEntityId

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultGroupRepository.java
```diff
@@ -1,39 +1,37 @@
...
 import org.springframework.stereotype.Repository;
 
+import javax.validation.constraints.NotNull;
 import java.util.List;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 @Repository
-public class GroupRepositoryImpl implements GroupRepository {
+public class DefaultGroupRepository implements GroupRepository {
```
malaffoon: JdbcGroupRepository?  Again, just thinking out loud here.

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultGroupService.java
```diff
@@ -0,0 +1,26 @@
...
+public class DefaultGroupService implements GroupService {
+
+    private final GroupRepository groupRepository;
+
+    @Autowired
+    public DefaultGroupService(final @NotNull GroupRepository groupRepository) {
+        this.groupRepository = groupRepository;
+    }
+
+    public Set<Group> getGroups(final User user) {
```
agorina: Nit: probably ```@NotNull``` would be good here.

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultGroupRepository.java
esotrope: i can roll with that

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
```diff
@@ -1,11 +1,20 @@
...
+
+import com.google.common.base.MoreObjects;
 
 public class Group {
+
     private long id;
     private String name;
     private String schoolName;
-    private String subjectId; //TODO:?
+    private long subject;
```
agorina: This is really subjectId. Do you view it differently?

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultGroupService.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import javax.validation.constraints.NotNull;
+import java.util.Set;
+
+import static com.google.common.collect.Sets.newHashSet;
+
+@Service
+public class DefaultGroupService implements GroupService {
```
malaffoon: RepositoryBackedGroupService?  A convention developed at a company i used to work for.

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
malaffoon: Do we want to leave this subjectId? Just for clarity?

esotrope: yeah it is - but as far as the client is concerned this is just what a subject is - a numeric value. the client will never have to distinguish subjectCode from subjectId etc its just a numeric value to associate with a subject. I'm ok with either way though. Might be appropriate to make it subjectId because it kind of self-documents where the value comes from

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/GroupController.java
```diff
@@ -24,8 +21,8 @@ public GroupController(final GroupService service) {
     }
 
     @GetMapping
-    @ResponseStatus(HttpStatus.OK)
-    public List<Group> getGroups(@AuthenticationPrincipal final User user) {
-        return service.findAll(user);
+    public Set<Group> getGroups(@AuthenticationPrincipal final User user) {
```
malaffoon: No objection, just curious why? Sometimes Set's are more problematic than List's but if the order is meaningless/nonpredictable this makes sense.

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultGroupService.java
esotrope: RepositoryBacked, I feel like there should be a better way to say the same thing but maybe not

webapp/src/main/resources/application.sql.yml
```diff
@@ -7,6 +7,17 @@ sql:
             join school s on s.id = sg.school_id
             where user_login=:user_login
 
+    findAllForUsername: >-
+      select
+      	  sg.id,
```
malaffoon: nit: consistent indenting

webapp/src/main/resources/application.sql.yml
```diff
@@ -7,6 +7,17 @@ sql:
             join school s on s.id = sg.school_id
             where user_login=:user_login
```
agorina: The PR does not show the full query above, but I think it is pretty much the same. Should it be removed?

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/GroupController.java
esotrope: this is kind of an intent thing - conceptually you are looking up a set of groups is why. It also informs the client that the values are not sorted yet and the client needs to sort them. If we return a list it may be assumed that it is pre-sorted in some meaningful way

webapp/src/main/resources/application.sql.yml
esotrope: yeah I can clean it up

webapp/src/main/resources/application.sql.yml
esotrope: is that two indents for continuation?

webapp/src/test/java/org/opentestsystem/rdw/reporting/group/GroupRepositoryIT.java
```diff
@@ -32,7 +33,12 @@
...
-        assertThat(repository.findAll("user-login").size()).isEqualTo(2);
-        //TODO: add more verifications
+        assertThat(repository.findAllForUsername("user-login"))
+                .usingFieldByFieldElementComparator()
+                .containsAll(newArrayList(
+                        Group.builder().id(-2).name("IT Student Group 2").schoolName("Test School 1").subject(-1L).build(),
+                        Group.builder().id(-1).name("IT Student Group 1").schoolName("Test School 1").subject(-1L).build()
+                ));
     }
+
```
agorina: Suggest to add a test with NULL subject, just because there is an extra step in the Repository code that converts it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Subjects.java
malaffoon: +1; it seems to be a best practice to make utility classes `final` with a private ctor instead of `abstract`. Some folk consider `abstract` to mean, "this class is intended to be extended" so they get confused by abstract utility classes. And, if it isn't abstract, the ctor has to be private.

webapp/src/test/java/org/opentestsystem/rdw/reporting/group/GroupRepositoryIT.java
esotrope: good point

webapp/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultGroupService.java
malaffoon: Repositorified?  ;)

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Subjects.java
agorina: Thank you for the explanation.

webapp/src/main/resources/application.sql.yml
agorina: Thank you.

webapp/src/main/resources/application.sql.yml
malaffoon: Oh i see, the sg.name is a continuation of the fields. Hmm, personally i'd avoid that issue and do something like:
```sql
      select sg.id, sg.name, school.name as school_name, subject_id
        from user_student_group usg
          join ...
          join ...
        where user_login=:user_login
```

esotrope: 👍

PR #79, opened 2017-05-10
Feature/cleanup

esotrope:
A decent sized cleanup exercise

1. moved all security related code to the security package
2. moved translation related code to i18n package
3. split security config into two files security.saml.WebSecurityConfiguration and security.SecurityConfiguaration and added prePostEnabled=true for @PreAuthorize annotations
4. added final to method params
5. changed impl naming to our new standards
6. changed junit tests to use assertj conventions


webapp/src/main/java/org/opentestsystem/rdw/reporting/config/DataSourceConfiguration.java
```diff
@@ -26,13 +26,14 @@ public DataSource dataSource() {
...
-    @Bean(name = "queryJdbcTemplate")
-    public NamedParameterJdbcTemplate queryJdbcTemplate() {
+    @Bean
+    public NamedParameterJdbcTemplate namedParameterJdbcTemplate() {
         return new NamedParameterJdbcTemplate(dataSource());
     }
 
     @Bean
-    public PlatformTransactionManager txManager() {
+    public PlatformTransactionManager transactionManager() {
```
malaffoon: I've been wondering if this is actually necessary in the reporting stack. Since it is only reading data, can we not have Transactional? Then you wouldn't need this (nor the annotation that goes with it).

fyi, @agorina

agorina: If I recall it correctly, it is needed for Integration Tests. Maybe there is a better way to configure it. I think removing it will cause the test to be non-transactional. This needs to be tested. I maybe wrong about it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/WebSecurityConfiguration.java
```diff
@@ -459,10 +457,10 @@ public AuthenticationManager authenticationManagerBean() throws Exception {
         return super.authenticationManagerBean();
     }
 
-	@Override
-	public void configure(WebSecurity web) {
-		web.ignoring().antMatchers("/iris/**");
-	}
+    @Override
+    public void configure(WebSecurity web) {
+        web.ignoring().antMatchers("/iris/**");
```
bgsavage: do we still need this?

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationController.java
```diff
@@ -16,7 +16,7 @@
     private TranslationService translationService;
 
     @RequestMapping("/api/translations/{locale}")
-    public Map<String, String> getTranslations(@PathVariable Locale locale) {
+    public Map<String, String> getTranslations(final @PathVariable Locale locale) {
```
esotrope: copy that

webapp/src/main/java/org/opentestsystem/rdw/reporting/config/DataSourceConfiguration.java
esotrope: should we introduce a TestConfiguration.java for the "test" profile with this bean and @EnableTransactionManagement?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/WebSecurityConfiguration.java
esotrope: yeah I think so - there is a story out there to add the api server proxy to iris. This would be so we can wrap security around iris.

webapp/src/main/java/org/opentestsystem/rdw/reporting/config/DataSourceConfiguration.java
agorina: Makes sense. Let me know if you want me to do it.

esotrope: I'll take a shot at it and you can review/change it from there maybe

agorina: Sounds good. Let me know if you have any question.

esotrope:
I added ITDataSourceConfiguration.java to the test dir.
It doesn't appear to be picked up for tests automatically - I had to add @ContextConfiguration({ ITDataSourceConfiguration.java }) to the repository IT

webapp/src/test/java/org/opentestsystem/rdw/reporting/ITDataSourceConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+import javax.sql.DataSource;
+
+@Profile("test")
+@Configuration
+@EnableTransactionManagement
+@Import({DataSourceConfiguration.class})
+public class ITDataSourceConfiguration {
```
malaffoon: 👍

agorina: +1. Good change.

PR #80, opened 2017-05-10
Feature/dwr 302 my groups widget

bgsavage:
Used the existing GroupsComponent to be the widget which is placeable by the the `<groups></groups>` markup.

Upgraded from angular 2 to 4 which had some depedency changes (mainly anything called ng2-xxxx is now ngx-xxxx).  The translate component had some api changes that needed to be accommodated as well.  @angular/cli had a few changes as well (now environmentSource instead of environment: source, and now accepts the name of the config to be prefixed with a dot.)

Using primeng component as our data table of choice as it had the best documentation, easiest implementation, and best customization to make the grid our own without giving up total control to the vendor.


webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -1,55 +1,41 @@
...
+              <a routerLink="/groups/{{group.id}}/students">{{group.name}}</a>
+            </ng-template>
+          </p-column>
+          <p-column field="schoolName" header="{{'labels.school.name' | translate}}" [sortable]="true"></p-column>
+          <p-column field="subjectId" header="{{'labels.assessment.subject' | translate}}" [sortable]="true">
+            <ng-template let-group="rowData" pTemplate type="body">
+              {{ group.subjectId | subject }}
+            </ng-template>
+          </p-column>
+        </p-dataTable>
```
bgsavage: primeng datatable markup is here.  This allows us to inject our own templates should we need some cells to have clickable links or if we want to pipe in a filter.  This was a big selling point of primeng for me as other vendors didn't give you this level granularity.

webapp/src/main/webapp/src/app/app.module.ts
```diff
@@ -1,41 +1,45 @@
...
+import { StudentExamsResolve } from "./student-exams/student-exam.resolve";
+import { StudentExamItemsResolve } from "./student-exam-items/student-exam-items.resolve";
+import { GroupExamItemResolve } from "./group-exam-item/group-exam-item.resolve";
+import { BreadcrumbsComponent } from "./breadcrumbs/breadcrumbs.component";
+import { DataTableModule, SharedModule } from 'primeng/primeng';
+import { SubjectPipe } from "./shared/subject.pipe";
 
 export function createTranslateLoader(http: Http) {
-  return new TranslateStaticLoader(http, '/api/translations', '');
+  return new TranslateHttpLoader(http, '/api/translations/', '');
```
bgsavage: @esotrope Please make sure this is still what we want.  The new translate api required me to pull in this TranslateHttpLoader from a seperate package as it wasn't available in their core.

chdietz:
Why are we using primeng ? Or why do we need a data table?
Just playing devils advocate. Are we sure we want to use a data table instead of styled lists? 
Have you used this before?

webapp/src/main/webapp/src/app/shared/subject.pipe.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  transform(value: number): string {
+    let result = "";
+
+    if(value >= 0 && value < this.subjects.length)
+      return this.subjects[value];
+
+    return result;
+  }
+}
```
esotrope: later this will be driven from the translation service

webapp/src/main/webapp/src/styles.less
```diff
@@ -560,3 +582,10 @@ a.active {
   color: gray;
   text-decoration: none;
 }
+
+.row-eq-height {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display:         flex;
+}
```
esotrope: let's make Aaron aware of this override

webapp/src/main/webapp/src/app/app.module.ts
esotrope: what is the actual difference here?

PR #81, opened 2017-05-11
reporting cleanup round 2

esotrope:
1. cleaned up DefaultExamItemService & unit test 
2. moves controllers to web package.
3. creates assessment, exam, exam.report, exam.report.export packages and moves to these new packages.


webapp/src/main/java/org/opentestsystem/rdw/reporting/DataSourceConfiguration.java
```diff
@@ -1,4 +1,4 @@
-package org.opentestsystem.rdw.reporting.config;
+package org.opentestsystem.rdw.reporting;
```
malaffoon: Could this go in `org.opentestsystem.rdw.reporting.repository`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -0,0 +1,255 @@
...
+    private XPathExpression secondarySampleListPath;
+    private XPathExpression namePath;
+    private XPathExpression valuePath;
+    private XPathExpression descriptionPath;
+    private XPathExpression scorePointPath;
+    private XPathExpression sampleContentPath;
+    private XPathExpression formatPath;
+
+    @Autowired
+    public DefaultExamItemService(@NotNull final ArtifactRepository artifactRepository) throws XPathExpressionException, ParserConfigurationException {
```
malaffoon: It's not great when a constructor throws exceptions, especially checked exceptions. Yeah, i get that it needs to successfully build all the xpath expressions to be in a runnable state, and it should fast-fail but these errors just aren't going to happen. I would favor removing the expressions from the ctor signature, catching these exceptions and logging an error. Perhaps break out a helper method that a unit test could run to establish that there are no errors in the expressions.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -0,0 +1,255 @@
...
+        this.scorePointPath = xPath.compile("./@scorepoint");
+        this.sampleContentPath = xPath.compile("./samplecontent");
+        this.formatPath = xPath.compile("/itemrelease/item/@format");
+    }
+
+    @Override
+    public ExamItemScoringGuide getScoringGuide(final String itemNumber) {
+        try {
+            return getExamItemSolution(getXmlDocument(itemNumber), itemNumber);
+        } catch (IOException exception) {
```
malaffoon: nit: `final`

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -0,0 +1,255 @@
...
+            final String message = "Error reading the item xml for item number: " + itemNumber;
+            logger.warn(message, exception);
```
malaffoon: log-and-throw anti-pattern?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+// TODO: Implement authorization
```
agorina: Is this TODO done?

webapp/src/main/java/org/opentestsystem/rdw/reporting/DataSourceConfiguration.java
esotrope: I can do that but package doesn't exist anymore and the config is specific to the core application.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: I'll add the try catch log + unit test

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: is this the exception that should be final here?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: I think what is missing is service method level annotations for PII/GROUP_PII

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: yeah I'll try and change this to catch wrap and throw

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
bgsavage: I'm curious to know why we want to pre compile every xpath expression.  A lot of them may not even be used if the `itemFormatIsValidToHaveAnswerKey(document)` condition is not met.

esotrope: The idea here is it adds processing overhead to each method call when it may be called over and over again in the lifetime of the app. Generally if it doesn't change and if it's thread safe it makes sense to sac memory for performance and make an instance to reuse.

bgsavage: Oh right, over the lifetime of the app.  So Spring will by default resolve this class as a single instance unless we specify otherwise?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
malaffoon: yup

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -53,203 +52,165 @@
...
-            final String message = "Error reading the item xml for item number: " + itemNumber;
-            logger.warn(message, exception);
-            throw new NoSuchElementException(message);
-        } catch (SAXException exception) {
-            final String message = "Error parsing the item xml for item number: " + itemNumber;
-            logger.error(message, exception);
-            throw new NoSuchElementException(message);
+            return getExamItemSolution(getXmlDocument(id), id);
+        } catch (final IOException | SAXException | XPathExpressionException exception) {
+            throw new RuntimeException(String.format("Error reading the xml for item: %s", id), exception);
```
bgsavage: I thought we might have wanted to know the difference between a parsing error (bad formatted xml) and a read error (ie the item doesn't exist).

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -53,203 +52,165 @@
...
+                .build();
     }
 
-    private ScoringCriterion getRubric(final Node node, final String id) {
+    private ScoringCriterion createRubric(final Node node, final String id) {
         return ScoringCriterion.builder()
-                .name(getXmlValue(namePath, node))
-                .scorepoint(getXmlValue(scorePointPath, node))
-                .template(encodeImages(getXmlValue(valuePath, node), id))
+                .name(safeValue(node, namePath, ""))
```
bgsavage: Why have this default parameter if it's always going to be "" ?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -53,203 +52,165 @@
...
-                        .getEncoder()
-                        .encodeToString(ByteStreams.toByteArray(imageStream));
-
-                String mimeType = URLConnection.guessContentTypeFromName(imageFileName);
-                result = result.replace(imageFileName, "data:" + mimeType + ";base64," + encoding);
-            } catch (IOException e) {
-                logger.warn("Unable to find referenced image({}) for item number: {}", imageFileName, itemNumber, e.getMessage());
+                final String base64Image = Base64.getEncoder()
+                        .encodeToString(ByteStreams.toByteArray(
+                                artifactRepository.getArtifactFromItemDirectory(id, imagePath)
```
bgsavage: I had these as two lines because it's easier to debug and read... What value does this add by making it one line?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: leaves the door open for low level customization later. 
If we ever do more xml parsing I would move these safe* methods to an xml util or something is the idea

bgsavage: Can default be optional and defaulted to ""?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: the wrapped exception in the runtime exception should indicate where/how the exception occurred. Also, when running the debugger we can always evaluate sub-expressions of code so the need for "two-lining" diminishes I think

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: idea here is the wrapped exception should tell us the origin of the problem.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
```diff
@@ -53,203 +52,165 @@
...
-                    .range(0, rubricNodes.getLength())
-                    .mapToObj(rubricNodes::item)
-                    .map(node -> getRubric(node, id))
-                    .collect(Collectors.toList());
-
-        } catch (XPathExpressionException exception) {
-            logger.warn("Item xml contained a rubriclist element but no rubrics for id: " + id, exception);
-            return new ArrayList<>();
-        }
+    private boolean hasAnswerKey(final Node node) throws XPathExpressionException {
```
esotrope: oh ok - I'll revert the method name

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultExamItemService.java
esotrope: yeah that could be an api improvement.

PR #82, opened 2017-05-11
Feature/dwr 302 my groups widget

bgsavage:
Adds unit tests and removes scrollable for grid since it's not sure we need it at this time.


webapp/src/main/webapp/src/app/groups/groups.components.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    }));
+
+    it('should search and match any part of the string', async(() => {
+      setSearchValue("eng");
+      expect(fixture.componentInstance.filteredGroups.length).toBe(1);
+    }));
+
+    it('should search and not be case sensitive', async(() => {
+      setSearchValue("math");
+      expect(fixture.componentInstance.filteredGroups.length).toBe(3);
```
esotrope: is it possible to test search outside the groups component? in case we reuse elsewhere?

webapp/src/main/webapp/src/app/groups/groups.components.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    }));
+
+    function setSearchValue(value: string) {
+      let input = fixture.debugElement.query(By.css('#search')).nativeElement;
+
+      input.value = value;
+      input.dispatchEvent(new Event('input'));
+    }
```
esotrope: angular doesn't do this for us with the [()] () [] bindings?

webapp/src/main/webapp/src/app/shared/subject.pipe.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    expect(actual).toBe("ELA");
+  }));
+
+  it('should return empty string if subjectId is out of range', async(() => {
+    let target = new SubjectPipe();
+    let actual = target.transform(23);
+
+    expect(actual).toBe("");
+  }));
+});
```
esotrope: hm not sure we should add tests for language specific things that can change. translation/display text change would break the test

webapp/src/main/webapp/src/app/groups/groups.component.ts
```diff
@@ -8,8 +8,9 @@ import {DataService} from "../shared/data.service";
 export class GroupsComponent implements OnInit {
 
   private groups = [];
-  private filteredGroups = [];
-  private searchTerm : string;
+
+  filteredGroups = [];
+  searchTerm : string;
```
esotrope: in ts is the default private? i'm wondering what the difference is when we do or don't specify the access modifier here

webapp/src/main/webapp/src/app/groups/groups.components.spec.ts
bgsavage: I don't think it does in the unit tests, it wasn't working without manually dispatching the event.  My other option was to call the onSearchChange in the test, but that doesn't seem right either.

webapp/src/main/webapp/src/app/shared/subject.pipe.spec.ts
bgsavage: Since we are having to hardcode this in the front end, I wanted to make sure we captured at least the "all" logic.

webapp/src/main/webapp/src/app/groups/groups.components.spec.ts
bgsavage: I thought about it but I think I would have to pass 3 parameters to get it to work, at which point I have to ask is it worth creating a shared method for?

webapp/src/main/webapp/src/app/groups/groups.component.ts
bgsavage: Yep you caught me.  It becomes public.  At this point I'm not really sure how to access the ng-model filteredGroups otherwise.  It's kind of weird that the template can access private properties.  I don't need access to the searchTerms, and I can get around not accessing the filteredGroups by checking how the dom is changed, but then I would have too much knowledge about the data table creates the rows.  I'm not really sure what was best here, but I did notice on the angular guide for testing they also didn't specify access modifiers on their properties:  https://angular.io/docs/ts/latest/guide/testing.html

I'm not sure if that was just to keep the tests simple or it was intentional.  I think this will definitely evolve over time as we get better with it.

webapp/src/main/webapp/src/app/groups/groups.components.spec.ts
esotrope: ok - interesting

webapp/src/main/webapp/src/app/shared/subject.pipe.spec.ts
esotrope: eventually I imagine doing a lookup like this in a template though: 

{{ 'group.subjects.' + subject | translate}}

where translation json is like:
{
  "group.subjects.0": "MATH, ELA",
  "group.subjects.1": "MATH",
  "group.subjects.2": "ELA"
}

webapp/src/main/webapp/src/app/groups/groups.component.ts
esotrope: good to know - i don't think there is an issue making things public anyway - it shouldn't affect anything really since we never manually create and access these component properties manually

PR #83, opened 2017-05-11
Updated karma to now run tests using phantoms instead of chrome.

bgsavage:


webapp/src/main/webapp/karma.conf.js
```diff
@@ -7,7 +7,7 @@ module.exports = function (config) {
     frameworks: ['jasmine', '@angular/cli'],
     plugins: [
       require('karma-jasmine'),
-      require('karma-chrome-launcher'),
+      require('karma-phantomjs2-launcher'),
```
esotrope: is phantomjs like a headless browser?

bgsavage: yeah so we don't have to launch chrome and the build machine can run it.

esotrope: awesomesauce

PR #87, opened 2017-05-12
First pass at the my groups page.  

bgsavage:
This mainly just markup and css.

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/results/group-results.component.html
+++ b/webapp/src/main/webapp/src/app/groups/results/group-results.component.html
@@ -1,0 +1,67 @@
+<div class="pad-top-20 container">
+  <h2><i class="fa fa-group color-green"></i> {{'labels.groups.name' | translate}}</h2>
+  <div class="panel panel-default">
+    <div class="panel-body">
+      <div class="col-md-4">
+        <label for="select-group">{{'labels.groups.results.select-group' | translate}}</label>
+
+        <!--type ahead? -->
+        <select id="select-group" class="dropdown form-control">
+          <option *ngFor="let group of groups">{{group.name}}</option>
+        </select>
+      </div>
+      <div class="col-md-3">
+        <label for="school-year">{{'labels.groups.results.select-year' | translate}}</label>
+
+        <!-- static or data driven? -->
+        <select id="school-year" class="dropdown form-control">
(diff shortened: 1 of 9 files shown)


webapp/src/main/webapp/src/scripts/iris/util_xdm.js
```diff
@@ -114,8 +114,12 @@ Util.XDM = (function ($) {
     // console.log('MESSAGE RECIEVED', evt);
 
     try {
+      if(!evt.data.data)
+        return;
+
       var data = XDM.deserialize(evt.data);
     } catch (ex) {
+
```
esotrope: Uh oh, what made us have to do this?
This should probably be filed as an issue in the IRIS repo for OSU unless its just a debug setting that slipped in

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
```diff
@@ -0,0 +1,67 @@
...
+      </div>
+      <div class="col-md-3">
+        <label for="school-year">{{'labels.groups.results.select-year' | translate}}</label>
+
+        <!-- static or data driven? -->
+        <select id="school-year" class="dropdown form-control">
+          <option>2017-2018</option>
+          <option>2016-2017</option>
+          <option>2015-2016</option>
+          <option>2014-2015</option>
```
esotrope: +todo: should be populated from controller/component.ts and "formatted"

webapp/src/main/webapp/src/scripts/iris/util_xdm.js
bgsavage: It's intercepting messages from webpack which are in a different format.  This causes the console log error in your browser / running unit tests.

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
bgsavage: Yep working on generating this list.  Eventually this would come from the API i think?

agorina: Yes, that should come from the API. We need to decided what the API will return. Just a note:  we store only the last digit of the school year. For example: 2014-2015 is stored in the db as 2015.

PR #88, opened 2017-05-13
adds group exam components for my groups page /api/groups/{id}/latest…

esotrope:
…exam api endpoint

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
@@ -1,0 +1,94 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+/**
+ * Represents an assessment taken by students
+ */
+public class Assessment {
+
+    private long id;
+    private String name;
+    private long gradeId;
+    private List<Long> cutPoints;
+
+    private Assessment() {
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * @return the assessment's entity ID
+     */
+    public long getId() {
+        return id;
+    }
+
+    /**
+     * @return the assessment's display name
+     */
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * @return the assessment's target grade level entity ID
+     */
+    public long getGradeId() {
+        return gradeId;
+    }
+
+    /**
+     * Assessment cut points represent the various scale scores that separate one performance level from another.
+     * This list will start with the minimum scale score and end with the maximum scale score and contain all other
+     * scale score breaking points in between in ascending order.
+     *
+     * @return the assessment's cut points
+     */
+    public List<Long> getCutPoints() {
+        return cutPoints;
+    }
+
+    public static class Builder {
+
+        private long id;
+        private String name;
+        private long gradeId;
+        private List<Long> cutPoints;
+
+        public Builder id(final long id) {
+            this.id = id;
+            return this;
+        }
+
+        public Builder name(final String name) {
+            this.name = name;
+            return this;
+        }
+
+        public Builder gradeId(final long gradeId) {
+            this.gradeId = gradeId;
+            return this;
+        }
+
+        public Builder cutPoints(final List<Long> cutPoints) {
+            this.cutPoints = cutPoints;
+            return this;
+        }
+
+        public Assessment build() {
+            final Assessment assessment = new Assessment();
+            assessment.id = id;
+            assessment.name = name;
+            assessment.gradeId = gradeId;
+            assessment.cutPoints = cutPoints != null ? ImmutableList.copyOf(cutPoints) : ImmutableList.of();
+            return assessment;
+        }
+
+    }
+
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultGroupExamService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultGroupExamService.java
@@ -1,0 +1,26 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+import org.opentestsystem.rdw.reporting.security.GroupGrant;
+import org.opentestsystem.rdw.reporting.security.User;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import javax.validation.constraints.NotNull;
+
+@Service
+public class DefaultGroupExamService implements GroupExamService {
+
+    private final GroupExamRepository repository;
+
+    @Autowired
+    public DefaultGroupExamService(@NotNull final GroupExamRepository repository) {
+        this.repository = repository;
+    }
+
+    @Override
+    public GroupExam getLatestGroupExam(@NotNull final User user, final long groupId) {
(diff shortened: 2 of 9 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Represents an assessment taken by students
+ */
+public class Assessment {
+
+    private long id;
+    private String name;
+    private long gradeId;
+    private List<Long> cutPoints;
```
agorina: There are three types of assessments: IAB, ICA and Summative assessments. IAB and others have different attributes. IAB does not have Calims, and I think we do not get cut points for it.  This needs to be reflected somehow here.

Another question: where do we need cutPoints now? Just trying to remember.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Assessment cut points represent the various scale scores that separate one performance level from another.
+     * This list will start with the minimum scale score and end with the maximum scale score and contain all other
+     * scale score breaking points in between in ascending order.
+     *
+     * @return the assessment's cut points
+     */
+    public List<Long> getCutPoints() {
```
agorina: There is a concept of the ```cut scores``` that is stored in the db and is delivered to the system with the assessment packages. Is this the same? Looks like this is something different?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultGroupExamService.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Autowired
+    public DefaultGroupExamService(@NotNull final GroupExamRepository repository) {
+        this.repository = repository;
+    }
+
+    @Override
+    public Optional<GroupExam> getLatestGroupExam(@NotNull final User user, final long groupId) {
+        final GroupGrant group = user.getGroupsById().get(groupId);
+        return repository.findLatestForGroupAndSubject(group.getId(), group.getSubjectId());
```
agorina: I am just trying to understand the UI flow.  I would think that the UI steps would be:
1. get all the data for the drop-down filters on the screen: get all subjects for the group, get all the assessments for the group, along with their types and most recent date. 
2.then choose one assessment (based on the group's subject). A group could have multiple subject. In this case default to one.
3. Using the chosen assessment : get all the sessions for the group, along with the dates.
4. finally get all the exams for the session.

If my logic is correct, then this method would have different input parameters. 
Could you please clarify?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
jtreuting: Looks like these are the breakdown of Below Standard, Near Standard and Above Standard.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultGroupExamService.java
jtreuting: I think the flow you described makes sense and I do think it will be easier to sit down and talk it through.

To answer the questions you posed: 
1) If the most recent is summative (or ICA) then that is what is shown by default.  and 
2) I think it depends on which selection you are speaking about.  Via email with Aaron based on some questions from Brandon, he was hoping that the Advanced Filters could be done client side instead of server side, but not sure that is possible (have to think thru the aggregate and All opportunities I think).  If the group or school year are changed then it is like reloading the page I would think.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
agorina: I do not think we get these data for IAB. We should chat about it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultGroupExamService.java
chdietz: As to the most recent exam this gets interesting: What if most of the class is taking IAB but one student transferred and took an ICA. So:

1. Regardless of most recent overall should we default to IAB vs ICA or Summative?
2. What if there are ICA and no IAB?
3. I think this is a usability issue and not a technical challenge.

chdietz: As far as getting drop downs there is a usability issue:

1. Should drop downs be filtered relevant to the data in the group for the school year?
2. Should also possible options be displayed but not available be grayed out?
3. Should all possible options be displayed regardless of available data?
4. I remember Aaron discussing this and not sure where we landed.  Options 1 and 2 add some complexity in queries but what is the ideal design for usability?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/GroupExam.java
```diff
@@ -0,0 +1,34 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+/**
+ * This class represents a group's exams for a specific assessment broken into sessions
+ */
+public class GroupExam {
```
chdietz: General question about domain objects:
It looks like the strategy is to have objects such as GroupExam that align to views that need these objects instead of a common domain model that all views need to use.  I usually prefer the strategy of objects that align to what the view needs which seems to be where we are going, I would like to confirm what direction we are following.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
esotrope: Right now the model is a WIP.
We will eventually need two models to represent IAB and ICA/Summative or one model with some optional fields.

The cut points represents the ICA/Summative score points that separate performance levels:
(e.g.
2000 -mimimum
2000-2400 - level1/below
...ect)
This list would store somethings like (2000, 2400, 2500, 2600, 2700)

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/GroupExam.java
esotrope: This model is aligned with a view - some underlying models are in a more generic state.
I think it would be good to reuse existing models and when an existing model doesn't fit the view make a new model to rearrange the data for that view/endpoint

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/DefaultGroupExamService.java
esotrope: This is a great point about the query - we would need some sort of selection logic after searching for the most recent IAB and ICA/Summative if we choose to go that way.

Some things I had in mind when designing this:

1. The view defaults to show the known selected group and school years - later these years can be populated by an endpoint that supplies relevant years
2. The selected assessments tab/advance filter options default to hidden. My assumption is that we make the query for the relevant assessments when the user opens it and not upfront. Also we don't necessarily need to know this information before looking for the most recent sessions.

PR #90, opened 2017-05-16
Feature/dwr 353 mygroups page

bgsavage:
My groups page basic filters only.  This will continue to evolve as we get eyes on this feature and understand the usability a bit better.  

- I added chrome launcher back as a secondary running to facilitate debugging unit tests.
- Route changes occur on group change and school year change.
- School year route param is optional and is set to the current year by default.


bgsavage:
@aaronsleeper  I added you for visibility into the less changes per our recent discussions on how we should name our color classes.  I understand these will change anyway when I start pulling in the ui-kit.

webapp/src/main/webapp/src/app/shared/data.service.ts
```diff
@@ -62,9 +66,30 @@ export class DataService {
...
+      }, () => {
+        observable = this.getRecentAssessmentBySchoolYear(groupId, schoolYear);
+      });
+
+      return observable;
+    }
+    else {
+      return this.getRecentAssessmentBySchoolYear(groupId, schoolYear);
+    }
+  }
```
esotrope: can we have the component resolve the years before it makes this request to make things simpler?

webapp/src/main/webapp/src/app/shared/staticData.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  private cache: { [key: string]: any } = {};
+
+  constructor(private http: Http) {
+  }
+
+  private get(url: string, options?: RequestOptionsArgs): Observable<any> {
+    if(this.cache[url]){
+      return this.cache[url]
+    }
```
esotrope: might want to make this parametrized - may not want to always cache all requests - not sure.
also we will want to rename "get" to "getJson" since it maps to json.

webapp/src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -359,3 +359,23 @@ export const mock_rubrics = [ {
...
+    id: 2015,
+    isCurrent: false
+  }, {
+    id: 2014,
+    isCurrent: false
+  }, {
+    id: 2013,
+    isCurrent: false
+  }
+];
```
esotrope: the component should make this model - the backend will return [2017, 2016...]

webapp/src/main/webapp/src/app/shared/data.service.ts
```diff
@@ -62,9 +66,30 @@ export class DataService {
...
+    params.set('ssid', searchFilter.ssid);
 
     return this.get('/students/search', { search: params });
   }
 
+  getMostRecentAssessment(groupId: number, schoolYear?: number) {
+    if (schoolYear == undefined) {
+      let observable = {};
+
+      this.staticDataService.getSchoolYears().subscribe(x => {
```
esotrope: not sure what static data service is for - can we have it go through the dataservice / mock response layer

bgsavage: This still goes through the mock response layer.  Static Data Service is basically data that is cacheable (ie in this case school years).  I think any "static" data which expect to not to change should go through a different service in which we can cache the results to avoid redundant api calls to it.

webapp/src/main/webapp/src/app/standalone/data/data.ts
bgsavage: How will the UI know what the current school year is?  Isn't this bus. logic we want the API to own?

webapp/src/main/webapp/src/app/shared/staticData.service.ts
bgsavage: If it shouldn't be cached, then it should go through the DataService instead.

webapp/src/main/webapp/src/app/shared/data.service.ts
bgsavage: I'd like to keep the component as light weight as possible and defer logic to the services.  I have a TODO in here to eventually break out all the methods we have in the data service to move to individual service.  Services are way easier to test than components.

webapp/src/main/webapp/src/app/shared/data.service.ts
esotrope: ok cool i missed that that was happening looking at the files. Can we call it cachingDataService then? if it was truly static i feel like it would be stored in a local file or something.

bgsavage: Sure I can do that, I thought about calling it a CachingDataService as first but wasn't sure if we wanted to leak impl details about how we are resolving the data.  But it could help developers to know which one they should be using.

webapp/src/main/webapp/src/app/standalone/data/data.ts
esotrope: I was thinking the api return an ordered list where the first entry is always the current year.
then which year is selected becomes an internal working of the component that the api is not concerned with.

malaffoon: 👍   (ordered list)

PR #91, opened 2017-05-16
adds /examyears /assessments endpoints.

esotrope:
1. Adds /examyears stub endpoint to return current year followed by previous four years
2. Adds /groups/{groupId}/years/{year}/assessments endpoint to return the latest assessment for a given year
3. Changes user object to expose permissions as map for easy lookup
4. Changes auth service to return set of permissions for simplicity


webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Optional;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+@Service
+public class DefaultGroupAssessmentService implements GroupAssessmentService {
+
+    private static final Comparator<GroupExam> groupExamCompletedDateTimeComparator = (a, b) -> {
+        if (a.getCompletionDateTime().isBefore(b.getCompletionDateTime())) {
+            return -1;
```
bgsavage: Java doesn't have a generic DateComparator that you could pass `getCompletionDateTime()` to?

esotrope: i did a quick look and didn't find a property based comparator factory thing so i just made it for now. but that would be nice if we could leverage a library for this

webapp/src/main/resources/application.sql.yml
```diff
@@ -14,6 +14,46 @@ sql:
             join student_group sg on sg.id = usg.student_group_id
           where user_login=:user_login
 
+    assessment:
+      findLatest: >-
```
bgsavage: might be good to put this under a parent `exam` tag similar to `iab`

esotrope: yeah i think it should be group.exam.findLatest now that you point that out

webapp/src/main/resources/application.sql.yml
```diff
@@ -14,6 +14,46 @@ sql:
...
+            a.id as assessment_id,
+            a.name as assessment_name,
+            a.grade_id as assessment_grade_id,
+            a.type_id as assessment_type_id,
+            a.subject_id as assessment_subject_id
+          from exam e
+            join student_group_membership sgm on e.student_id=sgm.student_id
+            join asmt a on e.asmt_id=a.id
+            join school s on e.school_id=s.id
+          where year(cast(e.completed_at as date))=:year
```
bgsavage: I'm curious to know how well this performs on a large data set.  Does casting here still allow us to hit our index?

webapp/src/test/java/org/opentestsystem/rdw/reporting/exam/StubExamYearServiceTest.java
```diff
@@ -0,0 +1,16 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+import org.junit.Test;
+
+import java.util.List;
+
+public class StubExamYearServiceTest {
+
+    @Test
+    public void test() throws Exception {
```
bgsavage: I'm guessing this is todo?

webapp/src/main/resources/application.sql.yml
esotrope: no clue.
This implementation also assumes the year parameter here is no longer school_year but calendar year.
 @agorina @malaffoon @rcordeau thoughts?

webapp/src/test/java/org/opentestsystem/rdw/reporting/exam/StubExamYearServiceTest.java
esotrope: let me remove this

webapp/src/main/resources/application.sql.yml
esotrope: after talking with JT - I think this line needs to change to

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupAssessmentController.java
```diff
@@ -0,0 +1,46 @@
...
+public class GroupAssessmentController {
+
+    private final GroupAssessmentService service;
+
+    @Autowired
+    public GroupAssessmentController(@NotNull final GroupAssessmentService service) {
+        this.service = service;
+    }
+
+    @GetMapping("/api/groups/{groupId}/schoolYears/{schoolYear}/assessments")
```
malaffoon: I feel that this URL is not very RESTful. This implies that there is a groups resource that contains years which contain assessments. Really, the resource is an Assessment and the list of assessments is being filtered by group and year. So it would be more like `/api/assessments?year={schoolYear}&group={groupId}`. Willing to discuss offline.

esotrope: That sounds right to me too Mark.
Our rational so far was just that required params be part of the path - but i like you're reasoning

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StubSchoolYearService.java
```diff
@@ -0,0 +1,27 @@
...
+public class StubSchoolYearService implements SchoolYearService {
+
+    /**
+     * @return list containing the current year followed by the four previous years
+     */
+    @Override
+    public List<Integer> getSchoolYears() {
+        final int to = Year.now().getValue();
+        final int from = to - 5;
+        return IntStream.range(from, to)
```
malaffoon: This seems oddly convoluted, why not just return the range?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StubSchoolYearService.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Service
+public class StubSchoolYearService implements SchoolYearService {
+
+    /**
+     * @return list containing the current year followed by the four previous years
+     */
+    @Override
+    public List<Integer> getSchoolYears() {
+        final int to = Year.now().getValue();
```
malaffoon: Personal opinion: for the stub service, just hardcode don't use now(). i.e.
```java
    private final List<Integer> Years = newArrayList(2014, 2015, 2016, 2017, 2018);
```
Or, if that really worries you, make it a property. 
```java
    @Value("${stub.years:"2014,2015,2016,2017,2018"}}
    private List<Integer> Years;
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
malaffoon: At the very least you can use `a.compareTo(b)`.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StubSchoolYearService.java
esotrope: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
esotrope: yes that would have definitely been better. 
I ended up with Comparator.comparing(GroupExam::getCompletionDateTime) which is best i think

PR #93, opened 2017-05-17
adds user endpoint for ui security/ username display

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/web/UserController.java
```diff
@@ -0,0 +1,17 @@
...
+import org.opentestsystem.rdw.reporting.security.User;
+import org.opentestsystem.rdw.reporting.web.response.BasicUserDetails;
+import org.springframework.security.core.annotation.AuthenticationPrincipal;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class UserController {
+
+    @GetMapping("/api/user")
```
agorina: Is there any security magic involved here?
Can a user only see his/her details?

malaffoon: It looks like it is reflecting the current user.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/response/BasicUserDetails.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+
+/**
+ * Represents a user's basic display and security information.
+ * This allows the UI to display the user name and show/hide content appropriately.
+ */
+public class BasicUserDetails {
+
+    private final User user;
```
malaffoon: This is fine but, due to being burned it a past life, i would copy the values out of the User and not keep a reference around.

esotrope: will do.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/UserController.java
esotrope: yeah @AuthenticationPrincipal gets the current user object out of a spring managed thread-local so it is not possible to get another user's user object here

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/response/BasicUserDetails.java
```diff
@@ -0,0 +1,47 @@
...
+     * @return the user's family name
+     */
+    public String getLastName() {
+        return user.getLastName();
+    }
+
+    /**
+     * @return all statewide permission IDs of the user
+     */
+    public Set<String> getStatewidePermissions() {
```
bgsavage: are these the only permissions we care about?

esotrope: hm actually - this won't be sufficient - you need to know if a user has lower level permissions.
I don't think the UI needs to know what school and district IDs the permissions are scoped to though. I cant see using that in practice can you?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/UserController.java
agorina: Cool, thank you for explaining it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/response/BasicUserDetails.java
bgsavage: Nope but if a user does try to access unauthorized district / school the api will need to return a generic error that the ui can intercept and show a generic authorization msg for.

PR #94, opened 2017-05-17
Feature/dwr 348 ui recent exam

bgsavage:
Adds the assessment-results component which only has a title, sessions, instructions, and is collapsible.


webapp/src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -99,42 +99,47 @@ export const assessments = [
...
     academicYear: 2014,
-    subject: AssessmentSubjectType.MATH
+    subject: AssessmentSubjectType.MATH,
+    sessions: [{id:"oat-01", dateTime:"2017-05-12T21:47:45Z"}]
   }
 ].map((assessment : any) => {
-  assessment.fullName = `Grade ${assessment.grade} ${assessment.name}`
+  assessment.fullName = `Grade ${assessment.grade} ${assessment.name}`;
   return assessment;
 });
```
esotrope: eventually I think this will look like this:
`
{
  assessment: {id, name, gradeId, typeId, subjectId, schoolYear},
  sessions: [
    {
       id: 'ma-01',
       dateTime: ...
       exams: [{ id:...}]
    }
  ]
}
`

PR #95, opened 2017-05-19

bgsavage:


webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -1,15 +1,70 @@
...
-  templateUrl: './assessment-results.component.html'
+  templateUrl: './assessment-results.component.html',
+  animations: [
+    trigger(
+      'fadeAnimation',
+      [
+        transition(
+          ':enter', [
+            style({ opacity: 0 }),
+            animate('500ms ease-in', style({ opacity: 1 }))
```
bgsavage: I wanted to do an translate animation instead of fade, but I couldn't get it to look right so I settled for a fade animation.  I felt this was needed to make the collapse / uncollapse feel smoother.  Otherwise it's kind of sudden and a bit disrupting.  @aaronsleeper  maybe have a better ideas to achieve a smoother transition in css.

webapp/src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -136,76 +232,91 @@ export const assessments = [
...
+    }]
   }
 ].map((assessment : any) => {
   assessment.fullName = `Grade ${assessment.grade} ${assessment.name}`;
   return assessment;
 });
 
-export const exams_of_student = [
+// Just to keep the old UI working for now...
+export const DEPRECATED_exams_of_student = [
```
bgsavage: This is just to keep the old UI working for reference right now.  It's needed because we changed the model in which exams used to reference assessments but now assessments have a collection of exams (indirectly through the sessions collection)

webapp/src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -109,7 +181,19 @@ export const assessments = [
     gradeId: 4,
     academicYear: 2013,
     subject: AssessmentSubjectType.MATH,
-    sessions: [{id:"md-01", dateTime:"2017-05-12T21:47:45Z"},{id:"md-02",dateTime:"2017-05-11T21:47:45Z"},{id:"md-03",dateTime:"2017-05-10T21:47:45Z"}]
+    sessions: [{
+      id:"md-01",
+      dateTime:"2017-05-12T21:47:45Z",
+      exams: exams_of_student.filter(x => x.id % 3 == 0)
```
bgsavage: This is still a TODO on the api, I just built this based on previous discussions.  @esotrope  let me know if this changes when you implement it.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
esotrope: Yeah we should also implement this in css I think.
We should just change the css class of the object and let css transitions take over no?

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -1,15 +1,70 @@
...
+  ],
 })
 
 export class AssessmentResultsComponent implements OnInit {
+  private _assessment;
+  private _examSessions = [];
 
   @Input()
-  assessment;
+  set assessment(assessment) {
```
bgsavage: I'm starting to think more and more we should be using getters / setters to expose private properties similar to how we do in other OO languages.  Even though it seems the templates can access the private properties without setters, I think it'll be cleaner to treat it as though they can't so that way we know what's internal and what's publicly available.  This also helps in unit testing.  In this case I'm taking advantage of the setter to do some processing.

webapp/src/main/webapp/src/app/standalone/data/data.ts
esotrope: yeah I have it slightly different -> top level object should have assessment field with assessment info and session field is in top level object instead of assessment object
{assessment:{}, sessions:[]}

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
esotrope: I'm ok with this - originally we didn't intend on adding any ui unit tests is why this wasn't considered. i would suggest we really only add unit tests for critical ui component functionality - like if we think we have some tricky logic in a component and want to cover all its edge cases. otherwise ui testing seems kind of overkill

PR #96, opened 2017-05-19
Feature/group assessments

esotrope:
1. Adds API endpoint for getting all assessments for a given group in a given school year
2. Simplifies existing "get latest group assessment" API endpoint by not accommodating broader assessment search in the same method

(!) new endpoints are as follows:
/api/groups/{groupId}/assessments?schoolYear={schoolYear}
/api/groups/{groupId}/latestassessment?schoolYear={schoolYear}

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
@@ -1,5 +1,7 @@
 package org.opentestsystem.rdw.reporting.assessment;
 
+import java.util.Objects;
+
 /**
  * Represents an assessment taken by an individual or group
  */
@@ -10,6 +12,7 @@ public class Assessment {
     private long gradeId;
     private long typeId;
     private long subjectId;
+    private long schoolYear;
 
     /**
      * @return the assessment entity ID
@@ -46,6 +49,35 @@ public class Assessment {
         return subjectId;
     }
 
+    /**
+     * @return the school year the assessment was designed for
+     */
+    public long getSchoolYear() {
+        return schoolYear;
+    }
+
+    /**
+     * This allows IAB and ICA/Summative exam's to not collide in {@link java.util.Set}s
+     *
+     * @param o the object to compare
+     * @return <code>true</code> if the given object is an Assessment and has the same entity ID and type ID
+     */
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        final Assessment that = (Assessment) o;
+        return id == that.id && typeId == that.typeId;
+    }
+
+    /**
+     * @return hash code based on the Assessment's entity ID and type ID
+     */
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, typeId);
+    }
+
     public static Builder builder() {
         return new Builder();
     }
@@ -57,6 +89,7 @@ public class Assessment {
         private long gradeId;
         private long typeId;
         private long subjectId;
+        private long schoolYear;
 
         public Builder id(final long id) {
             this.id = id;
@@ -83,6 +116,11 @@ public class Assessment {
             return this;
         }
 
+        public Builder schoolYear(final long schoolYear) {
+            this.schoolYear = schoolYear;
+            return this;
+        }
+
         public Assessment build() {
             final Assessment assessment = new Assessment();
             assessment.id = id;
@@ -90,6 +128,7 @@ public class Assessment {
             assessment.gradeId = gradeId;
             assessment.typeId = typeId;
             assessment.subjectId = subjectId;
+            assessment.schoolYear = schoolYear;
             return assessment;
         }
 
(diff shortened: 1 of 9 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupAssessmentController.java
```diff
@@ -26,21 +24,23 @@ public GroupAssessmentController(@NotNull final GroupAssessmentService service)
...
+    @GetMapping("/api/groups/{groupId}/latestassessment")
+    public Assessment getLatestGroupAssessment(
             @AuthenticationPrincipal final User user,
             @PathVariable final long groupId,
-            @PathVariable final long schoolYear,
-            final HttpServletRequest request) {
+            @RequestParam("schoolYear") final long schoolYear) {
+
+        return service.getLatestAssessment(user, groupId, schoolYear)
+                .orElseThrow(NoSuchElementException::new);
```
bgsavage: I think I need to catch this exception and show a "No results found" message on the ui

esotrope: correct - also this will come back as an HTTP 404 for you

webapp/src/main/resources/application.sql.yml
```diff
@@ -14,6 +14,42 @@ sql:
...
+          where e.school_year=:school_year
+          and sgm.student_group_id=:group_id
+          and (0=:group_subject_id or a.subject_id=:group_subject_id)
+          and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+
+      iab:
+        findAll: >-
+          select
+              distinct a.id,
+              a.name,
```
agorina: I believe you need a label, not a name. I think this needs to be confirmed with @jtreuting, but I recall Brandt saying that we should use a label.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
```diff
@@ -56,4 +59,32 @@
...
+
+        return template.query(
+                findAllQuery,
+                ImmutableMap.<String, Object>builder()
+                        .put("school_year", schoolYear)
+                        .put("group_id", groupId)
+                        .put("group_subject_id", groupSubjectId)
+                        .put("statewide", permissionScope.isStatewide())
+                        .put("district_ids", permissionScope.getDistrictIds().isEmpty() ? -1 : permissionScope.getDistrictIds())
+                        .put("school_ids", permissionScope.getInstitutionIds().isEmpty() ? -1 : permissionScope.getDistrictIds())
```
agorina: The review does not show it, but this section is repeated in this file twice and I expect it to be repeated a lot. I would suggest creating a helper for this.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
```diff
@@ -46,6 +49,35 @@ public long getSubjectId() {
...
+     *
+     * @param o the object to compare
+     * @return <code>true</code> if the given object is an Assessment and has the same entity ID and type ID
+     */
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        final Assessment that = (Assessment) o;
+        return id == that.id && typeId == that.typeId;
```
agorina: This highly depends on the knowledge of how the data is stored. I understand why you are doing it, but I feel it would be safer to compare all the attributes in this class.

webapp/src/main/resources/application.sql.yml
```diff
@@ -14,6 +14,42 @@ sql:
...
+            a.school_year
+          from exam e
+            join student_group_membership sgm on e.student_id=sgm.student_id
+            join asmt a on e.asmt_id=a.id
+            join school s on e.school_id=s.id
+          where e.school_year=:school_year
+          and sgm.student_group_id=:group_id
+          and (0=:group_subject_id or a.subject_id=:group_subject_id)
+          and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+
```
agorina: I am concerned about ```distinct```. It is usually one of the  the slowest function. This could be rewritten with ```EXISTS``` that theoretically should perform way better. 
```
select a.id,
  a.name,
  a.grade_id,
  a.type_id,
  a.subject_id,
  a.school_year
from asmt a where exists (
    select e.asmt_id
    from exam e
    where
      e.school_year=:school_year
      and a.id = e.asmt_id )
```

But until we have a large data set to try it, it is hard to say for certain.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
esotrope: Will do.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
esotrope: 👍

webapp/src/main/resources/application.sql.yml
esotrope: thanks - I'll try this out

agorina: thank you

PR #98, opened 2017-05-22
adds group security context to simplify repo api. adds improved find …

esotrope:
…all assessments of group query. adds common utils for dealing with empty collections in sql

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Queries.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Queries.java
@@ -7,7 +7,7 @@ public final class Queries {
     private Queries() {
     }
 
-    public static Object nullOrEmptyToDefault(final Collection collection, final Object defaultValue) {
+    public static <T> Collection<T> nullOrEmptyToDefault(final Collection<T> collection, final Collection<T> defaultValue) {
         if (collection == null || collection.isEmpty()) {
             return defaultValue;
         }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Queries.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Queries.java
@@ -1,0 +1,17 @@
+package org.opentestsystem.rdw.reporting.common.jdbc;
+
+import java.util.Collection;
+
+public final class Queries {
+
+    private Queries() {
+    }
+
+    public static Object nullOrEmptyToDefault(final Collection collection, final Object defaultValue) {
+        if (collection == null || collection.isEmpty()) {
+            return defaultValue;
+        }
+        return collection;
+    }
+
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
@@ -5,6 +5,7 @@ import org.opentestsystem.rdw.reporting.exam.GroupExam;
 import org.opentestsystem.rdw.reporting.exam.repository.GroupIcaSummativeExamRepository;
 import org.opentestsystem.rdw.reporting.exam.repository.GroupIabExamRepository;
 import org.opentestsystem.rdw.reporting.security.GroupGrant;
+import org.opentestsystem.rdw.reporting.security.GroupSecurityContext;
 import org.opentestsystem.rdw.reporting.security.PermissionScope;
 import org.opentestsystem.rdw.reporting.security.User;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -36,19 +37,11 @@ public class DefaultGroupAssessmentService implements GroupAssessmentService {
 
     public Optional<Assessment> getLatestAssessment(@NotNull final User user, final long groupId, final long schoolYear) {
 
-        // assured not null by spring security annotation
-        final GroupGrant groupGrant = user.getGroupsById().get(groupId);
-        final PermissionScope permissionScope = user.getPermissionsById().get("GROUP_PII_READ").getScope();
+        final GroupSecurityContext securityContext = new GroupSecurityContext(user, groupId, "GROUP_PII_READ");
 
         final List<GroupExam> exams = newArrayList();
-
-        groupIabExamRepository
-                .findLatest(groupGrant.getId(), groupGrant.getSubjectId(), schoolYear, permissionScope)
-                .ifPresent(exams::add);
-
-        groupIcaSummativeExamRepository
-                .findLatest(groupGrant.getId(), groupGrant.getSubjectId(), schoolYear, permissionScope)
-                .ifPresent(exams::add);
+        groupIabExamRepository.findLatest(securityContext, schoolYear).ifPresent(exams::add);
+        groupIcaSummativeExamRepository.findLatest(securityContext, schoolYear).ifPresent(exams::add);
 
         if (exams.size() == 0) {
             return Optional.empty();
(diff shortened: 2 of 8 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -70,26 +112,6 @@ sql:
...
-            where e.school_year=:school_year
-            and sgm.student_group_id=:group_id
-            and (0=:group_subject_id or a.subject_id=:group_subject_id)
-            and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
-            order by e.completed_at desc
-            limit 1
-
   district:
     findAllIds: >-
       select natural_id, id
```
esotrope: little hard to see what i did in this file but basically I added the "icasummative" prefix to house queries targeted for these resources (like how there is already the "iab" qualifier)
Second, I added in the improved get all assessment query Alla provided:


 +           select a.id,
 +             a.name,
 +             a.grade_id,
 +             a.type_id,
 +             a.subject_id,
 +             a.school_year,
 +             aa.min_score,
 +             aa.max_score,
 +             aa.cut_point_1,
 +             aa.cut_point_2,
 +             aa.cut_point_3
 +           from asmt a
 +             left join asmt_score aa on a.id=aa.asmt_id
 +           where exists (
 +               select e.asmt_id
 +               from iab_exam e
 +                 join student_group_membership sgm on e.student_id=sgm.student_id
 +                 join school s on e.school_id=s.id
 +               where
 +                 e.school_year=:school_year
 +                 and sgm.student_group_id=:group_id
 +                 and (0=:group_subject_id or a.subject_id=:group_subject_id)
 +                 and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
 +                 and a.id = e.asmt_id
 +           )

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
```diff
@@ -36,19 +37,11 @@ public DefaultGroupAssessmentService(
...
-                .findLatest(groupGrant.getId(), groupGrant.getSubjectId(), schoolYear, permissionScope)
-                .ifPresent(exams::add);
-
-        groupIcaSummativeExamRepository
-                .findLatest(groupGrant.getId(), groupGrant.getSubjectId(), schoolYear, permissionScope)
-                .ifPresent(exams::add);
+        groupIabExamRepository.findLatest(securityContext, schoolYear).ifPresent(exams::add);
+        groupIcaSummativeExamRepository.findLatest(securityContext, schoolYear).ifPresent(exams::add);
 
         if (exams.size() == 0) {
```
malaffoon: `exams.isEmpty()` > `exams.size() == 0`

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
```diff
@@ -61,19 +54,10 @@ public DefaultGroupAssessmentService(
 
     @Override
     public List<Assessment> getAllAssessments(@NotNull final User user, final long groupId, final long schoolYear) {
-
-        // assured not null by spring security annotation
-        final GroupGrant groupGrant = user.getGroupsById().get(groupId);
-        final PermissionScope permissionScope = user.getPermissionsById().get("GROUP_PII_READ").getScope();
-
+        final GroupSecurityContext securityContext = new GroupSecurityContext(user, groupId, "GROUP_PII_READ");
         final ImmutableList.Builder<Assessment> assessments = ImmutableList.builder();
```
malaffoon: We should probably be consistent with the techniques used to build a list of things from multiple sources. I have a slight preference for not returning an Immutable collection, especially if the method is building it explicitly for returning. That is, i prefer lines 42-44 over 58-60.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Queries.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.reporting.common.jdbc;
+
+import java.util.Collection;
+
+public final class Queries {
+
+    private Queries() {
+    }
+
+    public static Object nullOrEmptyToDefault(final Collection collection, final Object defaultValue) {
```
bgsavage: I think it's odd this method can return an Object or a Collection.  Wouldn't a caller of this method always have to check the result and then do some casting to be able to use it?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/GroupExamRepository.java
```diff
@@ -2,6 +2,7 @@
 
 import org.opentestsystem.rdw.reporting.assessment.Assessment;
 import org.opentestsystem.rdw.reporting.exam.GroupExam;
+import org.opentestsystem.rdw.reporting.security.GroupSecurityContext;
 import org.opentestsystem.rdw.reporting.security.PermissionScope;
```
bgsavage: nit:  looks like PermissionScope is unused now, yeah?

webapp/src/main/resources/application.sql.yml
```diff
@@ -15,14 +15,72 @@ sql:
           where user_login=:user_login
 
     assessment:
-      findAll: >-
-        select
-            distinct a.id,
+      icasummative:
```
bgsavage: nit: should be icaSummative ?

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Queries.java
esotrope: Good point.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
esotrope: Ah I think we talked about this before - just to return a mutable list? is that what you mean?

I would normally default to returning immutables.
The reason the collection starts mutable on 42,44 is because it needs to be sorted.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
esotrope: sorry can't use number operators on boolean expressions

malaffoon: java 10 is adding qualitative overloads for all operators  ;)

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
malaffoon: Ok. Out of context in the review tool the inconsistency stands out, but if there's a reason for it ... np.

webapp/src/main/resources/application.sql.yml
```diff
@@ -15,14 +15,72 @@ sql:
           where user_login=:user_login
 
     assessment:
-      findAll: >-
-        select
-            distinct a.id,
+      icaSummative:
+        findAll: >-
+          select a.id,
             a.name,
```
agorina: I still think this should be a label, not a name.

esotrope: meant to change that - thanks. I'll add that in the next PR

webapp/src/main/resources/application.sql.yml
agorina: "Improvement" is an overstatement until we test it :)  but thank you for changing it.

PR #99, opened 2017-05-22

bgsavage:
Updated API calls from UI with the what is in develop.  Then I broke that by what @esotrope plans to return for the the exams collection (flattened collection now instead of grouped by sessions).  I ended up creating UI models and mappers to guard against more changes coming from the API soon.  I also refactored out a couple of modules to make the app module smaller.

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/results/assessment/assessment-exam.mapper.ts
+++ b/webapp/src/main/webapp/src/app/groups/results/assessment/assessment-exam.mapper.ts
@@ -1,0 +1,43 @@
+import { Injectable } from "@angular/core";
+import { AssessmentExam } from "./model/assessment-exam.model";
+import { Assessment } from "./model/assessment.model";
+import { Exam } from "./model/exam.model";
+
+@Injectable()
+export class AssessmentExamMapper {
+  mapFromApi(apiModel): AssessmentExam {
+    let uiModel = new AssessmentExam();
+
+    uiModel.assessment = this.mapAssessmentFromApi(apiModel);
+    uiModel.exams = [];
+
+    apiModel.exams.forEach(x =>
+      uiModel.exams.push(this.mapExamFromApi(x))
+    );
+
+    return uiModel;
+  }
+
+  private mapAssessmentFromApi(apiModel): Assessment {
+    let uiModel = new Assessment();
+
+    uiModel.id = apiModel.id;
+    uiModel.name = apiModel.name;
+    uiModel.grade = apiModel.gradeId;
+
+    return uiModel;
+  }
+
+  private mapExamFromApi(apiModel): Exam {
+    let uiModel = new Exam();
+
+    uiModel.date = apiModel.dateTime;
+    uiModel.session = apiModel.sessionId;
+    uiModel.enrolledGrade = apiModel.grade;
+    uiModel.studentName = `${apiModel.studentName.last}, ${apiModel.studentName.first}`;
+    uiModel.score = apiModel.score;
+
+    return uiModel;
+  }
+}
+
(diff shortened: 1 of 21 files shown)


webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -7,8 +7,8 @@
       </button>
     </div>
     <h4 class="spacer-v-md">
-      <span class="label bg-color-grade-{{assessment.gradeId}} pad-v-lg">G{{assessment.gradeId}}</span>
-      <span class="color-grade-{{assessment.gradeId}} spacer-h-md"> {{assessment.name}}</span>
+      <span class="label bg-color-grade-{{assessmentExam.assessment.grade}} pad-v-lg">G{{assessmentExam.assessment.grade}}</span>
```
esotrope: Don't forget to make G here come from the translations

bgsavage: Good call, I updated this.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-exam.mapper.ts
```diff
@@ -0,0 +1,43 @@
...
+    return uiModel;
+  }
+
+  private mapExamFromApi(apiModel): Exam {
+    let uiModel = new Exam();
+
+    uiModel.date = apiModel.dateTime;
+    uiModel.session = apiModel.sessionId;
+    uiModel.enrolledGrade = apiModel.grade;
+    uiModel.studentName = `${apiModel.studentName.last}, ${apiModel.studentName.first}`;
```
esotrope: I see the want to reduce redundancy in templates by doing this but this formatting is specific to a individual views so I think it's best to keep this out of the model.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -18,10 +20,10 @@ <h4 class="spacer-v-md">
       <div class="spacer-v-lg">
         <strong>{{'labels.assessment.sessions-title' | translate}}</strong>
         <div class="spacer-v-sm">{{'labels.assessment.sessions-instruct' | translate}}</div>
-        <button *ngFor="let session of assessment.sessions" class="btn btn-gray spacer-h-sm" (click)="toggleSession(session)" [ngClass]="{'active':session.filter}">
-          <strong class="spacer-h-sm">{{session.dateTime | date }}</strong>
+        <button *ngFor="let session of sessions" class="btn btn-gray spacer-h-sm" (click)="toggleSession(session)" [ngClass]="{'active':session.filter}">
+          <strong class="spacer-h-sm">{{session.date | date }}</strong>
           <i class="fa fa-minus-square" [ngClass]="{'fa-minus-square':session.filter, 'fa-plus-square':!session.filter }"></i>
-          <div>{{session.id}}</div>
+          <div> {{session.id || '[Unknown]'}}</div>
```
esotrope: we'll need to run "[Unknown]" through the translate pipe

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-exam.mapper.ts
bgsavage: I'm not sure yet how this will evolve.  I may create a name model and use a pipe to for different formats for the name.  Right now this is the only format being shown so just looking to keep it simple for now.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
bgsavage: Good point.  I'll update on next check-in

jtreuting: One thing to consider with this, is that we might end up naming historical sessions somehow instead of having them be blank when coming back to the UI.  That is still unknown (see what I did there, very punny).

PR #100, opened 2017-05-23
Feature/group exams endpoint

esotrope:
TODOs
1. add exams to /latestassessment payload
2. add tests


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Queries.java
```diff
@@ -1,17 +0,0 @@
-package org.opentestsystem.rdw.reporting.common.jdbc;
-
-import java.util.Collection;
-
-public final class Queries {
-
-    private Queries() {
-    }
-
-    public static <T> Collection<T> nullOrEmptyToDefault(final Collection<T> collection, final Collection<T> defaultValue) {
```
esotrope: moved this method to the webapp module since the pdf processor probably wont use it

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
```diff
@@ -13,6 +16,7 @@
     private long typeId;
     private long subjectId;
     private long schoolYear;
+    private List<Long> cutPoints; // TODO field not in IAB assessments - decide if this warrants a separate model
```
esotrope: I think I saw the ingest project break exams and assessments out into different models. I'm wondering if this is warranted though. The field difference for both so far is one field.

bgsavage: Yeah I saw tons of similar code in Ingest that creating different models caused.  I would prefer one model as well.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StudentContext.java
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * @return <code>true</code> if the the student had an Individual Education Plan according to the IDEA act at the time of the exam
+     */
+    public boolean isIdeaIndicator() {
+        return ideaIndicator;
+    }
+
+    /**
+     * @return TODO - add description - http://www.smarterapp.org/documents/TestResults-DataModel.pdf says IEP is the AIR alias for idea indicator - are these fields redundant?
```
esotrope: @agorina , @jtreuting  - I'm not sure what this field is. Is this the same as the idea indicator?

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
```diff
@@ -57,6 +61,17 @@ public long getSchoolYear() {
...
 
     /**
+     * Cut points are the fixed point values which separate the achievement levels for ICA and Summative assessments.
+     * These points are sorted in ascending order beginning with the minimum score and ending with the maximum.
+     * the maximum possible assessment score.
+     *
+     * @return the assessment cut points if the assessment is of type ICA or Summative. Empty if IAB
+     */
+    public List<Long> getCutPoints() {
+        return cutPoints;
```
bgsavage: Just to clarify, the largest value in this collection will be the maximum value?  Curious if the server or the client should also make the percents for this as well?  I'm leaning toward the client since it will need it to make the scale visualization, but I'm not sure if the server would need it for doing some kind of export late.r

esotrope: yeah eventually the logic will end up in both places. we could do the pre computation on the backend. but the front end will still have to do the same or similar computations for aggregating the exam scale scores together

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
```diff
@@ -59,4 +68,37 @@ public DefaultGroupAssessmentService(
         return assessments.build();
     }
 
+    public List<Exam> getIabs(@NotNull final User user, final long groupId, final long assessmentId, final long schoolYear) {
```
bgsavage: This is kind of long parameter list, and should it change then it would cause a breaking change to anyone calling this method.  Would it be better practice to compose this data into an ExamRequest object (or something similar)?

bgsavage: Also, `User` is a common parameter that we will likely need for most service calls in order to authorize.  Could we come up with a pattern where we have some UserContext that gets injected through the constructor and we get the current user through that?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private Exam copy() {
+        final Exam copy = new Exam();
+        copy.id = id;
+        copy.sessionId = sessionId;
+        copy.dateTime = dateTime;
+        copy.opportunity = opportunity;
+        copy.schoolYear = schoolYear;
+        copy.student = student;
```
bgsavage: Does this copy the value of student or just the reference?

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
malaffoon: Just a reminder that the default pattern in Spring is that Services are stateless singletons.

bgsavage: On yeah that's right, not per request.  Sorry keep forgetting that one :)

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
malaffoon: If the student is immutable, does it matter?

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/DefaultGroupAssessmentService.java
```diff
@@ -59,4 +68,37 @@ public DefaultGroupAssessmentService(
...
+        return getExams(user, groupId, assessmentId, schoolYear, groupIcaSummativeExamRepository);
+    }
+
+    private List<Exam> getExams(final @NotNull User user, final long groupId, final long assessmentId, final long schoolYear, final GroupExamRepository repository) {
+
+        final GroupSecurityContext securityContext = new GroupSecurityContext(user, groupId, "GROUP_PII_READ");
+
+        final List<Exam> exams = repository.findAllForAssessment(securityContext, assessmentId, schoolYear);
+
+        final List<Long> studentIds = exams.stream().map(exam -> exam.getStudent().getId()).collect(toImmutableList());
```
malaffoon: I haven't gotten far in this review, but it seems odd that the service is doing all the ethnicity work. Why didn't the repository populate the student ethnicity?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StudentContext.java
chdietz: The TRDD Page six has:
For Field : IDEAIndicator

The Cross Reference Column has:
TRT:ExamineeAttribute - IDEAIndicator
DWSA:IDEAIndicator
AIR:IEP

So they may be the same, where did you pick up IDEAIndicator?

malaffoon: The schema has only `iep` in it, where is `idea` from?

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
malaffoon: Can you clarify? I don't think you mean "exams" and "assessments" should be combined. And, for the most part, i thought exams of various flavors did share a common base class. As do assessments of various flavors. I'm a bit tired though, so i could be wrong.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
```diff
@@ -129,6 +150,7 @@ public Assessment build() {
             assessment.typeId = typeId;
             assessment.subjectId = subjectId;
             assessment.schoolYear = schoolYear;
+            assessment.cutPoints = cutPoints != null ? ImmutableList.copyOf(cutPoints) : ImmutableList.of();
```
malaffoon: I think this is problematic. For IABs, `assessment.cutPoints` should be `null` not an empty list.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcEthnicityRepository.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Autowired
+    public JdbcEthnicityRepository(
+            @NotNull final NamedParameterJdbcTemplate template,
+            @Value("${sql.ethnicity.findAll}") final String findAllQuery,
+            @Value("${sql.ethnicity.findEthnicityIdsForStudentIds}") final String findEthnicityIdsForStudentIdsQuery) {
+
+        // lookup and cache ethnicity ID to code mappings at startup
+        final ImmutableMap.Builder<Long, String> builder = ImmutableMap.builder();
+        template.query(findAllQuery, row -> {
```
malaffoon: I don't suppose it matters a lot but you can use a RowCallbackHandler instead of a ResultSetExtractor so we don't have to have the loop in our code.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcEthnicityRepository.java
```diff
@@ -0,0 +1,68 @@
...
+                        final long studentId = row.getLong("student_id");
+                        final String ethnicityCode = ethnicityCodesById.get(row.getLong("ethnicity_id"));
+
+                        final List<String> studentEthnicityCodes = ethnicityCodesByStudentId.get(studentId);
+                        if (studentEthnicityCodes == null) {
+                            ethnicityCodesByStudentId.put(studentId, newArrayList(ethnicityCode));
+                        } else {
+                            studentEthnicityCodes.add(ethnicityCode);
+                        }
+                    } while (row.next());
```
malaffoon: ditto

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupAssessmentController.java
```diff
@@ -43,4 +44,24 @@ public Assessment getLatestGroupAssessment(
         return service.getAllAssessments(user, groupId, schoolYear);
     }
 
+    @GetMapping("/api/groups/{groupId}/assessments/{assessmentId}/iabs")
+    public List<Exam> getGroupIabs(
```
malaffoon: Are we ever going to want to get both IABs and ICA/Summatives in a single call? If so, i wonder if we should change this to a single end-point with a filter, e.g. `/api/groups/{groupId}/assessments/{assessmentId}/exams?type=IAB`. That would apply down the stack, i.e. Controller, Service, Repository.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
agorina: I am confused to. In the Ingest, there is one Assessment model, and two Exams, but they both extend AnyExam that has all the shared attributes. Not sure what you mean here.

It looks like this class represents an Assessment. Just to note: ICA has cut points and  Scored Claims, while IAB does not have them.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ScaleScore.java
```diff
@@ -0,0 +1,71 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+/**
+ * Represents a scale score of an assessment
+ */
+public class ScaleScore {
+
+    private long level;
+    private long value;
```
agorina: We are accepting unscored ICA. This means that these values could be NULL.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StudentContext.java
```diff
@@ -0,0 +1,206 @@
...
+public class StudentContext {
+
+    private long schoolId;
+    private long gradeId;
+    private boolean ideaIndicator;
+    private boolean iep;
+    private boolean lep;
+    private boolean section504;
+    private boolean economicDisadvantage;
+    private boolean migrantStatus;
```
agorina: Some of these attributes are optional and can be NULL. Please check the schema for  the ones that allow NULLs.

webapp/src/main/resources/application.sql.yml
```diff
@@ -18,7 +18,7 @@ sql:
       icaSummative:
         findAll: >-
           select a.id,
-            a.name,
+            a.label,
```
agorina: Thank you! :)

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ScaleScore.java
```diff
@@ -0,0 +1,71 @@
...
+            return this;
+        }
+
+        public Builder value(final long value) {
+            this.value = value;
+            return this;
+        }
+
+        public Builder standardError(final long standardError) {
+            this.standardError = standardError;
```
agorina: I believe we wanted to keep it as Double. When we calculate an error band, from the mathematical point of view it is more accurate to not cast to it Integer here.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupIcaSummativeExamRepository.java
```diff
@@ -1,19 +1,66 @@
...
+    }
+
+    @Override
+    protected Exam.Builder build(final Exam.Builder builder, final ResultSet row) throws SQLException {
+
+        final ImmutableList.Builder<ScaleScore> claimScaleScores = ImmutableList.builder();
+        for (int i = CLAIM_SCALE_SCORE_START_INDEX; i <= TOTAL_CLAIM_SCALE_SCORES; i++) {
+            claimScaleScores.add(ScaleScore.builder()
+                    .level(row.getLong(MessageFormat.format("claim{0}_category", i)))
+                    .value(row.getLong(MessageFormat.format("claim{0}_scale_score", i)))
```
agorina: Just to make sure you know: Math has 3 claim scores, while ELA has 4.

webapp/src/main/resources/application.sql.yml
```diff
@@ -74,33 +68,56 @@ sql:
...
+            left join asmt_score aa on e.asmt_id=aa.asmt_id
           where e.school_year=:school_year
           and sgm.student_group_id=:group_id
           and (0=:group_subject_id or a.subject_id=:group_subject_id)
           and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
           order by e.completed_at desc
           limit 1
 
+        findAllForAssessment: >-
+          select e.*,
```
agorina: Please use explicit columns here.

webapp/src/main/resources/application.sql.yml
```diff
@@ -112,6 +129,23 @@ sql:
...
```
agorina: Ditto.

webapp/src/main/resources/application.sql.yml
```diff
@@ -123,3 +157,12 @@ sql:
       select natural_id, id
         from school
         where natural_id in (:naturalIds)
+
+  ethnicity:
+    findAll: >-
+      select * from ethnicity
```
agorina: And here, could you please use explicit column names.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupAssessmentController.java
jtreuting: @malaffoon no they can't.  If you know an assessment ID you will be able to figure out if it's IAB, ICA or Summative that is part of the assessment metadata really.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
jtreuting: This one feels like wait a see a little bit.  Smarter is putting together their thoughts on the score bar for IABs and wil give us more info tomorrow.  Let's assess after that as we will know a little bit more.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupAssessmentController.java
bgsavage: > It doesn't make any sense to query on both assessment id and type.

Since we have two different tables in the db, the type needs to be known in order to know which table to query.  At this time, we don't have a parent table which could provide unique assessment ids across both exam / iab_exam.

malaffoon: Thanks Jeff. I think that means we should change this end-point, and possibly the data-flow.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
esotrope: Sorry - question was "should I make this an abstract class and add an IcaSummativeAssessment and IabAssessment class which extends it?"

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StudentContext.java
esotrope: This one is my mistake - iep/idea indicator are the same thing and I accidently duplicated the field with different names - will patch

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
malaffoon: Personally i wouldn't ... yet. There may be other reasons to in the future but this model if for interacting with the UI. If the UI is ok branching on cut-points being null then you're golden.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/Assessment.java
esotrope: I will leave as is for now and we can wait and see.
Also all forking of logic for iab vs ica/summative will be done on the ui using the assessment type ID value

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcEthnicityRepository.java
esotrope: Do you have an example mark - all the examples I see require looping yourself.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ScaleScore.java
esotrope: Ok. Good to know - thanks for the correction.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StudentContext.java
esotrope: I will check against the doc you sent me and update - thanks

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupIcaSummativeExamRepository.java
esotrope: Ok - I think what I'll do is add logic to ignore the columns if they come back null

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcEthnicityRepository.java
malaffoon: You can look at JdbcReportRepository in the task-service module. But something like ...
```
    template.query(sql, params, row -> { 
        builder.put(row.getLong("id"), row.getString("name"));
    });
```

esotrope: Great thanks - I'll use this. I think my issue was when you remove the {} java assumes I want to return the result of the inner block and therefore assumes I wanted a ResultSetExtractor

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
```diff
@@ -54,23 +76,112 @@
...
+                    .lastName(row.getString("student_last_name"))
+                    .genderId(row.getLong("student_gender_id"))
+                    .ethnicityCodes(ethnicityCodes)
+                    .build()
+            );
+            exams.add(build(examBuilder, row).build());
+        }
+
+        return exams;
+    }
```
esotrope: @jtreuting, @malaffoon , @agorina - hey guys looking for a re-review here now that I have moved the student ethnicity code logic into the repo - thanks

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupIcaSummativeExamRepository.java
```diff
@@ -1,19 +1,69 @@
...
+        super(ethnicityRepository, template, findLatestQuery, findAllQuery, findAllForAssessmentQuery);
+    }
+
+    @Override
+    protected Exam.Builder build(final Exam.Builder builder, final ResultSet row) throws SQLException {
+
+        final ImmutableList.Builder<ScaleScore> claimScaleScores = ImmutableList.builder();
+
+        // ELA exams only have 3 claims, MATH have 4
+        final int endIndex = row.getObject("claim4_category", Integer.class) != null ? TOTAL_MATH_CLAIMS : TOTAL_ELA_CLAIMS;
```
esotrope: @agorina - does this logic look correct? I noticed i wouldn't be able to look at the asmt type without a join in the query.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupIcaSummativeExamRepository.java
```diff
@@ -1,19 +1,69 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.text.MessageFormat;
 
 @Repository
-public class JdbcGroupIcaSummativeExamRepository extends AbstractJdbcGroupExamRepository implements GroupIcaSummativeExamRepository {
+public class JdbcGroupIcaSummativeExamRepository extends AbstractJdbcGroupExamRepository<Assessment.Builder, Exam.Builder> implements GroupIcaSummativeExamRepository {
+
+    private static final int TOTAL_MATH_CLAIMS = 4;
+    private static final int TOTAL_ELA_CLAIMS = 3;
```
agorina: It is the other way around: Math has 3, ELA has 4. If I am the one who told you this, then sorry :)

I am not sure if this is important here though. I think it may be ok to get them all, for the Math the last one will be Null.   

I would like to understand how it will be used. Maybe we can chat/whiteboard tomorrow? I think the UI may need to know more than Claim1-3/4. It may need to know the code for each claim to know how what label to show for it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupIcaSummativeExamRepository.java
agorina: See my comment above.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/GroupAssessmentService.java
```diff
@@ -16,24 +18,48 @@
...
+     * Gets all IAB exams for the given group, assessment and school year.
+     *
+     * @param user         the user accessing the group information
+     * @param groupId      the group for which the assessments will be searched
+     * @param assessmentId the assessment the group took
+     * @param schoolYear   the academic year for which the assessments will be searched
+     * @return all IAB exams for the given group, assessment and school year.
+     */
+    @PreAuthorize("hasAuthority('PERM_GROUP_PII_READ') and #user.getGroupsById().containsKey(#groupId)")
+    List<Exam> getIabs(@NotNull User user, long groupId, long assessmentId, long schoolYear);
```
malaffoon: I think that these two methods should be combined into getExams(). The logic of which type of exam should be handled inside the (service) method.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
```diff
@@ -0,0 +1,196 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+import java.time.Instant;
+import java.util.List;
+
+/**
+ * Represents a single assessment examination event
+ */
+public class Exam {
```
malaffoon: @esotrope i thought about yesterday's discussion ... given that this data model is to support our home-built ORM, i think we should use mutable objects. Believe me, this is not something i say lightly, i really really like immutable data objects. But the complexity of the query methods is not (IMO) worth it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/GroupExam.java
```diff
@@ -12,10 +14,12 @@
 
     private Instant completionDateTime;
     private Assessment assessment;
+    private List<Exam> exams;
 
-    public GroupExam(final Instant completionDateTime, final Assessment assessment) {
+    public GroupExam(final Instant completionDateTime, final Assessment assessment, final List<? extends Exam> exams) {
```
malaffoon: The `exams` param should probably be an Iterable. Why? General rule: make param types as flexible as necessary. Of course, if the order is critical the `List` type might act as a hint of that.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
malaffoon: More on this ...

I did come up with an approach of nested builders that might work as well. The Exam.Builder contains builders for Student and StudentContext. This allows the query method to use Exam.Builder throughout while producing an immutable result. But again, if the purpose of this data model is to simply move data up the stack, from the database to the UI, i'd lean towards mutable objects.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
```diff
@@ -38,13 +57,16 @@
                         .build(),
                 (row, index) -> new GroupExam(
                         row.getTimestamp("completed_at").toInstant(),
```
malaffoon: I still feel a separate Mapper would make all these more readable. I also realize this is just an opinion if there is no reuse of the Mapper.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
```diff
@@ -54,23 +76,112 @@
...
+                        .put("assessment_id", assessmentId)
+                        .put("school_year", schoolYear)
+                        .build(),
+                this::buildExams
+        );
+    }
+
+    private List<Exam> buildExams(final ResultSet row) throws SQLException {
+
+        final List<Exam> exams = newArrayList();
```
malaffoon: This is much cleaner, thanks.

I still have the willies about reseting the result set.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
```diff
@@ -54,23 +76,112 @@
...
+        }
+
+        return exams;
+    }
+
+    protected A createAssessmentBuilder() {
+        return (A) Assessment.builder();
+    }
+
+    protected A build(final A builder, final ResultSet row) throws SQLException {
```
malaffoon: Should add javadocs explaining that concrete classes should extract type-specific data from the result set and apply to builder.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupIabExamRepository.java
```diff
@@ -1,19 +1,36 @@
...
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Repository;
 
 import javax.validation.constraints.NotNull;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 
 @Repository
-public class JdbcGroupIabExamRepository extends AbstractJdbcGroupExamRepository implements GroupIabExamRepository {
+public class JdbcGroupIabExamRepository extends AbstractJdbcGroupExamRepository<Assessment.Builder, Exam.Builder> implements GroupIabExamRepository {
```
malaffoon: ~~public~~

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcEthnicityRepository.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.validation.constraints.NotNull;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Maps.newHashMap;
+
+@Repository
+public class JdbcEthnicityRepository implements EthnicityRepository {
```
malaffoon: ~~public~~

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
malaffoon: `build` is a bit of a misnomer since it doesn't actual build anything. Perhaps `processRow` or some such?

webapp/src/main/java/org/opentestsystem/rdw/reporting/jdbc/Parameters.java
```diff
@@ -5,17 +5,41 @@
...
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
-import static org.opentestsystem.rdw.reporting.common.jdbc.Queries.nullOrEmptyToDefault;
-
+/**
+ * Holds common methods for preparing parameters for SQL queries
+ */
 public class Parameters {
```
malaffoon: This is just me pontificating ...

The pattern that uses the class name pluralized for the utility class, e.g. Collections, Streams, is nice enough. But that doesn't really apply here since `Parameter` isn't a thing. Perhaps the older pattern of XxxUtils, e.g. IOUtils, makes more sense, i.e. `ParameterUtils`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupAssessmentController.java
malaffoon: Based on discussions it seems like these end-points should be combined, with the type removed from the path, either `api/groups/{groupId}/assessments/{assessmentId}/exams` or `api/groups/{groupId}/exams?assessment={assessmentId}`.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/GroupAssessmentService.java
```diff
@@ -16,24 +18,48 @@
...
+    /**
+     * Gets all IAB exams for the given group, assessment and school year.
+     *
+     * @param user         the user accessing the group information
+     * @param groupId      the group for which the assessments will be searched
+     * @param assessmentId the assessment the group took
+     * @param schoolYear   the academic year for which the assessments will be searched
+     * @return all IAB exams for the given group, assessment and school year.
+     */
+    @PreAuthorize("hasAuthority('PERM_GROUP_PII_READ') and #user.getGroupsById().containsKey(#groupId)")
```
bgsavage: What is the PERM prefix on the permission?  I didn't see that when I hit the api/users route.  

Also, what does @PreAuthorize return if this check fails?

esotrope: the PERM_ is for spring only - it is because it holds permissions and roles in the same "authorities" collection.

if pre authorize fails the UI will see a 403 forbidden

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
esotrope: that works

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/AbstractJdbcGroupExamRepository.java
esotrope: yeah i can split that out

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
esotrope: I'm a little lost here as to the focus on object mutability vs immutibility. I see the main driver of the repo method complexity being the need to make a second loop over the rows after ethnicity codes have been resolved.

webapp/src/main/java/org/opentestsystem/rdw/reporting/assessment/GroupAssessmentService.java
esotrope: Next PR I will have add this routing logic

PR #101, opened 2017-05-24

bgsavage:
Added calculations for aggregates of average, below standard, near standard, and above standard. Added getters/ setters for group-results component. Added functionality for showing percents and number radio button.


webapp/src/main/webapp/src/app/groups/results/assessment/assessment-exam.mapper.ts
```diff
@@ -36,6 +36,7 @@ export class AssessmentExamMapper {
     uiModel.enrolledGrade = apiModel.grade;
     uiModel.studentName = `${apiModel.studentName.last}, ${apiModel.studentName.first}`;
     uiModel.score = apiModel.score;
+    uiModel.level = apiModel.performance;
```
bgsavage: I realize these models don't reflect what the api is returning, these will be updated in a future iteration.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -13,17 +13,20 @@ <h4 class="spacer-v-md">
       <span class="color-grade-{{assessmentExam.assessment.grade}} spacer-h-md"> {{assessmentExam.assessment.name}}</span>
     </h4>
 
-      <div class="pull-right">
+      <div *ngIf="exams.length > 0" class="pull-right" style="width:250px;">
```
bgsavage: This style is just temp since I'm only displaying data points right now.  This won't be needed later.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -2,6 +2,8 @@ import { Component, OnInit, Input } from "@angular/core";
 import { trigger, transition, style, animate } from "@angular/animations";
 import { AssessmentExam } from "./model/assessment-exam.model";
 import { Exam } from "./model/exam.model";
+import { Utils } from "../../../shared/Utils";
```
bgsavage: oops this isn't needed anymore.

bgsavage: fixed

webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -8,44 +8,83 @@ import { AssessmentExam } from "./assessment/model/assessment-exam.model";
...
+
+  set filterBy(value: { schoolYear: number }) {
+    this._filterBy = value;
+  }
+
+  get selectedAssessments(): AssessmentExam[] {
+    return this._selectedAssessments;
+  }
+
+  private _selectedAssessments : AssessmentExam[] = [];
```
bgsavage: unrelated, creating getters / setters

webapp/src/main/webapp/src/app/shared/remove-comma.pipe.ts
```diff
@@ -0,0 +1,12 @@
+import {Pipe, PipeTransform} from "@angular/core";
+
+@Pipe({name: 'removeComma'})
+export class RemoveCommaPipe implements PipeTransform {
+
+  transform(value: string): string {
+    if(!value)
+      return value;
+
+    return value.replace(",", "");
```
bgsavage: The angular number pipe is great that it does rounding for us, but also can add commas depending on your browser locale.  For scores, we don't want commas so this filter is chained to strip commas.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -13,17 +13,20 @@ <h4 class="spacer-v-md">
...
+        <div title="{{statistics.average}}">Average: {{ statistics.average | number:'1.0-0' | removeComma }}</div>
```
bgsavage: removeComma usage

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -76,5 +95,24 @@ export class AssessmentResultsComponent implements OnInit {
...
+  }
+
+  private calculateStats() {
+    let stats : any = {
+      total: this._exams.length,
+      average: this._exams.reduce((x, y) => x + y.score, 0) / this._exams.length,
+      belowStandard: this._exams.filter(x => x.level == ExamResultLevel.BelowStandard).length,
+      nearStandard: this._exams.filter(x => x.level == ExamResultLevel.NearStandard).length,
+      aboveStandard: this._exams.filter(x => x.level == ExamResultLevel.AboveStandard).length,
+    };
```
agorina: Please be aware that for ICA/Summative the scores could be NULL. Not sure how we want to handle it. @jtreuting, could you please help here?

webapp/src/main/webapp/src/app/shared/exam-result-level.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum ExamResultLevel {
+  BelowStandard = 0,
+  NearStandard = 1,
+  AboveStandard = 2
```
agorina: This is for IAB only, right? ICA/Summative has 4 levels.  I assume you are aware, but just in case.

jtreuting: And to be clear the 4 levels are Level 1, Level 2, Level 3 and Level 4.  For ICA/Summative they are called Achievement Levels and for IAB the Below, Near and Above are called Reporting Category (in the Smarter terminology)

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
jtreuting: @agorina, do you mean how would a NULL score be handled when it comes to the aggregate?  I would assume it would be left out of the calculations, meaning that if there are 10 students in the group and 2 of them have NULL scores, the calculation is really summing the scores that are NOT NULL and dividing by 8 instead of 10.

jtreuting: Right now the statistics are specific to IAB (below, near, above).  At some point this should become more generic I think and called something like levels that is an array so it can handle either 3 or 4 levels depending on IAB or ICA/Summative.

agorina: Yes, this is what I meant. Thank you for clarifying it.

bgsavage: I wasn't sure how the other case was going to be calculated yet so I was going to let this calculation evolve once that part is known.  It may end up just being an array like you mentioned with the index being inferred depending on the assessment type.

webapp/src/main/webapp/src/app/shared/exam-result-level.enum.ts
bgsavage: I'm aware there's a difference, but not clear yet how ICA/Summative will look :)

esotrope: Not sure how it is useful to model this. what is the application?

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
esotrope: We could simplify this by just modeling the level as 0-3 or 4 and averaging all the levels i think

PR #102, opened 2017-05-24
Feature/dwr 346 ui authorize

bgsavage:
This PR creates the *sbAuthorize structural directive and auth guard.  This provides us with two generic ways in which we can enforce user permissions in the reporting web app.  

1.  By route
   a. add the canActivate: [ AuthorizeCanActivate ] on the route.
   b. add the permissions array on that route's data params.

2.  By element
    a. add the *sbAuthorize attribute on the attribute you want to enforce permissions in and pass in the permissions a user should have to see the element (and contents within).

Unit tests still to come.


webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -4,11 +4,11 @@
       <p>{{ 'labels.home.welcome' | translate }} <a href="">{{ 'labels.home.tutorial-link-text' | translate }}</a></p>
     </div>
     <div class="row row-eq-height">
-      <div class="col-md-8 card overflow-hidden margin-right-20">
+      <div *sbAuthorize="['GROUP_PII_READ','INDIVIDUAL_PII_READ']" class="col-md-8 card overflow-hidden margin-right-20">
```
bgsavage: sbAuthorize usage.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -25,8 +26,9 @@ export const routes: Routes = [
         path: 'groups/:groupId',
         pathMatch: 'full',
         resolve: { assessment: AssessmentsResolve },
-        data: { breadcrumb: { translate: 'labels.groups.name'} },
-        component: GroupResultsComponent
+        data: { breadcrumb: { translate: 'labels.groups.name'}, permissions: ['GROUP_PII_READ','INDIVIDUAL_PII_READ'] },
+        component: GroupResultsComponent,
+        canActivate: [ AuthorizeCanActivate ]
```
bgsavage: canActivate usage.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -25,8 +26,9 @@ export const routes: Routes = [
         path: 'groups/:groupId',
         pathMatch: 'full',
         resolve: { assessment: AssessmentsResolve },
-        data: { breadcrumb: { translate: 'labels.groups.name'} },
-        component: GroupResultsComponent
+        data: { breadcrumb: { translate: 'labels.groups.name'}, permissions: ['GROUP_PII_READ','INDIVIDUAL_PII_READ'] },
```
bgsavage: permissions required to activate this route.  Only need to have one in the array (not all).

webapp/src/main/webapp/src/app/user/user.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import { BrowserModule } from "@angular/platform-browser";
+import { AuthorizeCanActivate } from "./authorize.can-activate";
+
+@NgModule({
+  declarations: [ AuthorizeDirective ],
+  imports: [ CommonModule, BrowserModule ],
+  exports: [ AuthorizeDirective ],
+  providers: [ UserMapper, UserService, AuthorizeCanActivate ]
+})
+export class UserModule {
```
bgsavage: I'm moving toward more modularized web app than what we had before.  I think this will help keep us organized.  My thoughts are that if it warrants it's own directory in the app root, then it also warrants its own module.

webapp/src/main/webapp/src/app/shared/cachingData.service.ts
```diff
@@ -12,16 +12,21 @@ export class CachingDataService {
...
 
     let observable = this.http
       .get(`/api${url}`, options)
+      .share()
       .map(response => response.json());
 
-    this.cache[url] = observable;
+    observable
+      .subscribe(x => {
+        this.cache[url] = x;
```
esotrope: why do we need this cache? i thought the routes did this for us no?

webapp/src/main/webapp/src/app/user/authorize.directive.ts
```diff
@@ -0,0 +1,37 @@
...
+    private _userService : UserService,
+    private _templateRef: TemplateRef<any>,
+    private _viewContainer: ViewContainerRef) { }
+
+  @Input()
+  set sbAuthorize(permissions: string[]) {
+    if(!permissions || !permissions.length || permissions.length == 0 )
+      throw new Error("Specify at least one permission to authorize against.")
+
+    this._userService.doesCurrentUserHaveAtLeastOnePermission(permissions).subscribe(hasPermission => {
```
esotrope: im wondering if there is a 3rd party lib we can use for permission expressions similar to spring security expressions. I don't think we'll need anything but "or" expressions but it might be good to line one up in case we do all of a sudden

webapp/src/main/webapp/src/app/shared/cachingData.service.ts
bgsavage: Routes only provide us with context based caching depending how we construct the routes and resolves.  I don't think directives should try to get it's data from the route because the directive should be able to be used on any route, not the just the first one we develop for it.  Let me know if that's not clear

webapp/src/main/webapp/src/app/user/authorize.directive.ts
bgsavage: I did a quick peruse before implementing, but this seemed simple enough to do for now as opposed to bringing something else in.

webapp/src/main/webapp/src/app/user/user.service.ts
```diff
@@ -0,0 +1,35 @@
...
+  getCurrentUser() {
+    return this._dataService
+      .get("/user")
+      .map(x => this._mapper.mapFromApi(x));
+  }
+
+  doesCurrentUserHaveAtLeastOnePermission(permissions: string[]): Observable<boolean> {
+    return new Observable(observer => {
+      this.getCurrentUser().subscribe(user => {
+        observer.next(this.doesAtLeastOneExist(permissions, user.permissions));
```
esotrope: isn't this the same?
```javascript
user.permissions.some(p -> permissions.some(p))
```

bgsavage: Yeah I thought about doing this, but I kept confusing myself :)

I can go to this approach if you think it's cleaner.  Something to keep in mind, I didn't want this to fail because of case sensitivity or trailing white spaces.

webapp/src/main/webapp/src/app/user/authorize.directive.ts
```diff
@@ -0,0 +1,37 @@
...
+  private _hasView;
+
+  constructor(
+    private _userService : UserService,
+    private _templateRef: TemplateRef<any>,
+    private _viewContainer: ViewContainerRef) { }
+
+  @Input()
+  set sbAuthorize(permissions: string[]) {
+    if(!permissions || !permissions.length || permissions.length == 0 )
```
esotrope: I think you can reduce to this:
```javascript
if (!permissions || permissions.length == 0)
```
if permissions is not an array permissions.length will eval to undefined and undefined == 0 will eval to false

webapp/src/main/webapp/src/app/user/user.service.ts
esotrope: user permissions you can assume will be trimmed.
maybe do
```javascript
permissions.some(p -> user.permissions.some(p.trim()))
```

bgsavage: Sure can do.

webapp/src/main/webapp/src/app/user/authorize.directive.ts
bgsavage: Yeah that's right, thanks. Will update.

PR #106, opened 2017-05-29
Feature/ui tweaks

bgsavage:
Updated based on feedback from team demo and also in preparation for tomorrow's demo.


webapp/src/main/webapp/src/app/groups/results/assessment/model/exam.model.js
```diff
@@ -1,87 +0,0 @@
-"use strict";
```
bgsavage: not sure how this file snuck into a previous check-in, but shouldn't have been.

webapp/src/main/webapp/src/app/standalone/data/data.ts
```diff
@@ -192,6 +192,7 @@ export const exams_of_student = [
   exam.id = randomId();
   exam.items = iab_items;
   exam.studentContext = { gradeId: exam.grade };
+  exam.scaleScore.standardError = intBetween(10,100);
```
malaffoon: slightly more realistic would be to have decimal values in the data from the server (that we may or may not round in the UI); no biggie for demo though

webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -21,14 +21,14 @@
...
+        <p-dataTable [value]="filteredGroups" emptyMessage="{{'labels.groups.empty-message' | translate}}">
+          <p-column field="name" header="{{'labels.groups.cols.group' | translate}}" [sortable]="true">
             <ng-template let-group="rowData" pTemplate="body">
               <a routerLink="/groups/{{group.id}}">{{group.name}}</a>
             </ng-template>
           </p-column>
-          <p-column field="schoolName" header="{{'labels.school.name' | translate}}" [sortable]="true"></p-column>
-          <p-column field="subjectId" header="{{'labels.groups.subject' | translate}}" [sortable]="true">
+          <p-column field="schoolName" header="{{'labels.groups.cols.school' | translate}}" [sortable]="true"></p-column>
+          <p-column field="subjectId" header="{{'labels.groups.cols.subject' | translate}}" [sortable]="true">
```
esotrope: nit: cols -> columns or column-headers

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -53,7 +54,16 @@ <h4 class="spacer-v-md">
...
         </p-column>
-        <p-column field="score" header="{{'labels.groups.results.exam-cols.score' | translate}}" [sortable]="true"></p-column>
+        <p-column field="level" header="{{ performanceLevelHeader | translate }}" [sortable]="true">
+          <ng-template let-exam="rowData" pTemplate="body">
+            {{ examLevelEnum + exam.level | translate }}
+          </ng-template>
+        </p-column>
+        <p-column field="score" header="{{'labels.groups.results.exam-cols.score' | translate}}" [sortable]="true">
+          <ng-template let-exam="rowData" pTemplate="body">
+            {{ exam.score }} ± {{ exam.standardError }}
```
agorina: I am assuming that this is just for the demo tomorrow. Based on our conversation I am assuming that later you will make sure that this does not go higher/lower than the assessment min/max values.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -88,6 +88,11 @@ export class AssessmentResultsComponent {
       : "enum.achievement-level.";
   }
 
+  get performanceLevelHeader() {
+    return "labels.groups.results.exam-cols." +
+      (this.isIab ? "iab" : "ica") + ".performance";
```
esotrope: not sure about the isIab thing -> does this mean we should also have isSummative isIca?
I would prefer to see this.typeId == AssessmentType.IAB or something.
Also I feel like this logic belongs in the view and not the component as it is just translation mapping logic and not business logic

webapp/src/main/webapp/src/app/standalone/data/data.ts
bgsavage: The server / generated data returns integers right now, not decimals.

esotrope: it's modeled to return decimals its just that the data in the db happens to all be positive integers I think

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
bgsavage: This was mainly for the demo.  When I bring in the aggregate visualization I will look and see if I need to do some Math.min(scorepoint-max, exam.score + exam.standardError).  From the designs I've seen, we never actually show the std error, just use it to get the width of that pointer.

agorina: Sounds good. Thank you for the explanation.

webapp/src/main/webapp/src/app/groups/groups.component.html
bgsavage: I know we typically don't like abbreviations, but I'm trying to keep the pathing of the translate shorter / concise.  I think `cols` is one of those abbreviations that is know business domain specific that I can get away with right?  It's like using `Utils` instead of `Utilities`.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
bgsavage: This is another I'm trying to keep the logic more concise.  If we are interested in seeing how I determine isIab, you can go to the getter for that and see I'm actually comparing type.  Since we mostly care if it's iab or not, I don't think isSummative and isIca is not needed.

Also I prefer to keep logic out of the view when I can because the view can't be unit tested as easily as a component.

PR #107, opened 2017-05-30
reset schema pointer'

agorina:


esotrope:
Will accommodation translations not live in a more general purpose translations table in the reporting data mart?
The way I imagined it is we would have a translations table for general use and the accommodations language would just get pushed to it.

agorina:
All what I did it move the reporting pointer to the latest schema. I think @rcordeau  maybe able to address your comment better.

PR #109, opened 2017-05-31
Feature/dwr 378 consume ui kit

bgsavage:
This is consumption of the ui-kit within Reporting.  I've also pulled in the UI visualization, but I understand it will be changing soon so I didn't want to spend a ton of time binding up the dynamic levels just yet.  We will need a separate story that has a dependency on the ui-kit completion of this piece.

### DIFF
--- a/webapp/src/main/webapp/.angular-cli.json
+++ b/webapp/src/main/webapp/.angular-cli.json
@@ -9,7 +9,12 @@
       "outDir": "../resources/static",
       "assets": [
         "assets",
-        "favicon.ico"
+        "favicon.ico",
+        {
+          "glob": "**/*",
+          "input": "../node_modules/sbac-ui-kit/dist/images",
+          "output": "assets/image"
+        }
       ],
       "index": "index.html",
       "main": "main.ts",
@@ -18,10 +23,9 @@
       "prefix": "app",
       "mobile": false,
       "styles": [
-        "styles.less",
-        "../node_modules/primeng/resources/primeng.min.css",
-        "../node_modules/primeng/resources/themes/bootstrap/theme.css",
-        "../node_modules/font-awesome/css/font-awesome.css"
+        "../node_modules/font-awesome/css/font-awesome.css",
+        "../node_modules/sbac-ui-kit/dist/css/sbac-ui-kit.min.css",
+        "styles.less"
       ],
       "scripts": [
         "scripts/iris/jquery.js",
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -14,7 +14,7 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^4.1.1",
+    "@angular/animations": "4.1.1",
     "@angular/common": "4.1.1",
     "@angular/compiler": "4.1.1",
     "@angular/core": "4.1.1",
@@ -30,6 +30,7 @@
     "ngx-bootstrap": "^1.6.6",
     "primeng": "^4.0.0",
     "rxjs": "^5.0.1",
+    "sbac-ui-kit": "git+https://github.com/SmarterApp/SBAC-Global-UI-Kit.git#v0.0.1",
     "ts-helpers": "^1.1.1",
     "zone.js": "^0.8.4"
   },
@@ -51,7 +52,6 @@
     "karma-teamcity-reporter": "^1.0.0",
     "phantomjs-prebuilt": "^2.1.14",
     "protractor": "4.0.9",
-    "sbac-ui-kit": "git+https://github.com/SmarterApp/SBAC-Global-UI-Kit.git#feature/npm-standards",
     "ts-node": "1.2.1",
     "tslint": "^4.0.0",
     "typescript": "^2.2.1",
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -51,10 +51,10 @@
     "karma-teamcity-reporter": "^1.0.0",
     "phantomjs-prebuilt": "^2.1.14",
     "protractor": "4.0.9",
+    "sbac-ui-kit": "git+https://github.com/SmarterApp/SBAC-Global-UI-Kit.git#feature/npm-standards",
     "ts-node": "1.2.1",
     "tslint": "^4.0.0",
     "typescript": "^2.2.1",
     "webdriver-manager": "10.2.5"
   }
 }
-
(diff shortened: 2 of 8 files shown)


webapp/src/main/webapp/.angular-cli.json
```diff
@@ -18,10 +23,9 @@
       "prefix": "app",
       "mobile": false,
       "styles": [
-        "styles.less",
-        "../node_modules/primeng/resources/primeng.min.css",
-        "../node_modules/primeng/resources/themes/bootstrap/theme.css",
```
bgsavage: Removed primeng styles in favor of using only styles from our ui-kit for tables.

webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -21,7 +21,7 @@
       </div>
 
       <div>
-        <p-dataTable [value]="filteredGroups" emptyMessage="{{'labels.groups.empty-message' | translate}}">
+        <p-dataTable [value]="filteredGroups" emptyMessage="{{'labels.groups.empty-message' | translate}}" tableStyleClass="table table-striped">
```
bgsavage: tableStyleClass will set the specified styles on the generated <table> element.

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
```diff
@@ -1,33 +1,36 @@
...
+            <button id="advanced" class="form-control">{{'labels.groups.results.adv-filters-button' | translate}} <i
+              class="fa fa-chevron-down" aria-hidden="true"></i></button>
+          </div>
 
+        </div>
       </div>
+
     </div>
-    <div class="panel-footer overflow-hidden pad-bottom-md">
+    <div class="well gray-lightest">
```
bgsavage: Using wells now instead of panels.

webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -4,11 +4,11 @@
       <p>{{ 'labels.home.welcome' | translate }} <a href="">{{ 'labels.home.tutorial-link-text' | translate }}</a></p>
     </div>
     <div class="row row-eq-height">
-      <div *sbAuthorize="['GROUP_PII_READ','INDIVIDUAL_PII_READ']" class="col-md-8 card overflow-hidden margin-right-20">
+      <div *sbAuthorize="['GROUP_PII_READ','INDIVIDUAL_PII_READ']" class="col-md-8 well overflow-hidden margin-right-20">
```
bgsavage: Using wells now instead of card

webapp/src/main/webapp/src/styles.less
```diff
@@ -582,21 +462,21 @@ iframe.lg {height: 600px;}
...
-.color-grade-7 { color: @color-grade-7 }
-.color-grade-8 { color: @color-grade-8 }
-.color-grade-9 { color: @color-grade-9 }
+.bg-color-grade-3 { background-color: @color-grade-3 !important; }
+.bg-color-grade-4 { background-color: @color-grade-4 !important; }
+.bg-color-grade-5 { background-color: @color-grade-5 !important; }
+.bg-color-grade-6 { background-color: @color-grade-6 !important; }
+.bg-color-grade-7 { background-color: @color-grade-7 !important; }
+.bg-color-grade-8 { background-color: @color-grade-8 !important; }
+.bg-color-grade-9 { background-color: @color-grade-9 !important; }
```
bgsavage: The type of buttons using these styles haven't been implemented yet in the ui-kit, so just trying to give priority to these colors for now.

webapp/src/main/webapp/src/styles.less
```diff
@@ -669,3 +549,16 @@ a.active {
...
+.ui-unselectable-text {
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -o-user-select: none;
+  user-select: none;
+}
+
+.ui-datatable .ui-sortable-column {
+  cursor: pointer;
```
bgsavage: These are styles that I wanted to keep from the prime-ng as it changes the cursor to pointer on a sortable column and on spam clicking it doesn't highlight the text.

webapp/src/main/webapp/package.json
```diff
@@ -30,6 +30,7 @@
     "ngx-bootstrap": "^1.6.6",
     "primeng": "^4.0.0",
     "rxjs": "^5.0.1",
+    "sbac-ui-kit": "git+https://github.com/SmarterApp/SBAC-Global-UI-Kit.git#v0.0.1",
```
bgsavage: Here's the dependency, tied to tag #v.0.0.1 at the moment.

webapp/src/main/webapp/.angular-cli.json
esotrope: 👍

webapp/src/main/webapp/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^4.1.1",
+    "@angular/animations": "4.1.1",
```
esotrope: what is the functional diff between having a carrot and not in npm - do you know? the definition says it just means you are saying "is compatible with"

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
esotrope: Wondering if we should have Aaron port the "card"s from bootstrap4 since cards are semantically what we are using.

webapp/src/main/webapp/src/app/home/home.component.html
esotrope: Let me check again but i think "INDIVIDUAL_PII*" might actually just be "PII" and i need to change the docs. I'll get back to you on this one.

webapp/src/main/webapp/src/styles.less
esotrope: We should make a primeng.less inside of the ui kit for this eventually i think or like a primeng-plugin.min.css that the kit produces for us to optionally add in

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
bgsavage: They are recommending to use wells at the moment.

webapp/src/main/webapp/src/styles.less
bgsavage: Yeah that would be nice

PR #111, opened 2017-05-31
IT for recent repo additions

esotrope:
ignore the branch name

1. Moves everything from assessment package into exam package for now. Plan is to keep everything packaged here regarding exams and group exams - will move things around once we get a feel for what all the moving pieces are going to be.
2. adds integration tests for reporting.exam.repository.*
3. TODO: add more tests for group exam repos - right now it only tests findAllAssessments


agorina:
I do not mean to pick on you, it is more of a general request for everybody: could we please not merge our PRs until we have good ITs? I am about to start a big refactoring of the schema and I rely on the test to catch what was missed.

PR #113, opened 2017-06-02
more single exam table related simplifications

esotrope:
1. adds enums for subject and assessment type
2. removes now unused assessment repo
3. removes now unneeded logic in group exam service


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
```diff
@@ -13,10 +13,10 @@
     private long id;
     private String name;
     private long gradeId;
-    private long typeId;
-    private long subjectId;
+    private AssessmentType type;
+    private Subject subject;
```
esotrope: @bgsavage - these enums will break the frontend

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Subject.java
```diff
@@ -0,0 +1,33 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+/**
+ * Represents an assessment subject
+ */
```
agorina: Could you please add a comment that this has to match to the entry in the DB?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Subject.java
```diff
@@ -0,0 +1,33 @@
...
+public enum Subject {
+
+    MATH(1),
+    ELA(2);
+
+    private final long value;
```
agorina: Not a big deal, but ```int``` should be big enough.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentType.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Represents an assessment type
+ */
+public enum AssessmentType {
+
+    ICA(1, true),
+    IAB(2, false),
+    SUMMATIVE(3, true);
+
```
agorina: As we have talked, this should be moved into Common. I will do it at some point.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
esotrope: actually scratch that I changed the assessment mapper to fix it

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentType.java
```diff
@@ -0,0 +1,46 @@
...
+    private final long value;
+    private final boolean claims;
+
+    AssessmentType(final long value, final boolean claims) {
```
malaffoon: I know there was some push-back in an earlier review about putting business logic in enums. Not quite sure how i feel about this, but i understand the argument: whether an assessment has claims or not probably doesn't belong in the enum constraint that simply defines the types. OTOH, the enum could be used as the business-logic vehicle.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Subject.java
malaffoon: Hmm, may not be a big deal, but i strongly request it be changed to `int`

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Assessment.java
bgsavage: great thanks!

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentType.java
esotrope: I have no issue with moving it to the service layer - might end up creating repeated logic though. Also, if we adopt the "business logic not belonging in enums" idea fully I would suggest we also move the static "fromValue" mapping methods to the service layer as well. Thoughts?

malaffoon: Good point.

agorina: Just FYI: I am planning on moving this and Subject into Common. Will wait until all open items are resolved and do it then.

esotrope: If we think about it the from value mapping logic is kind of something that should be entirely encapsulated into the repo layer I think

malaffoon: The only counterargument i have is that it is very important that these values are consistent between the various repository implementations. So they should be shared. I guess we could have a shared utility class that does that mapping. Or, we could embed the necessary info in the enum  ;)

esotrope: Did everything just come full circle? x D
If we have to ship a util with every enum that seems a little bit overkill i think.
Should we just leave it the way it is for now?

PR #114, opened 2017-06-02
remove AssessmentType and Subject and use from rdw common

malaffoon:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/AuthorizationService.java
```diff
@@ -6,7 +6,6 @@
 import org.springframework.security.core.GrantedAuthority;
 
 import java.util.Collection;
-import java.util.Map;
```
malaffoon: unrelated, fix warning

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -14,7 +14,6 @@
 import org.springframework.stereotype.Service;
 
 import javax.validation.constraints.NotNull;
-import java.util.Map;
```
malaffoon: unrelated, fix warnings

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -55,7 +54,7 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
                 .firstName(firstName)
                 .lastName(lastName)
                 .email(email)
-                .permissionsById(Maps.uniqueIndex(authorizationService.getPermissions(grants), Permission::getId))
```
malaffoon: not showing in the diff, but permissions was being calculated twice

PR #115, opened 2017-06-02
adds exam item endpoints for items by points earned views

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcAssessmentItemRepository.java
```diff
@@ -0,0 +1,69 @@
...
+class JdbcAssessmentItemRepository implements AssessmentItemRepository {
+
+    private final NamedParameterJdbcTemplate template;
+    private final String findAllForAssessmentQuery;
+    private final String findDifficultyCutsForIdQuery;
+
+    @Autowired
+    public JdbcAssessmentItemRepository(
+            @NotNull final NamedParameterJdbcTemplate template,
+            @Value("${sql.assessment.item.findAllForAssessment}") final String findAllForAssessmentQuery,
```
jtreuting: Seems like there is a mix of constructor injection and private variable injection.  I think we decided on a group for the private fields

esotrope: I think our decision was to use constructor params until there were more than 3 params then use property injection for sql queries

jtreuting: We can get some other thoughts on this, but i thought the 3 parameters was for if a builder was needed and the property injecting of SQL was so that having to provide SQL in the constructor seemed "weird" (not sure the technical term).  :)

For example what you are doing below: https://github.com/SmarterApp/RDW_Reporting/pull/115/files/3e3fe76752bd74bffe3df55cd8ac9299f090c577#diff-e24454c9c3320bdab6085a493598a526R26

agorina: I recall it the same way as Jeff. We have agreed on a private fields for the SQL statements.

esotrope: that works I'll go ahead and change those

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItemDifficulty.java
```diff
@@ -0,0 +1,12 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+/**
+ * Represents the difficulty level of an assessment item
+ */
+public enum AssessmentItemDifficulty {
+
+    EASY,
+    MODERATE,
```
malaffoon: I can't find the SB reference material anywhere but, for some reason, i thought the middle one was MEDIUM. I guess it doesn't matter for the enum that much but just wanted to ask if these values are from SB docs somewhere.

jtreuting: All the design mockups we did had Moderate for what it's worth.  I think they gave us that.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItemScore.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Represents a score for a specific item of on a specific exam
+ */
+public class ExamItemScore {
+
+    private long id;
+    private long examId;
+    private long itemId;
+    private Long points;
+    private int position;
```
malaffoon: Wouldn't `position` be an attribute of the ExamItem, not the ExamItemScore?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItemScore.java
```diff
@@ -0,0 +1,101 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+/**
+ * Represents a score for a specific item of on a specific exam
```
malaffoon: grammar

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItemDifficulty.java
esotrope: I took these names from the view mockups. Either way our views will never directly render these names

webapp/src/main/resources/application.sql.yml
```diff
@@ -113,6 +113,49 @@ sql:
...
+          select idc.moderate_low_end, idc.difficult_low_end
+          from asmt a
+            join item_difficulty_cuts idc
+              on a.type_id=idc.asmt_type_id
+              and a.subject_id=idc.subject_id
+              and a.grade_id=idc.grade_id
+          where a.id=:assessment_id
+
+      findAllForAssessment: >-
+        select i.id,
```
agorina: Could you please clarify why you would need it? 
Just FYI: an assessment may have way more items that are given in an exam to a student. That is why I think I am confused.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItemScore.java
esotrope: We already had a model called ExamItem hence the ExamItemScore model -> the idea here is that the point of interest is the score not the entire item. These may end up being consolidated into the same model and the unneeded fields just left null.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupExamItemController.java
```diff
@@ -0,0 +1,35 @@
...
+public class GroupExamItemController {
+
+    private final GroupExamItemService service;
+
+    @Autowired
+    public GroupExamItemController(@NotNull final GroupExamItemService service) {
+        this.service = service;
+    }
+
+    @GetMapping("/api/groups/{groupId}/assessments/{assessmentId}/examitems")
```
malaffoon: i'm confused by the model here ... is this asking for all items from all exams for all the students in the group for the given assessment?

webapp/src/test/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupExamItemRepositoryIT.java
```diff
@@ -0,0 +1,124 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.opentestsystem.rdw.reporting.security.GroupSecurityContexts.statewide;
+
+
+@RunWith(SpringRunner.class)
+@RepositoryIT
+@ContextConfiguration(classes = {ITDataSourceConfiguration.class, JdbcGroupExamItemRepository.class})
+@Sql(statements = {
```
malaffoon: at this point i would seriously consider a sql file with all this

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupExamItemController.java
esotrope: yes - this is the data necessary to present the items by points earned view

webapp/src/test/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupExamItemRepositoryIT.java
esotrope: agreed

webapp/src/main/resources/application.sql.yml
```diff
@@ -113,6 +113,49 @@ sql:
           and (0=:group_subject_id or a.subject_id=:group_subject_id)
           and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
 
+      item:
+        findAllExamItemScoresForAssessment:
```
agorina: It maybe the name that gets me paused here. An item is not necessary scored. I believe we may have `unscored` items (score = -1).  I do not see you filtering just scored items.  

It looks like you want AllItemsForAnAssessmentAndGroup.

However, I still do not understand how it will work. Are you then will use an exam id to related it to the exam and session?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupExamItemController.java
malaffoon: alrighty then, looks good

webapp/src/main/resources/application.sql.yml
```diff
@@ -113,6 +113,49 @@ sql:
...
+          select ei.id,
+            ei.exam_id,
+            ei.item_id,
+            ei.score,
+            ei.position
+          from exam_item ei
+          where ei.exam_id in (
```
agorina: Assuming you still need this query, a ```join exam e on e.id = ei.exam_id``` will perform better.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItemDifficulty.java
bgsavage: Correct.

webapp/src/main/resources/application.sql.yml
esotrope: In the items by points earned view we need to show student item scores grouped by the claim and target and item position (for fixed form only I assume) - this query gets those data points

webapp/src/main/resources/application.sql.yml
esotrope: good to know - thanks

webapp/src/main/resources/application.sql.yml
esotrope: the "exam item scores" != "scored exam items" i put the "scores" wording in there so that it will be separate from a simple findAll - because this query isn't interested in all of the info associated with a specific exam item but is primarily interested in its score

webapp/src/main/resources/application.sql.yml
agorina: If I understand it correctly you need it for an item in the above query? If this is the case then I would do it all in one (the above) query.

webapp/src/main/resources/application.sql.yml
```diff
@@ -113,6 +113,49 @@ sql:
...
+  assessment:
+    item:
+      findDifficultyCutsForId: >-
+          select idc.moderate_low_end, idc.difficult_low_end
+          from asmt a
+            join item_difficulty_cuts idc
+              on a.type_id=idc.asmt_type_id
+              and a.subject_id=idc.subject_id
+              and a.grade_id=idc.grade_id
+          where a.id=:assessment_id
```
agorina: This is worth a discussion with the team. We may want to pre-compute the levels and store them in the data mart. Maybe we can talk about it on Monday.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupExamItemController.java
bgsavage: Would you mind updating our api docs?  https://confluence.fairwaytech.com/display/SBACDW/Reporting+Application+APIs

webapp/src/main/resources/application.sql.yml
esotrope: Is the suggestion to use a join? That would simplify things a bit.
One thing here is that the difficulty cuts are the same for every item in an assessment.

webapp/src/main/resources/application.sql.yml
esotrope: ok

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/GroupExamItemController.java
esotrope: sure - will do after merge

webapp/src/main/resources/application.sql.yml
```diff
@@ -113,6 +113,49 @@ sql:
...
+            where e.asmt_id=:assessment_id
+            and e.school_year=:school_year
+            and sgm.student_group_id=:group_id
+            and (0=:group_subject_id or a.subject_id=:group_subject_id)
+            and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+          );
+
+  assessment:
+    item:
+      findDifficultyCutsForId: >-
```
bgsavage: `findDifficultyCutsForAssessmentId` would be clearer.

webapp/src/main/resources/application.sql.yml
agorina: I may be wrong, but I thought that "In the items by points earned view we need to show student item scores for *the selected session(s)* grouped by the claim and target and item position. Would like @jtreuting to clarify this.

webapp/src/main/resources/application.sql.yml
agorina: Ok, that makes sense. 
I am still confused on why you would need the all at once. See my comment below.

webapp/src/test/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupExamItemRepositoryIT.java
bgsavage: Can we use our own repo layer to create these records?  If not, maybe a data seeder that iterates through some data and executes the same insert statements?  I'm only saying this because if the schema changes, this will be a pain to update.

webapp/src/main/resources/application.sql.yml
agorina: Just for the people who maybe curious:  Landon and I have talked and we understand the challenges here. Landon will have a follow up discussion to see if it needs to be changed or not.

webapp/src/main/resources/application.sql.yml
esotrope: this query might go away so i think i'll leave it for now - we might opt for precomputing this for the asmt items

PR #118, opened 2017-06-05
student now holds ethnicity ids instead of codes

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Student.java
```diff
@@ -84,8 +84,8 @@ public Builder genderId(final long genderId) {
             return this;
         }
 
-        public Builder ethnicityCodes(final List<String> ethnicityCodes) {
-            this.ethnicityCodes = ethnicityCodes;
+        public Builder ethnicityIds(final List<Long> ethnicityCodes) {
+            this.ethnicityIds = ethnicityCodes;
```
agorina: The ethnicity IDs is a DB generated value and could change from the installation to installation.
Some states may require different values. 
So the numeric ID really is not mapped to anything that is controlled by us.
How are you planning on converting it into a description? 

It seems to me that you may want a shorter code that is controlled by us.

esotrope: I have an ethnicity controller/service/repo on standby for this purpose - so the reporting app can ask the backend what ethnicities there are and what their codes are for translation. because of these concerns you just raised i will go ahead and add those to this PR

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcGroupExamRepository.java
```diff
@@ -13,40 +13,44 @@
...
 
+    @Value("${sql.ethnicity.findEthnicityIdsForStudentIds}")
+    private String findEthnicityIdsForStudentIdsQuery;
+
 
-    public JdbcGroupExamRepository(
-            @NotNull final EthnicityRepository ethnicityRepository,
-            @NotNull final NamedParameterJdbcTemplate template) {
-        this.ethnicityRepository = ethnicityRepository;
+    public JdbcGroupExamRepository(@NotNull final NamedParameterJdbcTemplate template) {
```
malaffoon: I'm assuming this is autowired. I'd ask to include the @Autowired annotation and i'd suggest that @NotNull is unnecessary for parameters in an autowired ctor.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Student.java
malaffoon: Are you saying you are changing this from ids to string codes?

esotrope: good catch - I'll be closing this PR for now since we're going with codes anyway - so please ignore

PR #119, opened 2017-06-05
adds exam filter options endpoint. moves student ethnicity lookups to…

esotrope:
… exam repo

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
@@ -1,0 +1,34 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+/**
+ * Represents all of the possible attributes a reporting client can filter exams on
+ */
+public class ExamFilterOptions {
+
+    private List<Integer> schoolYears;
+    private List<String> ethnicities;
+
+    public ExamFilterOptions(final List<Integer> schoolYears, final List<String> ethnicities) {
+        this.schoolYears = schoolYears != null ? ImmutableList.copyOf(schoolYears) : ImmutableList.of();
+        this.ethnicities = ethnicities != null ? ImmutableList.copyOf(ethnicities) : ImmutableList.of();
+    }
+
+    /**
+     * @return all school years available for searching and filtering
+     */
+    public List<Integer> getSchoolYears() {
+        return schoolYears;
+    }
+
+    /**
+     * @return all student ethnicities available for searching and filtering
+     */
+    public List<String> getEthnicities() {
+        return ethnicities;
+    }
+
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
@@ -12,7 +12,7 @@ public class ExamFilterOptions {
     private List<Integer> schoolYears;
     private List<String> ethnicities;
 
-    public ExamFilterOptions(final List<Integer> schoolYears, final List<String> ethnicities) {
+    public ExamFilterOptions(final Iterable<Integer> schoolYears, final Iterable<String> ethnicities) {
         this.schoolYears = schoolYears != null ? ImmutableList.copyOf(schoolYears) : ImmutableList.of();
         this.ethnicities = ethnicities != null ? ImmutableList.copyOf(ethnicities) : ImmutableList.of();
     }
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
@@ -1,33 +1,34 @@
 package org.opentestsystem.rdw.reporting.exam;
 
-import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 
 import java.util.List;
+import java.util.Set;
 
 /**
  * Represents all of the possible attributes a reporting client can filter exams on
  */
 public class ExamFilterOptions {
 
-    private List<Integer> schoolYears;
-    private List<String> ethnicities;
+    private Set<Integer> schoolYears;
+    private Set<String> ethnicities;
 
     public ExamFilterOptions(final Iterable<Integer> schoolYears, final Iterable<String> ethnicities) {
-        this.schoolYears = schoolYears != null ? ImmutableList.copyOf(schoolYears) : ImmutableList.of();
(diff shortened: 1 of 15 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Represents all of the possible attributes a reporting client can filter exams on
+ */
+public class ExamFilterOptions {
+
+    private List<Integer> schoolYears;
+    private List<String> ethnicities;
+
+    public ExamFilterOptions(final List<Integer> schoolYears, final List<String> ethnicities) {
```
malaffoon: params could be `Iterable<Integer>` -- usual argument of making input as least constrained as possible

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
```diff
@@ -0,0 +1,34 @@
...
+        this.schoolYears = schoolYears != null ? ImmutableList.copyOf(schoolYears) : ImmutableList.of();
+        this.ethnicities = ethnicities != null ? ImmutableList.copyOf(ethnicities) : ImmutableList.of();
+    }
+
+    /**
+     * @return all school years available for searching and filtering
+     */
+    public List<Integer> getSchoolYears() {
```
jtreuting: Since the school year can't be NULL, shouldn't this be `List<int>`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/JdbcEthnicityRepository.java
```diff
@@ -1,71 +1,29 @@
...
 
-    @Value("${sql.ethnicity.findAll}")
-    private String findAllQuery;
-
-    @Value("${sql.ethnicity.findEthnicityIdsForStudentIds}")
-    private String findEthnicityIdsForStudentIdsQuery;
-
-    private Map<Long, String> ethnicityCodesById;
+    @Value("${sql.ethnicity.findAllCodes}")
+    private String findAllCodesQuery;
```
malaffoon: Hmm, just noticed that we're inconsistent with naming these variables. Don't change anything but i might make a TODO note for myself to sweep through and make them consistent. Variations include:
* findAllCodesQuery
* findAllCodesSql
* sqlFindAllCodes

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
esotrope: Java doesn't allow primitives to be used as generics I think

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/EthnicityRepository.java
```diff
@@ -1,20 +1,15 @@
...
 
     /**
-     * Gets ethnicity codes for a group of students indexed by student ID
-     *
-     * @param studentIds the students for which the ethnicity codes are desired
-     * @return a map of ethnicity codes by student ID
+     * @return all system ethnicity codes
      */
-    Map<Long, List<String>> findEthnicityCodesForStudentIds(Collection<Long> studentIds);
+    List<String> findAllCodes();
```
malaffoon: Perhaps return `Set<String>` since it is implied that it is a group of unique values and order isn't significant.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
malaffoon: @esotrope is correct. 
Java sucks.
Long live java.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/EthnicityRepository.java
esotrope: Yeah - can do. I prefer that actually.
Also, student ehtnicityCodes should really be a set - I think I prematurely optimized that one

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
bgsavage: I can have List<int> in C# all day every day 🥇

jtreuting: i didn't see that response coming :) . okay, cool.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/EthnicityRepository.java
malaffoon: Hmm. Unless ethnicity order is important. If so, how do we determine it?

esotrope: order isn't important for filtering - there should be a client side order for the ethnicity filter selector though

esotrope: Actually, with this logic we should return school years as a set too no?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+/**
+ * Represents all of the possible attributes a reporting client can filter exams on
+ */
+public class ExamFilterOptions {
+
+    private List<Integer> schoolYears;
```
bgsavage: I'm trying to think if we would need schoolYears is some other context other than for filtering.

esotrope: right now I don't think we do. But if we do it will be easy to expose it again independent of this service.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/EthnicityRepository.java
malaffoon: I thought the school year list was sorted: first entry is "current year".

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamFilterOptions.java
agorina: Mark's next project must be in C#!

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/repository/EthnicityRepository.java
esotrope: I think we could go either way on this one - as long as the front end knows what to do with what it gets we should be fine.

PR #120, opened 2017-06-06
DWR-215

malaffoon:
This adds the diagnostic end-point to the reporting apps. I did the same thing for the web app that i did for import-service: the api (and ui) serves up on 8080 with no change, the health/status end-points are served up on 8008 with no security.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,8 +123,8 @@ The artifact is a Spring Boot executable jar so you can just run it. Just as whe
 is to run without a configuration server so the configuration must be specified on the command line. You can also
 change the port if desired, e.g.:
 ```bash
-java -jar build/libs/rdw-reporting*.jar --spring.config.location=/opt/rdw-reporting/config/application.yml --server.port=8081
-open http://localhost:8081
+java -jar build/libs/rdw-reporting*.jar --spring.config.location=/opt/rdw-reporting/config/application.yml --server.port=8088
+open http://localhost:8088
 ```
 
 ### Posterity
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -30,23 +30,24 @@ services:
     container_name: config-server
     ports:
       - 8888:8888
-    depends_on:
+    links:
       - rabbitmq
     environment:
       - CONFIG_SERVICE_REPO
       - ENCRYPT_KEY
       - GIT_USER
       - GIT_PASSWORD
+      - SPRING_RABBITMQ_ADDRESSES=rabbitmq:5672
+      - SPRING_RABBITMQ_USERNAME=guest
+      - SPRING_RABBITMQ_PASSWORD=guest
 
   webapp:
     image: fwsbac/rdw-reporting-webapp
     ports:
       - 8080:8080
+      - 8081:8008
     volumes:
       - /opt/rdw-reporting/config:/opt/rdw-reporting/config
-    depends_on:
-      - config-server
-      - rabbitmq
     links:
       - config-server
       - rabbitmq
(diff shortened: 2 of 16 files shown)


docker/docker-compose.yml
```diff
@@ -30,23 +30,24 @@ services:
...
+      - SPRING_RABBITMQ_PASSWORD=guest
 
   webapp:
     image: fwsbac/rdw-reporting-webapp
     ports:
       - 8080:8080
+      - 8081:8008
     volumes:
       - /opt/rdw-reporting/config:/opt/rdw-reporting/config
-    depends_on:
```
malaffoon: As noted elsewhere `depends_on` is redundant with `links`.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -482,6 +482,7 @@ protected void configure(HttpSecurity http) throws Exception {
         http
                 .authorizeRequests()
                 .antMatchers("/saml/**").permitAll()
+                .antMatchers("/{status:(status|health|info|env|configprops|autoconfig)}/**").permitAll()
```
malaffoon: This.

Because the SAML configuration completely takes over security, we have to define the permissions here. The problem, of course, is that there are other actuator end-points and this exposes only the hard-coded list given here. The only two that really matter are `status` and `health` but i still don't like hard-coding this.

webapp/src/main/resources/application.yml
```diff
@@ -1,3 +1,9 @@
+# actuator end-points should be served from a different port than the API/web content
+management:
+  port: 8008
+  security:
+    enabled: false
```
malaffoon: This line doesn't actually do anything since the management security is not actually being used. But i had it in during testing and just didn't want to remove it.

webapp/src/test/java/org/opentestsystem/rdw/reporting/web/GroupControllerIT.java
```diff
@@ -23,7 +22,7 @@
 @RunWith(SpringRunner.class)
 @WebMvcTest(GroupController.class)
 @WithMockUser(username = "test")
-public class GroupControllerTest {
+public class GroupControllerIT {
```
malaffoon: This stands up Spring Boot so is more properly an IT. I noticed because i hadn't finished configuring the Spring framework and my `gradle clean build` (without IT) was failing.

webapp/src/test/java/org/opentestsystem/rdw/reporting/web/TranslationControllerIT.java
```diff
@@ -20,7 +19,7 @@
 @RunWith(SpringRunner.class)
 @WebMvcTest(TranslationController.class)
 @WithMockUser(username="test")
-public class TranslationControllerTest {
+public class TranslationControllerIT {
```
malaffoon: ditto

CONTRIBUTING.md
```diff
@@ -123,8 +123,8 @@ The artifact is a Spring Boot executable jar so you can just run it. Just as whe
 is to run without a configuration server so the configuration must be specified on the command line. You can also
 change the port if desired, e.g.:
 ```bash
-java -jar build/libs/rdw-reporting*.jar --spring.config.location=/opt/rdw-reporting/config/application.yml --server.port=8081
-open http://localhost:8081
+java -jar build/libs/rdw-reporting*.jar --spring.config.location=/opt/rdw-reporting/config/application.yml --server.port=8088
+open http://localhost:8088
```
malaffoon: Just changed the example so it wasn't the same as the default docker mapping.

webapp/src/test/java/org/opentestsystem/rdw/reporting/web/GroupControllerIT.java
esotrope: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -463,10 +463,10 @@ public void configure(WebSecurity web) {
     }
 
     /**
-     * Defines the web based security configuration. .anyRequest().authenticated();
+     * Defines the web based security configuration.
      *
      * @param http It allows configuring web based security for specific http requests.
-     * @throws Exception
+     * @throws Exception if an error occurs
```
esotrope: isn't this like saying it throws an exception if an exception occurs?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
esotrope: There is probably a way to separate this http config so that it is not in the saml security config file. ideally different endpoints should be able to have different security models.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
malaffoon: yeah, the only reason i added that was to get rid of IntelliJ warning

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
malaffoon: Yeah, i'm sure a Spring security expert could figure out how to reinstall the actuator security filter but i just couldn't get it to work for me. I might spend another hour on it but ...

PR #124, opened 2017-06-07
Feature/grade search

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Maps.newHashMap;
+import static com.google.common.collect.Sets.newHashSet;
+import static org.opentestsystem.rdw.reporting.jdbc.QueryUtils.getNullable;
+
+
+public abstract class AbstractGroupExamRepository {
```
esotrope: This class is for the shared DB result mapping logic between the group and school grade exam search pages

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/DefaultDistrictService.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.google.common.collect.ImmutableSet;
+import org.opentestsystem.rdw.reporting.security.User;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import javax.validation.constraints.NotNull;
+import java.util.Set;
+
+@Service("schoolGradeDefaultDistrictService")
```
esotrope: added feature package prefixed names because spring threw an exception for default bean names being the same.
conflict was specifically:
...reporting.exam.search.grade.DistrictRepository
...reporting.security.auth.DistrictRepository

They both have the same default bean name: "districtRepository"

webapp/src/main/resources/application.sql.yml
```diff
@@ -34,17 +34,16 @@ sql:
...
-                e.school_year=:school_year
-                and sgm.student_group_id=:group_id
-                and (0=:group_subject_id or a.subject_id=:group_subject_id)
-                and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
-                and a.id = e.asmt_id
+              where e.school_year=:school_year
+              and sgm.student_group_id=:group_id
+              and (0=:group_subject_id or a.subject_id=:group_subject_id)
+              and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+              and a.id = e.asmt_id
```
esotrope: formatting change only

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/DistrictController.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.validation.constraints.NotNull;
+import java.util.Set;
+
+@RestController
+class DistrictController {
+
+    private final DistrictService service;
+
+    @Autowired
+    public DistrictController(@NotNull final DistrictService service) {
```
malaffoon: `@NotNull` isn't needed

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/ExamLocation.java
```diff
@@ -0,0 +1,31 @@
+package org.opentestsystem.rdw.reporting.exam.search.grade;
+
+/**
+ * Represents a named exam location or location qualifier.
+ * In practice, this equates to a district or school.
+ */
+public class ExamLocation {
```
malaffoon: I think `ExamLocation` is misleading. It isn't where the test was taken or anything like that, it is the responsible institution. I think `Institution` would be fine. Or `Organization`.

esotrope: I thought the responsible institution meant the location the exam was administered/taken?
Also, this model can be a district or school so I picked something more generic like location.
Another option is just to make a District and School model with the same props

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,209 @@
...
+        final Assessment.Builder builder = Assessment.builder()
+                .id(row.getLong("id"))
+                .name(row.getString("label"))
+                .gradeId(row.getLong("grade_id"))
+                .type(type)
+                .subject(Subject.valueOf(row.getInt("subject_id")))
+                .schoolYear(row.getLong("school_year"));
+
+        if (type == AssessmentType.IAB) {
+            // TODO: need to add IAB specific logic
```
agorina: Not sure you need anything specific here. As long as you support null values for cut_point_1 and 3, you should be fine with the same logic. UI, on the other hand may care.

esotrope: Did we every figure out what we are doing to display IAB scoring? This is kind of a placeholder for what we decide to do for that

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/ExamLocation.java
agorina: I think it means the school/district a student was enrolled into at the time of the test. I do not think it is a location of the test.
I believe we have agreed to call it with a generic name 'organization'.

esotrope: ok cool - I'll change the name and javadoc

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/GradeController.java
```diff
@@ -0,0 +1,30 @@
...
+    @Autowired
+    public GradeController(@NotNull final GradeService service) {
+        this.service = service;
+    }
+
+    @GetMapping("/api/schools/{schoolId}/grades")
+    public Set<Grade> getDistricts(
+            @AuthenticationPrincipal final User user,
+            @PathVariable final long schoolId) {
+        return service.getGradesInSchool(user, schoolId);
```
agorina: Since there are two concept of grades: grades the student are enrolled into and grade an assessment is designed for, I suggest we need to clarify what grade we are talking about.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/GradeService.java
```diff
@@ -0,0 +1,22 @@
...
+interface GradeService {
+
+    /**
+     * Gets all the grades for which there were exams in a given school
+     *
+     * @param user the user accessing the grades
+     * @param schoolId the ID of the school to search in
+     * @return all the grades for which there were exams in a given school
+     */
+    Set<Grade> getGradesInSchool(@NotNull User user, long schoolId);
```
agorina: Same comment about grade.

We have a requirement in phase 2 to support both. So we should be clear what we are talking about.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/JdbcSchoolGradeExamRepository.java
```diff
@@ -0,0 +1,93 @@
...
+                        .put("school_id", schoolId)
+                        .put("grade_id", gradeId)
+                        .put("school_year", schoolYear)
+                        .build()
+        );
+
+    }
+
+    @Override
+    public List<Assessment> findAll(
```
agorina: Since this is an Exam repository, it is strange that it returns Assessments. 
It this is the best way you see, then I suggest to name it explicitly: getAllAssessments, but I wonder why it is in this repo?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/GradeController.java
esotrope: Would javadoc explanation on method params/return suffice? this model could be used for both enrolled and asmt grade

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/JdbcSchoolGradeExamRepository.java
esotrope: It is related to exams in that it is getting assessments for which there are exams

webapp/src/main/resources/application.sql.yml
```diff
@@ -131,6 +130,102 @@ sql:
...
+          e.claim4_category,
+          st.id as student_id,
+          st.last_or_surname as student_last_name,
+          st.first_name as student_first_name,
+          st.gender_id as student_gender_id
+        from exam e
+          join student st on e.student_id=st.id
+          join asmt a on e.asmt_id=a.id
+          join school s on e.school_id=s.id
+        where e.asmt_id=:assessment_id
```
agorina: Since you are joining with ```assessment``` here and since it now has the claim score names, I think you can change the logic in the AbstractGroupExamRepository.java starting on line 134 to be dynamically driven by the data. Let me know if you want to chat about it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
agorina: I believe we did. The short version I think it that there are min/max and the middle cut point.

esotrope: ok. Will we be using cut point 1 for the midpoint?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/GradeController.java
agorina: I see how the model can be reused, but not the service/controller?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/JdbcSchoolGradeExamRepository.java
agorina: Maybe a name that clarifies it then?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
agorina: 2 is used for the midpoint

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private Map<Long, Set<String>> findEthnicityCodesForStudentIds(final Collection<Long> studentIds) {
+        if (studentIds == null || studentIds.isEmpty()) {
+            return newHashMap();
+        }
+        final Map<Long, String> ethnicityCodesById = findAll();
+        final Map<Long, Set<String>> ethnicityCodesByStudentId = newHashMap();
+        template.query(
+                findEthnicityIdsForStudentIdsQuery,
```
bgsavage: The method says `codes` but the property says `ids`.  Which is it? :)

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+    private Assessment buildAssessment(final ResultSet row, final int index) throws SQLException {
+
+        final AssessmentType type = AssessmentType.valueOf(row.getInt("type_id"));
+
+        final Assessment.Builder builder = Assessment.builder()
+                .id(row.getLong("id"))
+                .name(row.getString("label"))
+                .gradeId(row.getLong("grade_id"))
+                .type(type)
+                .subject(Subject.valueOf(row.getInt("subject_id")))
```
bgsavage: nit: `subject` should be `subjectId` to be consistent with `gradeId`

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private Assessment buildAssessment(final ResultSet row, final int index) throws SQLException {
+
+        final AssessmentType type = AssessmentType.valueOf(row.getInt("type_id"));
+
+        final Assessment.Builder builder = Assessment.builder()
+                .id(row.getLong("id"))
+                .name(row.getString("label"))
+                .gradeId(row.getLong("grade_id"))
```
bgsavage: gradeId is a `long`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+                .type(type)
+                .subject(Subject.valueOf(row.getInt("subject_id")))
+                .schoolYear(row.getLong("school_year"));
+
+        if (type == AssessmentType.IAB) {
+            builder.cutPoints(ImmutableList.of(
+                    row.getLong("min_score"),
+                    getNullable(row, row.getLong("cut_point_1")),
+                    row.getLong("cut_point_2"),
```
bgsavage: Why are cut points 1 and 3 nullable but 2 is not?  A comment might be good here.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+                            .iep(row.getBoolean("iep"))
+                            .lep(row.getBoolean("lep"))
+                            .section504(row.getBoolean("section504"))
+                            .economicDisadvantage(row.getBoolean("economic_disadvantage"))
+                            .migrantStatus(getNullable(row, row.getBoolean("migrant_status")))
+                            .build()
+                    );
+            //.student() set in post process
+
+            if (type.hasClaims()) {
```
bgsavage: Why not let the data drive this?  If a `type` (is this assessment type?  the context isn't clear here), shouldn't have claims, then there won't be any claims in the database right?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+                final ImmutableList.Builder<ScaleScore> claimScaleScores = ImmutableList.builder();
+
+                // MATH exams only have 3 claims, ELA have 4
+                final int endIndex = row.getObject("claim4_category", Integer.class) != null ? 4 : 3;
```
bgsavage: nit: constants for the magic numbers so you don't need the comment.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: for this one it's actually not the ID though it's the enum value

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: will change field name

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+                for (int i = 1; i <= endIndex; i++) {
```
bgsavage: @agorina , If this is truly a 1:many relationship, should we break this out to another table?  Now we are forced to have this weird looping through columns logic to build the collection.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: I haven't really been too concerned with the field memory footprint - I figured I would go through and optimize it when we get to performance testing

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: this is an IAB specific thing. basically the UI will be using min, max and the midpoint (cut_point_2) for IAB scale score computation. I guess the TRT sometimes supplies cut_point_1,3 for IAB but we aren't interested in it

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
agorina: The idea was to make it simpler for the reporting app.
 It is in a 1:many in the warehouse.
The challenge with that is that then you have many rows for one exam.
If it feels better, we can do it, but I do not see how, but open to discuss.

bgsavage: I was thinking we wouldn't even be returning claims here, only as a separate call once the user clicks "Items by Points"

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: Used to have constants but I found that less readable.
It's looking for a fourth - if found expect 4 - if not expect 3

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
bgsavage: Why wait until then if we know we don't need such a large range?  I'm okay with int as a default, but not long.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
agorina: Sure, we can change it. I would like to see the requirements for the export before we commit to this change. The use case we were trying to optimize for is "give me all exams for a group of users with all the data in one row". If this is not a valid use case, then we can change it. 
@jtreuting, would like to know if you have any comments about possible use cases.

Let's chat tomorrow after the stand up?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
```diff
@@ -0,0 +1,214 @@
...
+                    .ethnicityCodes(ethnicityCodes)
+                    .build()
+            );
+            exams.add(examBuilder.build());
+        }
+
+        return exams;
+    }
+
+    private Map<Long, Set<String>> findEthnicityCodesForStudentIds(final Collection<Long> studentIds) {
```
bgsavage: Is this method just needed until we denormalize ?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: We would have to make a second db call where we pass all the found exam ids, get a map of ids to exam claim lists and then map it. we basically do the same for student ethnicity - not sure which is better

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: will still be needed after denormalization - the findAll() method will go away though

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: Yeah I kind of feel the same - just add them if they exist and don't if they don't kind of thing?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
esotrope: I think i'm going to make it BigDecimal just for you : J

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/DefaultDistrictService.java
bgsavage: oh, can we put name spacing in the bean names as well?

esotrope: yeah good point - there might be some spring boot config for that

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/SchoolGradeExamRepository.java
```diff
@@ -0,0 +1,63 @@
...
+import org.opentestsystem.rdw.reporting.exam.Assessment;
+import org.opentestsystem.rdw.reporting.exam.Exam;
+import org.opentestsystem.rdw.reporting.exam.GroupExam;
+import org.opentestsystem.rdw.reporting.security.PermissionScope;
+
+import javax.validation.constraints.NotNull;
+import java.util.List;
+
+/**
+ * Repository responsible for accessing exams by school grade
```
bgsavage: The intent of this repo is confusing to me because it's getting exams by more than just grade, it's getting by school, grade, schoolYear, and permissions.

Also there's a method to get assessments.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/search/grade/JdbcSchoolGradeExamRepository.java
bgsavage: I'd advocate for separate repos for exams and assessments, where the assessments repo allows you to build some search parameters based on your exams.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
bgsavage: 💯  :D

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
bgsavage: Yeah exactly.  We should be able to trust our import / warehouse / migrate to do the right thing here for us, right?

agorina: @esotrope  and I talked offline that this logic should be data driven. I do not think it need to know the type of the exam or subject.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
bgsavage: Yeah let's discuss more tomorrow.  I was thinking we should return this data upon request as it's in a different view.  We don't even know if the user will make this click, and could have a lot of assessments loaded on their page, and would end up with a lot of waste.

agorina: Agree, but as I said before I think there is one view that lists them all together. We need to clarify it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractGroupExamRepository.java
bgsavage: I'm not following why we need to look up specific ethnicity codes for a given list of student ids?

esotrope: otherwise i would have to make one call per student.
this makes all calls at once basically.

PR #128, opened 2017-06-09
enable caching

malaffoon:


webapp/src/main/java/org/opentestsystem/rdw/reporting/Application.java
```diff
@@ -7,9 +7,16 @@
...
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.cloud.aws.autoconfigure.cache.ElastiCacheAutoConfiguration;
+import org.springframework.cloud.aws.autoconfigure.context.ContextStackAutoConfiguration;
 import org.springframework.context.annotation.Import;
 
-@SpringBootApplication
+@SpringBootApplication(exclude = {
+        ElastiCacheAutoConfiguration.class,
```
malaffoon: Without this, deployment on ec2 tries to use aws caching.

esotrope: interesting. is aws caching a distributed cache?

PR #129, opened 2017-06-09
Feature/dwr 381 apply student filters

bgsavage:
This pr adds functionality for all the student filters.  All advanced filters should now be functional.  Also fixes dwr 416.


webapp/src/main/webapp/src/app/groups/results/exam-filters/exam-filter-options.mapper.ts
```diff
@@ -0,0 +1,17 @@
...
+@Injectable()
+export class ExamFilterOptionsMapper {
+  mapFromApi(apiModel: any): ExamFilterOptions {
+    let uiModel: ExamFilterOptions = new ExamFilterOptions();
+
+    uiModel.schoolYears = [];
+    apiModel.schoolYears.forEach(year => uiModel.schoolYears.push(year));
+
+    uiModel.ethnicities = [];
+    apiModel.ethnicities.forEach(ethnicity => uiModel.ethnicities.push(ethnicity));
```
esotrope: this wouldn't be necessary with :any typed objects right?
eventually are we going to use the server response without mappers?
I think less inconsistency between ui and api server would be good in the end when things on the api side settle

bgsavage: I prefer to keep mappers around to protect against any future api changes, regardless of stability of the api.  It's nice to only have to update the mapper if an api model changes, instead of trying to search all usages in html and ts files.  Also ui models and api models after different concerns and therefore tend to be structured differently.

webapp/src/main/webapp/src/app/groups/results/adv-filters/adv-filters.component.html
```diff
@@ -95,7 +94,7 @@
             <strong>{{ translateRoot + 'student.ethnicity' | translate }}</strong>
           </label>
         </div>
-        <sb-checkbox-list [model]="filterBy" [values]="[1,2,3,4,5,6,8]" property="ethnicities" enum="enum.ethnicity"></sb-checkbox-list>
+        <sb-checkbox-list [model]="filterBy" [values]="filterOptions.ethnicities" property="ethnicities" enum="enum.ethnicity"></sb-checkbox-list>
```
bgsavage: This collection holds the ethnicity codes provided by the api/examFilterOptions route.

webapp/src/main/webapp/src/app/groups/results/model/filter-by.model.ts
```diff
@@ -16,30 +16,36 @@ export class FilterBy extends ObservableObject {
...
-        else if (property == "ethnicities" && this.filteredEthnicities.length > 0)
-          all.push("filteredEthnicities");
-        else if (this[ i ] >= 0)
+      if (this.hasOwnProperty(i)) {
+        let property = i.substring(1); // trim leading underscore
+
+        if (property == "ethnicities") {
+          let filteredEthnicities = this.filteredEthnicities;
+          for (let i of filteredEthnicities) {
+            all.push(property + "." + i );
```
bgsavage: I'm not thrilled about this solution of having to do a "ethnicities.ethnicityCode" approach to represent each selected ethnicity.  Ideally would like this all collection to be a collection of `{ property: 'ethnicities', value: 'code' }` but for some reason doing the *ngFor on the complex type instead of on strings was causing the the (click) event (to remove) to become unreliable.  I was never able to figure out why that was breaking so I had to fall back on this lackluster approach which you'll see muddies up the code here and when applying the filters.

webapp/src/main/webapp/src/app/shared/sb-checkbox-list.component.ts
```diff
@@ -59,31 +59,39 @@ export class SBCheckboxList {
...
 
   valueChange(value) {
     // Set all to true if all options are false.
-    this.modelValue[0] = this.allValuesAreFalse();
+    this.modelValue[0] = this.setAllValuesAreFalse();
+    this.modelValue = Object.assign({}, this.modelValue);
   }
 
-  private allValuesAreFalse() : boolean {
+  private setAllValuesAreFalse() : boolean {
```
bgsavage: oh wait what did I do here? :) I'll fix this.

webapp/src/main/webapp/src/app/groups/results/exam-filters/exam-filter-options.mapper.ts
esotrope: ok. If you feel strongly about it. I was thinking it would help me read things if the models are consistent client to api. also gets rid of extra files and stuff. But then again if i did it i wouldn't be using typed objects either so maybe its a good thing for converting to typed objects

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -248,14 +248,14 @@
...
-      "7": "Demographic Race of Two or More",
-      "8": "Filipino"
+      "HispanicOrLatinoEthnicity": "Hispanic / Latino",
+      "AmericanIndianOrAlaskaNative": "American Indian or Alaska Native",
+      "Asian": "Asian",
+      "BlackOrAfricanAmerican": "Black or African American",
+      "White": "White",
+      "NativeHawaiianOrOtherPacificIslander": "Native Hawaiian or Pacific Islander",
+      "DemographicRaceTwoOrMoreRaces": "Demographic Race of Two or More",
+      "Filipino": "Filipino"
```
bgsavage: bound by ethnicity codes now

webapp/src/main/webapp/src/app/groups/results/exam-filters/exam-filter.service.ts
```diff
@@ -0,0 +1,110 @@
...
+  private filterByEconomicDisadvantage(exam: Exam, filterValue: any) {
+    return exam.economicDisadvantage === Utils.polarEnumToBoolean(filterValue);
+  }
+
+  private filterByLimitedEnglishProficiency(exam: Exam, filterValue: any) {
+    return exam.limitedEnglishProficiency === Utils.polarEnumToBoolean(filterValue);
+  }
+
+  private filterByEthnicity(exam: Exam, filterValue: any) {
+    return exam.ethnicities.some(ethnicity => filterValue.some(code => code == ethnicity));
```
bgsavage: ethnicity intersection between the two collections.

webapp/src/main/webapp/src/app/groups/results/exam-filters/exam-filter-options.mapper.ts
bgsavage: Typed objects are the future on the frontend :) Gives me something I can rely on not to be typo'd or accidentally changed.

PR #131, opened 2017-06-12
Feature/school grade exam items endpoint

esotrope:
1. Adds school grade based exam item api endpoint components and IT
2. Fixes some IT test cases and moves some inline sql to an sql file


webapp/src/test/resources/integration-test-data.sql
```diff
@@ -33,4 +33,37 @@ insert into exam (id, type_id, grade_id, student_id, school_id, iep, lep, sectio
...
+  (-1, -1, -1, 0, 1),
+  (-2, -1, -2, 1, 2);
+
+-- groups
+
+insert into student_group (id, name, school_id, school_year, subject_id, import_id) values
+  (-1, 'group1', -10, 2017, 1, -1);
+
+insert into student_group_membership (student_group_id, student_id) values
+  (-1, -1);
```
esotrope: Right now I'm using pretty much one sql data init file for multiple IT. Is this too centralized for peoples' taste? If so should we have one per package or one per test class?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/GroupGrant.java
```diff
@@ -10,25 +10,25 @@
  */
 public class GroupGrant {
 
-    private final Long id;
-    private final Long subjectId;
+    private final long id;
+    private final int subjectId;
 
-    public GroupGrant(@NotNull final Long id, @NotNull final Long subjectId) {
+    public GroupGrant(@NotNull final Long id, @NotNull final Integer subjectId) {
```
malaffoon: If they can't be null, why not just change to primitive types?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/JdbcGroupGrantRepository.java
```diff
@@ -27,9 +29,9 @@ public JdbcGroupGrantRepository(@NotNull final NamedParameterJdbcTemplate jdbcTe
     public List<GroupGrant> findAllForUsername(final String username) {
         return jdbcTemplate.query(findAllForUsernameQuery,
                 new MapSqlParameterSource("user_login", username),
-                (results, index) -> new GroupGrant(
-                        results.getLong("id"),
-                        results.getLong("subject_id")
+                (row, index) -> new GroupGrant(
+                        row.getLong("id"),
+                        getNullable(row, row.getInt("subject_id"))
```
malaffoon: But up there it says it can't be null?

webapp/src/test/resources/integration-test-data.sql
malaffoon: Personally, i prefer my tests (and their data) to be very separate. Until recently i didn't even like having @Before set up the object-under-test. That said, there would be a large amount of duplication here since there is a fair amount of common data. So sharing might be reasonable. However, i think it would help a lot to put copious comments in this file to indicate which test is supported by which set of data. Otherwise, one test may blindly take advantage of data that it doesn't really "own" and later changes will break things.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/DefaultSchoolGradeExamItemService.java
```diff
@@ -0,0 +1,47 @@
...
+        this.examItemRepository = examItemRepository;
+    }
+
+    @Override
+    public GroupExamItemScores getExamItemScores(
+            @NotNull final User user,
+            final long schoolId,
+            final int gradeId,
+            final long assessmentId,
+            final int schoolYear) {
```
bgsavage: Might want to consider creating a class to pass around to avoid having such a long parameter list.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/SchoolGradeExamItemControllerIT.java
```diff
@@ -0,0 +1,63 @@
...
+                .andExpect(jsonPath("$.examItems", hasSize(0)));
+    }
+
+    @Test
+    public void getAllForAssessmentShouldReturnAllAssessmentsWhenFound() throws Exception {
+
+        final AssessmentItem assessmentItem = AssessmentItem.builder().id(900).build();
+        final ExamItemScore examItem = ExamItemScore.builder().id(200).build();
+
+        when(service.getExamItemScores(any(User.class), any(Long.class), any(Integer.class), any(Long.class), any(Integer.class)))
```
malaffoon: I would suggest making a couple of these matchers more specific, and then have input values that differ. For example, if the parameter order is messed up and schoolId and assessmentId are flipped, this test won't catch that.
```
    when(service.getExamItemScores(any(User.class), eq(1), eq(2), eq(3), eq(2017))).thenReturn(...);
    ...
    mvc.perform(get("/api/schools/1/assessmentGrades/2/assessments/3/examitems?schoolYear=2017))
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/DefaultSchoolGradeExamItemService.java
malaffoon: In a past life we always bundled stuff like this in a `SearchParams` class. Then you could create mvc utils to grab it, etc. But that would requiring changing the api to not have them as path variables, but request params.

webapp/src/test/resources/integration-test-data.sql
bgsavage: I think it's better to keep the sql separate from the actual data in order to keep a nice separation of concerns.  Let the tests own the data and have some test repo persist the data.  That being said, I would also encourage not to have so many integration tests, but add unit tests mocking the data layer where possible, in order to avoid complexities such as this :)

agorina: I would like to disagree with the comment about 'not to have so many integration tests' when it comes to the repository testing. I do not see the value in mocking the data layer. We do not have any other ways of testing SQL statements, so I feel that it is pretty important to have them in this case.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/GroupGrant.java
esotrope: There is some logic, probably in the repo, to convert null to 0 to represent both subjects I think though. I definitely did something goofy here - I'll have to fix this

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/DefaultSchoolGradeExamItemService.java
esotrope: Yeah that is on my todo list. I think I'll add it in in a cleanup PR after this one

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/SchoolGradeExamItemControllerIT.java
esotrope: good point I'll - I'll add that in

webapp/src/test/resources/integration-test-data.sql
bgsavage: I'm not as familiar as unit testing the data layer, but it can be done yeah?  

https://dev.mysql.com/doc/mysqltest/2.0/en/unit-tests.html

esotrope: Is this something that would go in the reporting app? That's where the queries live.

PR #135, opened 2017-06-13
fixes controller method names. fixes @PreAuthorize annotations on ser…

esotrope:
…vices


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamController.java
```diff
@@ -23,7 +23,7 @@ public GroupExamController(@NotNull final GroupExamService service) {
     }
 
     @GetMapping("/api/groups/{groupId}/assessments/{assessmentId}/exams")
-    public List<Exam> getGroupIabs(
+    public List<Exam> getExams(
```
bgsavage: that was fast :)

esotrope: 😆

PR #136, opened 2017-06-14
Renamed asmt table claim columns to be consistent with the exam table.

agorina:


agorina:
Merging to be in synch with the other modules in case somebody tries to depoy. If any comments come up - I will address them.

PR #137, opened 2017-06-15
Add localhost:8080 /api proxying.

wtritch:
Add generic HTML5 URI routing to index.html

Just some minor mods that helped me get set up for front-end development.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/NotFoundIndexTemplate.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/NotFoundIndexTemplate.java
@@ -1,0 +1,19 @@
+package org.opentestsystem.rdw.reporting.web;
+
+import org.springframework.boot.context.embedded.ConfigurableEmbeddedServletContainer;
+import org.springframework.boot.context.embedded.EmbeddedServletContainerCustomizer;
+import org.springframework.boot.web.servlet.ErrorPage;
+import org.springframework.http.HttpStatus;
+
+/**
+ * This customizer is responsible for re-directing any unbound request URI to the root
+ * path.  This allows us to handle HTML5-style Single-Page-App URIs.
+ */
+public class NotFoundIndexTemplate implements EmbeddedServletContainerCustomizer {
+
+    @Override
+    public void customize(final ConfigurableEmbeddedServletContainer container) {
+        container.addErrorPages(new ErrorPage(HttpStatus.NOT_FOUND, "/"));
+    }
+
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
@@ -6,8 +6,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
 @Controller
 public class ViewController {
 
-	// TODO: Find a better way to maintain all of our view routes as this will be hard to maintain over time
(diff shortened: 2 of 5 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/web/NotFoundIndexTemplate.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * This customizer is responsible for re-directing any unbound request URI to the root
+ * path.  This allows us to handle HTML5-style Single-Page-App URIs.
+ */
+public class NotFoundIndexTemplate implements EmbeddedServletContainerCustomizer {
+
+    @Override
+    public void customize(final ConfigurableEmbeddedServletContainer container) {
+        container.addErrorPages(new ErrorPage(HttpStatus.NOT_FOUND, "/"));
+    }
```
wtritch: This will re-direct unknown URIs to the home page, so we don't have to call out all our HTML5 routes in java.

webapp/src/main/webapp/package.json
```diff
@@ -5,7 +5,7 @@
   "angular-cli": {},
   "scripts": {
     "build": "ng build",
-    "start": "ng serve",
+    "start": "ng serve --proxy-config proxy.conf.json",
```
wtritch: Running `npm start` will now start the front-end live-watch environment with proxying to localhost:8080 /api endpoints with Real Live Data (TM)

webapp/src/main/webapp/proxy.conf.json
```diff
@@ -0,0 +1,6 @@
+{
+  "/api": {
+    "target": "http://localhost:8080",
+    "secure": false
+  }
+}
```
wtritch: Simple proxy config to send all requests to /api/** to the server running on localhost:8080, so we can have a live-watch front-end development environment that's hitting a developer's local running instance.

esotrope: I think this will cause CORs issues no?
The API webapp will always contain the angular client because...SAML so I'm not sure what this buys us.

wtritch: Since localhost shares cookies across ports, you can log in at :8080, then do live-development against :4200

wtritch: Ah, sorry, I misread the comment.  The angular-cli system is proxying requests to :8080/api behind the scenes.  So the browser *thinks* it's talking to localhost:4200

esotrope: Oh very cool. That is an big buy - thanks for the contribution

wtritch: And if we were doing hypermedia payloads, I'm pretty sure there's a way to turn on x-forwarded headers... but I'm not going to mess with it unless we need it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -6,8 +6,7 @@
 @Controller
 public class ViewController {
 
-	// TODO: Find a better way to maintain all of our view routes as this will be hard to maintain over time
-	@RequestMapping(value = {"/", "/groups", "groups/{id}", "/groups/{id}/students", "/groups/{id}/students/{id2}/exams/{id3}/items", "/irispoc"})
+	@RequestMapping(value = {"/"})
```
esotrope: Somehow originally we weren't able to navigate the angular app routes without specifying them here. It looks like that has gone away though - cool

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/NotFoundIndexTemplate.java
bgsavage: 👍

PR #138, opened 2017-06-15
Feature/dwr 382 get available assmts

bgsavage:
This PR is for both DWR 382 and 383.

DWR-382: Group Results Custom: Available Assessments
DWR-383: Group Results Custom: Add / Remove Assessments


bgsavage:
Example usage: 
image

webapp/src/main/webapp/src/app/groups/groups.module.ts
```diff
@@ -17,13 +17,15 @@ import { AdvFiltersComponent } from "./results/adv-filters/adv-filters.component
...
+import { SelectAssessmentsComponent } from './results/select-assessments/select-assessments.component';
 
 @NgModule({
   declarations: [
     GroupsComponent,
     GroupResultsComponent,
     AssessmentResultsComponent,
-    AdvFiltersComponent
+    AdvFiltersComponent,
+    SelectAssessmentsComponent
```
bgsavage: SelectAssessmentsComponent contains assessment grid grouped by grade.

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
```diff
@@ -17,23 +17,26 @@ <h2 class="blue-dark h3 mb-md">
           <div class="flex-child">
             <label for="select-group">{{'labels.groups.results.select-group' | translate}}</label>
             <!--TODO: type ahead? -->
-            <select id="select-group" [(ngModel)]="currentGroup" (change)="updateRoute($event)" class="form-control">
+            <select id="select-group" [(ngModel)]="currentGroup" (change)="updateRoute()" class="form-control">
```
bgsavage: $event wasn't used.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -45,33 +45,29 @@
...
-                      <span class="bar-label">Above Standard</span>
                     </div>
                 </div>
+              <div class="pull-right">
+                <!-- TODO: bind levels to visualization -->
+                <div *ngFor="let level of performance.levels">
+                  {{ examLevelEnum + level.id.toString() | translate }}:
+                  <span title="{{level.value}}">{{ level.value | number:'1.0-0' }}</span>{{ level.suffix }}
+                </div>
+              </div>
```
bgsavage: @aaronsleeper  I moved this back under the visualization to give more space for sessions, this is placeholder anyway for now

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -45,33 +45,29 @@
                         </span>
                     </span>
                     <div class="bar green left-side" data-width="28">
-                      <span class="bar-data">28%</span>
```
bgsavage: @aaronsleeper  I removed hard coded values as to not confuse with the real calculations below it.

webapp/src/main/webapp/src/app/shared/grade.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable } from "@angular/core";
+import { Grade } from "./model/grade.model";
+
+@Injectable()
+export class GradeService {
```
bgsavage: This service holds static grade data, such as their associated colors, and which grades are supported by the system.  I'm thinking we might eventually would like to have a function for the display text toShortString() i.e. which returns "G7" and also toLongString() i.e. which returns "Grade 7".  I think we have a requirement for when grade >=9 to just show it as "High School", so this is where that logic can be placed.

webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -93,7 +157,67 @@ export class GroupResultsComponent implements OnInit {
...
+    }
+    else {
+      this.removeUnselectedAssessmentExams();
+    }
+  }
+
+  private loadAssessmentExam(assessment: Assessment){
+    let subscription = this.assessmentService.getExams(this._currentGroup.id, this._currentSchoolYear, assessment.id)
+      .subscribe(exams => {
+        let assessmentLoaded = this.assessmentsLoading.splice(this.assessmentsLoading.findIndex(al => al.assessment.id == assessment.id), 1);
```
esotrope: Can you help me understand this block is doing? I'm a little thrown by the splice call

webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -93,7 +157,67 @@ export class GroupResultsComponent implements OnInit {
...
+        let assessmentExam = new AssessmentExam();
+        assessmentExam.assessment = assessmentLoaded[0].assessment;
+        assessmentExam.exams = exams;
+
+        this.assessmentExams.push(assessmentExam);
+        this.assessmentExams.sort((x, y) => {
+          if(x.assessment.grade == y.assessment.grade){
+            return x.assessment.name > y.assessment.name ? 1: -1;
+          }
+          return x.assessment.grade > y.assessment.grade ? 1: -1;
```
esotrope: we're probably going to want to include some sorting util since we will be doing client side sorting.

bgsavage: Yeah I was thinking something like that since the sorting is very similar elsewhere.

webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
bgsavage: So the assessmentLoading collection are active assessments in flight.  In this subscribe, we have completed the assessment loading so I'm removing it from the collection.  The reason I'm maintaining this collection so that the spam clicking user doesn't see their assessment result if it is not selected.  Might be overkill.

webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -93,7 +157,67 @@ export class GroupResultsComponent implements OnInit {
...
+  }
+
+  private removeUnselectedAssessmentExams() {
+    this.assessmentExams = this.assessmentExams.filter(loaded => this.selectedAssessments.some(selected => loaded.assessment.id == selected.id));
+
+    let assessmentsLoading = this.assessmentsLoading.filter(loading => !this.selectedAssessments.some(selected => loading.assessment.id == selected.id ));
+
+    // cancel api calls in flight.
+    for(let loading of assessmentsLoading){
+      loading.subscription.unsubscribe();
```
bgsavage: Cancel api calls if the user has unselected an assessment before the api could return a result.

PR #142, opened 2017-06-16
Feature/abstract exam search

esotrope:
1. Creates *Search pojos to pass data from controller down to repo
2. Change 1 allowed for shared methods through abstract classes so I added abstract base classes

Notes:
1. We might be able to remove the 6 classes for ArgumentResolvers if we used @PostMapping for exam search controllers
2. We could consolidate *AssessmentSearch and *ExamSearch objects to remove 3 classes - this would mean that there would be a dangling unset assessmentId on requests that do not use that search param though
3. Despite added abstraction there is redundant pojo to sql param mapping going on - not sure if i should delegate all that to the QueryUtils or not


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/AbstractExamRepository.java
```diff
@@ -25,33 +25,47 @@
...
     @Value("${sql.ethnicity.findAll}")
     private String findAllEthnicitiesQuery;
 
     @Value("${sql.ethnicity.findEthnicityIdsForStudentIds}")
     private String findEthnicityIdsForStudentIdsQuery;
 
-
-    @Autowired
-    public AbstractGroupExamRepository(@NotNull final NamedParameterJdbcTemplate template) {
+    public AbstractExamRepository(
```
malaffoon: this should be `protected`

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/AbstractExamSearch.java
```diff
@@ -0,0 +1,28 @@
+package org.opentestsystem.rdw.reporting.search.exam;
+
+public abstract class AbstractExamSearch extends AbstractAssessmentSearch {
+
+    protected long assessmentId;
```
malaffoon: this can't be `private`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/AssessmentRepository.java
```diff
@@ -0,0 +1,32 @@
...
+import org.opentestsystem.rdw.reporting.exam.Assessment;
+import org.opentestsystem.rdw.reporting.security.PermissionScope;
+
+import javax.validation.constraints.NotNull;
+import java.util.List;
+
+/**
+ * Repository responsible for accessing assessments by school and assessment grade
+ */
+public interface AssessmentRepository<T extends AbstractAssessmentSearch> {
```
malaffoon: I haven't gotten through the review yet, but it seems odd that interfaces and non-abstract impl's would have a T that extends an abstract search class. It seems like, at this point, it should be acting on a precisely defined set of search values.

I guess it might just be the fallout of the search object hierarchy. Perhaps multiple search object interfaces instead of a hierarchy would be more appropriate? I'll try to follow-up on this comment.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupAssessmentController.java
```diff
@@ -6,41 +6,36 @@
...
+    GroupAssessmentController(final GroupAssessmentService service) {
         this.service = service;
     }
 
     @GetMapping("/api/groups/{groupId}/latestassessment")
     public GroupAssessment getLatestGroupAssessment(
             @AuthenticationPrincipal final User user,
-            @PathVariable final long groupId,
-            @RequestParam("schoolYear") final long schoolYear) {
+            final GroupAssessmentSearch search) {
```
malaffoon: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/AssessmentRepository.java
malaffoon: Ok, i see. This interface is a base interface, nothing directly implements it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/JdbcGroupAssessmentRepository.java
```diff
@@ -1,51 +1,31 @@
...
-        return super.findLatest(
-                findLatestQuery,
-                ImmutableMap.<String, Object>builder()
-                        .putAll(getSecurityParameters(securityContext))
-                        .put("school_year", schoolYear)
-                        .build()
-        );
+    JdbcGroupAssessmentRepository(
+            @NotNull final NamedParameterJdbcTemplate template,
+            @Value("${sql.group.exam.findLatest}") final String findLatestQuery,
```
malaffoon: Instead of doing this, you could have an abstract getter in the base class; then the child class, instead of having to pass the queries into the base constructor, could have consistent `@Value` member variables and return them in the concrete implementation of the abstract getters.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/AbstractExamSearch.java
esotrope: because of the way it's built from the builder it had to be protected.
Could be private if we pass the builder itself to a private constructor i think

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/JdbcGroupAssessmentRepository.java
esotrope: Good point. I might sweep through and make sure that is done to all abstract repos

PR #143, opened 2017-06-16
Changes SQL after testing on larger volume of data.

agorina:
While researching https://jira.fairwaytech.com/browse/DWR-326 I am testing queries (one at a time) and this is the outcome. 
Here is the data set on the dev. that I tested with:
```
SELECT count(*) FROM reporting.exam; -- 4,230,339
SELECT count(*) FROM reporting.exam_item; -- 75,097,986
SELECT count(*) FROM reporting.school; -- 2,968
SELECT count(*) FROM reporting.district; -- 655
SELECT count(*) FROM reporting.student_group; -- 7,958
SELECT count(*) FROM reporting.student_group_membership WHERE student_group_id = 52; -- 27
```

### DIFF
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -16,7 +16,8 @@ sql:
 
     assessment:
       findAll: >-
-          select a.id,
+          select distinct
+            a.id,
             a.label,
             a.grade_id,
             a.type_id,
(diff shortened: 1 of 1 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -16,7 +16,8 @@ sql:
 
     assessment:
       findAll: >-
-          select a.id,
+          select distinct
```
agorina: Despite the fact that the execution plan looks way worse, it returns significantly faster. The speculation @malaffoon and I have is that because the group size is small.

PR #144, opened 2017-06-16
Defect/security annotations on abstracts

esotrope:
Fixes security rules on group components.
Group components incorrectly had PII_READ permission enforced
This will not affect dwtest@example.com because it had that permission anyway

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/ExamItemService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/ExamItemService.java
@@ -17,7 +17,6 @@ public interface ExamItemService<T extends AbstractExamSearch> {
      * @param search the search parameters to limit the results by
      * @return all the exam item scores and item information for a given assessment
      */
-    @PreAuthorize("hasAuthority('PERM_PII_READ')")
     GroupExamItemScores getExamItemScores(@NotNull User user, @NotNull T search);
 
 }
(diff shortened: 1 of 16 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/DefaultGroupExamItemService.java
```diff
@@ -25,7 +25,6 @@
         this.examItemRepository = examItemRepository;
     }
 
-    @Override
```
malaffoon: This isn't implementing the interface method?

esotrope: Do we wan't to have @Override for implementing methods? i found that it is inconsistently used so i arbitrarily removed it - can add it back

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/ExamItemService.java
```diff
@@ -17,7 +17,6 @@
      * @param search the search parameters to limit the results by
      * @return all the exam item scores and item information for a given assessment
      */
-    @PreAuthorize("hasAuthority('PERM_PII_READ')")
     GroupExamItemScores getExamItemScores(@NotNull User user, @NotNull T search);
```
malaffoon: It's not obvious to me why exam item scores doesn't need protection?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/DefaultGroupExamItemService.java
malaffoon: Hmm. I always put @Override for implementing methods. I thought that was standard (?)

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/ExamItemService.java
esotrope: Its a little hard to follow here but there are kind of "abstract" interfaces here and "concrete" interfaces - what i did here was move the security annotations from the abstract interfaces to the concrete in the packages because the security is package specific

PR #146, opened 2017-06-16
adds school search by name

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/district/DefaultDistrictService.java
```diff
@@ -1,29 +0,0 @@
-package org.opentestsystem.rdw.reporting.search.district;
```
esotrope: Removed district search related components as they are not needed by the UI anymore

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/school/JdbcSchoolRepository.java
```diff
@@ -11,28 +11,35 @@
...
 
     private final NamedParameterJdbcTemplate template;
 
-    @Value("${sql.school.findAllForDistrict}")
-    private String findAllForDistrictQuery;
+    @Value("${sql.school.findAllForNameContainingText}")
+    private String findAllForNameContainingTextQuery;
+
+    @Value("${app.search.schoolsByName.limit}")
+    private int findAllForNameContainingTextLimit;
```
esotrope: Added the query limit here as an app config

webapp/src/main/resources/application.sql.yml
```diff
@@ -295,32 +295,18 @@ sql:
...
       from school
-      where district_id=:district_id
-      and (
-        1=:statewide
-        or district_id in (:district_ids)
-        or id in (:school_ids)
-      );
+      where (1=:statewide or district_id in (:district_ids) or id in (:school_ids))
+      and name like :text
+      limit :limit
```
esotrope: @agorina @malaffoon here is the new school search by name

webapp/src/main/resources/application.yml
```diff
@@ -48,4 +48,9 @@ artifacts:
...
 
 permissionservice:
-  component: Reporting
\ No newline at end of file
+  component: Reporting
+
+app:
+  search:
+    schoolsByName:
+      limit: 10
```
esotrope: @wtritch - defaulted search result limit to 10

malaffoon: Does the default really need to be configurable? I guess it doesn't hurt but ...

esotrope: Ah I see what you're saying - i could put the default in the repo. - this way kind of shows that the property is there to change though so i may like this better

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/school/JdbcSchoolRepository.java
```diff
@@ -11,28 +11,35 @@
...
+            @NotNull final PermissionScope permissionScope,
+            @NotNull final String text) {
         return template.query(
-                findAllForDistrictQuery,
+                findAllForNameContainingTextQuery,
                 ImmutableMap.<String, Object>builder()
                         .putAll(getSecurityParameters(permissionScope))
-                        .put("district_id", districtId)
+                        .put("text", createMatcher(text))
+                        .put("limit", findAllForNameContainingTextLimit)
```
malaffoon: So, for now, the limit can be passed in?

esotrope: right now the only way to set the limit is through the yml file

webapp/src/main/resources/application.yml
malaffoon: Sorry, i was assuming that the value could be passed in at the REST API level, so having a configurable default was silly. But, if it can't be passed in, then it should be configurable, yes.

webapp/src/main/resources/application.sql.yml
agorina: I have a few questions:
1. Is it meant to be used for a type-ahead feature?
2. Is it really a requirement to have a wildcard at the beginning and end of the word?

esotrope: Yes this is meant for type ahead - do you think we can get away with just a wildcard at the end?

agorina: There are two different concerns:
1. What do users expect? I guess this is very subjective, but I would expect if I type 'a' to see everything that starts with 'a'.
2. Performance implications.  You can read more here: https://dev.mysql.com/doc/refman/5.6/en/index-btree-hash.html. In short: _"A B-tree index can be used for LIKE comparisons if the argument to LIKE is a constant string that  **does not** start with a wildcard character."_.  Interestingly: _"If you use ... LIKE '%string%' and string is **longer than three characters**, MySQL uses the Turbo Boyer-Moore algorithm to initialize the pattern for the string and then uses this pattern to perform the search more quickly."_ 
This table is small, and I have tried this query on AWS Dev. -
 it returns fast. I do not know how it will behave with a lot of users. I suspect it is fast now because it has enough memory to load the data. 

If it were up to me, I would just use a wildcard at the end and add an index to the table.

malaffoon: Oh goodie, text search ... never any controversy on how that should behave! A couple points:
1. The table will always be small; CA is by far the largest at about 14000 schools.
2. Although schools tend to have official names that don't match their common name, they usually start with the common name so a trailing wildcard will probably be close to the expected behavior.
3. Any user that is searching for a school is probably going to know their schools pretty well. These won't be teachers, they'll be district-level admin or curriculum leads.

bgsavage: _usually_ is what concerns me.  Maybe an exception to this rule would be when a school is named after a person (i.e. George Washington Elementary is probably more commonly known as Washington Elementary).  

I think it would be reasonable to require at least 4 characters before performing the search.

A possible solution could be to take a hybrid approach such as...

If string length is > 3
    then search by leading AND trailing wild card
else
    search by trailing wild card only.

This can satisfy @agorina 's  expected behavior #1 and the exception to @malaffoon 's assumption #2.

PR #148, opened 2017-06-16
integrates translation service with new table. adds IT

esotrope:


build.gradle
```diff
@@ -92,6 +92,9 @@ subprojects {
             dependency 'javax.validation:validation-api:1.1.0.Final'
             dependency 'org.springframework.boot:spring-boot-starter-jdbc:1.5.2.RELEASE'
             dependency 'org.springframework.boot:spring-boot-configuration-processor:1.5.2.RELEASE'
+
+            dependency 'org.springframework.boot:spring-boot-starter-test:1.5.2.RELEASE'
```
malaffoon: All of these version numbers should be coming from the bom, not being explicitly set here. 

Perhaps you could create a JIRA to fix this?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/SimpleTranslationService.java
```diff
@@ -20,7 +20,7 @@ public SimpleTranslationService(final TranslationRepository repository, final Tr
     }
 
     public Map<String, String> getTranslationsByCode(final Locale locale) {
-        return repository.findAllForLocaleAndNamespace(checkNotNull(locale), namespace)
+        return repository.findAllForLanguageAndNamespace(checkNotNull(locale).getISO3Language(), namespace)
```
bgsavage: Should this be `findAllBy` instead of `findAllFor`?

PR #152, opened 2017-06-20
Feature/dwr 433 items

bgsavage:
Partially implemented Items by points earned, just enough to be able to add Items By Points earned button click to the load test.  The table still needs apply proper sorting, need to be able go back to the previous view, and filtering still needs to be applied.

This also productinalizes the front end, and removes the old UI which we are no longer using.


webapp/src/main/webapp/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "angular-cli": {},
   "scripts": {
-    "build": "ng build",
+    "build": "ng build --prod",
```
bgsavage: This seems to have stricter compile rules than just `ng build` in which a lot of the old UI violates, mainly that you can't access privates from the template.  This seems to be a good time to just remove the old UI we no longer need.  We still need to create a label to that last working copy of the old UI so we can have it for reference for future development (i.e. IRIS / rubric / exemplar integrations)

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-exam.mapper.ts
```diff
@@ -20,21 +22,61 @@ export class AssessmentExamMapper {
...
+    let uiModels: AssessmentItem[] = [];
+
+    for(let apiAssessment of apiModel.assessmentItems){
+      let assessmentItem = this.mapAssessmentItemFromApi(apiAssessment);
+
+      for(let apiExamItem of apiModel.examItems.filter(x => x.itemId == assessmentItem.id)){
+        assessmentItem.scores.push(this.mapExamItemFromApi(apiExamItem));
+      }
+
+      assessmentItem.calculateBuckets();
```
bgsavage: partial implementation here.   I think I want to create a calculator to do this for me instead of having a function call in a mapper.

webapp/src/main/webapp/package.json
bgsavage: Also this brings the 5MB vendor bundle file down to 2.5MB.  main bundle is under 1MB.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -35,13 +36,20 @@ export class AssessmentResultsComponent {
...
+  pointColumns: number[];
+  showItemsByPoints: boolean = false;
 
   @Input()
   set assessmentExam(assessment: AssessmentExam) {
     this._assessmentExam = assessment;
     this.sessions = this.getDistinctExamSessions(assessment.exams);
 
     if (this.sessions.length > 0) {
+      // this.sessions.forEach(x => {
```
jtreuting: seems like this should be removed

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -117,6 +133,18 @@ export class AssessmentResultsComponent {
     this.updateExamSessions();
   }
 
+  viewItemsByPoints() {
+    if(this.loadAssessmentItems) {
```
jtreuting: is it just me that likes the space after the if?  Is this something that has been decided as our standard?  instead of `if (this.leadAssessmentItems) {`

webapp/src/main/webapp/src/app/groups/results/model/assessment-item.model.ts
```diff
@@ -0,0 +1,28 @@
+import { ExamItemScore } from "./exam-item-score.model";
+
+export class AssessmentItem {
+  id: number;
+  claim: string;
+  target: string;
+  difficulty: string;
+  maxPoints: number;
+  scores: ExamItemScore[] = [];
+  buckets = [];
```
esotrope: What is in the buckets? I would try to make this more descriptive

webapp/src/main/webapp/package.json
esotrope: Does this affect our ability to do development .ts debugging? Wondering if we need a separate script entry here for dev and prod build

webapp/src/main/webapp/src/app/groups/results/model/assessment-item.model.ts
bgsavage: partial implementation :)

webapp/src/main/webapp/package.json
bgsavage: These npm scripts only should be used by gradle I think.  We can still run `ng serve` with whatever arguments work best for debugging.

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
bgsavage: I don't think we've standardized on it but I'm open to it.

esotrope: We should probably just do whatever intellij defaults to - which should be a space there if i remember right. The other thing we should make consistent is using brackets for even one-line if blocks

bgsavage: Yeah I think I just need to run option + cmd + L on this file.

PR #153, opened 2017-06-21
Now sorting groups by name.

bgsavage:
Brought in a comparator library now using that to create our comparators.  Created an assessment comparator so for assessments to be sorted by grade then name.

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/groups.resolve.ts
+++ b/webapp/src/main/webapp/src/app/groups/groups.resolve.ts
@@ -2,11 +2,15 @@ import { Injectable } from "@angular/core";
 import { DataService } from "../shared/data.service";
 import { Resolve, ActivatedRouteSnapshot, RouterStateSnapshot } from "@angular/router";
 import { Observable } from "rxjs";
+import { byString } from "@kourge/ordering/comparator";
 
 @Injectable()
 export class GroupsResolve implements Resolve<any> {
   resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<any>|Promise<any>|any {
-    return this.service.getGroups();
+    return this.service.getGroups().toPromise()
+      .then( groups => {
+        return groups.sort((group1, group2) => byString(group1.name, group2.name));
+      });
   }
 
   constructor(private service: DataService) {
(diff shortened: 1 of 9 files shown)


webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -223,14 +219,15 @@ export class GroupResultsComponent implements OnInit {
...
 
     // cancel api calls in flight.
-    for(let loading of assessmentsLoading){
+    for (let loading of assessmentsLoading) {
       loading.subscription.unsubscribe();
     }
 
-    this.assessmentsLoading = this.assessmentsLoading.filter(loading => this.selectedAssessments.some(selected => loading.assessment.id == selected.id ));;
+    this.assessmentsLoading = this.assessmentsLoading.filter(loading => this.selectedAssessments.some(selected => loading.assessment.id == selected.id));
+    ;
```
jtreuting: This some be removed

jtreuting: Typo via phone and it won't let me edit.  That should say.  This should be removed.

bgsavage: oops yes.

webapp/src/main/webapp/src/app/groups/groups.resolve.ts
```diff
@@ -2,11 +2,15 @@ import { Injectable } from "@angular/core";
...
 import { Observable } from "rxjs";
+import { byString } from "@kourge/ordering/comparator";
 
 @Injectable()
 export class GroupsResolve implements Resolve<any> {
   resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<any>|Promise<any>|any {
-    return this.service.getGroups();
+    return this.service.getGroups().toPromise()
+      .then( groups => {
+        return groups.sort((group1, group2) => byString(group1.name, group2.name));
```
esotrope: The ordering lib also has this syntax which i think is a little more fluent

return groups.sort(ordering(byString).on(group => group.name).compare);

bgsavage: I agree, this is cleaner.  Just fyi, one nit from the compiler is we need to specify type T when calling `on`.

So in this case:  `groups.sort(ordering(byString).on<any>(group => group.name).compare);`

PR #156, opened 2017-06-21
Feature/student exams api

esotrope:
To start with, I went with the amorphism approach for the models - and added extra nullable fields to existing models to support the new feature. This allowed me to have a common base for DB row to model mapping.

TODOs:

Decide model strategy (inheritance, amorphism, duplication)

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
@@ -1,0 +1,57 @@
+package org.opentestsystem.rdw.reporting.search.exam;
+
+import com.google.common.collect.ImmutableList;
+import org.opentestsystem.rdw.common.model.AssessmentType;
+import org.opentestsystem.rdw.common.model.Subject;
+import org.opentestsystem.rdw.reporting.exam.Assessment;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public final class Assessments {
+
+    private static final String[] CLAIM_CODE_COLUMN_NAMES = {
+            "claim1_score_code",
+            "claim2_score_code",
+            "claim3_score_code",
+            "claim4_score_code"
+    };
+
+    private Assessments() {
+    }
+
+    public static <A extends Assessment.Builder> A map(final ResultSet row, final A builder) throws SQLException {
+
+        final AssessmentType type = AssessmentType.valueOf(row.getInt("type_id"));
+
+        builder.id(row.getLong("id"))
+                .name(row.getString("label"))
+                .gradeId(row.getInt("grade_id"))
+                .type(type)
+                .subject(Subject.valueOf(row.getInt("subject_id")))
+                .schoolYear(row.getInt("school_year"))
+                .cutPoints(ImmutableList.of(
+                        row.getInt("min_score"),
+                        row.getInt("cut_point_1"),
+                        row.getInt("cut_point_2"),
+                        row.getInt("cut_point_3"),
+                        row.getInt("max_score")
+                ));
+
+        // IABs do not have claims
+        if (type != AssessmentType.IAB) {
+            final ImmutableList.Builder<String> claimCodes = ImmutableList.builder();
+            for (String claimCodeColumnName : CLAIM_CODE_COLUMN_NAMES) {
+                final String claimCode = row.getString(claimCodeColumnName);
+                if (claimCode == null) {
+                    break;
+                }
+                claimCodes.add(claimCode);
+            }
+            builder.claimCodes(claimCodes.build());
+        }
+
+        return builder;
+    }
+
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Exams.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Exams.java
@@ -40,17 +40,7 @@ public final class Exams {
                 .opportunity(getNullable(row, row.getLong("opportunity")))
                 .completenessId(row.getInt("completeness_id"))
                 .administrativeConditionId(row.getInt("administration_condition_id"))
-                .scaleScore(scaleScore)
-                .studentContext(StudentContext.builder()
-                        .schoolYear(row.getInt("school_year"))
-                        .gradeId(row.getInt("grade_id"))
-                        .iep(row.getBoolean("iep"))
-                        .lep(row.getBoolean("lep"))
-                        .section504(row.getBoolean("section504"))
-                        .economicDisadvantage(row.getBoolean("economic_disadvantage"))
-                        .migrantStatus(getNullable(row, row.getBoolean("migrant_status")))
-                        .build()
-                );
+                .scaleScore(scaleScore);
 
         // Collects claim scores for ICA and Summative exams.
         // If a single claim is unscored a null entry will be put in the list to represent the unscored claim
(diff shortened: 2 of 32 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
```diff
@@ -0,0 +1,57 @@
...
+                .cutPoints(ImmutableList.of(
+                        row.getInt("min_score"),
+                        row.getInt("cut_point_1"),
+                        row.getInt("cut_point_2"),
+                        row.getInt("cut_point_3"),
+                        row.getInt("max_score")
+                ));
+
+        // IABs do not have claims
+        if (type != AssessmentType.IAB) {
```
esotrope: I'm leaning toward the nullable field/inheritance approach for the models because i would hate to have this complicated DB row mapping logic duplicated throughout the app

webapp/src/main/resources/application.sql.yml
```diff
@@ -1,4 +1,46 @@
...
+        e.claim2_scale_score_std_err,
+        a.claim3_score_code,
+        e.claim3_category,
+        e.claim3_scale_score,
+        e.claim3_scale_score_std_err,
+        a.claim4_score_code,
+        e.claim4_category,
+        e.claim4_scale_score,
+        e.claim4_scale_score_std_err,
+        e.asmt_id
```
esotrope: the only field difference for exam lookup by student

webapp/src/test/java/org/opentestsystem/rdw/reporting/TestData.java
```diff
@@ -0,0 +1,219 @@
...
+import org.opentestsystem.rdw.reporting.exam.StudentContext;
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.ZoneOffset;
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+public final class TestData {
```
esotrope: Created a test data holder to reduce repeating the creation of these complex objects.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentRepository.java
```diff
@@ -0,0 +1,61 @@
...
+                                (ethnicityRow, ethnicityIndex) -> ethnicityRow.getString("code")
+                        );
+
+                        return Student.builder()
+                                .id(studentId)
+                                .ssid(ssid)
+                                .firstName(row.getString("first_name"))
+                                .lastName(row.getString("last_or_surname"))
+                                .genderId(row.getLong("gender_id"))
+                                .ethnicityCodes(ImmutableSet.copyOf(ethnicityCodes))
```
esotrope: @jtreuting, @bgsavage do we actually need gender ID and ethnicity codes for the student exam history view? I imagine filtering on those things isn't useful because they are constant for the same student.

jtreuting: Nope.  The filters on that page will only be test related and not student related filters.  They are listed on the page spec: https://confluence.fairwaytech.com/display/SBACDW/2.0%3A+Student+History+Test+Results

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Exams.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static org.opentestsystem.rdw.reporting.jdbc.QueryUtils.getNullable;
+
+public final class Exams {
+
+    private static final ClaimScaleScoreColumn[] CLAIM_SCALE_SCORE_COLUMNS = {
+            new ClaimScaleScoreColumn("claim1_score_code", "claim1_category", "claim1_scale_score", "claim1_scale_score_std_err"),
```
bgsavage: Can this be generated based on some max value and string formatting? 

For example, you could have a string const of just "claim{0}_score_code" and format indexes in there to create your collection of ClaimScalScoreColumns.  I don't know if it's any simpler, just thinking how to avoid one of these string values from accidentally changing.

esotrope: Yeah we can do that - as long as its not done at method execution time like i had it i think people would be fine with that.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Exams.java
```diff
@@ -0,0 +1,101 @@
...
+
+        // Collects claim scores for ICA and Summative exams.
+        // If a single claim is unscored a null entry will be put in the list to represent the unscored claim
+        if (AssessmentType.valueOf(row.getInt("type_id")) != AssessmentType.IAB) {
+            final List<ScaleScore> claimScaleScores = newArrayList();
+            for (ClaimScaleScoreColumn column : CLAIM_SCALE_SCORE_COLUMNS) {
+
+                final String claimCode = getNullable(row, row.getString(column.code));
+                if (claimCode == null) {
+                    break;
```
bgsavage: Should we break here or continue?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentRepository.java
esotrope: Thanks

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Exams.java
esotrope: break - in reality assessments can have 0 (iab), 3 or 4 claims (ica/sum)
even if they had 2 they would be the first two and not spread out.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentRepository.java
bgsavage: I thought I saw a demographic view where we display this info, but probably not in the student exam history view.  There's no reason to have student data for filtering here because it's just a single student view, right?

esotrope: Yeah, according to the spec we wont need that info for this view. Also seems like we don't need the StudentContext populated on the exams

webapp/src/main/resources/application.sql.yml
bgsavage: Does this warrant a whole new query / repo get method?

webapp/src/test/java/org/opentestsystem/rdw/reporting/TestData.java
bgsavage: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Exams.java
bgsavage: I was just thinking in terms of defensive coding, whereas a value was populated null when we weren't expecting it to, should the other values still try to populate?

esotrope: if the asmt claim code is null then it would be invalid to even have not null claim scores in the corresponding claim columns

bgsavage: Can an assessment be partially scored?

jtreuting: depend on what you mean.  If there is a scale score there will be claim scores and if there are any claim scores there will be all claim scores.  There is a partial concept but it is related to the student didn't complete the entire test but it was scored anyways.

PR #159, opened 2017-06-22
changes PII_READ to INDIVIDUAL_PII_READ 

esotrope:
Permission server had role PII mapping to permission INDIVIDUAL_PII_READ instead of PII_READ

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/group/GroupService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/group/GroupService.java
@@ -17,7 +17,7 @@ public interface GroupService {
      * @param user the {@link User}
      * @return the {@link Group}s of the given {@link User}
      */
-    @PreAuthorize("hasAnyAuthority('PERM_PII_READ', 'PERM_GROUP_PII_READ')")
+    @PreAuthorize("hasAuthority('PERM_GROUP_PII_READ')")
     Set<Group> getGroups(@NotNull User user);
 
 }
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/school/SchoolService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/school/SchoolService.java
@@ -19,7 +19,6 @@ interface SchoolService {
      * @param text the text to match the school name against
      * @return all the schools in the given district which the given user has access to
      */
-    @PreAuthorize("hasAuthority('PERM_PII_READ')")
     Set<Organization> getSchoolsWithNameMatching(@NotNull User user, String text);
 
 }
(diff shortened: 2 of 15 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/group/GroupService.java
```diff
@@ -17,7 +17,7 @@
      * @param user the {@link User}
      * @return the {@link Group}s of the given {@link User}
      */
-    @PreAuthorize("hasAnyAuthority('PERM_PII_READ', 'PERM_GROUP_PII_READ')")
+    @PreAuthorize("hasAuthority('PERM_GROUP_PII_READ')")
```
esotrope: Role PII has permission GROUP_PII_READ so the hasAnyAuthority is redundant here

bgsavage: I believe they can change that at any time, right?  Jeff shared with me a link where you can change the permissions on the role.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/school/SchoolService.java
```diff
@@ -19,7 +19,6 @@
      * @param text the text to match the school name against
      * @return all the schools in the given district which the given user has access to
      */
-    @PreAuthorize("hasAuthority('PERM_PII_READ')")
```
esotrope: Removed security on school search by name as there is no NPI to protect behind this call

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/group/GroupService.java
esotrope: yeah we will have the ability to change role to permission mappings without changing code.
It's important to note that the app only loads those mappings once at startup though

bgsavage: Yeah so my point here is that the role has Group_PII_Read today, so it is redundant, but if the role changed to not have Group_PII_Read, does this produce an unexpected result for us?

jtreuting: @esotrope Are you saying that someone with a PII role should have the permission `GROUP_PII_READ` since we are doing permission based authorization?  So both the role PII and PII_GROUP would have that same permission?

esotrope: correct

chdietz: I think that makes sense.  For some reason if it was decided PII role did not warrant what permission GROUP_PII_READ allowed the role permission mapping change would allow that.

PR #160, opened 2017-06-22
adds permission collision workaround

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -57,7 +58,14 @@ public DefaultAuthorizationService(
...
             }
         }
-        return ImmutableSet.copyOf(grants);
+        return ImmutableSet.copyOf(sanitizeGrants(grants));
+    }
+
+    // TODO decide strategy for resolving case where users are assigned roles with colliding permission sets
+    private Collection<Grant> sanitizeGrants(final Collection<Grant> grants) {
+        return grants.stream().anyMatch(grant -> "PII".equals(grant.getRole()))
+                ? grants.stream().filter(grant -> !"PII_GROUP".equals(grant.getRole())).collect(toSet())
```
malaffoon: not quite grokking this but if it allows dwtest@example.com to login locally, then i'm all for it!

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -57,7 +58,14 @@ public DefaultAuthorizationService(
...
+                : grants;
```
esotrope: There is a tenancy chain precedence issue that arose with the following permission role mapping update:
PII -> INDIVIDUAL_PII_READ, GROUP_PII_READ, GROUP_READ
PII_GROUP -> GROUP_PII_READ, GROUP_READ
In this workaround the app ignores the tenancy chain PII_GROUP entries if a PII entry exists.

malaffoon: got it ... perhaps a comment to that affect in the code?

jtreuting: This doesn't feel right to me.  I don't think we should be hard-coding roles into the code because we shouldn't care about roles at all, we should only care about permissions that are derived from whatever roles there are.  Since we don't control roles. they could create a role `BOB` that has `INDIVIDUAL_PII_READ` permissions and never use a role called `PII`

esotrope: Yeah agreed. This is meant to be a hotfix and discussion opener as to how we want to resolve competing permissions.
Users could have 
PII|org set A
PII_GROUP|org set B
both of these roles have a GROUP_PII_READ permission which organizations should we associate with the permission? A, B or the union of A and B?

jtreuting: if we have the simple case of a teacher at a single school .... then no matter how they get a permission, meaning which role gives it to them, they have it so they get a union of all the permissions regardless of the role.

i think it can get more complex to calculate if they have a role at a district level and then different roles at a school level.

PR #162, opened 2017-06-23
Feature/permission scope union

esotrope:
Auth service now resolves the case of different tenancy chain role grants have overlapping permission sets with different access levels associated to them (state, district, school).
The solution implemented makes a union of the organizations associated with the permission. 

Example:

Tenancy chain:
PII|{districtA, schoolA}
PII_GROUP|{schoolB}

With permission mapping:
PII -> INDIVIDUAL_PII_READ, GROUP_PII_READ, GROUP_READ
PII_GROUP -> GROUP_PII_READ, GROUP_READ

will result in permissions:
INDIVIDUAL_PII_READ|{districtA, schoolA}
GROUP_PII_READ|{districtA, schoolA, schoolB}
GROUP_READ|{districtA, schoolA, schoolB}

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
@@ -131,7 +131,7 @@ public class DefaultAuthorizationService implements AuthorizationService {
 
         return permissionScopeBuildersByPermissionId.entrySet().stream()
                 // missing IDs in the natural ID to entity ID map can result in an invalid permission scope builder
-                // that is neither statewide containing any has schools and districts
+                // that is neither statewide nor containing any schools and districts so this will filter those out
                 .filter(entry -> entry.getValue().isValid())
                 .map(entry -> new Permission(entry.getKey(), entry.getValue().build()))
                 .collect(toImmutableSet());
@@ -204,12 +204,13 @@ public class DefaultAuthorizationService implements AuthorizationService {
             return this;
         }
 
+        /**
+         * @return <code>true</code> if the permission scope is statewide or has any districts or schools
+         */
         public boolean isValid() {
-            return statewide || (
-                    (districtIds != null ? districtIds.size() : 0)
-                            + (schoolIds != null ? schoolIds.size() : 0)
-                            > 0
-            );
+            return statewide
+                    || (districtIds != null && !districtIds.isEmpty())
+                    || (schoolIds != null && !schoolIds.isEmpty());
         }
 
         public PermissionScope build() {
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 
+import static com.google.common.collect.Maps.newHashMap;
 import static com.google.common.collect.Sets.newLinkedHashSet;
 import static java.util.stream.Collectors.toSet;
 
@@ -58,14 +59,7 @@ public class DefaultAuthorizationService implements AuthorizationService {
                 grants.add(grant);
             }
         }
-        return ImmutableSet.copyOf(sanitizeGrants(grants));
-    }
-
-    // TODO decide strategy for resolving case where users are assigned roles with colliding permission sets
-    private Collection<Grant> sanitizeGrants(final Collection<Grant> grants) {
-        return grants.stream().anyMatch(grant -> "PII".equals(grant.getRole()))
-                ? grants.stream().filter(grant -> !"PII_GROUP".equals(grant.getRole())).collect(toSet())
-                : grants;
+        return ImmutableSet.copyOf(grants);
     }
 
     public Set<Permission> getPermissions(final Collection<? extends Grant> grants) {
@@ -89,7 +83,7 @@ public class DefaultAuthorizationService implements AuthorizationService {
         final Map<String, Long> districtIds = districtRepository.findAllIds(districtNaturalIds);
         final Map<String, Long> institutionIds = institutionRepository.findAllIds(institutionNaturalIds);
 
-        final Set<Permission> permissions = newLinkedHashSet();
+        final Map<String, Permission> permissionsById = newHashMap();
 
         // Identify and store the permissions associated with the role
         for (String role : grantsByRole.keySet()) {
@@ -99,16 +93,21 @@ public class DefaultAuthorizationService implements AuthorizationService {
             // User may have roles not registered to this application in the permission service.
             if (permissionIds != null) {
                 for (String permissionId : permissionIds) {
-                    permissions.add(
-                            new Permission(
-                                    permissionId,
-                                    createScope(grantsByRole.get(role), districtIds, institutionIds)
-                            )
+                    final Permission existingPermission = permissionsById.get(permissionId);
+                    final Permission permission = new Permission(
+                            permissionId,
+                            createScope(grantsByRole.get(role), districtIds, institutionIds)
                     );
+                    if (existingPermission == null) {
+                        permissionsById.put(permissionId, permission);
+                    } else {
+                        // TODO document
+                        permissionsById.put(permissionId, union(existingPermission, permission));
+                    }
                 }
             }
         }
-        return ImmutableSet.copyOf(permissions);
+        return ImmutableSet.copyOf(permissionsById.values());
     }
 
     public Set<GrantedAuthority> getAuthorities(final Collection<? extends Grant> grants) {
@@ -196,4 +195,27 @@ public class DefaultAuthorizationService implements AuthorizationService {
         );
 
     }
+
+    private Permission union(final Permission a, final Permission b) {
+        if (a.getScope().isStatewide()) {
+            return a;
+        }
+        if (b.getScope().isStatewide()) {
+            return b;
+        }
+        return new Permission(
+                a.getId(),
+                new PermissionScope(
+                        ImmutableSet.<Long>builder()
+                                .addAll(a.getScope().getDistrictIds())
+                                .addAll(b.getScope().getDistrictIds())
+                                .build(),
+                        ImmutableSet.<Long>builder()
+                                .addAll(a.getScope().getInstitutionIds())
+                                .addAll(b.getScope().getInstitutionIds())
+                                .build()
+                )
+        );
+    }
+
 }
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
@@ -59,26 +59,33 @@ public class DefaultAuthorizationService implements AuthorizationService {
 
     @Override
     public Set<Grant> getGrants(final String[] encodedGrants) {
-        final Set<Grant> grants = newLinkedHashSet();
+        final ImmutableSet.Builder<Grant> grants = ImmutableSet.builder();
         for (String encodedGrant : encodedGrants) {
             final Grant grant = Grants.from(encodedGrant);
 
-            // Do not add do not belong to the application instance's state
+            // Only add grants for the application's configured state
             if (stateId.equals(grant.getStateId())) {
                 grants.add(grant);
             }
         }
-        return ImmutableSet.copyOf(grants);
+        return grants.build();
     }
 
+    // A Grant is a role granted for a state or organization identified by its natural id.
+    // A Permission is a permission and collection of organizations identified by their internal ids.
+    // This method maps the grant roles to the associated permissions, collects and dedups the
+    // organizations that share that permission, and resolves natural ids to internal ids.
     @Override
-    public Set<Permission> getPermissions(final Collection<? extends Grant> grants) {
+    public Set<Permission> getPermissions(final Collection<Grant> grants) {
 
         // Groups grants by role to later combine their districts and institutions into a single scope
         // This is to cover the case of multiple non-intersecting grants being given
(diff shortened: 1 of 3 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -146,54 +174,51 @@ public DefaultAuthorizationService(
...
+        public PermissionScope build() {
+            if (statewide) {
+                return PermissionScope.STATEWIDE;
+            }
+            return new PermissionScope(districtIds, schoolIds);
+        }
 
     }
+
 }
```
esotrope: This diff is pretty useless. May help to view the resulting file: https://github.com/SmarterApp/RDW_Reporting/blob/4389b4336955bd44696c807ffd93ccb476be12f1/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java

These are the changes:
1. Removed hotfix to ignore PII_GROUP role if user has PII role
2. Removed createPermissionScope methods -> no longer used.
3. Added private PermissionScopeBuilder class to manage proper construction and union of permissions

webapp/src/test/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationServiceTest.java
```diff
@@ -79,11 +97,68 @@ public void getPermissionsShouldCorrectlyDerivePermissionsFromGrants() throws Ex
...
+
+        final List<Grant> grants = ImmutableList.of(
+                Grant.builder().entityId("a").role("a").entityLevel(STATE).stateId("a").build(),
+                Grant.builder().entityId("a").role("a").entityLevel(DISTRICT).stateId("a").districtId("a").build()
+        );
+
+        assertThat(authorizationService.getPermissions(grants))
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyInAnyOrder(new Permission("a", PermissionScope.STATEWIDE));
+    }
```
esotrope: Added the above 4 unit test cases to cover overlapping permissions and the union of their organizations

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -76,46 +80,69 @@ public DefaultAuthorizationService(
...
+                            permissionScopeBuilder.addDistrictId(id);
+                        } else {
+                            logger.warn("grant district with natural ID \"{}\" not found.", grant.getEntityId());
+                        }
+                    } else if (grant.getEntityLevel() == INSTITUTION) {
+                        final Long id = schoolIdsByNaturalId.get(grant.getEntityId());
+                        if (id != null) {
+                            permissionScopeBuilder.addSchoolId(id);
+                        } else {
+                            logger.warn("grant district with natural ID \"{}\" not found", grant.getEntityId());
```
jtreuting: should say "grant school with ..."

esotrope: thanks - patching

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -19,15 +21,22 @@
...
 import static com.google.common.collect.Sets.newLinkedHashSet;
-import static java.util.stream.Collectors.toSet;
+import static org.opentestsystem.rdw.security.Grant.EntityLevel.DISTRICT;
+import static org.opentestsystem.rdw.security.Grant.EntityLevel.INSTITUTION;
+import static org.opentestsystem.rdw.security.Grant.EntityLevel.STATE;
 
 @Service
 public class DefaultAuthorizationService implements AuthorizationService {
 
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
malaffoon: We generally use static loggers, i.e.
```
    private static final Logger logger = LoggerFactory.getLogger(DefaultAuthorizationService.class); 
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -58,16 +68,10 @@ public DefaultAuthorizationService(
                 grants.add(grant);
```
malaffoon: there's a comment a couple lines above this (not showing in diff) that is confusing (just grammatically), `// Do not add do not belong to the application instance's state`. Could you tweak that to be something like `// Only add grants for the application's configured state`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -58,16 +68,10 @@ public DefaultAuthorizationService(
...
         }
-        return ImmutableSet.copyOf(sanitizeGrants(grants));
-    }
-
-    // TODO decide strategy for resolving case where users are assigned roles with colliding permission sets
-    private Collection<Grant> sanitizeGrants(final Collection<Grant> grants) {
-        return grants.stream().anyMatch(grant -> "PII".equals(grant.getRole()))
-                ? grants.stream().filter(grant -> !"PII_GROUP".equals(grant.getRole())).collect(toSet())
-                : grants;
+        return ImmutableSet.copyOf(grants);
```
malaffoon: I still don't understand why it is making an immutable copy of the set that was just created. But, if it really wants to return an immutable set why not use ImmutableSet.builder() instead of doing the copyOf?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -58,16 +68,10 @@ public DefaultAuthorizationService(
...
     }
 
+    @Override
     public Set<Permission> getPermissions(final Collection<? extends Grant> grants) {
```
malaffoon: Why is this a collection of `<? extends Grant>` instead of just `<Grant>`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -76,46 +80,69 @@ public DefaultAuthorizationService(
...
-
+        // prepare natural ID to entity ID maps
         final Set<String> districtNaturalIds = newLinkedHashSet();
         final Set<String> institutionNaturalIds = newLinkedHashSet();
         for (Grant grant : grants) {
-            if (grant.getEntityLevel() == Grant.EntityLevel.DISTRICT) {
+            if (grant.getEntityLevel() == DISTRICT) {
                 districtNaturalIds.add(grant.getEntityId());
-            } else {
+            } else if (grant.getEntityLevel() == INSTITUTION) {
```
malaffoon: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -76,46 +80,69 @@ public DefaultAuthorizationService(
...
-        for (String role : grantsByRole.keySet()) {
+        final Map<String, PermissionScopeBuilder> permissionScopeBuildersByPermissionId = newHashMap();
+        for (Map.Entry<String, Grant> entry : grantsByRole.entries()) {
 
+            final String role = entry.getKey();
+            final Grant grant = entry.getValue();
             final Collection<String> permissionIds = permissionsByRole.get(role);
 
             // User may have roles not registered to this application in the permission service.
             if (permissionIds != null) {
```
malaffoon: perhaps flip the logic to avoid extra indent:
```
            // Guard against roles that grant permissions not registered to this application in the permission service
            if (permissionIds == null) continue;

File: webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java [hunk 10]
```diff
@@ -76,46 +80,69 @@ public DefaultAuthorizationService(
...
-            } else {
+            } else if (grant.getEntityLevel() == INSTITUTION) {
                 institutionNaturalIds.add(grant.getEntityId());
             }
         }
-        final Map<String, Long> districtIds = districtRepository.findAllIds(districtNaturalIds);
-        final Map<String, Long> institutionIds = institutionRepository.findAllIds(institutionNaturalIds);
-
-        final Set<Permission> permissions = newLinkedHashSet();
+        final Map<String, Long> districtIdsByNaturalId = districtRepository.findAllIds(districtNaturalIds);
```
malaffoon: Minor comment on the algorithm: if you deferred the lookup of the natural ids you might not need them at all. Since statewide trumps institution grants, you only end up using the ids if it isn't a statewide grant. What i'm thinking is you could store the natural ids until you get through the inversion of the roles to permissions and resolve them all at the end.

Practically, it isn't going to be that much of an issue except for crazy test accounts. Just thinking out loud.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -58,16 +68,10 @@ public DefaultAuthorizationService(
...
-    // TODO decide strategy for resolving case where users are assigned roles with colliding permission sets
-    private Collection<Grant> sanitizeGrants(final Collection<Grant> grants) {
-        return grants.stream().anyMatch(grant -> "PII".equals(grant.getRole()))
-                ? grants.stream().filter(grant -> !"PII_GROUP".equals(grant.getRole())).collect(toSet())
-                : grants;
+        return ImmutableSet.copyOf(grants);
     }
 
+    @Override
     public Set<Permission> getPermissions(final Collection<? extends Grant> grants) {
```
malaffoon: I think an additional code comment might be warranted for this method, given its complexity. I'm having trouble coming up with good verbiage but something like:
```
         // A Grant is a role granted for an institution identified by its natural id.
        // A Permission is a permission and list of institutions identified by their internal ids.
        // This method maps the grant roles to the associated permissions, collects and dedups the
        // institutions that share that permission, and resolves natural ids to internal ids. 
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -76,46 +80,69 @@ public DefaultAuthorizationService(
...
-                            new Permission(
-                                    permissionId,
-                                    createScope(grantsByRole.get(role), districtIds, institutionIds)
-                            )
-                    );
+
+                    final PermissionScopeBuilder permissionScopeBuilder = permissionScopeBuildersByPermissionId
+                            .computeIfAbsent(permissionId, (key) -> new PermissionScopeBuilder());
+
+                    if (grant.getEntityLevel() == STATE) {
```
malaffoon: I know it's only three but this is kind of screaming `switch`

bgsavage: Typically I think it's cleaner if the cyclomatic complexity for a method is less than 10.  So it might be cleaner to extract each block inside each of these if blocks into their own methods.  That way at a glance I can understand what this code is doing, and if I want to know the "how" I can dive into deeper detail by looking inside each method.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
esotrope: Is there a reason to when its in a singleton?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
esotrope: K - I'll reword it

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
jtreuting: I can honestly say this is the first review comment i've read that throws out "cyclomatic complexity".

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
esotrope: *will change

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
esotrope: Good point - there is definitely room for performance optimization here. I take a look and see what would be involved in doing that

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
esotrope: that would end up looking like this:
```java
addSchool(permissionScopeBuilder, grant.getEntityId(), schoolIdsByNaturalId);
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
malaffoon: If it is a non-inheritable singleton then, no, there is no technical reason. However, i invoke consistency as a good reason: every non-trivial class tends to have a logger and if they are declared the same way in the same position it is just easier and more consistent.

PR #163, opened 2017-06-23
Updated readme.

agorina:


README.md
```diff
@@ -43,8 +43,7 @@ The service depends on the database being configured properly. This is done usin
 git clone https://github.com/SmarterApp/RDW_Schema
 cd RDW_Schema
 git checkout develop
-cd reporting
-../scripts/migrate
+./gradlew install
```
agorina: @jtreuting noticed that it was outdated. I think this is correct, but not sure if there is a better way?

malaffoon: You don't need to check out schema at all anymore, right? So this should probably be:
```
The service depends on the database being configured properly. This is done using RDW_Schema which
is a dependency in this project and has gradle tasks. 
```bash
./gradlew migrateReportingProd
```
Usually a migrate is all that is needed but you may also clean to wipe all data:
```bash
./gradlew cleanReportingProd migrateReportingProd
```
```

agorina: Thank you. Will update, here and in the ingest project too.

malaffoon: Thanks.

esotrope: btw does *Prod run the *Test tasks too? or do we need to be running those in addition?

malaffoon: *Prod does not run *Test that i can see. So, they need to be run separately. I think invoking `gradle IT` does run the *Test so you'd only need to do it if you are running ITs from within IntelliJ.

PR #165, opened 2017-06-26
Changes to the SQL to support DWR-427:Student Group data warehouse mo…

agorina:
…del changes

### DIFF
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -355,8 +355,9 @@ sql:
         ssid,
         first_name,
         last_or_surname
-      from student
-      where ssid=:ssid
+      from student s
+      where s.ssid=:ssid
+       and exists (select 1 from exam e where e.student_id = s.id)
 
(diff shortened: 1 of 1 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -355,8 +355,9 @@ sql:
         ssid,
         first_name,
         last_or_surname
-      from student
-      where ssid=:ssid
+      from student s
+      where s.ssid=:ssid
+       and exists (select 1 from exam e where e.student_id = s.id)
```
esotrope: I'm confused by this - won't the student record not even exist unless there is an exam the student took in the data mart? I thought the student table was derived from the exam table.

agorina: As part of supporting a student group upload from CSV, we are allowing students to be created from the CSV with just SSID and no exams.

esotrope: oh ok. thanks.

PR #166, opened 2017-06-26
Feature/user context

esotrope:
Adds schools and groups to /api/user endpoint

TODOs:
1. Remove deprecated classes and methods once angular client stops relying on them


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/DefaultUserContextService.java
```diff
@@ -0,0 +1,50 @@
...
+                .firstName(user.getFirstName())
+                .lastName(user.getLastName())
+                .permissions(user.getPermissionsById().keySet())
+                .schools(getSchools(user))
+                .groups(getGroups(user))
+                .build();
+    }
+
+    private Set<Organization> getSchools(final @NotNull User user) {
+        if (user.getPermissionsById().containsKey("INDIVIDUAL_PII_READ")) {
```
esotrope: Not sure this was the best way of handling this but it works.
Basically, it doesn't attempt to get the schools/groups of a user when it knows they are not authorized.
Another option for this could be to have the school/group services return empty sets when the user is not entitled.
Yet another option was to use try/catch here.
Not sure what's preferred here.

malaffoon: 1. Service returning an empty set makes sense to me, "here's the list of schools you know about ... and, yeah, it's empty." This would be my favored approach.
2. Leave it like it is with a comment something like `// check permissions first to avoid auth exception`
3. Why is getting the list of schools even gated by PII permissions? Schools have nothing to do with PII. If we want to do something like this, i think we should consider separating PII (information about students) from other "sensitive" data.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/UserContext.java
```diff
@@ -0,0 +1,104 @@
...
+
+import com.google.common.collect.ImmutableSet;
+import org.opentestsystem.rdw.reporting.exam.Organization;
+import org.opentestsystem.rdw.reporting.search.group.Group;
+
+import java.util.Set;
+
+class UserContext {
+
+    private String firstName;
```
malaffoon: Usual warning about first/last name stuff being a western thing. The best outcome i've experienced in previous jobs is to have a `displayName` field that the user is allowed to set. The next best was `familyName` + `givenName` where the UI ... you know what, that didn't turn out that great.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/group/DefaultGroupServiceTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Before
+    public void before() {
+        this.service = new DefaultGroupService(repository);
+    }
+
+    @Test
+    public void itShouldForwardCorrectUsername() throws Exception {
+        final List<Group> groups = newArrayList(Group.builder().id(10L).build());
+        when(user.getUsername()).thenReturn("user");
```
malaffoon: minor nit (especially since this is such a small test suite): for user i would either make it a real user object or do the mock behavior setup in the @Before method.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/UserContext.java
esotrope: The problem with just having one field is that there is context-dependent formatting.
Sometimes we might want to display FIRST LAST
other times LAST, FIRST

webapp/src/test/resources/org/opentestsystem/rdw/reporting/search/school/schools.sql
```diff
@@ -0,0 +1,9 @@
+insert into district(id, natural_id, name) values
```
malaffoon: Just arguing with myself ...

Usually test resources are not buried in the full package folder structure. 
But if the resource is need only for testing classes in that package ... maybe.
But then the resource can't be reused.
I don't think the resources are actually scoped so inadvertent name collision may occur.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/DefaultUserContextService.java
esotrope: For (3) you are right but on the other hand we can save a DB call if we know the only use for the schools is school grade based exam search (a PII only feature)

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/UserContext.java
malaffoon: For our users we want to do that?

webapp/src/test/resources/org/opentestsystem/rdw/reporting/search/school/schools.sql
esotrope: this way the resources are package scoped and collision can only occur if you use classpath:*resource or classpath:**resource I think

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/DefaultUserContextService.java
malaffoon: I'd vote for the extra DB call with a more correct permission. Could create a JIRA to postpone the work.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/UserContext.java
esotrope: Yeah we display LAST, FIRST at the top of our page

bgsavage: I'm usually not a fan of "display" fields on an API, because it makes assumptions on how the UI will want to display it.  I prefer to have to keep the separate fields (first / last) even if it does have a western connotation.

PR #167, opened 2017-06-26
Updated homepage

jtreuting:
- use translate instead of static text
- minor structure updates based on latest mockup

Please check the en.json hierarchy to see if it fits with what you had in mind as far as structure.  I went back and forth on whether the search widget content should be under homepage as a section or their own like groups is.

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/groups.component.html
+++ b/webapp/src/main/webapp/src/app/groups/groups.component.html
@@ -2,6 +2,7 @@
 
   <div class="flex-child grow">
     <label>{{'labels.groups.search-label' | translate }}</label>
+  </div>
 
   <!-- My Groups Search Input -->
   <div *ngIf="groups.length > 2" class="gray-lightest pt-sm pr-sm pb-sm pl-sm my-groups-search-form-container">
--- a/webapp/src/main/webapp/src/app/groups/groups.component.html
+++ b/webapp/src/main/webapp/src/app/groups/groups.component.html
@@ -1,5 +1,8 @@
 <div *ngIf="groups" class="home-view">
 
+  <div class="flex-child grow">
+    <label>{{'labels.groups.search-label' | translate }}</label>
+
   <!-- My Groups Search Input -->
   <div *ngIf="groups.length > 2" class="gray-lightest pt-sm pr-sm pb-sm pl-sm my-groups-search-form-container">
     <label for="search" class="sr-only">Search My Groups</label>
--- a/webapp/src/main/webapp/src/app/home/home.component.html
+++ b/webapp/src/main/webapp/src/app/home/home.component.html
@@ -4,20 +4,20 @@
 
     <!-- Left Column Content Area -->
     <div class="col-md-9">
-      <h1 class="sr-only">Reporting System Dashboard</h1>
+      <h1 class="sr-only">{{'labels.homepage.title' | translate}}</h1>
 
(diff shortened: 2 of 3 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -11,10 +11,29 @@
...
   "labels": {
+    "homepage": {
+      "title": "Access Assessment Results",
+      "search-student-id": {
+        "label": "Search by Student",
+        "placeholder": "Enter the Statewide Student Identifier (SSID)"
+      },
+      "search-school-grade": {
+        "school-label": "Search by School Name",
+        "school-placeholder": "Start typing the school name",
```
bgsavage: This is a combo box now so I'm not sure this is still accurate.  You have to click to open the dropdown and then you can type in the search field...

webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -4,38 +4,37 @@
 
     <!-- Left Column Content Area -->
     <div class="col-md-9">
-      <h1 class="sr-only">Reporting System Dashboard</h1>
+      <h1 class="sr-only">{{'labels.homepage.title' | translate}}</h1>
 
       <!-- Student ID Search -->
-      <h2 class="blue-dark h3 mb-md">Search by Student ID</h2>
+      <h1 class="blue-dark h3 mb-md">{{'labels.homepage.title' | translate}}</h1>
```
bgsavage: I think Aaron wants this to be an `h2` tag.

webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -1,5 +1,8 @@
 <div *ngIf="groups" class="home-view">
 
+  <div class="flex-child grow">
+    <label>{{'labels.groups.search-label' | translate }}</label>
+
```
bgsavage: Does this have a closing `</div>`  ?  I might be missing it.

webapp/src/main/webapp/src/app/home/home.component.html
jtreuting: The newly updated mockup has a main header and then the sub groups: https://xd.adobe.com/view/ad1f1339-5536-4f05-91b8-1a020a564bcf/ so figured that was an H1 since it was blue and seemed bigger?

webapp/src/main/webapp/src/app/groups/groups.component.html
jtreuting: it will very shortly :) oops

webapp/src/main/webapp/src/app/home/home.component.html
bgsavage: Maybe change the css class from h3 to h1 then?  I think this might be a best practice of having h1 / h2 / h3 / h4 / etc read in order down a page, though I'm not sure if there's place for that in angular since this isn't a static web page.

jtreuting: yep, changing class from h3 to h1, makes it looks more appropriate

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: We can change that once it's implemented I think to somehting that makes more sense.  What i see is not a combo box yet

PR #169, opened 2017-06-27
DWR-466 Remove invalid prompts from group results page when viewing "…

wtritch:
…Most recent" assessment

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/results/group-results.component.html
+++ b/webapp/src/main/webapp/src/app/groups/results/group-results.component.html
@@ -91,13 +91,13 @@
         </button>
       </div>
     </h2>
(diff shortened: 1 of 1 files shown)


wtritch:
This PR hides the "Click an item to remove it from the list" prompt text and "X" icon on assessments when viewing the most recent.

esotrope:
Is this the requirement? I could also imagine them maybe wanting to allow removing it and then auto switch the user to the custom tab if they change anything.

wtritch:
This was the suggested fix per the jira, and it's logically complete.  I could see arguments either way, but this seems small and changeable enough that I didn't want to delay development to hold a meeting over it.

esotrope:
Sounds good 👍  Just wanted to make sure it wasn't just what we thought was best but was also what sbac wanted.

PR #170, opened 2017-06-27
Updated front end to now use the codes instead of ids for administrat…

bgsavage:
…ive condition, completeness, and gender.  Updated filters and unit tests as well.  Now looping on a predetermined set of filters instead of trying to dynamically look at all the properties in the class.


webapp/src/main/webapp/src/app/groups/results/exam-filters/exam-filter.service.spec.ts
```diff
@@ -26,77 +26,77 @@ describe('ExamFilterService', () => {
...
   it('should not filter exams by summative status for IABs', () => {
-    filterBy.summativeStatus = 2;
+    filterBy.summativeStatus = 'Valid';
     assessmentExam.assessment.type = AssessmentType.IAB;
 
-    assessmentExam.exams[ 0 ].administrativeCondition = 2;
-    assessmentExam.exams[ 1 ].administrativeCondition = 3;
-    assessmentExam.exams[ 2 ].administrativeCondition = 2;
-    assessmentExam.exams[ 3 ].administrativeCondition = 3;
+    assessmentExam.exams[ 0 ].administrativeCondition = 'Valid';
```
esotrope: Is it possible to make enum for completeness and administration condition? I think that would help clients change everything in one place later. what do you think?

bgsavage: Might be a good idea.  Btw, do you know why 'Valid' is in a different format than the other 'SD', 'NS', 'IN' ?  Seems odd.

malaffoon: http://www.smarterapp.org/documents/TestResults-DataModel.pdf, top of page 10.

Looks like it might've come about because 'Valid' is like blank.

PR #171, opened 2017-06-27
Feature/find by school grade

bgsavage:
This PR is to complete story DWR 304 - school -grade widget on the home page.

image

This is using a primeng filterable dropdown and a model driven form to handle required fields and checking if the form is valid before allowing submit.  

There was also some code clean up and and improved logic for grabbing the userContext to prevent multiple calls to the api/users route when different parts of the app are racing to grab the same data.

This is still pending CSS updates by @jessicacolton to restore table CSS after primeng overrides.


webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -14,10 +14,10 @@ export class AppComponent {
     return this._user;
   }
 
-  constructor(public translate: TranslateService, private _userService : UserService) {
+  constructor(public translate: TranslateService, private _userService: UserService) {
 
-    let languages = ['en', 'ja'];
-    let defaultLanguage = languages[0];
+    let languages = [ 'en', 'ja' ];
+    let defaultLanguage = languages[ 0 ];
```
esotrope: not a fan of this part of the ootb intellij formatting...

webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -25,39 +25,18 @@ <h1 class="blue-dark h1 mb-md">{{'labels.homepage.title' | translate}}</h1>
...
-        </form>
+        </div>
       </div>
 
       <!-- My Groups Search -->
-      <div class="my-groups-search-container well">
-        <div *sbAuthorize="['GROUP_PII_READ','INDIVIDUAL_PII_READ']">
-          <!--TODO:  Implement authorize of PII or PII Group role. -->
-          <groups></groups>
+      <div *sbAuthorize="['GROUP_PII_READ','INDIVIDUAL_PII_READ']">
```
esotrope: This can be reduced to:
```html
*sbAuthorize="['GROUP_PII_READ']
```

webapp/src/main/webapp/src/app/shared/cachingData.service.js
```diff
@@ -0,0 +1,37 @@
+"use strict";
+var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
```
esotrope: This deserves some comments.
Why are we decorating and extending prototypes directly? I thought typescript had extension built in.

webapp/src/main/webapp/src/app/user/authorize.directive.ts
```diff
@@ -1,13 +1,13 @@
...
 
-/*
+/**
   Structural directive that removes element from the DOM should
   the user not have at least one of the specified permissions.
 
   @Input: an array of permissions in which a user must have at least one of to view the content.
 
-  Example usage: <div *sbAuthorize="['GROUP_UPDATE', 'GROUP_UPDATE']"></div>
+  Example usage: <div *sbAuthorize="['GROUP_READ', 'GROUP_UPDATE']"></div>
```
esotrope: may want to change GROUP_UPDATE to GROUP_WRITE -> this will be the actual permission

webapp/src/main/webapp/src/app/user/user.mapper.ts
```diff
@@ -14,6 +19,59 @@ export class UserMapper {
...
+  private mapSchoolsFromApi(schools: any[]): School[] {
+    return isNullOrUndefined(schools)
+      ? []
+      : schools
+        .filter(school => this.isSchoolValid(school))
+        .map(school => this.mapSchoolFromApi(school))
+        .sort(ordering(byString).on<School>(school => school.name).compare);
+  }
+
+  private isSchoolValid(school: any) {
```
esotrope: What is the production use of this?
The DB columns are not nullable.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -29,26 +31,109 @@
...
+        "name": "Third grade",
+        "short": "3"
+      },
+      "04": {
+        "name": "Fourth grade",
+        "short": "4"
+      },
+      "05": {
+        "name": "Fifth grade",
+        "short": "5"
```
esotrope: This is technically the "code" so I would call it that here.
Otherwise I would use "shortName"

webapp/src/main/webapp/src/app/shared/grade-display.pipe.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  constructor(private translate: TranslateService) {}
+
+  /**
+   * Method to transform a grade code to a desired display value.
+   * @param gradeCode
+   * @param format
+   * @returns transformed grade
+   */
+  transform(gradeCode: number, format: string): Observable<string> {
```
esotrope: Not sure I'm a fan of this. It would entail making many specialized pipes for certain enum lookups.
I think what you could do instead is make a general one that resolves dynamic keys with a map ->
```html
'code.with.${vars}.in.it' | translate2:{vars: 1}:{normalTranslateParams}
```

webapp/src/main/webapp/src/app/home/home.component.html
wtritch: What's the difference between GROUP_PII_READ and INDIVIDUAL_PII_READ and which does the user need to be able to search for a student by SSID?

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
```diff
@@ -0,0 +1,30 @@
...
+                    formControlName="school"
+                    [filter]="true"
+                    filterBy="label"
+                    placeholder="{{'prompt.select' | translate}}">
+        </p-dropdown>
+      </div>
+    </div>
+    <div class="flex-child">
+      <label for="grade-select">{{'labels.grade' | translate}}</label>
+      <select name="grade" formControlName="grade" id="grade-select" class="form-control" placeholder="Select">
```
wtritch: NIT Placeholder needs translation

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
```diff
@@ -0,0 +1,30 @@
...
+        <option disabled [ngValue]="selectNullOptionValue">{{'prompt.select' | translate}}</option>
+        <option *ngFor="let grade of availableGrades" [ngValue]="grade">{{grade.code | gradeDisplay:'short'}}</option>
```
wtritch: Recommend changing the default gradeDisplay value to "short" so we don't have to specify it every time.

webapp/src/main/webapp/src/app/home/home.component.html
bgsavage: Great yes I meant to ask about this.

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
```diff
@@ -0,0 +1,30 @@
...
+      </select>
+    </div>
+    <div class="flex-child">
+      <button type="submit" [disabled]="searchForm.invalid || gradesAreUnavailable" class="btn btn-primary mt-md">{{'prompt.search' | translate}}</button>
```
wtritch: If grades are unavailable, wouldn't the searchForm be invalid by default since the grade selection will fail validation?

webapp/src/main/webapp/src/app/shared/cachingData.service.js
bgsavage: This is generated code, I"m not sure how I was able to check this in?  I'll remove it.

esotrope: Oh interesting. yeah it's a .js file

webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    this._availableSchools = schools.map(school => {
+      return {
+        label: school.name,
+        value: school
+      };
+    });
+
+    if(this._availableSchools.length == 1){
+      this.schoolControl.setValue(this._availableSchools[0]);
+    }
```
wtritch: Will this happen after the ngOnInit and trigger the schoolChanged method?

webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.availableGrades = [];
+  }
+
+  ngOnInit() {
+    this.searchForm = new FormGroup({
+      school: new FormControl(null, Validators.required),
+      grade: new FormControl({ value: this.selectNullOptionValue, disabled: true }, Validators.required)
+    });
+
+    this.searchForm.controls[ "school" ].valueChanges.subscribe(school => this.schoolChanged(school));
```
wtritch: NIT Recommend using `this.schoolControl` here for consistency?

webapp/src/main/webapp/src/app/school-grade/school.service.ts
```diff
@@ -0,0 +1,44 @@
...
+   */
+  findGradesWithAssessmentsForSchool(school: School): Observable<Grade[]> {
+    return this.dataService
+      .get(`/schools/${school.id}/assessmentGrades`)
+      .map((apiGrades) => this.mapGradesFromApi(apiGrades));
+  }
+  private mapGradesFromApi(apiGrades: any[]): Grade[] {
+    if (isNullOrUndefined(apiGrades)) return [];
+    return apiGrades
+      .filter(apiGrade => !isNullOrUndefined(apiGrade) && !isNullOrUndefined(apiGrade.code))
```
wtritch: Might want to add a nullOrUndefined check on apiGrade.id as well, since we assume that both `code` and `id` will be present.

webapp/src/main/webapp/src/app/shared/grade-display.pipe.ts
```diff
@@ -0,0 +1,32 @@
+import { Pipe, PipeTransform } from "@angular/core";
+import { TranslateService } from "@ngx-translate/core";
+import { isNullOrUndefined } from "util";
+import { Observable } from "rxjs";
+
+/**
+ * This pipe is responsible for transforming a grade id number into a
```
wtritch: I think we decided to work on codes rather than numbers?

webapp/src/main/webapp/src/app/user/user.mapper.ts
wtritch: It looks like basic data sanitization, making sure we're getting back API Models that will properly map to front-end Models.  ...I'm wondering if we want more of an assertion that fast-fails, or at least logs a validation failure than a check that silently filters out a corrupt value...  What do you guys think?

webapp/src/main/webapp/src/app/user/user.resolve.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from "@angular/core";
+import { Resolve, ActivatedRouteSnapshot, RouterStateSnapshot } from "@angular/router";
+import { Observable } from "rxjs";
+import { UserService } from "./user.service";
+
+@Injectable()
+export class UserResolve implements Resolve<any> {
```
wtritch: NIT `Resolve<User>` or `Observable<User>`?

webapp/src/main/webapp/src/app/user/user.service.ts
```diff
@@ -2,16 +2,31 @@ import { Injectable } from "@angular/core";
...
 
 @Injectable()
 export class UserService {
   constructor(private _mapper: UserMapper, private _dataService: CachingDataService) {
   }
 
+  private currentUser: User;
+  private currentUserObservable: Observable<User>;
+
   getCurrentUser() {
```
wtritch: NIT return Observable\<User> ?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -11,8 +11,10 @@
     "notApplicable": "-",
     "logout": "Logout"
   },
+  "collapse": "Collapse",
+  "select": "Select",
   "labels": {
-    "homepage": {
+      "homepage": {
```
wtritch: nit: extra indent?

webapp/src/main/webapp/src/styles.less
```diff
@@ -109,3 +109,11 @@
   &:first-child { border-top-left-radius: 0px; }
   &:last-child { border-top-right-radius: 0px; }
 }
+
+.ui-dropdown{
+  width: 100% !important;
```
wtritch: !important always makes me cringe... is there *any* way to avoid using it?

webapp/src/main/webapp/src/app/user/user.mapper.ts
esotrope: Yeah logged or hard failure would be preferable. Not sure why we need this kind of data validation though. It makes sense if this is user input data but it's stuff from the backend that is contractually not null and can be constrained by the DB such that null is not every put in the table. Also the data is sanitized and validated during the input/ingest/migrate processes

webapp/src/main/webapp/src/app/shared/grade-display.pipe.ts
wtritch: Hmm... I'm thinking about this.
We could have a standard "default" value in each enum breakdown
```
"grades": {
  "PK": {
    "default": "PK"
    "long": "Prekindergarten"
  },
  ...
},
"anotherEnum": {
  "valA": {
    "default": "ValA"
    "long": "Another Enum: Value A"
  },
  ...
}
...
```
then translate `'PK' | translate2:{type: 'grades', display: 'long'}`
... hmmm, how many enums will we have?  If it's under 10 I think I'd vote for the simple pipe per enum.

esotrope: That's not exactly what I had in mind. This is how i would imagine translating with your examples:
```html
'grades.{grade}.long' | translate2:{grade: grade}
'anotherEnum.{value}.long' | translate2:{value: value}
```
Basically why we are making the pipe to begin with is to simplify the string substitution i think. In angular's HTML it doesn't let us simply use the es6 `${var}` string interpolation which is the pain point I thought.

wtritch: ahhh, I see what you're saying now.  I like that idea.  A simple interpolating translation pipe.

esotrope: yep exactly : j

webapp/src/main/webapp/src/app/user/user.mapper.ts
wtritch: I like to think that every time we change the back-end model, rename a property, remove a property, etc, we will remember where to go change it in the front-end... but I'm sad to say I've seen quite a few bugs crop up where something changes on the back-end, and the front-end doesn't get noticed until something stops working in some deep page.  I fear this may especially be the case since we've decided on a feature-based vertical architecture where a School model may be parsed within a number of verticals.  A little bit of gateway validation can save us a lot of future head banging. =)

webapp/src/main/webapp/src/styles.less
bgsavage: Also makes me cringe as well, and I wouldn't put if I didn't have to...

wtritch: Could we isolate it a little more using sub-selectors rather than apply it to all .ui-dropdown instances?  Or does it need to be applied to *all* .ui-dropdown instances?

bgsavage: I would think all for consistency.

bgsavage: Though this is the only one right now.  If we're able to solve the css issue, I'd like to put it on the group results page as well so you can search by group name there.

webapp/src/main/webapp/src/app/user/authorize.directive.ts
bgsavage: 👍

webapp/src/main/webapp/src/app/user/user.mapper.ts
bgsavage: What's the decision here? Honestly I prefer not to have to validate as it's more code to have to maintain for a "just in case" scenario.

Btw, if we put all api mapping logic in .mapper.ts files then api changes would only need to be updated by doing a CMD + N and search 'mapper' in IntelliJ.

webapp/src/main/webapp/src/assets/i18n/en.json
bgsavage: I can rename to `short-name`

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
bgsavage: Oops this placeholder doesn't even work, Removing now.

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
bgsavage: 👍

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
bgsavage: Yeah I know, it should be part of the form validation.  I was trying to find an easy way to create a custom validator for this but also didn't want to spin on it for too long.

webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
bgsavage: Yes this is after ngOnInit.

webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
bgsavage: Yep good catch, must've been before I created the getter.

webapp/src/main/webapp/src/app/user/user.mapper.ts
wtritch: My vote is validate, but I don't feel super strongly about it.

PR #172, opened 2017-06-27
DWR-489: Info popovers in grid headers

jtreuting:
- Implemented for 3 columns, but commented out for 1 of them for now since it messes the wrapping and a demo is coming up
- trigger of focus doesn't work on a span or an anchor tag so implemented show on click and hide on mouseleave.  don't love it so thoughts appreciated on this


webapp/src/main/webapp/src/app/groups/groups.module.ts
```diff
@@ -32,7 +33,8 @@ import { SelectAssessmentsComponent } from "./results/select-assessments/select-
     BrowserAnimationsModule,
     FormsModule,
     DataTableModule,
-    SharedModule
+    SharedModule,
+    PopoverModule.forRoot()
```
wtritch: This is where I'm a little fuzzy on .forRoot() If we use .forRoot() should it be in the app module?

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -87,19 +87,44 @@ <h3 class="gray-darkest mb-md">
             </ng-template>
           </p-column>
           <p-column field="session" header="{{'labels.groups.results.assessment.exams.cols.session' | translate}}" [sortable]="true"></p-column>
-          <p-column field="enrolledGrade" header="{{'labels.groups.results.assessment.exams.cols.grade' | translate}}" [sortable]="true" ></p-column>
-          <p-column field="administrativeCondition" header="{{'labels.groups.results.assessment.exams.cols.status' | translate}}" [sortable]="true" >
+          <p-column field="enrolledGrade" header="{{'labels.groups.results.assessment.exams.cols.grade' | translate}}" [sortable]="true" [style]="{'text-align':'center'}" ></p-column>
```
wtritch: Recommend no direct-injected styles (or [style] which creates a data-binding to a constant.)

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
```diff
@@ -121,6 +121,10 @@ export class AssessmentResultsComponent {
       (this.isIab ? "iab" : "ica") + ".performance";
   }
 
+  get performanceLevelHeaderInfo() {
```
wtritch: Is this being used?  It looked like you had to hard-code the status one

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
jtreuting: good call, will change

webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.ts
jtreuting: yeah, it's used in the commented out one for the performance column.  It is a dynamic column header and info based on whether it is IAB or ICA.  IAB calls it Reporting Category and ICA calls it Achievement Level but we back it with the same data.

PR #174, opened 2017-06-27
prunes endpoint data models

esotrope:
1. Removes unused Exam.opportunity, Exam.completenessId, Exam.administrationConditionId, StudentContext.schoolYear, Assessment.cutPoints, Student.genderId
2. Adds Exam.schoolYear, Exam.gradeId (moved from StudentContext)

TODO:
1. Move client to use exam.gradeId instead of exam.studentContext.gradeId

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
@@ -1,6 +1,5 @@
 package org.opentestsystem.rdw.reporting.exam;
 
-import com.google.common.base.MoreObjects;
 import org.apache.commons.lang.builder.ToStringBuilder;
 
 import java.time.Instant;
@@ -16,12 +15,11 @@ public class Exam {
     private Long assessmentId;
     private String sessionId;
     private Instant dateTime;
+    private Integer schoolYear;
+    private int gradeId;
     private Student student;
     private StudentContext studentContext;
-    private Long opportunity;
-    private int administrativeConditionId;
     private String administrativeConditionCode;
-    private int completenessId;
     private String completenessCode;
     private ScaleScore scaleScore;
     private List<ScaleScore> claimScaleScores;
@@ -58,10 +56,17 @@ public class Exam {
     }
 
     /**
-     * @return the student's attempt number on the assessment
+     * @return the school year in which the exam was taken
      */
-    public Long getOpportunity() {
-        return opportunity;
+    public Integer getSchoolYear() {
+        return schoolYear;
+    }
+
+    /**
+     * @return the grade entity ID of the student at the time of the exam
+     */
+    public int getGradeId() {
+        return gradeId;
     }
 
     /**
@@ -86,28 +91,10 @@ public class Exam {
     }
 
     /**
-     * @deprecated use {@link #getCompletenessCode()} instead.
-     * @return the completeness of the exam (partial, complete)
-     */
-    @Deprecated
-    public int getCompletenessId() {
-        return completenessId;
-    }
-
-    /**
      * @return the completeness of the exam
      */
     public String getCompletenessCode() {
         return completenessCode;
-    }
-
-    /**
-     * @deprecated use {@link #getAdministrativeConditionCode()} instead.
-     * @return the condition in which the exam was administered (valid, invalid, standard, non standard)
-     */
-    @Deprecated
-    public int getAdministrativeConditionId() {
-        return administrativeConditionId;
     }
 
     /**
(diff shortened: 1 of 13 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
```diff
@@ -58,10 +56,17 @@ public Instant getDateTime() {
...
+    public Integer getSchoolYear() {
+        return schoolYear;
+    }
+
+    /**
+     * @return the grade entity ID of the student at the time of the exam
+     */
+    public int getGradeId() {
+        return gradeId;
     }
```
wtritch: Interesting, why do we sometimes have an `int` and other times an `Integer`?  Is it for nullability?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Exams.java
```diff
@@ -37,10 +37,8 @@ private Exams(){
         builder.id(row.getLong("id"))
                 .sessionId(row.getString("session_id"))
                 .dateTime(row.getTimestamp("completed_at").toInstant())
-                .opportunity(getNullable(row, row.getLong("opportunity")))
-                .completenessId(row.getInt("completeness_id"))
+                .gradeId(row.getInt("grade_id"))
                 .completenessCode(row.getString("completeness_code"))
-                .administrativeConditionId(row.getInt("administration_condition_id"))
```
wtritch: We aren't setting school year?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
```diff
@@ -38,8 +38,8 @@
                         .build(),
                 (row, index) -> Exams.map(row, Exam.builder())
                         .assessmentId(row.getLong("asmt_id"))
+                        .schoolYear(row.getInt("school_year"))
```
wtritch: Ah, here it is.  Should this be inside the Exams.map call?

webapp/src/main/resources/application.sql.yml
```diff
@@ -75,12 +72,11 @@ sql:
...
           join student_group_membership sgm on e.student_id=sgm.student_id
           join school s on e.school_id=s.id
-          where
-              e.school_year=:school_year
-              and sgm.student_group_id=:group_id
-              and (0=:group_subject_id or a.subject_id=:group_subject_id)
-              and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
-              and a.id = e.asmt_id
+          where e.school_year=:school_year
+          and sgm.student_group_id=:group_id
```
wtritch: nit, indents

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
esotrope: Yes - nullability. student exam history displays school year and schoolgrade/group exam results do not.

wtritch: hmm... I've never had luck with optionally-saturated resource models that differ based upon how they were requested.  It's always bitten me sooner or later.

esotrope: The standing strategy is to provide targeted responses for each view since this is a tailored API and not a general use API.
I think eventually we can make use of inheritance (or variant duplicate models) instead of having one model with optional fields. That way it is more explicit what you can expect the model to have. Would this satisfy your concerns?

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
```diff
@@ -16,12 +15,11 @@
     private Long assessmentId;
     private String sessionId;
     private Instant dateTime;
+    private Integer schoolYear;
+    private int gradeId;
```
agorina: Just curious: why schoolYear is `Integer` and not `int`?

esotrope: It's nullable - student exam history uses it but other exam result views do not

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/Exam.java
agorina: I tend to agree with @wtritch on this subject.

wtritch: I realize we're trying to be lean and trim here, but I've learned to love a Student is a Student is a Student, regardless of where it's coming from.  I'm definitely not advocating for a change here and now, but I've never seen polymorphic resource models end well.

PR #175, opened 2017-06-27
Feature/item by points grid headers

jtreuting:
updated items by points earned header to match latest mockup
- points columns are 0, 1, 2 instead of 0 points, 1 point, etc.
- added popover for info on Full Credit column

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
+++ b/webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
@@ -162,7 +162,11 @@
           <p-column field="fullCredit" [sortable]="true">
             <ng-template pTemplate="header">
               {{'labels.groups.results.assessment.items.cols.full-credit' | translate}}
-              <span class="icon"><i class="fa fa-info-circle"></i></span>
+              <span class="icon"
+                    popover="{{'labels.groups.results.assessment.items.cols.full-credit-info' | translate}}"
+                    popoverTitle="{{'labels.groups.results.assessment.items.cols.full-credit' | translate}}"
+                    triggers="click:mouseleave"
+              ><i class="fa fa-info-circle"></i></span>
             </ng-template>
             <ng-template let-item="rowData" pTemplate="body">
                 <span [hidden]="showValuesAsPercent">{{ item.fullCredit }}</span>
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -122,7 +122,8 @@
               "difficulty": "Item Difficulty",
               "standard": "Standard",
               "full-credit": "Full Credit",
-              "x-points": "{{id}} Points"
+              "full-credit-info": "The full credit column represents ... need real data here",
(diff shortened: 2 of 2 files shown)


webapp/src/main/webapp/src/app/groups/results/assessment/assessment-results.component.html
```diff
@@ -162,7 +162,11 @@ <h3 class="gray-darkest mb-md">
           <p-column field="fullCredit" [sortable]="true">
             <ng-template pTemplate="header">
               {{'labels.groups.results.assessment.items.cols.full-credit' | translate}}
-              <span class="icon"><i class="fa fa-info-circle"></i></span>
+              <span class="icon"
+                    popover="{{'labels.groups.results.assessment.items.cols.full-credit-info' | translate}}"
+                    popoverTitle="{{'labels.groups.results.assessment.items.cols.full-credit' | translate}}"
+                    triggers="click:mouseleave"
```
esotrope: does this rely on mouseleave by chance? just checking because I don't think touch devices fire this event

jtreuting: yeah i mentioned in the last PR that i wasn't sure i loved it.  if you don't do that then you have to click the little icon to close it or it stays open since focus (meaning if you click somewhere else it goes away) only works on buttons and not span or anchors.

i guess we could create our own popover template with a close icon in the title and wire that up?

jtreuting: or we could make it the icon a button and style it to not look like one?

esotrope: yeah I don't know what's best - I'll leave this one for @aaronsleeper and @jessicacolton

jtreuting: i'll probably leave it for now and make a JIRA to followup/test it out

jtreuting: Created DWR-494 to research and adjust if needed

bgsavage: After trying it out, `:mouseleave` might be kind of annoying if you're trying to read something and then accidentally move your mouse.  This might make more sense if it was shown on `:mouseenter` but since it's a click to enter, it should be a click to leave (not necessarily the same button, but maybe a close icon like you mentioned?)

esotrope: I haven't tried out this specific piece of the ui yet but generally I would agree with you @bgsavage. Click to open and click to dismiss tends to be the best experience. Also it's compatible with mobile so it will work on all media devices the same

jtreuting: I created a story to address it.  I think the best solution is to use <button class="btn btn-link"><i ....></i></button> and get some custom styling to remove the button padding, border and font color for it so it looks "normal".  Then the focus will work, click to open, click anywhere else to close.

PR #176, opened 2017-06-27
DWR-303 Add API to quickly search for a student by id with at least o…

wtritch:
…ne exam.


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
```diff
@@ -47,4 +50,15 @@
...
 
+    @Override
+    public boolean existsForStudent(@NotNull final PermissionScope permissionScope, final long studentId) {
+        return !template.queryForList(
+                existsForStudentIdQuery,
+                ImmutableMap.<String, Object>builder()
+                        .putAll(getSecurityParameters(permissionScope))
+                        .put("student_id", studentId)
+                        .build(),
+                Long.class).isEmpty();
```
wtritch: The query has a limit 1.  If the resulting list is empty, no exams!

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamController.java
```diff
@@ -18,11 +19,27 @@
         this.service = service;
     }
 
-    @GetMapping("/api/studentExams")
+    @GetMapping("/api/students/{ssid}/exams")
```
wtritch: Changed the request API slightly to align with the additional endpoint below.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamController.java
```diff
@@ -18,11 +19,27 @@
...
+     * @param user      The authenticated user
+     * @param ssid      The student SSID
+     * @param hasExams  Filter parameter
+     * @return The student
+     */
+    @GetMapping("/api/students/{ssid}")
+    public Student getStudent(
+            @AuthenticationPrincipal final User user,
+            @PathVariable("ssid") final String ssid,
+            @RequestParam("hasExams") final String hasExams) {
```
wtritch: I added a "hasExams" query param here, just to isolate this endpoint from a more generic student API endpoint, should one crop up in the future.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamControllerIT.java
```diff
@@ -66,4 +66,24 @@ public void getStudentExamsShouldReturnHistoryWhenFound() throws Exception {
...
+        mvc.perform(get("/api/students/a?hasExams=true"))
+                .andExpect(status().isOk())
+                .andExpect(jsonPath("$.id").value((int) student.getId()));
+    }
+
+    @Test
+    public void itShouldReturnA404IfAStudentWithExamsCannotBeFound() throws Exception {
+        final Student student = Student.builder().id(1).build();
+        when(service.findStudentWithExams(any(User.class), eq("a")))
+                .thenThrow(new NoSuchElementException("Bad Juju"));
```
bgsavage: Never throw away an exotic...

webapp/src/main/resources/application.sql.yml
```diff
@@ -35,6 +35,14 @@ sql:
       where e.student_id=:student_id
         and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
 
+    existsForStudentId: >-
+      select e.id
+      from exam e
+        join school s on e.school_id=s.id
+      where e.student_id=:student_id
+        and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+      limit 1
```
wtritch: This should be a pretty fast exists query that returns a single exam id for a student.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
```diff
@@ -51,4 +51,22 @@ public StudentExamHistory getExams(@NotNull final User user, final String studen
         return new StudentExamHistory(student, assessments, exams);
     }
 
+    @Override
+    public Student findStudentWithExams(@NotNull final User user, final String studentSsid) {
+        final Student student = studentRepository.findOneBySsid(studentSsid);
+        if (student == null) {
+            throw new NoSuchElementException(String.format("No student with SSID: %s", studentSsid));
```
esotrope: Since we reuse this we could make a private method for this

webapp/src/main/resources/application.sql.yml
agorina: Any reason you do not want to combine it into a SQL statement that finds a student by SSID?

I do not think we ever would want to 'find a student' without the exams.

esotrope: should we do two db calls for this student exists/has exams method? or just join student table here and do a where on its ssid field?

wtritch: Good point

wtritch: Good point

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
esotrope: So queryForList doesn't throw an empty data exception like queryForObject can?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
```diff
@@ -51,4 +51,22 @@ public StudentExamHistory getExams(@NotNull final User user, final String studen
...
+        }
+
+        final boolean examsExist = examRepository.existsForStudent(
+                user.getPermissionsById().get("INDIVIDUAL_PII_READ").getScope(),
+                student.getId()
+        );
+        if (!examsExist) {
+            throw new NoSuchElementException(String.format("No student with SSID: %s", studentSsid));
```
malaffoon: is this message correct?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
malaffoon: Not thrilled with echoing ssid in exception. And i'll argue the exception type over a beer.

wtritch: Now that we're only doing it twice I'm tempted to leave it... Unless you feel strongly?

webapp/src/main/resources/application.sql.yml
wtritch: fixed

wtritch: fixed

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
wtritch: right.  But since I've changed it to return a single student, I opted for the same try/catch that you had before.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
```diff
@@ -51,4 +51,17 @@ public StudentExamHistory getExams(@NotNull final User user, final String studen
...
     }
 
+    @Override
+    public Student findStudentWithExams(@NotNull final User user, final String studentSsid) {
+        final Student student = examRepository.findStudentWithExams(
+                user.getPermissionsById().get("INDIVIDUAL_PII_READ").getScope(),
+                studentSsid
+        );
+        if (student == null) {
+            throw new NoSuchElementException(String.format("No student with SSID: %s", studentSsid));
```
malaffoon: hmph, changing the code to delete my comments ... still not thrilled with echoing ssid in the exception (which might escape out of a controller).

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
esotrope: It's fine

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
wtritch: I debated changing that... but I'm not sure I want to give more details away... regardless, we no longer do two calls to the DB so we can't distinguish any more =)

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamController.java
malaffoon: And that request param just has to exist? If i call `../students/1?hasExams=false`?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
wtritch: sometimes, when surfing logs to find problems it can be helpful.  Of course, if we have log threading with the original request information this becomes less useful =)

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
esotrope: That is one thing we need to shut down actually is stacktraces and error messages reaching the client. I think the client should always get something generic like "404: Not found"

webapp/src/main/resources/application.sql.yml
```diff
@@ -35,6 +35,22 @@ sql:
...
         and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
 
+    existsForStudentId: >-
+      select stu.id,
+        stu.ssid,
+        stu.first_name,
+        stu.last_or_surname
+      from student stu
+      where stu.ssid=:student_ssid
+        and exists
```
malaffoon: I'll leave it to @agorina for the final word but i've been finding joins execute faster than exists; mysql doesn't know how to optimize subselects.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
wtritch: Why wouldn't we want it to?  If somebody complains that Little Bobby isn't showing up, wouldn't we want to know how their trying to access his records?

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamServiceTest.java
```diff
@@ -100,4 +101,22 @@ public void getExamsShouldReturnHistoryWhenStudentFoundWithExams() throws Except
...
     }
 
+    @Test
+    public void itShouldReturnAStudentWithExams() throws Exception {
+        final Student student = Student.builder().id(1).build();
+        when(examRepository.findStudentWithExams(PermissionScope.STATEWIDE, "a"))
+                .thenReturn(student);
+
+        assertThat(service.findStudentWithExams(user, "a")).isEqualTo(student);
+        verify(examRepository).findStudentWithExams(PermissionScope.STATEWIDE, "a");
```
malaffoon: I don't think you need to verify here ... you set up the `when` and asserted the result.

webapp/src/main/resources/application.sql.yml
wtritch: weird... I always thought 'exists' were pretty fast.  My world view is crumbling... I can rework as a join if @agorina confirms or you're pretty sure.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamServiceTest.java
wtritch: Removed.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
wtritch: Ah... my bad, I was thinking in terms of logging in the @ControllerAdvice rather than actual response data.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamController.java
wtritch: Correct, until/unless we want a more generic StudentController/StudentService with a StudentFilter request parameter object.  At least, that was my thought.  Objections?

webapp/src/main/resources/application.sql.yml
wtritch: One thing about the join implementation is that I'd need to GROUP BY to avoid returning the same student for every exam... That felt messier than an exists sub-query.  But, again, whatever the group prefers.

esotrope: I would imagine the implementation being this:
```sql
select 1
from exam e
  join student st on e.student_id=st.id
where st.ssid=:ssid
limit 1
```
would the group by even matter with limit 1?

agorina: @malaffoon is correct that we have seen a way slower performance with `exists, but it is not consistent. I was the one who actually put the `exists` clause in here. Before doing it I tested it on awsdev and it worked fast. I would probably leave it as is but remember when we are testing with the large data set.

wtritch: ok, going to leave as-is for now.

wtritch: I ended up switching it to a simple join with a limit 1, much like @esotrope's thoughts above.

PR #177, opened 2017-06-28
Create Assessments Module

bgsavage:
This is first step of a multi-step front end code refactor.  In order to reuse all the components we created for the group results page, they need to be easily accessible on a shareable module.  This creating the assessments module.  There's still some cleanup to do in the GroupsModule (mostly renaming the "assessment" items that remained to "groupAssessments").  The next step will be to break out the group results page into an assessment results page.


bgsavage:
Current directory structure:
image

webapp/src/main/webapp/src/app/search.pipe.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
```
esotrope: another .js slipped in. 
we should just put a src/*.js in the gitignore I think

bgsavage: Sounds like a good idea

PR #179, opened 2017-06-28
DWR-303 Add widget to find student by SSID to homepage.

wtritch:
I've also added a new "test" directory to contain re-usable mock classes.  Right now it just contains a mock.data.service.ts file/class.

### DIFF
--- a/webapp/src/main/webapp/src/app/student/student.component.html
+++ b/webapp/src/main/webapp/src/app/student/student.component.html
@@ -3,10 +3,15 @@
     <div class="flex-children">
       <div class="flex-child grow">
         <label for="student-id-input">{{'labels.student.title' | translate}}</label>
-        <input formControlName="ssid" type="text" class="form-control" id="student-id-input" placeholder="{{'labels.student.prompt' | translate}}">
+        <input formControlName="ssid"
+               type="text"
+               class="form-control"
+               id="student-id-input"
+               required
+               placeholder="{{'labels.student.prompt' | translate}}">
       </div>
       <div class="flex-child">
-        <button type="submit" class="btn btn-primary">{{'buttons.search' | translate}}</button>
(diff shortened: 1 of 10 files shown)


webapp/src/main/webapp/src/app/student/student.component.html
```diff
@@ -0,0 +1,16 @@
+<form [formGroup]="searchForm" (ngSubmit)="performSearch()" class="form-group">
+  <div class="form-group">
+    <div class="flex-children">
+      <div class="flex-child grow">
+        <label for="student-id-input">{{'labels.student.title' | translate}}</label>
+        <input formControlName="ssid" type="text" class="form-control" id="student-id-input" placeholder="{{'labels.student.prompt' | translate}}">
+      </div>
+      <div class="flex-child">
+        <button type="submit" class="btn btn-primary">{{'buttons.search' | translate}}</button>
```
bgsavage: Should disable submit if form is not valid?

esotrope: Haven't tried out the current impl but yeah i think each form control needs to be filled in sequence and if a previous item is not filled the following items should be disabled

wtritch: works for me, I'll disable submit unless there's something in the input

webapp/src/main/webapp/src/test/mock.data.service.ts
```diff
@@ -0,0 +1,14 @@
+
+import Spy = jasmine.Spy;
```
bgsavage: This is a great start.  Eventually I think we should just having a TestingModule that other tests imports.

webapp/src/main/webapp/src/app/student/student.component.html
wtritch: done

PR #180, opened 2017-06-28
adds logic to ignore unsupported and invalid grants

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -61,11 +61,19 @@ public DefaultAuthorizationService(
...
-                grants.add(grant);
+                // Only add grants for the application's configured state
+                if (stateId.equals(grant.getStateId())) {
+                    grants.add(grant);
+                }
+            } catch (IllegalArgumentException exception) {
+
+                // Ignore grants with unexpected entity levels or unexpected formatting
+                if (logger.isDebugEnabled()) {
+                    logger.debug("invalid grant: {}", encodedGrant);
```
bgsavage: The implementation of `debug` requires you to to check `isDebugEnabled()` every time you want to write to debug?

esotrope: no but it's good practice for log levels that are usually not enabled. It prevents extra computation like string substitution from occurring when the log level is disabled

malaffoon: I am 90% sure that you don't need to do this, that it is repeating a check being done under the covers. Now, if you were constructing/formatting the string locally, your logic would apply.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -61,11 +61,19 @@ public DefaultAuthorizationService(
...
+                final Grant grant = Grants.from(encodedGrant);
 
-            // Only add grants for the application's configured state
-            if (stateId.equals(grant.getStateId())) {
-                grants.add(grant);
+                // Only add grants for the application's configured state
+                if (stateId.equals(grant.getStateId())) {
+                    grants.add(grant);
+                }
+            } catch (IllegalArgumentException exception) {
```
malaffoon: nit: could be `final`

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
esotrope: sure in this case it's unnecessary but if we ever build an object or do processing in the debug log I think we should add it

PR #183, opened 2017-06-29
removes state entity level filter when computing spring authorities

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -145,14 +145,12 @@ public DefaultAuthorizationService(
         final Map<String, Collection<String>> permissionsByRole = getPermissionsByRole();
         final ImmutableSet.Builder<GrantedAuthority> authorities = ImmutableSet.builder();
         for (Grant grant : grants) {
-            if (grant.getEntityLevel() == STATE) {
```
esotrope: Removed this line.
Previously this incorrect role/permission filtering was removed for the /user endpoint but not here.

jtreuting: was this fixing a bug?  what is the context for the change I guess

esotrope: Yes - this should fix the HTTP 403 described in DWR-500

PR #185, opened 2017-06-30
DWR-306 Modify student history API to include Assessment and School i…

wtritch:
…nformation for each Exam.

After reading through the Student Results table specs, we were missing some data from the student history exam payload.  This PR modifies the response payload from two disconnected collections of Exam and Assessment instances into a single collection of wrapper models, each containing an Exam, Assessment, and School.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
@@ -1,12 +1,15 @@
 package org.opentestsystem.rdw.reporting.search.exam;
 
 import com.google.common.collect.ImmutableList;
+import org.apache.commons.lang.StringUtils;
 import org.opentestsystem.rdw.common.model.AssessmentType;
 import org.opentestsystem.rdw.common.model.Subject;
 import org.opentestsystem.rdw.reporting.exam.Assessment;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+
+import static org.apache.commons.lang.StringUtils.EMPTY;
 
 public final class Assessments {
 
@@ -21,21 +24,25 @@ public final class Assessments {
     }
 
     public static <A extends Assessment.Builder> A map(final ResultSet row, final A builder) throws SQLException {
+        return map(row, builder, EMPTY);
+    }
 
-        final AssessmentType type = AssessmentType.valueOf(row.getInt("type_id"));
+    public static <A extends Assessment.Builder> A map(final ResultSet row, final A builder, final String prefix) throws SQLException {
 
-        builder.id(row.getLong("id"))
-                .name(row.getString("label"))
-                .gradeId(row.getInt("grade_id"))
+        final AssessmentType type = AssessmentType.valueOf(row.getInt(prefix + "type_id"));
+
+        builder.id(row.getLong(prefix + "id"))
+                .name(row.getString(prefix + "label"))
+                .gradeId(row.getInt(prefix + "grade_id"))
                 .type(type)
-                .subject(Subject.valueOf(row.getInt("subject_id")))
-                .schoolYear(row.getInt("school_year"));
+                .subject(Subject.valueOf(row.getInt(prefix + "subject_id")))
+                .schoolYear(row.getInt(prefix + "school_year"));
 
         // IABs do not have claims
         if (type != AssessmentType.IAB) {
             final ImmutableList.Builder<String> claimCodes = ImmutableList.builder();
             for (String claimCodeColumnName : CLAIM_CODE_COLUMN_NAMES) {
-                final String claimCode = row.getString(claimCodeColumnName);
+                final String claimCode = row.getString(prefix + claimCodeColumnName);
                 if (claimCode == null) {
                     break;
                 }
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
@@ -1,9 +1,12 @@
 package org.opentestsystem.rdw.reporting.search.exam.student;
 
 import com.google.common.collect.ImmutableMap;
+import org.opentestsystem.rdw.reporting.exam.Assessment;
 import org.opentestsystem.rdw.reporting.exam.Exam;
+import org.opentestsystem.rdw.reporting.exam.Organization;
 import org.opentestsystem.rdw.reporting.exam.Student;
 import org.opentestsystem.rdw.reporting.exam.StudentContext;
+import org.opentestsystem.rdw.reporting.search.exam.Assessments;
 import org.opentestsystem.rdw.reporting.search.exam.Exams;
 import org.opentestsystem.rdw.reporting.security.PermissionScope;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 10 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
```diff
@@ -21,21 +24,25 @@ private Assessments() {
     }
 
     public static <A extends Assessment.Builder> A map(final ResultSet row, final A builder) throws SQLException {
+        return map(row, builder, EMPTY);
+    }
+
+    public static <A extends Assessment.Builder> A map(final ResultSet row, final A builder, final String prefix) throws SQLException {
```
wtritch: I added an optional prefix so that I'm able to parse an Exam, School, and Assessment from a single record row.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
```diff
@@ -34,21 +37,40 @@
...
+                            row.getLong("school_id"),
+                            row.getString("school_name"));
+
+                    //Build the wrapper
+                    return new StudentHistoryExamWrapper.Builder()
+                            .withExam(exam)
+                            .withAssessment(assessment)
+                            .withSchool(school)
+                            .build();
+                }
```
wtritch: this is the big change, parsing three separate models from a single result row.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentHistoryExamWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.opentestsystem.rdw.reporting.exam.Assessment;
+import org.opentestsystem.rdw.reporting.exam.Exam;
+import org.opentestsystem.rdw.reporting.exam.Organization;
+
+/**
+ * This model represents an Exam within the context of a student's history.
+ * It contains the Exam's associated Assessment and School information.
+ */
+public class StudentHistoryExamWrapper {
```
wtritch: Here's the new composite model.

webapp/src/main/resources/application.sql.yml
```diff
@@ -1,39 +1,56 @@
...
+        asmt.claim2_score_code as asmt_claim2_score_code,
+        asmt.claim3_score_code as asmt_claim3_score_code,
+        asmt.claim4_score_code as asmt_claim4_score_code,
+        asmt.min_score as asmt_min_score,
+        asmt.max_score as asmt_max_score,
+        asmt.cut_point_1 as asmt_cut_point_1,
+        asmt.cut_point_2 as asmt_cut_point_2,
+        asmt.cut_point_3 as asmt_cut_point_3,
+        school.id as school_id,
+        school.name as school_name
```
wtritch: We were already joining on all the tables we need to saturate the models, so there shouldn't be much of a performance hit.  This is where I'm adding the "asmt_" prefix to the assessment columns to distinguish between exam.id and asmt.id

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentHistoryExamWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+            this.exam = exam;
+            return this;
+        }
+
+        public Builder withAssessment(final Assessment assessment) {
+            this.assessment = assessment;
+            return this;
+        }
+
+        public Builder withSchool(final Organization school) {
```
esotrope: Builder setters we've been naming exactly the same as the underlying property with no prefix

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentHistoryExamWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * @return The exam's School
+     */
+    public Organization getSchool() {
+        return school;
+    }
+
+    void setSchool(final Organization school) {
```
esotrope: I think protected properties is a little tighter than package-private right? If it's package private isn't this object then mutable within the package?

webapp/src/main/resources/application.sql.yml
esotrope: Do we absolutely need the prefix stuff? Can we avoid the conflict by just selecting exam.asmt_id instead of selecting asmt.id as asmt_id?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
esotrope: If we can possibly get away without the prefix that would be nice I think. Hopefully we can make everything as homogenous as possible.
Also I think the JVM internally pools strings so we wont lose any memory for using "" for empty strings.

agorina: I tend to agree with Landon. I am failing to see the benefits of using the prefix. I maybe missing something...?

wtritch: Both exam and assessment tables have an "id" and "grade_id" column.  The alternative is to add some per-column if checks into the assessment mapper which feels messier to me.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentHistoryExamWrapper.java
agorina: Just to share the pattern that we have been using: the Builder directly sets the private properties of the object it is building and we have no setters. Your code really does not break anything, but just if you want to stay consistent :)
```
public class RdwImport {
  private Long id;
....
   public static Builder builder() {
        return new Builder();
    }

    public static class Builder {
         private Long id;

        public RdwImport build() {
            final RdwImport rdwImport = new RdwImport();
            rdwImport.id = id;
...
           return rdwImport;
        }
....
```

wtritch: I've found package-private to be more restrictive, since it prevents subclasses from munging the properties. I don't feel too strongly about it, though.  Both are a sign to developers that the values shouldn't be modified, and nothing is sacred with reflection.  I just generally prefer setters over setting properties directly.  Would you prefer protected setters?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentHistoryExamWrapper.java
wtritch: I like consistency, will change.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentHistoryExamWrapper.java
wtritch: ...ok, that's two votes for setting properties directly.  Will conform :-p

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
agorina: Why couldn't you change the the SQL to return "select a.id as asmt_id" and so on.
I am really not trying to be difficult, just want to understand. Is it because the same object can be loaded from different repos?

wtritch: That's exactly right, we load assessments from a few queries, in a few repositories, and I'd like to re-use the one-true-mapper rather than copy and modify it, so that if an Assessment model ever changes we only have to update one mapper.

webapp/src/main/resources/application.sql.yml
wtritch: Prefixing seemed like the simplest, most consistent way to avoid column conflicts between model columns.  I'm open to suggestions?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/Assessments.java
agorina: Thank you for the explanation.

esotrope: I'm curious just about sql behavior in general
If I run "select e.id, a.id ... join a on e" there will be a conflict right?
But If I run "select e.id ... join a on e" will there still be a conflict on the "id" field?

wtritch: nope, because you're not selecting it.  When you select it, you need to make sure the things you're selecting don't try to use the same column name.

esotrope: Ok cool. I think i get the need for this now btw - thanks 👍

PR #193, opened 2017-07-04
Sorting by array property in datatable doesn't work, fixed by creatin…

jtreuting:
…g property names for each

This feels weird and only works because we know that there can only be up to 4 claims so we can create denormalized properties.

I might be missing something so want people's thoughts on this.  @wtritch are you having this issue on your student test history results page as well?

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/model/exam.model.ts
+++ b/webapp/src/main/webapp/src/app/assessments/model/exam.model.ts
@@ -17,4 +17,20 @@ export class Exam {
   limitedEnglishProficiency: boolean;
   ethnicities: string[];
   claimLevels: number[];
+
+  get claimLevel0() {
+    return this.claimLevels[0];
+  }
+
+  get claimLevel1() {
+    return this.claimLevels[1];
+  }
+
+  get claimLevel2() {
+    return this.claimLevels[2];
+  }
+
+  get claimLevel3() {
+    return this.claimLevels[3];
+  }
 }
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -147,7 +147,7 @@ <h3 class="gray-darkest mb-md">
           <!-- Claim Score columns -->
           <p-column *ngFor="let claim of claimCodes; let idx = index"
                     header="{{'enum.claim-code.' + claim | translate}}"
-                    field="{{'exam.claimLevels[' + idx + ']'}}"
+                    field="{{'claimLevel' + idx}}"
```
bgsavage: Yeah I had to do something similar for the dynamic columns on the items by points earned.  I didn't use getters because the size can be unknown, so I set it dynamically like:

`exam['identifier' + idx] = value`

webapp/src/main/webapp/src/app/assessments/model/exam.model.ts
```diff
@@ -17,4 +17,20 @@ export class Exam {
   limitedEnglishProficiency: boolean;
   ethnicities: string[];
   claimLevels: number[];
+
+  get claimLevel0() {
```
bgsavage: nit: Maybe just a comment of why we had to add these props ?

wtritch:
@jtreuting I hadn't run into this because my sample data set wasn't large enough =(
Looks like you don't need those extra properties, you can just update your `field` attribute to avoid using array annotation:
`field="{{'claimScores.' + idx + '.level'}}"` and it should work find without the need for extra properties on the model.

jtreuting:
Nice.  The one thing I didn't think of trying.  I'll change before merging.  Thanks

wtritch:
Thanks for identifying the issue :-). I fixed it on my PR

PR #195, opened 2017-07-04
Feature/student history ui

wtritch:
My apologies for the stupidly-large PR.  I'll add comments below at the interesting bits.


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
```diff
@@ -32,11 +32,11 @@
     }
 
     @Override
-    public StudentExamHistory getExams(@NotNull final User user, final String studentSsid) {
+    public StudentExamHistory getExams(@NotNull final User user, final Long studentId) {
```
wtritch: The back-end changes are just in support of navigating by student ID rather than SSID.  We now need to look up a student's history by the primary key.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
```diff
@@ -32,11 +32,11 @@
...
 
-        final Student student = studentRepository.findOneBySsid(studentSsid);
+        final Student student = studentRepository.findOneById(studentId);
         if (student == null) {
-            throw new NoSuchElementException(String.format("No student with SSID: %s", studentSsid));
+            throw new NoSuchElementException(String.format("No student with SSID: %s", studentId));
```
agorina: Nit: the error msg needs to be adjusted, it is not SSID anymore.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -29,6 +31,20 @@ export const routes: Routes = [
...
+        resolve: { examHistory: StudentExamHistoryResolve },
+        data: {
+          breadcrumb: {
+            translate: 'labels.student.results.crumb',
+            translateResolve: 'examHistory.student'
+          },
+          permissions: ['INDIVIDUAL_PII_READ']
+        },
+        component: StudentResultsComponent,
+        canActivate: [ AuthorizeCanActivate ]
```
wtritch: This adds the route to the student exam history page.  Note the "translateResolve" breadcrumb property, which injects the resolved object as an argument into the TranslatePipe so I can get a nice display like "Student Id: {{ssid}}"

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -52,6 +53,59 @@ export class AssessmentExamMapper {
...
+      if (apiModel.student.ethnicityCodes) {
+        apiModel.student.ethnicityCodes.forEach(code => uiModel.ethnicities.push(code));
+      }
+    }
+
+    if (apiModel.scaleScore) {
+      uiModel.score = apiModel.scaleScore.value;
+      uiModel.level = apiModel.scaleScore.level;
+      uiModel.standardError = apiModel.scaleScore.standardError;
+    }
```
wtritch: Since our Exam models have optional properties, I wrapped all of the sub-object references in an existence check to avoid NPEs while mapping.

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -75,53 +129,18 @@ export class AssessmentExamMapper {
...
-    uiModel.name = apiModel.name;
-    uiModel.grade = apiModel.gradeId;
-    uiModel.type = AssessmentType[ apiModel.type as string ];
-    uiModel.subject = apiModel.subject;
-    uiModel.claimCodes = apiModel.claimCodes || [];
-
-    return uiModel;
+    return apiScaleScores
+      .map((apiScore) => this.mapClaimScaleScoreFromApi(apiScore));
   }
```
wtritch: I added a simple ClaimScore model that contains the score, standardError, and level of each claim score, this is the simple collection mapper, with the item mapper below.

webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -19,36 +19,15 @@
...
-          <span *ngIf="property.indexOf('ethnicities') != -1">
-              <strong>{{ filters['ethnicities'].enumValue + property.substring(property.indexOf('.')) | translate}}</strong> {{ filters['ethnicities'].label | translate }}
-            </span>
-          <span class="icon"><i class="fa fa-close"></i></span>
-        </a>
-      </div>
-    </div>
+    <adv-filters-toggle [displayOptions]="filterDisplayOptions"
+                        [filterBy]="clientFilterBy"
+                        (removed)="removeFilter($event)"></adv-filters-toggle>
```
wtritch: Since I wanted to re-use the advanced filters and advanced filters toggling in the student results page, I wrapped the toggle behavior in a component for re-use in the student results component.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -42,12 +41,12 @@ export class AssessmentsComponent implements OnInit {
...
   showValuesAsPercent: boolean = true;
-  expandFilterOptions: boolean = false;
+  filterDisplayOptions: any = {
+    expanded: true
+  };
   clientFilterBy: FilterBy;
 
-  filters: any[] = [];
   filterOptions: ExamFilterOptions = new ExamFilterOptions();
-  currentFilters = [];
```
wtritch: removed some unused properties.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -42,12 +41,12 @@ export class AssessmentsComponent implements OnInit {
   assessmentProvider: AssessmentProvider;
 
   showValuesAsPercent: boolean = true;
-  expandFilterOptions: boolean = false;
+  filterDisplayOptions: any = {
+    expanded: true
+  };
```
wtritch: I wrapped the display options in an object so I could pass/bind to values by reference.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -125,16 +124,11 @@ export class AssessmentsComponent implements OnInit {
...
               private filterOptionService: ExamFilterOptionsService,
-              private examFilterService: ExamFilterService,
               private gradeService: GradeService) {
     this.clientFilterBy = new FilterBy()
   }
 
   ngOnInit() {
-    this.examFilterService.getFilterDefinitions().forEach(filter => {
-      this.filters[ filter.name ] = filter;
-    });
```
wtritch: Since this was only used by the filter toggle display, I moved it down into the re-usable adv-filter-toggle component.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
```diff
@@ -32,11 +32,11 @@
     }
 
     @Override
-    public StudentExamHistory getExams(@NotNull final User user, final String studentSsid) {
+    public StudentExamHistory getExams(@NotNull final User user, final Long studentId) {
 
-        final Student student = studentRepository.findOneBySsid(studentSsid);
+        final Student student = studentRepository.findOneById(studentId);
```
bgsavage: nit: this is no longer a `findOne` but a `get` yeah?

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -43,7 +43,7 @@
   </div>
 
   <!-- Student Filters -->
-  <div id="filter-by-student" class="flex-child grow">
+  <div *ngIf="showStudentFilter" id="filter-by-student" class="flex-child grow">
```
wtritch: To be able to re-use the adv-filters component, I added a toggle to optionally display student-scoped filters (which don't make sense when viewing exams for a single student.)

webapp/src/main/webapp/src/app/assessments/model/exam.model.ts
```diff
@@ -16,5 +18,6 @@ export class Exam {
   economicDisadvantage: boolean;
   limitedEnglishProficiency: boolean;
   ethnicities: string[];
-  claimLevels: number[];
+  schoolYear: number;
+  claimScores: ClaimScore[];
```
wtritch: Small change to @jtreuting's code to map all of the claim score information rather than just the level values.

webapp/src/main/resources/application.sql.yml
```diff
@@ -360,13 +360,13 @@ sql:
...
-    findOneBySsid: >-
+    findOneById: >-
       select id,
         ssid,
         first_name,
         last_or_surname
       from student s
-      where s.ssid=:ssid
+      where s.id=:studentId
        and exists (select 1 from exam e where e.student_id = s.id)
```
bgsavage: @agorina do we need this existence check?  My assumption is the student wouldn't be in the student table if it didn't have exams.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -113,7 +114,7 @@ <h3 class="gray-darkest mb-md">
               ><i class="fa fa-info-circle"></i></span>
             </ng-template>
             <ng-template let-exam="rowData" pTemplate="body">
-              <span *ngIf="exam.administrativeCondition == 'IN' || exam.administrativeCondition == 'SD'" class="label border-only maroon">{{ 'enum.administrative-condition.' + exam.administrativeCondition | translate }}</span>
+              <span *ngIf="exam.administrativeCondition !== 'Valid'" class="label border-only maroon">{{ 'enum.administrative-condition.' + exam.administrativeCondition | translate }}</span>
```
wtritch: We want to show "Invalid, Standardized, Non-Standardized"  The only option we never want to show is "Valid" ...

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -140,18 +141,18 @@ <h3 class="gray-darkest mb-md">
               ><i class="fa fa-info-circle"></i></span>
             </ng-template>
             <ng-template let-exam="rowData" pTemplate="body">
-              {{ exam.score }} ± {{ exam.standardError }}
+              {{ exam | scaleScore }}
```
wtritch: I created a scaleScore pipe to turn an exam into "{{exam.score}} += {{exam.standardError * 1.5}}" since we use that in a few tables.

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.html
```diff
@@ -10,10 +10,10 @@
       </li>
       <li *ngFor="let crumb of breadcrumbs; let last = last;" [ngClass]="{'active': last }">
         <a *ngIf="!last" [routerLink]="[crumb.url, crumb.params]">
-          {{ crumb.requiresTranslate? (crumb.label | translate) : crumb.label }}
+          {{ crumb.requiresTranslate? (crumb.label | translate:crumb.translateParams) : crumb.label }}
         </a>
         <span *ngIf="last" [routerLink]="[crumb.url, crumb.params]">
-          {{ crumb.requiresTranslate? (crumb.label | translate) : crumb.label }}
+          {{ crumb.requiresTranslate? (crumb.label | translate:crumb.translateParams) : crumb.label }}
```
wtritch: This is the breadcrumb change that passes a resolved value into the translate pipe so it can be referenced by the translation string.

webapp/src/main/webapp/src/app/app.routes.ts
bgsavage: nit:  It'd be nice to have this as a translate object such as: 
`translate: { path: 'labels.student.results.crumb', resolve: 'examHistory.student' }`

webapp/src/main/webapp/src/app/shared/assessment-type.pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+import { AssessmentType } from "./enum/assessment-type.enum";
+
+/**
+ * This pipe is responsible for transforming an AssessmentType enum
+ * into display values.
+ * Default pipe displays human-readable text.
+ * 'color' argument displays associated assessment type color.
+ */
+@Pipe({name: 'assessmentType'})
+export class AssessmentTypePipe implements PipeTransform {
```
wtritch: This is a pipe for displaying AssessmentType enum values as strings... I'm not sure I like that we convert the strings 'ICA', 'IAB', etc from the server into an enum value, because it makes it difficult to work with on the front end, leading to methods like 'isIAB()' on the Assessment model.  Thoughts?
Regardless, the values are data and probably shouldn't be translated, so this pipe just turns them back into their enum string values.
I also added a color option to retrieve the associated color for an AssessmentType.

webapp/src/main/webapp/src/app/shared/enum/assessment-type.enum.ts
```diff
@@ -1,5 +1,5 @@
 export enum AssessmentType {
-  ICA = 1,
-  IAB = 2,
+  IAB = 1,
+  ICA = 2,
```
wtritch: Since it didn't look like anybody was relying on the specific order of this enum, I changed it to match the sorting in my mockup so I can easily sort by assessment type.

webapp/src/main/webapp/src/app/shared/scale-score.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+ * This pipe is responsible for transforming an Exam into a
+ * Scale Score display string.
+ */
+@Pipe({name: 'scaleScore'})
+export class ScaleScorePipe implements PipeTransform {
+
+  transform(exam: Exam): string {
+    let score: number = exam.score;
+    let stdErr: number = exam.standardError;
+    let errorBand: number = Math.round(stdErr * 1.5);
```
wtritch: According to the docs, the "error band" should be 1.5 * standardError.

webapp/src/main/webapp/src/app/student/model/student-history-exam-wrapper.model.ts
```diff
@@ -0,0 +1,29 @@
+
+import { Assessment } from "../../assessments/model/assessment.model";
+import { School } from "../../user/model/school.model";
+import { Exam } from "../../assessments/model/exam.model";
+
+/**
+ * This model represents an exam from a student's history, with associated
+ * Assessment and School information.
+ */
+export class StudentHistoryExamWrapper {
```
wtritch: This is the primary model used by the student results page.  Each (exam, assessment, school) set defines one result row.

webapp/src/main/webapp/src/app/student/model/student.model.ts
```diff
@@ -0,0 +1,25 @@
+export class Student {
```
wtritch: I was surprised we didn't already have a Student model floating around...

webapp/src/main/webapp/src/app/student/results/model/student-results-filter-state.model.ts
```diff
@@ -0,0 +1,11 @@
...
+export class StudentResultsFilterState {
+
+  public years: number[];
+  public schoolYear: number = 0;
+  public subjects: string[];
+  public subject: string = "";
+
+  public filterBy: FilterBy = new FilterBy();
+
+}
```
wtritch: This filter object contains the current filter options/state for the student results component.  It's just a simple POJO passed by reference and holding state.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
bgsavage: Just want to double check you checked the template wasn't using these?  IntelliJ will warn that these are unused but isn't smart enough to to check the html, only checks the typescript.

webapp/src/main/webapp/src/app/student/results/student-results-filter.component.html
```diff
@@ -0,0 +1,60 @@
...
+                  [(ngModel)]="filterState.subject"
+                  (change)="onFilterChange()"
+                  class="form-control">
+            <option selected value="">{{ 'buttons.all' | translate }}</option>
+            <option *ngFor="let subject of filterState.subjects" [ngValue]="subject">{{subject}}</option>
+          </select>
+        </div>
+
+        <!-- Advanced Filters -->
+        <div class="flex-child ml-auto">
```
wtritch: I had to add an "ml-auto" (margin-left: auto) class to use a simple flex layout with the last object pulled-right.  FYI @aaronsleeper

webapp/src/main/webapp/src/app/student/results/student-results-filter.component.ts
```diff
@@ -0,0 +1,53 @@
...
+  selector: 'student-results-filter',
+  templateUrl: './student-results-filter.component.html'
+})
+export class StudentResultsFilterComponent implements OnInit {
+
+  @Input()
+  filterState: StudentResultsFilterState = new StudentResultsFilterState();
+
+  @Output()
+  filterChange: EventEmitter<any> = new EventEmitter();
```
wtritch: The \@Input passes in the filter state by reference, and the filterChange emits that it has been updated and it's state should be applied.

webapp/src/main/webapp/src/app/student/results/student-results.component.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+});
+
+function getFilteredExams(examMap: Map<AssessmentType, Map<string, StudentHistoryExamWrapper[]>>): StudentHistoryExamWrapper[] {
+  return Array.from(examMap.values())
+    .map((bySubject) => Array.from(bySubject.values()))
+    .reduce((a, b) => a.concat(b))
+    .reduce((a, b) => a.concat(b));
+}
+
+class MockBuilder {
```
wtritch: I'm wondering if we should have a re-usable test builder for creating generic test models across our tests... Thoughts?

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    return this.displayState[assessmentType]
+      && this.displayState[assessmentType][subject]
+      && this.displayState[assessmentType][subject].collapsed;
+  }
+
+  toggleCollapsed(assessmentType: AssessmentType, subject: string): void {
+    this.displayState[assessmentType] = this.displayState[assessmentType] || {};
+    this.displayState[assessmentType][subject] = this.displayState[assessmentType][subject] || {};
+    this.displayState[assessmentType][subject].collapsed = !this.displayState[assessmentType][subject].collapsed;
+  }
```
wtritch: Since the expand/collapse controller of each result table is outside of the scope of the table itself, I need this container component to track whether or not the component is collapsed.  Anyone have any null-safe typescript accessing techniques that could clean this up?

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -0,0 +1,196 @@
...
+      let type: AssessmentType = wrapper.assessment.type;
+      let subject: string = wrapper.assessment.subject;
+      let byType: Map<string, StudentHistoryExamWrapper[]> = examsByTypeAndSubject.get(type) || new Map();
+      let bySubject: StudentHistoryExamWrapper[] = byType.get(subject) || [];
+      bySubject.push(wrapper);
+      byType.set(subject, bySubject);
+      examsByTypeAndSubject.set(type, byType);
+    });
+
+    this.examsByTypeAndSubject = examsByTypeAndSubject;
```
wtritch: applying a filter involves filtering out the flat list of source exams, then constructing the type/subject/exam[] tree used by the display.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+    if (this.filterState.subject) {
+      searchParams.set("subject", this.filterState.subject);
+    }
+
+    //Update the current route without navigating
+    this.location.replaceState(
+      `/students/${this.examHistory.student.ssid}`,
+      searchParams.toString()
+    );
```
wtritch: We don't want to navigate when updating the filters, but we do want to deep-link the state.  This updates the browser location without performing a navigation.

webapp/src/main/webapp/src/app/assessments/model/exam.model.ts
bgsavage: 👍

webapp/src/main/webapp/src/app/student/results/tables/student-history-iab-table.component.html
```diff
@@ -0,0 +1,40 @@
+<p-dataTable [value]="exams"
```
wtritch: Since the IAB and ICA/Summitive tables have different columns/values/behavior, I broke them out into two table components.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
bgsavage: Oh... This is actually temporary and will be replaced by a visualization in all the mockups I've seen.

webapp/src/main/webapp/src/app/app.routes.ts
wtritch: hmm... except I expect it to be unnecessary 90% of the time, and then we'd have to change all our:
`translate: 'blah'` to `translate: { path: 'blah' }` for corner cases which might get annoying.  If you feel strongly, I can make the change, though.

webapp/src/main/resources/application.sql.yml
agorina: Yeah, we need the check.  With the 'student groups' feature, one could upload a student group with a student without the exams.

webapp/src/main/webapp/src/app/student/results/tables/student-history-ica-summitive-table.component.html
```diff
@@ -0,0 +1,81 @@
...
+    </ng-template>
+  </p-column>
+
+  <!-- Overall Score columns -->
+  <p-column field="exam.administrativeCondition"
+            header="{{'labels.student.results.cols.status' | translate}}"
+            sortable="true"
+            [hidden]="displayState.table !== 'overall'">
+    <ng-template let-exam="rowData.exam" pTemplate="body">
+      <span *ngIf="exam.administrativeCondition !== 'Valid'" class="label border-only maroon">{{ 'enum.administrative-condition.' + exam.administrativeCondition | translate }}</span>
```
wtritch: here's where we hide "Valid" labels for Summitive exams.

webapp/src/main/webapp/src/app/shared/enum/assessment-type.enum.ts
bgsavage: I think the order here was to match the api?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
wtritch: This is the StudentRepository finding one student by id, isn't it?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
wtritch: doh, fixed

webapp/src/main/webapp/src/app/shared/enum/assessment-type.enum.ts
agorina: I am not sure if it makes the difference, but the original mapping was reflecting the DB ids. Checkout the asmt_type table.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
wtritch: Yeah, I double-checked with a full-text-search before removing anything =)

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
wtritch: Once we no longer need it, we can just blow it away.

webapp/src/main/webapp/src/app/shared/enum/assessment-type.enum.ts
wtritch: I'll change it back, but the API just returns strings, I believe.  Which implies that all communication with it w.r.t. assessment types will be in those strings, won't it?

webapp/src/main/webapp/src/app/student/model/student-exam-history.model.ts
```diff
@@ -0,0 +1,20 @@
...
+  public get student(): Student {
+    return this._student;
+  }
+
+  public get exams(): StudentHistoryExamWrapper[] {
+    return this._exams;
+  }
+
+  constructor(
+    private _student: Student,
```
bgsavage: nit: I'm not a fan of trying to create immutability in typescript as it's still javascript under the hood so nothing is actually immutable.  

Also, I prefer we discuss before trying to alter the established pattern, as now our models are inconsistent depending on the developer's preference when creating them.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
malaffoon: Shouldn't parameter be `long` (primitive)? It can't be null.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamController.java
```diff
@@ -19,11 +19,11 @@
         this.service = service;
     }
 
-    @GetMapping("/api/students/{ssid}/exams")
+    @GetMapping("/api/students/{id}/exams")
     public StudentExamHistory getStudentExams(
             @AuthenticationPrincipal final User user,
-            @PathVariable("ssid") final String ssid) {
-        return service.getExams(user, ssid);
+            @PathVariable("id") final Long studentId) {
```
malaffoon: ditto about `long`

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamService.java
```diff
@@ -11,7 +11,7 @@
 interface StudentExamService {
 
     @PreAuthorize("hasAuthority('PERM_INDIVIDUAL_PII_READ')")
-    StudentExamHistory getExams(@NotNull User user, String studentSsid);
+    StudentExamHistory getExams(@NotNull User user, Long studentId);
```
malaffoon: ditto about `long`

webapp/src/main/webapp/src/app/student/model/student.model.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  get lastName(): string {
+    return this._lastName;
+  }
+
+  constructor(
+    private _id: number,
+    private _ssid: string,
+    private _firstName: string,
+    private _lastName: string
```
bgsavage: nit: I find this constructor set params to be error prone.  

`new Student(apiModel.id, apiModel.firstName, apiModel.lastName, apiModel.ssid)` compiles and doesn't look like there is anything wrong at the surface.

This doesn't happen with the use of setters (or even builder pattern if we're going for immutability)

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
wtritch: good point, will change

webapp/src/main/webapp/src/app/shared/scale-score.pipe.ts
malaffoon: What is this used for? The 1.5x is only for claims so it applies only to IABs and individual claims in ICA and Summatives.

webapp/src/main/webapp/src/app/student/model/student.model.ts
wtritch: good point, I like consistency.  Will change to public properties.
Tangent: Nothing is truly protected if you have access to the source, (i.e. reflection in Java,) it's more about demonstrating intent of what "should" be mutable vs what "shouldn't."

webapp/src/main/webapp/src/app/student/model/student-exam-history.model.ts
wtritch: Good point, I like consistency.  Will change to public properties.
Tangent: Nothing is really protected if you have access to the source, (i.e. reflection in Java.) It's more about demonstrating intent of what "should" be mutable vs what "shouldn't."

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamController.java
wtritch: I'll change this one too.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamService.java
wtritch: No!  This far, and no farther! I'm keeping this one. (j/k)

webapp/src/main/webapp/src/app/student/model/student.model.ts
wtritch: Will change to public properties.

webapp/src/main/webapp/src/app/student/results/student-results.component.spec.ts
bgsavage: Maybe, but those tend to get very large and cumbersome over time.  Maybe whenever we find the need to reuse we create model-based test builders (i.e. MockExamBuilder, MockAssessmentBuilder)

bgsavage: Another thought would be to use the standalone data `data.ts`, though they represent api models so they would need to be mapped first, but it'd be a great way to salvage that work when we throw away the standalone service.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  isCollapsed(assessmentType: AssessmentType, subject: string): boolean {
+    return this.displayState[assessmentType]
+      && this.displayState[assessmentType][subject]
+      && this.displayState[assessmentType][subject].collapsed;
+  }
+
+  toggleCollapsed(assessmentType: AssessmentType, subject: string): void {
+    this.displayState[assessmentType] = this.displayState[assessmentType] || {};
+    this.displayState[assessmentType][subject] = this.displayState[assessmentType][subject] || {};
```
bgsavage: nit: might be cleaner to do some initialization here so you don't have to check for existence when toggling collapsed and checking collapsed.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
bgsavage: Oh just saw this :)  Yeah might be cleaner just to initialize your object type for each of your assessmentTypes and subjects.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
wtritch: Just got hit up by @jtreuting:
"In the tables it is going to be scale score +/- standard error with scale score bolded"

So it sounds like we need more of a small directive than a pipe since it'll contain display information.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
bgsavage: I think we should deep link advanced filter state, but not now, only if we have time later.  It seems like more of a nice to have, and if we do it, we want it to be consistent across the app.  @jtreuting  thoughts?

bgsavage: Also, the reason I wanted to re-navigate on groups and school/grade because I want to hit my resolve again to load my initial data with the new params.  Any data I didn't want refreshed is in a parent resolve.  For you, you always get all the data immediately regardless of your parameters, so you don't need to hit your resolve again.  Thinking maybe we should call that out somewhere...

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    this.examsByTypeAndSubject = examsByTypeAndSubject;
+  }
+
+  /**
+   * Filter an exam by the current filter state.
+   *
+   * @param wrapper The exam wrapper to filter
+   * @returns {boolean} True if the exam should be visible
+   */
+  private filterExam(wrapper: StudentHistoryExamWrapper): boolean {
```
bgsavage: Filter logic might be a good candidate to remove from the component and into some service class (keep our components as light as we can, right?).  Service classes are much easier to test and run much faster as there's so much less setup.

bgsavage: nit: the name of this method should indicate the return value.  ie: `isExamVisible` .  The current name and return contradict each other.  I would think you should return true if the exam is filtered out based on `filterExam`, which is the opposite of visible.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+    // Completion state
+    if (this.filterState.filterBy.completion != -1) {
+      if (wrapper.exam.completeness !== this.filterState.filterBy.completion) {
+        return false;
+      }
+    }
+    // Off-Grade Assessment filter
+    if (this.filterState.filterBy.offGradeAssessment) {
+      if (wrapper.exam.enrolledGrade !== wrapper.assessment.grade) {
```
bgsavage: Yeah we already have a lot of this logic in the `ExamFilterService`, you weren't able to use this?  I'd have to have to maintain it in two spots.

webapp/src/main/webapp/src/app/shared/scale-score.pipe.ts
wtritch: This is used in the universal "1234 +/- 54" score table display.  Per @jtreuting we aren't going to be putting visualizations in the tables.  The standard-error * 1.5 is used to calculate the "blow/near/above standard" levels, so there's a debate as to whether we should display +/- (stdError * 1.5) which is more consistent with their data or +- (stdError) which is more consistent with data science.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: I called out that this method shouldn't navigate in the jsdocs =)

wtritch: Created DWR-523

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: Good point... I can probably modify it to work on StudentHistoryExamWrapper instances in addition to AssessmentExam instances.  I'll play with it and see what I can do.

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
```diff
@@ -31,29 +32,65 @@ export class ExamFilterService {
...
+   * @param filterBy        The currently-applied filters
+   * @returns {Exam[]} The filtered exams
+   */
+  filterExams(assessmentExam: AssessmentExam, filterBy: FilterBy): Exam[] {
+    return this.filterItems(
+      () => assessmentExam.assessment,
+      (exam) => exam,
+      assessmentExam.exams,
+      filterBy);
+  }
```
wtritch: @bgsavage I abstracted out this method to allow filtering a collection of "thingies" as long as each "thingy" can be resolved to both an Assessment and Exam.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: done

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: Renamed and changed to use exam filter service

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: done

webapp/src/main/webapp/src/app/shared/scale-score.pipe.ts
malaffoon: To be clear, if it is "below/near/above" then it is a claim performance and the +- 1.5x can be discussed as a reasonable alternative. If it is "level 1-4" then it is overall performance and the +- 1.5x doesn't make any sense at all.

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
```diff
@@ -31,29 +32,65 @@ export class ExamFilterService {
...
     let filters = this.getFilters(filterBy);
     for (let filter of filters) {
       let filterDefinition = this.getFilterDefinitionFor(filter);
 
-      if (filterDefinition.precondition(assessmentExam.assessment)) {
-        let filterValue = filterBy[filter];
+      results = results
+        .filter((item) => {
+          let assessment: Assessment = assessmentProvider(item);
+          if (!filterDefinition.precondition(assessment)) return true;
```
bgsavage: It seems like this will cause the precondition (and other evaluations being done here) way more times then it was before, once per exam per filter instead of just once per filter.

bgsavage: It might be cleaner just to have two seperate methods, one for the wrapper, and one for the assessmentExam, since the assessment -> exam relationship is different.

wtritch: Discussed offline: We ran a test against 600k exams. The existing AssessmentExam filtering took 24ms, with this change it took 233ms.  We reverted the change to filtering AssessmentExam items, but left the ability to filter generic items as a separate method.

PR #196, opened 2017-07-05
Feature/merge translations

bgsavage:


webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -0,0 +1,25 @@
...
+  private uiLoader = new TranslateHttpLoader(this.http, '/assets/i18n/', '.json');
+
+  getTranslation(lang: string) {
+    return this.uiLoader
+      .getTranslation(lang)
+      .mergeMap(uiTranslation => this.apiLoader
+        .getTranslation(lang)
+        .mergeMap(apiTranslation => {
+          // TODO: Should the API unflatten the JSON before returning?
+          let merged = _.merge(uiTranslation, JsonUnFlat.unflat(apiTranslation));
```
bgsavage: I brought in lodash so I could use _.merge, which does a deep merge.  Typescript's `Object.assign` I discovered only does a shallow merge, replacing entire objects at the first level rather than nested properties.

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -0,0 +1,25 @@
...
+  private apiLoader = new TranslateHttpLoader(this.http, '/api/translations/', '');
+  private uiLoader = new TranslateHttpLoader(this.http, '/assets/i18n/', '.json');
+
+  getTranslation(lang: string) {
+    return this.uiLoader
+      .getTranslation(lang)
+      .mergeMap(uiTranslation => this.apiLoader
+        .getTranslation(lang)
+        .mergeMap(apiTranslation => {
+          // TODO: Should the API unflatten the JSON before returning?
```
wtritch: ...my vote is "no."  Let the back-end do a simple `deep-key: value` flat mapping.

bgsavage: My concern is that the `api/translations` endpoint returns in a different format than the others, which creates an inconsistency any consumer now has to be aware of.  

Also, I think it leaks the underlying implementation a bit because flattened keys is how we store it in the mysql/ aurora database.  If we were storing in Mongo or server file system, we would be returning an unflattened JSON by default.

wtritch: hmm... I'm not particularly concerned about any implementation exposure.  But the lack of consistency is a little concerning.  I just wonder if it's worth the couple of hours it'll take to find a good unflattening back-end repository and wire it into the Jackson framework as a custom serializer...  Maybe create a nice-to-have jira for it?

PR #200, opened 2017-07-05
Re-enable table column sorting when clicking on a table header with a…

wtritch:
…n info popover.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
@@ -1,5 +1,5 @@
 {{title}}
-<button class="btn btn-link icon-only"
+<button class="btn btn-xs btn-link icon-only"
       popover="{{content}}"
       popoverTitle="{{title}}"
       triggers="focus"
--- a/webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
@@ -1,6 +1,8 @@
 {{title}}
-<button class="btn btn-xs btn-link icon-only"
+<button class="btn btn-icon btn-xs icon-only"
       popover="{{content}}"
       popoverTitle="{{title}}"
-      triggers="focus"
-><i class="fa fa-info-circle"></i></button>
+      triggers="focus">
+  <span class="sr-only">{{'labels.more-info' | translate}}</span>
+  <i class="fa fa-info-circle"></i>
+</button>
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
```diff
@@ -1,5 +1,5 @@
 {{title}}
-<button class="btn btn-link icon-only"
+<button class="btn btn-xs btn-link icon-only"
```
jtreuting: Aaron's example had an "sr-only" tag for screen readers that said "More Info"

<i class="fa fa-info-circle"></i><span class="sr-only">More info</span>

Reference: DWR-526

webapp/src/main/webapp/src/app/assessments/results/information-label.component.ts
```diff
@@ -6,8 +6,8 @@ import { Component, Input } from "@angular/core";
  * an information popover icon.
  */
 @Component({
-  selector: 'info-label',
-  templateUrl: './information-label.component.html',
+  selector: 'info-label,[info-label]',
```
bgsavage: Might be worth adding a comment of when to use which selector.

esotrope: Is this attribute selection strategy so that the markup can select span or div depending on the situation?

wtritch: @esotrope yessir
@bgsavage I'm not sure I want to call out table headers specifically, because then we're bleeding prime-ng datatable knowledge into this component that could be used anywhere.

webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
wtritch: done

webapp/src/main/webapp/src/styles.less
```diff
@@ -242,3 +242,9 @@ body .ui-datatable .ui-datatable-data > tr > td {
 .my-groups-list-container .ui-datatable-tablewrapper {
   border-radius: 0 0 5px 5px;
 }
+
+// Icon-only button styling
+.btn.btn-icon {
+  border-width: 0;
+  background-color: transparent;
+}
```
wtritch: this fixes the styling on the info icons (gets rid of the blue and hover-underline)

bgsavage: May consider asking Aaron to promote this to the sbac-ui-kit.  His vision is we should rarely contribute styles to our own styles.less.

PR #202, opened 2017-07-06
removes now unused gradeId from student context

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/StudentContext.java
```diff
@@ -7,8 +7,6 @@
  */
 public class StudentContext {
 
-    // TODO: remove gradeId when client starts using Exam.gradeId
```
malaffoon: TODONE!

esotrope: 😆

PR #203, opened 2017-07-06
moves common models to common module for use in reports

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/ScaleScores.java
```diff
@@ -0,0 +1,31 @@
...
+import java.sql.SQLException;
+
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
+
+/**
+ * Holds utility methods for mapping SQL query results to {@link ScaleScore}
+ */
+public final class ScaleScores {
+
+    private ScaleScores() {
```
malaffoon: nit (and probably a personal preference, not an absolute guideline): i like "hiding" the private ctor at the bottom of utility classes, just so you don't have to read past it.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/user/DefaultUserContextServiceTest.java
```diff
@@ -2,13 +2,11 @@
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
```
malaffoon: i'm assuming this is unrelated cmd-alt-O cleanup?

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/ScaleScores.java
esotrope: Our intellij autoformat settings like it there though : J

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/user/DefaultUserContextServiceTest.java
esotrope: This is interesting. I don't think I made this change but somehow intellij did when I dragged files around.

PR #204, opened 2017-07-06
Feature/report repo

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamReportRepository.java
```diff
@@ -0,0 +1,11 @@
+package org.opentestsystem.rdw.reporting.common.report;
+
+import org.opentestsystem.rdw.reporting.common.security.PermissionScope;
+
+import javax.validation.constraints.NotNull;
+
+public interface ExamReportRepository<T extends AbstractExamReport> {
+
+    T findOneByStudentIdAndSchoolYear(@NotNull PermissionScope permissionScope, long studentId, int schoolYear);
```
esotrope: It may be sort of an abuse of the spring 1:1 entity:repository pattern but it made sense to create an IcaReportRepo and an IabReportRepo because they can both be created with one query each respectively.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
```diff
@@ -0,0 +1,75 @@
...
+                    }
+                    final Assessment assessment = Assessments.map(row, Assessment.builder(), "asmt_").build();
+                    final Exam exam = Exams.map(row, Exam.builder()).build();
+                    assessments.put(assessment, exam);
+                }
+        );
+
+        return assessments.isEmpty()
+                ? null
+                : reportBuilder.examsByAssessment(assessments.asMap()).build();
```
esotrope: Both report repos return null/no report if the student/year search yields no exams. Is this something we want or do we want to return a report that says the specific student has no exams for the year?

common/src/main/resources/common.sql.yml
```diff
@@ -4,4 +4,130 @@ sql:
...
+      from exam e
+        join asmt a on e.asmt_id=a.id
+        join student st on e.student_id=st.id
+        join school s on e.school_id=s.id
+        join district d on s.district_id=d.id
+      where e.student_id=:student_id
+      and e.school_year=:school_year
+      and a.type_id=2
+      and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+      order by e.completed_at desc
```
esotrope: This ordering is not really necessary but it allows the repo to figure out the most recent school/district/grade of the student. This can be done in the repo layer with added logic in O(n) computation complexity vs. here with O(n log(n)) complexity. Do you guys think this should belong in the repo logic or here?

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamReportRepository.java
malaffoon: All the kerfuffle combining exam types and now they are separate again?  ;)

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/IabReport.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private IabReport() {
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public Map<Assessment, Collection<Exam>> getExamsByAssessment() {
+        return examsByAssessment;
```
malaffoon: This is returning the same collection that was passed to the builder. Probably want to make a copy (or javadoc this behavior to warn about possible abuse).

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/IcaReport.java
```diff
@@ -0,0 +1,43 @@
+package org.opentestsystem.rdw.reporting.common.report.ica;
+
+
+import org.opentestsystem.rdw.reporting.common.model.Assessment;
+import org.opentestsystem.rdw.reporting.common.model.Exam;
+import org.opentestsystem.rdw.reporting.common.report.AbstractExamReport;
+
+public class IcaReport extends AbstractExamReport {
+
+    private Assessment assessment;
```
malaffoon: Why does an IabReport have a collection of assessment/exams, but an IcaReport has a single assessment/exam?   A student can take multiple ICA's per year.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepository.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public IcaReport findOneByStudentIdAndSchoolYear(
+            @NotNull final PermissionScope permissionScope,
+            final long studentId,
+            final int schoolYear) {
+
+        try {
+            return template.queryForObject(
+                    findLatestIcaByStudentIdAndSchoolYearQuery,
```
malaffoon: i guess this technically answers my question but i don't understand why ICA->latest and IAB->all

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/IcaReport.java
jtreuting: The ICA report is a single instance of an ICA report.  Only ever showing one.  But the IAB report lists all IABs across all assessments and previously taken exams of each.

common/src/main/resources/common.sql.yml
malaffoon: I'm sure it won't matter a lot since there're not a lot of records but i try to avoid `ORDER BY` on non-indexed columns; and adding an index just for this doesn't make sense. So, probably, if i were coding this i would have done the sorting in java.

common/src/main/resources/common.sql.yml
```diff
@@ -4,4 +4,130 @@ sql:
...
+      from exam e
+        join asmt a on e.asmt_id=a.id
+        join student st on e.student_id=st.id
+        join school s on e.school_id=s.id
+        join district d on s.district_id=d.id
+      where e.student_id=:student_id
+      and e.school_year=:school_year
+      and a.type_id=1
+      and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+      order by e.completed_at desc
```
malaffoon: Oh, but you need it here because of the limit. Hmm.

common/src/test/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepositoryIT.java
```diff
@@ -0,0 +1,108 @@
...
+        assertThat(repository.findOneByStudentIdAndSchoolYear(districts(0L), -1, 0)).isNull();
+    }
+
+    @Test
+    public void findIcaReportShouldReturnNullForBadSchool() throws Exception {
+        assertThat(repository.findOneByStudentIdAndSchoolYear(schools(0L), -1, 0)).isNull();
+    }
+
+    @Test
+    public void findIvsReportShouldReportForCorrectStudentAndYear() throws Exception {
```
agorina: Nit: the method name has a typo.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamReportRepository.java
esotrope: haha yeah the reports are different formats is why

common/src/main/resources/common.sql.yml
esotrope: i think the column will be indexed - we use it to get latest exam of a group/schoolgrade

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/IabReport.java
esotrope: Good point - will add.

common/src/test/java/org/opentestsystem/rdw/reporting/common/test/support/Instants.java
```diff
@@ -0,0 +1,18 @@
...
+import java.time.ZoneOffset;
+
+/**
+ * Holds utility methods for dealing with instants
+ */
+public final class Instants {
+
+    private Instants(){}
+
+    public static Instant from(final String value) {
```
malaffoon: This method needs some javadocs. I think i'd vote for a different name too, it is clearly doing more than just getting the instant from a string value.

common/src/main/resources/common.sql.yml
agorina: I think having an 'order by' here is good.

common/src/main/resources/common.sql.yml
```diff
@@ -4,4 +4,130 @@ sql:
...
+        e.claim2_scale_score,
+        e.claim2_scale_score_std_err,
+        a.claim3_score_code,
+        e.claim3_category,
+        e.claim3_scale_score,
+        e.claim3_scale_score_std_err,
+        a.claim4_score_code,
+        e.claim4_category,
+        e.claim4_scale_score,
+        e.claim4_scale_score_std_err,
```
agorina: Since you have a separate repo for IAB, do you need to pull all these data that will be NULL for IAB?

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepository.java
jtreuting: The legacy system only allowed the printing of the latest ICA, but see we have a display student history, I believe we will need to support latest as well as print this specific ICA that was taken based on the ID

esotrope: So in the UI will we be prompting the user to select IAB, ICA or specific a ICA? Trying to imagine the UI for that

jtreuting: Let's have an offline discussion and bring in UI/UX for their thoughts.  The original mockups have a menu option when looking at a table row representing a specific exam.  This works fine for ICA's because the report is for a single ICA, so if they choose one they could print it.  But it breaks down for IABs, as they choose any of them and they would get the same report which seems like a weird experience.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/IabReport.java
esotrope: I just realized this will be kind of awkward because the immutable copy would have to maintain the sort order too.

common/src/main/resources/common.sql.yml
esotrope: *removed claim related fields from the select

PR #209, opened 2017-07-07
adds gradeCode fields to Assessment, Exam and StudentEnrollment

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/report/StudentEnrollment.java
```diff
@@ -2,16 +2,15 @@
...
     private String districtName;
     private String stateCode;
 
     private StudentEnrollment(){}
 
-    public int getGradeId() {
-        return gradeId;
+    public String getGradeCode() {
+        return gradeCode;
     }
```
wtritch: Should we "replace" gradeId with gradeCode everywhere?  Or is leaving both properties to prevent front-end breakage during the adjustment?

esotrope: yeah this is a two part refactor. once gradeId isn't used in the ui then we will remove it from the endpoint models

PR #211, opened 2017-07-07
Fix deep linking on filters in student results component.

wtritch:
Small PR to fix an issue I found with deep linking during a functionality discussion earlier today.

### DIFF
--- a/webapp/src/main/webapp/src/app/student/results/student-results.component.ts
+++ b/webapp/src/main/webapp/src/app/student/results/student-results.component.ts
@@ -141,7 +141,7 @@ export class StudentResultsComponent implements OnInit {
 
     //Update the current route without navigating
     this.location.replaceState(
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -141,7 +141,7 @@ export class StudentResultsComponent implements OnInit {
 
     //Update the current route without navigating
     this.location.replaceState(
-      `/students/${this.examHistory.student.ssid}`,
+      `/students/${this.examHistory.student.id}`,
```
wtritch: When I updated the student results page to route on student id I missed the change to deep linking on filter values.

bgsavage: oops :)

PR #212, opened 2017-07-07
Added frontend configuration for iris vendorId and url and consuming …

bgsavage:
…from the front end.


webapp/src/main/resources/application.yml
```diff
@@ -51,6 +51,9 @@ permissionservice:
   component: Reporting
 
 app:
+  iris:
+    vendorId: 2B3C34BF-064C-462A-93EA-41E9E3EB8333
+    url: https://tds-stage.smarterbalanced.org/iris/
```
bgsavage: @malaffoon  new required configurations... Do I need to sync this merge with a config repo merge so I don't break your next qa deploy?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/DefaultUserContextService.java
```diff
@@ -13,19 +14,31 @@
...
+    private String irisVendorUrl;
+
     DefaultUserContextService(final GroupService groupService, final SchoolService schoolService) {
         this.groupService = groupService;
         this.schoolService = schoolService;
     }
 
     @Override
     public UserContext getUserContext(@NotNull final User user) {
+        final UserConfiguration configuration = UserConfiguration.builder()
```
esotrope: This is definitely an abuse of the user* naming. We should probably decide on better naming/object wrapping soon.

webapp/src/main/resources/application.yml
```diff
@@ -51,6 +51,9 @@ permissionservice:
...
   search:
```
esotrope: i dont think "search.schoolsByName.limit" is used anymore - mind removing it?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/UserConfiguration.java
```diff
@@ -0,0 +1,54 @@
+package org.opentestsystem.rdw.reporting.search.user;
+
+class UserConfiguration {
```
malaffoon: I agree with Landon that this doesn't feel like the right name but, regardless, i think class javadocs would help a lot.

webapp/src/main/resources/application.yml
malaffoon: Yeah, we should coordinate this. It isn't good to put any environment-specific values in the main application.yml file. Is the vendorId a sensitive value? We need to get the values out of here somehow. Can the system run without this IRIS info? Are there stub values that can go in here?

malaffoon: Coordination-wise, it is pretty easy when you have new settings: just add to the configuration repo first and commit it.

esotrope: So what is our actual vendor ID? Who makes/provides this and what is it used for? I noticed the iris API doesn't require it

wtritch: In TDS, we just referenced environment variables in the application.yml config if they were environment-specific.
```
app:
  iris:
    vendorId: ${iris.vendorid}
    url: ${iris.url}
```
Then had developers define these properties in their environment:
```
export IRIS_VENDORID='blabityBlahBlah'
export IRIS_URL='moreBlah'
```
That way any new-comers know what needs to be defined and we don't bleed environment knowledge into the source code.

malaffoon: I guess we can do that. I'm not sure i see the advantage over having to set properties on the command-line in the run/debug configuration. But maybe that is my generic phobia of environment variables.

bgsavage: @esotrope RE: vendor ID

All I know about vendorID is what we have on confluence:
 > A value at the end of the URI represents a vendor ID. Both the sample page and Student application use the same value. 

It is required by the IRiS client in order to call `loadToken`

bgsavage: @malaffoon required configs...

I can set empty values for the application.yml if you prefer, though I think these are reasonable defaults I've been using up until this point that are public and available.  

If the values don't exist, Spring will refuse to deploy.

If the values exist but are blank, Spring will deploy but your Item Viewer won't render in the UI.

webapp/src/main/resources/application.yml
bgsavage: I was wondering that, I can remove.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/UserConfiguration.java
bgsavage: I'm open to suggested alternatives.  `FrontEndConfiguratoin`, `AppConfiguration`, etc ?

esotrope: ApplicationSettings? 
ClientContext container that holds ApplicationSettings and UserContext? I don't think its best to have iris/app details in the user context

bgsavage: And this requires renaming the endpoint and controller?

esotrope: Yeah it makes sense to do that if we are putting more than user info in the /user endpoint

bgsavage: Let's parking lot this for Monday then since it's going to incur a larger refactor.  I asked about that this morning about adding to UserContext and it seemed we were fine.

bgsavage: For this PR I'll rename to UserConfiguration -> `AppSettings`

webapp/src/main/resources/application.yml
malaffoon: Okay, Will talked me into it. The advantages:
* there's a visible placeholder in the application.yml file so, even if you don't have instructions, you should be able to figure out what needs setting
* env variables may be used by more than one application
* encouraging the use of env variables doesn't preclude the use of command-line properties

webapp/src/main/resources/application.yml
bgsavage: FYI, It seems `JdbcSchoolRepository` currently uses this config.

PR #214, opened 2017-07-07
removes unused school search by name. removes unused school year cont…

esotrope:
…roller


webapp/src/main/resources/application.yml
```diff
@@ -49,8 +49,3 @@ artifacts:
 
 permissionservice:
   component: Reporting
-
-app:
```
bgsavage: this is probably your conflict.

PR #215, opened 2017-07-08
DWR-543 - Add configurable properties to source application.yml with …

wtritch:
…placeholders.

This adds placeholders to the source application.yml file that can be configured via any spring property configuration method: system properties, environment variables, spring boot properties, etc.

See [confluence](https://confluence.fairwaytech.com/display/SBACDW/RDW_Reporting%3A+Web+App+Local+Setup) for how to configure a development environment.


webapp/src/main/resources/application.yml
```diff
@@ -32,25 +41,33 @@ spring:
...
-    credentials:
-      instance-profile: false
 
 artifacts:
-  root: "classpath:"
+  #Root location of assessment files [s3://some-bucket/ | classpath: | etc]
+  root: ${rdw.resource.root:"classpath:"}
   exam-item:
-    directory: data/irp/Item-{0}/
+    directory: item-content/local/items/Item-{0}/
```
wtritch: I renamed the classpath directory/package to match the development S3 directory path.  It's one less property that needs to be configured on developer machines.

webapp/src/main/resources/application.yml
```diff
@@ -32,25 +41,33 @@ spring:
...
     file-name: item-{0}.xml
 
+application:
+  state: CA
```
wtritch: I think defaulting to a single state is both demonstrative for future maintainers and reasonable for local development.

webapp/src/main/resources/application.yml
```diff
@@ -32,25 +41,33 @@ spring:
...
 permissionservice:
   component: Reporting
+  endpoint: ${permission.service}
 
 app:
   iris:
-    vendorId: 2B3C34BF-064C-462A-93EA-41E9E3EB8333
-    url: https://tds-stage.smarterbalanced.org/iris/
+    vendorId: ${iris.vendorId}
+    url: ${iris.url}
```
wtritch: I moved these values into the "Getting Started" guide on the wiki.

webapp/src/main/resources/application.yml
esotrope: this shouldn't have a default probably

esotrope:
Overall idea here is great - i like being upfront about what props there are to configure. My only issue is the added layer of indirection with the naming.
If I want to override a property in the file I want to use the same key and not a new one.
Also, I liked using the external application.yml for the same reason - it is in the same format as the app configs and requires less context switching

wtritch:
@esotrope I can see your point, but I like the abstraction for shared properties to both allow and prevent conflicts based upon the application/project space.  For instance, "saml.key-store-file" may be a property used by multiple applications, likewise "cloud.aws.access-key" may be used across multiple projects.  I attempted to break it out by naming based upon what I considered to be "reporting-specific" "rdw-specific" and "multi-project-wide." Thoughts?

esotrope:
@wtritch I would see the reuse as a con. What if I want to just change a specific project's props on my local but there are multiple projects on my local configured to use the same env variable, I will have to come up with my own startup -D overrides outside the instruction of the project setup wiki.
I would strongly advocate using an app specific .yml here to avoid colliding with other projects and also to have the same keys and no context switching. Maybe we should parking lot this to get the broader team's feedback

webapp/src/main/resources/application.yml
wtritch: It's one less property that *must* be configured, is easily overridden, isn't tightly-bound to a runtime environment and provides a concrete example value.  Do you have a strong objection to leaving it?

wtritch:
@esotrope sounds like a plan, we can bring it up on Monday =)

webapp/src/main/resources/application.yml
esotrope: Ah yeah - actually i was thinking it would be nice to have the app fail to start if it wasn't set so you couldn't forget to set it to the correct state but that doesn't make much sense now that i think about it. Keeping it here is good i guess

webapp/src/main/resources/application.yml
```diff
@@ -32,25 +41,33 @@ spring:
...
-    credentials:
-      instance-profile: false
 
 artifacts:
-  root: "classpath:"
+  #Root location of assessment files [s3://some-bucket/ | classpath: | etc]
+  root: ${rdw.resource.root:"classpath:"}
   exam-item:
-    directory: data/irp/Item-{0}/
+    directory: local/items/Item-{0}/
```
bgsavage: shouldn't these be env vars too?  The path is different in the s3 location vs. local.

bgsavage:
I'd be curious to hear @malaffoon 's feedback if this makes it easier or harder to dockerize, configure, and deploy to (for example) QA.

webapp/src/main/resources/application.yml
wtritch: I changed them so they didn't have to be.  But again, you *can* override them using env vars if you like, this isn't a restricting change.
Example:
```
export ARTIFACTS_EXAMITEM_DIRECTORY='some/directory/path'
```

webapp/src/main/resources/application.yml
```diff
@@ -4,6 +4,15 @@ management:
...
     enabled: false
 
+saml:
+  # SAML keystore path: file:/some/path/saml.jks
+  key-store-file: ${rdw.reporting.keystore.path}
+  key-store-password: ${rdw.reporting.keystore.password}
+  private-key-entry-alias: ${rdw.reporting.keystore.alias}
+  private-key-entry-password: ${rdw.reporting.keystore.entry.password}
+  idp-metadata-url: ${rdw.reporting.idp}
+  sp-entity-id: ${rdw.reporting.entityId}
```
wtritch: I namespaced the properties under "rdw." and "rdw.reporting"

PR #216, opened 2017-07-08
Feature/student full page report

esotrope:
1. removed now unused pdf gen components
2. moved components to "report" package
3. added /api/students/<id>/examReport endpoint


webapp/src/test/java/org/opentestsystem/rdw/reporting/report/student/StudentExamReportPdfControllerIT.java
```diff
@@ -0,0 +1,78 @@
...
+    @MockBean
+    private StudentExamReportPdfService service;
+
+    @MockBean
+    private ExamReportPdfMessageConverter converter;
+
+    @Test
+    public void getExamReportShouldSend404WhenServiceThrowsException() throws Exception {
+        when(service.getExamReport(any(User.class), any(StudentExamReportPdfRequest.class)))
+                .thenThrow(new NoSuchElementException());
```
esotrope: wasn't able to get mockito refEq to match the passed request for some reason. working on making these tests cover more.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/student/StudentExamReportPdfRequestArgumentResolver.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public Object resolveArgument(
+            final MethodParameter parameter,
+            final ModelAndViewContainer mavContainer,
+            final NativeWebRequest request,
+            final WebDataBinderFactory binderFactory) throws Exception {
+
+        final Map<String, String> pathVariables = getPathVariables(request);
+        return StudentExamReportPdfRequest.builder()
+                .assessmentType(ExamReportAssessmentType.valueOf(request.getParameter("assessmentType")))
```
malaffoon: I admit to waffling on this but I think it is better to put the string parsing logic in the builder. Yeah, that is tainting the builder a bit knowing something about the type of strings coming in but it allows this class to be simpler by separating concerns: just pass the strings into the builder. Then the builder can do whatever machinations are necessary to interpret the strings.

If you really don't like the base builder interpreting the param strings, extending the builder with a ParamReportPdfRequestBuilder that accepts strings keeps things very clean.

webapp/src/test/java/org/opentestsystem/rdw/reporting/report/student/StudentExamReportPdfControllerIT.java
malaffoon: You have three choices here:
1. Leave it as is assuming other tests will cover the transformation of the request object. Not great because i think there is a gap in coverage.
2. Add equals() (and hashCode) to the request object and then use eq() in the when() mock. This is probably the most readable/elegant solution. 
3. Continue to use any() in the when() mock and then do a verify() using an argument captor, checking the passed value. This works and is what i do when i don't have control of the argument class code, but is a bit more obfuscated than #2.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
```diff
@@ -0,0 +1,65 @@
...
+            this.language = language;
+            return (B) this;
+        }
+
+        public B options(final T options) {
+            this.options = options;
+            return (B) this;
+        }
+
+        protected A build(final A search) {
```
bgsavage: Is this a Java pattern for builders with generics where 'B' is the builder for 'A' ?  If not, this might be very confusing... :)  But I'm wondering if I was only confused because I'm not as familiar with Java patterns.  Otherwise I would suggest give your generics more meaningful names such as `TOptions` and `TBuilder`

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamReportOrder.java
```diff
@@ -0,0 +1,7 @@
+package org.opentestsystem.rdw.reporting.common.report;
+
+public enum ExamReportOrder {
+
+    STUDENT_NAME
```
bgsavage: Do you plan on adding more here?

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/student/DefaultStudentExamReportPdfService.java
```diff
@@ -0,0 +1,49 @@
...
+        this.icaReportRepository = icaReportRepository;
+    }
+
+    @Override
+    public ExamReportPdfMessage getExamReport(@NotNull final User user, @NotNull final StudentExamReportPdfRequest request) {
+
+        final PermissionScope permissionScope = user.getPermissionsById().get("INDIVIDUAL_PII_READ").getScope();
+
+        final ExamReportRepository repository = request.getAssessmentType() == ExamReportAssessmentType.IAB
+                ? iabReportRepository : icaReportRepository;
```
bgsavage: nit: I think it's cleaner to give each part of a ternary expression it's own line.
```
? iabRepository
: icaRepository
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/student/StudentExamReportPdfRequestArgumentResolver.java
esotrope: Hmm. wouldn't that just be putting the mapping logic one layer deeper and make it less upfront?
In the past I have always used POST+@RequestBody for all this kind of mapping stuff but spring doesn't like GET+@RequestBody I guess.
What I don't like most about the argument resolver approach is (1) 2 extra files (2) we lose the automagic datatype casting spring usually provides which is what I think you are annoyed with here too.

I'm wondering if we shouldn't just go back to having the mapping in the controller method and then build the objects in the controller method.
That way we can drop the extra classes, get most data type castings and also have a more self documenting API endpoint

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
esotrope: I'm not opposed, I think that would be more readable. I'm not sure what the java convention is here though. Generally I only ever see one letter generic class names so I don't know. Maybe the rest of the team can weigh in here.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamReportOrder.java
esotrope: Yup that's the idea. Not sure what yet - but something for sure : j

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/student/DefaultStudentExamReportPdfService.java
esotrope: I'm good with that - will change

webapp/src/test/java/org/opentestsystem/rdw/reporting/report/student/StudentExamReportPdfControllerIT.java
esotrope: Can I go with option 4 where I strangle mockito a little bit before choosing a working option? : J

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
malaffoon: I'm okay with this pattern. I've done other things in the past but this works. It would be better with some javadocs and, personally, i'd change the name of this method but i don't care that much.

esotrope: I think I'll hold off and then do a sweeping change for all builders on this one

PR #217, opened 2017-07-10
Fix build by providing test property values on required properties an…

wtritch:
…d default values on optional properties.

### DIFF
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -43,8 +43,8 @@ cloud:
   aws:
     credentials:
       #AWS credentials are necessary if reading assessment files from S3
-      accessKey: ${rdw.aws.access.key}
-      secretKey: ${rdw.aws.secret.key}
+      accessKey: ${rdw.aws.access.key:""}
+      secretKey: ${rdw.aws.secret.key:""}
       instance-profile: false
     region:
       auto: false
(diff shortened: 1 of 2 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -43,8 +43,8 @@ cloud:
   aws:
     credentials:
       #AWS credentials are necessary if reading assessment files from S3
-      accessKey: ${rdw.aws.access.key}
-      secretKey: ${rdw.aws.secret.key}
+      accessKey: ${rdw.aws.access.key:""}
+      secretKey: ${rdw.aws.secret.key:""}
```
wtritch: If not using S3 for item information, I believe these can be null/empty.  @malaffoon?

webapp/src/test/resources/application-test.yml
```diff
@@ -9,4 +9,9 @@ permissionservice:
...
 
 application:
-  state: CA
\ No newline at end of file
+  state: CA
+
+app:
+  iris:
+    vendorId: vendorId
+    url: http://my-iris.somewhere.org/iris/
```
wtritch: I added test-only values when a property is required.

webapp/src/main/resources/application.yml
bgsavage: Yes it should serve up your local item xmls then.

webapp/src/test/resources/application-test.yml
bgsavage: It will only "fail fast" if you navigate to the right section in the UI.

PR #219, opened 2017-07-11
Feature/495 google analytics

jtreuting:
Integrated angulartics2 (https://github.com/angulartics/angulartics2)
- Only Google Analytics at this point
- Tracking ID stored in env variable `RDW_REPORTING_ANALYTICS_TRACKINGID` but no harm if not present
- Automatically tracks route changes as page views, but we do a lot of things without a route change so tracking other "important" things via events (like clicking Items by Points Earned, or Filtering, or toggling claim scores, etc.)
- Using attributes on buttons, links, etc. when possible.  Couldn't get change event on dropdowns to work so had to do specific code in the Components to handle that

Known TODOs:
- Wasn't able to get sp-radio-button-list.component to wire up properly, need another set of eyes for that


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
```diff
@@ -29,6 +37,7 @@ public static Builder builder() {
 
         private String irisVendorId;
         private String irisUrl;
+        private String analyticsTrackingId;
```
jtreuting: returned as part of existing UserContext since it is the same type of global as IRIS info.  i'm thinking these 3 should be moved to a separate endpoint for system level config as a nice to have

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -25,11 +26,17 @@ export class AppComponent {
...
     this._userService.getCurrentUser().subscribe(user => {
       this._user = {
         firstName: user.firstName,
         lastName: user.lastName
       };
+
+      if (window['ga'] && user.configuration && user.configuration.analyticsTrackingId) {
+        console.log('Initialize analytics tracking with ID: ' + user.configuration.analyticsTrackingId);
+        window['ga']('create', user.configuration.analyticsTrackingId, 'auto');
+      }
```
jtreuting: The google analytics JS snippet they give you looks like:

```
<script>
      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
          (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
      })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
            ga('create', 'UA-XXXXXX-1', 'auto');
    </script>
```

But in order to get the tracking ID needed to pull the call to the `ga` function out of there and call once we got the info back.  Since it's not defined for angular to know about had to access it as `window['ga']` instead of `window.ga` because it blocked that at compile time.

The rest of the snippet is still in index.html so it pulls in the GA javascript and defines the ga function.

webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -9,7 +9,8 @@
         <!-- Advanced Filters Toggle -->
           <label for="advanced" class="pull-right">{{'labels.groups.results.adv-filters.label' | translate}}</label>
           <button id="advanced" class="form-control btn btn-default pull-right"
-                  (click)="showAdvancedFilters = !showAdvancedFilters">
+                  (click)="showAdvancedFilters = !showAdvancedFilters"
+                  angulartics2On="click" angularticsEvent="ToggleAdvancedFilters" angularticsCategory="AssessmentResults">
```
jtreuting: basic implementation with a category and event name

webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -37,24 +38,29 @@ <h2 class="h3 gray-darkest">
...
+                [ngClass]="{'active': !showOnlyMostRecent}"
+                angulartics2On="click" angularticsEvent="ShowCustomAssessmentSelection" angularticsCategory="AssessmentResults"
+        >{{'labels.groups.results.filter-type-custom' | translate}}
         </button>
       </div>
     </h2>
     <p class="gray-darker" [hidden]="showOnlyMostRecent">{{'labels.groups.results.selected-assessments-instruct' | translate }}</p>
     <div class="flex-children">
       <a *ngFor="let assessment of selectedAssessments" (click)="removeAssessment(assessment)"
+         angulartics2On="click" angularticsEvent="RemoveAssessment" angularticsCategory="AssessmentResults" [angularticsProperties]="{label: assessment.name}"
```
jtreuting: example of providing more info via the GA label.  So I thought it would be nice to see which assessments are actually being used and what older years are being accessed, etc.  so using label for a subset of events where it seems that would be good to know

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -27,17 +29,20 @@
       <!-- Manner of Administration -->
       <div class="block-children">
         <label>{{ translateRoot + 'status.administration' | translate }}</label>
-        <sb-radio-button-list [model]="filterBy" [values]="['SD','NS']" property="administration" enum="enum.administrative-condition" ></sb-radio-button-list>
+        <sb-radio-button-list [model]="filterBy" [values]="['SD','NS']" property="administration" enum="enum.administrative-condition"
+                              analyticsEvent="AddFilter" analyticsCategory="AdvancedFilters"></sb-radio-button-list>
```
jtreuting: this is not currently working (no errors so left it for now).  passing the event and category to the sb-radio-button-list component and then using them there to wire it up

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -127,7 +127,11 @@ export class AssessmentResultsComponent implements OnInit {
   }
 
   get isIca(): boolean {
-    return this._assessmentExam.assessment.isIca
+    return this._assessmentExam.assessment.isIca;
+  }
+
+  get showClaimToggle(): boolean {
+    return this._assessmentExam.assessment.isIca || this._assessmentExam.assessment.isSummative;
```
jtreuting: unrelated, but noticed that the claim toggle should be available for summative as well

webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -68,10 +70,19 @@ export class GroupResultsComponent implements OnInit {
...
     });
+
+    // track change event since wiring select boxes on change as HTML attribute is not possible
+    this.angulartics2.eventTrack.next({
+      action: 'Change' + changeSource,
+      properties: {
+        category: 'AssessmentResults',
+        label: changeSource === 'Group' ? this._currentGroup.id : this._currentSchoolYear
+      }
+    });
```
jtreuting: dropdown wiring didn't work with the attributes, so wiring up in the change event.  passing in the source of the change so I can log which actual event it was and the relevant data selected.  so did the group or the year change and what was it changed to

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
```diff
@@ -7,6 +7,11 @@ import { Utils } from "./Utils";
   (All) [ (Option 1) (Option 2) ]
  */
 
+/*
+  TODO: Fix error and get angulartics working in this component
```
malaffoon: At this point do we want to be creating JIRA's for this kind of thing so it doesn't get lost? Or do we want to create a JIRA to visit all remaining TODO's in the code?

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -14,7 +15,7 @@ export class AppComponent {
     return this._user;
   }
 
-  constructor(public translate: TranslateService, private _userService: UserService) {
+  constructor(public translate: TranslateService, private _userService: UserService, angulartics2GoogleAnalytics: Angulartics2GoogleAnalytics) {
```
bgsavage: Are you still using the analytics service here?  It seems not as no one will have access to it without defining an access modifier by default.

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters-toggle.component.html
```diff
@@ -1,7 +1,8 @@
 <h2 class="h3 gray-darkest">
   <i class="fa fa-check-square"></i> {{'labels.groups.results.adv-filters.label' | translate }}
   <button id="adv-filters-expand-collapse" class="btn btn-default btn-sm pull-right"
-          (click)="displayOptions.expanded = !displayOptions.expanded">
+          (click)="displayOptions.expanded = !displayOptions.expanded"
+          angulartics2On="click" angularticsEvent="ToggleExpandCollapse" angularticsCategory="AdvancedFilters">
```
bgsavage: Their api is a bit verbose, it'd be nice to have a layer of abstraction here (i.e. a directive/component which can shorten this syntax a bit and add the angulartics2 required tags).  This would also give us an easy pivot point if we need to change libraries for whatever reason, since we are using it a lot.

webapp/src/main/webapp/src/app/app.component.ts
jtreuting: At the time I remember that if it wasn't passed in at all, be even though it isn't specifically called, the attributes didn't work.  I'll double check now that it is complete to make sure.  Could have been something else that got fixed since then

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
jtreuting: I was hoping to pair program with someone before the final merge to address.  If that doesn't happen then I'll create a jira for it

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters-toggle.component.html
bgsavage: This also would give us an opportunity to implement some "parent category" so we don't have keep setting the same "AdvancedFilters" category to tags in the same feature area.

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
```diff
@@ -20,6 +25,7 @@ import { Utils } from "./Utils";
         </label>
```
bgsavage: You'd have to update the input template here.

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
```diff
@@ -38,6 +44,12 @@ export class SBRadioButtonComponent {
   @Input()
   public model: any;
 
+  @Input()
+  public analyticsEvent: string;
+
+  @Input()
+  public analyticsCategory: string;
```
bgsavage: This to me feels odd, should the generic controls be concerned with analytics ?  If anything, maybe the control takes a param to accept any attributes / tag that can be placed on the `input`.  I haven't done anything like this in angular4 yet, but I remember in angularJS it required some `$compile` that happened after the attributes were added. 

That being said, it might be easier to just fire the analytic events in code when we're detecting and / or applying the advanced filters rather than on clicking.  We're mostly just interested in what filters they are using, yeah?

webapp/src/main/webapp/src/app/app.component.ts
bgsavage: Try with`ng build --prod` as well, angular doesn't seem to really enforce access until you build `--prod` for some reason...

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -25,11 +26,17 @@ export class AppComponent {
...
   ngOnInit() {
+
     this._userService.getCurrentUser().subscribe(user => {
       this._user = {
         firstName: user.firstName,
         lastName: user.lastName
       };
+
+      if (window['ga'] && user.configuration && user.configuration.analyticsTrackingId) {
+        console.log('Initialize analytics tracking with ID: ' + user.configuration.analyticsTrackingId);
```
wtritch: nit: recommend removing console logs before merging

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
wtritch: Hmmm... I could see that both ways.  To play devil's advocate, I can also see the SBRadioButtonComponent as an abstracted Input, and adding these inputs here make sense, rather than coming up with a generic AnalyticsAwareComponent interface/abstract class.

I can see complexity arising from re-used components however.  The assessments table component is used from both the student/grade and groups routes.  When a control fires an event, should we record which route the user is actually in?  If so, how?  Do we embed routing information in every event?  Do we pass the "current context/route" into each component? Do something else?

webapp/src/main/webapp/src/app/app.component.ts
jtreuting: I verified that if that is not passed into the constructor, no logging happens at all.  it is not used directly when passed in and works fine if it is private or there is no scope modifier and it is local to the constructor.  both work fine.  i'll make it private so it's available outside if that feels "more right" but it doesn't seem to matter either way in practice.

jtreuting: also confirmed that building with `--prod` does not make a difference and it's fine without private as well there

bgsavage: hm, alright.  Maybe just a comment so this doesn't get accidentally "cleaned up" by those with itchy fingers to removed grayed out params?

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters-toggle.component.html
jtreuting: I don't see the need to change libraries as a factor in whether we do a separate directive for it as it is a generic analytics library and we can swap out Google Analytics for another implementation if we want and this code wouldn't change at all.

I'm not sure I see the benefit of it in cutting down what you have to provide.  We could get rid of the `angulartics2On` attribute since i've only seen click work (i tried `update` for a select but that didn't appear to work).  but other than that you do need to pass in the other info it seems.

i'm happy to go with consensus on this though if others think there is a benefit that outweighs the extra code/complexity/abstraction.  @wtritch interested in weighing in?

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
jtreuting: I'm leaning towards using the approach that actually gets it to work since it isn't working right now :) . i'm up for pair programming it to figure out a solution if one of you is up for it later today.

webapp/src/main/webapp/src/app/app.component.ts
jtreuting: yep we were on the same page, added already.

PR #220, opened 2017-07-11
DWR-436 Add a placeholder route for student exam responses.

wtritch:
Add a resolver to fetch student exam responses.
Update student results table menu items to navigate to the selected student/exam responses.
Update breadcrumb component to build crumb links by navigation commands rather than url.

Since the code changes were starting to mount, I decided this was a good place to stop and file an interim PR.

### DIFF
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -43,8 +43,8 @@ cloud:
   aws:
     credentials:
       #AWS credentials are necessary if reading assessment files from S3
-      accessKey: ${rdw.aws.access.key:""}
-      secretKey: ${rdw.aws.secret.key:""}
+      accessKey: ${rdw.aws.access.key:}
+      secretKey: ${rdw.aws.secret.key:}
       instance-profile: false
     region:
       auto: false
--- a/webapp/src/main/webapp/src/app/app.routes.ts
+++ b/webapp/src/main/webapp/src/app/app.routes.ts
@@ -48,8 +48,7 @@ export const routes: Routes = [
         children: [ {
           path: '',
           pathMatch: 'full',
-          component: StudentResultsComponent,
-          canActivate: [ AuthorizeCanActivate ]
+          component: StudentResultsComponent
         }, {
           path: 'exams/:examId',
           pathMatch: 'full',
@@ -60,8 +59,7 @@ export const routes: Routes = [
             },
             permissions: ['INDIVIDUAL_PII_READ']
           },
-          component: StudentResponsesComponent,
-          canActivate: [ AuthorizeCanActivate ]
+          component: StudentResponsesComponent
         } ]
       }
     ]
--- a/webapp/src/main/webapp/src/app/app.routes.ts
+++ b/webapp/src/main/webapp/src/app/app.routes.ts
@@ -9,6 +9,8 @@ import { SchoolResultsComponent } from "./school-grade/results/school-results.co
 import { CurrentSchoolResolve } from "./school-grade/results/current-school.resolve";
 import { StudentResultsComponent } from "./student/results/student-results.component";
 import { StudentExamHistoryResolve } from "./student/results/student-exam-history.resolve";
+import { StudentResponsesResolve } from "./student/responses/student-responses.resolve";
+import { StudentResponsesComponent } from "./student/responses/student-responses.component";
 
 export const routes: Routes = [
   {
@@ -34,7 +36,6 @@ export const routes: Routes = [
       },
       {
         path: 'students/:studentId',
-        pathMatch: 'full',
         resolve: { examHistory: StudentExamHistoryResolve },
         data: {
           breadcrumb: {
@@ -43,8 +44,25 @@ export const routes: Routes = [
           },
           permissions: ['INDIVIDUAL_PII_READ']
         },
-        component: StudentResultsComponent,
-        canActivate: [ AuthorizeCanActivate ]
+        canActivate: [ AuthorizeCanActivate ],
+        children: [ {
+          path: '',
+          pathMatch: 'full',
+          component: StudentResultsComponent,
+          canActivate: [ AuthorizeCanActivate ]
+        }, {
+          path: 'exams/:examId',
+          pathMatch: 'full',
+          resolve: { assessmentItems: StudentResponsesResolve },
+          data: {
+            breadcrumb: {
+              translate: 'labels.student.responses.crumb'
+            },
+            permissions: ['INDIVIDUAL_PII_READ']
+          },
+          component: StudentResponsesComponent,
+          canActivate: [ AuthorizeCanActivate ]
+        } ]
       }
     ]
   }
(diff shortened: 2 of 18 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -43,8 +43,8 @@ cloud:
   aws:
     credentials:
       #AWS credentials are necessary if reading assessment files from S3
-      accessKey: ${rdw.aws.access.key:""}
-      secretKey: ${rdw.aws.secret.key:""}
+      accessKey: ${rdw.aws.access.key:}
+      secretKey: ${rdw.aws.secret.key:}
```
wtritch: unrelated, default the values to an empty string rather than a string of two quote characters.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -43,8 +44,25 @@ export const routes: Routes = [
...
+          resolve: { assessmentItems: StudentResponsesResolve },
+          data: {
+            breadcrumb: {
+              translate: 'labels.student.responses.crumb'
+            },
+            permissions: ['INDIVIDUAL_PII_READ']
+          },
+          component: StudentResponsesComponent,
+          canActivate: [ AuthorizeCanActivate ]
+        } ]
```
wtritch: This adds the child path, so we now have '/students/:studentId' and '/students/:studentId/exams/:examId'

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.html
```diff
@@ -9,10 +9,10 @@
         </a>
       </li>
       <li *ngFor="let crumb of breadcrumbs; let last = last;" [ngClass]="{'active': last }">
-        <a *ngIf="!last" [routerLink]="[crumb.url, crumb.params]">
+        <a *ngIf="!last" [routerLink]="crumb.commands">
           {{ crumb.requiresTranslate? (crumb.label | translate:crumb.translateParams) : crumb.label }}
         </a>
-        <span *ngIf="last" [routerLink]="[crumb.url, crumb.params]">
+        <span *ngIf="last" [routerLink]="crumb.commands">
```
wtritch: I was running into issues trying to merge a url string with a single set of route.params.  I changed the breadcrumb to build an array of routing commands rather than join a single url string to a single set of parameters.

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -34,12 +35,18 @@ export class BreadcrumbsComponent implements OnInit{
...
       let crumbData = child.snapshot.data[BreadcrumbsKeyword];
-      let routeURL: string = child.snapshot.url.map(segment => segment.path).join("/");
-      url += `/${routeURL}`;
+      let route = child.snapshot;
+      let urlSegments: UrlSegment[] = route.url;
+      urlSegments.forEach(segment => {
+        commands.push(segment.path);
+        commands.push(segment.parameters);
+      });
+      let routeCommands = _.clone(commands);
```
wtritch: This will build the router commands array from the current route.  Each Url Segment contains a path string and a parameters object.  We will end up representing: '/groups;schoolYear=2017/students/1/exams/123' as:
```
['groups', { schoolYear: 2017 }, 'students', {}, '1', {}, 'exams', {}, '123', {}]
```
which allows us to keep parameters associated with their route segment.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -128,29 +126,26 @@ export class StudentResultsComponent implements OnInit {
...
+      params.subject = this.filterState.subject;
     }
 
-    //Update the current route without navigating
-    this.location.replaceState(
-      `/students/${this.examHistory.student.id}`,
-      searchParams.toString()
-    );
+    //Update the current route
+    this.router.navigate(['students', this.examHistory.student.id, params]);
```
wtritch: This concerns me a bit.  To provide deep linking consistent with the groups and school year results pages, we were previously updating the browser location.  This provided a deep link to the user without re-fetching data.

Unfortunately, since we rely upon the angular router to provide breadcrumb links, we loose the parameters when navigating to a student responses page and back using the breadcrumb.

Updating the current route causes angular to re-resolve the data, which re-fetches a student's history whenever the user filters on year or subject in the UI.  It's not a major flaw, but it's an extra unnecessary request if the user decides to filter by year or subject.  Thoughts?

webapp/src/main/webapp/src/app/student/results/tables/student-history-ica-summitive-table.component.ts
```diff
@@ -56,8 +59,9 @@ export class StudentHistoryICASummitiveTableComponent {
       let responsesAction: PopupMenuAction = new PopupMenuAction();
       responsesAction.displayName = (() => responsesLabel);
       responsesAction.perform = ((wrapper) => {
-        console.log(`Show Responses: ${wrapper.assessment.name}`);
-      }).bind(this);
+        let examId: number = wrapper.exam.id;
+        this.router.navigate(['exams', examId], { relativeTo: this.route });
+      });
```
wtritch: This navigates the user when they click on the "Bobby's Responses" menu item, preserving the current route history.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -43,8 +44,25 @@ export const routes: Routes = [
...
           permissions: ['INDIVIDUAL_PII_READ']
         },
-        component: StudentResultsComponent,
-        canActivate: [ AuthorizeCanActivate ]
+        canActivate: [ AuthorizeCanActivate ],
+        children: [ {
+          path: '',
+          pathMatch: 'full',
+          component: StudentResultsComponent,
+          canActivate: [ AuthorizeCanActivate ]
```
bgsavage: I wonder if you need the `AuthorizeCanActivate` and both child and parent, or just parent?  Since there's no permissions defined at the level, I'm assuming it's authorizing you twice against the same set of permissions (defined in the parent)

wtritch: Looks like having it on the parent and children does make it fire twice.  And putting it on the parent protects child routes.
Will remove from child routes.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
bgsavage: @jtreuting thoughts?

jtreuting: go back to the table replacement and not do a new page ;)  j/k

i'm not sure i know what the possible remedies are for it other than stating it is a concern because it is pulling the data again.  is a solution to add localStorage caching for something like this so that we have it and don't have to make the call back to the server if it's in local cache?

wtritch: We could possibly add a caching layer, or do something funky like see if there is a previous route with the same parameters that already has data... We could probably find a workaround but it's likely to be a little fugly.  I'd be tempted to just leave this for now and implement short/medium-term caching on the back-end if it becomes an issue.

bgsavage: Oof, I'd hate to have caching solution to this, I think that would just cause more problems.  Right now we're just being consistent with the other results pages where if you change a basic filter, then you hit the network.  Only this time we just don't need to, so your data is refreshed.

We could consider sharing data via services rather than resolves as well in special cases like this.

PR #223, opened 2017-07-12
Feature/standardize grades

wtritch:
This PR standardizes the text display, coloring, and format of grade values across the UI.
Grade 11 will now only be displayed as either 'G11' or 'Grade 11'  and will have a consistent color provided by the new ColorService.

I've updated our mappers/models to use the "gradeCode" api model property rather than the "gradeId" property and removed the deprecated model properties from the back-end models and SQL scripts.


webapp/src/main/webapp/package.json
```diff
@@ -58,7 +58,7 @@
     "protractor": "4.0.9",
     "ts-node": "1.2.1",
     "tslint": "^4.0.0",
-    "typescript": "2.3.4",
+    "typescript": "2.4.1",
```
wtritch: The latest rxjs version fixes the compatibility issues it was having with typescript 2.4, we're now able to bump our typescript version up to the latest.

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -73,7 +73,7 @@ export class AssessmentExamMapper {
     uiModel.id = apiModel.id;
     uiModel.date = apiModel.dateTime;
     uiModel.session = apiModel.sessionId;
-    uiModel.enrolledGrade = apiModel.gradeId;
+    uiModel.enrolledGrade = apiModel.gradeCode;
```
wtritch: I left the UI model properties with the same name, just updated their types to string and values to grade codes.

webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -48,8 +48,8 @@ <h2 class="h3 gray-darkest">
     <p class="gray-darker" [hidden]="showOnlyMostRecent">{{'labels.groups.results.selected-assessments-instruct' | translate }}</p>
     <div class="flex-children">
       <a *ngFor="let assessment of selectedAssessments" (click)="removeAssessment(assessment)"
-         class="tag tag-xs {{ gradeService.getColor(assessment.grade)}}">
-        <span class="label">{{'labels.groups.results.grade-prefix' | translate}}{{assessment.grade}}</span>
+         class="tag tag-xs {{ colorService.getColor(getGradeIdx(assessment.grade)) }}">
+        <span class="label">{{assessment.grade | gradeDisplay}}</span>
```
wtritch: This is an example of using the ColorService to get an indexed color based upon the grade code index.

webapp/src/main/webapp/src/app/assessments/filters/select-assessments/select-assessments.component.ts
```diff
@@ -42,10 +42,11 @@ export class SelectAssessmentsComponent implements OnInit {
   private groupAssessmentsByGrade() {
     let assessmentsByGrade = [];
 
-    for (let grade of this.grades) {
-      let assessments = grade.id == 9
-        ? this._assessments.filter(x => x.grade >= grade.id)
-        : this._assessments.filter(x => x.grade == grade.id);
+    let grades: string[] = _.uniq(this._assessments.map(assessment => assessment.grade))
+      .sort((a, b) => a.localeCompare(b));
```
wtritch: This component was still based on the "3-8, HS" grade breakdown.  I changed to to work against any set of grades provided by the injected assessments.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -104,8 +104,7 @@ <h3 class="gray-darkest mb-md">
           <p-column field="session" header="{{'labels.groups.results.assessment.exams.cols.session' | translate}}" [sortable]="true"></p-column>
           <p-column field="enrolledGrade" header="{{'labels.groups.results.assessment.exams.cols.grade' | translate}}" [sortable]="true" [hidden]="isClaimScoreSelected">
             <ng-template let-exam="rowData" pTemplate="body">
-              <!-- TODO once grade codes are returned, pipe through translation -->
-              {{ exam.enrolledGrade }}
+              {{ exam.enrolledGrade | gradeDisplay }}
```
wtritch: Now that the models are using grade codes, we can translate the grade values in the data tables.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
```diff
@@ -26,7 +26,7 @@ <h2 class="blue-dark h3 mb-md">
       title="{{ (gradesAreUnavailable ? 'labels.school-grade.messages.no-grades-available': '') | translate}}"
       [disabled]="!availableGrades || availableGrades.length == 0" id="current-grade" [(ngModel)]="currentGrade" (change)="updateRoute()"
             class="form-control">
-      <option *ngFor="let grade of availableGrades" [ngValue]="grade">{{grade.code | gradeDisplay }}</option>
+      <option *ngFor="let grade of availableGrades" [ngValue]="grade">{{grade.code | gradeDisplay:'long-name' }}</option>
```
wtritch: After discussing options with @aaronsleeper he preferred to see the long grade name rather than the short grade name in the grade selection dropdowns.

webapp/src/main/webapp/src/app/assessments/assessments.component.html
esotrope: Can we just add the CSS class grade-XYZ and then in the .less put
```css
.grade-XYZ {
  .green
}
```
I would rather we map everything in stylesheets as opposed to .ts business logic.

webapp/src/main/webapp/src/app/shared/assessment-type.pipe.ts
```diff
@@ -4,26 +4,11 @@ import { AssessmentType } from "./enum/assessment-type.enum";
...
 @Pipe({name: 'assessmentType'})
 export class AssessmentTypePipe implements PipeTransform {
 
-  private colorMap: Map<AssessmentType, string> = new Map();
-
-  constructor() {
-    this.colorMap.set(AssessmentType.IAB, 'green-dark');
-    this.colorMap.set(AssessmentType.ICA, 'maroon');
-    this.colorMap.set(AssessmentType.SUMMATIVE, 'blue');
-  }
```
wtritch: I removed color transformations from the assessment type pipe in favor of using the ColorService.

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
esotrope: 👍

webapp/src/main/webapp/src/app/shared/color.service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  private colors: string[] = [
+    'teal',
+    'green',
+    'orange',
+    'blue-dark',
+    'maroon',
+    'green-dark',
+    'blue-dark'
+  ];
```
wtritch: These are the colors that can be used to identify distinct values with a standard palette of choices.  Using a round-robin index-based color mapping we can handle all 19 grades without requiring 19 distinct colors with some minor re-use. (KG will have the same color as 7th Grade)

webapp/src/main/webapp/src/app/shared/enum/grade-code.enum.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * This utility class contains an ordered list of available
+ * grade code values.
+ */
+export class GradeCode {
```
wtritch: I would have rather implemented this as a standard enum.  However, typescript doesn't allow enum value names like '10' or '13'

webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -14,7 +14,7 @@ <h3 class="blue mb-md">{{'labels.student.results.title' | translate}}</h3>
 <!-- Assessments Grouped by Type and Subject -->
 <div *ngFor="let assessmentType of assessmentTypes">
   <div *ngFor="let subject of getSubjectsForType(assessmentType)" class="well">
-    <span class="tag tag-md {{ assessmentType | assessmentType:'color' }}">
+    <span class="tag tag-md {{ colorService.getColor(assessmentType) }}">
```
wtritch: If we have an enum value, we can just pass the value to the color service to get the same color every time.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -34,56 +34,56 @@
         "short-name": "KG"
       },
       "01": {
-        "long-name": "First grade",
-        "short-name": "1"
+        "long-name": "Grade 1",
+        "short-name": "G1"
```
wtritch: After discussing with Brandon, "Grade 1" looked better and matched the mockups, (where "First grade" did not.)  Also, we didn't particularly want to have to create a third naming option for each grade.

webapp/src/main/webapp/src/app/assessments/assessments.component.html
wtritch: The problem with that is that we'd have 19 of them, and we'd be duplicating the colors across all enums (grades, assessment types, etc)

wtritch:
@esotrope np, thanks for leaving those in until the UI could shift =)

webapp/src/main/webapp/src/app/assessments/assessments.component.html
esotrope: Hmm.. that is good separation of concerns though and more descriptive. I would really like to keep styles in stylesheets if we can.
Later if someone wants to re-theme the app the should just have to point to a new stylesheet and not have to change business logic I think.

wtritch: I suppose I could create .item-color-1 through .item-color-7 ?  This would be a generic color wheel for providing colors that look reasonable together and provide reasonable differentiation for any number of items, just a scalar->color transformation.  We can use it for grades, subject, assessment types, multi-series line charts, bar charts, etc, etc

wtritch: After playing with this for a little while and discussing with @aaronsleeper having duplicate names for the color classes is both difficult and increases the compiled css size significantly.  I'm going to leave the class names as color names in here for now.

PR #224, opened 2017-07-12
Fixed Bug 555

bgsavage:
This broke because the translate was now taking longer to resolve now that we are merging ui / api translations.  When the primeng dropdown was being initialized, a placeholder (from translate) hadn't finished resolving so the primeng was selecting the first item in the dropdown rather than using it's placeholder.  Primeng never updated itself once the translation completed.

So now optimized the translate loader by parallelizing the api / ui translation requests and added a translate resolve so that the HomeComponent won't load until translate is complete.

### DIFF
--- a/webapp/src/main/webapp/src/app/app.component.ts
+++ b/webapp/src/main/webapp/src/app/app.component.ts
@@ -1,7 +1,6 @@
 import { Component } from "@angular/core";
 import { TranslateService } from "@ngx-translate/core";
 import { UserService } from "./user/user.service";
-import { RdwTranslateLoader } from "./shared/rdw-translate-loader";
 
 @Component({
   selector: 'app-component',
@@ -10,13 +9,12 @@ import { RdwTranslateLoader } from "./shared/rdw-translate-loader";
 export class AppComponent {
 
   private _user;
-  public loadingTranslations: boolean = true;
 
   get user() {
     return this._user;
   }
 
-  constructor(public translate: TranslateService, private _userService: UserService, private translateLoader: RdwTranslateLoader) {
+  constructor(public translate: TranslateService, private _userService: UserService) {
 
     let languages = [ 'en', 'ja' ];
     let defaultLanguage = languages[ 0 ];
--- a/webapp/src/main/webapp/src/app/app.component.ts
+++ b/webapp/src/main/webapp/src/app/app.component.ts
@@ -1,6 +1,7 @@
 import { Component } from "@angular/core";
 import { TranslateService } from "@ngx-translate/core";
 import { UserService } from "./user/user.service";
+import { RdwTranslateLoader } from "./shared/rdw-translate-loader";
 
 @Component({
   selector: 'app-component',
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -9,12 +10,13 @@ import { UserService } from "./user/user.service";
...
 
   private _user;
+  public loadingTranslations: boolean = true;
 
   get user() {
     return this._user;
   }
 
-  constructor(public translate: TranslateService, private _userService: UserService) {
+  constructor(public translate: TranslateService, private _userService: UserService, private translateLoader: RdwTranslateLoader) {
```
bgsavage: oops these changes can be discarded, was trying to something else for a while.

webapp/src/main/webapp/src/app/shared/common.module.ts
```diff
@@ -41,8 +37,7 @@ export function createTranslateLoader(http: Http) {
     TranslateModule.forRoot({
       loader: {
         provide: TranslateLoader,
-        useFactory: createTranslateLoader,
-        deps: [ Http ]
+        useClass: RdwTranslateLoader
```
bgsavage: unrelated refactor: now useClass instead of a factory.

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -1,25 +1,42 @@
...
-          return Observable.of(merged);
-        }));
+    let uiObservable = this.uiLoader.getTranslation(lang);
+    let apiObservable = this.apiLoader.getTranslation(lang);
+
+    let translateObserver: Observer<any>;
+    this._observable = new Observable<any>(observer => translateObserver = observer).share();
+
+    Observable
+      .forkJoin([uiObservable, apiObservable])
```
bgsavage: forkJoin is the new $q.all

webapp/src/main/webapp/src/app/shared/common.module.ts
esotrope: Oh the factory here was originally to support some compile time optimization angular cli warned about. You don't get a warning with this change?

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
esotrope: Is spoonJoin the new Promise.all?

bgsavage: I think it's dinner time for someone :)

esotrope: 🍽  😋

webapp/src/main/webapp/src/app/shared/common.module.ts
bgsavage: no compile warning here

PR #227, opened 2017-07-12
Bugfix/revert responsive datatable wraper

wtritch:
After playing with the 'table-responsive' datatable wrapper class, @aaronsleeper and I found some non-trivial issues with regards to table header popovers and row menu dropdown display.  Reverting the change for now.


wtritch:
Oops, included some commits in the PR that were unintended.

PR #229, opened 2017-07-12
DWR-436 - Add row expansion with item information tab component

wtritch:
Small modification to the student results table to include the item information tabs as a row expansion.

### DIFF
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
@@ -27,7 +27,9 @@
   <p-dataTable [value]="assessmentItems"
                emptyMessage="{{'messages.noExamsMatchingCriteria' | translate}}"
                sortField="assessmentItem.scores.0.position"
-               tableStyleClass="table table-striped table-hover">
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -55,6 +57,9 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
         {{ correctness | number:'1.2-2' }}
       </ng-template>
     </p-column>
+    <ng-template let-item pTemplate="rowexpansion">
+      <item-tab [item]="item.assessmentItem" [showItemDetails]="assessment.isInterim"></item-tab>
+    </ng-template>
```
wtritch: That was a lot simpler than I thought it was going to be.

bgsavage: well, you still have to pass in and code for the student responses :D  (or someone does)

wtritch: hmmm good point.  I'll abort this PR and get those in there.

PR #230, opened 2017-07-13
Feature/proxy iris

bgsavage:
Branch name is a misnomer at this point, the real fix here is to use setTimeout(function(),0) to give the new Iris more time to finish loading.

Note:  A qa config change will need to be merged before this first.


webapp/src/main/webapp/src/scripts/iris/util_xdm.js
```diff
@@ -111,12 +111,13 @@ Util.XDM = (function ($) {
   // recieve a request/response
   function messageHandler(evt) {
 
-    // console.log('MESSAGE RECIEVED', evt);
+    if (!evt.data
+      || typeof evt.data != 'string'
+      || evt.data.toUpperCase().indexOf("IRIS") < 0) {
+      return;
+    }
```
esotrope: Is this file something we should be editing here? I think eventually we should be sourcing this from iris. I think there is a URL on the iris server we can hit to get this JS file even.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
```diff
@@ -43,26 +44,36 @@ export class ItemViewerComponent implements OnInit {
...
-    })
+    });
   }
 
   irisframeOnLoad() {
     IRiS.setFrame(this._irisFrame);
+    this.irisIsLoading = false;
+
+    // This will fire after iris finishes running it's setup js.
+    setTimeout(this.loadToken.bind(this), 0);
```
bgsavage: This is the real fix.  Not ideal, but the best we can do for the time being.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
```diff
@@ -43,26 +44,36 @@ export class ItemViewerComponent implements OnInit {
...
-    IRiS.loadToken(this.vendorId, token);
-    this.irisIsLoading = false;
+    IRiS.loadToken(this.vendorId, token)
+      .fail((function(err) {
+        if (this._currentAttempt-- > 0) {
+          console.log("Failed to load token, attempting again", err);
+          setTimeout(this.loadToken.bind(this), 2000);
+        } else {
+          console.log("Max failures attempted, aborting.");
+        }
```
bgsavage: Retry logic here, thanks @wtritch

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
```diff
@@ -43,26 +44,36 @@ export class ItemViewerComponent implements OnInit {
...
+      }).bind(this));
   }
 
   private getToken(bankItemKey){
-    return `{"passage":{"autoLoad":"false"},"items":[{"id":"I-${bankItemKey}"}],"layout":"WAI"}`;
-
-    // return `{"items":[{"id":"I-${bankItemKey}"}], "accommodations": []}`;
-    // return `{"items":[{"response":"","id":"I-187-1437"}],"accommodations":[]}`
+    return `{"items":[{"response":"","id":"I-${bankItemKey}"}], "accommodations": []}`;
```
bgsavage: This token format breaks against the old iris, so we shouldn't merge until after the config is merged.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -174,7 +174,7 @@ <h3 class="gray-darkest mb-md">
             <i class="fa fa-chevron-left"></i> {{'labels.groups.results.assessment.button-exams' | translate}}
           </a>
         </h3>
-        <p-dataTable [value]="filteredAssessmentItems" tableStyleClass="table table-striped table-hover" [tableStyle]="{'table-layout':'auto'}" expandableRows="true">
+        <p-dataTable [value]="filteredAssessmentItems" tableStyleClass="table table-striped table-hover" [tableStyle]="{'table-layout':'auto'}" expandableRows="true" rowExpandMode="single">
```
bgsavage: Now only allowing one row to be expanded at a time.

webapp/src/main/webapp/src/scripts/iris/util_xdm.js
bgsavage: Only allow iris events to come through.  Otherwise we start receiving other messages such as from webpack.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
esotrope: Should we parametrize maxTimeoutAttempts, baseTimeout and backoff multiplier?

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
esotrope: @jtreuting are we expected to send the available exam accommodations over to IRIS for our item views?

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
jtreuting: i believe the solution is the setTimeout(..., 0) and not the retry logic

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
jtreuting: I'm guessing that if this fails there is no message shown to the user and they just see the spinning loading?  can we have the spinner stop and display an error message in the div instead

webapp/src/main/webapp/src/scripts/iris/util_xdm.js
bgsavage: Ideally we would be integrating with IVS instead of iris, so we wouldn't need this file at all.  For the time being though, we need to limit the messageHandler to not get errors from non-iris messages firing such as webpack.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
bgsavage: I can't stop the iris spinner but I can show a message above the iris iframe.

esotrope: Can we just remove/hide the iframe once we give up on it?

jtreuting: ahh so the spinner is in the iframe .. yeah we should probably show a message and hide/remove the iframe seems appropriate

webapp/src/main/webapp/src/scripts/iris/util_xdm.js
```diff
@@ -111,12 +111,13 @@ Util.XDM = (function ($) {
   // recieve a request/response
```
malaffoon: can you fix the spelling?

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
bgsavage: well there's two spinners, the "Loading" spinner is ours, we switch to the iris one once the iframe is loaded.

I can remove the iframe once we give up.

bgsavage: fyi, I'll be taking error handling in a different PR.

PR #233, opened 2017-07-14
Removed Print option from the exam row context menu as that is moving

jtreuting:
Removed Instrucitonal Resource menu option from My Groups and School Grade
Added Student's Test History as an option and wired it up for the My Groupr and School Grade

### DIFF
--- a/webapp/src/main/webapp/src/app/app.routes.ts
+++ b/webapp/src/main/webapp/src/app/app.routes.ts
@@ -15,7 +15,8 @@ import { TranslateResolve } from "./home/translate.resolve";
 import { StudentHistoryResponsesExamResolve } from "./student/responses/student-history-responses-exam.resolve";
 import { StudentHistoryResponsesAssessmentResolve } from "./student/responses/student-history-responses-assessment.resolve";
 
-const studentTestHistoryRoute = {
+
+const studentTestHistoryChildRoute = {
   path: 'students/:studentId',
   resolve: { examHistory: StudentExamHistoryResolve },
   data: {
@@ -23,7 +24,6 @@ const studentTestHistoryRoute = {
       translate: 'labels.student.results.crumb',
       translateResolve: 'examHistory.student'
     },
-    permissions: ['INDIVIDUAL_PII_READ']
   },
   canActivate: [ AuthorizeCanActivate ],
   children: [ {
@@ -41,8 +41,7 @@ const studentTestHistoryRoute = {
     data: {
       breadcrumb: {
         translate: 'labels.student.responses.crumb'
-      },
-      permissions: ['INDIVIDUAL_PII_READ']
+      }
     },
     component: StudentResponsesComponent
   } ]
@@ -64,7 +63,7 @@ export const routes: Routes = [
           resolve: { assessment: GroupAssessmentsResolve },
           component: GroupResultsComponent
         },
-        studentTestHistoryRoute
+        studentTestHistoryChildRoute
         ]
       },
       {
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -280,21 +282,15 @@ export class AssessmentResultsComponent implements OnInit {
...
     }).bind(this);
-    actions.push(resourcesAction);
-
-    let reportLabel: string = this.translateService.instant('labels.menus.print-report');
-    let reportAction: PopupMenuAction = new PopupMenuAction();
-    reportAction.displayName = (() => reportLabel);
-    reportAction.perform = ((exam: Exam) => {
-      console.log(`Print Report: ${exam.student.lastName}`)
+    responsesAction.perform = ((exam: Exam) => {
+      this.router.navigateByUrl(`/students/${exam.student.id}`);
```
jtreuting: This is not preserving the breadcrumb, so if you come from My Groups it replaces the breadcrumb with just the student test history as if you went directly there.  Any thoughts /guidance on getting that in place?  I feel like that is done somewhere but having a hard time finding it.

@wtritch, @bgsavage ?

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -15,6 +15,39 @@ import { TranslateResolve } from "./home/translate.resolve";
...
 
+const studentTestHistoryRoute = {
+  path: 'students/:studentId',
+  resolve: { examHistory: StudentExamHistoryResolve },
+  data: {
+    breadcrumb: {
+      translate: 'labels.student.results.crumb',
+      translateResolve: 'examHistory.student'
+    },
+    permissions: ['INDIVIDUAL_PII_READ']
```
wtritch: This is a little strange.  You need GROUP_PII_READ permissions to view the group results, which now includes a menu option to navigate to a path that requires INDIVIDUAL_PII_READ permissions.  I'm not very familiar with the security sub-system, is there an easy way in code to not append that menu option if the user doesn't have individual read writes?

wtritch: Hmmm... it also looks like that's the only row action.  We should probably just hide the menu column if the menu has no registered actions.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
wtritch: Looks like you found it, but for posterity we're appending a route to the current route: `router.navigate([...], {relativeTo: this.route});`

jtreuting: hmm, i know i found that and used it, now I'm getting my PRs mixed up I think cause I know I switched to that method.  let me figure that out

webapp/src/main/webapp/src/app/app.routes.ts
jtreuting: i copied the existing route and pulled it into a constant so it could be reused as child routes.

we should talk this through a bit more, because if you have access to a group you have access to the students in that group.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
jtreuting: looks like this comment is on an outdated version of the code.  i had first pushed this without the child route working and did a later commit last night to fix it.  not sure how you saw the origi version to comment on it.  but if you look at the final version it is using the proper syntax, otherwise that route wouldn't work as expected

webapp/src/main/webapp/src/app/app.routes.ts
wtritch: Ah, I had questions about that after the review yesterday, where there were concerns about displaying PII within the group results.  

Sounds like we just may not be able to re-use the route definition in its entirety.  Maybe just re-use the "children" property, and remove the 'data.permissions' property from the '/exams/:examId' route.  (I don't believe it's used since we're only authorizing the root path.)

PR #234, opened 2017-07-14
Remove session options from School/Grade results page

jtreuting:
Do not display session toggles for School/Grade, always show all results since sessions don't make sense at that level according to SBAC.  Since this is a shared component for Groups and School/Grade, each sets an input boolean to represent whether to use session filters or not.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.html
@@ -105,6 +105,7 @@
   <assessment-results [assessmentExam]="assessment"
                       [showValuesAsPercent]="showValuesAsPercent"
                       [filterBy]="clientFilterBy"
+                      [useSessions]="useSessions"
                       [loadAssessmentItems]="boundLoadAssessmentItems"></assessment-results>
 </div>
 <!-- No Results Error -->
--- a/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
@@ -56,10 +56,14 @@ export class AssessmentResultsComponent implements OnInit {
   @Input()
   set assessmentExam(assessment: AssessmentExam) {
     this._assessmentExam = assessment;
-    this.sessions = this.getDistinctExamSessions(assessment.exams);
 
-    if (this.sessions.length > 0) {
-      this.toggleSession(this.sessions[ 0 ]);
+    // if we aren't going to display the sessions, don't waste resources computing them
+    if (this.useSessions) {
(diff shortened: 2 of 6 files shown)


webapp/src/main/webapp/src/app/groups/results/group-results.component.html
```diff
@@ -3,7 +3,7 @@ <h2 class="blue-dark h3 mb-md">
   {{'labels.groups.name' | translate}}
 </h2>
 
-<assessments [assessmentExams]="assessmentExams" [assessmentProvider]="assessmentProvider">
+<assessments [assessmentExams]="assessmentExams" [assessmentProvider]="assessmentProvider" [useSessions]="true">
```
wtritch: Nit: I don't think you need the brackets around [useSessions] since you're passing a static value...

jtreuting: without it, it treats it as a string instead of a boolean.  surprising to me because the @Input() useSessions is defined as a boolean.  but logging it out it shows a a type of string and then the checks later on fail.

webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -105,6 +105,7 @@ <h2 class="blue-dark h3 mt-lg mb-md">
   <assessment-results [assessmentExam]="assessment"
                       [showValuesAsPercent]="showValuesAsPercent"
                       [filterBy]="clientFilterBy"
+                      [useSessions]="useSessions"
```
bgsavage: nit: clarify the name of what it means to "use sessions".  maybe something like `allowFilterBySessions` ?

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -56,10 +56,14 @@ export class AssessmentResultsComponent implements OnInit {
   @Input()
   set assessmentExam(assessment: AssessmentExam) {
     this._assessmentExam = assessment;
-    this.sessions = this.getDistinctExamSessions(assessment.exams);
 
-    if (this.sessions.length > 0) {
-      this.toggleSession(this.sessions[ 0 ]);
+    // if we aren't going to display the sessions, don't waste resources computing them
+    if (this.useSessions) {
```
bgsavage: I'm curious since this is being done in setters, is it possible that `assessmentExam` can be set before `useSessions` is?

webapp/src/main/webapp/src/app/assessments/assessments.component.html
jtreuting: i like that name better.  i'll change that.

PR #235, opened 2017-07-14
Disable the Interim Only or Summative Only advanced filter based on i…

jtreuting:
…f there is at least one interim or summative assessment in the current view

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -132,6 +132,8 @@ export class AssessmentsComponent implements OnInit {
   ngOnInit() {
     this.filterOptionService.getExamFilterOptions().subscribe(filterOptions => {
       this.filterOptions = filterOptions;
+
+      this.updateFilterOptions();
     });
 
     this.boundLoadAssessmentItems = this.loadAssessmentItems.bind(this);
@@ -167,6 +169,8 @@ export class AssessmentsComponent implements OnInit {
     if (latestAssessment) {
       this._assessmentExams.push(latestAssessment);
     }
+
+    this.updateFilterOptions();
   }
 
   removeAssessment(assessment: Assessment) {
@@ -181,6 +185,13 @@ export class AssessmentsComponent implements OnInit {
     else {
       this.removeUnselectedAssessmentExams();
     }
+
+    this.updateFilterOptions();
+  }
+
+  private updateFilterOptions() {
+    this.filterOptions.hasInterim = this.selectedAssessments.some(a => a.isInterim);
+    this.filterOptions.hasSummative = this.selectedAssessments.some(a => a.isSummative);
   }
 
   private loadAssessmentItems(assessmentId: number): Observable<AssessmentItem[]> {
--- a/webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
@@ -28,16 +28,16 @@
     <div class="block-children">
       <!-- Manner of Administration -->
       <div class="block-children">
-        <label>{{ translateRoot + 'status.administration' | translate }}</label>
+        <label [ngClass]="{'text-muted': !filterOptions.hasInterim}">{{ translateRoot + 'status.administration' | translate }}</label>
         <sb-radio-button-list [model]="filterBy" [values]="['SD','NS']" property="administration" enum="enum.administrative-condition"
-                              label="{{ translateRoot + 'status.label' | translate }}"
+                              label="{{ translateRoot + 'status.label' | translate }}" [disabled]="!filterOptions.hasInterim"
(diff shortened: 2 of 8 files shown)


webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -28,16 +28,16 @@
     <div class="block-children">
       <!-- Manner of Administration -->
       <div class="block-children">
-        <label>{{ translateRoot + 'status.administration' | translate }}</label>
+        <label [ngClass]="{'text-muted': !filterOptions.hasInterim}">{{ translateRoot + 'status.administration' | translate }}</label>
```
jtreuting: using text-muted cause that appears to be a standard bootstrap thing, but currently the color is very very similar so it is hard to tell the label is a slightly different color.

wtritch: @aaronsleeper thoughts?

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -181,6 +185,13 @@ export class AssessmentsComponent implements OnInit {
...
       this.removeUnselectedAssessmentExams();
     }
+
+    this.updateFilterOptions();
+  }
+
+  private updateFilterOptions() {
+    this.filterOptions.hasInterim = this.selectedAssessments.some(a => a.isInterim);
+    this.filterOptions.hasSummative = this.selectedAssessments.some(a => a.isSummative);
   }
```
wtritch: We should probably do the same thing with the student history advanced filters.  I twinge a bit at exposing/requiring individual filter knowledge outside of the filter component, but I don't have a better solution.  The only alternative I can think of is to pass a collection of Assessment models into the filter component to let it configure itself.  If we start enabling/disabling/adding/removing more options maybe we can re-examine.

jtreuting: yeah i went back and forth on that one as well.  i felt like it was a small step that didn't push the boundaries too much :) but yeah if someone has a strong feeling we could make bigger changes, but I'm leaning towards leaving as it is right now

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
aaronsleeper: we can modify .text-muted. you're correct @jtreuting it is a standard Bootstrap class. @wtritch can you send me a screenshot?

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
jtreuting: oh and I'll get the change in place for the student history filters as well

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
jtreuting: i can do that in a bit.  thanks

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
wtritch: SGTM

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
```diff
@@ -10,14 +10,14 @@ import { Utils } from "./Utils";
   selector: 'sb-radio-button-list',
   template: `
     <div class="nested-btn-group btn-group-sm toggle-group" data-toggle="buttons">
-      <label class="btn btn-primary" [ngClass]="{'active': model[property] == -1 }">
-        <input [value]="-1" [(ngModel)]="model[property]" type="radio" [name]="name" 
+      <label class="btn btn-primary" [ngClass]="{'active': model[property] == -1, 'disabled': disabled }">
+        <input [value]="-1" [(ngModel)]="model[property]" type="radio" [name]="name" [attr.disabled]="disabled?'disabled':null"
```
bgsavage: What is `attr` here?

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
```diff
@@ -41,6 +41,10 @@ export class SBRadioButtonComponent {
   @Input()
   public model: any;
 
+  // When true, disables the buttons so they can not be changed
+  @Input()
+  public disabled: boolean = false;
```
bgsavage: Might be nice to have a generic attrs parameter so we don't have to keep adding props to this component every time we want to add new attrs (similar to the analytics below).

webapp/src/main/webapp/src/app/shared/sb-radio-button-list.component.ts
jtreuting: `attr` as a prefix is a way of setting an HTML attribute based on some boolean passed in.

so it will add `disabled="disabled"` to disable the input.  I don't think I had luck using `[disabled]` directly.

PR #236, opened 2017-07-14
Feature/add student responses

wtritch:


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItems.java
```diff
@@ -11,11 +12,24 @@
...
+        return builder.build();
+    }
+
+    private static boolean hasColumn(final ResultSet row, final String columnName) throws SQLException {
+        final ResultSetMetaData metaData = row.getMetaData();
+        for (int i = 1; i <= metaData.getColumnCount(); i++) {
+            final String name = metaData.getColumnName(i);
+            if (columnName.equals(name)) return true;
+        }
+        return false;
```
wtritch: This kind of weirds me out.  Depending upon how the ExamItems are being requested we may or may not have a "response" column in the ResultSet.  Thoughts? @esotrope

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.html
```diff
@@ -1,13 +1,16 @@
-<tabset>
+<tabset #itemTabs>
   <!-- TODO: This tab should be toggleable depending on the context -->
-  <tab heading="{{translateRoot + 'student-scores' | translate }}">
+  <tab *ngIf="showStudentScores" heading="{{translateRoot + 'student-scores' | translate }}">
```
wtritch: The "Student Scores" tab makes no sense when viewing responses of a single student.

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.html
```diff
@@ -1,13 +1,16 @@
...
       <!--TODO -->
     </div>
   </tab>
   <tab *ngIf="showItemDetails" (select)="loadItemViewer=true" heading="{{translateRoot + 'item-viewer' | translate }}">
     <div class="tab-content well">
-      <item-viewer *ngIf="loadItemViewer" [bankItemKey]="item.bankItemKey"></item-viewer>
+      <item-viewer *ngIf="loadItemViewer"
+                   [bankItemKey]="item.bankItemKey"
+                   [response]="response"
+                   [position]="position"></item-viewer>
```
wtritch: I added the ability to inject both a response and "position" that will be displayed as the item label so that the item appears as it did on the student's exam.

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
```diff
@@ -34,9 +54,19 @@ export class ItemTabComponent implements OnInit {
...
-  ngOnInit() {
+  ngOnInit(): void {
+
+    //Unfortunate hack to "select" the initial tab
+    //The timeout is required to give the TabComponent time to process
+    //its template-defined TabDirectives.
+    setTimeout((function() {
+      let tab: TabDirective = this.itemTabs.tabs[0];
+      tab.select.emit(tab);
+    }).bind(this), 0);
```
wtritch: This is a little unfortunate.  The ngx-bootstrap TabComponent doesn't fire a "select" event when displaying the first tab in the tabset.  That was causing the item-viewer tab to not initially display the iris frame even though it's tab was selected by default.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
```diff
@@ -60,9 +66,15 @@ export class ItemViewerComponent implements OnInit {
...
 
+  /**
+   * Send a request to load the specified assessment item.
+   */
   loadToken() {
     let token = this.getToken(this.bankItemKey);
     IRiS.loadToken(this.vendorId, token)
+      .done((function() {
+        this.loadResponse();
+      }).bind(this))
```
wtritch: After loading the token (which fetches and renders the assessment item in the Iris frame) we fire another XDM request to both populate the student's response in the assessment item and set the item's label.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
```diff
@@ -73,7 +85,35 @@ export class ItemViewerComponent implements OnInit {
...
+    IRiS.setResponses(response);
+  }
+
+  private getToken(bankItemKey: string): string {
+    return JSON.stringify({
+      items: [{
+        id: `I-${bankItemKey}`
+      }]
+    });
+  }
```
wtritch: I don't generally like hand-jamming JSON.  Even simple json is finicky and it's easy to make a typo. (Which I did regularly while trying to initially set the label through the token request.)

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
```diff
@@ -73,7 +85,35 @@ export class ItemViewerComponent implements OnInit {
...
+
+    if (response && response.length > 0) {
+      payload.response = response;
+    }
+
+    if (position && position > 0) {
+      payload.label = position.toString();
+    }
+
+    return [payload];
```
wtritch: This sets the student response and item label, if they're defined.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -104,6 +104,13 @@ <h3 class="gray-darkest mb-md">
         </h3>
 
         <p-dataTable [value]="exams" emptyMessage="{{'labels.groups.results.assessment.exams.empty-message' | translate}}" sortField="student.lastName" tableStyleClass="table table-striped table-hover" expandableRows="true">
+          <!-- Row Menu -->
+          <p-column styleClass="cell-dropdown">
+            <ng-template let-data="rowData" pTemplate="body">
+              <popup-menu [item]="data" [actions]="actions"></popup-menu>
+            </ng-template>
+          </p-column>
```
wtritch: Semi-related: After discussing with @aaronsleeper, I moved the row menu buttons to the left-hand side and styled them more like buttons since developers and QA were missing the unstyled vertical ellipsis on the right-hand side of the table.

webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
```diff
@@ -1,5 +1,5 @@
 {{title}}
-<button class="btn btn-icon btn-xs icon-only"
+<button class="btn btn-borderless btn-xs icon-only"
```
wtritch: After discussing with @aaronsleeper I replaced all usages of my hand-jammed "btn-icon" class with style-kit-sanctioned "btn-borderless" =)

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
jtreuting: Did you mean to remove the accommodations empty array?  I think @bgsavage ran into issues when that was not provided.

wtritch: Yup, it's no longer required.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItems.java
esotrope: How we've been solving that is having the mapper just return the builder with all the basic fields mapped and then the repo using the mapper maps the rest of the non-basic fields. That way there is no hasColumn check.
There is definitely room for improvement in this strategy though.

What you have here is nice because it is flexible, but also not nice in that every ExamItems.map call does the check. 

This is kind of the ongoing design challenge of this app.
Our agreed approach when we started was feature-based object modeling which implies limited reuse of models and mappers. Mid-project though I moved to reusing the exam & assessment mappers because they have some logic there that would be annoying to duplicated everywhere. Which is how we've wound up where we are.

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
bgsavage: This only because I have that logic where I don't want IRIS to load if it's not selected.  Couldn't you just update that logic to check if it's the first tab as well (and therefore would be loaded by default)?

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
bgsavage: I think you can pass the response as part of the original loadToken request?

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -55,6 +59,13 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
         {{ correctness | number:'1.2-2' }}
       </ng-template>
     </p-column>
+    <ng-template let-item pTemplate="rowexpansion">
+      <item-tab [item]="item.assessmentItem"
+                [response]="item.assessmentItem.scores[0].response"
+                [position]="item.assessmentItem.scores[0].position"
```
bgsavage: Can we reduce the parameters by composing / modeling related data?

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
wtritch: I think only the tabset knows what the first tab is.  I don't think individual tabs know their own positions, do they?

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
wtritch: I believe it's part of the API, but it wasn't working for me...  Plus, we cannot send the label as part of the original loadToken request, and there is no "setLabel" method in the API

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
wtritch: We could create a "[data]" input, with a wrapper model, but since response and position are optional I wasn't sure I wanted to create an extra model for the complexity.  Thoughts?

PR #239, opened 2017-07-14
Event bind to the keyup to detect when the SSID field is changing

jtreuting:
remove error message if the user has altered the input

https://jira.fairwaytech.com/browse/DWR-518

Clears out the error message if the user attempts to correct the input.

Thoughts?

### DIFF
--- a/webapp/src/main/webapp/src/app/student/student.component.html
+++ b/webapp/src/main/webapp/src/app/student/student.component.html
@@ -3,12 +3,13 @@
     <div class="flex-children">
       <div class="flex-child grow">
         <label for="student-id-input">{{'labels.student.title' | translate}}</label>
-        <input formControlName="ssid"
+        <input #ssid formControlName="ssid"
                type="text"
                class="form-control"
                id="student-id-input"
                required
-               placeholder="{{'labels.student.prompt' | translate}}">
+               placeholder="{{'labels.student.prompt' | translate}}"
+               (keyup)="onSsidChange(ssid.value)">
       </div>
       <div class="flex-child">
         <button type="submit" class="btn btn-primary" [disabled]="searchForm.invalid">{{'buttons.search' | translate}}</button>
--- a/webapp/src/main/webapp/src/app/student/student.component.html
+++ b/webapp/src/main/webapp/src/app/student/student.component.html
@@ -3,13 +3,13 @@
     <div class="flex-children">
       <div class="flex-child grow">
         <label for="student-id-input">{{'labels.student.title' | translate}}</label>
-        <input #ssid formControlName="ssid"
+        <input formControlName="ssid"
                type="text"
                class="form-control"
                id="student-id-input"
                required
                placeholder="{{'labels.student.prompt' | translate}}"
-               (keyup)="onSsidChange(ssid.value)">
+               (keyup)="onSsidChange()">
       </div>
       <div class="flex-child">
         <button type="submit" class="btn btn-primary" [disabled]="searchForm.invalid">{{'buttons.search' | translate}}</button>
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/student/student.component.html
```diff
@@ -3,12 +3,13 @@
...
         <label for="student-id-input">{{'labels.student.title' | translate}}</label>
-        <input formControlName="ssid"
+        <input #ssid formControlName="ssid"
                type="text"
                class="form-control"
                id="student-id-input"
                required
-               placeholder="{{'labels.student.prompt' | translate}}">
+               placeholder="{{'labels.student.prompt' | translate}}"
+               (keyup)="onSsidChange(ssid.value)">
```
wtritch: Do we have to bind to keypress rather than (change) ?

webapp/src/main/webapp/src/app/student/student.component.html
```diff
@@ -3,12 +3,13 @@
     <div class="flex-children">
       <div class="flex-child grow">
         <label for="student-id-input">{{'labels.student.title' | translate}}</label>
-        <input formControlName="ssid"
+        <input #ssid formControlName="ssid"
```
wtritch: Do we need the #ssid identifier?

webapp/src/main/webapp/src/app/student/student.component.ts
```diff
@@ -26,8 +26,12 @@ export class StudentComponent implements OnInit {
     return this.searchForm.controls["ssid"];
   }
 
+  onSsidChange(ssidValue: string): void {
+    this.studentNotFound = false;
+  }
```
wtritch: nit: jsdocs =)

webapp/src/main/webapp/src/app/student/student.component.ts
```diff
@@ -26,8 +26,12 @@ export class StudentComponent implements OnInit {
...
+
   /**
-   * Search for the entered sudent SSID.  If a student exists with exams,
+   * Search for the entered student SSID.  If a student exists with exams,
```
wtritch: oops.  Downside of jsdocs: more room for spellign erros.

webapp/src/main/webapp/src/app/student/student.component.html
```diff
@@ -8,7 +8,8 @@
                class="form-control"
                id="student-id-input"
                required
-               placeholder="{{'labels.student.prompt' | translate}}">
+               placeholder="{{'labels.student.prompt' | translate}}"
+               (keyup)="onSsidChange()">
```
bgsavage: Does this still work if user uses the enter key to submit?

bgsavage: Might be better to use a model changed event instead.

jtreuting: yes enter key still works ... i'm fine switching to the ngModal and ngModelChange to get the same logic, but it seems like 2 attributes instead of one for the same behavior.  not sure if ngModalChange tracks the keyup JS event change or not or how it is implemented.

bgsavage: What if user is only using the mouse to copy and paste a student id?

jtreuting: then their error message will stay on the page until they get moved to the next page by clicking search with a valid one :) 

i'm happy moving it to ngModel and ngModelChange as I said before, it felt like those are not a favored way of doing things but i might be wrong on that.  i really don't have a strong opinion about it

bgsavage: I don't think you should use ngModel / ngModelChange here since we are using a [model driven form](https://blog.thoughtram.io/angular/2016/06/22/model-driven-forms-in-angular-2.html) to capture and use the input data.  You could just subscribe to changes made to the `ssid` control instead.

jtreuting: didn't know about that, thanks.  changed to use that approach

PR #240, opened 2017-07-15
Feature/batch exam report endpoints

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/report/BatchExamReportRequestHolder.java
```diff
@@ -1,13 +1,29 @@
...
+
 public class BatchExamReportRequestHolder {
 
+    private String id;
+    private PermissionScope permissionScope;
     private AbstractExamReportRequest<PrintOptions> request;
 
-    public BatchExamReportRequestHolder(final AbstractBatchExamReportRequest<PrintOptions> request) {
+    public BatchExamReportRequestHolder(final AbstractBatchExamReportRequest<PrintOptions> request, final PermissionScope permissionScope) {
+        this.id = UUID.randomUUID().toString();
```
esotrope: This is kind of placeholder until the interface is flushed out.
Right now each batch request just creates a uuid handle to return to the client for lookup and passes it to the queue for the processor to save the completed batch under.

webapp/src/test/java/org/opentestsystem/rdw/reporting/report/group/GroupExamReportPdfControllerIT.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private User user = User.builder()
+            .id("testUser")
+            .groupsById(ImmutableMap.of(1L, new GroupGrant(1L, 1)))
+            .authorities(ImmutableSet.of(new SimpleGrantedAuthority("ROLE_USER")))
+            .build();
+
+    @Test
+    public void getExamReportShouldSend404WhenServiceThrowsException() throws Exception {
+        when(service.queueExamReports(any(User.class), any(GroupExamReportPdfRequest.class)))
```
esotrope: * I've noted a todo to go in and add hashcode/equals so this works with eq()

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/schoolgrade/SchoolGradeExamReportPdfController.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private final SchoolGradeExamReportPdfService service;
+
+    @Autowired
+    SchoolGradeExamReportPdfController(final SchoolGradeExamReportPdfService service) {
+        this.service = service;
+    }
+
+    @GetMapping("/api/schools/{schoolId}/assessmentGrades/{gradeId}/examReports")
+    public String queueExamReports(
```
esotrope: Should batch submissions be GET calls? or POST? you are technically creating a batch and getting its ID

common/src/main/resources/messages.properties
```diff
@@ -60,9 +60,9 @@ report.iab.title.grade.11=Grade 11 Interim Assessment Blocks {0}-{1}
 report.iab.result-category.current=Current Results
 report.iab.result-category.previous=Previous Results
 
-report.iab.level.0.label=Below Standard
-report.iab.level.1.label=Near Standard
-report.iab.level.2.label=Above Standard
+report.iab.level.1.label=Below Standard
```
esotrope: DB stores performance levels starting at 1 instead of 0.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/schoolgrade/SchoolGradeExamReportPdfController.java
malaffoon: The way it is here i'd suggest a POST. 

Not thrilled with returning just a String. I think it should be a JSON payload (i.e. an object) even if there is only one field.

esotrope: Noted. I'll add the change minus HATEOAS in a followup PR

PR #241, opened 2017-07-15
Feature/notification system

wtritch:
After discussing with @aaronsleeper and @jtreuting the way we will deal with session timeouts is to refresh the browser window when an AJAX request returns 401 Unauthenticated.

To notify the user that their session timed out and the browser was refreshed, we will display a notification once the refresh has completed and they have logged back in (which may or may not require providing their SSO credentials.)

This PR introduces a NotificationService and NotificationComponent that can be used to display both immediate notifications to the user and queue Notifications to be displayed on next application initialization.
The queueing is performed via storing serialized notifications in a browser cookie for up to 30 minutes. (We don't want to display a "Your session expired and you were logged out" notification if the user closes the application and returns the next day.)

### DIFF
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -33,6 +33,7 @@
     "json-unflat": "^1.0.1",
     "lodash": "^4.17.4",
     "ngx-bootstrap": "^1.6.6",
+    "ngx-cookie": "^1.0.0",
     "primeng": "^4.1.0",
     "rxjs": "^5.4.2",
     "sbac-ui-kit": "git+https://github.com/SmarterApp/SBAC-Global-UI-Kit.git#v0.0.7",
--- a/webapp/src/main/webapp/src/app/app.component.html
+++ b/webapp/src/main/webapp/src/app/app.component.html
@@ -1,3 +1,6 @@
+<!-- Notifications -->
+<app-notifications></app-notifications>
+
 <!-- Header -->
 <header>
   <div class="container">
--- a/webapp/src/main/webapp/src/app/notification/notification.component.html
+++ b/webapp/src/main/webapp/src/app/notification/notification.component.html
@@ -1,0 +1,11 @@
+<div class="navbar-fixed-top mt-md mr-lg">
+  <div class="clearfix" *ngFor="let notification of notifications">
+    <alert class="pull-right"
+           [type]="notification.configuration.type"
+           [dismissible]="notification.configuration.dismissible"
+           [dismissOnTimeout]="notification.configuration.dismissOnTimeout"
+           (onClose)="onDismissed(notification)">
+      <span>{{notification.translationKey || translate}}</span>
+    </alert>
+  </div>
+</div>
--- a/webapp/src/main/webapp/src/app/notification/notification.component.ts
+++ b/webapp/src/main/webapp/src/app/notification/notification.component.ts
@@ -19,7 +19,7 @@ export class NotificationComponent implements OnInit {
 
   ngOnInit(): void {
     //Display qeued notifications
-    this.notificationService.unQueueNotifications().forEach(notification => {
+    this.notificationService.dequeueNotifications().forEach(notification => {
       this.onNotification(notification);
     });
   }
--- a/webapp/src/main/webapp/src/app/notification/notification.component.ts
+++ b/webapp/src/main/webapp/src/app/notification/notification.component.ts
@@ -1,0 +1,51 @@
+import { Component, OnInit } from "@angular/core";
+import { NotificationService } from "./notification.service";
+import { Notification } from "./notification.model";
+
+/**
+ * This component is responsible for displaying user notifications.
+ */
+@Component({
+  selector: 'app-notifications',
+  templateUrl: './notification.component.html'
+})
+export class NotificationComponent implements OnInit {
+
+  public notifications: Notification[] = [];
+
+  constructor(private notificationService: NotificationService) {
+    notificationService.notification.subscribe(this.onNotification.bind(this));
+  }
+
+  ngOnInit(): void {
+    //Display qeued notifications
+    this.notificationService.unQueueNotifications().forEach(notification => {
+      this.onNotification(notification);
+    });
+
+    // let note: Notification = new Notification("Test Notification");
+    // this.notificationService.queueNotification(note);
+  }
+
+  /**
+   * When a notification is dismissed, remove it from the list.
+   *
+   * @param notification The dismissed notification
+   */
+  public onDismissed(notification) {
+    let idx: number = this.notifications.indexOf(notification);
+    if (idx < 0) return;
+
+    this.notifications.splice(idx, 1);
+  }
+
+  /**
+   * When a notification is received, display it to the user.
+   *
+   * @param notification A notification
+   */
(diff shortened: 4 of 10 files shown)


webapp/src/main/webapp/package.json
```diff
@@ -33,6 +33,7 @@
     "json-unflat": "^1.0.1",
     "lodash": "^4.17.4",
     "ngx-bootstrap": "^1.6.6",
+    "ngx-cookie": "^1.0.0",
```
wtritch: I'm using the ngx-cookie library for interacting with browser cookies.

webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -1,3 +1,6 @@
+<!-- Notifications -->
+<app-notifications></app-notifications>
```
wtritch: Add the fixed-position notification glasspane to the application html.

webapp/src/main/webapp/src/app/notification/notification.component.html
```diff
@@ -0,0 +1,11 @@
...
+  <div class="clearfix" *ngFor="let notification of notifications">
+    <alert class="pull-right"
+           [type]="notification.configuration.type"
+           [dismissible]="notification.configuration.dismissible"
+           [dismissOnTimeout]="notification.configuration.dismissOnTimeout"
+           (onClose)="onDismissed(notification)">
+      <span>{{notification.translationKey || translate}}</span>
+    </alert>
+  </div>
+</div>
```
wtritch: We're using standard Bootstrap alerts to display notifications to the user.  The type, message, whether it's dismissible or auto-dismisses after a set period of time is all configurable when dispatching the Notification.

webapp/src/main/webapp/src/app/notification/notification.component.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Notification } from "./notification.model";
+
+/**
+ * This component is responsible for displaying user notifications.
+ */
+@Component({
+  selector: 'app-notifications',
+  templateUrl: './notification.component.html'
+})
+export class NotificationComponent implements OnInit {
```
wtritch: This is the component that binds to the NotificationService and actually renders the notifications to the UI.

webapp/src/main/webapp/src/app/notification/notification.model.ts
```diff
@@ -0,0 +1,22 @@
+import { isNullOrUndefined } from "util";
+import * as _ from "lodash";
+
+/**
+ * This model represents a notification message.
+ */
+export class Notification {
```
wtritch: Notification model with default configuration.

webapp/src/main/webapp/src/app/notification/notification.service.ts
```diff
@@ -0,0 +1,55 @@
+import { Injectable, EventEmitter } from "@angular/core";
+import { Notification } from "./notification.model";
+import { CookieService } from "ngx-cookie";
+
+/**
+ * This service is responsible for displaying notification alerts
+ * to the user.
+ */
+@Injectable()
+export class NotificationService {
```
wtritch: This is the injectable service that can be used to either queue or display a notification from anywhere.

webapp/src/main/webapp/src/app/notification/notification.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.cookieService.putObject(this.cookieName, existing, {expires: expiration});
+  }
+
+  /**
+   * Retrieve notifications queued for display on next application initialization.
+   * Note that this will purge the queue.
+   *
+   * @returns {Array} The queued notifications
+   */
+  public unQueueNotifications(): Notification[] {
```
jtreuting: I'm used to seeing queue and dequeue.  So I would think this should be named `dequeueNotifications`

webapp/src/main/webapp/src/app/notification/notification.module.ts
```diff
@@ -0,0 +1,27 @@
...
+import { CookieModule } from "ngx-cookie";
+
+@NgModule({
+  declarations: [
+    NotificationComponent
+  ],
+  imports: [
+    AlertModule,
+    BrowserModule,
+    CommonModule,
```
bgsavage: What do you use in CommonModule?  I would think this dependency should be reversed (i.e. have CommonModule pull this in instead of AppModule)

wtritch: It uses the Translation package from the CommonModule.  I'll look into inverting the dependencies.

webapp/src/main/webapp/src/app/notification/notification.service.ts
wtritch: I like that better.  Changed.

webapp/src/main/webapp/src/app/notification/notification.module.ts
wtritch: That doesn't look like it'll work so well, since app.component.html contains the messaging glasspane

webapp/src/main/webapp/src/app/notification/notification.component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  public notifications: Notification[] = [];
+
+  constructor(private notificationService: NotificationService) {
+    notificationService.onNotification.subscribe(this.onNotification.bind(this));
+  }
+
+  ngOnInit(): void {
+    //Display qeued notifications
```
bgsavage: nit: spelling

webapp/src/main/webapp/src/app/notification/notification.model.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  // Default configuration
+  public configuration: any = {
+    type: "info",
```
bgsavage: should we have an enum for the types?

wtritch: I'd rather not, since we'd have to translate to/from enums.  I can add default bootstrap values as a comment on the default configuration?

webapp/src/main/webapp/src/app/notification/notification.component.ts
wtritch: doh, fixed

webapp/src/main/webapp/src/app/notification/notification.service.ts
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * @param notification A notification message
+   */
+  public queueNotification(notification: Notification): void {
+    let existing: Notification[] = this.dequeueNotifications();
+    existing.push(notification);
+
+    let expiration: Date = new Date();
+    expiration.setMinutes(expiration.getMinutes() + this.expirationMinutes);
+    this.cookieService.putObject(this.cookieName, existing, {expires: expiration});
```
bgsavage: We should consider wrapping `cookieService` with some `PersistableStorage` service we write that under the hood either saves as cookie or local storage (I'd advocate local storage here instead of cookies)

wtritch: The benefit of using cookies is that we can set an expiration.  If we used local storage, we'd need to manage notification queue expiration ourselves.

PR #242, opened 2017-07-15
On route change scroll to the top of the window, except when that rou…

jtreuting:
…te change is the back button

Implemented solution found here: https://stackoverflow.com/questions/39601026/angular-2-scroll-to-top-on-route-change

Based on someone's comment, I tested in Safari and it works fine as-is without doing the `setTimeout(...,0)` trick for calling the scroll, so I didn't add that in

### DIFF
--- a/webapp/src/main/webapp/src/app/app.component.ts
+++ b/webapp/src/main/webapp/src/app/app.component.ts
@@ -2,6 +2,8 @@ import { Angulartics2GoogleAnalytics } from 'angulartics2';
 import { Component } from "@angular/core";
 import { TranslateService } from "@ngx-translate/core";
 import { UserService } from "./user/user.service";
+import { Router, NavigationEnd } from '@angular/router';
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -41,5 +45,25 @@ export class AppComponent {
...
+    });
+
+    this.router.events.subscribe((evt) => {
+      if (!(evt instanceof NavigationEnd)) {
+        return;
+      }
+
+      // if the user is going back, don't do the window scroll
+      // let the browser take the user back to the position where they last were
+      if (evt.url == this._lastPoppedUrl) {
```
bgsavage: This doesn't work then if the user goes 2+ more back in history then?

jtreuting: I will test that

PR #244, opened 2017-07-17
Return a no-challenge 401 Unauthorized response for /api/** requests …

wtritch:
…when not authenticated

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/Http401NotAuthorizedNoChallengeEntryPoint.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/Http401NotAuthorizedNoChallengeEntryPoint.java
@@ -1,0 +1,25 @@
+package org.opentestsystem.rdw.reporting.security;
+
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.AuthenticationEntryPoint;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+import static javax.servlet.http.HttpServletResponse.SC_UNAUTHORIZED;
+
+/**
+ * An authentication entry point which returns 401 Unauthorized without
+ * a challenge or redirection for SSO authentication.
+ */
+public class Http401NotAuthorizedNoChallengeEntryPoint implements AuthenticationEntryPoint {
+
+    @Override
+    public void commence(final HttpServletRequest request,
+                         final HttpServletResponse response,
+                         final AuthenticationException authException) throws IOException, ServletException {
+        response.sendError(SC_UNAUTHORIZED, "Unauthorized");
+    }
+}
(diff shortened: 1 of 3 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/Http401NotAuthorizedNoChallengeEntryPoint.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class Http401NotAuthorizedNoChallengeEntryPoint implements AuthenticationEntryPoint {
+
+    @Override
+    public void commence(final HttpServletRequest request,
+                         final HttpServletResponse response,
+                         final AuthenticationException authException) throws IOException, ServletException {
+        response.sendError(SC_UNAUTHORIZED, "Unauthorized");
+    }
+}
```
wtritch: This is the no-challenge entry point that just sends a 401 response

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -245,6 +250,19 @@ public SAMLEntryPoint samlEntryPoint() {
...
+    public DelegatingAuthenticationEntryPoint delegatingEntryPoint() {
+        final LinkedHashMap<RequestMatcher, AuthenticationEntryPoint> entryPoints = new LinkedHashMap<>();
+
+        final Http401NotAuthorizedNoChallengeEntryPoint noChallengeEntryPoint = new Http401NotAuthorizedNoChallengeEntryPoint();
+        entryPoints.put(new AntPathRequestMatcher("/api/**"), noChallengeEntryPoint);
+
+        final DelegatingAuthenticationEntryPoint delegatingEntryPoint = new DelegatingAuthenticationEntryPoint(entryPoints);
+        delegatingEntryPoint.setDefaultEntryPoint(samlEntryPoint());
+        return delegatingEntryPoint;
+    }
```
wtritch: This delegating entry point will use the no-challenge entry point on "/api/**" requests and the samlEntryPoint otherwise

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -477,7 +495,7 @@ public void configure(WebSecurity web) {
     protected void configure(HttpSecurity http) throws Exception {
         http
                 .httpBasic()
-                .authenticationEntryPoint(samlEntryPoint());
+                .authenticationEntryPoint(delegatingEntryPoint());
```
wtritch: This just registers the delegating entry point as the general system entry point impl

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
esotrope: This is great. Way better than the way i figured out before 👍

wtritch: Thanks! =)

PR #246, opened 2017-07-17
Bugfix/unittest against aurora

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/246)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,12 @@ allprojects {
     if (project.hasProperty("buildNumber")) {
         project.setVersion(project.version.toString().replaceAll("SNAPSHOT", "${buildNumber}"))
     }
+
+    processTestResources {
+        filesMatching("**/application-test.yml") {
+            expand(project.properties)
+        }
+    }
 }
 
 subprojects {
--- a/common/src/test/java/org/opentestsystem/rdw/reporting/common/test/ITDataSourceConfiguration.java
+++ b/common/src/test/java/org/opentestsystem/rdw/reporting/common/test/ITDataSourceConfiguration.java
@@ -6,12 +6,13 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.Profile;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.test.context.ActiveProfiles;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
 
 import javax.sql.DataSource;
 
-@Profile("test")
+@ActiveProfiles("test")
 @Configuration
 @EnableTransactionManagement
 @Import({DataSourceConfiguration.class})
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,3 +10,7 @@
 dockerHubUser=
 dockerHubPassword=
 dockerHubEmail=
+
+database_url=jdbc:mysql://localhost:3306/
+database_user=root
+database_password=
--- a/webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
+++ b/webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
@@ -26,7 +26,7 @@ public class JdbcEthnicityRepositoryIT {
 
     @Test
     public void findAllCodesShouldReturnAllCodes() throws Exception {
-        assertThat(repository.findAllCodes()).containsExactlyInAnyOrder("e1", "e2", "e3");
+        assertThat(repository.findAllCodes()).contains("e1", "e2", "e3");
     }
 
 }
--- a/webapp/src/test/resources/application-test.yml
+++ b/webapp/src/test/resources/application-test.yml
@@ -1,6 +1,8 @@
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/reporting_test?useSSL=false&useLegacyDatetimeCode=false
+    url: ${database_url}reporting_test?useSSL=false&useLegacyDatetimeCode=false
+    password: ${database_password}
+    username: ${database_user}
 
 artifacts:
   root: "classpath:"
--- a/common/src/test/resources/application-test.yml
+++ b/common/src/test/resources/application-test.yml
@@ -1,8 +1,8 @@
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/reporting_test?useSSL=false&useLegacyDatetimeCode=false
-    username: root
-    password:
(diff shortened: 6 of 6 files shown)


build.gradle
```diff
@@ -48,6 +48,12 @@ allprojects {
     if (project.hasProperty("buildNumber")) {
         project.setVersion(project.version.toString().replaceAll("SNAPSHOT", "${buildNumber}"))
     }
+
+    processTestResources {
```
rcordeau: Allows for gradle to fill in the placeholders in the application-test.yml files with information present in the project properties.

common/src/test/java/org/opentestsystem/rdw/reporting/common/test/ITDataSourceConfiguration.java
```diff
@@ -6,12 +6,13 @@
...
 import org.springframework.context.annotation.Profile;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.test.context.ActiveProfiles;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
 
 import javax.sql.DataSource;
 
-@Profile("test")
+@ActiveProfiles("test")
```
rcordeau: I could be wrong on this, but I thought it should be ActiveProfiles("test") like the other places in the code.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
```diff
@@ -26,7 +26,7 @@
 
     @Test
     public void findAllCodesShouldReturnAllCodes() throws Exception {
-        assertThat(repository.findAllCodes()).containsExactlyInAnyOrder("e1", "e2", "e3");
+        assertThat(repository.findAllCodes()).contains("e1", "e2", "e3");
```
rcordeau: Without this, the test expects that only e1, e2, and e3 will be in the table. The test should be able to handle any other stray data and just verify that the test data is contained.

gradle.properties
```diff
@@ -9,4 +9,8 @@
 
 dockerHubUser=
 dockerHubPassword=
-dockerHubEmail=
\ No newline at end of file
+dockerHubEmail=
+
+database_url=jdbc:mysql://localhost:3306/
```
esotrope: nit: remove the / here and let the consumers of the property add that

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
esotrope: So did we find out why non IT data found its way into the _test schema?

build.gradle
esotrope: Do we need this with the ability to use environment variables and java options?
I would think that would be our primary way of overriding anything in the application*.yml files

rcordeau: Yes, we do. This is the way to allow environment variables (Gradle specific variables) to override place holders in the .yml files

esotrope: I guess my question is why do we need gradle specific variables?
Can't we just keep the variables in the application.yml file and then when we want to override them in an env we use environment variables/java options?

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
rcordeau: Yeah, it turns out that when run locally, the order of Ingest integration tests magically cleans up data. If the tests are run in a different order, then data migrated from warehouse is kept in the reporting tables. The integration test data is cleaned up, but the initial .dml data in the warehouse is migrated and stays around.

gradle.properties
rcordeau: Yeah, habit of adding the trailing slash.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
esotrope: Are the ethnicities in the .dml?
I thought they were migrated i guess

build.gradle
rcordeau: We need gradle specific environment variables when running in the CI environment. The CI environment for both Reporting and Ingest is using Gradle to execute tests. It can be done using Spring environment variables, but then we will have Spring env variables for Reporting and Gradle env variables that are needed for schema. Using Gradle variables keeps everything consistent.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
rcordeau: Yes, initial ethnicities are in the warehouse .dml. They are migrated during an ingest integration test, along with some IT test values (indicated by negative id numbers). The test data is cleaned out, but the migrated data remains.

esotrope: ok cool - thanks for explaining

build.gradle
malaffoon: I would like to talk this out. I'm not understanding the inconsistency. And i'm a bit worried that the source code now has more stuff that relies on environment variables being set.

rcordeau: In order to run integration tests against Aurora, we need to specify the URL of the server. This really wasn't done before.

common/src/test/java/org/opentestsystem/rdw/reporting/common/test/ITDataSourceConfiguration.java
agorina: I do not think this is the right change. It meant to limit when it is loaded.

rcordeau: Ok, I can put it back.

webapp/src/test/resources/application-test.yml
```diff
@@ -1,6 +1,8 @@
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/reporting_test?useSSL=false&useLegacyDatetimeCode=false
+    url: ${database_url}reporting_test?useSSL=false&useLegacyDatetimeCode=false
```
malaffoon: Perhaps only substitute the dbhost, here?  ie. `url: jdbc:mysql://${dbhost}/reporting_test...`

malaffoon: Perhaps follow the pattern we have for the "secret" properties, e.g. `${rdw.reporting.test.dbhost}`

PR #249, opened 2017-07-17
moves yml props away from deprecated ones. e.g. spring.datasource to …

esotrope:
…app.datamart.datasource

1. application.state -> app.state.code
2. spring.datasource -> app.datamart.datasource
3. added app.batch.datasource to pdf processor

Gitlab has both the new and old prop keys


bgsavage:
Do you plan to update the config repo prior to merging this?

esotrope:
Gitlab has the new and the old prop keys already.

PR #252, opened 2017-07-18
Added position to AssessmentItem

bgsavage:
Now the Items By Points Earned table now shows item position rather than item id.  Updated api and ui.

Also fixed issue with the mapper bombing with claim scale scores are null, so now handling and showing "-" when this is the case.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
@@ -10,6 +10,7 @@ import { ordering } from "@kourge/ordering";
 import { byNumber } from "@kourge/ordering/comparator";
 import { ClaimScore } from "./model/claim-score.model";
 import { Student } from "../student/model/student.model";
+import { isNullOrUndefined } from "util";
 
 @Injectable()
 export class AssessmentExamMapper {
@@ -77,7 +78,10 @@ export class AssessmentExamMapper {
     uiModel.administrativeCondition = apiModel.administrativeConditionCode;
     uiModel.completeness = apiModel.completenessCode;
     uiModel.schoolYear = apiModel.schoolYear;
-    uiModel.claimScores = this.mapClaimScaleScoresFromApi(apiModel.claimScaleScores);
+
+    if(apiModel.claimScaleScores) {
+      uiModel.claimScores = this.mapClaimScaleScoresFromApi(apiModel.claimScaleScores);
+    }
 
     if (apiModel.studentContext) {
       uiModel.migrantStatus = apiModel.studentContext.migrantStatus;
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/student/results/tables/student-history-ica-summitive-table.component.html
```diff
@@ -90,7 +90,8 @@
             sortable="true"
             [hidden]="displayState.table !== 'claim'">
     <ng-template let-exam="rowData.exam" pTemplate="body">
-      {{ 'enum.iab-category.' + exam.claimScores[idx].level.toString() | translate }}
+      <span *ngIf="exam.claimScores[idx].level">{{ 'enum.iab-category.' + exam.claimScores[idx].level.toString() | translate }}</span>
```
bgsavage: @jtreuting @wtritch  seems odd we are using iab-category here, is this right?  Also summative is mispelled in this file :)

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -145,14 +149,17 @@ export class AssessmentExamMapper {
...
   private mapClaimScaleScoreFromApi(apiScaleScore: any): ClaimScore {
     let uiModel: ClaimScore = new ClaimScore();
-    uiModel.level = apiScaleScore.level;
-    uiModel.score = apiScaleScore.value;
-    uiModel.standardError = apiScaleScore.standardError;
+
+    if(apiScaleScore) {
+      uiModel.level = apiScaleScore.level;
+      uiModel.score = apiScaleScore.value;
+      uiModel.standardError = apiScaleScore.standardError;
```
bgsavage: Now only set if apiScaleScore is not null.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -168,7 +168,8 @@ <h3 class="gray-darkest mb-md">
                     sortable="true"
                     [hidden]="!isClaimScoreSelected">
             <ng-template let-exam="rowData" pTemplate="body">
-              {{ 'enum.iab-category.' + exam.claimScores[idx].level.toString() | translate }}
+              <span *ngIf="exam.claimScores[idx].level">{{ 'enum.iab-category.' + exam.claimScores[idx].level.toString() | translate }}</span>
+              <span *ngIf="!exam.claimScores[idx].level">-</span>
```
bgsavage: Show a "-" if there is no scale score

webapp/src/main/webapp/src/app/student/results/tables/student-history-ica-summitive-table.component.html
jtreuting: The values "Below Standard", "Above Standard", and "Near Standard" are the same values that are used as the IAB category values.  We could separate them out i guess in case they need to change independently, but really they are common to both of the situations they were just put in iab-category first.  So maybe the name should be changed to be more general for IAB category and claim values

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
jtreuting: I'm assuming we will need to address this not be included in the average scale score also.  i don't see that included.  if so, i'll make a story for it

PR #254, opened 2017-07-18
Added level up/down to Items By Points Earned

bgsavage:


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -201,6 +201,10 @@ export class AssessmentResultsComponent implements OnInit {
     return GradeCode.getIndex(gradeCode);
   }
 
+  getPointRowStyleClass(index: number){
+    return index == 0 ? 'level-down' : '';
```
bgsavage: conditionally add the class 'level-down' only on the first points column.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -213,7 +213,7 @@ <h3 class="gray-darkest mb-md">
               <span [hidden]="!showValuesAsPercent">{{ item.fullCreditAsPercent | number:'1.0-0' }}%</span>
             </ng-template>
           </p-column>
-          <p-column *ngFor="let column of pointColumns" [field]="column.numberField" header="{{'labels.groups.results.assessment.items.cols.x-points' | translate:{id: column.points} }}" [sortable]="true">
+          <p-column *ngFor="let column of pointColumns; let i = index" [field]="column.numberField" header="{{'labels.groups.results.assessment.items.cols.x-points' | translate:{id: column.points} }}" [sortable]="true" [styleClass]="getPointRowStyleClass(i)">
```
bgsavage: I'm not sure how we can add the 'row-end' class here, as we need access to the row data at the p-column level, which I don't think we have.  I talked to @aaronsleeper and he said he might have a css solution on all empty `<td>`s here.

jtreuting: could it be done in the body template instead or is that what you are saying as the work around

bgsavage: The body template is in isde the `<td>` in which @aaronsleeper said he wouldn't be able to traverse up tot he <tr> from there.  But I think Aaron has a css fix for this.

PR #255, opened 2017-07-18
Feature/session expiration

wtritch:
This PR adds a session-expiration route that the user is navigated to when we have detected that their session has expired.

After discussing with @aaronsleeper and @bgsavage we will no longer be using growl/alert notifications *after* a browser reload after detecting session expiration.


webapp/src/main/webapp/package.json
```diff
@@ -33,7 +33,6 @@
     "json-unflat": "^1.0.1",
     "lodash": "^4.17.4",
     "ngx-bootstrap": "^1.6.6",
-    "ngx-cookie": "^1.0.0",
```
wtritch: I removed the ngx-cookie dependency since we don't need to preserve notifications across a browser refresh.

webapp/src/main/webapp/src/app/notification/notification.module.ts
```diff
@@ -1,27 +0,0 @@
...
-    CookieModule.forChild()
-  ],
-  exports: [
-    NotificationComponent
-  ],
-  providers: [
-    NotificationService
-  ]
-})
-export class NotificationModule {
```
wtritch: The NotificationModule was removed and sources moved into the CommonModule to avoid circular dependencies.

webapp/src/main/webapp/src/app/shared/authentication.service.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  /**
+   * On authentication failure, navigate the user to the session-timeout route.
+   */
+  handleAuthenticationFailure(): void {
+    this.storageService.getStorage(StorageType.Session)
+      .setItem(this.locationKey,  window.location.href);
+    this.router.navigate(["session-expired"]);
+  }
```
wtritch: On authentication failure, we store the current url and navigate to the "session-expired" route.

webapp/src/main/webapp/src/app/shared/authentication/session-expired.component.ts
```diff
@@ -0,0 +1,25 @@
...
+   * On OK, navigate the browser to their previous route with a full browser refresh.
+   * If a previous route is not available, navigate the user to the home page.
+   */
+  onOk(): void {
+    let location = this.authenticationService.getReauthenticationLocation();
+    if (!location) {
+      location = this.location.prepareExternalUrl("");
+    }
+    window.location.href = location;
+  }
```
wtritch: When the user clicks "Ok" we navigate the browser back to the route where their authentication failure was detected.  If we no longer have the route (due to a deep link to the session-expiration page or some other funkiness) we navigate them to the home page.

Note that in order to refresh the browser, we need to use the old-skool javascript navigation method of setting the window.location.href property.  There doesn't appear to be a way to tell Angular Router to refresh the browser on navigation.

webapp/src/main/webapp/src/app/shared/cachingData.service.ts
```diff
@@ -1,26 +1,26 @@
...
+import { DataService } from "./data/data.service";
 
 @Injectable()
 export class CachingDataService {
 
   private cache: { [key: string]: any } = {};
 
-  constructor(private http: Http) {
+  constructor(private dataService: DataService) {
   }
```
wtritch: In order to trap all requests to "/api/**" I changed the CachingDataService to use the DataService rather than Angular's Http service.

webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -1,21 +1,28 @@
...
+              private authenticationService: AuthenticationService) {
   }
 
   get(url, options?: RequestOptionsArgs): Observable<any> {
     return this.http
       .get(`/api${url}`, options)
+      .catch((response: Response) => {
+        if (response.status === 401) {
+          this.authenticationService.handleAuthenticationFailure();
+        }
```
wtritch: If any request piped through the DataService responds with a 401, ask the AuthenticationService to handle it.

webapp/src/main/webapp/src/app/shared/notification/notification.component.ts
```diff
@@ -7,23 +7,16 @@ import { Notification } from "./notification.model";
...
   constructor(private notificationService: NotificationService) {
     notificationService.onNotification.subscribe(this.onNotification.bind(this));
   }
 
-  ngOnInit(): void {
-    //Display qeued notifications
-    this.notificationService.dequeueNotifications().forEach(notification => {
-      this.onNotification(notification);
-    });
-  }
```
wtritch: I removed the concept of "queued notifications" in favor of directly-displaying notification messages.

webapp/src/main/webapp/src/app/shared/notification/notification.service.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable, EventEmitter } from "@angular/core";
+import { Notification } from "./notification.model";
+
+/**
+ * This service is responsible for displaying notification alerts
+ * to the user.
+ */
+@Injectable()
+export class NotificationService {
```
wtritch: Since nothing in our code-base or mockups appears to need a notification service, I'm tempted to file a follow-up PR to remove it.  Leaving unused code in the code-base makes me twitch a bit, and we can always recover the code from GitHub's history.  Thoughts?

webapp/src/main/webapp/src/app/shared/storage.service.ts
```diff
@@ -0,0 +1,43 @@
+import { Injectable } from "@angular/core";
+
+/**
+ * This service is responsible for providing Storage implementations.
+ */
+@Injectable()
+export class StorageService {
```
wtritch: This is a StorageService that provides instances of SBStorage based upon a type enum.  Right now, it just supports Session and Local backed by browser sessionStorage and localStorage.  We could potentially expand this to a memory or cookie implementation as well in the future if we see a need.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -9,8 +9,13 @@
...
     "noExamsSelected": "Please select from the assessments above",
     "notApplicable": "-",
+    "ok": "Ok",
     "logout": "Logout",
-    "loading": "Loading"
+    "loading": "Loading",
+    "session-expired": {
+      "title": "Authentication Expired",
+      "message": "Click Ok to re-authenticate."
+    }
```
wtritch: I've created a UI Design jira to have our UI/UX gurus give the page a once-over.

wtritch:
session-expired

webapp/src/main/webapp/src/app/shared/data/data.service.ts
jtreuting: right now it will throw right? i don't think this is an exception really, it's communicating a state back to us that we can and are handling.  so i don't think it should throw

webapp/src/main/webapp/src/app/shared/notification/notification.service.ts
jtreuting: right now the notification service is not being used because it was switched to the session-expired right?  so this will be here if we have the need but not active at the moment.  just checking.

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: I think we will want the language to be in English and not techie :)

webapp/src/main/webapp/package.json
bgsavage: Awesome, was wondering about this when I had to install a new npm dependency yesterday.

webapp/src/main/webapp/src/app/shared/notification/notification.service.ts
bgsavage: It's probably better to remove the unused code so we don't have to maintain it, and though there's always the potential of use, in that case we would bring it from history like you mentioned and update it.

webapp/src/main/webapp/src/app/shared/data/data.service.ts
wtritch: Nope, no throwing. It'll intercept and navigate the user.  (Although the underlying system will still log a communication failure to the console)

webapp/src/main/webapp/src/assets/i18n/en.json
wtritch: Recommendations welcome =)

webapp/src/main/webapp/src/app/shared/notification/notification.service.ts
wtritch: Right, but since we don't want to maintain it, I'm going to remove it in a small follow-up PR.  We can recover from github history if we find a use for it later. =)

PR #256, opened 2017-07-19
Feature/report download

esotrope:


webapp/src/main/webapp/package.json
```diff
@@ -26,9 +26,11 @@
     "@kourge/ordering": "^1.0.1",
     "@ngx-translate/core": "^6.0.1",
     "@ngx-translate/http-loader": "0.0.3",
+    "@types/file-saver": "0.0.1",
     "@types/lodash": "^4.14.69",
     "angulartics2": "^2.2.2",
     "core-js": "^2.4.1",
+    "file-saver": "^1.3.3",
```
esotrope: library used to save files with a name

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -0,0 +1,76 @@
...
+              </div>
+            </div>
+          </form>
+          <div class="alert alert-danger" *ngIf="errorMessage">
+            <p>{{errorMessage}}</p>
+          </div>
+
+        </div>
+        <div class="modal-footer">
+          <button class="btn btn-primary" (click)="form.ngSubmit.emit()"><i class="fa fa-download"></i> Download</button>
```
esotrope: This whole template is a placeholder for whatever spec gets approved later.
This is why you will find static english text and other silliness.

webapp/src/main/webapp/src/app/report/report-download.service.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /**
+   * Gets an exam report download if ready, otherwise throws an exception
+   *
+   * @param token the token used to lookup the download
+   * @returns {Observable<ReportDownload>}
+   */
+  public getBatchExamReport(token: ReportDownloadToken): Observable<ReportDownload> {
+    return this.getExamReport(`/api/examReports/${token.id}`);
```
esotrope: the batch related methods here are placeholder until the api is flushed out

webapp/src/main/webapp/src/app/report/report-download.service.ts
```diff
@@ -0,0 +1,120 @@
...
+      params: options != null ? this.toParameters(options) : null,
+      headers: new Headers({
+        Accept: 'application/pdf'
+      }),
+      responseType: ResponseContentType.Blob
+    }).map((response: Response) => new ReportDownload(
+        this.getFileNameFromResponse(response),
+        new Blob([ response.blob() ], { type: 'application/pdf' })
+      )
+    );
```
esotrope: There's potential here to make this into a more generic "getPDF" method later if we need it

webapp/src/main/webapp/src/app/report/report-download.service.ts
```diff
@@ -0,0 +1,120 @@
...
+    };
+  }
+
+  /**
+   * Gets the file name specified in the Content-Disposition HTTP response header
+   *
+   * @param response the HTTP response
+   * @returns {string} the file name
+   */
+  private getFileNameFromResponse(response: Response): string {
```
esotrope: http response utility method candidate

webapp/src/main/webapp/src/app/report/report-download-token.model.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Holds information used to get the exam report
+ */
+export class ReportDownloadToken {
+
+  public constructor(public id :string){}
+
+}
```
bgsavage: This is todo?

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -0,0 +1,76 @@
...
+                      <option selected [ngValue]="false">Color</option>
+                      <option [ngValue]="true">Grayscale</option>
+                    </select>
+                  </label>
+                </div>
+
+              </div>
+            </div>
+          </form>
+          <div class="alert alert-danger" *ngIf="errorMessage">
```
bgsavage: Since you're using a form, ideally would be if form is not valid.

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  public schoolYears: Array<number>;
+  public languages: Array<string> = [ 'eng', 'spa', 'vie' ];
+  public options: ReportOptions;
+
+  constructor(private examFilterOptionsService: ExamFilterOptionsService) {
+  }
+
+  ngOnInit(): void {
+    this.examFilterOptionsService.getExamFilterOptions()
+      .subscribe((options: ExamFilterOptions) => {
```
bgsavage: Consider creating a Resolve instead of having to subscribe in the ngOnInit

webapp/src/main/webapp/src/app/report/report-download.service.ts
```diff
@@ -0,0 +1,120 @@
...
+import { ReportDownloadToken } from "./report-download-token.model";
+import { Observable } from "rxjs";
+import { ReportDownload } from "./report-download.model";
+import { AssessmentType } from "../shared/enum/assessment-type.enum";
+import { AssessmentSubjectType } from "../shared/enum/assessment-subject-type.enum";
+
+@Injectable()
+export class ReportDownloadService {
+
+  constructor(private http: Http) {
```
bgsavage: Never use Http directly, instead use the our wrapper, DataService

webapp/src/main/webapp/src/app/report/report-download.service.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Gets a student exam report PDF download
+   *
+   * @param studentId the student ID
+   * @param options settings which to shape the report content
+   * @returns {Observable<ReportDownload>}
+   */
+  public getStudentExamReport(studentId: number, options: ReportOptions): Observable<ReportDownload> {
+    return this.getExamReport(`/api/students/${studentId}/examReport`, options);
```
bgsavage: Each of these api calls should be going through some mapper to map to the ui model to protect against api model changes.

webapp/src/main/webapp/src/app/report/report-download.service.ts
```diff
@@ -0,0 +1,120 @@
...
+  private getExamReport(url: string, options?: ReportOptions): Observable<ReportDownload> {
+    return this.http.get(url, {
+      params: options != null ? this.toParameters(options) : null,
+      headers: new Headers({
+        Accept: 'application/pdf'
+      }),
+      responseType: ResponseContentType.Blob
+    }).map((response: Response) => new ReportDownload(
+        this.getFileNameFromResponse(response),
+        new Blob([ response.blob() ], { type: 'application/pdf' })
```
bgsavage: I'm guessing we can't get the content-type from response should we decide we want to support multiple file types?

webapp/src/main/webapp/src/app/report/report-download-token.model.ts
esotrope: Yeah this is placeholder for handling batch requests later

webapp/src/main/webapp/src/app/report/report-download.component.html
esotrope: I'm not sure there will be a way to make this form invalid.
The only error cases I can think of are 
1. there are no exams to put in the report you request (404) - we may just want to return an empty report in this case i'm not sure of our requirement here yet.
2. something goes terribly wrong on the backend (500)

webapp/src/main/webapp/src/app/report/report-download.component.ts
esotrope: Ok. I'll take a look at what you've done with resolves and see if i can mimic that.

webapp/src/main/webapp/src/app/report/report-download.service.ts
esotrope: Just curious - what are we using the indirection for right now?

webapp/src/main/webapp/src/app/report/report-download.service.ts
esotrope: Not sure what you mean, this wraps the response in a domain model called ReportDownload.

webapp/src/main/webapp/src/app/report/report-download.service.ts
esotrope: yeah we probably can get that from the response. This definitely has potential to be more generic. It could be a "getDownload" kind of method because I think this would handle all downloads.

webapp/src/main/webapp/src/app/report/report-download.service.ts
esotrope: ok cool - I'll run the requests through there 👍

esotrope:
Incorporated feedback.
@bgsavage, @wtritch please re-review, especially data.service.ts

PR #260, opened 2017-07-19
Created item-info tab and populating claim & difficulty.

bgsavage:
I'm going to have quite a few changes to populate each of these data points, so I'm going to break it up as I implement each field.  In this case, it's just claim & difficulty which is data we already have loaded from the examitems call.  Other fields will require updates to the examitems call to get the needed data.


webapp/src/main/webapp/src/app/assessments/items/item-info/item-info.component.html
```diff
@@ -0,0 +1,40 @@
+<div>
+  <p>
+    <b>{{ 'labels.assessments.items.tabs.item-info.claim' | translate}}</b>
+    {{ 'definition.claim.' + item.claim + '.description' | translate }}
+  </p>
+  <p class="green">
+    // TODO:
```
bgsavage: Lots of todos here.

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -40,7 +40,7 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
                     content="{{'labels.groups.results.assessment.items.cols.claim-info' | translate}}"></span>
       </ng-template>
       <ng-template let-item="rowData.assessmentItem" pTemplate="body">
-        {{ 'enum.claim.' + item.claim | translate }} / {{ 'labels.groups.results.assessment.items.target' | translate:item }}
+        {{ 'definition.claim.' + item.claim + '.name' | translate }} / {{ 'labels.groups.results.assessment.items.target' | translate:item }}
```
bgsavage: moved the json location as it's more than just enum now.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -357,7 +367,7 @@
       "3": "Met Standard",
       "4": "Exceeded Standard"
     },
-    "claim-code": {
+    "subject-claim-code": {
```
bgsavage: renamed for clarity

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -388,25 +398,60 @@
...
       "E": "Easy",
       "M": "Moderate",
       "D": "Difficult"
     }
   },
+  "definition": {
+    "claim": {
+      "1-IT": {
+        "name":"Read Analytically: Informational Text",
+        "description": "Read Analytically: Informational Text - Students can read closely and analytically to comprehend a range of increasingly complex literary and informational texts."
```
bgsavage: If the table is small enough, we will load the description in the en.json.  For common core standard & target, I'll make individual api endpoints to get that data that at some future point we can run through translate (but for now will just get the description directly)

webapp/src/main/webapp/src/app/assessments/items/item-info/item-info.component.html
esotrope: Is it green so it stands out as a todo or will it actually be green? : j

bgsavage: As a todo so it looks like a comment if you open the tab :) . It'll help me keep track and let any testers know it's not done.

image

PR #263, opened 2017-07-19
Added depth of knowledge to item-tab

bgsavage:


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItem.java
```diff
@@ -83,6 +91,7 @@ public static Builder builder() {
         private String bankItemKey;
         private String claim;
         private String target;
+        private String depthOfKnowledge;
```
bgsavage: Even though it's a smallint in the database today, I'm creating this as a string on the model in anticipation of this being denormalized as a code which is a string.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItems.java
```diff
@@ -22,6 +22,7 @@ public static AssessmentItem map(final ResultSet row) throws SQLException {
                 .maximumPoints(row.getInt("max_points"))
                 .claim(row.getString("claim_code"))
                 .target(row.getString("target_code"))
+                .depthOfKnowledge(row.getString("dok_id")) // TODO: Should use code when it's available
```
bgsavage: todo: https://jira.fairwaytech.com/browse/DWR-613

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItem.java
```diff
@@ -52,6 +53,13 @@ public String getTarget() {
     }
 
     /**
+     * @return the depth of knowledge code
+     */
+    public String getDepthOfKnowledge() {
```
esotrope: I'm thinking we should probably add *Code to depthOfKnowledge, claim, target to remain consistent with our *Id, *Code naming.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItem.java
esotrope: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItem.java
bgsavage: Well it's not really a code yet, so I'm trying to keep it flexible so that when it does change it doesn't create churn.

esotrope: Well the idea here is that it will be a code right?
We should make the field naming consistent at some point. Thought this would be a good op.

bgsavage: I'm not sure if code is already implied if it's not an id and if clarification is required than the javadocs are good enough?

PR #265, opened 2017-07-19
Added math practice to item info tab

bgsavage:


webapp/src/main/webapp/src/app/assessments/items/item-info/item-info.component.html
```diff
@@ -22,10 +22,9 @@
     <b>{{ 'labels.assessments.items.tabs.item-info.ccs' | translate}}</b>
     Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse nec placerat velit.
   </p>
-  <p class="green">
-    // TODO:
+  <p *ngIf="item.mathPractice > 0">
```
bgsavage: The spec says this field is "optional", so it will only show if mathPractice is > 0, does this work for you @jtreuting ?

esotrope: If it is optional can you model it with type Integer and use QueryUtils.getNullable() - this is what the other repos are doing

bgsavage: Oh yeah you're right I should make this nullable.

bgsavage: Done.  Also updated my mispelling :D

PR #268, opened 2017-07-20
Bugfix/614 additional analytics

jtreuting:
Added analytics for:
- Expanding item row
- Selecting item tabs
- Going back to responses from Items by Points Earned
- School/Grade filters

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
@@ -2,6 +2,7 @@ import { Component, OnInit, Input, ViewChild } from "@angular/core";
 import { AssessmentItem } from "../model/assessment-item.model";
 import { TabsetComponent, TabDirective } from "ngx-bootstrap";
 import { Exam } from "../model/exam.model";
+import { Angulartics2 } from 'angulartics2';
 
 @Component({
   selector: 'item-tab',
@@ -63,7 +64,7 @@ export class ItemTabComponent implements OnInit {
 
   private _position: number = -1;
 
-  constructor() { }
+  constructor(private angulartics2: Angulartics2) { }
 
   ngOnInit(): void {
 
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
```diff
@@ -73,7 +74,36 @@ export class ItemTabComponent implements OnInit {
...
+  }
+
+  selectExemplarTab() {
+    this.loadExemplar = true;
+    this.trackAnalyticsEvent('ItemTabSelection', 'Rubric Exemplar');
+  }
+
+  selectItemInformationTab() {
+    this.trackAnalyticsEvent('ItemTabSelection', 'Item Information');
+  }
```
wtritch: We could combine these into a single function:
```
selectTab(tabName: string): void {
  this.trackAnalyticsEvent('ItemTabSelection', tabName);
}
```
then change the (click) handlers to something like:
`(click)=selectTab('my-tab-name')`

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
```diff
@@ -73,7 +74,36 @@ export class ItemTabComponent implements OnInit {
     setTimeout((function() {
       let tab: TabDirective = this.itemTabs.tabs[0];
       tab.select.emit(tab);
+
+      this.trackAnalyticsEvent('ExpandItem', tab.heading + ' for ' + this.item.bankItemKey);
```
wtritch: You shouldn't need this, the same select event should fire whether the user clicked on the tab or this initializer selects the tab.

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
jtreuting: the only reason I didn't do that was because because of the additional switch statement logic, it felt more hidden than the explicit method for each.  but i'm willing to change it

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
jtreuting: hmmmm, not sure I was seeing that behavior, but i'll double check

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -143,6 +145,39 @@ export class SchoolResultsComponent implements OnInit {
...
+    this.angulartics2.eventTrack.next({
+      action: 'ChangeYear',
+      properties: {
+        category: 'AssessmentResults',
+        label: this.currentSchoolYear
+      }
+    });
+
+    this.updateRoute();
   }
```
wtritch: I realize metrics are important, but this is starting to feel a little intrusive and duplicative... maybe modify updateRoute to take a string label?

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
wtritch: Maybe a switch is too bulky... if-chain with 2 cases?  I don't feel too strongly about it but a function per action feels like it will become a maintenance issue.  But maybe I'm jumping at shadows.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
jtreuting: yeah i like that.  i'll have a switch statement to determine which additional details to pass along.

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
jtreuting: tested and you are right, i must of had another issue when i first didn't see it.  thanks, removing.

PR #269, opened 2017-07-20
change to ethnicity IT to handle existing data

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/269)


webapp/src/test/java/org/opentestsystem/rdw/reporting/search/ethnicity/JdbcEthnicityRepositoryIT.java
```diff
@@ -26,7 +26,7 @@
 
     @Test
     public void findAllCodesShouldReturnAllCodes() throws Exception {
-        assertThat(repository.findAllCodes()).containsExactlyInAnyOrder("e1", "e2", "e3");
+        assertThat(repository.findAllCodes()).contains("e1", "e2", "e3");
```
rcordeau: From previous PR:
"Without this, the test expects that only e1, e2, and e3 will be in the table. The test should be able to handle any other stray data and just verify that the test data is contained.
       @esotrope
esotrope 2 days ago Member
So did we find out why non IT data found its way into the _test schema?
       @rcordeau
rcordeau 2 days ago • edited Owner
Yeah, it turns out that when run locally, the order of Ingest integration tests magically cleans up data. If the tests are run in a different order, then data migrated from warehouse is kept in the reporting tables. The integration test data is cleaned up, but the initial .dml data in the warehouse is migrated and stays around. It seems that running on AWS makes the order different, thus keeping data around.
       @esotrope
esotrope 2 days ago Member
Are the ethnicities in the .dml?
I thought they were migrated i guess
       @rcordeau
rcordeau 2 days ago Owner
Yes, initial ethnicities are in the warehouse .dml. They are migrated during an ingest integration test, along with some IT test values (indicated by negative id numbers). The test data is cleaned out, but the migrated data remains.
       @esotrope
esotrope 2 days ago Member
ok cool - thanks for explaining
"

PR #272, opened 2017-07-21
Feature/as tablewidth solution

aaronsleeper:
@jtreuting @bgsavage @wtritch @esotrope @jessicacolton look at the "my groups" page. I put a class of overflow on the table that was breaking out and restyled the header and footer. I think this is a better solution than the full-width tables


jtreuting:
@aaronsleeper i pull down this branch locally tonight and see what it looks like and provide any feedback.  thanks.

jtreuting:
It looks fine to me.  I'll let you and @jessicacolton decide on the UI difference of the header and the breadcrumb change.

This appears to solve the main results tables but I don't think it solves the problem of the Items by Points Earned when there is a max score of 10, so we need 11 columns for that.  The really wide table would have solved that as well, so i'm guessing we'll need another fix for that?  @aaronsleeper, is that right?

aaronsleeper:
It allows for infinitely wide tables as well. In either scenario 11 columns
will be a lot of data to look at. We can condense the breakdown into a
tooltip if you'd like or do they really need to be able to sort on every
point value?

PR #273, opened 2017-07-21
Fix session expiration 401 handling

wtritch:


webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -29,17 +29,22 @@ export class AssessmentExamMapper {
   }
 
   mapAssessmentsFromApi(apiModels): Assessment[] {
+    if (!apiModels) return [];
```
wtritch: I added error handlers to the DataService calls and handling for empty results to the mappers.

webapp/src/main/webapp/src/app/shared/cachingData.service.ts
```diff
@@ -23,9 +23,11 @@ export class CachingDataService {
...
 
     observable
-      .subscribe(x => {
-        this.cache[url] = x;
-      });
+      .subscribe((value) => {
+          this.cache[url] = value;
+        }, () => {
+          //ignore Errors
+        });
```
wtritch: This was strange, I needed to handle errors here or the error handling gets aborted...

webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -26,7 +26,7 @@ export class DataService {
   public get(url: string, options?: RequestOptionsArgs): Observable<any> {
     return this.http
       .get(`/api${url}`, options)
-      .catch(this.handleException)
+      .catch(this.handleException.bind(this))
```
wtritch: This was the biggest issue, binding the handleException method using a reference like this was loosing the "this" reference at runtime.

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter-options.service.ts
```diff
@@ -10,6 +11,10 @@ export class ExamFilterOptionsService {
   getExamFilterOptions() {
     return this.service
       .get("/examFilterOptions")
+      .catch((err) => {
+        console.warn(err);
+        return Observable.empty()
```
jtreuting: did this basically get added to every service?

wtritch: Yes, this handling was already on many of our DataService calls, I added it to the remaining calls for consistency and to allow the 401 handling to process.  If the error is not handled, the 401 handling does not process properly.  
However, the more I think about it, the less I like it.  It feels like we're fighting error handling rather than using it, by forcing services to handle unrecoverable errors.

We have a few choices:
1. leave the 401 handler on the DataService and percolate the error which must be handled by every service to allow the navigation to occur (this review)
2. leave the 401 handler on the DataService and return an Observable.empty() which must be handled by every service to allow the navigation to occur
3. move the 401 handler to a custom global Angular ErrorHandler and every service should *only* handle recoverable errors (e.g. a 404) and percolate unrecoverable errors (5xx, 401, 403, etc)
4. ??

Number 1 requires us to catch every error in every service, which feels bulky and non-DRY.
Number 2 re-interprets a 401 as NO CONTENT before passing back to the services, which forces services to either throw a dissociated NPE when assuming a data response.
Number 3 feels like the "angular way" but 401 handling could be mistakenly intercepted and prevented by any swallowing `.catch((err) => console.warn(err))` error handling between the DataService and the ErrorHandler.

...I'm leaning towards 3. as the most DRY, however.  Thoughts?

wtritch:
Closing in favor of https://github.com/SmarterApp/RDW_Reporting/pull/275

PR #274, opened 2017-07-21
Feature/average scale score component

jtreuting:
PR for review purposes to see if this approach makes sense.

Implemented the following:
- Average Scale Score component that can handle different displays for IAB and ICA/Summative
- Calculation of the average standard error (i think my stats calc is correct, @malaffoon would know)
- Determine level that an average scale score falls into.  This logic is simple for ICA since it is just where the score falls between the cutPoints.  For IABs the logic uses the midpoint to go +/- 1.5 * standard error and compare that tot he score to see if it is above or below to determine the level.  The logic used in TIS for the same calculation is here: https://github.com/SmarterApp/TDS_TestIntegrationSystem/blob/master/TestScoringEngine/ScoringEngine/Scoring/SpecificComputationRules.cs#L501
- Moved some things around as far as where the calculations are done, made some objects know how to calcualte things (went back and forth on this).

NOTE: The UI Kit classes are not yet in the latest release.  They will get pushed tomorrow and I'll validate it looks appropriate before moving forward.


webapp/src/main/webapp/src/app/assessments/model/assessment.model.spec.ts
```diff
@@ -0,0 +1,60 @@
+import {Assessment} from "./assessment.model";
+import {AssessmentType} from "../../shared/enum/assessment-type.enum";
+
+describe('Assessment Model', () => {
```
jtreuting: i didn't see any tests for models since most don't have logic in them, so maybe that should have told me something.  since I added a calculateLevelNumber method on the Assessment object I added tests here.  thoughts?

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
```diff
@@ -24,4 +25,61 @@ export class Assessment {
...
   }
+
+  /**
+   * Determines what level a scale score is in based on the cut points
+   *
+   * @param score the scale score
+   * @param standardError the standard error
+   * @returns {number} the level number (starting with 1)
+   */
+  calculateLevelNumber(score: number, standardError: number): number {
```
jtreuting: thought about just having 2 methods for IAB and then ICA/Sumamtive and the caller would decide, but the Assessment knows if it's IAB or ICA so that seemed strange to make the client determine.  The odd part is standardError is only needed for the IAB one.

webapp/src/main/webapp/src/app/assessments/model/exam-statistics.model.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  it('should calculate level percents', () => {
+    let stats = new ExamStatistics();
+    stats.total = 10;
+    stats.levels = [5, 2, 3].map(x => {
+      let level = new ExamStatisticsLevel();
+      level.value = x;
+      return level;
+    });
+
+    let actual = stats.percents;
```
jtreuting: instead of calculating the percents in the exam statistics calculator it seemed like the class itself could know how to do that since it had all the info, so added tests around that change.

webapp/src/main/webapp/src/app/assessments/model/exam-statistics.model.ts
```diff
@@ -0,0 +1,22 @@
...
+      return {
+        id: x.id,
+        value: x.value / this.total * 100,
+        suffix: '%'
+      }
+    })
+  };
+}
+
+export class ExamStatisticsLevel {
```
jtreuting: this felt like a subclass if i was writing this in C#, so I left it in the same model file, but I could be convinced to move it if that is not a thing done in typescript

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -46,9 +47,9 @@ enum ScoreViewState {
   ],
 })
 export class AssessmentResultsComponent implements OnInit {
-  exams = [];
+  exams: Exam[] = [];
   sessions = [];
-  statistics: any = { percents: {} };
+  statistics: ExamStatistics;
```
jtreuting: made these strong classes now

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -0,0 +1,72 @@
+<!-- IAB display -->
```
jtreuting: i took the UI Kit HTML and put both versions in the same template, also the HTMl is similar so I probably could try to combine them and have the classes determined dynamically but i thought it would make the file much harder to read.  but again went back and forth on those things.

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  get performanceLevels(): ExamStatisticsLevel[] {
+    if (this.showValuesAsPercent)
+      return this.statistics.percents;
+    else
+      return this.statistics.levels;
+  }
+
+  getLevelPercent(num: number): number {
```
jtreuting: the width of the bars in the visualization always need percent, so the performanceLevels method above gives the data in either percent or raw numbers based on the toggle, but we needed a way to always get the percents

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.spec.ts
```diff
@@ -5,6 +5,18 @@ import { ExamItemScore } from "../model/exam-item-score.model";
 
 describe('Exam Calculator', () => {
 
+  it('should return only scored exams', () => {
```
jtreuting: added logic to remove unscored exams from the average calculation.  I made it a javascript truthy type check because I honestly wasn;t sure if it would be returned as null, undefined or 0.

@agorina Are unscored exams NULL in the DB?

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -1,16 +1,33 @@
...
   }
 
-  groupLevels(exams, numberOfLevels) {
+  calculateAverageStandardError(exams: Exam[]) {
+    let scoredExams = this.getOnlyScoredExams(exams);
+
+    return Math.sqrt(
+      scoredExams.reduce((x, y) => x + (y.standardError * y.standardError), 0)
+      / scoredExams.length
+    );
```
jtreuting: So to calculate the average standard error for a group i'm using the square root of the average of the sum of the squares which I think is the right calculation based on previous conversations and me nodding along as @malaffoon described how simple it was to calculate.

and this was my cheat sheet: https://stats.stackexchange.com/questions/209585/the-average-of-mean-and-standard-deviation

webapp/src/main/webapp/src/app/assessments/results/scale-score.component.html
```diff
@@ -1 +1,4 @@
-<strong>{{score}}</strong> ± {{standardError}}
+<span class="scale-score">
+  <span class="score">{{ score }}</span>
+  <span class="error-band">± {{standardError}}</span>
+</span>
```
jtreuting: from the UI Kit example

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
wtritch: I know you've been waffling on this, and it feels pretty innocuous, but I have never seen ModelServices end well.  I'd recommend moving this into a ScaleScoreService that accepts the Assessment as well as these parameters, and expose it in your component.

webapp/src/main/webapp/src/app/assessments/model/exam-statistics.model.spec.ts
wtritch: I'd recommend calculating this in the Mapper and leaving the Model as a POJO/property holder.

webapp/src/main/webapp/src/app/assessments/model/exam-statistics.model.ts
wtritch: I don't have a problem with it =)

webapp/src/main/webapp/src/app/assessments/model/exam-statistics.model.ts
```diff
@@ -0,0 +1,22 @@
...
+  average: number;
+  standardError: number;
+  levels: ExamStatisticsLevel[];
+
+  get percents(): ExamStatisticsLevel[] {
+    return this.levels.map(x => {
+      return {
+        id: x.id,
+        value: x.value / this.total * 100,
+        suffix: '%'
```
wtritch: I'd recommend returning a numeric value and using translation to append the '%'

webapp/src/main/webapp/src/app/assessments/model/exam-statistics.model.ts
```diff
@@ -0,0 +1,22 @@
+export class ExamStatistics {
+  total: number;
+  average: number;
+  standardError: number;
+  levels: ExamStatisticsLevel[];
+
+  get percents(): ExamStatisticsLevel[] {
```
wtritch: A potential issue with having logic in the models is that this will be called on digest if it's bound to the HTML.

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
wtritch: I'm debating if you really want to average the standard errors, or if you just want to calculate a new standard error from your available scores...

jtreuting: we discussed with SBAC at some point and I'm 95% sure this is the appraoch that made sense to them.  we can't really do it based on the raw data because the individual standard error for a scale score depends on each item that goes into creating the scale score and the standard error.  So I think we would need to use all the raw items and know the details of how the score tests (which we don't have).

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
jtreuting: k, will do.  (i think Angular2's tagline should be "more files is better")

webapp/src/main/webapp/src/app/assessments/model/exam-statistics.model.ts
jtreuting: ahhhh, that makes sense. didn't think about that.  yeah we don't want that.  i'll move these out of here.  great point.

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
malaffoon: FWIW, looks right to me

PR #275, opened 2017-07-21
Bugifx/fix session expiration take 2

wtritch:
This is take-2 of the session expiration fix that uses an application ErrorHandler which allows services to *only* handle recoverable errors and pass other errors up the chain.


webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
```diff
@@ -28,8 +28,11 @@ export class GroupAssessmentService implements AssessmentProvider {
   getAvailableAssessments() {
     return this.dataService.get(`/groups/${this.groupId}/assessments`, { search: this.getSchoolYearParams(this.schoolYear) })
       .catch(response => {
-        console.warn(response);
-        return Observable.empty();
+        if (response.status == 404) {
+          return Observable.empty();
+        } else {
+          return Observable.throw(response);
```
wtritch: Now these error catch statements just need to handle recoverable/expected errors.  Others are percolated up to the ErrorHandler.

webapp/src/main/webapp/src/app/shared/cachingData.service.ts
```diff
@@ -25,7 +25,7 @@ export class CachingDataService {
     observable
       .subscribe(x => {
         this.cache[url] = x;
-      });
+      }, (err) => Observable.throw(err));
```
wtritch: Something about the .subscribe not having an explicit error handler causes the error to be swallowed here...

webapp/src/main/webapp/src/app/shared/common.module.ts
```diff
@@ -69,7 +70,10 @@ import { StorageService } from "./storage.service";
     CachingDataService,
     ColorService,
     RdwTranslateLoader,
-    StorageService
+    StorageService, {
+      provide: ErrorHandler,
+      useClass: SBErrorHandler
+    }
```
wtritch: This registers our custom ErrorHandler

webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -40,24 +37,10 @@ export class DataService {
...
-   *
-   * @param response the API server response
-   * @returns {Observable<Response>}
-   */
-  private handleException(response: Response): Observable<Response> {
-    if (response.status === 401) {
-      this.authenticationService.handleAuthenticationFailure();
-    }
-    return Observable.throw(response);
-  }
```
wtritch: We don't need to trap this at the DataSource layer any longer.

webapp/src/main/webapp/src/app/shared/sb-error-handler.service.ts
```diff
@@ -0,0 +1,35 @@
...
+   * @param error An error
+   */
+  handleError(error: any): void {
+    if (this.isUnauthenticated(error)) {
+      let authService: AuthenticationService = this.injector.get(AuthenticationService);
+      setTimeout(() => authService.handleAuthenticationFailure(), 0);
+    } else {
+      super.handleError(error);
+    }
+  }
```
wtritch: This is the error handler that passes 401s to the AuthenticationService for handling.  Attempting to inject the AuthenticationService directly causes a circular reference error, since the application context isn't initialized when angular attempts to instantiate the ErrorHandler.  Injecting the injector was the cleanest workaround I could find.

webapp/src/main/webapp/src/app/shared/sb-error-handler.service.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  private isUnauthenticated(error: any): boolean {
+    return (error instanceof Response && error.status == 401) ||
+      (error && error.rejection instanceof Response && error.rejection.status == 401);
+  }
```
wtritch: Sometimes Angular "helpfully" wraps the Response error object in an Error object as the "response" property...

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
esotrope: nit: no else block is needed here.

Also, is the idea here that these catch statements be expanded on? or can we just have some utility ErrorHandler.standardCatcher variable to hold this repeated method?

wtritch: Doh, will remove the else.
hmm... we could create a static utility handle404AsEmpty method, maybe in a ResponseUtils class?

esotrope: Yeah that sounds good 👍

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-scoring.service.ts
```diff
@@ -12,6 +13,11 @@ export class ItemScoringService {
   getGuide(bankItemKey: string): Observable<ItemScoringGuide>{
     return this.dataService
       .get(`/examitems/${bankItemKey}/scoring`)
-      .map(guide => this.mapper.mapFromApi(guide));
+      .catch(ResponseUtils.notFoundToNull)
+      .map((guide) => {
+        if (guide == null) return null;
+
+        return this.mapper.mapFromApi(guide);
+      });
```
wtritch: After a long chat with @esotrope we decided on this standard error-handling approach at the service layer, with the following goals:
1. components shouldn't know about how services work (http, cached value, etc)
2. services should return value, no-value, or opaque error

PR #285, opened 2017-07-24
Feature/student grade csv export

wtritch:
Small PR to add CSV export to the third and last results page.


webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -165,6 +178,48 @@ export class SchoolResultsComponent implements OnInit {
     return Number.parseInt(params[ "schoolYear" ]) || this.filterOptions.schoolYears[ 0 ];
   }
 
+  exportCsv(): void {
```
bgsavage: I think I'd prefer to see this work done in a service rather than a component.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -165,6 +178,48 @@ export class SchoolResultsComponent implements OnInit {
...
+    let getExam = (item) => item.exam;
+    let getAssessment = (item) => item.assessment;
+    let getIABExam = (item) => item.assessment.isIab ? item.exam : null;
+    let getNonIABExam = (item) => item.assessment.isIab ? null: item.exam;
+    let getNonIABMathExam = (item) => !item.assessment.isIab && item.assessment.subject === 'MATH' ? item.exam : null;
+    let getNonIABElaExam = (item) => !item.assessment.isIab && item.assessment.subject === 'ELA' ? item.exam : null;
+
+    let data: string[][] = this.csvBuilder
+      .newBuilder()
+      .withStudentIdAndName(getStudent)
```
bgsavage: maybe this function could just be called `withStudent`.  I don't think we'd want to rename this function should the requirement of what parts of the student we want to export changes.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -165,6 +178,48 @@ export class SchoolResultsComponent implements OnInit {
...
+      .withMathClaimScores(getNonIABMathExam)
+      .withELAClaimScores(getNonIABElaExam)
+      .withGender(getStudent)
+      .withStudentContext(getExam)
+      .build(sourceData);
+
+    let fileName: string = this._currentSchool.name +
+      "-" + this.translateService.instant(`labels.grades.${this._currentGrade.code}.short-name`) +
+      "-" + new Date().toDateString();
+    new Angular2Csv(data, fileName);
```
bgsavage: I was under the impression the CsvBuilder would build the actually csv, and invoke `Angular2Csv`.  Otherwise, it's not really a `CsvBuilder`, but a `CsvAngular2CsvDataFormatBuilder`, if that makes sense... :)

wtritch: I could move that into the builder and add a withFilename() method.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
wtritch: renamed

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
wtritch: I'm trying to decide how to break this into a re-usable service rather than just move the logic into a separate file.
I think columns should be defined at the component level since the component knows what needs to be exported.  I think source data should be defined at the component level since the component knows how to access/compose the export data. Since the component knows the format of the source data, it is also responsible for defining the data filters... Thoughts?

bgsavage: I would disagree.  Sure, the component knows how to access the data but only for the purposes of presenting the data.  The component shouldn't know how to compose data that will be exported csv, that's implementation details that should be in a service.  Right now we are exporting via client side.  Think about if we wanted to swap out the implementation to be done server side.  Now the component should have to change because the underlying implementation change?  Or should we have a layer of abstraction and lean closer into an SRP approach.

What's interesting here also is we are doing it at the school-grade level rather than the assessments level.  I would think the exported data would be the same if it came from school / grade search vs. group results, yeah?

One last point I'd like to make to advocate to moving to a service... Testing.  It's much easier to write unit test against a service rather than a component.  Component unit tests run much slower, and require much more setup.

wtritch: Done.

PR #288, opened 2017-07-24
Fix build

wtritch:


webapp/src/main/webapp/src/app/csv-export/angular-csv.provider.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from "@angular/core";
+import { Angular2Csv } from "angular2-csv";
+
+/**
+ * This provider wraps the Angular2Csv library.
+ */
+@Injectable()
+export class Angular2CsvProvider {
```
wtritch: In order to still test the CsvBuilder, I needed to wrap the Angular2Csv library in a provider for testing.  Unfortunately the library doesn't provide an angular module for easy testing/mocking.

PR #290, opened 2017-07-24
Feature/improve pdf processor configs

esotrope:


pdf-report-processor/src/main/resources/application.yml
```diff
@@ -63,7 +63,7 @@ app:
       initialize: false
 
   wkhtmltopdf:
-    url: http://wkhtmltopdf/
+    url: http://localhost:8082
```
malaffoon: i assume (from the sbac_config_repo review) that the loss of the trailing slash is intentional?

esotrope: It should resolve to the same URL with or without

PR #292, opened 2017-07-25
adds header, student and footer styles/markup

esotrope:


common/src/main/resources/templates/ica-report.html
```diff
@@ -6,40 +6,42 @@
...
+            </div>
 
-        <div class="well">
-            <div class="well-body">
-                <div th:text="#{report.accommodations.label}"></div>
-                <!-- <li th:each="accommodation : ${report.exam.accommodationCodes}>" -->
+            <div class="well">
+                <div class="well-body">
+                    <div th:text="#{report.accommodations.label}"></div>
+                    <!-- <li th:each="accommodation : ${report.exam.accommodationCodes}>" -->
```
bgsavage: nit: commented code

common/src/main/resources/templates/fragments.html
```diff
@@ -1,65 +1,103 @@
...
+</head>
+<body>
 
-        <div th:fragment="styles">
-            <!-- TODO: add path to stylesheet that wkhtmltopdf can reach -->
-            <!--<link rel="stylesheet" href="sbac-ui-kit/dist/css/sbac-ui-kit.min.css">-->
-        </div>
+<div th:fragment="styles">
+    <link rel="stylesheet"
+          href="https://raw.githack.com/SmarterApp/SBAC-Global-UI-Kit/develop/dist/css/sbac-ui-kit-wkhtmltopdf.min.css">
```
esotrope: This is a development config. This hosting service allows us to use github as a CDN (not something we want in prod)
Unfortunately the only working alternative to this now is to hardcode embedded CSS into this file.

I think a complete solution here will be to include the CSS as a dependency of the gradle project and host it out of /resources/css or /static/css for the pdf service to consume.

common/src/main/resources/templates/ica-report.html
esotrope: noted

PR #294, opened 2017-07-25
makes the pdf templates source the stylescheet from the webapp

esotrope:
Goes along with gitlab change: https://gitlab.com/fairwaytech/sbac-config-repo/merge_requests/59
Will merge when config repo is ready.

### DIFF
--- a/pdf-report-processor/src/main/resources/application.yml
+++ b/pdf-report-processor/src/main/resources/application.yml
@@ -67,3 +67,4 @@ app:
     options:
       dpi: 276
       page-size: Letter
+      user-style-sheet: http://localhost:8080/assets/css/sbac-ui-kit-wkhtmltopdf.min.css
--- a/common/src/main/resources/templates/fragments.html
+++ b/common/src/main/resources/templates/fragments.html
@@ -4,11 +4,6 @@
 </head>
 <body>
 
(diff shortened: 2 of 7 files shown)


pdf-report-processor/src/main/resources/application.yml
```diff
@@ -67,3 +67,4 @@ app:
     options:
       dpi: 276
       page-size: Letter
+      user-style-sheet: http://localhost:8080/assets/css/sbac-ui-kit-wkhtmltopdf.min.css
```
esotrope: with this solution the docker wkhtmltopdf service will be hitting the webapp service to get the pdf stylesheet.
the only alternative is embedding the css into the html template or creating our own custom dockerized wkhtmltopdf service rest interface that has the stylesheet.

Thoughts?

malaffoon: Seems like a good approach but perhaps a parking-lot brainstorming discussion?

bgsavage: Is there a way to use relative pathing here?

esotrope: if you did the host:port would resolve to the wkhtmltopdf service's (localhost:8082)

PR #297, opened 2017-07-25
Bugfix/629 ui issues

jtreuting:
- made the expander arrow on the students responses page consistent with item by points earned change
- move the date on the student response page back to the right as it used to be

### DIFF
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
@@ -9,15 +9,15 @@
 
 <div class="well">
   <div class="mb-md">
-    <div class="row">
-      <div class="col-md-9">
+    <div class="flex-children">
+      <div class="flex-child">
         <span class="tag tag-md {{colorService.getColor(getGradeIndex(assessment.grade))}}">
           <span class="label">{{assessment.grade | gradeDisplay}}</span>
           <span>{{assessment.name}}</span>
         </span>
       </div>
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -9,15 +9,15 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
 
 <div class="well">
   <div class="mb-md">
-    <div class="flex-children">
-      <div class="flex-child">
+    <div class="row">
+      <div class="col-md-9">
```
jtreuting: pinged @aaronsleeper on Slack to see if this was appropriate or not and waiting to hear.  i was not able to get the pull-right class to work with the flex-child class.  but wanting to make sure this change doesn't break something else since i'm not fully aware of what flex-child does

wtritch: The problem with col-md-9 is that if the screen shrinks below "md" size, it'll just be another column.  I think we're trying to avoid using bootstrap gridding.  Paging @aaronsleeper

jtreuting: yep just chatted with aaron and the HTML is right since the ml-auto class is meant to pull it right, but currently the flex-child class is more specific with the margin left so ml-auto is ignored.  will revert back to the HTML and Aaron will fix with CSS changes to address that.

PR #299, opened 2017-07-25
Implemented target info in item-info tab

bgsavage:
- Created api endpoint which pulls the item info by id.  api/targets/{id}
- Added targetId to existing items endpoint
- Now calling targets endpoint on item info tab load via cachingDataService.


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/ExamItemService.java
```diff
@@ -1,5 +1,8 @@
 package org.opentestsystem.rdw.reporting.search.item;
 
+import org.springframework.security.access.prepost.PreAuthorize;
+
+@PreAuthorize("hasAuthority('PERM_GROUP_PII_READ')")
```
bgsavage: Unrelated, but wanted to protect the rubrics / exemplars endpoint.

webapp/src/main/webapp/src/app/assessments/items/item-info/item-info.component.html
```diff
@@ -1,12 +1,11 @@
-<div>
+<div class="white-space-normal">
```
bgsavage: @aaronsleeper required since the table will cause white-space:nowrap, which I don't think we want inside the item-info tab.

webapp/src/main/webapp/src/app/assessments/items/item-info/item-info.component.ts
```diff
@@ -11,12 +12,17 @@ export class ItemInfoComponent implements OnInit {
...
-  constructor(private userService: UserService) { }
+  constructor(private userService: UserService, private cachingDataService: CachingDataService) { }
 
   ngOnInit() {
     this.userService.getCurrentUser().subscribe(user => {
       this.interpretiveGuide = user.configuration.interpretiveGuide
+    });
+
+    this.cachingDataService.get(`/targets/${this.item.targetId}`).subscribe(target => {
+      this.targetDescription = target.description;
```
bgsavage: Invocation of targets endpoint

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
```diff
@@ -84,6 +89,9 @@ export class ItemTabComponent implements OnInit {
     else if (tabName == 'Item Viewer') {
       this.loadItemViewer = true;
     }
+    else if (tabName == 'Item Information') {
```
bgsavage: I'm not sure why we are using string comparisons here? But just following the existing convention.

bgsavage:
Hm, i'm not sure why my entire Target folder did not get committed?

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
esotrope: wow planting sleeper bugs to discover when the page is translated are we?

wtritch: I think those strings are defined in the HTML and only used for metrics collection =)

PR #301, opened 2017-07-26
adds pdf report html and configuration

esotrope:
Todos:
1. Add scale score/cut point positioning segment widths
2. Add images/font awesome icons (could not embed whole font into css file)
3. Determine icons to use for MATH claims: 
Concepts & Procedures
Problem Solving
Communicating Reasoning

Qs:
1. Are we showing accommodations for IAB?
2. Should we show more than 2 past IABs? The list to the right of an IAB can get long.


common/src/main/resources/messages.properties
```diff
@@ -11,25 +11,31 @@ report.date.label=Date Taken:
...
+report.grade.06.color=blue-dark
 report.grade.07=7
 report.grade.07.label=G7
+report.grade.07.color=maroon
 report.grade.08=8
 report.grade.08.label=G8
+report.grade.08.color=green-dark
 report.grade.11=11
 report.grade.11.label=G11
+report.grade.11.color=blue-dark
```
esotrope: @aaronsleeper @wtritch are these the correct grade to color mappings for GX labels?

common/src/main/resources/messages.properties
```diff
@@ -83,24 +92,39 @@ report.ica.section.claim.header=Student Achievement by Claim
...
+report.claim.SOCK_LS.icon=fa-lightbulb-o
 report.claim.4-CR.name=Research & Inquiry
+report.claim.4-CR.icon=fa-question
 report.claim.1.name=Concepts & Procedures
-report.claim.SOCK_2.name=Problem Solving and Modeling & Data Analysis
+report.claim.1.icon=fa-question
+report.claim.SOCK_2.name=Problem Solving
+report.claim.SOCK_2.icon=fa-question
 report.claim.3.name=Communicating Reasoning
+report.claim.3.icon=fa-question
```
esotrope: @aaronsleeper what icons should we use for the Math claims?
Also, these icons are the only things requiring custom font. Until we have a robust pdf service can we make image versions of these icons or perhaps role a custom subset of the font-awesome font just for pdf?

common/src/main/resources/templates/ica-report.html
```diff
@@ -1,47 +1,106 @@
...
+        <hr class="mt-sm mb-sm"/>
+
+        <h2 th:text="#{report.ica.section.claim.header}"></h2>
+        <p th:text="#{('report.student-achievement-by-claim.grade.' + ${report.assessment.gradeCode} + '.subject.' + ${report.assessment.subject})(${report.student.firstName})}"></p>
+
+        <div class="row" th:with="columnClass=${report.assessment.claimCodes.size() == 4} ? 'col-xs-3' : 'col-xs-4'">
+            <div th:class="${columnClass}" th:each="claimCode : ${report.assessment.claimCodes}">
+                <h3 class="text-center" th:text="#{'report.claim.' + ${claimCode} + '.name'}"></h3>
+                <!-- TODO enable when fonts are embedded -->
+                <!--<span class="h4 gray-dark"><i class="fa fa-5x text-center" th:classappend="#{'report.claim.' + ${claimCode} + '.icon'}"></i></span><br />-->
```
esotrope: Disabled font icons until we can embed the icons.
This will cause an error because wkhtmltopdf cant find the file.

common/src/main/resources/templates/ica-report.html
```diff
@@ -1,47 +1,106 @@
...
-                    <div th:text="#{report.accommodations.label}"></div>
-                    <!-- <li th:each="accommodation : ${report.exam.accommodationCodes}>" -->
+        <div class="row" th:if="${!report.exam.accommodationCodes.isEmpty()}">
+            <div class="col-xs-12">
+                <div class="well">
+                    <div class="well-body">
+                        <p class="gray-darker" th:text="#{report.accommodations.label}"></p>
+                        <!-- TODO: print over accommodation codes when markup is clear -->
+                        <!--<p><strong>Alternative Response<br />
+                            Printed passages/stimuli</strong></p>-->
```
esotrope: @aaronsleeper I wasn't clear on the markup here, the accommodations will need to be spit out in some sort of list and this looks like a list of <strong> with <br> in between. is that right?

webapp/src/main/resources/application.yml
```diff
@@ -85,6 +85,9 @@ app:
     options:
       dpi: 276
       page-size: Letter
+      # Hotfix for issue with wkhtmltopdf 0.12.2.1 ~ 0.12.3 (pdf markup/styles were tested against 0.12.4)
+      # See https://github.com/wkhtmltopdf/wkhtmltopdf/issues/3241
+      zoom: 0.7
```
esotrope: @aaronsleeper This setting was about as close as i could get to making 0.12.2.1 (the version in the docker image) to produce the pdf you produced with 0.12.4

esotrope:
current output:

screen shot 2017-07-25 at 8 02 53 pm

screen shot 2017-07-25 at 8 03 46 pm

common/src/main/resources/templates/iab-report.html
```diff
@@ -6,34 +6,78 @@
...
+                    <div class="well assessment-latest-result" data-width="60" th:with="assessment=${entry.key},exam=${entry.value[0]}">
+                        <div class="well-body">
+                            <div class="row">
+                                <div class="col-xs-9">
+                                    <span class="tag h3" th:classappend="#{'report.grade.' + ${assessment.gradeCode} + '.color'}"><span class="label" th:text="#{'report.grade.' + ${assessment.gradeCode} + '.label'}"></span> <span th:text="${assessment.name}" th:remove="tag"></span></span>
+                                    <div class="scale-score-graph individual">
+                                        <div class="bar-container">
+                                            <span class="scale-score" data-left="42"><span class="score label" th:classappend="#{'report.iab.level.' + ${exam.scaleScore.level} + '.color'}" th:text="${exam.scaleScore.value}"></span> <span class="error-band">&plusmn; <span th:text="${#numbers.formatDecimal(exam.scaleScore.standardError, 0, 'NONE', 0, 'NONE')}" th:remove="tag"></span></span></span>
```
esotrope: I went ahead and did a Math.floor() on scale score std err. Is this what we want?
I noticed if i left the decimal in the number wrapped to a new line in the reports

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepository.java
```diff
@@ -65,4 +80,12 @@ public IcaReport findOneByStudentIdAndSchoolYear(
         }
     }
 
+    private List<String> getAccommodationCodes(final ResultSet row) throws SQLException {
+        final String pipeDelimitedCodes = row.getString("available_accommodation_codes");
+        if (pipeDelimitedCodes == null) {
+            return null;
+        }
+        return ImmutableList.copyOf(pipeDelimitedCodes.split("|"));
```
esotrope: Thinking about making this a utility. Pretty sure we do pipe delim parsing somewhere else too.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepository.java
```diff
@@ -65,4 +80,12 @@ public IcaReport findOneByStudentIdAndSchoolYear(
         }
     }
 
+    private List<String> getAccommodationCodes(final ResultSet row) throws SQLException {
+        final String pipeDelimitedCodes = row.getString("available_accommodation_codes");
+        if (pipeDelimitedCodes == null) {
```
malaffoon: Might want to use `if (!hasText(pipeDelimitedCodes))` to catch empty strings.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepository.java
malaffoon: this seems readable enough ...

common/src/main/resources/templates/iab-report.html
malaffoon: it's probably more correct to round it than to floor it

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepository.java
esotrope: sounds good

common/src/main/resources/templates/iab-report.html
esotrope: k I'll checkout how to do that in thymeleaf

common/src/main/resources/messages.properties
aaronsleeper: what's important is that two adjacent grades don't have the same color rather than a specific color being assigned to a specific grade. so @wtritch wrote a function to iterate over a list of colors and assign them. as long as adjacent grades are different, it's correct. @jessicacolton chime in if you disagree.

common/src/main/resources/messages.properties
esotrope: *update looks like http://fontello.com/ will allow us to roll tiny font-awesome with just the icons we need

webapp/src/main/resources/application.yml
aaronsleeper: these seem to need some more work . the overlapping elements are the only critical fixes . it'd be nice to fix the other differences if we have time . i suspect perfection isn't required for UAT, @jtreuting thoughts?

common/src/main/resources/templates/ica-report.html
aaronsleeper: technically it's a list, yes. we can just wrap each item in a <p> though

common/src/main/resources/messages.properties
aaronsleeper: yes. that's an option if we don't want to include the whole font. we can also save images of the icons

common/src/main/resources/messages.properties
bgsavage: I would think we'd want the colors to be consistent between web app / pdf?  @esotrope  I'm not sure if there's an easy way for you to reuse the same function to ensure consistency in case the function changes.

wtritch: wait, we have a general color wheel.  Here's a snippet:
`class="tag tag-xs {{colorService.getColor(getGradeIdx(byGrade.grade))}}"`

wtritch: The idea is that any iterable "thingy" (grade, assessment type, subject, etc, etc) will always use the same color, the colors look good next to each other, and we only manage a single list of "identity" colors that can handle any number of thingies.  This is so that we don't have to keep going to the UX team to get a color for grade "PS" or "IT" or any of the other 19 potential grades.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepository.java
wtritch: I'm a fan of the Guava Splitter class... but I'm a Guava-junkie.

common/src/main/resources/messages.properties
esotrope: For now i will just make the colors consistent

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/processor/DefaultExamReportPdfProcessor.java
```diff
@@ -16,22 +22,37 @@
...
     public byte[] process(
             final Object report,
             final AbstractExamReportRequest<PrintOptions> request) {
-        final Map<String, Object> model = ImmutableMap.of("report", report, "request", request);
+        final Map<String, Object> model = ImmutableMap.of("report", report, "request", request, "accommodations", getAccommodations(report, request));
         final byte[] html = templateProcessor.process(request.getAssessmentType().template(), model, request.getLanguage());
         return htmlToPdfProcessor.process(html, request.getOptions());
     }
 
+    public List<String> getAccommodations(
```
esotrope: This will likely find a new and better home in a follow up PR

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
```diff
@@ -70,13 +73,18 @@ private QueryUtils() {
...
+     * @return the delimited values split or null if the delimited values are null or empty
      */
-    public static String createMatcher(final String value) {
-        return "%" + value + "%";
+    public static List<String> safeSplit(final String delimitedValues, final String delimiter) {
+        if (isNullOrEmpty(delimitedValues)) {
+            return null;
+        }
+        return ImmutableList.copyOf(Splitter.on(delimiter).split(delimitedValues));
     }
```
esotrope: @wtritch @malaffoon added utility for splitting DB results since we do it in two separate places

PR #302, opened 2017-07-26
disable + icon for adding assessments when they have all been selected

jtreuting:


bgsavage:
Hmm, should we restrict this?  What if they want to just open the assessment flyout to see the groupings?

webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -67,7 +67,8 @@ <h2 class="h3 gray-darkest pull-left">
             <span class="icon" [hidden]="showOnlyMostRecent"><i class="fa fa-close"></i></span>
           </a>
           <div>
-            <a *ngIf ="!expandAssessments" class="btn btn-default btn-sm img-circle icon-only" (click)="expandAssessments=true"
+            <a *ngIf ="!expandAssessments" class="btn btn-default btn-sm img-circle icon-only"
+               (click)="expandAssessments=!allSelected" [attr.disabled]="allSelected?'disabled':null"
```
bgsavage: I don't think you need `attr.` here.

bgsavage: Also I think it would just be [disabled]="allSelected"

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -108,10 +108,14 @@ export class AssessmentsComponent implements OnInit {
...
 
   get selectedAssessments() {
+    this._isAllSelected = false;
     if (this.showOnlyMostRecent && this._assessmentExams)
       return this._assessmentExams.map(x => x.assessment);
-    else if (this.availableAssessments)
-      return this.availableAssessments.filter(x => x.selected);
+    else if (this.availableAssessments) {
+      let assessments = this.availableAssessments.filter(x => x.selected);
+      this._isAllSelected = (assessments.length == this.availableAssessments.length);
```
bgsavage: Seems odd that our getter would have this side effect setting `_isAllSelected`.  Is there a better place to make this check?

jtreuting: i can look for one, my thinking was that since it is doing the filter and not saving it there shouldn't be a need to redo that processing.

webapp/src/main/webapp/src/app/assessments/assessments.component.html
jtreuting: when i used [disabled] it said ti wasn't a valid attribute for an `a` element.  And for the disabled attribute the true/false does not translate properly and wil always set the `disabled` attribute on the anchor tag.

esotrope: Looks like the ```<a>``` element is already class ```btn``` shoudn't we just make this a ```<button>``` then ```[disabled]``` would work i think

jtreuting: i can look at that, i'm thinking it will mess up the look of it and I don't really want to iterate over the style if it's to be able to set it to disabled with a true/false value instead of a conditional.  but, yeah let me see if that swaps out nicely.  good idea.

jtreuting: the button looks exactly the same, so switching to that.  good idea, thanks

esotrope: 👍

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
jtreuting: moved to the `selectedAssessmentsChanged` method .. much better place for it.  thanks.

bgsavage: 👍

PR #306, opened 2017-07-27
adds width/positioning of scale score graphs in pdf reports

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Gets the position and width attributes necessary for rendering a scale score graph
+     *
+     * @param exam       the exam to use for computing the scale score position
+     * @param assessment the assessment to use for computing the cut point bar attributes
+     * @return the position and width attributes necessary for rendering a scale score graph
+     */
+    public ScaleScoreView getScaleScoreView(final Exam exam, final Assessment assessment) {
```
esotrope: This method has the meat of the change. I have never coded math ops in Java before, so critique and suggestions are very welcome.

common/src/main/resources/templates/iab-report.html
```diff
@@ -67,8 +69,8 @@
...
                         <div class="well-body">
                             <ul class="previous-results">
-                                <li th:with="assessment=${entry.key}" th:each="exam : ${entry.value}"><span  th:if="${examStat.index} > 0"><span class="previous-result"><span class="date" th:text="${#dates.format(exam.dateTime, 'MM-dd-yyyy')}"></span> <span
-                                        class="scale-score"><span th:text="${exam.scaleScore.value}" th:remove="tag"></span> <span class="error-band">&plusmn; <span th:text="${#numbers.formatInteger(exam.scaleScore.standardError, 0, 'NONE')}" th:remove="tag"></span></span></span> <span
+                                <li th:with="assessment=${entry.key}" th:each="exam : ${entry.value}"><span th:if="${examStat.index} > 0"><span class="previous-result"><span class="date" th:text="${#dates.format(exam.dateTime, 'MM-dd-yyyy')}"></span> <span
+                                        class="scale-score"><span th:text="${exam.scaleScore.value}" th:remove="tag"></span> <span class="error-band">&plusmn; <span th:text="${support.round(exam.scaleScore.standardError)}" th:remove="tag"></span></span></span> <span
```
esotrope: Now uses Math.round() with default rounding which i think is HALF_EQUALS

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param value the integer to get the last two digits from
+     * @return the last two digits of the given integer
+     */
+    public String getLastTwoDigits(final int value) {
+        final String nextYear = Integer.toString(value);
+        return nextYear.substring(nextYear.length() - 2);
+    }
+
+    /**
+     * Gets the product of {@link Math#round(double)} on the given value
```
malaffoon: nit: use of the word `product` here is confusing; i know what you mean but in math-like functions this will be misread; i spent a few seconds trying to figure out what was being multiplied

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Gets the last two digits of the given integer
+     *
+     * @param value the integer to get the last two digits from
+     * @return the last two digits of the given integer
+     */
+    public String getLastTwoDigits(final int value) {
+        final String nextYear = Integer.toString(value);
```
malaffoon: Is this being used in a tight loop or anything? There are much more efficient ways of doing this but i won't bother you with them if it isn't being used very often.

esotrope: Nope used once per report generation. But don't just tease me like that, now I'm interested in the efficient ways.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
esotrope: will change to *result

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
malaffoon: Hmm, i misread the method signature. If you're returning a string anyway this is fine. 

I was thinking you wanted the value of the last two digits of the integer. In that case you'd just want to do `value % 100` or some such.

esotrope: gotcha - this is to turn a year into the last two digits of the year basically

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
```diff
@@ -0,0 +1,137 @@
...
+    public ScaleScoreView getScaleScoreView(final Exam exam, final Assessment assessment) {
+
+        final List<Integer> cutPoints = assessment.getType() == AssessmentType.ICA
+                ? assessment.getCutPoints()
+                // Process IAB cut points for use in display
+                : newArrayList(
+                assessment.getCutPoints().get(0),
+                // The second and third cut point are the middle cut point +/- the product of the individual exam
+                // result scale score standard error * 1.5
+                toIntExact(assessment.getCutPoints().get(2) - Math.round(exam.getScaleScore().getStandardError() * 1.5)),
```
malaffoon: toIntExact seems overkill given what we know about the input values but i guess it doesn't hurt

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Gets the percent distance covered by the given position along the given length in whole percent (0 - 100)
+     * Assumes position and length are positive values
+     *
+     * @param position the position along the length
+     * @param length   the distance
+     * @return the percent distance covered by the given position along the given length in whole percent (0 - 100)
+     */
+    private int getPercendRounded(final float position, final float length) {
```
malaffoon: typo in name?

esotrope:
screen shot 2017-07-26 at 5 35 38 pm

@aaronsleeper 
Both IAB and ICA can run into this stderr overlap issue
Do we want the +/- to change sides after data-left is > 90 or something?

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
malaffoon: The variable name, nextYear, implies something about the value that isn't really appropriate. I'd use something more generic.

esotrope:
screen shot 2017-07-26 at 5 37 51 pm

@jtreuting @aaronsleeper 
Not sure if this is something that can happen IRL (I think max 3 IABs of a type can be taken in a year)
But either way this is how the current view deals with the situation 👍

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
malaffoon: Seems like this method should have a robust set of unit tests.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
esotrope: yup thanks - that was left over

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportViewSupport.java
esotrope: Yes! that is what is what is coming next

esotrope:
@jtreuting @aaronsleeper do we have a strategy for handling unscored IAB? ICA? and ICA claims?

jtreuting:
We've always says that unscored will just be represented by a dash.  But that was more when displayed in a table.  It might be as simple as saying "not scored" in place of where the visualization would be.

In practice CA will not be sending unscored but we need to be able to handle it.

PR #307, opened 2017-07-27
Feature/ui kit from artifactory

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/307)


webapp/src/main/webapp/.npmrc
```diff
@@ -0,0 +1 @@
+@sbac:registry=https://airdev.jfrog.io/airdev/api/npm/npm
```
rcordeau: Adding this file (.npmrc) to just the package level, which will allow npm to locate the sbac-ui-kit

PR #312, opened 2017-07-27
DWR-643

malaffoon:
More help in diagnosing wkhtmltopdf issues in k8s.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/status/WkhtmltopdfStatusIndicator.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/status/WkhtmltopdfStatusIndicator.java
@@ -1,0 +1,51 @@
+package org.opentestsystem.rdw.reporting.status;
+
+import org.opentestsystem.rdw.common.status.AbstractStatusIndicator;
+import org.opentestsystem.rdw.common.status.DiagnosticLevel;
+import org.opentestsystem.rdw.common.status.Rating;
+import org.opentestsystem.rdw.common.status.Status;
+import org.opentestsystem.rdw.reporting.common.report.client.Request;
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/status/WkhtmltopdfStatusIndicator.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public String name() {
+        return "wkhtmltopdf-service";
+    }
+
+    @Override
+    protected void doStatusCheck(final Status.Builder builder, final int level) {
+        final byte[] html = "<!DOCTYPE html><html lang=\"en\"><head><meta charset=\"utf-8\"><title>Status Check</title></head><body></body></html>".getBytes(utf8);
```
malaffoon: I just wanted an empty html doc that it would convert; this works but any advice appreciated.

webapp/src/main/java/org/opentestsystem/rdw/reporting/status/WkhtmltopdfStatusIndicator.java
```diff
@@ -0,0 +1,51 @@
...
+        final byte[] pdf = responseTime(builder, 500, () -> {
```
malaffoon: Not entirely sure what a reasonable response time is for an empty doc but, on my system, it takes about 300ms.

webapp/src/main/java/org/opentestsystem/rdw/reporting/status/WkhtmltopdfStatusIndicator.java
esotrope: Looks good.

PR #315, opened 2017-07-28
changed statistics to a setter so that the corresponding display perc…

jtreuting:
…ent can be recalculated when it changes

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit } from "@angular/core";
+import { Component, Input } from "@angular/core";
 import { AssessmentExam } from "../model/assessment-exam.model";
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -33,16 +34,20 @@ export class AverageScaleScoreComponent {
   public assessmentExam: AssessmentExam;
 
   @Input()
-  public statistics: ExamStatistics;
+  set statistics(value: ExamStatistics) {
+    this._statistics = value;
 
-  constructor(private scaleScoreService: ScaleScoreService) {
+    if (this._statistics) {
+      this.levelPercents = this.scaleScoreService.calculateDisplayScoreDistribution(this._statistics.percents);
```
bgsavage: I have a filter function where I think all these types of sets should be happening together (in one place)

bgsavage: Oh nevermind I see this is a separate component now.

PR #316, opened 2017-07-28
Feature/pdf download popover

esotrope:
@aaronsleeper @jessicacolton I introduced your recommended changes for the download button & popover. Additionally I changed the button color. Let me know what you think. I'm pretty sure you'll have suggestions about responsive compatibility and the popover jutting out to the right.

screen shot 2017-07-27 at 10 26 58 pm

*Added the pill header styling that is in the PDF reports

screen shot 2017-07-27 at 10 18 13 pm

### DIFF
--- a/common/src/main/resources/templates/fragments.html
+++ b/common/src/main/resources/templates/fragments.html
@@ -5,11 +5,17 @@
 <body>
 
 <div th:fragment="styles">
-    <link rel="stylesheet"
-          href="https://cdn.rawgit.com/SmarterApp/SBAC-Global-UI-Kit/3f548f92980dbfcc7fe08d78944ea610eaca27d6/dist/css/sbac-ui-kit-wkhtmltopdf.min.css">
+    <link rel="stylesheet" href="https://cdn.rawgit.com/SmarterApp/SBAC-Global-UI-Kit/3f548f92980dbfcc7fe08d78944ea610eaca27d6/dist/css/sbac-ui-kit-wkhtmltopdf.min.css">
     <style>
-        @media print {
-            .accommodations {page-break-inside: avoid;}
+        /* TODO: move to styelsheet */
+        header nav.navbar .navbar-header {
+            display: block;
+        }
+        header .application-title {
+            margin: 12px 0 0 0;
+        }
+        .assessment-latest-result .performance-level{
+            display:inline-block;
         }
     </style>
 </div>
@@ -20,7 +26,7 @@
             <nav class="navbar navbar-default">
                 <div class="navbar-header">
                     <div class="navbar-brand"></div>
-                    <span class="application-title ml-auto" th:text="#{${prefix} + '.title'}"></span>
+                    <span class="application-title text-right pull-right mt-sm" th:text="#{${prefix} + '.title'}"></span>
                 </div>
             </nav>
         </div>
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -56,6 +56,7 @@ services:
       - spring.profiles.active=local-docker
       - CONFIG_SERVICE_ENABLED=true
       - CONFIG_SERVICE_URL=http://config-server:8888
+      - app.wkhtmltopdf.options.zoom=0.75
     extra_hosts:
       - "dwhost:$DATAWAREHOUSE_HOST"
 
@@ -76,6 +77,6 @@ services:
 #      - "dwhost:$DATAWAREHOUSE_HOST"
 
   wkhtmltopdf:
-    image: openlabs/docker-wkhtmltopdf-aas
+    image: fwsbac/wkhtmltopdf-image:latest
     ports:
       - 8082:80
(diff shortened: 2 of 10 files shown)


webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -1,58 +1,45 @@
...
+        </label>
+
+        <label class="ml-xs">
+          <button class="btn btn-info" (click)="form.ngSubmit.emit()">{{'labels.reports.download' | translate}}</button>
+        </label>
+
       </div>
+    </form>
+
+    <div *ngIf="errorMessage" class="alert alert-danger">
```
esotrope: Not clear on what to do for error conditions. Will an alert-danger with "Unable to generate report at this time" work?

webapp/src/main/webapp/src/styles.less
```diff
@@ -86,3 +86,14 @@ footer {
...
+  display: inline-block;
+}
+.report-download-popover {
+  display: block;
+  max-width: 440px;
+  width:auto;
+  select {
+    font-weight: normal;
+  }
+}
```
esotrope: @aaronsleeper @jessicacolton styles to make the popover width fit the contents

bgsavage:
I wonder if we should consider moving the advanced filters to a similar popover for A.  consistency and B. save on vertical screen real estate to see real-time changes to the data

webapp/src/main/webapp/src/app/report/report-download.component.html
bgsavage: Really should be *ngIf invalid for this specific reason with the contents pointing to that specific error message in translate.  How many potential errors can you have?

esotrope: do we use form invalid when its an error server response? This is if something goes terribly wrong and the pdf service can't be reached.

bgsavage: Well my point here is we should separate the concern of whether or not to show the error message from what the error message should be.  Keep the translate in the html when possible and control visibility of the error message with a different boolean.  This is the approach we've been taking in other components until now.  Sorry about the "invalid" terminology confusion.

webapp/src/main/webapp/src/styles.less
aaronsleeper: @esotrope was the font bolded previously? i feel like i'm missing context

esotrope: yeah somehow the font weight was bold on the select boxes in the popover.

webapp/src/main/webapp/src/app/report/report-download.component.html
esotrope: Yeah totally on board with that - I'll put in a placeholder message

webapp/src/main/webapp/src/app/report/student-report-download.component.ts
```diff
@@ -41,8 +37,8 @@ export class StudentReportDownloadComponent extends ReportDownloadComponent {
           saveAs(download.content, download.name);
         },
         (error: any) => {
-          // TODO: add handler for 404/500 errors
-          this.errorMessage = `${error.status}: ${error.statusText}`;
+          // TODO: add handler for 404/500 errors - consider publishing to global notification "snack drawer"
+          console.error(error);
```
esotrope: @wtritch @bgsavage got rid of the error message in the popover. Do we still have some global message service where we can shoot an error message to the user? I'm thinking that might be most appropriate if they async download request fails.

esotrope:
*Merging now
Follow up PR will include suggested vertical popover layout and left/down popover direction.

common/src/main/resources/templates/fragments.html
```diff
@@ -5,11 +5,17 @@
 <body>
 
 <div th:fragment="styles">
-    <link rel="stylesheet"
-          href="https://cdn.rawgit.com/SmarterApp/SBAC-Global-UI-Kit/3f548f92980dbfcc7fe08d78944ea610eaca27d6/dist/css/sbac-ui-kit-wkhtmltopdf.min.css">
+    <link rel="stylesheet" href="https://cdn.rawgit.com/SmarterApp/SBAC-Global-UI-Kit/3f548f92980dbfcc7fe08d78944ea610eaca27d6/dist/css/sbac-ui-kit-wkhtmltopdf.min.css">
```
esotrope: @wtritch one thing we will want to add to the docker image is downloading a version of the ui kit from artifactory and using its assets

docker/docker-compose.yml
```diff
@@ -76,6 +77,6 @@ services:
 #      - "dwhost:$DATAWAREHOUSE_HOST"
 
   wkhtmltopdf:
-    image: openlabs/docker-wkhtmltopdf-aas
+    image: fwsbac/wkhtmltopdf-image:latest
```
esotrope: now points at will's wkhtmltopdf 0.12.4 image

PR #319, opened 2017-07-28
Now formatting target code to remove trailing dash.

bgsavage:
This is a global change for the app and will propagate all places in the app where target is displayed & used.
Added unit test.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
@@ -138,8 +138,8 @@ export class AssessmentExamMapper {
     uiModel.id = apiModel.id;
     uiModel.bankItemKey = apiModel.bankItemKey;
     uiModel.position = apiModel.position;
-    uiModel.claim = apiModel.claimCode
-    uiModel.target = apiModel.targetCode;
+    uiModel.claim = apiModel.claimCode;
+    uiModel.target = this.formatTarget(apiModel.targetCode);
     uiModel.targetId = apiModel.targetId;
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -151,6 +151,14 @@ export class AssessmentExamMapper {
     return uiModel;
   }
 
+  formatTarget(target) {
```
bgsavage: Ideally this should be private inside this mapper and is only public to make it easy to unit test.  If we have another formatting function, we should consider creating an AssessmentExamFormatter to facilitate unit testing of all the formatting functions.

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -151,6 +151,14 @@ export class AssessmentExamMapper {
...
+    let dashIndex = target.indexOf("-");
```
esotrope: are all target codes two segments?
wondering if this should be lastIndexOf() or not

bgsavage: That's a good point, let me look and see what we have in UAT and ETS

bgsavage: I did queries in UAT and ETS counting the dashes and didn't find any where the number of dashes was more than one, so I think we're good here.

bgsavage:
It does not since we are using target id to look it up.  Good thought though

PR #322, opened 2017-07-28
connections stab-in-the-dark

malaffoon:
i can't build anything locally so i'm submitting this blind; basically i'm hoping the reduced connections helps alleviate the connection problem; this isn't the final fix, just a hack to get all the tests running again


webapp/src/test/resources/application-test.yml
```diff
@@ -13,6 +13,10 @@ app:
   datamart:
     datasource:
       url: jdbc:mysql://localhost:3306/reporting_test?useSSL=false&useLegacyDatetimeCode=false
+      initialSize: 4
+      maxActive: 10
+      maxIdle: 10
+      maxWait: 10000
```
bgsavage: looks promising

PR #324, opened 2017-07-29
fixes problem solving text

esotrope:
@aaronsleeper @jessicacolton should we set a height for the titles so the claim views will line up?

screen shot 2017-07-28 at 5 11 20 pm


webapp/src/main/webapp/src/app/report/student-report-download.component.ts
```diff
@@ -34,6 +34,7 @@ export class StudentReportDownloadComponent extends ReportDownloadComponent {
         (error: any) => {
           // TODO: add handler for 404/500 errors - consider publishing to global notification component
           console.error(error);
+          this.subscription = null;
```
esotrope: unrelated change - the "completed" callback wasn't firing after the "error" callback

common/src/main/resources/messages.properties
```diff
@@ -126,7 +126,7 @@ report.claim.4-CR.name=Research & Inquiry
 report.claim.4-CR.icon=fa-question
 report.claim.1.name=Concepts & Procedures
 report.claim.1.icon=fa-book
-report.claim.SOCK_2.name=Problem Solving
+report.claim.SOCK_2.name=Problem Solving & Modelling / Data Analysis
```
bgsavage: Isn't modeling one `l`?

esotrope: thanks - fixing

webapp/src/main/webapp/src/app/report/student-report-download.component.ts
bgsavage: Hm, that's odd, I though the completed callback was supposed to be like a `finally` block.

PR #325, opened 2017-07-29
Now showing the standard error as a rounded value in the UI.

bgsavage:
I only had to update the component the tables use,  the aggregate scale score was already rounding.

I used this script to assign random decimal values to my standard errors:
```
update exam
set scale_score_std_err = scale_score_std_err + round(rand() * 0.49 + 0.01, 4)
```


webapp/src/main/webapp/src/app/assessments/results/scale-score.component.html
```diff
@@ -1 +1,4 @@
-<span class="scale-score"><span class="score">{{ score ? score : '-' }}</span> <span *ngIf="score" class="error-band">± {{ standardError }}</span></span>
+<span class="scale-score">
+  <span class="score">{{ score ? score : '-' }}</span>
+  <span *ngIf="score" class="error-band">± {{ standardError | number:'1.0-0' }}</span>
```
bgsavage: The real change is adding the number pipe here.  The rest is just formatting for readability.

PR #329, opened 2017-07-31
When route update is required from student results, now navigating re…

bgsavage:
…lative to the parent's parent url which fixes issue when coming from /groups and /schools.  Also no longer updating route when basic filters have not changed (Changing advanced filters should not cause the route to change in the first place)

### DIFF
--- a/webapp/src/main/webapp/src/app/student/results/student-results.component.ts
+++ b/webapp/src/main/webapp/src/app/student/results/student-results.component.ts
@@ -72,8 +72,18 @@ export class StudentResultsComponent implements OnInit {
    * Handle a filter state change.
    */
   onFilterChange(): void {
-    this.updateRoute();
+    if(this.hasBasicFilterChanged()) {
+      this.updateRoute();
+    }
+
     this.applyFilter();
+  }
+
+  hasBasicFilterChanged(): boolean {
+    let params:any = this.route.snapshot.params;
+
+    return this.filterState.schoolYear != params.schoolYear
+      || this.filterState.subject != params.subject;
   }
 
   isCollapsed(assessmentType: AssessmentType, subject: string): boolean {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -186,8 +196,11 @@ export class StudentResultsComponent implements OnInit {
       params.subject = this.filterState.subject;
     }
 
-    //Update the current route
-    this.router.navigate(['students', this.examHistory.student.id, params]);
+    let navigationExtras = this.route.parent.parent.snapshot.url.length > 0
+      ? { relativeTo: this.route.parent.parent }
```
bgsavage: This is the real fix.  The bug was because we were at /schools/id/students/id (or /groups/id/students/id) and then navigating to /students/id (which cleared the adv filter).  Now navigating relative to the parent's parent if there is one.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -72,10 +72,20 @@ export class StudentResultsComponent implements OnInit {
    * Handle a filter state change.
    */
   onFilterChange(): void {
-    this.updateRoute();
+    if(this.hasBasicFilterChanged()) {
+      this.updateRoute();
+    }
```
bgsavage: Added this optimization as well to not update the route when an advanced filter has been changed which we don't keep in the route anyway.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: Can't you just navigate relative to the parent?

bgsavage: The if check is for the case when we came from the home page and there is not grandparent.

PR #330, opened 2017-07-31
fixes and tests double apostrophe occurrence in messsages.properties

esotrope:


common/src/main/resources/messages.properties
```diff
@@ -38,15 +38,15 @@ report.subject.MATH.name=Math
 report.subject.ELA.name=ELA/Literacy
 
 report.scale-score.error-band=Error Band
-report.scale-score.description=A student''s test score can vary if the test is taken several times. If this student was tested again, it is likely that the scale score would fall within this range.
+report.scale-score.description=A student's test score can vary if the test is taken several times. If this student was tested again, it is likely that the scale score would fall within this range.
```
esotrope: So the discovery here is that messages with substitutions like:
```{0}'s``` need two consecutive ```''``` apostrophes for a single apostrophe to appear in the view.
And for unformatted messages without ```{0}``` we do not need to escape apostrophes with a double apostrophe....

File: common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/MessageFormattingTest.java [hunk 2]
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * In property files, formatted messages require single apostrophes to be escaped by a leading single apostrophe.
+     * Unformatted messages do not require escaping single apostrophes.
+     * This test asserts after processing of messages, there is no instance of consecutive apostrophes in the message.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void itShouldNotShowConsecutiveApostrophes() throws Exception {
```
esotrope: This test makes sure all messages we ever put in the messages[_lang].properties with {0} do not end up with consecutive single apostrophes and messages without {0} also do not end up with consecutive single apostrophes.

PR #333, opened 2017-07-31
Popover for disabled text buttons.

bgsavage:
Would be good to eventually create this as a component as this code could potentially be repeated in a lot of places.

Buttons affected:
 - Instructional Resources (button on assessments well)
 - Instructional Resources (menu item)
 - Items By Points Earned

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
@@ -4,18 +4,20 @@
     <div class="mb-md">
         <!-- Top Right Buttons -->
         <div class="pull-right text-right">
-            <button
-              class="btn btn-default btn-xs"
-              angulartics2On="click"
-              angularticsEvent="InstructionalResource"
-              angularticsCategory="AssessmentResults"
-              [angularticsProperties]="{label: assessmentExam.assessment.name}"
-              [disabled]="!assessmentExam.assessment.hasResourceUrl"
-              (click)="openInstructionalResource()"
-              title="{{ assessmentExam.assessment.hasResourceUrl ? '' : 'labels.menus.resources-disabled-message' | translate }}"
-            >
+          <span popover="{{ 'labels.menus.resources-disabled-message' | translate }}" triggers="{{assessmentExam.assessment.hasResourceUrl ? '': 'mouseenter:mouseleave'}}">
+                        <button
+                          class="btn btn-default btn-xs"
+                          angulartics2On="click"
+                          angularticsEvent="InstructionalResource"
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -4,18 +4,20 @@
...
+                          angulartics2On="click"
+                          angularticsEvent="InstructionalResource"
+                          angularticsCategory="AssessmentResults"
+                          [angularticsProperties]="{label: assessmentExam.assessment.name}"
+                          [disabled]="!assessmentExam.assessment.hasResourceUrl"
+                          (click)="openInstructionalResource()"
+                        >
               <i class="fa fa-list-alt"></i> {{'labels.groups.results.assessment.instruct' | translate }}
             </button>
+          </span>
```
bgsavage: Why do I have to wrap with a span?  I always bump up against this, and wish it wasn't so:

https://github.com/twbs/bootstrap/issues/9294

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -59,17 +61,18 @@
...
         <h3 class="gray-darkest mb-sm">
           {{'labels.groups.results.assessment.exams.title' | translate}}
 
-          <button *ngIf="assessmentExam.assessment.isInterim"
-                  [disabled]="!hasItemLevelData"
-                  title="{{ hasItemLevelData ? '' : 'messages.no-item-level-data' | translate }}"
-                  class="btn btn-default btn-xs pull-right" (click)="viewItemsByPoints(true)">
-            {{'labels.groups.results.assessment.button-items' | translate}} <i class="fa fa-chevron-right"></i>
-          </button>
+          <span class="pull-right" popover="{{ 'messages.no-item-level-data' | translate }}" triggers="{{ hasItemLevelData ? '': 'mouseenter:mouseleave'}}">
```
jtreuting: assuming a trigger of '' means none so it won't go

bgsavage:
On the menu item:
image


On the button: 
image

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
bgsavage: correct, empty string trigger means it won't trigger, it's how they show to implement manual triggers in the docs (even though I'm not doing a manual trigger here, in this case, I just don't want it to trigger itself)

bgsavage: If you do empty text popover like how we were doing for `title` the popover bubble still pops just with no text.

PR #334, opened 2017-07-31
Bugfix/handle missing exam items

wtritch:
The student responses page was having issues when the examitems response payload contained more assessment items than exam items.  This change handles both -1 scores and missing exam items when displaying user responses.

### DIFF
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses-item.model.ts
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses-item.model.ts
@@ -6,5 +6,6 @@ import { AssessmentItem } from "../../assessments/model/assessment-item.model";
  */
 export class StudentResponsesAssessmentItem {
   assessmentItem: AssessmentItem;
+  score: number;
   correctness: number;
 }
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses-item.model.ts
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses-item.model.ts
@@ -8,4 +8,5 @@ export class StudentResponsesAssessmentItem {
   assessmentItem: AssessmentItem;
   score: number;
   correctness: number;
+  response: string;
 }
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
@@ -53,7 +53,7 @@
     </p-column>
     <p-column field="score" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true">
       <ng-template let-score="rowData.score" pTemplate="body">
-        <span *ngIf="score !== null">score</span>
+        <span *ngIf="score !== null">{{score}}</span>
         <span *ngIf="score === null">&mdash;</span>
       </ng-template>
     </p-column>
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
@@ -70,7 +70,7 @@
     </p-column>
     <ng-template let-item pTemplate="rowexpansion">
       <item-tab [item]="item.assessmentItem"
-                [response]="item.assessmentItem.scores[0].response"
+                [response]="item.response"
                 [position]="item.assessmentItem.position"
                 [showItemDetails]="assessment.isInterim"
                 [showStudentScores]="false"></item-tab>
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
@@ -51,7 +51,12 @@
         <span *ngFor="let standard of standards" class="label border-only blue mr-xs">{{standard}}</span>
       </ng-template>
     </p-column>
-    <p-column field="assessmentItem.scores.0.points" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true"></p-column>
+    <p-column field="score" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true">
+      <ng-template let-score="rowData.score" pTemplate="body">
+        <span *ngIf="score !== null">score</span>
+        <span *ngIf="score === null">&mdash;</span>
+      </ng-template>
+    </p-column>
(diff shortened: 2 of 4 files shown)


webapp/src/main/webapp/src/app/student/responses/student-responses-item.model.ts
```diff
@@ -6,5 +6,7 @@ import { AssessmentItem } from "../../assessments/model/assessment-item.model";
  */
 export class StudentResponsesAssessmentItem {
   assessmentItem: AssessmentItem;
+  score: number;
   correctness: number;
+  response: string;
 }
```
wtritch: I didn't want to do null-safe de-referencing in the HTML, so I moved these values into the view model.

webapp/src/main/webapp/src/app/student/responses/student-responses.component.ts
```diff
@@ -44,9 +44,11 @@ export class StudentResponsesComponent implements OnInit {
     let responseItem = new StudentResponsesAssessmentItem();
     responseItem.assessmentItem = item;
 
-    let studentScore = item.scores[0].points;
+    let score: ExamItemScore = item.scores.length === 1 ?  item.scores[0] : null;
+    responseItem.score = score && item.scores[0].points >= 0 ? item.scores[0].points : null;
     let maxScore = item.maxPoints;
-    responseItem.correctness = studentScore / maxScore;
+    responseItem.correctness = responseItem.score !== null ? responseItem.score / maxScore : null;
+    responseItem.response = score ? score.response : null;
```
wtritch: Provide null score, correctness, and response values if the exam_item is missing altogether.
Provide null score, correctness values if the exam_item score is < 0

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -51,20 +51,26 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
         <span *ngFor="let standard of standards" class="label border-only blue mr-xs">{{standard}}</span>
       </ng-template>
     </p-column>
-    <p-column field="assessmentItem.scores.0.points" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true"></p-column>
+    <p-column field="score" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true">
+      <ng-template let-score="rowData.score" pTemplate="body">
+        <span *ngIf="score !== null">score</span>
+        <span *ngIf="score === null">&mdash;</span>
```
bgsavage: what happens when it's `undefined` ?

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -51,20 +51,26 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
         <span *ngFor="let standard of standards" class="label border-only blue mr-xs">{{standard}}</span>
       </ng-template>
     </p-column>
-    <p-column field="assessmentItem.scores.0.points" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true"></p-column>
+    <p-column field="score" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true">
+      <ng-template let-score="rowData.score" pTemplate="body">
+        <span *ngIf="score !== null">score</span>
```
bgsavage: Call me crazy, but shouldn't this be {{score}} ?

wtritch: uh... doh

jtreuting: crazy

bgsavage: I guess it's good job security to introduce new bugs while fixing bugs :D

wtritch:
@jtreuting I re-generated package-lock.json.  Since I'm able to run the app, everything should be fine.  But it will bump vaguely-defined dependency versions to my current exact versions during the build process.

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -51,20 +51,26 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
...
     <p-column field="assessmentItem.maxPoints" header="{{'labels.student.responses.cols.max-score' | translate}}" sortable="true"></p-column>
     <p-column field="correctness" sortable="true">
       <ng-template pTemplate="header">
               <span info-label title="{{'labels.student.responses.cols.correctness' | translate}}"
                     content="{{'labels.student.responses.cols.correctness-info' | translate}}"></span>
       </ng-template>
       <ng-template let-correctness="rowData.correctness" pTemplate="body">
-        {{ correctness | number:'1.2-2' }}
+        <span *ngIf="correctness !== null">{{ correctness | number:'1.2-2' }}</span>
+        <span *ngIf="correctness === null">&mdash;</span>
```
bgsavage: when we have more time, it'd be nice to create a `ValueOrDashPipe` instead.

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
wtritch: I don't believe:
`responseItem.score = score && item.scores[0].points >= 0 ? item.scores[0].points : null;`
leaves any room for it to be undefined

webapp/src/main/webapp/src/app/student/responses/student-responses.component.ts
bgsavage: I think `undefined` is preferred in js, no?  The default value is always `undefined` so introducing `null` values creates one "thing" to check when looking for nulls / undefined / etc.

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
wtritch: hmmm, would it accept a number format?  What if we wanted to use it on other values than numbers?

bgsavage: I think it could expect `:any` and if it's a null value (null, undefined, "") then it returns a "-"

PR #337, opened 2017-08-01
Feature/add report processor api

wtritch:
This PR includes the API that will be used by the Reporting web-app to communicate with the report processor microservice.  It contains endpoints for generating, listing, and retrieving contents of reports.


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
```diff
@@ -0,0 +1,166 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.opentestsystem.rdw.reporting.common.report.AbstractExamReportRequest;
+
+import java.net.URI;
+import java.time.Instant;
+
+/**
+ * The model represents the request, generation, and current status of a report.
+ */
+public class Report {
```
wtritch: I put the Report model in the common package because it'll be piped through the Reporting back-end.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportStatus.java
```diff
@@ -0,0 +1,13 @@
...
+ * An enum of available report statuses.
+ */
+public enum ReportStatus {
+    PENDING,
+    RUNNING,
+    COMPLETED,
+    FAILED,
+    EXPIRED,
+    NO_RESULTS
+}
```
wtritch: We probably don't need all these states.  We can prune them once we have a fully-functioning system.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
```diff
@@ -1,9 +1,11 @@
 package org.opentestsystem.rdw.reporting.common.report;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import org.opentestsystem.rdw.common.model.Subject;
 
 import java.util.Locale;
 
+@JsonTypeInfo(use=JsonTypeInfo.Id.CLASS, include=JsonTypeInfo.As.PROPERTY, property="@class")
```
wtritch: This tells Jackson to append class information when serializing instances of this abstract class.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/BatchExamReportRequestHolder.java
```diff
@@ -8,12 +8,24 @@
 
     private String id;
     private PermissionScope permissionScope;
+    private String user;
+    private String label;
```
wtritch: I added the user identifier and user-entered report label to the wrapping request holder.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/PrintOptions.java
```diff
@@ -5,7 +5,12 @@
  */
 public class PrintOptions {
 
-    private final boolean grayscale;
+    private boolean grayscale;
+
+    /**
+     * No-arg constructor for serialization
+     */
+    private PrintOptions() {}
```
wtritch: I had to add some private no-arg constructors for jackson serialization.

pdf-report-processor/build.gradle
```diff
@@ -2,28 +2,28 @@ apply plugin: 'org.springframework.boot'
...
-    compile 'com.google.guava:guava'
-    compile 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
-    compile 'mysql:mysql-connector-java'
 
+    testCompile 'org.assertj:assertj-core'
     testCompile('org.springframework.boot:spring-boot-starter-test') {
         exclude(module: 'commons-logging')
     }
-    testCompile 'org.springframework.security:spring-security-test'
     testCompile 'org.springframework.cloud:spring-cloud-stream-test-support'
```
wtritch: These changes consist of:
1. alpha-ordering
2. adding assertj-core test dependency
3. removing spring-security dependency

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.InputStream;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Default implementation of a ReportService
+ */
+@Service
+public class DefaultReportService implements ReportService {
```
wtritch: Temporary shim to support the API.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportService.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.io.InputStream;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Implementations of this interface are responsible for creating and retrieving Reports.
+ */
+public interface ReportService {
```
wtritch: This is the report service that will be responsible for generating/listing/retrieve contents for reports.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ExceptionHandlingAdvice.java
```diff
@@ -0,0 +1,30 @@
...
+     * This catches Optional.get() when it is empty.
+     *
+     * @param exception original exception
+     * @param request   request
+     * @return NOT_FOUND (404) response
+     */
+    @ExceptionHandler(value = NoSuchElementException.class)
+    protected ResponseEntity<Object> handleNoSuchElementException(final RuntimeException exception, final WebRequest request) {
+        return handleExceptionInternal(exception, "no such element", new HttpHeaders(), HttpStatus.NOT_FOUND, request);
+    }
```
wtritch: Just adding some 404 handling here.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -0,0 +1,99 @@
...
+        final Report report = reportService.findOneById(reportId)
+                .orElseThrow(NoSuchElementException::new);
+        try (final InputStream pdfBytes = reportService.getContent(reportId);
+             final OutputStream responseOutput = response.getOutputStream()) {
+            response.setStatus(200);
+            response.setHeader(Headers.CONTENT_TYPE, APPLICATION_PDF_VALUE);
+            response.setHeader(Headers.CONTENT_DISPOSITION, "attachment; filename=" + report.getLabel() + ".pdf");
+            IOUtils.copy(pdfBytes, responseOutput);
+            response.flushBuffer();
+        }
```
wtritch: I built this out to just stream from the service-returned InputStream to the response OutputStream rather than have to copy all the PDF bytes into memory.

pdf-report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/web/ReportControllerIT.java
```diff
@@ -0,0 +1,174 @@
...
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.header;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@WebMvcTest(ReportController.class)
+public class ReportControllerIT {
```
wtritch: Tests showing the controller does what I think it does.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/PrintOptions.java
esotrope: Alternatively we can add @JsonCreator + @JsonProperty("propNameToMapToArg")

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Retrieve the report content for the given report.
+     *
+     * @param reportId  A report id
+     * @param response  The HttpServletResponse
+     * @throws IOException  If there is a problem reading/writing the response
+     */
+    @RequestMapping(value = "{reportId}/content", method = GET)
+    public void downloadReport(@PathVariable final Long reportId,
```
esotrope: For this we should eventually use a MessageConverter like for the individual student

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/PrintOptions.java
wtritch: I could go either way.  No-arg constructor feels more natural to me, but that's just what I'm used to.  Would you prefer the other way?

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Retrieve the list of Reports for a user, optionally filtered by id.
+     *
+     * @param user      The user identifier
+     * @param reportIds The report ids, empty will retrieve the list of all reports for the user
+     * @return The list of Reports
+     */
+    @GetMapping(produces = APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
```
esotrope: for my education - what does this annotation actually do? Isn't the default behavior to always return 200 when no error is thrown? I'm always curious about why this is explicitly put in.

wtritch: Mainly for consistency =)  It weirds me out when some in a controller have it and some don't.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/PrintOptions.java
esotrope: either way works - i take it jackson can do private property setting?

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
esotrope: K. I'd prefer we only place it in if its not 200 when successful if we can

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
esotrope: Just read this - this is a good point. I'm wondering if we need to do this instead of the message converter for the other student pdf service to save time/memory

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/PrintOptions.java
wtritch: yessir

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
wtritch: I'm not sure... I'm tempted to back-burner until mid-to-late next week when we can run some tests on the two implementations.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportStatus.java
agorina: Just curious: if the report is generated from the UI, when would it be NO_RESULTS? Or is there any other way to generate it?

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -0,0 +1,98 @@
+package org.opentestsystem.rdw.reporting.processor.web;
+
+import com.amazonaws.services.s3.Headers;
```
malaffoon: hmm, there is probably a more generic header import helper to use?

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -0,0 +1,98 @@
...
+     * @param response  The HttpServletResponse
+     * @throws IOException  If there is a problem reading/writing the response
+     */
+    @RequestMapping(value = "{reportId}/content", method = GET)
+    public void downloadReport(@PathVariable final Long reportId,
+                               final HttpServletResponse response) throws IOException {
+
+        final Report report = reportService.findOneById(reportId)
+                .orElseThrow(NoSuchElementException::new);
+        try (final InputStream pdfBytes = reportService.getContent(reportId);
```
malaffoon: Shouldn't some of this work be in the service?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
```diff
@@ -0,0 +1,166 @@
...
+            report.status = status;
+            report.reportResourceUri = reportResourceUri;
+            report.label = label;
+            report.reportRequest = reportRequest;
+            report.created = created;
+
+            return report;
+        }
+
+        public Builder copy(final Report report) {
```
esotrope: I'm curious why the @JsonIgnore and copy() methods are here but maybe it will come together for me when later code is submitted.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
wtritch: You mean pass the OutputStream into the service rather than have the service return an InputStream? ...we can look into that, but I'm tempted to leave it as a stretch task for next week.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
wtritch: doh, changed to Spring's HttpHeaders.BLAH

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
wtritch: The \@jsonignore is there because we don't have any other resource objects (for example a ReportResource) and we're just returning a Report object.  We can change that later if you like.  The copy() is there because the model instance is immutable but will be updated throughout the report generation process.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportStatus.java
wtritch: It could be NO_RESULTS during a re-generation if the user's permissions have changed. It probably won't be used by next week...

agorina: thank you for the explanation.

PR #339, opened 2017-08-01
adds reports page

esotrope:
image


.gitignore
```diff
@@ -1,6 +1,6 @@
 build/
 !.mvn/wrapper/maven-wrapper.jar
-src/main/resources/static/*.bundle.*
+src/main/resources/static/
```
esotrope: added this to make adding things to git from command line nicer. Everything in this folder is autogenerated by the npm build anyway

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
```diff
@@ -1,42 +0,0 @@
...
-import java.io.InputStream;
-import java.util.Collection;
-import java.util.List;
-import java.util.Optional;
-
-/**
- * Default implementation of a ReportService
- */
-@Service
-public class DefaultReportService implements ReportService {
```
esotrope: renamed to StubReportService for now until it is not stubbed

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -90,29 +97,36 @@ export const routes: Routes = [
...
+          }
+        ]
+      },
+      {
+        path: 'reports',
+        pathMatch: 'full',
+        data: { breadcrumb: { translate: 'labels.reports.heading' }, permissions: [ 'GROUP_PII_READ' ]},
+        canActivate: [ AuthorizeCanActivate ],
+        resolve: { reports: ReportsResolve },
+        component: ReportsComponent
```
esotrope: here is the new entry - rest is formatting

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -35,7 +35,7 @@
  * This controller is responsible for providing access to Reports.
  */
 @Controller
-@RequestMapping("/reports")
+@RequestMapping("/api/reports")
```
wtritch: 👍

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
wtritch: Nice, that'll avoid conflicts once we start building it out.

webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -0,0 +1,19 @@
+<h3 class="blue mb-md">{{'labels.reports.heading' | translate}}</h3>
+<div class="well">
+  <p-dataTable [value]="reports"
+               emptyMessage="{{'labels.reports.no-reports' | translate}}"
+               sortField="created"
+               sortOrder="-1"
+               tableStyleClass="table table-striped table-hover">
+
+    <p-column field="created" header="{{'labels.reports.table-column-header.created' | translate}}" sortable="true">
```
wtritch: Think we should have a menu action column?  Right now, there won't be any option for failed reports, but that's where we could stick the download link for completed reports.

esotrope: Yup that's what i was thinking just didn't add it yet.
Actually, what if the downloadable ones were just links and the ones you can regenerate have a regenerate button appear next to them that goes away when they are complete?

wtritch: I like that idea, but the regenerate button should only appear next to failed/expired reports?  But that'll be a problem for late next week, or the week after most likely =P  Unless this all just magically falls into place.

jtreuting: I like the more explicit button than the menu.  Also looked at the screenshot.  I think the first column should just be Date, and maybe the status can have the styling like the "Standardized", "Partial" so they have a little more to them than just text, since they are sort of tag like statuses.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
bgsavage: We can't define a default prefix in spring?

PR #342, opened 2017-08-02
adds remaining report download button components to each page

esotrope:
image

screen shot 2017-08-02 at 4 38 06 pm


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -377,14 +377,20 @@
...
       "download": "Download",
       "assessment-type": "Assessment Type",
       "subject": "Subject",
       "school-year": "School Year",
       "messages" : {
         "404": "This student does not have any test results for your selection. Please update your selection and try again.",
-        "500": "There was an unexpected error while generating this report.  Please try again."
+        "500": "There was an unexpected error while generating this report.  Please try again.",
+        "submitted": "Creating report. See <a href=\"/reports\">status</a>.",
+        "submission-failed": "Error creating report. See <a href=\"/reports\">status</a>."
```
esotrope: Not sure if we do this in other places but i thought it might be nice to have a link to the report page in at least the failed request case.

webapp/src/main/webapp/src/app/shared/notification/notification.component.html
```diff
@@ -6,7 +6,7 @@
            [dismissOnTimeout]="notification.configuration.dismissOnTimeout"
            (onClose)="onDismissed(notification)">
       <h4 *ngIf="notification.titleKey" class="alert-title">{{notification.titleKey | translate}}</h4>
-      <span>{{notification.messageKey | translate: notification.messageObject}}</span>
+      <span [innerHtml]="notification.messageKey | translate: notification.messageObject"></span>
```
esotrope: this allowed me to add ```<a>``` to my alert message

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -12,15 +12,20 @@ export abstract class ReportDownloadComponent implements OnInit {
   @Input()
   public schoolYears: Array<number>;
 
+  @Input()
+  public schoolYearsOnly: boolean = false;
```
esotrope: quick way to tell the component to only show school years which is what we want for school grade/group report generation.
Later if we add sort order/ language back this naming wont make sense anymore.

webapp/src/main/webapp/src/app/report/group-report-download.component.ts
```diff
@@ -0,0 +1,48 @@
+import { Component, Input } from "@angular/core";
+import { ReportService } from "./report.service";
+import { saveAs } from "file-saver";
+import { Subscription } from "rxjs";
+import { ReportDownloadComponent } from "./report-download.component";
+import { Download } from "../shared/data/download.model";
+import {NotificationService} from "../shared/notification/notification.service";
+import {Notification} from "../shared/notification/notification.model";
```
jtreuting: nit: spacing on this.  i know IntelliJ does it this way when it adds the import

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -128,4 +128,19 @@ export class ReportService {
     };
   }
 
+  /**
+   * Maps a API report model to a local report model
+   *
+   * @param remote the API model
+   * @returns {Report} the local model
+   */
+  private toReport(remote: any): Report {
```
jtreuting: We don't have a strong type for this API model

webapp/src/main/webapp/src/app/report/school-grade-report-download.component.ts
```diff
@@ -0,0 +1,51 @@
+import { Component, Input } from "@angular/core";
+import { ReportService } from "./report.service";
+import { saveAs } from "file-saver";
+import { Subscription } from "rxjs";
+import { ReportDownloadComponent } from "./report-download.component";
+import { Download } from "../shared/data/download.model";
+import {NotificationService} from "../shared/notification/notification.service";
+import {Notification} from "../shared/notification/notification.model";
```
jtreuting: git: same spacing

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: I think it makes sense.  We will also need a more dedicate report button somewhere so that people can get there without having to submit a report.  Not sure where that would go but might be a UI/UX question.

esotrope: Yeah, I almost feel like we should also allow a user to create any kind of report once on the report page, would feel a little strange if they had to go back and forth to make and view reports.

webapp/src/main/webapp/src/app/report/school-grade-report-download.component.ts
esotrope: Cool, does that git command make everything have consistent spacing? : P

webapp/src/main/webapp/src/app/report/group-report-download.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  private subscription: Subscription;
+
+  constructor(private service: ReportService, private notificationService: NotificationService) {
+    super('labels.reports.button-label.group');
+  }
+
+  public submit(): void {
+
+    // Keep handle on subscription for disabling submit button
+    this.subscription = this.service.createGroupExamReport(this.groupId, this.options)
```
bgsavage: Is there a reason to keep track of this subscription?

webapp/src/main/webapp/src/app/report/school-grade-report-download.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  private subscription: Subscription;
+
+  constructor(private service: ReportService, private notificationService: NotificationService) {
+    super('labels.reports.button-label.school-grade');
+  }
+
+  public submit(): void {
+
+    // Keep handle on subscription for disabling submit button
+    this.subscription = this.service.createSchoolGradeExamReport(this.schoolId, this.gradeId, this.options)
```
bgsavage: same question, why hold onto this subscription?

bgsavage: Also since this code is repeated in components, any chance to move dup code to a service that the components call?

webapp/src/main/webapp/src/app/shared/notification/notification.component.html
bgsavage: hm, I would prefer to put that under the `html` translation parent key then, I don't want to confuse whoever does our translations to allow them to think they can add html to any translate value.

webapp/src/main/webapp/src/styles.less
```diff
@@ -86,7 +86,9 @@ footer {
 .white-space-normal {
   white-space: normal;
 }
-student-report-download {
+student-report-download,
+group-report-download,
+school-grade-report-download {
```
bgsavage: Call me crazy.... but don't you need css selectors here?  i.e. `.student-report-download` or `#student-report-download` ?

webapp/src/main/webapp/src/app/report/group-report-download.component.ts
esotrope: Nope don't think so, I'll take it out

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -4,18 +4,18 @@
...
     <form #form="ngForm" (ngSubmit)="submit()">
-      <div class="form-group">
+      <div class="form-group" [hidden]="schoolYearsOnly">
         <label for="assessment-type-select">{{'labels.reports.assessment-type' | translate}}</label>
         <select id="assessment-type-select" class="form-control input-sm" name="assessment-type" [(ngModel)]="options.assessmentType">
           <option *ngFor="let type of assessmentTypes" [ngValue]="type">{{'labels.assessmentTypes.' + type + '.short-name' | translate}}</option>
         </select>
       </div>
-      <div class="form-group mt-sm">
+      <div class="form-group mt-sm" [hidden]="schoolYearsOnly">
```
esotrope: k makes sense

webapp/src/main/webapp/src/styles.less
esotrope: You crazy : j there i did it.

The way angular renders the components out it actually puts a ```<your-component-selector ...>``` named element in the dom. the default behavior the browsers is to treat unknown elements like a ```<div>``` which is not what i wanted. I noticed when I gave the ```display:inline-block``` style to the first element inside that element there was unwanted margin to the right. If you can think of a cleaner way to do this let me know.

bgsavage: Ah, these are actual html elements.  Got it.

So instead of doing <your-component-selector > can you do <span your-componenet-selector></span>.  It might require a change in your component's selector to something like `selector: '[school-grade-report-download]`

esotrope: ok ill try it out

webapp/src/main/webapp/src/styles.less
```diff
@@ -82,26 +88,25 @@ footer {
...
   select {
     font-weight: normal;
   }
 }
 
-
-//
-// Responsive Overrides
-// ----------------------------------------
+.notifications {
```
esotrope: for whatever reason bootstrap class navbar-fixed-top was creating an invisible div width:100% accross the page that prevented you from clicking things below the notifications so I added the styles we wanted in this class

webapp/src/main/webapp/src/app/shared/notification/notification.model.ts
```diff
@@ -1,31 +1,40 @@
...
-      this.configuration = _.merge(this.configuration, _configuration);
-    }
+  /**
+   * Set this flag if the notification can be dismissed
+   */
+  public dismissible: boolean = true;
+
+  constructor(options: any) {
+    Object.assign(this, options || {});
   }
```
esotrope: changed how we create & submit notificications.

webapp/src/main/webapp/src/app/shared/notification/notification.service.ts
```diff
@@ -22,5 +21,25 @@ export class NotificationService {
...
+    this.showNotification(this.createNotification('warning', options));
+  }
+
+  public error(options: any): void {
+    this.showNotification(this.createNotification('danger', options));
+  }
+
+  private createNotification(type: string, options: any): Notification {
+    return new Notification(Object.assign(options, {type: type}));
+  }
```
esotrope: added convenience methods for alert types so this would be kind of like a logger or the Console.

webapp/src/main/webapp/src/app/shared/notification/notification.component.html
```diff
@@ -1,12 +1,13 @@
...
-           [dismissible]="notification.configuration.dismissible"
-           [dismissOnTimeout]="notification.configuration.dismissOnTimeout"
+           [type]="notification.type"
+           [dismissible]="notification.dismissible"
+           [dismissOnTimeout]="notification.dismissOnTimeout"
            (onClose)="onDismissed(notification)">
-      <h4 *ngIf="notification.titleKey" class="alert-title">{{notification.titleKey | translate}}</h4>
-      <span>{{notification.messageKey | translate: notification.messageObject}}</span>
+      <span *ngIf="!notification.html">{{notification.id | translate: notification.args}}</span>
+      <span *ngIf="notification.html" [innerHtml]="notification.id | translate: notification.args"></span>
```
esotrope: @bgsavage now you have to explicitly flag the notification with ```html: true``` to make it render the html

webapp/src/main/webapp/src/app/shared/notification/notification.model.ts
esotrope: Also, I got rid of the title related stuff.
If we want to add title now we can flag the message content as HTML and put ```<strong>TITLE</strong>``` in the message. Bootstrap convention also seemed to suggest using ```<strong>``` for titles http://getbootstrap.com/components/#alerts

webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -6,11 +6,24 @@ <h3 class="blue mb-md">{{'labels.reports.heading' | translate}}</h3>
...
-    <p-column field="created" header="{{'labels.reports.table-column-header.created' | translate}}" sortable="true">
+    <p-column field="created" header="{{'labels.reports.column.created.header' | translate}}" sortable="true">
       <ng-template let-report="rowData" pTemplate="body">{{ report.created | date }}</ng-template>
     </p-column>
-    <p-column field="label" header="{{'labels.reports.table-column-header.label' | translate}}" sortable="true"></p-column>
-    <p-column field="status" header="{{'labels.reports.table-column-header.status' | translate}}" sortable="true"></p-column>
+    <p-column field="label" header="{{'labels.reports.column.label.header' | translate}}" sortable="true">
+      <ng-template let-report="rowData" pTemplate="body">
+        <a *ngIf="report.status === 'COMPLETED'" href="javascript:void(0)" (click)="getReport(report)">{{report.label}}</a>
+        <span *ngIf="report.status !== 'COMPLETED'">{{report.label}}</span>
```
bgsavage: Not saying we have to do this now, but we may want to consider start using ngIf else syntax:

https://stackoverflow.com/questions/43006550/how-to-use-ngif-else-in-angular-4

webapp/src/main/webapp/src/app/shared/notification/notification.component.html
bgsavage: 👍

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -372,19 +372,41 @@
...
       "download": "Download",
       "assessment-type": "Assessment Type",
       "subject": "Subject",
       "school-year": "School Year",
       "messages" : {
         "404": "This student does not have any test results for your selection. Please update your selection and try again.",
-        "500": "There was an unexpected error while generating this report.  Please try again."
+        "500": "There was an unexpected error while generating this report.  Please try again.",
+        "submitted": "Creating report. See <a href=\"/reports\">status</a>.",
+        "submission-failed": "Error creating report. See <a href=\"/reports\">status</a>.",
```
bgsavage: I still would prefer to have html elements in a specific spot in the en.json so that it's clear to people editing this file and wanting to override translations where they can use html and where they cannot.

webapp/src/main/webapp/src/app/report/reports.component.html
esotrope: Hm looks funky. I almost like the double if syntax better : p

webapp/src/main/webapp/src/app/shared/notification/notification.model.ts
bgsavage: I liked how we were doing the _.merge before because then you don't have to provide all the other defaults if you only wanted one property to be different.

bgsavage: Please test the Interpretive Guide / User Guide clicks didn't change as I think I was expecting the merge to happen.

webapp/src/main/webapp/src/app/report/reports.component.html
bgsavage: haha, yeah just a thought.

webapp/src/main/webapp/src/app/shared/notification/notification.model.ts
esotrope: The behavior is the same. If you only put {type: ...} it only will override the default type property on the notification object.
*Did test the homepage notifications.

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: Maybe what we can do is suffix the id with ```.html``` or ```-html```?
That way the messages can still be related to the page/component/feature they belong to and wont be found in the file far away from it

bgsavage: I'd be good with tagging html in the name somehow like that.

wtritch:
We may want to hide the buttons by default for now, until the UX is ready for prime-time. (Or add the same "Coming Soon" behavior we had on the instructional resources.)  We'll likely continue pushing to UAT as we develop.

esotrope:
@jtreuting would you prefer hiding the group/school report download buttons or showing "coming soon..." ?

jtreuting:
That's a good idea.  Let's do the coming soon popover like Instructional Resources has.

It should say "Available in week 2 of UAT"

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -396,9 +396,14 @@
...
-        "submission-failed": "Error creating report. See <a href=\"/reports\">status</a>.",
-        "download-failed": "There was an unexpected error while getting this report. Please try again."
+        "submitted": {
+          "html": "Creating report. See <a href=\"/reports\">status</a>."
+        },
+        "submission-failed": {
+          "html": "Error creating report. See <a href=\"/reports\">status</a>."
+        },
+        "download-failed": "There was an unexpected error while getting this report. Please try again.",
+        "coming-soon": "Available in week 2 of UAT"
```
esotrope: Coming soon text

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -37,4 +38,11 @@ export abstract class ReportDownloadComponent implements OnInit {
    */
   public abstract submit(): void;
 
+  /**
+   * TODO: remove when feature is ready
+   */
+  public showComingSoon(): void {
+    this.notificationService.info({id: 'labels.reports.messages.coming-soon'});
```
esotrope: Group/School Grade Download button now invokes this notification and doesn't show any popup

webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -28,6 +28,9 @@
                     <i class="fa fa-user-circle"></i> <span *ngIf="user">{{'welcome' | translate:user }}</span> <span class="caret"></span>
                   </button>
                   <ul *dropdownMenu class="dropdown-menu" role="menu">
+                    <li *ngIf="!production">
+                      <a class="dropdown-item" href="/reports">{{'labels.reports.link' | translate}}</a>
+                    </li>
```
esotrope: ```/reports``` link is now exposed when running locally with ```npm start```

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -1,9 +1,9 @@
 <span *ngIf="options" class="report-download-container">
 
-  <button *ngIf="batch" class="btn btn-sm btn-default"
+  <button *ngIf="production" class="btn btn-sm btn-default"
           (click)="showComingSoon()"><i class="fa fa-cloud-download" aria-hidden="true"></i> {{buttonLabel | translate}}</button>
```
esotrope: coming soon now shown based on production mode instead of if batch is true or not - this will allow us to develop it when running ```npm start``` and still not see it in QA/UAT

PR #344, opened 2017-08-02
set client character set

malaffoon:


common/src/test/resources/application-test.yml
```diff
@@ -4,7 +4,7 @@ app:
 
   datamart:
     datasource:
-      url: jdbc:mysql://localhost:3306/reporting_test?useSSL=false&useLegacyDatetimeCode=false
+      url: jdbc:mysql://localhost:3306/reporting_test?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8
```
esotrope: without this did the connection not support reading columns with utf8 columns?

malaffoon: Actually, reading data was fine in all my manual tests. But i wanted to make this change, just in case. For example, the user_report stuff that is being added is writing data that could be UTF-8.

When we address the JIRA about read-only database access, we can discuss whether to remove this flag for those read-only connections.

PR #346, opened 2017-08-02
Feature/student group permissions

bgsavage:
This PR updates the reporting permissions to implement GROUP_PII_READ alongside INDIVIDUAL_PII_READ.  DWR-701 has the details.


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
```diff
@@ -18,7 +18,7 @@
  */
 public class QueryUtils {
 
-    private static final List<Long> UNMATCHABLE_IDS = ImmutableList.of(-1L);
+    public static final List<Long> UNMATCHABLE_IDS = ImmutableList.of(-1L);
```
bgsavage: I'm using this outside the class now.

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/PermissionScope.java
```diff
@@ -27,18 +30,29 @@
...
+    private PermissionScope(final boolean isStatewide){
+        this.isStatewide = isStatewide;
     }
 
     public PermissionScope(
             final Collection<Long> districtIds,
             final Collection<Long> institutionIds) {
         this.districtIds = districtIds != null ? ImmutableSet.copyOf(districtIds) : ImmutableSet.of();
         this.institutionIds = institutionIds != null ? ImmutableSet.copyOf(institutionIds) : ImmutableSet.of();
+        this.isStatewide = this.districtIds.isEmpty() && this.institutionIds.isEmpty();
```
bgsavage: We were making this assumption as a getter before, now I only make the assumption in the constructor to keep original functionality / implementation.

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/PermissionScope.java
```diff
@@ -27,18 +30,29 @@
      */
     private Set<Long> institutionIds = ImmutableSet.of();
 
+    /**
+     * true if the permission scope is statewide
+     */
+    private boolean isStatewide;
+
     private PermissionScope() {
+        this.isStatewide = true;
```
bgsavage: default was true before as well, we just didn't have an explicit property for it.

common/src/main/resources/common.sql.yml
```diff
@@ -50,7 +50,26 @@ sql:
...
+              where sgm.student_group_id IN (:group_ids)
+              and sgm.student_id = :student_id
+              and (sg.subject_id is NULL OR sg.subject_id = a.subject_id))
+            then 1
+            else 0
+            end
+          )=1
+          and (1=:gpr_statewide or s.district_id in (:gpr_district_ids) or e.school_id in (:gpr_school_ids))
+        )
+      )
```
bgsavage: @agorina looking for optimization tips here, if any... :)

common/src/main/resources/common.sql.yml
```diff
@@ -50,7 +50,26 @@ sql:
...
+              select sg.subject_id
+              from student_group_membership sgm
+                join student_group sg ON sgm.student_group_id = sg.id
+              where sgm.student_group_id IN (:group_ids)
+              and sgm.student_id = :student_id
+              and (sg.subject_id is NULL OR sg.subject_id = a.subject_id))
+            then 1
+            else 0
+            end
+          )=1
```
bgsavage: Basically just looking to see if we can find a matching assessment subject id in the one of the user's groups.  If it exists, then it is true, otherwise it is false.

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/PermissionScope.java
bgsavage: I can change this, I just didn't research if we were using the default constructor anywhere else besides the statewide one.

common/src/test/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepositoryIT.java
```diff
@@ -42,27 +44,66 @@
...
+    }
+
+    @Test
+    public void findOneShouldNotReturnNullForGroupPiiReadStatewide() throws Exception {
+        assertThat(repository.findOneByStudentIdAndSchoolYear(empty(), statewide(), ImmutableSet.of(-1L), -1, 2017, 1)).isNotNull();
+    }
+
+    @Test
+    public void findOneShouldReturnNullForGroupPiiReadStatewideStudentNotInGroup() throws Exception {
+        assertThat(repository.findOneByStudentIdAndSchoolYear(empty(), statewide(), ImmutableSet.of(-1L), -2, 2017, 1)).isNull();
```
esotrope: These tests will need to change I think. 
The UNMATCHABLE_IDS is a set containing -1
We will need to use a group ID that doesn't collide with this reserved number.
This is why the schools and districts you see are all -1,2,3 etc * 10 to avoid the collision but keep the simple numbering.

bgsavage: ahhh i was wondering about that, ok I can update this data.

common/src/main/resources/common.sql.yml
esotrope: Looking at this again I realize we really need comments in the sql sometimes

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/student/DefaultStudentExamReportPdfService.java
```diff
@@ -31,14 +30,18 @@
...
 
         final AbstractExamReport report = repository
-                .findOneByStudentIdAndSchoolYear(permissionScope, request.getStudentId(), request.getSchoolYear(), (int) request.getSubject().id());
+                .findOneByStudentIdAndSchoolYear(
+                        user.getPermissionScopeByPermissionId("INDIVIDUAL_PII_READ"),
+                        user.getPermissionScopeByPermissionId("GROUP_PII_READ"),
+                        user.getGroupsById().keySet(),
+                        request.getStudentId(),
+                        request.getSchoolYear(),
+                        (int) request.getSubject().id());
```
esotrope: Is the ID not an int?

bgsavage: I guess not?  I didn't check, we were already casting to int prior to my change.

bgsavage: It's `long`

common/src/main/resources/common.sql.yml
agorina: My first reaction is to break it into two SQLs, but I am looking into it.

agorina: It seems to be performing OK on the QA instance. So for now I would keep it and see if we run into performance issues when we test for it. 
You can simplify it a little bit:
```
   (
        (1 = :ipr_statewide OR s.district_id IN (:ipr_district_ids) OR e.school_id IN (:ipr_school_ids))
        OR (
          EXISTS(
              SELECT 1
              FROM student_group_membership sgm
                JOIN student_group sg ON sgm.student_group_id = sg.id
              WHERE sgm.student_group_id IN (:group_ids)
                    AND sgm.student_id = :student_id
                    AND (sg.subject_id IS NULL OR sg.subject_id = :subject_id)
          )
          AND (1 = :gpr_statewide OR s.district_id IN (:gpr_district_ids) OR e.school_id IN (:gpr_school_ids))
        )
      )
```

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
```diff
@@ -39,16 +45,50 @@ private QueryUtils() {
...
+                .forEach(entry -> {
+                    final Permission permission = permissionsById.get(entry.getKey());
+                    final PermissionScope scope = permission == null
+                            ? PermissionScope.EMPTY
+                            : permission.getScope();
+
+                    securityParameters.putAll(getSecurityParameters(scope, entry.getValue()));
+                });
+
+        return securityParameters.build();
```
bgsavage: Java experts might be able to tell me a better way to build this map, unless using the builder as I am is good.

esotrope: Interesting. I think we might want to add to the javadocs exactly what permissions this is for, because it will only apply to those which have a prefix? The use and name of PERMISSION_PREFIXES is kind of confusing too. Because (1) yes we are holding their prefixes but we are also interested in only the two permissions that are in there. Maybe change to PII_READ_PERMISSION_PREFIXES?

esotrope: Likewise we might want to change the names of these utility methods to reflect what they are getting.
Like getPiiReadPermissionScopeParameters(permission map) and getPermissionScopeParameters(scope). Maybe not the best names but just a thought.

bgsavage: I could call it STUDENT_READ_PERMISSION_PREFIXES , if it's not too much of a mouthful.  My thinking here is if there is a 3rd permission that comes in that we need to also look at, in java, all we'd have to do is update that const.

common/src/main/resources/common.sql.yml
esotrope: nit: inconsistent sql reserved word casing

common/src/test/java/org/opentestsystem/rdw/reporting/common/report/ica/JdbcIcaReportRepositoryIT.java
```diff
@@ -38,72 +40,72 @@
...
     @Test
     public void findOneShouldFindByGroupPIIReadStatewide() throws Exception {
-        assertThat(repository.findOneByStudentIdAndSchoolYear(empty(), statewide(), ImmutableSet.of(-1L), -2, 2017, 1)).isNull();
+        assertThat(repository.findOneByStudentIdAndSchoolYear(permissions(groupOf(statewide())), ImmutableSet.of(-1L), -2, 2017, 1)).isNull();
     }
 
     @Test
     public void findOneShouldFindByGroupPIIReadStatewideStudentNotInGroup() throws Exception {
-        assertThat(repository.findOneByStudentIdAndSchoolYear(empty(), statewide(), ImmutableSet.of(-1L), -2, 2017, 1)).isNull();
+        assertThat(repository.findOneByStudentIdAndSchoolYear(permissions(groupOf(statewide())), ImmutableSet.of(-1L), -2, 2017, 1)).isNull();
```
esotrope: Really need to make the user just hold a permission ID to permission scope map. too much indirection here

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
esotrope: method body can be changed to this i think too:
return PERMISSION_PREFIXES.stream()
                .map(entry -> {
                    final Permission permission = permissionsById.get(entry.getKey());
                    return permission == null
                            ? PermissionScope.EMPTY
                            : permission.getScope()
                })
                .collect(toImmutableMap());

bgsavage: Yeah I spent probably way too much time yesterday trying to get this to work with a map function and wasn't successful.  If you're able to get this to compile, I'm happy to take it.

esotrope: Commit what you have and i can play with it once it's in.

bgsavage:
@malaffoon @agorina @esotrope @jtreuting @wtritch  Please take one more look if you can, I've made many commits since yesterday since I wasn't allowed to merge to develop, but pretty much I just followed the same pattern for updating student id lookup, student history, and student responses.

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/PermissionScope.java
```diff
@@ -27,18 +30,29 @@
      */
     private Set<Long> institutionIds = ImmutableSet.of();
 
+    /**
+     * true if the permission scope is statewide
+     */
+    private boolean isStatewide;
```
esotrope: convention is to just call this "statewide"

PR #347, opened 2017-08-02
set metadata base url from settings

malaffoon:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlSettings.java
```diff
@@ -170,5 +177,19 @@ public boolean isIncludeServerPort() {
...
             this.includeServerPort = includeServerPort;
         }
+
+        /**
+         * @return base url built from scheme, host and optional port; null if proxy is not enabled
+         */
+        public String getBaseUrl() {
+            if (!enabled) return null;
+
+            final StringBuilder sb = new StringBuilder();
```
malaffoon: may be brute force, but i pulled it right out of core spring saml2 stuff

PR #348, opened 2017-08-03
DWR-706 Adjusted test data to work with the latest schema.

agorina:


common/src/test/resources/common-integration-test-data.sql
```diff
@@ -1,8 +1,8 @@
 insert into district (id, natural_id, name) values
   (-10, 'district1', 'district1');
 
-insert into school (id, district_id, natural_id, name, import_id, updated) VALUES
-  (-10, -10, 'school1', 'school1', -1, '2017-07-18 20:14:34.000000');
+insert into school (id, district_id, natural_id, name, import_id, updated, migrate_id) VALUES
```
agorina: migrate_id is a NOT NULL column.

PR #349, opened 2017-08-03
Revert "adds remaining report download button components to each page"

esotrope:
Reverts SmarterApp/RDW_Reporting#342


esotrope:
* reverted due to ```ng test``` failing on the [selectors]:
Error: Template parse errors:
        Can't bind to 'schoolId' since it isn't a known property of 'span'. ("
            <span school-grade-report-download class="ml-xs"
                *ngIf="currentGrade"
                [ERROR ->][schoolId]="currentSchool.id"
                [gradeId]="currentGrade.id"
                [schoolYears]="filterOption"): ng:///DynamicTestModule/SchoolResultsComponent.html@6:8
        Can't bind to 'gradeId' since it isn't a known property of 'span'. ("
                *ngIf="currentGrade"
                [schoolId]="currentSchool.id"
                [ERROR ->][gradeId]="currentGrade.id"
                [schoolYears]="filterOptions.schoolYears">
            </span>
        "): ng:///DynamicTestModule/SchoolResultsComponent.html@7:8

This appears to work when doing ```ng build``` or ```ng build --prod``` but not ```ng test```
I think it may have to do with having an abstract base component - will investigate

bgsavage:
You probably just need to include the component in the test that is failing.

esotrope:
@bgsavage yeah that's what it was. Luckily I remembered angular had some cryptic messages for missing module imports so i could fix it

PR #351, opened 2017-08-03
Feature/batch report order

esotrope:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-download.component.html
+++ b/webapp/src/main/webapp/src/app/report/report-download.component.html
@@ -11,25 +11,31 @@
   <template #popoverTemplate>
     <form #form="ngForm" (ngSubmit)="submit()">
       <div class="form-group" *ngIf="!batch">
-        <label for="assessment-type-select">{{'labels.reports.assessment-type' | translate}}</label>
+        <label for="assessment-type-select">{{'labels.reports.form.assessment-type' | translate}}</label>
         <select id="assessment-type-select" class="form-control input-sm" name="assessment-type" [(ngModel)]="options.assessmentType">
           <option *ngFor="let type of assessmentTypes" [ngValue]="type">{{'labels.assessmentTypes.' + type + '.short-name' | translate}}</option>
         </select>
       </div>
       <div class="form-group mt-sm" *ngIf="!batch">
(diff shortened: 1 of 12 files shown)


webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -11,25 +11,31 @@
   <template #popoverTemplate>
     <form #form="ngForm" (ngSubmit)="submit()">
       <div class="form-group" *ngIf="!batch">
-        <label for="assessment-type-select">{{'labels.reports.assessment-type' | translate}}</label>
+        <label for="assessment-type-select">{{'labels.reports.form.assessment-type' | translate}}</label>
```
jtreuting: Noticed while in the code, above on line 8 it is currently `[placement]="'bottom'"` but you shouldn't need to bind to an expression like that and can just have it parse the string like `placement="bottom"`

bgsavage:
General question unrelated to this pr about this control, but I've noticed you have to click the button again in order for the popover to close.  Was wondering if this is intuitive?  My initial attempt was trying to click outside the popover to close it, but that might be just me because of other systems I've worked in.

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -134,6 +135,21 @@ export class ReportService {
...
+   *
+   * @param options the options to convert to url parameters
+   * @returns {{schoolYear: number, language: string, grayscale: boolean, order: any}}
+   */
+  private toBatchReportRequestParameters(options: ReportOptions): Object {
+    return {
+      schoolYear: options.schoolYear,
+      language: options.language,
+      grayscale: options.grayscale,
+      order: ReportOrder[ options.order ]
```
bgsavage: interesting, are we sorting on the backend?

bgsavage: Oh wait, this is a dumb question, of course we have to sort on the backend because this is for pdf generation :)

Carry on.

webapp/src/main/webapp/src/app/report/report-download.component.html
esotrope: K thanks, I'll play with it.

PR #352, opened 2017-08-03
Bugfix/657 csv null scale score

jtreuting:
Fixed bug where NULL scale score and claim and levels were showing incorrect values in CSV

NULL scale score was showing 0
NULL levels was showing the translation string with undefined at the end

While testing the fix I found a problem with the display of the Reporting Category or Achievement Level value when NULL in the data table

### DIFF
--- a/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
+++ b/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
@@ -169,7 +169,7 @@ export class CsvBuilder {
       this.translateService.instant('labels.groups.results.assessment.exams.cols.ica.performance'),
       (item) => {
         let exam: Exam = getExam(item);
-        if (!exam) return "";
+        if (!exam || !exam.level) return "";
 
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -191,7 +191,10 @@ export class CsvBuilder {
   withScaleScore(getExam: (item: any) => Exam) {
     return this.withColumn(
       this.translateService.instant('labels.groups.results.assessment.exams.cols.score'),
-      (item) => getExam(item).score
+      (item) => {
+        let score = getExam(item).score;
+        return !score ? '' : score;
```
bgsavage: what happens when score is 0 and considered falsy here?

jtreuting: 0 is not a valid score so I wasn't worried about it, considering now if I should be or not.  hmmm

PR #358, opened 2017-08-04
removes spring cloud stream - now using rest interface to report micr…

esotrope:
…oservice


pdf-report-processor/build.gradle
```diff
@@ -18,13 +18,11 @@ dependencies {
     compile 'org.springframework.cloud:spring-cloud-starter-aws'
     compile 'org.springframework.cloud:spring-cloud-starter-config'
     compile 'org.springframework.cloud:spring-cloud-starter-stream-rabbit'
```
wtritch: Should we get rid of spring-cloud-starter-stream-rabbit as well?

webapp/build.gradle
```diff
@@ -19,7 +19,6 @@ dependencies {
     compile 'org.springframework.boot:spring-boot-starter-web'
     compile 'org.springframework.cloud:spring-cloud-starter-config'
     compile 'org.springframework.cloud:spring-cloud-starter-stream-rabbit'
```
wtritch: likewise here, remove the rabbit stream impl?

pdf-report-processor/build.gradle
esotrope: I'll test it without and remove it if we can get along without it

malaffoon:
Shouldn't this have affected the docker config?

PR #359, opened 2017-08-04
fixes double submit and adds code style fixes

esotrope:


webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -35,7 +35,7 @@
         </select>
       </div>
       <div class="form-group mt-sm">
-          <button class="btn btn-info btn-sm btn-block" [disabled]="subscription != null" (click)="form.ngSubmit.emit()">{{'labels.reports.form.submit' | translate}}</button>
```
esotrope: Don't need the click handler now that the ```<button>``` is in the ```<form>```. If you have it it fires the submit twice

bgsavage: Can you explicitly set `type="submit"` so this doesn't get lost if another button was added here?

esotrope: noted - will make another pr for it

PR #360, opened 2017-08-04
Bug/dwr 514 error handle group

bgsavage:
Now handling 400 / 403 / 404 errors by returning null and handling nulls correctly in reporting webapp pages.

image


webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -53,12 +53,24 @@ export class BreadcrumbsComponent implements OnInit{
...
       }
-
       if(crumbData.translateResolve){
-        translateParams = Utils.getPropertyValue(crumbData.translateResolve, child.snapshot.data);
+        try {
+          translateParams = Utils.getPropertyValue(crumbData.translateResolve, child.snapshot.data);
+        }
+        catch(ex) {
+          console.error("Unable to get property value for breadcrumb.", ex);
+          label = this.getDefaultLabel(commands);
```
bgsavage: So in the case of the screenshot, we aren't able to resolve student, so an exception is thrown when trying to get student.firstName.  Here, if we get any exception at all, try to fallback and just creating a label based on route information.  In the screenshot, you can see the bookmark is the attempted student id.

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
```diff
@@ -27,31 +27,31 @@ export class GroupAssessmentService implements AssessmentProvider {
...
+      .catch(ResponseUtils.badResponseToNull)
       .map(x => {
         return this.mapper.mapAssessmentsFromApi(x);
       });
   }
 
   getExams(assessmentId: number) {
     return this.dataService.get(`/groups/${this.groupId}/assessments/${assessmentId}/exams`, { search: this.getSchoolYearParams(this.schoolYear) })
-      .catch(ResponseUtils.notFoundToEmptyArray)
+      .catch(ResponseUtils.badResponseToNull)
```
bgsavage: method rename since I'm using it to handle 400, 403, and 404

webapp/src/main/webapp/src/app/groups/results/group-assessments.resolve.ts
```diff
@@ -6,12 +6,7 @@ import { GroupAssessmentService } from "./group-assessment.service";
 @Injectable()
 export class GroupAssessmentsResolve implements Resolve<any> {
   resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<any>|Promise<any>|any {
-    // TODO: Check params to return most recent or custom.
-    return this.service.getMostRecentAssessment(route.params[ "groupId" ], route.params[ "schoolYear" ])
-      .toPromise()
-      .then(value => {
-        return value;
-      })
+    return this.service.getMostRecentAssessment(route.params[ "groupId" ], route.params[ "schoolYear" ]);
```
bgsavage: no longer need promises since we should always return something with the new observable catch pattern we are using.

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
```diff
@@ -5,7 +5,8 @@
         {{'labels.school-grade.school-name.label' | translate}}
       </label>
       <div>
-        <p-dropdown [options]="availableSchools"
+        <p-dropdown id="rdw-school-grade-select-school"
```
bgsavage: Added ids for Nohemi

webapp/src/main/webapp/src/app/shared/response-utils.ts
```diff
@@ -4,30 +4,16 @@ import { Observable } from "rxjs";
...
-  }
-
-  /**
-   * Handle a 404 Response code as an empty array response.
-   * Any other response code is re-thrown.
-   *
-   * @param response  The response
-   * @returns {Observable} An empty array response for 404, otherwise an error
-   */
-  static notFoundToEmptyArray(response: Response): Observable<Response> {
```
bgsavage: It's hard to tell on the diff, but I actually removed `notFoundToEmptyArray` because the implementation was an exact copy of `notFoundToNull`.  I'm not sure if there was some original intention to have different implementations or?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentExamItemController.java
```diff
@@ -20,6 +20,7 @@ public StudentExamItemController(final ExamItemService examItemService) {
         this.examItemService = examItemService;
     }
 
+    //TODO: Investigate removing unused {studentId} here.
     @GetMapping("/api/students/{studentId}/exams/{examId}/examitems")
```
bgsavage: Wasn't sure about this so a stuck a TODO in there.  Maybe we just wanted {studentId} for some sort of consistency?

webapp/src/main/webapp/src/app/student/results/student-exam-history.resolve.ts
```diff
@@ -7,8 +7,16 @@ import { StudentExamHistoryService } from "../student-exam-history.service";
...
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<any>|Promise<any>|any {
+    return this.service
+      .findOneById(route.params[ "studentId" ])
+      .toPromise()
+      .then(value => {
+        return value;
+      })
+      .catch(ex => {
+        return null;
+      });
```
bgsavage: Oh you know what I meant to remove these.

webapp/src/main/webapp/src/app/school-grade/results/school-assessment.service.ts
```diff
@@ -29,31 +29,31 @@ export class SchoolAssessmentService implements AssessmentProvider {
 
   getAvailableAssessments() {
     return this.dataService.get(`/schools/${this.schoolId}/assessmentGrades/${this.gradeId}/assessments`, { search: this.getSchoolYearParams(this.schoolYear) })
-      .catch(ResponseUtils.notFoundToEmptyArray)
+      .catch(ResponseUtils.badResponseToNull)
```
esotrope: After these catch block, I would expect to see an if (... = null) in the map block - do all these mappers handle null case? I would think a mapper would only handle copying properties and not logic like that

webapp/src/main/webapp/src/app/student/responses/student-history-responses-assessment.resolve.ts
```diff
@@ -14,13 +14,13 @@ export class StudentHistoryResponsesAssessmentResolve implements Resolve<Assessm
     let examId: number = route.params[ "examId" ];
     let history: StudentExamHistory = route.parent.data[ "examHistory" ];
     if (!history) {
-      throw("Cannot resolve parent's StudentExamHistory");
+      return null;
```
bgsavage: If we throw here, we never actually resolve.  So now returning null.

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
esotrope: I'm curious, is it sometimes the case though that we want an empty array to represent a 404 response?

bgsavage:
I'm still adding unit tests before I merge.

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
bgsavage: It's easier to just trap for nulls than to trap for nulls and empty arrays.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -341,7 +343,8 @@
       "prompt": "Enter the Statewide Student Identifier (SSID)",
       "name": "{{lastName}}, {{firstName}}",
       "messages": {
-        "no-results": "Student does not exist or has no available test results."
+        "no-results": "Student does not exist or has no available test results.",
+        "null-student": "This student is unavailable.  Contact your administrator if you think you should have access to this student."
```
jtreuting: Change "Contact your administrator" to "Contact your building administrator" as that is the language they gave us when a user does not have groups to show.

And I think we can make the 2nd sentence more generic for each of the 3, maybe something like "Contact your building administrator if you feel your access is incorrect."

PR #361, opened 2017-08-04
Feature/report api

esotrope:
Mostly cleanup here.

1. Added ExamReportController + IT
2. Removes/moves content of report.schoolgrade/group into ExamReportService/Controller
3. Removes *Pdf from naming
4. Hooks up UI to ExamReportController endpoints


webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
```diff
@@ -0,0 +1,63 @@
...
+        return service.getReports(user);
+    }
+
+    @GetMapping("/api/reports/{reportId}")
+    public void getReport(
+            @AuthenticationPrincipal final User user,
+            @PathVariable final long reportId,
+            final HttpServletResponse response) throws IOException {
+
+        service.streamReport(user, reportId, response.getOutputStream());
```
esotrope: Streams PDF by passing output stream down through service/controller/client

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -16,33 +16,13 @@ export class ReportService {
...
-
-    let report3 = new Report();
-    report3.id = 3;
-    report3.label = 'Report 3';
-    report3.created = new Date(2017, 1, 3);
-    report3.status = 'COMPLETED';
-
-    return Observable.of([report1, report2, report3]);
+    return this.dataService.get('/reports')
+      .map(reports => (reports || []).map(this.toReport));
```
esotrope: @wtritch @bgsavage I still need to add the new error handling standard here I think right?

bgsavage: yes please, and make sure your component can init on null values (in case of 400/403/404)

esotrope: Ok I'll mock that out to see what happens and patch it up from there - thanks

pdf-report-processor/src/main/resources/application.yml
```diff
@@ -1,11 +1,12 @@
 # this app is expected to run inside the security perimeter so disable security for actuator end-points and,
 # since those are the only end-points, switch to a less standard port using the main server.port config
 management:
+  port: 8009
```
malaffoon: and what is wrong with 8008 like all the other services?

pdf-report-processor/src/main/resources/application.yml
```diff
@@ -1,11 +1,12 @@
...
   security:
     enabled: false
 
 server:
-  port: 8009
+  port: 8083
```
malaffoon: in the embedded application yml this should be 8080, it gets mapped to something else by the orchestration environment or run-time command line

wtritch: Good call, will change to 8080 in application.yml in a cleanup PR.

pdf-report-processor/src/main/resources/application.yml
wtritch: Gah, I had these changed for local development.  Will revert in a follow-up PR.  That'll teach me to use configuration properties rather than file modifications.

PR #364, opened 2017-08-05
Feature/report api error handling

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
```diff
@@ -46,18 +52,62 @@ public void getReport(
...
+     *
+     * TODO: find a better solution
+     */
+    private static class AnnotationIgnoringObjectMapper extends ObjectMapper {
+
+        private static final TypeReference<GroupExamReportRequest> GROUP_REQUEST_TYPE = new TypeReference<GroupExamReportRequest>() {};
+        private static final TypeReference<SchoolGradeExamReportRequest> SCHOOL_GRADE_REQUEST_TYPE = new TypeReference<SchoolGradeExamReportRequest>() {};
+
+        AnnotationIgnoringObjectMapper(){
+            configure(USE_ANNOTATIONS, false);
```
esotrope: Pretty hacky but isolated workaround - please read comments

webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -28,8 +28,8 @@
                     <i class="fa fa-user-circle"></i> <span *ngIf="user">{{'welcome' | translate:user }}</span> <span class="caret"></span>
                   </button>
                   <ul *dropdownMenu class="dropdown-menu" role="menu">
-                    <li *ngIf="!production">
-                      <a class="dropdown-item" href="/reports">{{'labels.reports.link' | translate}}</a>
+                    <li role="menuitem" *sbAuthorize="['GROUP_PII_READ']">
```
esotrope: Added "My Reports" link to user name dropdown. It is not not limited to development mode builds

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
wtritch: Hopefully this isn't needed any longer =)

esotrope: Just tested locally without it - It's not needed anymore - thanks : j

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
```diff
@@ -84,6 +84,11 @@ public Report createReport(final BatchExamReportRequestHolder request) {
         public Void extractData(final ClientHttpResponse response) throws IOException {
             // TODO copy Content-Disposition/Content-Type headers over too?
             // To copy the headers this will need the entire response object
+            if (!response.getStatusCode().is2xxSuccessful()) {
+                throw new RuntimeException(String.format(
```
bgsavage: Does this result in a 500 response ?

webapp/src/main/webapp/src/app/report/report.model.ts
```diff
@@ -10,4 +10,22 @@ export class Report {
...
 
+  public isCompleted(): boolean {
+    return this.status === 'COMPLETED';
+  }
+
+  public isProcessing(): boolean {
+    return this.status === 'PENDING' || this.status === 'RUNNING';
+  }
+
+  public getStatusColor(): string {
```
bgsavage: Can we run this through the `ColorService` wheel instead?

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -77,8 +79,17 @@ export class ReportService {
...
   private createBatchExamReport(url: string, options: ReportOptions): Observable<Report> {
-    return this.dataService.post(url, { params: this.toBatchReportRequestParameters(options) })
-      .map(this.toReport);
+    return this.dataService
+      .post(url, this.toBatchReportRequestParameters(options), {
+        headers: new Headers({ 'Content-Type': 'application/json' })
+      })
+      .catch(ResponseUtils.badResponseToNull)
+      .map(report => {
+        if (report == null) {
```
bgsavage: curious if you need this check to pass if it's null or undefined

webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -1,27 +1,32 @@
...
 <div class="well">
-  <p-dataTable [value]="reports"
-               emptyMessage="{{'labels.reports.no-reports' | translate}}"
-               sortField="created"
-               sortOrder="-1"
-               tableStyleClass="table table-striped table-hover">
+
+  <p *ngIf="!reports.length" [innerHtml]="'labels.reports.no-reports-html' | translate"></p>
+
+  <p-dataTable *ngIf="reports.length"
```
bgsavage: I wonder since you are using badResponseToNull - will this page render if `reports` is null?

webapp/src/main/webapp/src/app/home/home.component.ts
```diff
@@ -23,6 +23,7 @@ export class HomeComponent implements OnInit {
   }
 
   showComingSoon(featureName: string) {
-    this.notifService.info({id: 'messages.coming-soon', args: { featureName: featureName }, dismissOnTimeout: 5000000 });
+    this.notificationService.info({id: 'messages.coming-soon', args: { featureName: featureName }, dismissOnTimeout: 5000 });
```
bgsavage: Wow that was a crazy long timeout before :P

esotrope: Yeah, that was a testing config that slipped in. Made it that long to debug why the notifications were blocking clicking on the page

webapp/src/main/webapp/src/app/report/reports.component.html
esotrope: Null will never reach the view. The service maps null to an exception.

webapp/src/main/webapp/src/app/report/report.service.ts
esotrope: The service always returns an array, never null

webapp/src/main/webapp/src/app/report/report.model.ts
esotrope: I don't mind delegating to the service, but it shouldn't be a wheel of color probably. Certain colors are semantically associated with success/failure states

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
esotrope: It does. Not in place yet - but the goal is to map all 5xx to 404 on the API side

PR #367, opened 2017-08-05
DWR-652 - Persist report contents using the common ArchiveService

wtritch:


pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.io.InputStream;
+import java.net.URI;
+
+/**
+ * This implementation of a ReportContentService is backed by an ArchiveService.
+ */
+@Service
+public class ArchiveBackedReportContentService implements ReportContentService {
+    private final static String BasePath = "REPORTS/";
```
wtritch: This wraps the ArchiveService with Report knowledge.  I'm going to stuff the reports into a configurable bucket under a "/REPORTS/" directory.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
```diff
@@ -34,11 +33,6 @@ public Report insert(final Report report) {
     }
 
     @Override
-    public InputStream getContent(final long reportId) {
-        return null;
-    }
```
wtritch: I pulled this out into a separate ReportContentService.

wtritch:
Merging in the name of time, will continue to resolve comments as they come in.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
malaffoon: There are going to be a lot of `REPORTS`, perhaps we should be more specific; or add a subfolder. For example, we will have import reconciliation reports, aggregate reports, reliability reports. So, not sure how to decorate this one, maybe start with `REPORTS/USER/` or `USER_REPORTS` or ???

wtritch: I like that idea.  REPORTS/USER/ it is

PR #370, opened 2017-08-05
DWR-652 Update our report and student id repositories to allow search…

wtritch:
…ing for students / reports by student group.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
@@ -38,8 +38,11 @@ class JdbcIabReportRepository implements IabReportRepository {
     @Value("${sql.exam.findAllIabsByStudentIdAndSchoolYear}")
     private String findAllIabsByStudentIdAndSchoolYear;
 
-    @Value("${sql.exam.findAllIabsByStudentIdsAndFilter}")
-    private String findAllIabsByStudentIdsAndFilter;
+    @Value("${sql.exam.findAllIabsByStudentIdsAndQueryParams}")
+    private String findAllIabsByStudentIdsAndQueryParams;
+
+    @Value("${sql.exam.findAllIabsByStudentIdsAndQueryParamsWithGroup}")
+    private String findAllIabsByStudentIdsAndQueryParamsWithGroup;
 
     @Value("${app.state.code}")
     private String stateCode;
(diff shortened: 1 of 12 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
```diff
@@ -38,8 +38,11 @@
     @Value("${sql.exam.findAllIabsByStudentIdAndSchoolYear}")
     private String findAllIabsByStudentIdAndSchoolYear;
 
-    @Value("${sql.exam.findAllIabsByStudentIdsAndFilter}")
-    private String findAllIabsByStudentIdsAndFilter;
+    @Value("${sql.exam.findAllIabsByStudentIdsAndQueryParams}")
+    private String findAllIabsByStudentIdsAndQueryParams;
+
+    @Value("${sql.exam.findAllIabsByStudentIdsAndQueryParamsWithGroup}")
+    private String findAllIabsByStudentIdsAndQueryParamsWithGroup;
```
wtritch: Since the join tables will change based upon whether or not we're filtering by student_group, I need two queries... which leads to a bit of bloat in the .sql.yml file.

common/src/main/resources/common.sql.yml
```diff
@@ -169,7 +221,76 @@ sql:
       order by e.completed_at desc
       limit 1
 
-    findAllIcasByStudentIdsAndSchoolAndYear: >-
+    findAllIcasByStudentIdsAndQueryParams: >-
```
wtritch: This is an odd diff.  This query was renamed, and the query below was added.

common/src/main/resources/common.sql.yml
```diff
@@ -101,6 +101,58 @@ sql:
...
+        d.name as district_name
+      from exam e
+        join asmt a on e.asmt_id=a.id
+        left join instructional_resource ir on ir.name = a.name
+        join student st on e.student_id=st.id
+        join school s on e.school_id=s.id
+        join district d on s.district_id=d.id
+        join student_group_membership sgm on st.id = sgm.student_id
+        join student_group sg on sgm.student_group_id = sg.id
+          and (sg.subject_id is null or a.subject_id = sg.subject_id)
```
wtritch: If we're filtering by student_group, I need an inner-join on student to student_group_membership to student_group.

common/src/main/resources/common.sql.yml
```diff
@@ -101,6 +101,58 @@ sql:
...
+      where e.student_id in (:student_ids)
+        and sg.id=:group_id
+        and (:school_year is null or e.school_year=:school_year)
+        and (:school_id is null or s.id=:school_id)
+        and (:subject_id is null or a.subject_id=:subject_id)
```
wtritch: I still give the ExamQueryParams object the ability to further filter the subject should we ever want to support that in the UI.

wtritch:
Merging in the name of time, will continue to resolve comments as they arrive.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
malaffoon: And the student ids and group couldn't be part of the query params?

PR #374, opened 2017-08-07
adds header forwarding on report stream so we don't lose the file names

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
```diff
@@ -74,22 +71,24 @@ public Report createReport(final BatchExamReportRequestHolder request) {
...
+            this.response = response;
         }
 
         @Override
         public Void extractData(final ClientHttpResponse response) throws IOException {
-            // TODO copy Content-Disposition/Content-Type headers over too?
-            // To copy the headers this will need the entire response object
+            for (final Map.Entry<String, List<String>> entry : response.getHeaders().entrySet()) {
+                this.response.setHeader(entry.getKey(), HEADER_VALUE_JOINER.join(entry.getValue()));
+            }
```
esotrope: Quick way to forward the content-disposition header to the angular client. This way it knows what to name the report based on the backend data.

Discussion on better ways to get the report names to follow.

esotrope:
Merging for now. Possibly better implementation will follow

PR #375, opened 2017-08-07
Respect JAVA_OPTS when running container images

wtritch:
While testing deploying the report processing service to awsdev I found that our reporting containers aren't respecting the JAVA_OPTS environment variables which means we can't control java heap sizes.

This PR changes our container ENTRYPOINT into a CMD that passes the JAVA_OPTS environment variable to the java command.

We should probably look at our other projects to see if something similar is required.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -148,7 +148,7 @@ subprojects {
             label([maintainer: 'Fairway Technologies'])
             copyFile("${jar.archivePath.name}", "${jar.archivePath.name}")
             exposePort(8008, 8080)
-            entryPoint("java", "-jar", "${jar.archivePath.name}")
+            defaultCommand("sh", "-c", "java \$JAVA_OPTS -jar ${jar.archivePath.name}")
         }
 
         // task to build the docker image
--- a/build.gradle
+++ b/build.gradle
@@ -148,7 +148,14 @@ subprojects {
             label([maintainer: 'Fairway Technologies'])
             copyFile("${jar.archivePath.name}", "${jar.archivePath.name}")
             exposePort(8008, 8080)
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -148,7 +148,7 @@ subprojects {
             label([maintainer: 'Fairway Technologies'])
             copyFile("${jar.archivePath.name}", "${jar.archivePath.name}")
             exposePort(8008, 8080)
-            entryPoint("java", "-jar", "${jar.archivePath.name}")
+            defaultCommand("sh", "-c", "java \$JAVA_OPTS -jar ${jar.archivePath.name}")
```
malaffoon: Why the switch from entryPoint to cmd? Do we really want to suggest that it is okay to change what the image does?

wtritch: EntryPoint is executed outside of a shell context which means no access to environment variables.  If we want to use EntryPoint, we'll need to create a shell script file and point the entrypoint at that, which is also undesirable... thoughts?

build.gradle
```diff
@@ -148,7 +148,14 @@ subprojects {
...
             exposePort(8008, 8080)
-            entryPoint("java", "-jar", "${jar.archivePath.name}")
+            ext.java_opts = []
+            entryPoint {
+                List<String> entryPointArgs = ["java"]
+                entryPointArgs.addAll(ext.java_opts)
+                entryPointArgs.add("-jar")
+                entryPointArgs.add("${jar.archivePath.name}")
+                entryPointArgs.toArray()
+            }
```
wtritch: After working through this with @malaffoon we want to use the ENTRYPOINT so that SIGTERM and SIGKILL commands reach the java process.
Unfortunately, that precludes the use of environment variables for configuring java memory limits.  This PR adds a default max heap size 300m to the report-processor-service and allows configuring default memory limits for other report services in the future.

PR #376, opened 2017-08-07
Added error page

bgsavage:
Added an error page to be handled by angular.  This is the page any unexpected error from spring should navigate to, and/or if we are unable to resolve the user.

image


webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -23,7 +23,7 @@ export class RdwTranslateLoader {
 
   getTranslation(lang: string) {
     let uiObservable = this.uiLoader.getTranslation(lang);
-    let apiObservable = this.apiLoader.getTranslation(lang);
+    let apiObservable = this.apiLoader.getTranslation(lang).catch(res => Observable.of({}));
```
bgsavage: Added this catch in case so that if there is any error from the api translate loader, we can still always still fall back to the ui translate loader (en.json).

webapp/src/main/webapp/src/app/user/user.service.ts
```diff
@@ -22,8 +24,17 @@ export class UserService {
     if(isNullOrUndefined(this.currentUserObservable)) {
       this.currentUserObservable = this._dataService
         .get("/user")
+        .catch(res => {
+          return Observable.of(null);
```
bgsavage: We always need the `getCurrentUser` to resolve to something otherwise it can block child routes from loading (ie the /error route).

webapp/src/main/webapp/src/app/error/error.component.html
```diff
@@ -0,0 +1,6 @@
+<div class="alert alert-danger">
+  {{ 'messages.unknown-error' | translate }}
```
bgsavage: We can further fine tune this later to take some error status codes and provide a different message for each.

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -41,15 +42,19 @@ export class AppComponent {
...
-      this.interpretiveGuide = user.configuration.interpretiveGuide;
+        this.interpretiveGuide = user.configuration.interpretiveGuide;
 
-      if (window[ 'ga' ] && user.configuration && user.configuration.analyticsTrackingId) {
-        window[ 'ga' ]('create', user.configuration.analyticsTrackingId, 'auto');
+        if (window[ 'ga' ] && user.configuration && user.configuration.analyticsTrackingId) {
+          window[ 'ga' ]('create', user.configuration.analyticsTrackingId, 'auto');
+        }
+      } else {
+        this.router.navigate(["error"]);
```
bgsavage: If we are unable to resolve a user, then forward to the error page.  Without a user, we can do nothing.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -465,7 +466,8 @@
     },
     "search": "Search",
     "display-overall": "Overall Scores",
-    "display-claim": "Claim Scores"
+    "display-claim": "Claim Scores",
+    "go-home": "Go Home"
```
esotrope: I think you misspelled <i>Away</i>

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
esotrope: So for this page will it fallback on the translation codes? or default english messages if the API server explodes?

bgsavage: If we don't have a ja.son on the client, then yeah it will still fail to load.  If we do have a ja.json, it will fallback to that.

PR #377, opened 2017-08-08
adds client side default report naming

esotrope:
Basically the change here is to create batch report names client side in this format:
```
{groupName} {schoolYear}[ {language}]
{schoolName} {gradeName} {schoolYear}[ {language}]
```
When downloaded the spaces will be replaced with underscores.

Though inputing custom report names is not implemented yet, this strategy lends itself well to later presenting a default suggested report name that can be changed by the user before requesting the report.

<strong>Notice: there is currently no server-side sanitization of this text input. The angular client displaying the text is safe due to angular XSS sanitization built-ins but any other client of this text could be vulnerable to XSS if it doesn't take the proper precautions when displaying the data.</strong>

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
@@ -11,6 +11,7 @@ public abstract class AbstractExamReportRequest<T> {
     protected int schoolYear;
     protected Locale language;
     protected T options;
+    protected String name;
 
     public ExamReportAssessmentType getAssessmentType() {
         return assessmentType;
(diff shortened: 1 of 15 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
```diff
@@ -11,6 +11,7 @@
     protected int schoolYear;
     protected Locale language;
     protected T options;
+    protected String name;
```
esotrope: @wtritch, since the user may potentially be inputing a report name through the UI, I changed the report request model to now hold the report name. This makes the BatchRequestHolder label field redundant.

Also, this made me think a nicer user experience might be to have the app control the report name and just give the user the ability to add custom labels to a report that we display similar to status in the report table and also let the user filter on in the table. - thoughts?

PR #378, opened 2017-08-08
adds report status polling on reports page

esotrope:


webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -22,6 +22,55 @@ export class ReportsComponent implements OnInit {
 
   ngOnInit(): void {
     this.reports = this.route.snapshot.data[ 'reports' ];
+
+    /*
+     Start report status polling
+     Since reports currently cannot be generated on this page we do not have to dynamically update the IDs to pull
+     */
+
+    let pollingInterval: number = 5000;
```
esotrope: @wtritch @jtreuting I imagine us making this something closer to a minute or more between polls. What do you guys think a reasonable starting point for poll rate should be?

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
```diff
@@ -46,6 +47,14 @@
     }
 
     @Override
+    public List<Report> getReports(final String userId, final Iterable<Long> reportIds) {
+        final ResponseEntity<List<Report>> response = template.exchange(
+                "{url}/api/reports?user={userId}&id={reportIds}", HttpMethod.GET, null,
```
bgsavage: should `id` be `ids` ?

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -24,7 +24,28 @@ export class ReportService {
   public getReports(): Observable<Report[]> {
     return this.dataService.get('/reports')
       .catch(ResponseUtils.badResponseToNull)
-      .map(reports => (reports || []).map(this.toReport));
+      .map(reports => {
+        if (reports == null) {
+          throw new Error('Error getting reports');
```
bgsavage: Just want to make sure if you throw this error, does the User see some error message or does the page not load?

webapp/src/main/webapp/src/app/report/reports.component.ts
bgsavage: It'd be nice to do some kind of SignalR approach here (I'm not sure what the Java equivalent is) where the server posts a message back to the UI when it's done.  I'm curious if our polling should be as large as a minute, do we think it's likely users will wait on a page for a minute waiting for some status to change?  Or likely they would just come back and/or refresh which kind of diminishes the need for polling, yeah?

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -22,6 +22,55 @@ export class ReportsComponent implements OnInit {
...
+
+    setInterval(() => {
```
bgsavage: I think this function would be better in the service than the component in the spirit of keeping our components light for service reuse and quicker / easier unit tests.

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -22,6 +22,55 @@ export class ReportsComponent implements OnInit {
...
+              if (remote !== undefined && remote.status !== local.status) {
+                updated = true;
+                return remote;
+              }
+              return local;
+            });
+
+            // optimally updates the local report collection only when a change is detected
+            if (updated) {
+              this.reports = updatedReports;
```
bgsavage: Prefer to set reports where you set the flag instead of having to keep track of a flag.

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -22,6 +22,55 @@ export class ReportsComponent implements OnInit {
...
+
+        // optimally only send IDs of reports that are in progress
+        this.service.getReportsById(ids).subscribe(
+          remoteReports => {
+
+            // flag set when one or more reports are found to have a new status
+            let updated: boolean = false;
+
+            // creates a copy of the existing report collection and updates it with reports that have changed
+            let updatedReports: Report[] = this.reports.map(local => {
```
bgsavage: This really isn't a map as we're not mapping anything, just deciding if we should take the latest or not., maybe this should just be a .forEach?

webapp/src/main/webapp/src/app/report/reports.component.ts
esotrope: Good Qs - not sure what we expect. Also, we can kind of coach the user's expectations with notification messages when the submit the request and also with a blurb on the reports page

webapp/src/main/webapp/src/app/report/report.service.ts
esotrope: Thanks for noticing, right now it's not set up - will add.

webapp/src/main/webapp/src/app/report/reports.component.ts
esotrope: .forEach would modify the collection in-place while map makes a copy and doesn't modify the original. I think for the change detection to kick in i need to set the reports array reference to a new array instance no?

webapp/src/main/webapp/src/app/report/reports.component.ts
esotrope: When using prime ng does the table bind to the array elements or the array itself? or both? Will it detect report property changes automagically?

webapp/src/main/webapp/src/app/report/reports.component.ts
bgsavage: That's a good point, you do what to copy the array otherwise angular can miss the change and you'd have to implement your own specialized change detection.

webapp/src/main/webapp/src/app/report/reports.component.ts
bgsavage: I think just to the array, is there harm in always setting this.reports to updatedReports?

esotrope: Unnecessary re-render of a possibly long list

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
wtritch: Nope, although this might have to change, depending upon how smart the RestTemplate is.  The URI should look like: `/api/reports?user=bob&id=1&id=2&id=3`

webapp/src/main/webapp/src/app/report/reports.component.ts
wtritch: A little spinny with a message that alternates between "Updating" and "Updated" or something visual would help prevent users from habitually refreshing.

jtreuting: i think a websocket would be ideal, but for the first pass I think doing polling like every 20 or 30 seconds would be good.  And only doing it when we know there is something that is processing.

wtritch: @aaronsleeper ?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ExceptionHandlerControllerAdvice.java
```diff
@@ -14,58 +13,78 @@
...
+     * Catches all spring security access denied exceptions and masks them as HTTP 404 not found
+     *
+     * @param exception original exception
+     * @param request   request resulting in exception
+     * @return NOT_FOUND (404) response
+     */
+    @ExceptionHandler(AccessDeniedException.class)
+    public ResponseEntity<Object> handleAccessDeniedException(RuntimeException exception, WebRequest request) {
+        return notFound(exception, request);
     }
```
esotrope: Server side mapping of 403 -> 404

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ExceptionHandlerControllerAdvice.java
```diff
@@ -14,58 +13,78 @@
...
      * @param exception original exception
-     * @param request   request
-     * @return 4xx response
+     * @param request   request resulting in exception
+     * @return NOT_FOUND (404) response
+     */
+    @ExceptionHandler(RuntimeException.class)
+    protected ResponseEntity<Object> handleRemainingExceptions(final RuntimeException exception, final WebRequest request) {
+        return notFound(exception, request);
+    }
```
esotrope: Created catch-all for 5xx and other non-spring runtime exceptions - this handler maps these exceptions to 404s

webapp/src/main/webapp/src/app/shared/response-utils.ts
```diff
@@ -19,4 +19,6 @@ export class ResponseUtils {
     return Observable.throw(response);
   }
 
+  static toNull = (): Observable<any> => Observable.of(null);
```
esotrope: @wtritch, @jtreuting 
Talked extensively with @bgsavage about error handling and came up with a modified standard I'm hoping we can apply across the app.

API endpoint contract:

  /singleResource
    found -> {}
    not found -> 404

  /resourceCollection
    found -> [{},{},...]
    found but empty -> []
    not found -> []

API HTTP code masking:
  403 -> 404
  5xx -> 404
  (all other HTTP codes come back as-is and either are 401 unauthenticated or 4xx indicating developer error)

Client behavior:
```typescript
    // in service
    dataService.get(...)
      .map(report => /* do some mapping */ )

    // component consuming service
    service.getResource()
      .subscribe(
        data => { this.data = data; },
        error => {
          // as expected, use the error callback to handle errors encounted when invoking the service
          
        }
      );

    // in resolve
    service.getResource()
      .catch(() => Observable.of(null));

    // component consuming resolve
    // if data is null go to error state in view
    this.data = this.route.snapshot.data[ 'data' ];
```
This change is centered around noticing that we had our services coercing HTTP exception states to null which forced us to (1) have null checking in each service object mapper and (2) have null checks in our service subscription success handlers when they really belong in the error handler.
In this new approach we don't have to check for error cases in service subscription success handlers and mappers. We let all exceptions propagate to (1) null resolve which can be displayed as an opaque error or (2) a service exception which can be displayed as an opaque or http response-aware error

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ExceptionHandlerControllerAdvice.java
malaffoon: You would reduce the amount of code in here by either eliminating all RuntimeExceptions that are explicitly handled (e.g. NoSuchElementException) or combining them into one handler (@ExceptionHandler takes a list). Personally, i'd do the latter so you keep it explicit, even adding comments, e.g.
```
    @ExceptionHandler({
            NoSuchElementException.class,   // catch Optional.get() when it is empty
            RuntimeException.class          // catch everything else not explicitly called out
    })
    protected ResponseEntity<Object> handleRuntimeExceptions(final RuntimeException exception, final WebRequest request) {
        return notFound(exception, request);
    }
```

esotrope: Thanks - I'll change that.

PR #381, opened 2017-08-08
Implemented a route reuse strategy

bgsavage:
The strategy looks for a new data.CanReuse data property to determine when to save an instance of route and reattach later to keep state such as filters, assessments, and settings of a page when returning.


webapp/src/main/webapp/src/app/shared/rdw-route-reuse.strategy.ts
```diff
@@ -0,0 +1,77 @@
...
+   */
+  store(route: ActivatedRouteSnapshot, handle: DetachedRouteHandle|any): void {
+    // Only allow one handler per component type to minimize memory usage
+    for (let key in this.handlers) {
+      if (this.handlers[ key ]
+        && this.handlers[ key ].componentRef
+        && handle
+        && handle.componentRef
+        && this.handlers[ key ].componentRef.componentType == handle.componentRef.componentType) {
+        this.handlers[ key ] = null;
```
bgsavage: I'm very concerned about the memory impact this could have if we had multiple GroupComponents or SchoolComponents sitting in memory.  So for now, I'm limiting to only one per component type so state technically is only restored for those you can see in your breadcrumbs.  If you go back far enough, then state will not be restored.

webapp/src/main/webapp/src/app/shared/rdw-route-reuse.strategy.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Determines if this route should be reattached.  If the current route navigates to the HomeComponent, the handlers are cleared.
+   * @param route the current route
+   * @returns true if the route's data.canReuse evaluates as truthy and a handler exists.
+   */
+  shouldAttach(route: ActivatedRouteSnapshot): boolean {
+    console.debug("Route Reuse handlers: ", this.handlers);
+    if (route.component == HomeComponent) {
+      this.handlers = {};
```
bgsavage: Again, concerned about memory, so I'm clearing all the handlers when navigating home.

webapp/src/main/webapp/src/app/shared/rdw-route-reuse.strategy.ts
```diff
@@ -0,0 +1,77 @@
...
+   * @param future The route being navigated to
+   * @param curr The route being navigated from
+   * @returns true if the route should be reused.
+   */
+  shouldReuseRoute(future: ActivatedRouteSnapshot, curr: ActivatedRouteSnapshot): boolean {
+    return future.routeConfig === curr.routeConfig;
+  }
+
+  private getKey(route: ActivatedRouteSnapshot) {
+    return route.parent.url.join("/");
```
bgsavage: The current implementation only creates a key up to the first parent.  so for a schoolcomponent from route `groups/34/students/23` will be stored as `students/23`.  The current implementation shouldn't have any collisions because if you navigate home to then navigate to students a different way, your handlers will be cleared.  If we find that we don't want to clear on the home component, then we might need to consider keying all the way up to the root.

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-scoring.service.ts
```diff
@@ -13,7 +13,6 @@ export class ItemScoringService {
   getGuide(bankItemKey: string): Observable<ItemScoringGuide>{
     return this.dataService
       .get(`/examitems/${bankItemKey}/scoring`)
-      .catch(ResponseUtils.badResponseToNull)
```
bgsavage: unrelated: But I noticed we were getting the rubric / exemplar error in the tab when the item wasn't available anymore. It was caused by this catch block which we don't need as we are already catching in the caller.

webapp/src/main/webapp/src/app/shared/rdw-route-reuse.strategy.ts
jtreuting: not sure i fully understand but if a user can get to the same student test history page from group and grade and does that without going back to the home page, are you saying that could be an issue?  i'm assuming the only issue would be that the breadcrumb is wrong?

webapp/src/main/webapp/src/app/shared/rdw-route-reuse.strategy.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.handlers[ this.getKey(route) ] = handle;
+  }
+
+  /**
+   * Determines if this route should be reattached.  If the current route navigates to the HomeComponent, the handlers are cleared.
+   * @param route the current route
+   * @returns true if the route's data.canReuse evaluates as truthy and a handler exists.
+   */
+  shouldAttach(route: ActivatedRouteSnapshot): boolean {
+    console.debug("Route Reuse handlers: ", this.handlers);
```
jtreuting: nit: removing console logging

webapp/src/main/webapp/src/app/shared/rdw-route-reuse.strategy.ts
bgsavage: Possibly?  Though I'm not sure how they can do that without going back to the home page or changing the url themselves (changing the url manually causes a full app reload)

jtreuting: yeah i don't think they can unless they have saved links or something, just trying to understand if it happened what is the downside/risk to it.  but in practice a user will need to go to the homepage to change that

PR #383, opened 2017-08-09
Revert the revert of group permissions.

bgsavage:
Which is basically bringing by group permissions granular checks and resolves conflicts after merging from develop.

Original PR was reviewed here: https://github.com/SmarterApp/RDW_Reporting/pull/346


bgsavage:
@wtritch  and @esotrope  please review my conflict resolutions here.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
```diff
@@ -20,7 +21,12 @@
  */
 public class QueryUtils {
 
-    private static final List<Long> UNMATCHABLE_IDS = ImmutableList.of(-1L);
+    public static final List<Long> UNMATCHABLE_IDS = ImmutableList.of(-1L);
+    private static final Map<String, String> PERMISSION_PREFIXES = ImmutableMap
+            .<String, String>builder()
+            .put("GROUP_PII_READ", "group_")
+            .put("INDIVIDUAL_PII_READ", "individual_")
```
wtritch: Don'e we have these as constants somewhere?

common/src/main/resources/common.sql.yml
```diff
@@ -50,7 +50,21 @@ sql:
...
+              select 1
+              from student_group_membership sgm
+                join student_group sg on sgm.student_group_id = sg.id
+              where sgm.student_group_id in (:group_ids)
+                    and sgm.student_id = :student_id
+                    and (sg.subject_id is null or sg.subject_id = :subject_id)
+          )
+          and (1 = :group_statewide or s.district_id in (:group_district_ids) or e.school_id in (:group_school_ids))
+        )
+      )
```
wtritch: This is large, complex, and re-used enough SQL that I'd really, really, like to break it into it's own definition...  If this needs to be changed the updates across multiple projects are going to be painful and buggy.

bgsavage: Yep, I absolutely agree.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
bgsavage: I don't think so, all the uses to this point have just been of the string values.

common/src/main/resources/common.sql.yml
bgsavage: Given, there are slight variations depending on the query it's attached to.

wtritch: Do we need this chunk on every security-filtered query?  If not, why do we need it on some and not others?  Is there a way to document it in the code such that a new developer will understand why they need this one vs the simpler security filtering sql?

esotrope: We could put a comment blurb in the file about it.
There are pretty much 3 distinct security where clauses (individual_pii, group_pii, both)

bgsavage: We only need this for individual student look ups since we can have permission to look them either through INDIVIDUAL_PII_READ or through GROUP_PII_READ.  Other queries already have the context of how the look up is being applied and can pick which permission to use based on that.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
wtritch: Not a problem for the moment, but constants would help prevent future typos =)

bgsavage: I agree with that as well :)

common/src/main/resources/common.sql.yml
wtritch: A small comment blurb maybe in the sql yml file about this looking different from the others BECAUSE it may be executed from either a group_pii or individual_pii context might be nice?

bgsavage: And honestly I'm not sure about these other queries are for in this file.  When I added these, there were only two queries I needed to add these too.

PR #384, opened 2017-08-10
Feature/reports ui improvements

esotrope:
1. API 5xx no longer maps to 404
2. Revised error catching/handling on the UI
3. Changed look and feel of reports page to that seen below:

screen shot 2017-08-10 at 3 01 02 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report.model.ts
+++ b/webapp/src/main/webapp/src/app/report/report.model.ts
@@ -18,6 +18,10 @@ export class Report {
     return this.status === 'PENDING' || this.status === 'RUNNING';
   }
 
+  public canRegenerate(): boolean {
+    return !(this.isCompleted() || this.isProcessing());
+  }
+
   public getStatusColor(): string {
     if (this.isCompleted()) {
       return 'blue-dark';
--- a/webapp/src/main/webapp/src/app/report/reports.resolve.ts
+++ b/webapp/src/main/webapp/src/app/report/reports.resolve.ts
@@ -3,17 +3,18 @@ import { RouterStateSnapshot, ActivatedRouteSnapshot, Resolve } from "@angular/r
 import { Injectable } from "@angular/core";
 import { ReportService } from "./report.service";
 import { Report } from "./report.model";
-import { ResponseUtils } from "../shared/response-utils";
+import { Resolution } from "../shared/resolution.model";
 
 /**
  * This resolver is responsible for fetching assessment items for a student and exam.
  */
 @Injectable()
-export class ReportsResolve implements Resolve<Report[]> {
+export class ReportsResolve implements Resolve<Resolution<Report[]>> {
 
-  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Report[]> {
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Resolution<Report[]>> {
     return this.service.getReports()
-      .catch(ResponseUtils.toNull);
+      .map(Resolution.ok)
+      .catch(Resolution.errorObservable);
   }
 
   constructor(private service: ReportService) {
--- a/webapp/src/main/webapp/src/app/report/student-report-download.component.ts
+++ b/webapp/src/main/webapp/src/app/report/student-report-download.component.ts
@@ -34,7 +34,11 @@ export class StudentReportDownloadComponent extends ReportDownloadComponent {
           saveAs(download.content, download.name);
         },
         (error: any) => {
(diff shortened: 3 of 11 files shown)


webapp/src/main/webapp/src/app/report/report.model.ts
```diff
@@ -18,6 +18,10 @@ export class Report {
     return this.status === 'PENDING' || this.status === 'RUNNING';
   }
 
+  public canRegenerate(): boolean {
+    return !(this.isCompleted() || this.isProcessing());
```
bgsavage: Might just be my personal preference, but this would be easier to understand as `!this.IsCompleted() && !this.isProcessing()`

webapp/src/main/webapp/src/app/shared/resolution.model.ts
```diff
@@ -0,0 +1,40 @@
+import { NotFoundError } from "./not-found.error";
+import { Observable } from "rxjs";
+
+export class Resolution<T> {
```
bgsavage: I like it, but there's nothing in the framework already for this?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -426,7 +428,9 @@
           "html": "Error creating report. See <a href=\"/reports\">status</a>."
         },
         "download-failed": "There was an unexpected error while getting this report. Please try again.",
-        "coming-soon": "Available in week 2 of UAT"
+        "coming-soon": {
+          "report-regeneration": "Report regeneration coming soon."
```
bgsavage: nit: `*is coming soon`

webapp/src/main/webapp/src/app/report/report.model.ts
esotrope: Preferences are interesting.
Can reports be regenerated if they are not complete or processing OR if they are not complete and not processing? : j

webapp/src/main/webapp/src/app/shared/resolution.model.ts
esotrope: Not that I could find.
1. Typed Errors in TS are not really a thing i guess - they don't work as expected hence the error type checking via error.name == 'CustomErrorName'
2. Seemed like the community generally throws up a generic error view via route changing when resolves fail as opposed to letting components handle them individually.

webapp/src/main/webapp/src/app/report/student-report-download.component.ts
```diff
@@ -34,7 +34,11 @@ export class StudentReportDownloadComponent extends ReportDownloadComponent {
           saveAs(download.content, download.name);
         },
         (error: any) => {
-          this.notificationService.error({ id: 'labels.reports.messages.500' });
+          this.notificationService.error({
+            id: error.name === 'NotFoundError'
```
esotrope: One gotcha with TS is that there is no easy way to use typeof with custom error classes.
Seems like there is a workaround for this every TS release that breaks in the following release hence the reliance on the seemingly stable error.name checking.

webapp/src/main/webapp/src/app/report/reports.resolve.ts
```diff
@@ -3,17 +3,18 @@ import { RouterStateSnapshot, ActivatedRouteSnapshot, Resolve } from "@angular/r
...
 @Injectable()
-export class ReportsResolve implements Resolve<Report[]> {
+export class ReportsResolve implements Resolve<Resolution<Report[]>> {
 
-  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Report[]> {
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Resolution<Report[]>> {
     return this.service.getReports()
-      .catch(ResponseUtils.toNull);
+      .map(Resolution.ok)
+      .catch(Resolution.errorObservable);
```
esotrope: Another oddity I found is that catch blocks need to return an iterable of some sort but the map block doesn't need to? Can anyone explain this one to me?

webapp/src/main/webapp/src/app/report/student-report-download.component.ts
esotrope: Another thing to note about this change is that now - service consumers need not know about HTTP nuances, only that the data is not found or there was an error.
Also, if a resolve returns a Resolution, views can now distinguish if there was a 404 or 5xx during the resolution of their data before the page loaded.

webapp/src/main/webapp/src/app/report/reports.resolve.ts
bgsavage: Maps always return a single object, so the catch needs to return the same type as success, or throw.

PR #385, opened 2017-08-11
changes for ArchiveService

malaffoon:


pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/PdfExamReportProcessorApplication.java
```diff
@@ -23,7 +22,6 @@
 @EnableCaching
 @PropertySource(value = "classpath:build-info.properties", ignoreResourceNotFound = true)
 @Import({
-        ArchiveConfiguration.class,
```
malaffoon: moved the archive configuration into its own Configuration class

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportGenerationTemporaryStorage.java
```diff
@@ -22,16 +15,13 @@
...
 
     private final LocalArchiveService archiveService;
-    private final String root;
     private final TreeSet<Integer> chunks = new TreeSet<>();
 
     public ReportGenerationTemporaryStorage(final long executionId,
                                             final String rootUri) {
-        root = rootUri + "execution_" + executionId + "/";
-        this.archiveService = (LocalArchiveService) ArchiveServiceFactory.newInstance(root);
+        this.archiveService = new LocalArchiveService(rootUri + "execution_" + executionId + "/");
```
malaffoon: The factory no longer has that newInstance method so, for now, just instantiate a local archive service. We may switch this to point to a shared S3 bucket/folder, in which case we'll need to have to call the archive service factory for our beans.

pdf-report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportGenerationTemporaryStorage.java
```diff
@@ -64,11 +54,6 @@ public void writeChunk(final int chunkIdx, final InputStream content) {
      * Remove any persistent storage.
      */
     public void cleanup() {
-        final File file = new File(URI.create(root));
-        try {
-            FileUtils.deleteDirectory(file);
-        } catch (final IOException e) {
-            logger.warn("Unable to purge report generation directory", e);
-        }
+        archiveService.delete("");
```
malaffoon: New method means we don't have to dip into the file system directly, let archive service do it.

PR #386, opened 2017-08-11
adds dropdown form initialization based on filters

esotrope:
This addresses https://jira.fairwaytech.com/browse/DWR-730
It makes it so that when you open the student report download dropdowns that they start with the current filters selected on the page but do not change when they change and vice versa.

### DIFF
--- a/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
+++ b/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
@@ -14,6 +14,7 @@ import { AssessmentsComponent } from "../../assessments/assessments.component";
 import { TranslateService } from "@ngx-translate/core";
 import { CsvExportService } from "../../csv-export/csv-export.service";
 import { ItemByPointsEarnedExportRequest } from "../../assessments/model/item-by-points-earned-export-request.model";
+import { SchoolGradeDownloadComponent } from "../../report/school-grade-report-download.component";
 
 @Component({
   selector: 'app-group-results',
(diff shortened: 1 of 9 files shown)


webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -226,4 +227,14 @@ export class SchoolResultsComponent implements OnInit {
...
     });
   }
+
+  /**
+   * Initializes SchoolGradeDownloadComponent options with the currently selected filters
+   *
+   * @param downloader
+   */
+  private initializeDownloader(downloader: SchoolGradeDownloadComponent): void {
+    downloader.options.schoolYear = this.currentSchoolYear;
```
bgsavage: Can you clarify the approach here?  You're emitting an event just so you can set schoolYear here?   You can't pass this directly to the component or share an underlying model instead?

esotrope: Yeah so. The idea here is every time you open the popover the default values reflect the default values of the primary filters on the page.
Reason for this approach is
1. The user gets the filter state they currently have on the page as the default
2. Once they open it if they change the values in the popover those values change independently of the page filters
3. If the user changes the page filters the popover does not change (not as relevant if we have click-outside to close coming though)

If it shares the underlying model - when you change the schoolYear in the popover it will change the whole page's schoolYear right?

I think i could get away with [schoolYear]="pageSchoolYear"  on the school/group pages but for the individual student page that does not refresh data when you change subject/year this approach works best

PR #387, opened 2017-08-11
Bugfix/620 popover

jtreuting:
Changed to use ngx-popover:

- item info icons updated
- hover for instructional resource button and popup menu
- hover for item by points earned

Still needs UI adjustment as the popver placement is off


webapp/src/main/webapp/package.json
```diff
@@ -39,6 +39,7 @@
     "lodash": "^4.17.4",
     "mathjs": "^3.15.0",
     "ngx-bootstrap": "^1.6.6",
+    "ngx-popover": "0.0.16",
```
malaffoon: why does this one not have a ^? I think that locks the specific release version which we want, right?

bgsavage: Absent of caret and tilda will lock in the specific version, which is what we should be doing at this point.  The caret means it can update to the latest minor / patch version (i.e for ngx-bootstrap it will take 1.7 if it's available).

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.html
```diff
@@ -10,8 +10,12 @@
   </button>
   <ul *ngIf="open" class="dropdown-menu" role="menu">
     <li *ngFor="let action of actions; let idx = index" role="menuitem">
-
-      <a class="dropdown-item" popover="{{action.tooltip(item)}}" triggers="{{ action.tooltip(item) == '' ? '': 'mouseenter:mouseleave'}}" placement="right" container="body">
+      <a class="dropdown-item"
+         popover="{{action.tooltip(item)}}"
+         [popoverOnHover]="true"
```
bgsavage: Does this one need the `[ ]` ?

webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
```diff
@@ -1,9 +1,14 @@
...
-      triggers="focus">
+      [popover]="infoPopover">
   <span class="sr-only">{{'labels.more-info' | translate}}</span>
   <i class="fa fa-info-circle"></i>
 </button>
+
+<popover-content #infoPopover
+                 title="{{ title }}"
+                 placement="top"
+                 [animation]="true"
```
bgsavage: This one has an animation, should the others?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -263,7 +263,7 @@
...
+              "claim-info": "Claims and targets are a way of classifying test content.  The claim is the major topic area.  For example, in English language arts, reading is a claim.  Within each claim, there are targets that describe the particular knowledge and skills that the test measures.  Each target may encompass one or more standards from the Common Core State Standards.  Within the Reading claim, for example, one of the targets is concerned with finding the central idea in a text.  For more information, see the <a href=\"http://www.smarterbalanced.org/assessments/development/\" target=\"_blank\">Smarter Balanced Content Specifications</a>.",
```
bgsavage: @esotrope and I discussed a way to indicate that "This translate value can contain html" , I think it was to append `-html`?

esotrope: correct.

jtreuting: the generic info icon component shows these, so right now that is set to display HTML.  are you saying you think we should add -html to all of the info related text items in en.json?  or add it to this one and then make the info display component smart enough to only use `innerHtml` when the key ends in `-html`?

esotrope: The smart part would be cool. But no, the idea was just to suffix the message codes so it's obvious later which messages support HTML.

bgsavage: I think we need some way to document which of these translate values can contain html, and which cannot.  My original intent was to put all html-able translate values underneath a parent `html` tag, but the concern about that is that it moves the html translate values away from other translate values in the same feature / component.  

If we expect a lot of the info popovers to contain links, maybe they all get moved to the `html` section?  If this is a one or two-off, then leave it where it is, indicate it as `-html`, and then pass a flag to the info component to `renderAsHtml=true` ?  

At least those are my thoughts... :)

webapp/src/main/webapp/package.json
jtreuting: That was the default it gave me for `npm install ngx-popover` and it seemed appropriate

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.html
jtreuting: yes.  otherwise the value gets passed in as a string and is not considered an expression.  So it would be passed in as "true" and not `true`

webapp/src/main/webapp/src/app/assessments/results/information-label.component.html
jtreuting: Oh i think that was a copy paste from an example.  I don't remember it actually doing anything special.  I'll remove it and make sure it looks appropriate.  Good catch.

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.html
esotrope: Hover is something that only works with desktop/mouse.
Will this automagically be driven by clicks on mobile by chance?

jtreuting: This was replacing the mouseenter:mouseleave from the previous triggers so it is what matches that.  I can see if having both hover and clickable would work, but not sure if it will since the link gets disabled when this message should be displayed.  but worth a shot.

PR #388, opened 2017-08-13
Bugfix/731 print report menu option

jtreuting:
Added single PDF report generation option to the menu for an exam row in a table.  Now a single PDF report can be generated without needing to go to the student history results page.

- The menu is contextual so it knows the year, subject and assessment type
- Displays different menu option verbiage depending on if IAB, ICA or summative


webapp/src/main/webapp/src/app/assessments/menu/menu-action.builder.ts
```diff
@@ -3,8 +3,8 @@ import { PopupMenuAction } from "./popup-menu-action.model";
 import { TranslateService } from "@ngx-translate/core";
 import { ActivatedRoute, Router } from "@angular/router";
 import { Student } from "../../student/model/student.model";
-import { isNull } from "util";
```
jtreuting: not related, just saw it wasn't used anymore

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
```diff
@@ -12,6 +13,10 @@ export class Assessment {
   claimCodes: string[];
   cutPoints: number[];
 
+  get assessmentSubjectType(): AssessmentSubjectType {
+    return AssessmentSubjectType[this.subject];
+  }
```
jtreuting: Needed the subject as an AssessmentSubjectType for the report service.  I didn't want to change it everywhere else and went back and forth on whether to just do the conversion from string to enum in the submit report logic or include it here.  To me it felt like this enum would be nice to use in other places so put it here.  but am open to pulling it out.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -339,6 +346,39 @@ export class AssessmentResultsComponent implements OnInit {
...
+                this.notificationService.error({
+                  id: error.name === 'NotFoundError'
+                    ? 'labels.reports.messages.404'
+                    : 'labels.reports.messages.500'
+                });
+              }
+            )
+          ;
+        }
+      )
```
jtreuting: This is the submit report callback method used when a user clicks the menu option

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -1,6 +1,5 @@
 import { OnInit, Input, ViewChild, Output, EventEmitter } from "@angular/core";
 import { ReportOptions } from "./report-options.model";
-import { saveAs } from "file-saver";
```
jtreuting: unrelated unused import

webapp/src/main/webapp/src/app/assessments/menu/menu-action.builder.ts
```diff
@@ -120,6 +120,36 @@ export class MenuActionBuilder {
...
+        assessmentType = 'iab';
+      }
+      else if (assessmentExam.assessment.isIca) {
+        assessmentType = 'ica';
+      }
+      else if (assessmentExam.assessment.isSummative) {
+        assessmentType = 'summative';
+      }
+      else {
+        return '';
```
bgsavage: Is this a valid case where we don't have a valid assessment type?  If yes, shouldn't we be returning `this` or throwing some kind of exception.

bgsavage: Oh nevermind, this is the displayName function, ignore.

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
bgsavage: Hm, we should attempt to figure out some sort of consistency here.  This will cause problems later when someone asks "Well should I use `subject` or `assessmentSubjectType`" ?

esotrope: @bgsavage generally is it the direction on the frontend to use enum or string? Sounds like we just need to pick one and stick with it moving forward.
I'll vote enum.

webapp/src/main/webapp/src/app/assessments/menu/menu-action.builder.ts
esotrope: Simplification suggestion:
```typescript
return this.translateService.instant(`labels.menus.student-report.${getAssessmentExam(actionable).assessmentType.toLowerCase()}`, getStudent(actionable));
```

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -339,6 +346,39 @@ export class AssessmentResultsComponent implements OnInit {
...
+      .withStudentHistory(exam => exam.student)
+      .withStudentReport(
+        () => this._assessmentExam,
+        exam => exam.student,
+        exam => {
+          let reportOptions: ReportOptions = new ReportOptions();
+          reportOptions.assessmentType = this._assessmentExam.assessment.type;
+          reportOptions.subject = this._assessmentExam.assessment.assessmentSubjectType;
+          reportOptions.schoolYear = exam.schoolYear;
+          reportOptions.language = 'eng';
```
esotrope: I wonder how this would work later with being able to select multiple languages.

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
bgsavage: I'd lean towards using string enums made available in ts 2.4

esotrope: Looks like dev is pointing at
"typescript": "2.4.1"
and package-lock is on
"typescript": "2.2.2"
Should we upgrade this?

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
jtreuting: I'm guessing we would need a modal at that point to do that selection.

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
jtreuting: Wasn't there an issue with upgrading to 2.4 that you ran into before @bgsavage

webapp/src/main/webapp/src/app/assessments/menu/menu-action.builder.ts
jtreuting: since the assessment has an AssessmentType property which is an enum you can't do the toLowerCase like that and it seems like to make that a string would add some complexity to that simplified statement.   Also, if there isn't a valid assessment type (very unlikely) it would return the placeholder string.

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
bgsavage: Yeah it caused a breaking change in rxjs, but they've since updated and resolved this.

PR #389, opened 2017-08-13
Show item viewer in the Items by Points Earned -> Student Scores.

jtreuting:
Expanding a student score row will display the Item Viewer with their response.

### DIFF
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -258,7 +258,8 @@ sql:
             ei.exam_id,
             ei.item_id,
             ei.score,
-            ei.position
+            ei.position,
+            ei.response
           from exam_item ei
             join exam e on ei.exam_id=e.id
             join student_group_membership sgm on e.student_id=sgm.student_id
@@ -386,7 +387,8 @@ sql:
             ei.exam_id,
             ei.item_id,
             ei.score,
-            ei.position
+            ei.position,
+            ei.response
           from exam_item ei
             join exam e on ei.exam_id=e.id
             join asmt a on e.asmt_id=a.id
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -40,7 +40,7 @@
     "mathjs": "^3.15.0",
     "ngx-bootstrap": "^1.6.6",
     "ngx-popover": "0.0.16",
-    "primeng": "^4.1.0",
+    "primeng": "^4.1.3",
     "rxjs": "^5.4.2",
     "ts-helpers": "^1.1.1",
     "zone.js": "^0.8.13"
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -40,7 +40,7 @@
     "mathjs": "^3.15.0",
     "ngx-bootstrap": "^1.6.6",
     "ngx-popover": "0.0.16",
-    "primeng": "^4.1.3",
+    "primeng": "4.1.2",
     "rxjs": "^5.4.2",
     "ts-helpers": "^1.1.1",
     "zone.js": "^0.8.13"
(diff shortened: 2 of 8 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -258,7 +258,8 @@ sql:
             ei.exam_id,
             ei.item_id,
             ei.score,
-            ei.position
+            ei.position,
+            ei.response
```
jtreuting: We now need the response for the groups and grades when they go to Item by Points by Earned, which means the response is now heavier, so something to think about.

webapp/src/main/webapp/src/app/assessments/items/item-scores/item-scores.component.html
```diff
@@ -22,4 +24,11 @@
       {{ correctness | number:'1.2-2' }}
     </ng-template>
   </p-column>
+  <ng-template let-score pTemplate="rowexpansion">
+    <div class="tab-content well">
+      <item-viewer [bankItemKey]="item.bankItemKey"
+                   [response]="score.response"
+                   [position]="item.position"></item-viewer>
```
jtreuting: The only thing changing from row to row is the response but the HTTP calls to IRIS happen each time.  I'd like to be able to reuse the item viewer from row to row and just call the javascript for setting the response.  Thoughts on best way to do that?

webapp/src/main/webapp/package.json
```diff
@@ -40,7 +40,7 @@
     "mathjs": "^3.15.0",
     "ngx-bootstrap": "^1.6.6",
     "ngx-popover": "0.0.16",
-    "primeng": "^4.1.0",
+    "primeng": "^4.1.3",
```
jtreuting: I was having an issue getting the row to expand so figured i'd try upgrading to latest since it was a patch, then found out my issue which had nothing to do with that.  I noticed the package-lock.json was at 4.1.2 anyways, so figured this was low risk to go to 4.1.3, but we can revert if there are concerns.

webapp/src/main/webapp/src/app/assessments/items/item-scores/item-scores.component.html
bgsavage: You would need a way to move the same element around on the dom instead of the tables remove/add on collapse / expand.  I think you would need to build your own custom table to make that happen.

If we are flexible on the UI here, another option would be to have a dedicated iris panel that updates based on row selection so that where there's less dom shifting.

webapp/src/main/webapp/src/app/assessments/items/item-scores/item-scores.component.html
```diff
@@ -22,4 +24,11 @@
       {{ correctness | number:'1.2-2' }}
     </ng-template>
   </p-column>
+  <ng-template let-score pTemplate="rowexpansion">
+    <div class="tab-content well">
+      <item-viewer [bankItemKey]="item.bankItemKey"
```
bgsavage: Should we hide the existing item viewer tab without the response then?

jtreuting: We went back and forth on that and Smarter is going to decide if they like it or not.  If we keep it we will add some introduction text to say how it is representative of the item and no response is included so it;s clear.

bgsavage: Alright, just something to consider... the more iris iframes we allow to be rendered on the page simultaneously, the less stable our app becomes.

webapp/src/main/webapp/package.json
```diff
@@ -40,7 +40,7 @@
     "mathjs": "^3.15.0",
     "ngx-bootstrap": "^1.6.6",
     "ngx-popover": "0.0.16",
-    "primeng": "^4.1.0",
+    "primeng": "4.1.2",
```
jtreuting: specific version as 4.1.3 introduced a bug with filtering data tables

PR #390, opened 2017-08-14
Adds spring-session + redis

esotrope:
This RP will come paired with config-repo settings:
```yaml
spring:
  session:
    store-type: redis
  redis:
    host: redis
    port: 6379
```

### This will break k8s images until they include redis/ellasticache.

For some context on spring-session/ellasticache integration/AWS setup see:

1. http://docs.spring.io/spring-session/docs/current/reference/html5/#api-redisoperationssessionrepository-sessiondestroyedevent
2. https://github.com/spring-projects/spring-session/issues/124

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/GroupGrant.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/GroupGrant.java
@@ -1,11 +1,12 @@
 package org.opentestsystem.rdw.reporting.common.security;
 
+import java.io.Serializable;
 import java.util.Objects;
 
 /**
  * This class represents an access grant to a specific group in the system and is used for access control
  */
-public class GroupGrant {
+public class GroupGrant implements Serializable {
 
     private long id;
     private int subjectId;
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/GroupGrant.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/GroupGrant.java
@@ -1,12 +1,11 @@
 package org.opentestsystem.rdw.reporting.common.security;
 
-import java.io.Serializable;
 import java.util.Objects;
 
 /**
  * This class represents an access grant to a specific group in the system and is used for access control
  */
-public class GroupGrant implements Serializable {
+public class GroupGrant {
 
     private long id;
     private int subjectId;
@@ -14,7 +13,8 @@ public class GroupGrant implements Serializable {
     /**
      * No-arg contructor for JSON serialization and deserialization
      */
-    private GroupGrant(){}
+    private GroupGrant() {
+    }
 
     public GroupGrant(final long id, final int subjectId) {
         this.id = id;
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SecuredRedisConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SecuredRedisConfiguration.java
@@ -1,0 +1,29 @@
+package org.opentestsystem.rdw.reporting.security.session;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.session.data.redis.config.ConfigureRedisAction;
+
+/**
+ * This class allows spring-redis to work in a secured redis environment
+ */
+@Configuration
+@ConditionalOnProperty(name = "spring.session.store-type", havingValue = "redis")
+public class SecuredRedisConfiguration {
+
+    /**
+     * This bean informs spring-redis to not configure session expiration and deletion keyspace events.
+     * In a secured environment, these events must be configured manually by a redis admin.
+     *
+     * Without this instruction, spring-redis will try to configure redis to dispatch these events and cause a startup
+     * exception.
+     *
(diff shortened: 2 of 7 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/security/GroupGrant.java
```diff
@@ -1,11 +1,12 @@
 package org.opentestsystem.rdw.reporting.common.security;
 
+import java.io.Serializable;
 import java.util.Objects;
 
 /**
  * This class represents an access grant to a specific group in the system and is used for access control
  */
-public class GroupGrant {
+public class GroupGrant implements Serializable {
```
esotrope: Needed for session replication

webapp/src/main/resources/application.yml
```diff
@@ -23,6 +23,9 @@ saml:
 
 
 spring:
+  session:
+    store-type: none
```
esotrope: Tells spring-session to not do anything when running locally not in docker.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SecuredRedisConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.session.data.redis.config.ConfigureRedisAction;
+
+/**
+ * This class allows spring-redis to work in a secured redis environment
+ */
+@Configuration
+@ConditionalOnProperty(name = "spring.session.store-type", havingValue = "redis")
```
esotrope: This will make sure the configuration only loads if we are using redis.
However, it also makes sure that in our local-docker mode spring wont auto-configure session expiration/deletion keyspace events and they will not work.

One option is to add @Profile("!local-docker") if we think local docker working perfectly is necessary.

webapp/src/main/resources/application.yml
malaffoon: Bit confused here. Why bother using redis in local-docker at all? It's not like you're going to run multiple instances of the reporting app in that environment.

esotrope: This config specifically tells spring session not to kick in by default. This allows us to start the app outside of docker/k8s without failure.

And about docker - you're correct. The idea was mainly to show that spring-session + redis works and that we have it configured correctly before promoting it up the environment chain.

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/GroupGrant.java
wtritch: Interesting, would you rather use java serialization for persisting session data or Jackson with an ObjectMapper?  See [here](https://github.com/SmarterApp/TDS_Common/blob/master/common/src/main/java/tds/common/cache/RedisJsonSerializer.java) for an example redis serializer using jackson.

esotrope: Cool. @wtritch curious though what the drive was to use json?
The only thing i can think of is (1) consistency with api, (2) lenient object versioning.

wtritch: lenient object versioning and for supporting @Cacheable annotations, where some of the cached models didn't support Java serialization.   (We were caching legacy objects and didn't have time to go support Java serialization on all of them.)  Plus, I've been bitten by java serialization issues too many times.

malaffoon: I'd like to weigh in on: java serialization sucks. Doesn't version well, doesn't recurse well, and ghod forbid you have more than a couple objects involved.

esotrope: Ok. JSON sounds like the way to go. I'll add this in.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SecuredRedisConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+    @Primary
+    public RedisTemplate<String, Object> redisTemplate(
+            final RedisConnectionFactory connectionFactory,
+            final ObjectMapper objectMapper) {
+
+        final RedisTemplate<String, Object> template = new RedisTemplate<>();
+        template.setConnectionFactory(connectionFactory);
+        template.setKeySerializer(new StringRedisSerializer());
+        template.setValueSerializer(new RedisJsonSerializer(objectMapper));
+        return template;
```
esotrope: Adds JSON serialization to Redis connection factory

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SecuredRedisConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * TODO: move to org.opentestsystem.rdw.common
+     *
+     * This Redis serializer uses an ObjectMapper to serialize objects.
+     */
+    public class RedisJsonSerializer implements RedisSerializer<Object> {
```
esotrope: @wtritch I added the JSON serializer from TDS as suggested

PR #392, opened 2017-08-15
Feature/disable report download when report will be empty

esotrope:
This PR makes the group, school grade and student assessment results page "Student Report" button appear disabled when the current year selection has no results.

screen shot 2017-08-15 at 11 02 21 am

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/schoolyear/StubSchoolYearService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/schoolyear/StubSchoolYearService.java
@@ -9,12 +9,11 @@ import java.util.Set;
 public class StubSchoolYearService implements SchoolYearService {
 
     /**
-     * @return set containing (2017, 2016, 2015, 2014, 2013)
+     * @return set containing (2017, 2016)
      */
     @Override
     public Set<Integer> getSchoolYears() {
         return ImmutableSet.of(2017, 2016);
-        //return ImmutableSet.of(2017, 2016, 2015, 2014, 2013);
     }
 
 }
--- a/webapp/src/main/webapp/src/app/report/school-grade-report-download.component.ts
+++ b/webapp/src/main/webapp/src/app/report/school-grade-report-download.component.ts
@@ -37,10 +37,14 @@ export class SchoolGradeDownloadComponent extends ReportDownloadComponent {
     this.service.createSchoolGradeExamReport(this.school.id, this.grade.id, this.options)
       .subscribe(
         (report: Report) => {
-          this.notificationService.info({ id: 'labels.reports.messages.submitted.html', html: true });
+          this.notificationService.info({ id: 'labels.reports.messages.create.success-html', html: true });
         },
         (error: any) => {
-          this.notificationService.error({ id: 'labels.reports.messages.submission-failed.html', html: true });
+          this.notificationService.error({
+            id: error.name === 'NotFoundError'
+              ? 'labels.reports.messages.create.school-grade-no-content'
+              : 'labels.reports.messages.create.error'
+          });
         }
       );
   }
--- a/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
+++ b/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
@@ -1,7 +1,9 @@
 <div *ngIf="currentSchool" class="mb-md">
   <span class="h3 blue-dark">{{currentSchool.name}}<span *ngIf="currentGrade"> - {{currentGrade.code | gradeDisplay:'long-name'}}</span></span>
   <span class="pull-right text-right">
-    <button class="btn btn-default btn-sm" (click)="exportCsv()"><i class="fa fa-bold fa-table"></i> {{'labels.export.csv' | translate}}</button>
+    <button class="btn btn-default btn-sm"
+            (click)="exportCsv()"
+            [disabled]="assessmentExams.length == 0"><i class="fa fa-bold fa-table"></i> {{'labels.export.csv' | translate}}</button>
     <span school-grade-report-download #downloader class="ml-xs"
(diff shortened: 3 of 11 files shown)


webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -8,6 +8,7 @@
     <span student-report-download #downloader class="ml-xs"
           [studentId]="examHistory.student.id"
           [schoolYears]="filterState.years"
+          [disabled]="examHistory.exams.length == 0"
```
esotrope: The way we designed the app, I don't think it's possible for the student assessment result history page to have no assessments but I added the disabling just for safe measure.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
```diff
@@ -7,6 +7,7 @@
         [school]="currentSchool"
         [grade]="currentGrade"
         [schoolYears]="filterOptions.schoolYears"
+        [disabled]="assessmentExams.length == 0"
```
esotrope: Currently the app allows the user to select a different year than the page when they open the download dropdown which means the button disabling wont always prevent them from attempting to create a report for a "resultless" year.

screen shot 2017-08-15 at 11 02 03 am

We can solve this by
1. Adding fail fast logic to the backend to return 404 when a user requests a report with zero results. This would come paired with a "There are no results matching your search" notification.
OR
2. Removing school year from the popover and making the page's school year be the school year the report is created for (like the CSV download)

Which is

wtritch: I like disabling both buttons if the results are known to be empty.  I think we'll still need that NO_CONTENT ReportStatus that can be set by the report-processor in the case of regenerated reports.  But that may be a problem for the future?

esotrope: @wtritch Yeah I agree.

@jtreuting did we get any feedback about having the year in the download popover? Do you think it would be confusing if the page's school year was used instead?

bgsavage: I like option 2...  But then the question arises should we be feeding other context to the report as well (such as advanced filters?) . If not, then how will a user now which filters feed the report, and which do not?

esotrope: I think you're right. I think either way we go there will be some confusion to overcome regarding the filters applying to csv and not reports though.

I think for now, I will implement option 1 (as I already have most the code done) and wait for uat feedback for direction on using the page's school year filter or not

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/schoolyear/StubSchoolYearService.java
```diff
@@ -9,12 +9,11 @@
...
 
     /**
-     * @return set containing (2017, 2016, 2015, 2014, 2013)
+     * @return set containing (2017, 2016)
      */
     @Override
     public Set<Integer> getSchoolYears() {
         return ImmutableSet.of(2017, 2016);
-        //return ImmutableSet.of(2017, 2016, 2015, 2014, 2013);
     }
```
esotrope: Created DWR-758 & 759 to address making this not a stub.

webapp/src/main/webapp/src/app/report/school-grade-report-download.component.ts
```diff
@@ -37,10 +37,14 @@ export class SchoolGradeDownloadComponent extends ReportDownloadComponent {
...
         (report: Report) => {
-          this.notificationService.info({ id: 'labels.reports.messages.submitted.html', html: true });
+          this.notificationService.info({ id: 'labels.reports.messages.create.success-html', html: true });
         },
         (error: any) => {
-          this.notificationService.error({ id: 'labels.reports.messages.submission-failed.html', html: true });
+          this.notificationService.error({
+            id: error.name === 'NotFoundError'
+              ? 'labels.reports.messages.create.school-grade-no-content'
+              : 'labels.reports.messages.create.error'
```
esotrope: added 404 handling in group/school grade downloader popovers

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -418,17 +417,20 @@
...
+          "error": "There was an unexpected error while getting your reports. Please click the button below to reload the page and try again.",
+          "no-results-html": "<p>To create a report, first select a group or school from the <a href=\"/\">Home page</a>. Next click the \"Student Reports\" button to get started."
+        },
+        "create": {
+          "success-html": "Creating report. See <a href=\"/reports\">status</a>.",
+          "error": "There was an unexpected error while generating this report.  Please try again..",
+          "student-no-content": "This student does not have any test results for your selection. Please update your selection and try again.",
+          "school-grade-no-content": "This school and assessment grade does not have any test results for your selection. Please update your selection and try again.",
+          "group-no-content": "This group does not have any test results for your selection. Please update your selection and try again."
         },
```
esotrope: remapped the translation keys to hopefully be more logical?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/schoolyear/StubSchoolYearService.java
jtreuting: wasn't this 2018, 2017, 2016, 2015?  i thought i made that change because that is what is expected in production?

esotrope: Don't worry - this isn't getting merged - I thought i closed this already actually.
As for the years I don't know when/why they have been changed.

esotrope:
This issue will be solved by making student reports go through the same process as batch reports.

PR #394, opened 2017-08-15
Bug/673 alignment

aaronsleeper:


package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
```
esotrope: Not sure why this file is here. I think the only one we need is webapp/src/main/webapp/package-lock.json

webapp/src/main/webapp/package-lock.json
```diff
@@ -884,6 +893,11 @@
         "hoek": "2.16.3"
       }
     },
+    "bootstrap": {
+      "version": "3.3.7",
+      "resolved": "https://registry.npmjs.org/bootstrap/-/bootstrap-3.3.7.tgz",
+      "integrity": "sha1-WjiTlFSfIzMIdaOxUGVldPip63E="
+    },
```
esotrope: Curious why we need this. Was this put in manually or just happened when npm build was run?

package-lock.json
bgsavage: Agreed, please remove this file.

webapp/src/main/webapp/src/primeng-overrides.less
```diff
@@ -164,6 +166,18 @@ body .ui-widget.ui-datatable {
...
+    &.level-up {
+      border-right: 1px solid @white;
+    }
+    &.level-down {
+      display: flex !important;
+      border-bottom: 1px solid @table-border-color;
+      border-left: none;
+      overflow: hidden;
+    }
+  }
```
aaronsleeper: admin should have the same styles as front

webapp/src/main/webapp/package-lock.json
aaronsleeper: i didn't make that change. how do i un-include it

bgsavage: Yes, if this is the version of bootstrap @aaronsleeper  wants to lock in.

aaronsleeper: i haven't seen it break anything and i didn't include it . i'm fine keeping it up to date

PR #395, opened 2017-08-15
adds failfast for empty batch reports - like how student reports works

esotrope:
Adds fail fast to batch report generation so that "POST /reports" requests return 404s when the report will have no contents

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
@@ -17,26 +17,37 @@ import javax.servlet.http.HttpServletResponse;
 import javax.validation.constraints.NotNull;
 import java.io.OutputStream;
 import java.util.List;
+import java.util.NoSuchElementException;
 
 @Service
 class DefaultExamReportService implements ExamReportService {
 
     private final ReportWebServiceClient client;
+    private final ExamReportExamRepository repository;
 
 
-    DefaultExamReportService(final ReportWebServiceClient client) {
+    DefaultExamReportService(final ReportWebServiceClient client, final ExamReportExamRepository repository) {
         this.client = client;
+        this.repository = repository;
     }
 
     @Override
     public Report createReport(@NotNull final User user, @NotNull final SchoolGradeExamReportRequest request) {
         final PermissionScope permissionScope = user.getPermissionsById().get("INDIVIDUAL_PII_READ").getScope();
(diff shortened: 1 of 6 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
```diff
@@ -17,26 +17,37 @@
...
+        }
         return createReport(user, request, permissionScope);
     }
 
     @Override
     public Report createReport(@NotNull final User user, @NotNull final GroupExamReportRequest request) {
         final PermissionScope permissionScope = user.getPermissionsById().get("GROUP_PII_READ").getScope();
+        if (!repository.existsForGroupAndSchoolYear(
+                permissionScope, request.getGroupGrant().getId(), request.getGroupGrant().getSubjectId(), request.getSchoolYear())) {
+            throw new NoSuchElementException("There are no exams matching the given search criteria");
```
bgsavage: So just for my own clarification, we have to check for existence here because don't want to queue it up only to find out later in the downstream service that there are no results?

wtritch:
I have to admit, I don't think I'd do any of this.  I'd kick off the report as normal, then if the student list retrieved in FetchStudents is empty, update the Report to NO_CONTENT status and fail the job.  Then the User can still see that they requested a report, it ran, and there were no students, rather than interpreting a 404 as NO_CONTENT, which we also return in other cases, I believe.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
esotrope: You got it.

esotrope:
For context, this is a response to JIRA: https://jira.fairwaytech.com/browse/DWR-722

@wtritch yeah, we  need to decide a consistent behavior here between single/batch reports I think.
Do we want (option 1) to use the above impl and have batch behave more like single student reports? Or, do we want (option 2) single student reports to behave like batch reports?

My vote would be that we do our best to disable report download (without 404s) via the UI knowledge of no report content + making student reports also go through the report service.
We can then add a user_report.type column so that the UI can filter on report type?

esotrope:
Closing PR.
Decided to fix the issue with the following implementation:

1. Making the report download popover respect the current page’s school year filter and remove the popover’s school year filter. This would allow the UI to always disable the “Student Reports” button before they create a report with no content.
2. Making single student reports go through the report-service and also appear in the “My Reports” page like other reports. This would make the user experience consistent for all reports and also help us guard against abusive users who want to create tons of one-off single

PR #397, opened 2017-08-17
Feature/utf8 download names

esotrope:
Silly but working solution for adding support for utf8 filenames for report downloads.
https://jira.fairwaytech.com/browse/DWR-704

screen shot 2017-08-16 at 6 07 20 pm

Suggestions welcome : j

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/web/Headers.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/web/Headers.java
@@ -1,0 +1,119 @@
+package org.opentestsystem.rdw.reporting.common.web;
+
+import javax.mail.internet.MimeUtility;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/web/Headers.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Encode the given header field param as describe in RFC 5987.
+     *
+     * @param input   the header field param
+     * @param charset the charset of the header field param string
+     * @return the encoded header field param
+     * @see {@link org.springframework.http.HttpHeaders#encodeHeaderFieldParam(String, Charset)} for source of derivation
+     */
+    public static String encodeHeaderFieldParam(final String input, final Charset charset) {
```
esotrope: This method is the meat of the encoding logic and is lifted from ```org.springframework.http.HttpHeaders.encodeHeaderFieldParam()```

The other options I explored were:

1. Using ```URLEncoder.encode()``` - was the simplest solution but poses some potential problems and could result in unforeseen bugs down the road
2. Using ```javax.mail.internet.MimeUtility.encodeText()``` - this only supported ASCII encoding of UTF8 which is not simple to decode on the front-end

PR #398, opened 2017-08-17
operational settings

malaffoon:


webapp/build.gradle
```diff
@@ -4,6 +4,10 @@ plugins {
 
 apply plugin: 'org.springframework.boot'
 
+createDockerfile {
+    ext.java_opts = ["-Xmx768m"]
```
malaffoon: I'll set the container memory limit to 1g to go along with this.

webapp/src/main/resources/application.yml
```diff
@@ -21,6 +21,11 @@ saml:
     port: 8080
     include-server-port: false
 
+server:
+  # straw-man values for restricting resource utilization
+  tomcat:
+    max-connections: 100
+    max-threads: 20
```
malaffoon: I find a ton of conflicting advice for this all the way from one thread per core to 150 threads per core, and 2-3 connections per thread up to 50 connections per thread. This is a compromise.

PR #402, opened 2017-08-22
Feature/admin link

esotrope:


webapp/src/main/resources/application.yml
```diff
@@ -102,6 +102,8 @@ app:
   reportservice:
     url: http://localhost:8083
 
+  admin-webapp-url: /
```
esotrope: not sure what a sensible default for this would be.

jtreuting: not sure there is one.  does it make sense to have it be blank, and then if it isn't configured the Manage Student Groups button doesn't show?  Could give the flexibility to just use the Admin URL directly instead of as part of the reporting app, if someone wanted to totally separate them.

esotrope: That would be good yeah.

webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -32,9 +32,11 @@ <h1 class="blue-dark h1 mb-md">{{'labels.homepage.title' | translate}}</h1>
     <!-- Right Column Content Area -->
     <div class="col-md-3 pt-lg">
       <!-- Student Groups Admin Link -->
-      <p class="text-right" *sbAuthorize="['GROUP_WRITE']">
-        <a href="#" class="btn btn-xs btn-info"><i class="fa fa-gear"></i> {{'labels.groups.manage-group' | translate}}</a>
-      </p>
+      <span *ngIf="user.configuration.adminWebappUrl.length != 0">
```
esotrope: default is now blank string and the link is not displayed if it is blank

PR #406, opened 2017-08-25
Defect/enrolled grade used instead of asmt grade

esotrope:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -125,7 +125,7 @@
       },
       "11": {
         "long-name": "Grade 11",
-        "short-name": "G11"
+        "short-name": "High School"
```
esotrope: Do we want G9-12 to be "High School" ? or just G11?

malaffoon: i think the short form is `HS`?

esotrope: Yeah I'm looking at the UI and realized i need to be setting the long-name as well.
Same Q stands though - do we want G9-12 to say HS?

jtreuting: Yeah i think short name should be `HS`

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -124,8 +124,8 @@
         "short-name": "G10"
       },
       "11": {
-        "long-name": "Grade 11",
-        "short-name": "G11"
+        "long-name": "High School",
+        "short-name": "HS"
```
esotrope: changed grade 11 text to HS/High School

esotrope: changed grade 11 text to HS/High School

PR #408, opened 2017-08-25
changing dockerhub naming

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/408)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,6 +34,7 @@ apply plugin: 'io.spring.dependency-management'
 
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "0.0.1-36"
 String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-128"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -34,6 +34,7 @@ apply plugin: 'io.spring.dependency-management'
 
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "0.0.1-36"
 String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-128"
+String dockerPrefix = "smarterbalanced"
```
esotrope: They didn't choose opentestsystem?

malaffoon:
We should change the docker-compose file and README as well.  

Basically, ctrl-shift-F for "fwsbac" and get rid of it.

build.gradle
rcordeau: guess not

PR #409, opened 2017-08-25
DWR-823 Display answer key in item viewer.

wtritch:
This PR displays a simple answer key if available at the top of the item viewer.  If no simple answer key is available, it displays a message letting the user know that more information can be found in the rubric tab.
If no rubric available at all, no message is displayed to the user.

I considered caching the rubric/exemplar information should the user navigate to the exemplar screen, but decided against it since we don't have a good expiring cache impl and I don't want to continually shove rubric information into memory.

dwr_823_complex_answer
dwr_823_with_answer

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
@@ -6,7 +6,7 @@
   </div>
   <div *ngIf="rubricModel" style="white-space: normal">
     <div *ngIf="rubricModel.answerKeyValue" class="mb-md">
-      <b>{{ 'labels.assessments.items.tabs.exemplar.answer-key' | translate }}</b> {{rubricModel.answerKeyValue}}
+      <strong>{{ 'labels.assessments.items.tabs.exemplar.answer-key' | translate }}</strong> {{rubricModel.answerKeyValue}}
     </div>
     <div *ngIf="!rubricModel.answerKeyValue" class="mb-md">
       {{ 'labels.assessments.items.tabs.item-viewer-see-rubric' | translate }}
--- a/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
@@ -4,5 +4,14 @@
       <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'messages.loading' | translate }}</span>
     </div>
   </div>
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
```diff
@@ -4,5 +4,14 @@
       <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'messages.loading' | translate }}</span>
     </div>
   </div>
+  <div *ngIf="rubricModel" style="white-space: normal">
+    <div *ngIf="rubricModel.answerKeyValue" class="mb-md">
+      <b>{{ 'labels.assessments.items.tabs.exemplar.answer-key' | translate }}</b> {{rubricModel.answerKeyValue}}
```
esotrope: I think ```<strong>``` is advised for bold instead of ```<b>```. Can't tell you why, but I notice bootstrap seems to favor it too.

wtritch: doh, I think you're right.  <strong> tags are generally preferred.  Wish I could remember why...

bgsavage:
So.... then are we making the rubric / exemplar api call twice?  Once when we open the item-viewer tab, and once we when open the Rubric / Exemplar tab?

wtritch:
@bgsavage yes, however without metrics for actual user flows and/or load-tests we won't know how large of an issue that is.
Hmm... I suppose we could store the rubric data in the parent item-tab.component.ts?
What I'd *like* is a session-storage-backed LRU/TTL cache, but I couldn't find a good library and didn't have time to build one.

chdietz:
@wtritch thinking about this after you mentioned it Friday, if this is a nice to have feature should we make sure it's efficient before we add it?

wtritch:
There are two inefficiencies here:
1. We request the rubric for a NON-answer-key rubric (non-multiple-choice)
2. We request the rubric twice for a single answer *if* the user opens the rubric/exemplar tab after opening the item-viewer tab.

Both are single-call-per-user-action, as opposed to a systematic, wide-spread inefficiency.  i.e. at worst, it's like having 2 users browsing the item-viewer and exemplar tabs at once rather than 1.  While it could be improved, I don't believe this will make the difference between system success and failure even under heavy load.  We can certainly discuss in the parking lot, however.

jtreuting:
It was a pretty high priority feedback request from smarter and CA.  We also have the option to add caching on the server side if we want since the Item XML will not be changing unless there is a fix that happens.  So having an hour cache or something would not hurt (in my opinion)

PR #410, opened 2017-08-27
Don't show the same exact notification multiple times.  This prevents…

jtreuting:
… multiple messages when the same link is clicked.

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/notification/notification.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/notification/notification.component.ts
@@ -36,9 +36,11 @@ export class NotificationComponent {
    * @param notification A notification
    */
   private onNotification(notification) {
-    if (!this.notifications.some(x => _.isEqual(x, notification))) {
-      this.notifications.push(notification);
-    }
+    // if the same message is being generated again then remove it so a new one can be added
+    _.remove(this.notifications, x => _.isEqual(x, notification));
+
+    // add a brief pause to this so that Angular can remove it from display and a visual reference is made so the user knows the existing was replaced
+    setTimeout(() => this.notifications.push(notification), 50);
   }
 
 }
--- a/webapp/src/main/webapp/src/app/shared/notification/notification.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/notification/notification.component.ts
@@ -1,6 +1,7 @@
-import { Component, OnInit } from "@angular/core";
+import { Component } from "@angular/core";
 import { NotificationService } from "./notification.service";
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/shared/notification/notification.component.ts
```diff
@@ -35,7 +36,9 @@ export class NotificationComponent {
    * @param notification A notification
    */
   private onNotification(notification) {
-    this.notifications.push(notification);
+    if (!this.notifications.some(x => _.isEqual(x, notification))) {
+      this.notifications.push(notification);
```
bgsavage: Will this prevent the future error notifications from showing if the context / cause was different?

esotrope: If it's easy it would be cool to not put a new message but to reset the timeout of the message if it's the same. I think the easy way to do that is to remove the existing same message and just push the new one possibly.

bgsavage: I'd be good with that if we had some kind of ui / animation to indicate the error happened again.

jtreuting: the reason i'm doing the equality check on the object itself is because a message ID is provided and `args` to do the translation.  So I couldn't just check on the type of message in case it is actually different.  It only impacts when the notification is actually displayed as it gets removed from the notifications array when it goes away from the UI.

i'll test what it looks like display-wise with removing and adding.  I'm guessing the user won't be able to notice unless there is some delay between those actions so angular can remove the currently displayed one and redraw a new one.

jtreuting: in order to get the notification to hide then show again so that there is an indicator that a new one replaced the other one, we have to do a `setTimeout` to get Angular time to process it.  Here is what it looks like:

    _.remove(this.notifications, x => _.isEqual(x, notification));
    
    setTimeout(() => this.notifications.push(notification), 50);

Not sure if I like this.  Thoughts?

esotrope: Looks like a good enough solution assuming it looks ok.

bgsavage: Does this render on slower browsers ie IE ?

jtreuting: i have no idea.  we can see once it's in QA :) i can't test that locally.

also, it would just extend the length of when the notification is visible which is not a bad degraded state.

PR #411, opened 2017-08-29
Unlock metrics endpoints, since they're on a non-exposed port

wtritch:
This removes SSO security from all metrics endpoints.


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -519,8 +524,14 @@ protected void configure(HttpSecurity http) throws Exception {
                 .addFilterAfter(samlFilter(), BasicAuthenticationFilter.class);
         http
                 .authorizeRequests()
-                .antMatchers("/saml/**").permitAll()
-                .antMatchers("/{status:(status|health|info|env|configprops|autoconfig|loggers|metrics|mappings|trace|dump|beans)}/**").permitAll()
+                .antMatchers("/saml/**").permitAll();
+
+        for (final AbstractEndpoint endpoint : actuatorEndpoints) {
```
malaffoon: thanks

PR #412, opened 2017-08-29
fixes report polling

esotrope:


webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -40,6 +40,10 @@ export class ReportsComponent implements OnInit {
     }
   }
 
+  ngOnDestroy(): void {
+    this.stopPollingStatus();
```
esotrope: Without this it was polling even when you leave the page.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
```diff
@@ -50,9 +50,11 @@
 
     @Override
     public List<Report> getReports(final String userId, final Iterable<Long> reportIds) {
+        final String expandedReportIds = StreamSupport.stream(reportIds.spliterator(), false)
+                .map(a -> "&id=" + a.toString()).reduce("", String::concat);
         final ResponseEntity<List<Report>> response = template.exchange(
-                "{url}/api/reports?user={userId}&id={reportIds}", HttpMethod.GET, null,
-                new ParameterizedTypeReference<List<Report>>() {}, url, userId, reportIds);
+                url + "/api/reports?user={userId}" + expandedReportIds, HttpMethod.GET, null,
```
esotrope: Hand jammed this URL expansion. spring's ```UriComponentsBuilder``` expanded it as ```&id=[...]``` instead of repeated params. apache's ```URIBuilder``` did the same and the spring controllers don't accept this for some reason.

webapp/src/main/webapp/src/app/report/reports.component.ts
bgsavage: 👍

PR #413, opened 2017-08-29
applies session-store of groups as whitelist to group list ui

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/group/DefaultGroupService.java
```diff
@@ -24,7 +25,9 @@ public DefaultGroupService(@NotNull final GroupRepository groupRepository) {
         if (!user.getPermissionsById().containsKey("GROUP_PII_READ")) {
             return ImmutableSet.of();
         }
-        return newHashSet(groupRepository.findAllForUsername(user.getUsername()));
+        return groupRepository.findAllForUsername(user.getUsername()).stream()
```
malaffoon: I think classic collection functions like retainAll would be more efficient (avoid creating extra containers) but this'll work ...

esotrope: I think retain all relies on ```Object.equals``` though which i don't think we want here.

bgsavage: I thought you would want to pass the group ids to the query like we do everywhere else.

esotrope: True, that would be slightly cleaner - thanks

PR #414, opened 2017-08-29
changes group whitelist to database driven whitelisting + fixes tests

esotrope:


webapp/src/main/resources/application.sql.yml
```diff
@@ -134,6 +134,7 @@ sql:
         join student_group sg on sg.id = usg.student_group_id
         join school on school.id = sg.school_id
       where user_login=:user_login
+      and sg.id in (:group_ids)
```
esotrope: Just curious, Is there any performance gained by making this ```user_student_group.student_group_id```?

malaffoon: Theoretically, no. However, you'd have to `EXPLAIN` to see if it is affecting the optimizer.

PR #415, opened 2017-08-29
Bug/dwr 656 school typeahead

bgsavage:
If the # of schools > 25, then show the user a typeahead instead of a dropdown.

image

Right now this is a client side typeahead as the schools are loaded in already with the usercontext.  We may want to consider separating that out to a separate call, which could then throw a specific error the results are too many, at which point we switch to a server side typeahead.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/School.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/School.java
@@ -1,0 +1,21 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+/**
+ * Represents a school
+ */
+public class School extends Organization {
+
+    private final Organization district;
+
+    public School(long id, String name, Organization district) {
+        super(id, name);
+        this.district = district;
+    }
+
+    /**
+     * @return The district that this school belongs to.
+     */
+    public Organization getDistrict() {
+        return district;
+    }
+}
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -470,9 +470,10 @@ sql:
       where natural_id in (:naturalIds)
 
     findAll: >-
-      select name, id
-      from school
-      where (1=:statewide or district_id in (:district_ids) or id in (:school_ids))
+      select s.name, s.id, d.id as district_id, d.name as district_name
+      from school s
+        join district d on s.district_id = d.id
+      where (1=:statewide or s.district_id in (:district_ids) or s.id in (:school_ids))
 
     findAllForNameContainingText: >-
       select name, id
--- a/webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
+++ b/webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
@@ -4,16 +4,7 @@
       <label>
         {{'labels.school-grade.school-name.label' | translate}}
       </label>
-      <div>
-        <p-dropdown id="rdw-school-grade-select-school"
-                    [options]="availableSchools"
-                    formControlName="school"
-                    [filter]="true"
-                    filterBy="label"
-                    placeholder="{{'prompt.select' | translate}}"
-                    filterPlaceholder="{{'labels.school-grade.school-name.prompt' | translate}}">
-        </p-dropdown>
-      </div>
+      <school-select [availableSchools]="availableSchools" (schoolChanged)="schoolChanged($event)"></school-select>
     </div>
     <div class="flex-child">
       <label for="grade-select">{{'labels.grade' | translate}}</label>
--- a/webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
+++ b/webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
@@ -54,15 +54,7 @@ export class SchoolGradeComponent implements OnInit {
     }
   }
 
-  private get gradeControl() {
-    return this.searchForm.controls[ "grade" ];
-  }
-
-  private get schoolControl() {
-    return this.searchForm.controls[ "school" ];
-  }
-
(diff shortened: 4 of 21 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/School.java
```diff
@@ -0,0 +1,21 @@
+package org.opentestsystem.rdw.reporting.exam;
+
+/**
+ * Represents a school
+ */
+public class School extends Organization {
```
bgsavage: New school model which extends Organization and has a district.

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
```diff
@@ -4,16 +4,7 @@
...
-        <p-dropdown id="rdw-school-grade-select-school"
-                    [options]="availableSchools"
-                    formControlName="school"
-                    [filter]="true"
-                    filterBy="label"
-                    placeholder="{{'prompt.select' | translate}}"
-                    filterPlaceholder="{{'labels.school-grade.school-name.prompt' | translate}}">
-        </p-dropdown>
-      </div>
+      <school-select [availableSchools]="availableSchools" (schoolChanged)="schoolChanged($event)"></school-select>
```
bgsavage: New school select component to prevent dup code.

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.html
```diff
@@ -0,0 +1,22 @@
...
+    placeholder="{{'prompt.select' | translate}}"
+    filterPlaceholder="{{'labels.school-grade.school-name.prompt' | translate}}">
+  </p-dropdown>
+
+  <input *ngIf="!showDropdown"
+         [(ngModel)]="selectedSchoolName"
+         (ngModelChange)="schoolTextChanged($event)"
+         [typeahead]="availableSchools"
+         typeaheadOptionField="name"
+         typeaheadGroupField="districtName"
```
bgsavage: Groups by district

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  selectedSchool: School;
+
+  selectedSchoolName: string;
+  schoolsMatchingSearch: School[] = [];
+  schoolDropdownOptions: any[] = [];
+  constructor() {
+  }
+
+  get showDropdown() {
+    return this.availableSchools && this.availableSchools.length < 25;
```
esotrope: Would someone really want to scroll through 25 schools? I would make this 10 or 15 max before we cutover to a typeahead.

webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
```diff
@@ -77,6 +63,7 @@ export class SchoolGradeComponent implements OnInit {
   }
 
   private schoolChanged(school: School) {
+    this.schoolControl.setValue(school);
```
esotrope: What is this for? Does this sync form controls with form state? for some reason i thought angular did that for us no?

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
bgsavage: I thought 25 was still reasonable as the dropdown does still have a search.

webapp/src/main/resources/application.sql.yml
```diff
@@ -471,9 +471,10 @@ sql:
       where natural_id in (:naturalIds)
 
     findAll: >-
-      select name, id
-      from school
-      where (1=:statewide or district_id in (:district_ids) or id in (:school_ids))
+      select s.name, s.id, d.id as district_id, d.name as district_name
+      from school s
+        join district d on s.district_id = d.id
+      where (1=:statewide or s.district_id in (:district_ids) or s.id in (:school_ids))
```
jtreuting: @malaffoon We are fine doing this join because we are rejecting schools that don't have a district as the parent and just have the state right?

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
jtreuting: For what it's worth.  It looks like there are 1900 districts, 1774 have less than 25 schools and 1717 have less than 20 schools and 1528 have less than 10 schools

webapp/src/main/resources/application.sql.yml
malaffoon: Correct, district_id is not null.

malaffoon: It seems like we could pass a "page size" parameter down the stack so we don't waste time and effort pulling 14000 schools. Or is that in here somewhere and i'm not seeing it.

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
malaffoon: For either the list of <25 or the type-ahead, i don't suppose this fixes the issue that you can't just arrow down into the list?

webapp/src/main/resources/application.sql.yml
bgsavage: We are pulling 14000 schools right now, no paging.  This has been implemented as a client side typeahead as noted.  To move this to a server side type ahead, it will require further refactoring.

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
bgsavage: The typeahead is keyboard friendly.

PR #416, opened 2017-08-29
adds guard against null/unscored claims in icas

esotrope:
Resolution for DWR-782.

Should we display an explanation as to why the claim might not be scored?

screen shot 2017-08-29 at 3 28 44 pm


jtreuting:
No I think that is good.

PR #417, opened 2017-08-29
DWR-758 Datamart: add school_year table to datamart

agorina:
This is due to https://github.com/SmarterApp/RDW_Schema/pull/143.
School_year is not preloaded into reporting. Hence the test script need to insert it.

I still need to make a change to read the school years from the db.


webapp/src/test/resources/integration-test-data.sql
```diff
@@ -17,6 +17,10 @@ insert into grade (id, code, name) values
 insert into gender (id, code) values
   (-1, 'g1');
 
+insert into school_year (year) values
+  (2016),
+  (2017);
```
esotrope: i think in practice we will only make a JdbcSchoolYearRepository in the reporting module and it will use this file to init its data so we probably only need the inserts here.

common/src/test/java/org/opentestsystem/rdw/reporting/common/report/ExamReportMessageConverterTest.java
```diff
@@ -68,7 +68,7 @@ public ExamReportAssessmentType getAssessmentType() {
 
             @Override
             public int getSchoolYear() {
-                return 2017;
+                return 1997;
```
agorina: I had to change all 2017 to 1997 and 2016 to 1996.

In a real life the reporting should not have any data pre-loaded into school_year table. But because the incremental updates are meant to be applied to the db with the data, it inserts 2005-2018. So the test that were inserting this year had a conflict.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/schoolyear/DefaultSchoolYearService.java
```diff
@@ -0,0 +1,25 @@
+package org.opentestsystem.rdw.reporting.search.schoolyear;
```
agorina: This is a real change to read school years from the db.

PR #418, opened 2017-08-29
Updated Enrolled Grade to display G9, G10, G11, G12 instead of being …

jtreuting:
…tied to the assessment grade where grade 11 is HS for High School.  Needed them to be independent of each other

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -82,83 +82,103 @@
     "grades": {
       "KG": {
         "long-name": "Kindergarten",
-        "short-name": "KG"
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -82,83 +82,103 @@
...
-        "short-name": "G10"
+        "long-name": "High School",
+        "short-name": "HS",
+        "enrolled-name": "G10"
       },
       "11": {
         "long-name": "High School",
-        "short-name": "HS"
+        "short-name": "HS",
+        "enrolled-name": "KG"
```
malaffoon: this can't be right?

PR #419, opened 2017-08-29
DWR-201 proxy and secure requests to Iris

wtritch:
This PR removes the need for the front-end client to directly communicate with IRIS.  Instead, IRIS requests are proxied through the back-end via requests to "/iris/**"

This allows us to provide security to all IRIS requests via standard SSO authentication.

Note that since IRIS is still sending mis-labeled JSON requests as "application/x-www-form-urlencoded" a relatively ugly workaround is required to pass the JSON content through the Zuul proxy.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
@@ -9,9 +9,6 @@ class AppSettings {
     @Value("${app.iris.vendorId}")
     private String irisVendorId;
 
-    @Value("${app.iris.url}")
-    private String irisUrl;
-
     @Value("${app.analytics.trackingId}")
     private String analyticsTrackingId;
 
@@ -32,13 +29,6 @@ class AppSettings {
      */
     public String getIrisVendorId() {
         return irisVendorId;
-    }
-
-    /**
-     * @return the Iris url in which to embed the Iris iframe.
-     */
-    public String getIrisUrl() {
-        return irisUrl;
     }
 
     /**
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
@@ -1,0 +1,197 @@
+package org.opentestsystem.rdw.reporting.zuul;
+
+import com.netflix.zuul.context.RequestContext;
+import com.netflix.zuul.http.HttpServletRequestWrapper;
+import com.netflix.zuul.http.ServletInputStreamWrapper;
+import org.springframework.cloud.netflix.zuul.filters.pre.FormBodyWrapperFilter;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpOutputMessage;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.StringHttpMessageConverter;
+import org.springframework.util.AntPathMatcher;
+import org.springframework.util.Assert;
+import org.springframework.util.ReflectionUtils;
+import org.springframework.util.StringUtils;
+
+import javax.servlet.ServletInputStream;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletRequestWrapper;
+import javax.servlet.http.HttpServletRequest;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.reflect.Field;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static com.google.common.base.Charsets.UTF_8;
+import static org.apache.commons.lang3.StringUtils.EMPTY;
+
+/**
+ * This class is mostly taken from {@link FormBodyWrapperFilter} which is responsible for
+ * converting a client request with content-type MediaType.APPLICATION_FORM_URLENCODED or
+ * MediaType.MULTIPART_FORM_DATA into a forwarded request to the proxied service.
+ *
+ * This class is responsible for being aware that IRIS requests are malformed and although
+ * they advertise a content-type of MediaType.APPLICATION_FORM_URLENCODED, the content actually
+ * consists of JSON.  This class re-interprets the form key of:
+ * {"items":[{"id":"I-xxx-xxxxx"}]}
+ * with an empty-string value as a JSON payload and wraps the client request in a suitable
+ * RequestWrapper.
+ */
+public class IrisAwareFormBodyWrapperFilter extends FormBodyWrapperFilter {
+
+    private static final StringHttpMessageConverter converter = new StringHttpMessageConverter(UTF_8);
+    private static final AntPathMatcher PathMatcher = new AntPathMatcher();
+    private static final String IrisPathPattern = "/iris/**";
+
+    private final Field requestField;
+    private final Field servletRequestField;
+
+    public IrisAwareFormBodyWrapperFilter() {
+        this.requestField = ReflectionUtils.findField(HttpServletRequestWrapper.class,
+                "req", HttpServletRequest.class);
+        this.servletRequestField = ReflectionUtils.findField(ServletRequestWrapper.class,
+                "request", ServletRequest.class);
+        Assert.notNull(this.requestField,
+                "HttpServletRequestWrapper.req field not found");
+        Assert.notNull(this.servletRequestField,
+                "ServletRequestWrapper.request field not found");
+        this.requestField.setAccessible(true);
+        this.servletRequestField.setAccessible(true);
+    }
+
+    @Override
+    public Object run() {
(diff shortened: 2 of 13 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
```diff
@@ -9,9 +9,6 @@
     @Value("${app.iris.vendorId}")
     private String irisVendorId;
 
-    @Value("${app.iris.url}")
-    private String irisUrl;
```
wtritch: Since the front-end no longer requires the IRIS url, I stripped it out of the UserContext sent back to the front-end.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
```diff
@@ -0,0 +1,197 @@
...
+        this.servletRequestField.setAccessible(true);
+    }
+
+    @Override
+    public Object run() {
+        final RequestContext ctx = RequestContext.getCurrentContext();
+        final HttpServletRequest request = ctx.getRequest();
+        if (!PathMatcher.match(IrisPathPattern, request.getRequestURI())) {
+            return super.run();
+        }
```
wtritch: If we're not making a call to "/iris/**" then this delegates to the standard parent Zuul FormBodyWrapperFilter

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
```diff
@@ -0,0 +1,197 @@
...
+                "ServletRequestWrapper.request field not found");
+        this.requestField.setAccessible(true);
+        this.servletRequestField.setAccessible(true);
+    }
+
+    @Override
+    public Object run() {
+        final RequestContext ctx = RequestContext.getCurrentContext();
+        final HttpServletRequest request = ctx.getRequest();
+        if (!PathMatcher.match(IrisPathPattern, request.getRequestURI())) {
```
esotrope: Can we not 'precompile' this pattern for matching?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
```diff
@@ -0,0 +1,197 @@
...
+            return super.run();
+        }
+
+        final FormBodyRequestWrapper wrapper;
+        if (request instanceof HttpServletRequestWrapper) {
+            final HttpServletRequest wrapped = (HttpServletRequest) ReflectionUtils
+                    .getField(this.requestField, request);
+            wrapper = new FormBodyRequestWrapper(wrapped);
+            ReflectionUtils.setField(this.requestField, request, wrapper);
+            ReflectionUtils.setField(this.servletRequestField, request, wrapper);
```
wtritch: This conditional was copied from the Zuul implementation.  I believe it has to do with the Zuul-wrapped request no longer being able to provide accurate request.getParameterMap() results.  Or request.getInputStream() results...

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
```diff
@@ -0,0 +1,197 @@
...
+    private static final String IrisPathPattern = "/iris/**";
+
+    private final Field requestField;
+    private final Field servletRequestField;
+
+    public IrisAwareFormBodyWrapperFilter() {
+        this.requestField = ReflectionUtils.findField(HttpServletRequestWrapper.class,
+                "req", HttpServletRequest.class);
+        this.servletRequestField = ReflectionUtils.findField(ServletRequestWrapper.class,
+                "request", ServletRequest.class);
```
wtritch: The problem with pulling back the covers on some projects is that it exposes the fugly...

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
```diff
@@ -0,0 +1,197 @@
...
+        private synchronized void buildContentData() {
+            try {
+                String payload = EMPTY;
+                final Set<String> queryParams = findQueryParams();
+                for (final Map.Entry<String, String[]> entry : this.request.getParameterMap().entrySet()) {
+                    if (queryParams.contains(entry.getKey())) continue;
+
+                    payload = entry.getKey();
+                    break;
+                }
```
wtritch: This is the custom logic.  The POST content has been interpreted as a request parameter due to the mislabeled content-type header.  I retrieve the "parameter" that is not a query param AND has no assigned value as the JSON payload.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
```diff
@@ -0,0 +1,197 @@
...
+
+                final FormHttpOutputMessage data = new FormHttpOutputMessage();
+                this.contentType = MediaType.valueOf(this.request.getContentType());
+                data.getHeaders().setContentType(this.contentType);
+                converter.write(payload, this.contentType, data);
+                // copy new content type including multipart boundary
+                this.contentType = data.getHeaders().getContentType();
+                this.contentData = data.getInput();
+                this.contentLength = this.contentData.length;
```
wtritch: This re-writes the payload as a "String" HttpOutputMessage.  I then grab the generated message's contents and use them as this wrapper's request attributes.  (Taken from the original Zuul implementation.)

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
jtreuting: it is publicly available since all open source uses of IRIS are directed to use that same ID.  so it's not an issue having it out in the public

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/Servlet30RequestWrapper.java
```diff
@@ -0,0 +1,26 @@
+package org.opentestsystem.rdw.reporting.zuul;
+
+import com.netflix.zuul.http.HttpServletRequestWrapper;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Duplicated from {@link org.springframework.cloud.netflix.zuul.filters.pre.Servlet30RequestWrapper}
```
wtritch: As the comment says, this is cloned from a Zuul class that is marked as package-local.  If/when Zuul is fixed/updated, this may require modification/cleanup.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
esotrope: In an ideal situation (maybe a future situation?) we wont need the wrapper filter/ request wrapper right?
I would try to keep it where it is ideal wherever that is. Is it possible to bake it into the zuul config?

webapp/src/main/resources/application.yml
```diff
@@ -108,3 +108,16 @@ app:
...
+  set-content-length: true
+  FormBodyWrapperFilter:
+    pre: disable
+  host:
+    socket-timeout-millis: 100000
+  routes:
+    iris:
+      path: /iris/**
+      url: ${app.iris.url}
+      stripPrefix: true
```
wtritch: This configures the zuul proxy to route requests to the configured Iris server.

webapp/src/main/webapp/proxy.conf.json
```diff
@@ -2,5 +2,9 @@
   "/api": {
     "target": "http://localhost:8080",
     "secure": false
+  },
+  "/iris": {
+    "target": "http://localhost:8080",
+    "secure": false
```
wtritch: This enables the iris proxy when running the front-end development environment on port 4200.  (Yes, we proxy to a proxy to a server.  And it works.)

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
wtritch: We couldn't bake it into the application.yml zuul config.  We'd need to register a filter that adds it to a wrapped HttpServletRequest.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareFormBodyWrapperFilter.java
wtritch: I didn't see a way to do that using the AntPathMatcher.  If I wanted to hand-jam some regex, I suppose I could.  I'm tempted to use Ant patterns when it comes to matching paths, however...

PR #421, opened 2017-08-30
Feature/public landing page

bgsavage:
Created a new public landing page that will be shown when a user navigates to "/" and is not logged in.

The page will always show if navigated directly (i.e. an existing bookmark) via "/assets/public/landing.html".

image


webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -1,14 +1,20 @@
...
 
 @Controller
 public class ViewController implements ErrorController {
 
 	@RequestMapping(value = {"/"})
-	public String index() {
+	public String index(@AuthenticationPrincipal final User user) {
+		if(user == null) {
+			return "forward:/assets/public/landing.html";
+		}
```
bgsavage: The "magic".  Try to get the AuthenitcationPrincipal, and if it is null, this means the user is not logged in.  Send them to the landing page.

webapp/src/main/resources/application.yml
```diff
@@ -46,6 +46,9 @@ spring:
   thymeleaf:
     mode: HTML
 
+security:
+  filter-dispatcher-types: FORWARD, INCLUDE, REQUEST, ASYNC, ERROR
```
bgsavage: This was required in order for the ViewController to get the AuthenticationPrincipal after a redirect from the `NotFoundIndexTemplate` when spring is navigating to a nested angular route (i.e. a refresh on the assessment results page).

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -58,6 +58,11 @@ const studentTestHistoryChildRoute = {
 
 export const routes: Routes = [
   {
+    path: 'home',
+    redirectTo: '',
+    pathMatch: 'full'
```
bgsavage: Created a 'home' route that the landing page navigates too.  It's sole purpose is just to trigger the sso login and then once we get here, we redirect back to root (which will be the authenticated home page now).

webapp/src/main/webapp/src/assets/public/landing.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+  <title>Smarter Balanced Reporting</title>
+  <link rel="stylesheet" type="text/css" href="/assets/public/bootstrap.min.css">
+  <link rel="stylesheet" type="text/css" href="/assets/public/ui.jqgrid.css">
+  <link rel="stylesheet" type="text/css" href="/assets/public/style.css">
```
bgsavage: Notice I'm going to /assets/public/* for my resources.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -518,6 +518,8 @@ protected void configure(HttpSecurity http) throws Exception {
                 .addFilterAfter(samlFilter(), BasicAuthenticationFilter.class);
         http
                 .authorizeRequests()
+                .antMatchers("/").permitAll()
+                .antMatchers("/assets/public/**").permitAll()
```
bgsavage: Allow public access to the root path (which will redirect to landing page) and then the public folder which has all the static resources required by the landing page.  

@wtritch We discussed a different matching pattern, but does this work for you?  I thought it might be better to only expose items in the `/assets/public/` route.

webapp/src/main/webapp/src/assets/public/landing.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
+<html>
```
bgsavage: This file and related resources are taken from the existing reporting system: https://reporting.smarterbalanced.org/assets/public/landing.html

I trimmed down to just what we want to display.

webapp/src/main/webapp/src/assets/public/style.css
```diff
@@ -0,0 +1,738 @@
+/*!
```
bgsavage: This file could probably be trimmed down further, but was trimmed down from 6000 lines to just what I found the landing page to be using.  There might be more unused classes here.

webapp/src/main/webapp/src/assets/public/style.css
```diff
@@ -0,0 +1,738 @@
...
+(c) 2014 The Regents of the University of California. All rights reserved,
+subject to the license below.
```
bgsavage: Does this copyright need to be updated?  Or?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -1,14 +1,20 @@
...
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.RequestMapping;
 
 @Controller
 public class ViewController implements ErrorController {
 
 	@RequestMapping(value = {"/"})
-	public String index() {
+	public String index(@AuthenticationPrincipal final User user) {
+		if(user == null) {
```
malaffoon: nit: whitespace betwixt if and paren

webapp/src/main/webapp/src/assets/public/style.css
malaffoon: In theory, the copyright should include all years that the thing being copyrighted was published. In practice people go two ways: the first year of publication (i.e. (c) 2014) or range to current (i.e. (c) 2014-2017) because software is "published" all the time. 

I am not a lawyer and don't play one on TV.

jtreuting: According to my lawyers advice (@malaffoon) I think we should leave it as-is right now.

webapp/src/main/webapp/src/assets/public/bootstrap.min.css
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Bootstrap v2.3.2
```
esotrope: we couldn't use BS 3.x in the node_modules?

webapp/src/main/webapp/src/assets/public/landing.html
esotrope: There a plan to incorporate this into the existing styles.less/bs3 markup & styles instead of these css deps?

bgsavage: I'd like to keep this page as it's own separate entity and avoid any cross dependencies between the angular app and this static page since one is exposed to the open world and the other is behind a security layer.

PR #422, opened 2017-08-30
Feature/advanced filter fab

esotrope:
Before click:
screen shot 2017-08-30 at 4 14 09 pm

After click:
screen shot 2017-08-30 at 4 14 23 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -227,6 +227,13 @@ export class AssessmentsComponent implements OnInit {
     this.onExportItemsByPointsEarned.emit(exportRequest);
   }
 
+  openAndScrollToAdvancedFilters() {
+    this.showAdvancedFilters = true;
+    setTimeout(() => {
+      document.getElementById('results-adv-filters').scrollIntoView();
+    }, 0);
+  }
+
   private updateFilterOptions() {
     this.filterOptions.hasInterim = this.selectedAssessments.some(a => a.isInterim);
     this.filterOptions.hasSummative = this.selectedAssessments.some(a => a.isSummative);
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -235,7 +235,7 @@
             "ethnicity": "Race/Ethnicity",
             "ethnicity-info": "Student race and ethnicity"
           },
-          "fab": "Filter Results"
+          "floating-action-button": "Filter Results"
         },
         "selected-assessments-title": "Selected Assessments",
         "selected-assessments-instruct": "Click an assessment to remove it from the list",
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -234,7 +234,8 @@
             "limited-english-proficiency-info": "Students who use a language other than English as their primary language and who have been assessed as not possessing the English language skills necessary to succeed in the regular instructional program.",
             "ethnicity": "Race/Ethnicity",
             "ethnicity-info": "Student race and ethnicity"
-          }
+          },
+          "fab": "Filter Results"
         },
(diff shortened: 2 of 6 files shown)


webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -227,6 +227,13 @@ export class AssessmentsComponent implements OnInit {
     this.onExportItemsByPointsEarned.emit(exportRequest);
   }
 
+  openAndScrollToAdvancedFilters() {
+    this.showAdvancedFilters = true;
+    setTimeout(() => {
+      document.getElementById('results-adv-filters').scrollIntoView();
+    }, 0);
+  }
```
esotrope: open to suggestions on a good way to not duplicate this code on both assessment result pages (student / group+school)

webapp/src/main/webapp/src/styles.less
```diff
@@ -114,6 +114,13 @@ footer {
   margin-bottom: @line-height-computed / 2;
 }
 
+.fab {
+  position: fixed;
+  right: @grid-gutter-width / 2;
+  bottom: @line-height-computed / 2;
+  z-index: 2000;
```
esotrope: @aaronsleeper had to add zindex to make the button appear over the assessment result tables.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
bgsavage: I think the angular way is to create a `@ViewChild` and then tag the element with that `#name` to get at the element.

bgsavage: re: dup code - could this button live in the adv filters component?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -234,7 +234,8 @@
             "limited-english-proficiency-info": "Students who use a language other than English as their primary language and who have been assessed as not possessing the English language skills necessary to succeed in the regular instructional program.",
             "ethnicity": "Race/Ethnicity",
             "ethnicity-info": "Student race and ethnicity"
-          }
+          },
+          "fab": "Filter Results"
```
bgsavage: nit: abbreviations - what's a fab? :)

webapp/src/main/webapp/src/styles.less
bgsavage: nit: may want to consider defining variables of known z-indexes so it is known why 2000 was chosen.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
aaronsleeper: the screenshots above look right to me. thanks :)

esotrope: on dup code - could do that but would have to pass in callbacks to advance filters like: ```(fabClick)=this.showAdvanceFilters=true``` not sure if that is better or not. Also, conceptually I feel like the button should live beside advanced filters and not in the component, but that's my feel.

webapp/src/main/webapp/src/styles.less
esotrope: maybe ```@z-index-greater-than-tables``` ?

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: You're fab Brandon : j

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
esotrope: Realized the show/hide will show/hide the fab if we put it in the advanced filter component. I think i'll leave it as is for now.
```html
<!-- Advanced Filters -->
   <div id="group-results-adv-filter-options" [hidden]="!showAdvancedFilters || !filterDisplayOptions.expanded" class="well gray-lighter">
     <adv-filters [filterBy]="clientFilterBy" [filterOptions]="filterOptions"></adv-filters>
   </div>
```

PR #423, opened 2017-08-30
Bug/dwr 728 leak student data

bgsavage:
Now throwing when no exams are found on the get student exams call.  This solution seems simpler than implementing all the sql permission checks in the studentRepository.findOneById if we are already making these same checks in the exam lookup anyway.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
@@ -37,6 +37,10 @@ class DefaultStudentExamService implements StudentExamService {
                 student.getId()
         );
 
+        if(exams.size() == 0) {
+            throw new NoSuchElementException(String.format("No exams for student with ID: %s", studentId));
+        }
+
         return new StudentExamHistory(student, exams);
     }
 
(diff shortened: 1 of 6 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/DefaultStudentExamService.java
```diff
@@ -37,6 +37,10 @@ public StudentExamHistory getExams(@NotNull final User user, final long studentI
                 student.getId()
         );
 
+        if(exams.size() == 0) {
```
esotrope: nit: use ```isEmpty()```

malaffoon: +1

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/ReportingExamItemService.java
```diff
@@ -28,10 +28,13 @@ public ReportingExamItemService(final ExamItemRepository examItemRepository,
 
     @Override
     public GroupExamItemScores getExamItems(@NotNull final User user, final long examId) {
-        final PermissionScope scope = user.getPermissionsById().get("GROUP_PII_READ").getScope();
-
         final List<ExamItem> examItems = examItemRepository.findAllByExamId(user.getPermissionsById(), user.getGroupsById().keySet(), examId);
         final List<AssessmentItem> assessmentItems = assessmentItemRepository.findAllForExam(examId);
+
+        if(examItems.isEmpty() || assessmentItems.isEmpty()) {
```
esotrope: you could optimize this by moving this block before the assessment item lookup. if there are no exam items we dont need to lookup the assessment items

bgsavage: Ideally these two calls could be made asynchronously as they don't depend on each other.  

But I can optimize as you suggest for now.

PR #425, opened 2017-08-31
version 1.0.0

malaffoon:
The CI process is such that develop should have the version number set to what we're working towards, in this case 1.0.0.

I've already bumped common and schema.

Note that i am not resetting the build number cause it really should've been 1.0.0 for a while.


common/src/test/resources/common-integration-test-data.sql
```diff
@@ -40,13 +40,13 @@ insert into exam (id, type_id, grade_id, grade_code, student_id, school_id, oppo
...
-  (-7, 2, -1, 'g1', -1, -10, 1, 0, 0, 0, 0, 2016, -4, 'v1', 2, 'Complete', 1, 'Valid', 'session7', 2100, 21, 2, '1997-01-05 00:00:00.000000', null, null, null, null, null, null, null, null, null, null, null, null, -1, '1997-07-18 20:14:34.000000', -1),
-  (-8, 2, -1, 'g1', -1, -10, 1, 0, 0, 0, 0, 2016, -4, 'v1', 2, 'Complete', 1, 'Valid', 'session8', 2100, 21, 2, '1997-01-06 00:00:00.000000', null, null, null, null, null, null, null, null, null, null, null, null, -1, '1997-07-18 20:14:34.000000', -1),
+  (-7, 2, -1, 'g1', -1, -10, 1, 0, 0, 0, 0, 1996, -4, 'v1', 2, 'Complete', 1, 'Valid', 'session7', 2100, 21, 2, '1997-01-05 00:00:00.000000', null, null, null, null, null, null, null, null, null, null, null, null, -1, '1997-07-18 20:14:34.000000', -1),
+  (-8, 2, -1, 'g1', -1, -10, 1, 0, 0, 0, 0, 1996, -4, 'v1', 2, 'Complete', 1, 'Valid', 'session8', 2100, 21, 2, '1997-01-06 00:00:00.000000', null, null, null, null, null, null, null, null, null, null, null, null, -1, '1997-07-18 20:14:34.000000', -1),
```
malaffoon: This must've been missed in an earlier change; we were getting away with it because the school-year table was duplicated in reporting.

webapp/src/main/webapp/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "0.0.0",
+  "version": "1.0.0",
```
chdietz: I think if you edit the version in package.json it will carry over to package-lock.json

rcordeau: It will. When you run an update, it should regenerate the package-lock.json file, but we would still have to check that in, which would bump the version number.

rcordeau:
Where is the package.json file?

PR #426, opened 2017-08-31
DWR-953 Performance: review reporting app for slow queries

agorina:
You can find more details here: https://jira.fairwaytech.com/browse/DWR-953

### DIFF
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -334,7 +334,7 @@ sql:
         where e.school_year=:school_year
         and e.school_id=:school_id
         and a.grade_id=:grade_id
-        and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
+        and (1=:statewide or s.district_id in (:district_ids) or s.id in (:school_ids))
         order by e.completed_at desc
         limit 1
 
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -80,7 +80,7 @@ app:
 
   datamart:
     datasource:
-      url: jdbc:mysql://localhost:3306/reporting?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8
+      url: jdbc:mysql://localhost:3306/reporting?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&useCompression=true
       username: root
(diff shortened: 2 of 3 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -334,7 +334,7 @@ sql:
         where e.school_year=:school_year
         and e.school_id=:school_id
         and a.grade_id=:grade_id
-        and (1=:statewide or s.district_id in (:district_ids) or e.school_id in (:school_ids))
```
agorina: According to the first comment on this page: https://dev.mysql.com/doc/refman/5.7/en/order-by-optimization.html, the optimizer does not like an indexed fields to be in the OR. 

Note: this by itself did not seem to change anything, but in combination with the new index https://github.com/SmarterApp/RDW_Schema/pull/147 improved the response time.

webapp/src/main/resources/application.yml
```diff
@@ -80,7 +80,7 @@ app:
 
   datamart:
     datasource:
-      url: jdbc:mysql://localhost:3306/reporting?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8
```
agorina: There are a lot of data being pulled. When I put a trace, this made a difference.

webapp/src/main/resources/application.yml
```diff
@@ -80,7 +80,7 @@ app:
...
+      url: jdbc:mysql://localhost:3306/reporting?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&useCompression=true
```
jtreuting: i know we have a story to separate the parameter part from the url part.  just made me wonder if this needs to be anywhere else or is it specific to here.  i'm guessing specific to here but figured i'd ask

agorina: Oh, good point. @malaffoon, I think this needs to be added to the configuration server?

And if we need it somewhere else: I did try it before in the migrate and did not see any effect. I highly doubt that ingest will benefit. Group Admin maybe a candidate, but not sure.

Just to note: it comes with the price of compressing/decompressing so I would not add it anywhere else without testing.

jtreuting: I assumed reporting was the best candidate for it as there is overhead, so I'm guessing it is correct only there, but just checking.

PR #427, opened 2017-08-31
modifying how we tag the docker file based on release or not

rcordeau:
The change is that, if doing a ./gradlew dockerPushImage -Pbuildnumber=RELEASE -Prelease, the -Prelease will force the docker image tag to be a versioned tag, which will not overwrite LATEST in dockerHub, but will create a version of the image.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_reporting/427)


rcordeau:
@esotrope versus the other changes that I introduce that are totally unreasonable?? ;-)

esotrope:
@rcordeau notice, i even said it just <i>sounds</i> reasonable. Meaning it might actually be unreasonable xD

PR #429, opened 2017-08-31
Bugfix/fix landing page

bgsavage:
It seems Spring doesn't like to keep the AuthPrincipal around when switching between http / https.  So instead of using the error page handling approach, now we're using a regex matcher to forward angular routes to the index (or landing page if not logged in)

https://stackoverflow.com/questions/16952718/springs-securitycontextholder-getcontext-getauthentication-returns-null-aft?rq=1


webapp/src/main/java/org/opentestsystem/rdw/reporting/web/NotFoundIndexTemplate.java
```diff
@@ -1,19 +0,0 @@
...
-
-/**
- * This customizer is responsible for re-directing any unbound request URI to the root
- * path.  This allows us to handle HTML5-style Single-Page-App URIs.
- */
-public class NotFoundIndexTemplate implements EmbeddedServletContainerCustomizer {
-
-    @Override
-    public void customize(final ConfigurableEmbeddedServletContainer container) {
-        container.addErrorPages(new ErrorPage(HttpStatus.NOT_FOUND, "/"));
```
bgsavage: The previous approach was to send any page not found in spring to the root "/".  However, now that we opened public access to "/", we need to check if a user is logged in order to determine which page to show them, the home page or the landing page.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -18,8 +18,14 @@ public String index(@AuthenticationPrincipal final User user) {
         return "forward:/index.html";
     }
 
+    @RequestMapping("/{path:^(?!api|assets|iris|.*\\..*).+}/**")
+    public String forward(@AuthenticationPrincipal final User user) {
+        return  index(user);
+    }
```
bgsavage: Since path forwarding seems to drop the @AuthPrincipal, now we are using regex matching to determine that the route is not to the api, assets, or iris.

malaffoon: Perhaps a code comment that says that last part, i.e. `match all routes that are not to the api, assets, or iris`.

wtritch: Interesting note:  In QA we can resolve the AuthenticationPrincipal here.  If we instead we internally forwarded to "/" then the "/" handler method above CAN NOT resolve the AuthenticationPrincipal in QA.
My *guess* is that it has to do with internal-forwarding losing the session due to HTTPS/HTTP protocol mismatching.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -518,7 +518,7 @@ protected void configure(HttpSecurity http) throws Exception {
                 .addFilterAfter(samlFilter(), BasicAuthenticationFilter.class);
         http
                 .authorizeRequests()
-                .antMatchers("/").permitAll()
+                .antMatchers("/", "/*.ico").permitAll()
```
esotrope: Does this fix the ico 404s i see now and then?

wtritch: It should, since we're not relying on a 404 returning the index.html SPA page

wtritch: The downside is that now we need to be aware of all requests that *shouldn't* go to the index page.

wtritch: I suppose we could flip it on it's head and be aware of all requests that *should* go to the index page, but that feels like a larger and more dynamic list.

bgsavage: Maybe in a future update where we prefix all angular routes with some `/app/**`

PR #430, opened 2017-08-31
Allow /status through security

wtritch:
Our StatusEndpoint isn't registered as a Bean in StatusConfiguration which means we need to manually account for it when poking security holes.

UPDATE: I instead modified RDW_Common to register the StatusEndpoint as a bean, so our standard passthrough of all actuator endpoints now picks it up.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
@@ -520,7 +520,8 @@ public class SamlWebSecurityConfiguration extends WebSecurityConfigurerAdapter {
                 .authorizeRequests()
                 .antMatchers("/", "/*.ico").permitAll()
                 .antMatchers("/assets/public/**").permitAll()
-                .antMatchers("/saml/**").permitAll();
+                .antMatchers("/saml/**").permitAll()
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -520,7 +520,8 @@ protected void configure(HttpSecurity http) throws Exception {
                 .authorizeRequests()
                 .antMatchers("/", "/*.ico").permitAll()
                 .antMatchers("/assets/public/**").permitAll()
-                .antMatchers("/saml/**").permitAll();
+                .antMatchers("/saml/**").permitAll()
+                .antMatchers("/status/**").permitAll();
```
wtritch: ...the more I think about this the more I think we should just register the endpoint as a bean

bgsavage: Maybe I'm misunderstanding, but should status be public?  Can't it contain environment sensitive details?

wtritch: It's only available on port 8008 which isn't available to the outside world.  I'm not sure why we're binding a security filter to our 8008 dispatcher, but that was before my time. =)

PR #431, opened 2017-09-01
Gradle script to gzip all bundles js files and spring configuration t…

bgsavage:
…o serve them when found.

### DIFF
--- a/webapp/build.gradle
+++ b/webapp/build.gradle
@@ -62,22 +62,34 @@ test.finalizedBy(jacocoTestReport)
 task buildFe(type: NpmTask) {
     inputs.files(fileTree("src/main/webapp/src"))
     inputs.file("src/main/webapp/package.json")
-    workingDir = file ("src/main/webapp")
+    workingDir = file("src/main/webapp")
     outputs.dir('build/resources/static')
     dependsOn npm_install
     args = ['run', 'build']
 }
 
+
 task testFe(type: NpmTask) {
     inputs.files(fileTree("src/main/webapp/src"))
     inputs.file("src/main/webapp/package.json")
-    workingDir = file ("src/main/webapp")
+    workingDir = file("src/main/webapp")
     outputs.dir('build/resources/static')
     dependsOn npm_install
     args = ['run', 'test-ci']
 }
 
+task gzipAllJs(dependsOn: buildFe) {
+    doLast {
+        FileTree tree = fileTree(dir: "${project.projectDir}/src/main/resources/static/", include: "**/*.bundle.js")
+        tree.each { File file ->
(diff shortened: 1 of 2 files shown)


webapp/build.gradle
```diff
@@ -62,22 +62,34 @@ test.finalizedBy(jacocoTestReport)
...
 }
 
+task gzipAllJs(dependsOn: buildFe) {
+    doLast {
+        FileTree tree = fileTree(dir: "${project.projectDir}/src/main/resources/static/", include: "**/*.bundle.js")
+        tree.each { File file ->
+            String fileName = file.getName()
+            println "$fileName => ${fileName}.gz"
+            ant.gzip(src: file, destfile: file.toString() + ".gz")
+        }
```
bgsavage: I spent a long time trying to get a 3rd party plugin (gradle-js) gzip task to work with multiple files only later to find out that there's an ant.gzip method I could use instead...

webapp/src/main/resources/application.yml
```diff
@@ -31,6 +31,11 @@ server:
     max-threads: 20
 
 spring:
+  resources:
+    chain:
+      enabled: true
+      gzipped: true
```
bgsavage: Configuration in which spring will look for a `.gz` variant of any resource requested and will serve that instead if it finds one.

webapp/build.gradle
rcordeau: So you want one .gz per bundle.js file?

webapp/src/main/resources/application.yml
jtreuting: does this flag also tell Spring to do the gzip on the fly?  i'm guessing no, which i think is good for the first step

webapp/build.gradle
jtreuting: yeah, if there is a somefile.js.gz file on the server, spring can return it for browsers that make a request for somefile.js and have the Accept-Encoding header of gzip.  brandon said it made the vendor bundle go down from 1.4 MG to 400K so a lot faster response to load the initial page for users

bgsavage: Yeah I'm thinking we still want the separate bundles so they can be cached individually and grabbed in parallel.

webapp/src/main/resources/application.yml
bgsavage: Nope, server.compression.enabled tells spring to do the gzip on the fly for a list of specified MIME types, which is not in this PR.

webapp/build.gradle
rcordeau: Sounds good to me. Faster responses to loading is great.

PR #432, opened 2017-09-01
Feature/loading spinner

esotrope:
Added loading spinner that gets triggered when a route change takes 2+ seconds.
This will cover the loading times of latest assessment, group, school & year filter changes on the group & school assessment pages.

screen shot 2017-09-01 at 2 58 32 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/app.component.ts
+++ b/webapp/src/main/webapp/src/app/app.component.ts
@@ -5,7 +5,6 @@ import { UserService } from "./user/user.service";
 import { Router, NavigationEnd } from "@angular/router";
 import { Location, PopStateEvent } from "@angular/common";
 import { NotificationService } from "./shared/notification/notification.service";
-import { environment } from "../environments/environment";
 import { isNullOrUndefined } from "util";
 
 @Component({
@@ -19,6 +18,7 @@ export class AppComponent {
   private _lastPoppedUrl: string;
   private _user;
 
+
   get user() {
     return this._user;
   }
@@ -26,8 +26,10 @@ export class AppComponent {
   /*
    Even though the angulartics2GoogleAnalytics variable is not explicitly used, without it analytics data is not sent to the service
    */
-  constructor(public translate: TranslateService, private _userService: UserService,
-              private router: Router, private location: Location,
+  constructor(public translate: TranslateService,
+              private userService: UserService,
+              private router: Router,
+              private location: Location,
               private notifService: NotificationService,
               private angulartics2GoogleAnalytics: Angulartics2GoogleAnalytics) {
 
@@ -41,8 +43,8 @@ export class AppComponent {
 
   ngOnInit() {
 
-    this._userService.getCurrentUser().subscribe(user => {
-      if(!isNullOrUndefined(user)) {
+    this.userService.getCurrentUser().subscribe(user => {
+      if (!isNullOrUndefined(user)) {
         this._user = {
           firstName: user.firstName,
           lastName: user.lastName
@@ -54,7 +56,7 @@ export class AppComponent {
           window[ 'ga' ]('create', user.configuration.analyticsTrackingId, 'auto');
         }
       } else {
-        this.router.navigate(["error"]);
+        this.router.navigate([ "error" ]);
       }
     });
 
@@ -84,6 +86,6 @@ export class AppComponent {
   }
 
   showInterpretiveGuideComingSoon() {
-    this.notifService.info({id: 'messages.interpretive-guide-coming-soon', args: { }, dismissOnTimeout: 5000 });
+    this.notifService.info({ id: 'messages.interpretive-guide-coming-soon', args: {}, dismissOnTimeout: 5000 });
   }
 }
--- a/webapp/src/main/webapp/src/app/shared/loader/loader.component.html
+++ b/webapp/src/main/webapp/src/app/shared/loader/loader.component.html
@@ -1,0 +1,9 @@
+<div class="delayed-fade" [class.delayed-fade-in]="loading">
+  <div class="loader-modal">
+    <div class="loader">
+      <div class="inner">
+        <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i>
+      </div>
+    </div>
+  </div>
+</div>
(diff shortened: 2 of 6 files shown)


webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -84,6 +86,6 @@ export class AppComponent {
   }
 
   showInterpretiveGuideComingSoon() {
-    this.notifService.info({id: 'messages.interpretive-guide-coming-soon', args: { }, dismissOnTimeout: 5000 });
+    this.notifService.info({ id: 'messages.interpretive-guide-coming-soon', args: {}, dismissOnTimeout: 5000 });
   }
 }
```
esotrope: this file just has code format/style changes

webapp/src/main/webapp/src/styles.less
```diff
@@ -23,8 +23,7 @@
 @import "primeng-overrides.less";
 
 // Local varaibles
```
malaffoon: nit: could fix typo while you're in here

webapp/src/main/webapp/src/app/shared/loader/loader.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="delayed-fade" [class.delayed-fade-in]="loading">
```
esotrope: was hoping to take advantage of bootstrap built in: ```modal-backdrop modal fade in``` css classes but so far have not had success with this.

webapp/src/main/webapp/src/styles.less
```diff
@@ -122,14 +121,50 @@ footer {
...
+  opacity: 0;
+  visibility: hidden;
+  transition: opacity 0s linear, visibility 0s linear;
+}
+
+.delayed-fade-in {
+  opacity: 1;
+  visibility: visible;
+  transition: opacity 0.15s linear, visibility 0.15s linear;
+  transition-delay: 2s;
```
malaffoon: Is 2s too long? I just timed myself and, i admit i'm a bit hyper, but 1.2s - 1.5s was when i started to wonder if i had missed the click.

webapp/src/main/webapp/src/app/shared/loader/loader.component.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor(private router: Router) {
+    router.events.subscribe(event => {
+      if (event instanceof NavigationStart) {
+        this.loading = true;
+      } else if (event instanceof NavigationEnd
+        || event instanceof NavigationCancel
+        || event instanceof NavigationError) {
+        this.loading = false;
+      }
```
esotrope: made the component bind itself to router events.
It will show when you change page and hide when a navigation ends.

webapp/src/main/webapp/src/styles.less
esotrope: Woah, chill out there Mark. Maybe lower the caffein intake a bit.
j/k yeah that makes sense to me too.

webapp/src/main/webapp/src/styles.less
```diff
@@ -122,14 +121,50 @@ footer {
...
+.loader-modal {
+  //.modal without display: none;
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: @zindex-modal;
+  overflow: hidden;
+  -webkit-overflow-scrolling: touch;
```
bgsavage: do we need this for other browsers?

webapp/src/main/webapp/src/styles.less
esotrope: definitely my type too x l

webapp/src/main/webapp/src/app/shared/loader/loader.component.html
bgsavage: I'm curious on initial login / initial loading if now we will see 2 spinners on top of one another.

webapp/src/main/webapp/src/styles.less
esotrope: these styles are copy-pasted from bs ```.modal``` so good q for bs folks.

webapp/src/main/webapp/src/styles.less
jtreuting: i thought @aaronsleeper wanted it to happen no matter what and the transition would allow it to not be jumpy when it was a quick page load.  if we are going to delay i think more like 0.5 is appropriate for the delay before showing

webapp/src/main/webapp/src/app/shared/loader/loader.component.html
esotrope: Now you will see them in sequence. so the "angularless" page will have the original spinner.
Then once angular loads the app.component.html will be loaded without the page body and you will see the new spinner modal atop that (if it takes > 2s) otherwise you wont see it

webapp/src/main/webapp/src/app/app.component.ts
aaronsleeper: the system should "respond" in some way between 0.15 and 0.5 seconds at a maximum

the spinner is an acceptable "response" so they know something is happening.

the spinner should fade in so if the interaction is only 0.4s it wasn't all the way faded in and it isn't too obtrusive. 

we shouldn't wait 2 seconds to put it up

webapp/src/main/webapp/src/styles.less
aaronsleeper: the system should respond in some manner within 0.15s
we should not wait 2s to put it up :)

aaronsleeper: that's correct @jtreuting

esotrope: I got 2s from somewhere in our conversations so i threw it in there, but I'll change it to .15s
Also the JIRA says fade the modal in over .4s - i changed this to .15s to match bootstrap modals so it would be consistent with other modals - thoughts?

webapp/src/main/webapp/src/styles.less
```diff
@@ -122,14 +121,50 @@ footer {
...
+  opacity: 0;
+  visibility: hidden;
+  transition: opacity 0s linear, visibility 0s linear;
+}
+
+.delayed-fade-in {
+  opacity: 1;
+  visibility: visible;
+  transition: opacity 0.4s linear, visibility 0.4s linear;
+  transition-delay: 0.15s;
```
esotrope: changed delay to .15s and duration to .4s

PR #433, opened 2017-09-01
Feature/user guide link

esotrope:


webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -38,25 +35,19 @@ export class AppComponent {
...
   }
 
   ngOnInit() {
 
     this.userService.getCurrentUser().subscribe(user => {
       if (!isNullOrUndefined(user)) {
-        this._user = {
-          firstName: user.firstName,
-          lastName: user.lastName
-        };
```
esotrope: is there a reason why we were only passing a 'view' of the user to the ui? I changed it to pass the whole user object.

jtreuting:
Isn't there more than 1 place for the user guide link.  Top nav and homepage right?

esotrope:
@jtreuting added href to link on homepage

webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -42,7 +42,7 @@ <h1 class="blue-dark h1 mb-md">{{'labels.homepage.title' | translate}}</h1>
       <div class="welcome-container well">
         <p class="lead">{{'labels.homepage.welcome.title' | translate}}</p>
         <p>{{'labels.homepage.welcome.sub-title' | translate}}</p>
-        <p><a href="javascript:void(0)" (click)="showComingSoon('User Guide')" class="btn btn-primary">{{'labels.navigation.user-guide' | translate}}</a></p>
+        <p><a href="{{user.configuration.userGuideUrl}}" target="_blank" class="btn btn-primary">{{'labels.navigation.user-guide' | translate}}</a></p>
```
jtreuting: i think we need to remove the showComingSoon() method from `home.component.ts`

webapp/src/main/webapp/src/app/app.component.ts
bgsavage: Just all that was needed at the time.

webapp/src/main/webapp/src/app/home/home.component.html
esotrope: removed with latest commit

webapp/src/main/webapp/src/app/app.component.ts
esotrope: 👍

PR #435, opened 2017-09-04
Now caching all static assets except for index.html and landing.html

bgsavage:
Disables caching for landing.html and index.html, but sets caching for up to a month for all other static assets.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
@@ -535,6 +535,7 @@ public class SamlWebSecurityConfiguration extends WebSecurityConfigurerAdapter {
                 .logoutSuccessUrl("/");
 
         http.headers().frameOptions().sameOrigin();
+        http.headers().cacheControl().disable();
     }
 
     /**
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
@@ -12,7 +12,7 @@ public class ViewController implements ErrorController {
     @RequestMapping(value = {"/"})
     public String index(@AuthenticationPrincipal final User user) {
         if (user == null) {
-            return "forward:/assets/public/landing.html";
+            return "forward:/landing.html";
         }
 
         return "forward:/index.html";
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
@@ -26,7 +26,6 @@ public class WebMvcConfiguration extends WebMvcConfigurerAdapter implements Embe
                 .setCachePeriod(SecondsInMonth)
                 .resourceChain(true).addResolver(new GzipResourceResolver());
 
-
         registry.addResourceHandler("/assets/public/**")
                 .addResourceLocations("classpath:static/assets/public/")
                 .setCachePeriod(SecondsInMonth);
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
@@ -21,6 +21,10 @@ public class WebMvcConfiguration extends WebMvcConfigurerAdapter implements Embe
                 .addResourceLocations("classpath:static/assets/public/landing.html")
                 .setCachePeriod(0);
 
+        registry.addResourceHandler("/assets/i18n/**")
+                .addResourceLocations("classpath:static/assets/i18n/")
+                .setCachePeriod(0);
+
         registry.addResourceHandler("/**")
                 .addResourceLocations("classpath:static/")
                 .setCachePeriod(SecondsInMonth)
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
@@ -5,10 +5,11 @@ import org.springframework.boot.context.embedded.EmbeddedServletContainerCustomi
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
+import org.springframework.web.servlet.resource.GzipResourceResolver;
 
 @Configuration
 public class WebMvcConfiguration extends WebMvcConfigurerAdapter implements EmbeddedServletContainerCustomizer {
-    private final static Integer SecondsInMonth = 3600 * 24 * 30;
+    private final static int SecondsInMonth = 3600 * 24 * 30;
 
     @Override
     public void addResourceHandlers(ResourceHandlerRegistry registry) {
@@ -22,7 +23,9 @@ public class WebMvcConfiguration extends WebMvcConfigurerAdapter implements Embe
 
         registry.addResourceHandler("/**")
                 .addResourceLocations("classpath:static/")
-                .setCachePeriod(SecondsInMonth);
+                .setCachePeriod(SecondsInMonth)
+                .resourceChain(true).addResolver(new GzipResourceResolver());
+
 
         registry.addResourceHandler("/assets/public/**")
                 .addResourceLocations("classpath:static/assets/public/")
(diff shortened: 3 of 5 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -535,6 +535,7 @@ protected void configure(HttpSecurity http) throws Exception {
                 .logoutSuccessUrl("/");
 
         http.headers().frameOptions().sameOrigin();
+        http.headers().cacheControl().disable();
```
bgsavage: By default spring security adds cache-control: no-store headers to all requests as to prevent a malicious user from viewing a cached page once a user has already logged out.  I tested that we still get the login page when trying to do this with this impl, but looking to spring security experts ( @malaffoon @wtritch ) to poke a hole in this if there is one.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Configuration
+public class WebMvcConfiguration extends WebMvcConfigurerAdapter implements EmbeddedServletContainerCustomizer {
+    private final static Integer SecondsInMonth = 3600 * 24 * 30;
+
+    @Override
+    public void addResourceHandlers(ResourceHandlerRegistry registry) {
+        registry.addResourceHandler("/index.html")
+                .addResourceLocations("classpath:static/index.html")
+                .setCachePeriod(0);
```
bgsavage: Always get latest of index.html as it contains the manifest of all our cacheable assets.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+
+        registry.addResourceHandler("/landing.html")
+                .addResourceLocations("classpath:static/assets/public/landing.html")
+                .setCachePeriod(0);
```
bgsavage: Also not caching landing.html so that we don't ever cache `/` since we always need that check to see if a user is logged in to either serve `landing.html` or `index.html` when navigating to `/`.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+
+        registry.addResourceHandler("/**")
+                .addResourceLocations("classpath:static/")
+                .setCachePeriod(SecondsInMonth);
+
+        registry.addResourceHandler("/assets/public/**")
+                .addResourceLocations("classpath:static/assets/public/")
+                .setCachePeriod(SecondsInMonth);
```
bgsavage: For all others, setting the cache period to a month.

webapp/src/main/webapp/src/index.html
```diff
@@ -4,6 +4,9 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate"/>
+    <meta http-equiv="Pragma" content="no-cache"/>
+    <meta http-equiv="Expires" content="0"/>
```
bgsavage: I'm not convinced that adding these headers actually do anything in our spring environment, but they also don't hurt anything.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -12,7 +12,7 @@
     @RequestMapping(value = {"/"})
     public String index(@AuthenticationPrincipal final User user) {
         if (user == null) {
-            return "forward:/assets/public/landing.html";
+            return "forward:/landing.html";
```
malaffoon: I don't quite understand how this doesn't make the landing page fall under an authenticated location. `/assets/public/**` is explicitly allowed in the security configuration but `/**` requires auth.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.springframework.boot.context.embedded.ConfigurableEmbeddedServletContainer;
+import org.springframework.boot.context.embedded.EmbeddedServletContainerCustomizer;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
+
+@Configuration
+public class WebMvcConfiguration extends WebMvcConfigurerAdapter implements EmbeddedServletContainerCustomizer {
+    private final static Integer SecondsInMonth = 3600 * 24 * 30;
```
malaffoon: nit: could be `int`

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
bgsavage: Is it because this is an internal `forward` and not a client `redirect` and we've already gone past secuity at this point?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+                .addResourceLocations("classpath:static/index.html")
+                .setCachePeriod(0);
+
+        registry.addResourceHandler("/landing.html")
+                .addResourceLocations("classpath:static/assets/public/landing.html")
+                .setCachePeriod(0);
+
+        registry.addResourceHandler("/**")
+                .addResourceLocations("classpath:static/")
+                .setCachePeriod(SecondsInMonth);
```
malaffoon: Do our static resources have generated names that change when we make changes? If not, this means that a modified resource won't be reloaded for up to a month after it is changed? If so, why have a an expiration at all?

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
bgsavage: Ah, I chose `Integer` because `setCachePeriod` takes an `Integer`.  But `int` also works.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/WebMvcConfiguration.java
bgsavage: Our js and css files have generated names based on the content, however our images do not.

Maybe I should treat images as separate type of resource that have a different expiration?

As far as the cache period for js/css being one month, I feel I should explicitly set it as something right?  Or if I choose not to set, I saw chrome was still caching by default anyways.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
wtritch: "/" is also explicitly allowed by security.  I don't believe the internal forwarding goes through security again.  This is fielding a browser request for "/", the browser never actually makes another request for "/landing.html" or "index.html"

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
wtritch: I'd like to see what the actual cache-control headers sent to the browser look like for requests to:
"/" (should never be cached)
"/api/something" (should never be cached)
"/something.js" (should be cached)
"/public/assets/something.css" (should be cached)

Am I missing any other paths/desired caching behavior?

webapp/src/main/webapp/src/index.html
wtritch: I'd ditch these in favor of the happy-path for cache control headers (using spring above)  We'll have forgotten that we've defined caching in two places within a month =P

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
bgsavage: I agree with those.

So for js|css|png|etc..
`Cache-Control:max-age=2592000`

For index.html / landing.html:
`Cache-Control:no-store`

And for api endpoints there is no cache-control header.

bgsavage: Oh but you know what, en.json is also being cached and it should not.  I'll call that one out specifically too.

wtritch: I'd also prefer if our /api/** calls had a no-store cache-control header.  Otherwise we're leaving it up to the browser.  And browsers like to cache =P

bgsavage: Sounds good, I can add that

PR #437, opened 2017-09-04
fix claim target component to only make one http request to load the …

esotrope:
…target

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/claim-target.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/claim-target.component.html
@@ -3,9 +3,12 @@
    popoverTitle="{{'definition.claim.' + item.claim + '.name' | translate}}"
    outsideClick="true">{{'definition.claim.' + item.claim + '.name' | translate}}</a>
 /
-<a href="javascript:void(0)" #me
+<a href="javascript:void(0)"
+   (click)="loadTargetDescription()"
    [popover]="targetPopover"
    popoverTitle="{{'labels.groups.results.assessment.items.target' | translate:item}}"
    outsideClick="true">{{'labels.groups.results.assessment.items.target' | translate:item}}</a>
 
-<ng-template #targetPopover>{{getTargetDescription() | async}}</ng-template>
+<ng-template #targetPopover>
+  <span *ngIf="target">{{target}}</span>
+</ng-template>
--- a/webapp/src/main/webapp/src/app/assessments/results/claim-target.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/claim-target.component.ts
@@ -1,7 +1,7 @@
 import { Component, Input } from "@angular/core";
-import { Observable } from "rxjs";
 import { ItemInfoService } from "../items/item-info/item-info.service";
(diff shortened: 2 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/results/claim-target.component.html
```diff
@@ -3,9 +3,12 @@
...
    outsideClick="true">{{'definition.claim.' + item.claim + '.name' | translate}}</a>
 /
-<a href="javascript:void(0)" #me
+<a href="javascript:void(0)"
+   (click)="loadTargetDescription()"
    [popover]="targetPopover"
    popoverTitle="{{'labels.groups.results.assessment.items.target' | translate:item}}"
    outsideClick="true">{{'labels.groups.results.assessment.items.target' | translate:item}}</a>
 
-<ng-template #targetPopover>{{getTargetDescription() | async}}</ng-template>
```
esotrope: not sure why using ```| async``` resulted in multiple backend calls but it did.

bgsavage: We probably shouldn't bind directly to an service / api call from the view anyway, it could get checked for changes on every digest cycle.

webapp/src/main/webapp/src/app/assessments/results/claim-target.component.html
```diff
@@ -3,9 +3,12 @@
    popoverTitle="{{'definition.claim.' + item.claim + '.name' | translate}}"
    outsideClick="true">{{'definition.claim.' + item.claim + '.name' | translate}}</a>
 /
-<a href="javascript:void(0)" #me
+<a href="javascript:void(0)"
+   (click)="loadTargetDescription()"
```
bgsavage: So do we reload the same target description again and again on every click?

webapp/src/main/webapp/src/app/assessments/results/claim-target.component.ts
```diff
@@ -14,13 +14,20 @@ import { AssessmentItem } from "../model/assessment-item.model";
...
+
+  target: string;
 
   constructor(private service: ItemInfoService) {
   }
 
-  public getTargetDescription(): Observable<string> {
-    return this.service.getTargetDescription(this.item.targetId);
+  loadTargetDescription(): void {
+    if (isNullOrUndefined(this.target)) {
```
bgsavage: Oh nevermind, you're checking for undefined here. :)

webapp/src/main/webapp/src/app/assessments/results/claim-target.component.html
esotrope: I think where i strayed from the docs was using a method here instead of binding to the returned observable of the method called at init time. But if i did that, it wouldn't lazy load, which would defeat the purpose : /

webapp/src/main/webapp/src/app/assessments/results/claim-target.component.ts
esotrope: Also i think the getTarget() service call goes through the caching dataservice you set up anyway 👍

bgsavage:
looks good

PR #439, opened 2017-09-05
Now enabling the cache control headers again provided by security.  A…

bgsavage:
…s it turns out, the WebMvcConfiguration overrides these so static assets are still cached while api end points are not.


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -535,7 +535,6 @@ protected void configure(HttpSecurity http) throws Exception {
                 .logoutSuccessUrl("/");
 
         http.headers().frameOptions().sameOrigin();
-        http.headers().cacheControl().disable();
```
bgsavage: So I actually don't need to disable cache control behavior in security as I'm overriding it in the WebMvcConfiguration, so for our api routes, this gives us back our 
`Cache-Control:no-cache, no-store, max-age=0, must-revalidate` provided by spring security's default behavior.

PR #440, opened 2017-09-05
Now no longer storing the location.href if it is the session-expired …

bgsavage:
…link to prevent sending back to the session-expired page after login.  Also sending to /home now instead of / so the user doesn't have to see the landing page again if they were coming from the home route.

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/authentication/authentication.service.ts
+++ b/webapp/src/main/webapp/src/app/shared/authentication/authentication.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from "@angular/core";
 import { Router } from "@angular/router";
 import { StorageService, StorageType } from "../storage.service";
+import { WindowRefService } from "../window-ref.service";
 
 /**
  * This service is responsible for handling authentication errors.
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/shared/authentication/authentication.service.ts
```diff
@@ -18,8 +18,21 @@ export class AuthenticationService {
...
    */
   handleAuthenticationFailure(): void {
-    this.storageService.getStorage(StorageType.Session)
-      .setItem(this.locationKey,  window.location.href);
+    let urlToStore = window.location.href;
+
+    // If the url was the root, navigate to home instead so the user doesn't get the
+    // landing page.
+    if(window.location.pathname === "/") {
+      urlToStore += "home";
```
esotrope: so what's the diff between / and /home?
Is it that /home triggers login redirect and / goes to /public/landing.html?
Do we have anything to redirect you /home if you are authenticated already and go to /public/landing.html?

bgsavage: Yep `/home` will trigger a login redirect as it's not public like `/` is.  `/` will show the landing page if you're not logged in instead of the sso redirect.

I think if you want to explicitly go to `assets/public/landing.html` then I won't stop you, though there's no links in the app that will take you directly there, it's always through conditions set by hitting `/`.

PR #442, opened 2017-09-06
Hotfix/update links and text

malaffoon:
Okay, let's try this again ... with more cowbell.

@bgsavage and @jtreuting please check that all the changes shown here are intended for the hotfix. Thanks.

### DIFF
--- a/webapp/src/main/webapp/src/app/app.component.html
+++ b/webapp/src/main/webapp/src/app/app.component.html
@@ -18,7 +18,7 @@
             <a class="navbar-brand hidden-xs" href="/" title="{{'application.home' | translate}}"><img src="assets/image/SmarterBalanced-Logo.png" alt="{{'messages.organization' | translate}}"></a>
             <span class="application-title">{{'application.title' | translate}}</span>
           </div>
-          <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
+          <div *ngIf="user" class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
             <ul class="nav navbar-nav navbar-right">
               <li class="navbar-link"><a href="{{user.configuration.userGuideUrl}}" target="_blank">{{'labels.navigation.user-guide' | translate}}</a></li>
               <li class="navbar-link"><a href="javascript:void(0)" (click)="showInterpretiveGuideComingSoon()" target="_blank">{{'labels.navigation.interpretive-guide' | translate}}</a></li>
--- a/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
@@ -12,6 +12,11 @@
       {{ 'labels.assessments.items.tabs.item-viewer-see-rubric' | translate }}
     </div>
   </div>
-  <p>{{ 'labels.assessments.items.tabs.item-viewer-intro-text' | translate }}</p>
+  <p>
+    {{ 'labels.assessments.items.tabs.item-viewer-intro-text' | translate }}
+    <span *ngIf="!showResponse">
+      {{ 'labels.assessments.items.tabs.item-viewer-response-not-included' | translate }}
+    </span>
+  </p>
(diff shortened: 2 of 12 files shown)


webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -18,7 +18,7 @@
             <a class="navbar-brand hidden-xs" href="/" title="{{'application.home' | translate}}"><img src="assets/image/SmarterBalanced-Logo.png" alt="{{'messages.organization' | translate}}"></a>
             <span class="application-title">{{'application.title' | translate}}</span>
           </div>
-          <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
+          <div *ngIf="user" class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
```
bgsavage: unrelated: This has already been fixed in develop, but without this, we get a flurry of exceptions in the console.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
```diff
@@ -12,6 +12,11 @@
       {{ 'labels.assessments.items.tabs.item-viewer-see-rubric' | translate }}
     </div>
   </div>
-  <p>{{ 'labels.assessments.items.tabs.item-viewer-intro-text' | translate }}</p>
+  <p>
+    {{ 'labels.assessments.items.tabs.item-viewer-intro-text' | translate }}
+    <span *ngIf="!showResponse">
+      {{ 'labels.assessments.items.tabs.item-viewer-response-not-included' | translate }}
+    </span>
```
bgsavage: If by context (i.e. from Items By Points earned table) we don't want to show responses, then show the message: This does not include student responses.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.ts
```diff
@@ -32,6 +32,9 @@ export class ItemViewerComponent implements OnInit {
   @Input()
   public position: number;
 
+  @Input()
+  public showResponse: boolean = true;
```
bgsavage: Default to true as it's the most common case.  There's only one place where we want it to be false.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -231,6 +231,7 @@ <h3 class="gray-darkest mb-sm">
           </p-column>
           <ng-template let-item pTemplate="rowexpansion">
             <item-tab [item]="item"
+                      [showResponse]="false"
```
bgsavage: Don't show the response in the Items By Points earned table.

webapp/src/main/webapp/src/assets/public/landing.html
```diff
@@ -41,11 +41,9 @@
...
     <ul>
       <li class="">
-        <a href="https://portal.smarterbalanced.org/library/en/reporting-system-interpretive-guide.pdf" target="_blank">
-          <div class="annotatedReports asset"></div>
-          <h2>Interpretive Guide</h2>
-          <p>Learn how to interpret scores in this reporting tool</p>
-        </a>
+|       <div class="annotatedReports asset"></div>
+        <h2>Interpretive Guide</h2>
+        <p>Learn how to interpret scores in this reporting tool <b>- Available September 15th</b></p>
```
bgsavage: If we had more time to work on this page, I think it'd benefit us to fold this into our angular app so we can pipe the text through translate and links through configurations so we can alter it without requiring a release.

But then again, who knows if this will ever change again after Sept 15?

webapp/src/main/webapp/src/app/app.component.html
jtreuting: @bgsavage Can you confirm that QA tested and signed off of that JIRA?

bgsavage: I don't think a Jira was created for this.

webapp/src/main/webapp/src/assets/public/landing.html
esotrope: The sarcasm is real x l

PR #443, opened 2017-09-08
Added a public error and not-permitted.html pages to reporting webapp…

bgsavage:
… to prevent infinite looping of sso login when a user doesn't have credentials to this tenant.  Now writing the warn to the logger explicitly and throwing a UserNameNotFoundException rather than the RuntimeException we were throwing previously which has different meaning.

This fixes DWR 913 (stack trace logged with bad login) and DWR 1002 (inf. sso loop)

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
@@ -6,6 +6,8 @@ import org.opentestsystem.rdw.reporting.common.security.Permission;
 import org.opentestsystem.rdw.reporting.security.User;
 import org.opentestsystem.rdw.reporting.security.auth.AuthorizationService;
 import org.opentestsystem.rdw.security.Grant;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
@@ -19,6 +21,7 @@ import java.util.Set;
 
 @Service
 public class SamlUserDetailsService implements SAMLUserDetailsService {
+    private static final Logger logger = LoggerFactory.getLogger(SamlUserDetailsService.class);
 
     private static final String REDACTED = "[REDACTED]";
     private AuthorizationService authorizationService;
@@ -41,8 +44,10 @@ public class SamlUserDetailsService implements SAMLUserDetailsService {
 
 		// Secures the application by throwing exception if the user has no authority to access data for this state
         if (grants.isEmpty()) {
-            throw new RuntimeException(String.format(
-                    "Grants do not permit access to tenant for user \"%s\"", username));
+            String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
+            logger.warn(message);
+
+            throw new UsernameNotFoundException(message);
         }
 
         final Set<Permission> permissions = authorizationService.getPermissions(grants);
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
@@ -67,6 +67,7 @@ import org.springframework.security.saml.websso.WebSSOProfileECPImpl;
 import org.springframework.security.saml.websso.WebSSOProfileImpl;
 import org.springframework.security.saml.websso.WebSSOProfileOptions;
 import org.springframework.security.web.AuthenticationEntryPoint;
+import org.springframework.security.web.DefaultRedirectStrategy;
 import org.springframework.security.web.DefaultSecurityFilterChain;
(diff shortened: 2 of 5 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -520,6 +521,7 @@ protected void configure(HttpSecurity http) throws Exception {
                 .authorizeRequests()
                 .antMatchers("/", "/*.ico").permitAll()
                 .antMatchers("/assets/public/**").permitAll()
+                .antMatchers("/error", "/error/**").permitAll()
```
bgsavage: Permitting error path.  This was the cause of the infinite loop because the exception was redirecting to /error, which requires sso, so redirects back to sso, which threw exception that redirects to /error ..... etc. etc

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -19,11 +20,26 @@ public String index(@AuthenticationPrincipal final User user) {
...
+        return "redirect:/assets/public/not-permitted.html";
+    }
+
+    @RequestMapping(value="/error")
+    public String error(@AuthenticationPrincipal final User user) {
+        if (user == null) {
+            return "forward:/assets/public/error.html";
+        }
+
+        return "forward:/index.html";
```
bgsavage: This will go to the angular /error route in this case.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -41,8 +44,10 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
 
 		// Secures the application by throwing exception if the user has no authority to access data for this state
         if (grants.isEmpty()) {
-            throw new RuntimeException(String.format(
-                    "Grants do not permit access to tenant for user \"%s\"", username));
+            String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
```
malaffoon: nit: final

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -19,11 +20,26 @@ public String index(@AuthenticationPrincipal final User user) {
     }
 
     // matches all routes that are not /api/**, /assets/**, /iris/**, or /*.*
```
malaffoon: nit: perhaps change comment to be more useful, like `matches all routes that aren't expected, like /api/**`

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
malaffoon: perhaps a comment in the code; i see myself trying to equate `/error` with `/index.html` and coming up blank

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
esotrope: I think we run similar logic inside ```getGrants()``` should we remove it from there?
Also the logic inside that method throws a ```RuntimeException``` instead of the ```UsernameNotFoundException``` so we might get different reaction from spring security if it has a catch for that typed exception.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -19,11 +20,26 @@ public String index(@AuthenticationPrincipal final User user) {
...
-    @RequestMapping("/{path:^(?!api|assets|iris|.*\\..*).+}/**")
+    @RequestMapping("/{path:^(?!api|assets|iris|error|.*\\..*).+}/**")
     public String forward(@AuthenticationPrincipal final User user) {
         return  index(user);
     }
 
+    @PostMapping(value="/error/notPermitted")
```
esotrope: ```/accessDenied``` maybe?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
bgsavage: Yeah we do, I notice I wasn't getting forwarded to the authenticationFailureUrl when a RuntimeException is thrown.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
esotrope: Why do we have to forward to public non-translated resources?

bgsavage: The translated resources are protected by security at the moment.  Wasn't sure if wanted to make them publicly available at this point.

esotrope: I think they should be public - there is nothing worth protecting in there.

PR #444, opened 2017-09-08
adds 401 handling http service

esotrope:
Adds global 401 handle to redirect user to /session-expired when the session has expired and an async http request is made.

This solution has one quirk which is that angular logs an error if you are navigating to a new page when your session is expired. This error however does not affect the behavior of being navigated to the session-expired page.

screen shot 2017-09-08 at 11 17 24 am


webapp/src/main/webapp/src/app/shared/authentication/authenticated-http.service.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  request(url: string | Request, options?: RequestOptionsArgs): Observable<Response> {
+    return super.request(url, options)
+      .catch((error: Response) => {
+        if (error.status === 401) {
+          this.service.handleAuthenticationFailure();
+          return Observable.of();
+        }
+        return Observable.throw(error);
+      });
```
esotrope: this method is the meat of the solution.
Extend angular's Http service and catch if the response is a 401. If it is, redirect to /session-expired.

webapp/src/main/webapp/src/app/shared/response-utils.ts
```diff
@@ -31,6 +31,9 @@ export class ResponseUtils {
    * @param response the HTTP response
    */
   static throwError(response: Response): Observable<any> {
+    if (response.status === 401) {
+      return Observable.empty();
+    }
```
esotrope: There is probably a more "ReactiveX" way of handling this but basically this prevents 401s from bubbling up the ```catch``` error listener chain.
Specifically, this prevents page-specific error handlers from handling the 401 error after the Http service has handled it.

webapp/src/main/webapp/src/app/shared/authentication/authenticated-http.service.ts
bgsavage: Weren't we already doing this?

esotrope: From what I gathered from @wtritch we had something roughly like this at first and then there was some issue with using it with Resolves so we moved to the sb-error-handler.
I'm not sure what the original issue was with this approach but it seems to work fine with this impl.

bgsavage: I guess I'm just curious what prompted the change from the error handler direction to this approach?

esotrope: Ah. So the purpose of the error handler is to catch and handle uncaught/unhandled errors and do something with them. This was problematic for the reports page which catches and handles errors but would also like to redirect if the error is a 401. The alternative would have been identifying that the error was a 401 and letting it bubble up the error chain.
This way we catch 401s immediately and handle them before allowing page-level components to handle them.

bgsavage: Ah makes sense.

LGTM 👍

PR #445, opened 2017-09-08
Improved error logging / handling for item scoring guide.  The getXml…

bgsavage:
…Document method is now synchronized as the parse method is not thread-safe and will throw an exception if two different threads attempt to call parse at the same time.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.net.URLConnection;
 import java.util.Base64;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -82,12 +83,18 @@ class DefaultExamItemService implements ExamItemService {
     public ExamItemScoringGuide getScoringGuide(final String id) {
         try {
             return getExamItemSolution(getXmlDocument(id), id);
-        } catch (final IOException | SAXException | XPathExpressionException exception) {
+        }
+        catch(SAXException exception) {
+            logger.error(String.format("SAXException while trying to read the item xml: %s", id), exception);
             throw new RuntimeException(String.format("Error reading the xml for item: %s", id), exception);
+        }
+        catch (final IOException | XPathExpressionException exception) {
+            throw new NoSuchElementException(String.format("Rubric/Exemplar data is not available for item: %s", id));
         }
     }
 
-    private Document getXmlDocument(final String id) throws IOException, SAXException {
+    // DocumentBuilder.parse is not thread-safe: https://stackoverflow.com/questions/12455602/is-documentbuilder-thread-safe
+    private synchronized Document getXmlDocument(final String id) throws IOException, SAXException {
         return documentBuilder.parse(artifactRepository.getItemXml(id));
     }
 


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
```diff
@@ -82,12 +83,18 @@
     public ExamItemScoringGuide getScoringGuide(final String id) {
         try {
             return getExamItemSolution(getXmlDocument(id), id);
-        } catch (final IOException | SAXException | XPathExpressionException exception) {
+        }
+        catch(SAXException exception) {
+            logger.error(String.format("SAXException while trying to read the item xml: %s", id), exception);
             throw new RuntimeException(String.format("Error reading the xml for item: %s", id), exception);
```
bgsavage: I don't think we should ever expect a SAXException in normal business flow, so I want to explicitly log the error for it.  This should now be more apparent in the logs should it continue to occur.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
```diff
@@ -82,12 +83,18 @@
...
         }
+        catch (final IOException | XPathExpressionException exception) {
+            throw new NoSuchElementException(String.format("Rubric/Exemplar data is not available for item: %s", id));
```
bgsavage: An IOxception or XPathExpresionException would happen if the xml doens't exist for this item or the rubric / exemplar does not exist in the item xml.  I don't think we should ever get an XPathExpressionException here at this level as protect at lower levels and set defaults when they do occur.  

Regardless, they should be considered normal flow and should return 404 instead of 500.  The UI can handle this just fine without any updates.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
```diff
@@ -82,12 +83,18 @@
...
+        }
     }
 
-    private Document getXmlDocument(final String id) throws IOException, SAXException {
+    // DocumentBuilder.parse is not thread-safe: https://stackoverflow.com/questions/12455602/is-documentbuilder-thread-safe
+    private synchronized Document getXmlDocument(final String id) throws IOException, SAXException {
         return documentBuilder.parse(artifactRepository.getItemXml(id));
```
bgsavage: This parse method is the offender and is not thread safe, so unfortunately, we have to block multiple threads from trying to use parse at the same time with the `synchronized` keyword.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
```diff
@@ -82,12 +83,18 @@
...
             throw new RuntimeException(String.format("Error reading the xml for item: %s", id), exception);
         }
+        catch (final IOException | XPathExpressionException exception) {
+            throw new NoSuchElementException(String.format("Rubric/Exemplar data is not available for item: %s", id));
+        }
     }
 
-    private Document getXmlDocument(final String id) throws IOException, SAXException {
+    // DocumentBuilder.parse is not thread-safe: https://stackoverflow.com/questions/12455602/is-documentbuilder-thread-safe
+    private synchronized Document getXmlDocument(final String id) throws IOException, SAXException {
```
esotrope: Could this alternatively be resolved by making a new DocumentBuilder per method invocation?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
esotrope: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
bgsavage: Yep we could.  Not sure which would perform better though.  One suggestion was to implement a DocumentBuilder pool, which I thought would be overkill for what we're trying to do.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/item/DefaultExamItemService.java
malaffoon: I'm not sure i agree that a well-behaved server should *ever* return a 500. Clearly this is expected because there is code for it. But it is merged and i don't think i can clearly describe a better solution.

malaffoon: As an ops person i don't think this should be loglevel ERROR. ERROR's in the log trigger pager calls. This is not deserving of a pager call. I'd suggest WARN.

PR #447, opened 2017-09-08
Welcome username is now same font size as other links in the nav.

bgsavage:
Before:
image

Now:
image


webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -23,12 +23,12 @@
               <li class="navbar-link"><a href="{{user.configuration.userGuideUrl}}" target="_blank">{{'labels.navigation.user-guide' | translate}}</a></li>
               <li class="navbar-link"><a href="javascript:void(0)" (click)="showInterpretiveGuideComingSoon()" target="_blank">{{'labels.navigation.interpretive-guide' | translate}}</a></li>
               <li class="navbar-link" *sbAuthorize="['GROUP_PII_READ']"><a routerLink="/reports">{{'labels.reports.link' | translate}}</a></li>
-              <li class="navbar-link mt-sm">
+              <li class="navbar-link">
```
bgsavage: This margin was there before to line it up correctly.  No longer needed.

webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -23,12 +23,12 @@
...
                 <div class="btn-group" dropdown>
-                  <button type="button" dropdownToggle class="btn btn-link btn-sm dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                  <button type="button" dropdownToggle class="btn btn-link dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
bgsavage: Removed btn-sm which was making the button smaller.

webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -23,12 +23,12 @@
...
                     <i class="fa fa-user-circle"></i> <span *ngIf="user">{{'welcome' | translate:user }}</span> <span class="caret"></span>
                   </button>
-                  <ul *dropdownMenu class="dropdown-menu" role="menu">
+                  <ul *dropdownMenu class="dropdown-menu mt-md" role="menu">
```
bgsavage: This margin is now needed so that the dropdown shows up below the button instead of on top.

webapp/src/main/webapp/src/styles.less
```diff
@@ -39,6 +39,21 @@
...
+
+.navbar-nav > li > div {
+  line-height: 23px;
+}
+
+.nav > li > div {
+  position: relative;
+  display: block;
+  padding: 10px 15px;
+}
```
bgsavage: These styles are needed to imitate the `.nav > li > a` styles provided by bootstrap.  We need the `div` version of these because the list item here is a dropdown, not a link.

PR #449, opened 2017-09-12
merges dropdown menu and expander icons with first column of tables

esotrope:
Merged menu icons with first column of tables.
Took some artistic liberty with this and added ```btn-block``` to ```btn``` and ```mr-xs``` to ```<i>```

image

screen shot 2017-09-12 at 10 35 10 am
screen shot 2017-09-12 at 10 35 20 am
screen shot 2017-09-12 at 10 35 40 am

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
@@ -25,6 +25,14 @@ export class PopupMenuComponent {
   @Input()
   public actions: PopupMenuAction[];
 
+  private _open: boolean;
+
+  private removeListener: () => void;
+
+  constructor(private renderer: Renderer2,
+              private domElement: ElementRef) {
+  }
+
   public get open(): boolean {
     return this._open;
   }
@@ -36,21 +44,15 @@ export class PopupMenuComponent {
    * @param value True to open the menu, false to close
    */
   public set open(value: boolean) {
-    if (this._open === value) return;
-
+    if (this._open === value) {
+      return;
+    }
     this._open = value;
     if (value) {
       this.removeListener = this.renderer.listen(document, 'click', this.onClick.bind(this));
     } else if (!isNullOrUndefined(this.removeListener)) {
       this.removeListener();
     }
-  }
-
-  private _open: boolean;
-  private removeListener: () => void;
-
-  constructor(private renderer: Renderer2,
-              private domElement: ElementRef) {
   }
 
   /**
@@ -73,6 +75,10 @@ export class PopupMenuComponent {
     event.preventDefault();
 
     action.perform(this.item);
+  }
+
+  hasText(): boolean {
+    return this.text !== null && this.text.length !== 0;
   }
 
 }
--- a/webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
@@ -1,7 +1,7 @@
 import { Component, Input, ElementRef, Renderer2 } from "@angular/core";
 import { isNullOrUndefined } from "util";
 import { PopupMenuAction } from "./popup-menu-action.model";
-
+import { Utils } from "../../shared/Utils";
 /**
  * This component is responsible for displaying a table-row menu button
  * with configurable actions.
@@ -31,6 +31,10 @@ export class PopupMenuComponent {
 
   constructor(private renderer: Renderer2,
               private domElement: ElementRef) {
+  }
+
+  public get hasText(): boolean {
+    return !Utils.isNullOrEmpty(this.text);
   }
 
   public get open(): boolean {
@@ -75,10 +79,6 @@ export class PopupMenuComponent {
     event.preventDefault();
 
     action.perform(this.item);
-  }
-
-  hasText(): boolean {
-    return this.text !== null && this.text.length !== 0;
   }
 
 }
--- a/webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
@@ -17,6 +17,9 @@ import { PopupMenuAction } from "./popup-menu-action.model";
 export class PopupMenuComponent {
 
   @Input()
+  public text: string = '';
+
+  @Input()
   public item: any;
 
   @Input()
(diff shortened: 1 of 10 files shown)


aaronsleeper:
looks good :)

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
```diff
@@ -17,11 +17,22 @@ import { PopupMenuAction } from "./popup-menu-action.model";
...
   @Input()
+  public text: string = '';
+
+  @Input()
   public item: any;
 
   @Input()
   public actions: PopupMenuAction[];
 
+  private _open: boolean;
```
bgsavage: nit: would name this as `isOpen`

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
```diff
@@ -72,4 +77,8 @@ export class PopupMenuComponent {
     action.perform(this.item);
   }
 
+  hasText(): boolean {
+    return this.text !== null && this.text.length !== 0;
```
bgsavage: Seems like we call this in the template a lot and the value seems like it wouldn't change after `ngInit`.  Could you set a property on the component with the result of this evaluation instead?

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -104,15 +104,9 @@ <h3 class="gray-darkest mb-sm">
...
-          <p-column styleClass="cell-dropdown">
-            <ng-template let-data="rowData" pTemplate="body">
-              <popup-menu [item]="data" [actions]="actions"></popup-menu>
-            </ng-template>
-          </p-column>
-
           <p-column sortField="student.lastName" header="{{'labels.groups.results.assessment.exams.cols.name' | translate}}" [sortable]="true">
-            <ng-template let-exam="rowData" pTemplate="body">
-              {{ 'labels.personName' | translate:exam.student }}
+            <ng-template let-data="rowData" let-exam="rowData" pTemplate="body">
```
bgsavage: Hm... do we need both lets?  Or could we just use one instead?

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
esotrope: Not exactly. Though it will never happen ```text``` is an ```@Input()``` and could be bound so ```ngInit``` wouldn't be good practice. Also, this JS execution time is pretty neglidgeable I think. What i think I could though is break it out into the utils.ts because the same logic is in both components.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
esotrope: Is there a consequence to having both? It is to improve the semantic readability of the code.

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
bgsavage: Would be good to temporarily put a log statement here just to see how often it's called.  In angularJS, it alway surprised me to see how often controller methods were called from the template.

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
esotrope: Is that the best practice in ```.ts```? In java we would name it ```open```. Also, I didn't modify this variable just moved it to the top of the class def.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
bgsavage: I would pick a name that can be fitting for both usages then...  The consequence is consistency in usage.  If I were to come in here and add to this template, I wouldn't know if I should use `data` or `exam`.

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
bgsavage: I think it's just a naming practice in general; the verb `open` doesn't tell me it's a boolean so I had to go and find it's type to know what it was when I first saw it.  Especially in ts it's important because `open` could just as easily be a function.

webapp/src/main/webapp/src/app/shared/datatable/datatable-row-expander.component.html
```diff
@@ -0,0 +1 @@
+<button class="btn btn-info btn-xs btn-block text-left" (click)="toggle()"><i class="fa" [ngClass]="{'fa-chevron-right': !isExpanded(), 'fa-chevron-down': isExpanded(), 'mr-xs': hasText()}"></i> {{text}}</button>
```
bgsavage: nit: formatting should be multiple lines if it doesn't change how it looks.

webapp/src/main/webapp/src/app/shared/datatable/datatable-row-expander.component.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  @Input()
+  text: string = '';
+
+  toggle(): void {
+    this.datatable.toggleRow(this.row);
+  }
+
+  isExpanded(): boolean {
+    return this.datatable.isRowExpanded(this.row);
```
bgsavage: You could create these as getters and then you can treat it as a public readonly field from the template.

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
esotrope: True about the possibility of being a function. But I'm not really inclined to change this here unless we decide we want to do it everywhere. I don't think this is the only place we use this kind of naming in our TS.

webapp/src/main/webapp/src/app/shared/datatable/datatable-row-expander.component.ts
esotrope: Good point. That might look nicer.

webapp/src/main/webapp/src/app/assessments/menu/popup-menu.component.ts
```diff
@@ -17,11 +17,22 @@ import { PopupMenuAction } from "./popup-menu-action.model";
...
 
+  private _open: boolean;
+
+  private removeListener: () => void;
+
+  constructor(private renderer: Renderer2,
+              private domElement: ElementRef) {
+  }
+
   public get open(): boolean {
```
esotrope: One thing I notice in all of our code is that we inconsistently declare and omit the ```public``` access modifier. When no access modifier is specified TS makes the method/field ```public```. This class for instance both omits and declares it.

PR #451, opened 2017-09-13
Updated text and button for the home component right side instruction…

bgsavage:
…s to support html.

image

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -36,8 +36,9 @@
       "title": "Access Assessment Results",
       "welcome": {
         "title": "Welcome!",
-        "sub-title": "Get the most from the Reporting System:",
-        "interpret-instructions": "Learn how to interpret the scores in this reporting tool:"
+        "user-guide-html": "<a href=\"{{userGuideUrl}}\" target=\"_blank\">The User Guide</a> describes features of the Smarter Balanced Reporting System and instructions for using each feature.",
+        "interpretive-guide-html": "<a target=\"_blank\" class=\"cursor-pointer\">The Interim Assessments Interpretive Guide</a> is designed to help educators, parents, and other stakeholders interpret interim assessments reports for the Smarter Balanced Reporting System.",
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -36,8 +36,9 @@
       "title": "Access Assessment Results",
       "welcome": {
         "title": "Welcome!",
-        "sub-title": "Get the most from the Reporting System:",
-        "interpret-instructions": "Learn how to interpret the scores in this reporting tool:"
+        "user-guide-html": "<a href=\"{{userGuideUrl}}\" target=\"_blank\">The User Guide</a> describes features of the Smarter Balanced Reporting System and instructions for using each feature.",
+        "interpretive-guide-html": "<a target=\"_blank\" class=\"cursor-pointer\">The Interim Assessments Interpretive Guide</a> is designed to help educators, parents, and other stakeholders interpret interim assessments reports for the Smarter Balanced Reporting System.",
```
bgsavage: This one does not have a href yet as it should just show the "Coming soon" notification for now, right?  Though that's going away and might be a pita to implement a function call from the translate for something going away 9/15.

aaronsleeper: we have a link to a page that says "coming soon"?
are you saying it'll be launched on 9/15 and no longer coming soon?

esotrope: You may want to just use ```href="#"```
Then you wont need the ```cursor-pointer``` css class

bgsavage: href="#" reloads the page, which is not what I want to happen.

bgsavage: Yep, though I'm not sure what the plan is to redeploy on 9/15?  @malaffoon @jtreuting Will we have a hotfix on 9/15 to remove the notifications?

malaffoon: Presumably yes.

PR #453, opened 2017-09-13
No longer throwing exception but instead returning a disabled user ob…

bgsavage:
…ject.  Now when navigating to index.html, checking the user is enabled first otherwise sending them to accessDenied.

Users will now be able to successfully log out from the accessDenied page.  They were not before because the authentication does not get added to the SecurityContextHolder if we throw an exception.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
@@ -42,7 +42,7 @@ public class SamlUserDetailsService implements SAMLUserDetailsService {
 
         final Set<Grant> grants = authorizationService.getGrants(encodedGrants);
 
-		// Secures the application by throwing exception if the user has no authority to access data for this state
+		// Secures the application by setting enabled to false if the user has no authority to access data for this state
         if (grants.isEmpty()) {
             final String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
             logger.warn(message);
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
@@ -47,7 +47,15 @@ public class SamlUserDetailsService implements SAMLUserDetailsService {
             final String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
             logger.warn(message);
 
-            throw new UsernameNotFoundException(message);
+            return User.builder()
+                    .id(id)
+                    .firstName(firstName)
+                    .lastName(lastName)
+                    .email(email)
+                    .username(username)
+                    .password(REDACTED)
+                    .enabled(false)
+                    .build();
         }
 
         final Set<Permission> permissions = authorizationService.getPermissions(grants);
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -47,7 +47,15 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
...
-            throw new UsernameNotFoundException(message);
+            return User.builder()
+                    .id(id)
+                    .firstName(firstName)
+                    .lastName(lastName)
+                    .email(email)
+                    .username(username)
+                    .password(REDACTED)
+                    .enabled(false)
+                    .build();
```
bgsavage: So I'm no longer throwing the exception if a user doesn't have any grants to the tenant, and instead returning a disabled user.  

The problem with throwing the exception is that then the user never gets added to the SecurityContextHolder so then we can never log the user out correctly.  Technically the user is authenticated, just not authorized here.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -14,6 +14,8 @@
     public String index(@AuthenticationPrincipal final User user) {
         if (user == null) {
             return "forward:/landing.html";
+        } else if(!user.isEnabled()) {
+            return "redirect:/error/accessDenied";
```
bgsavage: The problem with this approach is now I have to handle what should be a security concern in the ViewController.   This feels a bit hackish to me but I've been stuck on it for longer than I'd like to admit.  I'm definitely open to alternatives.

esotrope: I think there is an auth failure handler that can route to a URL instead of doing this in the view controller

bgsavage: Will the auth failure handler hit if the auth is successful?  This is what I was hitting before by throwing the exception.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
esotrope: This is kind of an abuse of the purpose of the enabled flag. It is technically for SSO systems where an admin can go in and turn a user account off and on. Similar to "locked" but not related to password retry failure.
I still think we should throw an exception here and I think we can configure an auth failure handler in the SamlWebSecurityConfiguration to handle the view routing logic

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
esotrope: It shouldn't.
I think we set this: ```HttpSecurity#authenticationFailureHandler()```

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
bgsavage: I'm all for it if you can get that to work.  I have been unsuccessful with it, and frankly, I've struggled to find examples online where others throw exceptions in this service and how they are handled.

I can also create a new property on the `User` and use that instead.  Something like `hasAccessToTennant` ?

esotrope: I think ```getPermissionsById()``` will be empty if they don't have tenant access so we could check for that too. Let me fiddle with the failure handler and let you know what I find.

bgsavage:
I'm going to look into putting this in the success handler.

PR #454, opened 2017-09-14
fix iris pathing

wtritch:
It looks like iris hand-jams it's resource paths based upon the frame source path.  So the previous frame source of "/iris" caused malformed resource requests like "/irisScripts/blahblah" rather than "/iris/Scripts/blahblah"

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
@@ -18,5 +18,5 @@
       {{ 'labels.assessments.items.tabs.item-viewer-response-not-included' | translate }}
     </span>
   </p>
-  <iframe class="center" #irisframe src="/iris" width="950" height="500" frameborder="0" ></iframe>
+  <iframe class="center" #irisframe src="/iris/" width="950" height="500" frameborder="0" ></iframe>
 </div>


webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
```diff
@@ -18,5 +18,5 @@
       {{ 'labels.assessments.items.tabs.item-viewer-response-not-included' | translate }}
     </span>
   </p>
-  <iframe class="center" #irisframe src="/iris" width="950" height="500" frameborder="0" ></iframe>
+  <iframe class="center" #irisframe src="/iris/" width="950" height="500" frameborder="0" ></iframe>
```
esotrope: If removing a forward slash where a forward slash is optional breaks things we should add a comment here and probably submit a bug to the IRIS project

bgsavage:
I think you might start to see requests coming in now like /iris//blah/blah/blah

I do believe we had the trailing slash at one point but then removed it.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
malaffoon: 👍  to submitting a bug to IRIS

PR #455, opened 2017-09-14
hotfix/1.0.2

malaffoon:


webapp/src/main/webapp/src/assets/public/landing.html
```diff
@@ -40,11 +40,12 @@
   <div id="about" class="content">
     <h3>Learn More About Smarter Balanced Reporting:</h3>
     <ul>
-      <!--TODO: Remove inline style -->
-      <li class="" style="cursor:default;">
-        <div class="annotatedReports asset"></div>
-        <h2>Interpretive Guide</h2>
-        <p>Learn how to interpret scores in this reporting tool <b>- Available September 15th</b></p>
+      <li class="">
```
esotrope: what does this ghost css class do?

webapp/src/main/webapp/src/styles.less
```diff
@@ -167,4 +167,6 @@ footer {
   }
 }
 
-
+.cursor-pointer {
```
esotrope: I don't think we'll need this now that the interpretive guide href has a value

webapp/src/main/webapp/src/assets/public/landing.html
bgsavage: A whole lot of nothing.

webapp/src/main/webapp/src/styles.less
esotrope: Don't see the comment anymore but I agree this should be removed on develop and isn't critical for this hotfix.

PR #456, opened 2017-09-15
Bugfix/logout access denied

bgsavage:
Now redirecting to an angular access-denied route if a user is able to authenticate but is not authorized with any permissions.  By not throwing an exception, this allows the session to still get added to the `SecurityContextHolder` and allow for a successful logout.

image

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
@@ -42,7 +42,7 @@ public class SamlUserDetailsService implements SAMLUserDetailsService {
 
         final Set<Grant> grants = authorizationService.getGrants(encodedGrants);
 
-		// Secures the application by throwing exception if the user has no authority to access data for this state
+		// Secures the application by setting enabled to false if the user has no authority to access data for this state
         if (grants.isEmpty()) {
             final String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
             logger.warn(message);
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
@@ -47,7 +47,15 @@ public class SamlUserDetailsService implements SAMLUserDetailsService {
             final String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
             logger.warn(message);
 
-            throw new UsernameNotFoundException(message);
+            return User.builder()
+                    .id(id)
+                    .firstName(firstName)
+                    .lastName(lastName)
+                    .email(email)
+                    .username(username)
+                    .password(REDACTED)
+                    .enabled(false)
+                    .build();
         }
 
         final Set<Permission> permissions = authorizationService.getPermissions(grants);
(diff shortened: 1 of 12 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -42,12 +42,19 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
...
         final Set<Grant> grants = authorizationService.getGrants(encodedGrants);
 
-		// Secures the application by throwing exception if the user has no authority to access data for this state
+		// Secures the application by returning a user without any permissions, authorities, or groups.
         if (grants.isEmpty()) {
             final String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
             logger.warn(message);
 
-            throw new UsernameNotFoundException(message);
+            return User.builder()
```
esotrope: Still don't think this is necessary as the ```DefaultAuthorizationService``` filters out all non-tenant grants for the user. The service uses those grants to create the spring authorities and permissions objects so with no grants the user will already have no permissions.
```
@Override
    public Set<Grant> getGrants(final String[] encodedGrants) {
        final ImmutableSet.Builder<Grant> grants = ImmutableSet.builder();
        for (String encodedGrant : encodedGrants) {
            try {
                final Grant grant = Grants.from(encodedGrant);

                // Only add grants for the application's configured state
                if (stateCode.equals(grant.getStateId())) {
                    grants.add(grant);
                }
            } catch (final IllegalArgumentException exception) {

                // Ignore grants with unexpected entity levels or unexpected formatting
                logger.debug("invalid grant: {}", encodedGrant);
            }
        }
        return grants.build();
    }
```

I think removing the ```throw``` here is sufficient.

bgsavage: You don't think it's more efficient to short circuit here?

webapp/src/main/webapp/src/app/user/authorize-at-least-one.can-activate.ts
```diff
@@ -0,0 +1,29 @@
...
+export class AuthorizeAtleastOneCanActivate implements CanActivate {
+
+  constructor(private _service : UserService, private router: Router) {
+  }
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean>|Promise<boolean>|boolean {
+    let observable = this._service.doesCurrentUserHaveAnyPermissions().share();
+
+    observable.subscribe(hasPermissions => {
+      if(hasPermissions) {
```
esotrope: nit: spacing after if
nit: else block not needed since the if block returns

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
esotrope: Good point. That's probably best : j

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -19,6 +19,10 @@
       "title": "Authentication Expired",
       "message": "Click Ok to re-authenticate."
     },
+    "access-denied": {
+      "title": "Access Denied",
+      "message": "Your user account is not permitted to access this reporting system. If you believe you have reached this page in error, please contact your building administrator."
```
malaffoon: `building` administrator?

jtreuting: That is the phrase that Gail wanted us to use.  She gave it to us when there were no student groups

PR #457, opened 2017-09-15
consolidates student and batch report generation in the UI

esotrope:
I don't plan on merging this - just want to get some eyes on the changes earlier than later.
We will need @wtritch's report processor service changes to accommodate the student report before merging.

Notice, individual student reports and batch report dropdowns are now all the same (save for the Sort Order field which is not relevant for individual student reports) and school year selection is offloaded to the top level filters on the page.
They also will now behave the same and submit a report request to the report-processor-service and theoretically show up as records on the "My Reports" page once the backend changes are in place.

screen shot 2017-09-15 at 11 07 22 am

Not sure if we want to, but we can, remove "Subject" from the individual student report download dropdown and delegate to the page filters to decide the subject. I flip flop on doing this however because it would take away from the uniformity of the dropdowns and possibly lead to confusion.

screen shot 2017-09-15 at 11 07 39 am

screen shot 2017-09-15 at 11 07 49 am


webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
```diff
@@ -29,6 +28,12 @@
     }
 
     @Override
+    public Report createReport(@NotNull final User user, @NotNull final StudentExamReportRequest request) {
+        final PermissionScope permissionScope = user.getPermissionsById().get("GROUP_PII_READ").getScope();
+        return createReport(user, request, permissionScope);
```
esotrope: On the final pass we will send the user's permissionsById map and groups to the report processor so it can perform the proper security checks that we have in place for other individual student data shared between PII and GROUP_PII users.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
```diff
@@ -35,9 +39,11 @@
...
             @RequestParam(value = "id", required = false) final Set<Long> reportIds) {
 
-        return reportIds != null
-                ? service.getReports(user, reportIds)
-                : service.getReports(user);
+//        return reportIds != null
+//                ? service.getReports(user, reportIds)
+//                : service.getReports(user);
+
+        return reports();
```
esotrope: mock response - will clean this up

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -44,6 +44,26 @@ export class ReportsComponent implements OnInit, OnDestroy {
...
+          saveAs(download.content, download.name);
+        },
+        () => {
+          this.notificationService.error({ id: 'labels.reports.messages.download-failed' });
+        }
+      );
+  }
+
+  regenerateReport(report: Report) {
+    this.notificationService.info({ id: 'labels.reports.messages.coming-soon.report-regeneration' });
```
esotrope: On the final pass we will add report regeneration of "copying" where we will resubmit the original report request when a user clicks the report when it is in a failed/expired state.

webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -67,7 +67,7 @@ export class DataService {
    * @returns {string} thre formatted name
    */
   private safelyFormatFileName(name: string) {
-    return name == null ? null : name.replace(/\s+/g, '_');
+    return name == null ? null : name.replace(/[&~@#$^*_+=/:?;\\|<>"',]/g, '').replace(/\s+/g, '_');
```
esotrope: This allows us to make the report name/label anything but the saved file will not have the path unfriendly symbols in it.

bgsavage:
So yesterday we had the discussion about making these popovers as modals instead as we continue to add more options (discussion came up because of the feature to optionally add accommodations).  If we're making changes now, should we consider moving to a modal sooner rather than later?

esotrope:
@bgsavage yes sooner than later would be good.
@jtreuting @aaronsleeper - any vision on the layout of the modal in the works?

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
```diff
@@ -49,6 +55,16 @@ public void getReport(
...
     }
 
+    @PostMapping("/api/students/{studentId}/report")
+    @ResponseStatus(HttpStatus.CREATED)
+    public Report createReport(
+            @PathVariable final long studentId,
+            @AuthenticationPrincipal final User user,
+            @RequestBody final StudentExamReportRequest request) throws IOException {
+
+        return service.createReport(user, request.copy().studentId(studentId).build());
```
bgsavage: What's the purpose of copying?

esotrope: <i>Imutatability</i>? Saves us from writing out a builder instantiation and copying over all the fields here.

aaronsleeper: let's just start by moving what we have from a popover to a modal
i don't think it'll need a design but we can tweak after you move it if needed

jtreuting:
Here is what we put together for adding the Show/Hide accommodations in the modal: https://confluence.fairwaytech.com/display/SBACDW/sow-10+-+Prevent+accommodations+displayed+on+student+report

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
bgsavage: Why isn't the `StudentExamReportRequest` sent with a populated `studentId` then?

webapp/src/main/webapp/src/app/report/report-naming.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  nameSchoolGradeExamReport(school: School, grade: Grade, options: ReportOptions): string {
+    return this.name(this.concat(school.name, this.translate.instant(`labels.grades.${grade.code}.short-name`)), options);
+  }
+
+  private name(target: string, options: ReportOptions): string {
+    return this.concat(
+      target,
+      options.schoolYear.toString(),
+      this.nameLanguage(options.language)
+    );
```
bgsavage: What does `name` refer to in this context?

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
malaffoon: The perennial question for REST API's ... if the payload and the resource url both have the same identifier, which do you trust? Is it design smell?

webapp/src/test/java/org/opentestsystem/rdw/reporting/report/student/StudentExamReportControllerIT.java
```diff
@@ -8,6 +8,7 @@
 import org.opentestsystem.rdw.reporting.common.report.ExamReportMessage;
 import org.opentestsystem.rdw.reporting.common.report.ExamReportPdfMessageConverter;
 import org.opentestsystem.rdw.reporting.common.report.PrintOptions;
+import org.opentestsystem.rdw.reporting.common.report.StudentExamReportRequest;
```
malaffoon: needed?

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
esotrope: Has to do with how spring maps POST requests. It will take the request body json and map it to the @RequestBody annotated POJO omitting the path variables. So if the setting is in the path we have to add it there. We could do this in the angular code too i guess but at least for the groups we will need this because we actually have to resolve the group to which the user is referring to with the groupId path variable.

webapp/src/test/java/org/opentestsystem/rdw/reporting/report/student/StudentExamReportControllerIT.java
esotrope: There is a place and a time for everything Mark. And right now it is time for this line to be deleted : j

webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -13,8 +13,8 @@ <h3 class="blue mb-md">{{'labels.reports.heading' | translate}}</h3>
     <p-column field="label" header="{{'labels.reports.column.label.header' | translate}}" sortable="true">
       <ng-template let-report="rowData" pTemplate="body">
         <span *ngIf="report.isProcessing()">{{report.label}}</span>
-        <button *ngIf="report.isCompleted()" class="btn btn-info btn-xs" (click)="getReport(report)"><i class="fa fa-cloud-download" aria-hidden="true"></i> {{'labels.reports.form.submit' | translate}} {{report.label}}</button>
-        <button *ngIf="report.canRegenerate()" class="btn btn-default btn-xs" (click)="regenerateReport(report)"><i class="fa fa-refresh" aria-hidden="true"></i> {{'labels.reports.form.regenerate' | translate}} {{report.label}}</button>
+        <button *ngIf="report.isCompleted()" class="btn btn-info btn-xs" (click)="getReport(report)"><i class="fa fa-cloud-download mr-xs" aria-hidden="true"></i> {{report.label}}</button>
+        <button *ngIf="report.canRegenerate()" class="btn btn-default btn-xs" (click)="regenerateReport(report)"><i class="fa fa-refresh mr-xs" aria-hidden="true"></i> {{report.label}}</button>
```
bgsavage: I still will advocate these be props instead of methods when doing `*ngIf`.

esotrope: sounds good. Under the hood though will it be called any more or less? a getter/setter is a method after all.

webapp/src/main/webapp/src/app/report/reports.component.ts
bgsavage: Should `id` here be `translateKey` ?

aaronsleeper:
https://confluence.fairwaytech.com/display/SBACDW/sow-10+-+Prevent+
accommodations+displayed+on+student+report looks great

webapp/src/main/webapp/src/app/report/reports.component.ts
esotrope: Yeah but that is a long var name : j

esotrope:
Changed popovers to modals and added accommodations toggle.

screen shot 2017-09-18 at 5 26 46 pm

screen shot 2017-09-18 at 5 26 59 pm

screen shot 2017-09-18 at 5 26 12 pm

screen shot 2017-09-18 at 5 25 48 pm

webapp/src/main/webapp/src/app/shared/common.module.ts
```diff
@@ -40,12 +42,14 @@ import { DataTableRowExpanderComponent } from "./datatable/datatable-row-expande
...
     SBCheckboxList,
+    SBToggleComponent,
     SchoolYearPipe,
     SearchPipe,
     SessionExpiredComponent,
     SubjectPipe,
     LoaderComponent,
-    DataTableRowExpanderComponent
+    DataTableRowExpanderComponent,
+    InformationLabelComponent
```
esotrope: moved information label from assessment module to common module because it is used in reports module which shouldn't depend on the assessment module just for that label component.

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -0,0 +1,63 @@
...
+        <input [value]="true" [(ngModel)]="value" type="radio" [name]="name">{{'buttons.visibility.show' | translate}}
+      </label>
+      <label class="btn btn-primary" [ngClass]="{'active': !value}">
+        <input [value]="false" [(ngModel)]="value" type="radio" [name]="name">{{'buttons.visibility.hide' | translate}}
+      </label>
+    </div>
+  `,
+  providers: [ CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR ]
+})
+export class SBToggleComponent implements ControlValueAccessor {
```
esotrope: Toggle component for the accommodations flag.
Implementing ControlValueAccessor allows you to use [(ngModel)] on your component which was pretty neat.
```<sb-toggle [(ngModel)]="options.accommodationsVisible"></sb-toggle>```

File: webapp/src/main/webapp/src/styles.less [hunk 11]
```diff
@@ -23,7 +23,7 @@
 @import "primeng-overrides.less";
 
 // Local variables
-@zindex-fab: 1065;
+@zindex-fab: 1040;
```
esotrope: This makes sure the floating action button goes below modal backgrounds

bgsavage:
I think we have a mismatch in the title of the modal saying "Print" Ethel's Report and the actionable btn-primary saying "Download".  Which is it? :)

webapp/src/main/webapp/src/app/shared/common.module.ts
bgsavage: 👍

esotrope:
@bgsavage my local has "Print {{firstName}}'s  ..." as the menu text. probably a recent change.

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -13,23 +13,41 @@ import { Report } from "./report.model";
...
+  lockAssessmentType: boolean = false;
 
   @Input()
   subject: string;
 
+  @Input()
+  lockSubject: boolean = false;
+
+  @Input()
+  displayOrder: boolean = true;
```
bgsavage: Lots and lots of @Input()s here.  Would consider creating a ReportDownloadOptions model to compose all the options into a single parameter just like we would in java with an ever growing parameter list.

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -0,0 +1,63 @@
...
+
+@Component({
+  selector: 'sb-toggle',
+  template: `
+    <div class="btn-group btn-group-sm toggle-group" data-toggle="buttons">
+      <label class="btn btn-primary" [ngClass]="{'active': value}">
+        <input [value]="true" [(ngModel)]="value" type="radio" [name]="name">{{'buttons.visibility.show' | translate}}
+      </label>
+      <label class="btn btn-primary" [ngClass]="{'active': !value}">
+        <input [value]="false" [(ngModel)]="value" type="radio" [name]="name">{{'buttons.visibility.hide' | translate}}
```
bgsavage: I believe the other button components pass in a translation enum so that you can customize the text based on the enum.

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
bgsavage: Does this call your register functions down below for you?

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
esotrope: Hm. I was thinking to generalize this for broader consumption i would make an "option" model where you can set the values like 
```<sb-toggle [options]="[{value: true, text: '...'}, {value: false, text: '...'}]">```
thoughts?

File: webapp/src/main/webapp/src/app/report/report-download.component.ts (hunk 12 above)
Comment (esotrope) [reply webapp/src/main/webapp/src/app/report/report-download.component.ts]: Hm. That might introduce complexity when binding though. I would have to make sure the changes of the model propagate to the component.

File: webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts (hunk 10 above)
Comment (esotrope) [reply webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts]: Yeah, the only method i added outside the interface was ```set value``` I think.

File: webapp/src/main/webapp/src/app/report/report-download.component.ts (hunk 12 above)
Comment (bgsavage) [reply webapp/src/main/webapp/src/app/report/report-download.component.ts]: Pretty sure it'll pass by ref.

File: webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts (hunk 13 above)
Comment (wtritch) [reply webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts]: I like that, and the text would be the translation key?

File: webapp/src/main/resources/application.yml [hunk 14]
```diff
@@ -115,7 +115,7 @@ app:
       zoom: 0.75
 
   reportservice:
-    url: http://localhost:8083
+    url: http://localhost:8085
```
wtritch: Does this match the port in the docker-compose.yml?  If so, maybe a small comment would be appropriate?

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
wtritch: hmm... does TypeScript have an @Override annotation so we can see what's an interface at-a-glance?

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
esotrope: Yeah - something like that. It would be cool if we could use the translation pipe inline here hm...
Otherwise our sb comps will depend on the ng-translate lib (i think they already do) but ideally they would be lib agnostic

webapp/src/main/resources/application.yml
esotrope: Yeah it does, this speeds up the local debug process when inserting your intellij stood up instances into the docker relm

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
esotrope: Looks like there is talk about adding an override keyword to TS but nothing of yet. I think for now we can add /** @override */ JS doc

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -0,0 +1,117 @@
...
+        <input type="radio" [name]="name" [disabled]="disabled" [value]="options[0].value" [(ngModel)]="value">{{options[0].text}}
+      </label>
+      <label class="btn btn-primary" [ngClass]="{'active': options[1].value === value, disabled: disabled}">
+        <input type="radio" [name]="name" [disabled]="disabled" [value]="options[1].value" [(ngModel)]="value">{{options[1].text}}
+      </label>
+    </div>
+  `,
+  providers: [ CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR ]
+})
+export class SBToggleComponent implements ControlValueAccessor, OnInit {
```
esotrope: Made this class more flexible and ready for sharing among ui kit consuming projects

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -1,42 +1,86 @@
...
+                <span info-label
+                      title="{{'labels.reports.form.accommodations.title' | translate}}"
+                      content="{{'labels.reports.form.accommodations.info' | translate}}"></span>
+              </label>
+              <sb-toggle name="accommodations-toggle"
+                         [(ngModel)]="options.accommodationsVisible"
+                         [disabled]="options.assessmentType === AssessmentType.IAB"
+                         [options]="[
+                           {value: true, text: 'buttons.visibility.show' | translate},
+                           {value: false, text: 'buttons.visibility.hide' | translate}
```
esotrope: Honestly didn't expect this to work but it does. Angular EL supports pipes nested in objects 👍

webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -12,14 +12,14 @@ <h3 class="blue mb-md">{{'labels.reports.heading' | translate}}</h3>
...
-        <button *ngIf="report.canRegenerate()" class="btn btn-default btn-xs" (click)="regenerateReport(report)"><i class="fa fa-refresh" aria-hidden="true"></i> {{'labels.reports.form.regenerate' | translate}} {{report.label}}</button>
+        <span *ngIf="report.processing">{{report.label}}</span>
+        <button *ngIf="report.completed" class="btn btn-info btn-xs" (click)="download(report)"><i class="fa fa-cloud-download mr-xs" aria-hidden="true"></i> {{report.label}}</button>
+        <button *ngIf="report.regenerable" class="btn btn-default btn-xs" (click)="regenerate(report)"><i class="fa fa-refresh mr-xs" aria-hidden="true"></i> {{report.label}}</button>
       </ng-template>
     </p-column>
     <p-column field="status" sortable="true" header="{{'labels.reports.column.status.header' | translate}}">
       <ng-template let-report="rowData" pTemplate="body">
-        <span class="label border-only {{report.getStatusColor()}}">{{ 'labels.reports.status.' + report.status | translate }}</span>
+        <span class="label border-only {{report.statusColor}}">{{ 'labels.reports.status.' + report.status | translate }}</span>
```
esotrope: @bgsavage changed the report methods to getters

common/src/main/resources/templates/batch-reports.html
```diff
@@ -1,28 +1,28 @@
...
-    </div>
-    <div th:if="${studentCompositeReport.icaEla != null}"
-         th:with="report=${studentCompositeReport.icaEla}">
-        <div th:replace="ica-body::body"></div>
-    </div>
-</div>
-</body>
+    <head>
+        <div th:replace="fragments::meta"></div>
+        <title th:text="${request.name}"></title>
```
esotrope: The real change here was to switch to using the request's report name.
This way a meaninful browser tab title will display instead of the static "Student Reports" text.

PR #459, opened 2017-09-19
Feature/refactor report processor

wtritch:
This PR refactors the report processor to use Spring Stream for distributed processing flow rather than Spring Batch for an internal-only processing flow.

UPDATE:
A concern I have is that although the Rabbit (and presumably Kafka) bindings have a retry policy implemented, we don't trap processing failures and flag the report as FAILED.  Does anyone have a good idea for handling that?  My first instinct is to turn off the built-in retry impl, add an X-RETRY-COUNT header to the message, wrap each @ServiceActivator method in a try/catch on Exception, and handle failures manually with either repost or retry exhaustion manually.  Thoughts?


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
```diff
@@ -12,11 +12,12 @@
 public class Report {
     private Long id;
     private String user;
-    private Long jobExecutionId;
```
wtritch: We no longer need to track a spring batch job execution

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
```diff
@@ -12,11 +12,12 @@
...
     private ReportStatus status;
     private URI reportResourceUri;
     private String label;
     private AbstractExamReportRequest reportRequest;
+    private int totalChunkCount;
+    private int completeChunkCount;
```
wtritch: Now that chunk processing is potentially distributed among multiple microservice instances we need a central source of truth for how many chunks have been completed for a report.  This is required so that only the last chunk completion triggers the final report collation step where all the chunks are combined into the single final report.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/ReportProcessorConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+import static com.google.common.collect.Maps.newHashMap;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+@Configuration
+@EnableConfigurationProperties(ReportGenerationProperties.class)
+@EnableBinding({
+        FetchStudentsSource.class,
+        FetchStudentsProcessor.class,
+        GenerateStudentsReportProcessor.class,
+        CollateStudentReportsSink.class
```
wtritch: These interfaces define the spring stream flow.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ReportRequestMessage.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.opentestsystem.rdw.reporting.common.security.PermissionScope;
+
+/**
+ * This message represents a basic report processing message.
+ */
+public class ReportRequestMessage {
+
+    protected long reportId;
+    protected PermissionScope permissionScope;
```
wtritch: @esotrope This will need to change to provide the UserPermission object containing the map of all user permissions as well as the explicitly-assigned groups/districts.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
```diff
@@ -126,6 +129,23 @@ public Report findById(final long reportId) {
...
+                        .addValue("report_id", report.getId()),
+                keyHolder);
+
+        if (updatedCount != 1) {
+            throw new IllegalStateException("Unable to update report");
+        }
+        return Report.builder()
+                .copy(report)
+                .completeChunkCount(keyHolder.getKey().intValue())
+                .build();
```
wtritch: Marking a chunk completed increments the completion count on the report record and returns the incremented value.  This happens in a thread-safe manner, so only one request from one thread from one microservice instance believes that it has completed the final chunk.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportGenerationTemporaryStorage.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.function.Supplier;
+
+/**
+ * This class is responsible for temporarily storing student Report chunk
+ * content while generating a Report.  It is backed by an ArchiveService.
+ *
+ * It is responsible for cleaning up the temporary storage when requested.
+ */
+@Service
+public class ArchiveBackedReportGenerationTemporaryStorage implements ReportGenerationTemporaryStorage {
```
wtritch: Since we now have chunk processing spread across multiple microservice instances, we need to use the ArchiveService to provide a central storage location for each chunk report PDF.

report-processor/src/main/resources/application.yml
```diff
@@ -18,11 +18,34 @@ spring:
...
+          group: default
+          consumer:
+            concurrency: 2
+          producer:
+            requiredGroups: default
+
+        REPORT_COLLATE:
+          group: default
+          producer:
+            requiredGroups: default
```
wtritch: These are the binding configurations.  Note that both fetching the students and collating the reports happens on 1 thread per channel, but generating a PDF for a chunk of students is multi-threaded.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportServiceIT.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        executorService.shutdownNow();
+        template.update("DELETE FROM user_report;", new MapSqlParameterSource());
+    }
+
+    @Test
+    public void itShouldOnlyReturnChunksCompletedOnce() throws Exception {
```
wtritch: This multi-threaded test is ensuring that only one chunk completion believes it is the last completion.  I've run this locally with 100 threads, and tried it using @agorina 's MultiThreadTest implementation in the RDW_Ingest project.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/web/ReportControllerIT.java
```diff
@@ -89,10 +93,12 @@ public void itShouldGenerateAReport() throws Exception {
                 .content(serializedPayload))
 
                 .andExpect(status().isAccepted())
-                .andExpect(jsonPath("$.*", hasSize(4)))
+                .andExpect(jsonPath("$.*", hasSize(6)))
                 .andExpect(jsonPath("$.id").value(report.getId()))
                 .andExpect(jsonPath("$.status").value(report.getStatus().name()))
                 .andExpect(jsonPath("$.label").value(report.getLabel()))
+                .andExpect(jsonPath("$.totalChunkCount").value(10))
+                .andExpect(jsonPath("$.completeChunkCount").value(9))
```
wtritch: @esotrope In case we want to report on this (I'm not sure we do) we can now give the user a sense of progress once the totalChunkCount is set higher than 0 during the initial processing step.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+                .iabMath(value.get(MATH))
+                .build()
+        ));
+    }
+
+    private void appendICAReports(final Map<Long, StudentCompositeReport> compositeReports,
+                                  final PermissionScope permissionScope,
+                                  final Collection<Long> studentIds,
+                                  final ExamQueryParams examQueryParams) {
+        if (examQueryParams.getAssessmentType() != null && examQueryParams.getAssessmentType() != ICA) {
```
esotrope: should these guards be delegated to the caller? Having them in the method is safest I realize but also makes less sense. It would read more logically if we check this then conditionally call appendICAReports() otherwise we might want to add "conditionally" in the method name or something.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/web/ReportControllerIT.java
esotrope: Cool 👍 . Not sure we want to unless we use websockets to stream the progress to the user. Otherwise we would probably want to poll in a tighter interval update a progress bar.
OR we can do one of those pseudo progress bars that just animates to 80% when started then animates to 100% when done X D

esotrope:
@wtritch change looks great will.
Looking through I didn't see any explicit support/test for single student reports - is this to come with the "UserPermission" param change in a follow up PR?

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/web/ReportControllerIT.java
wtritch: bleh, the only think I dislike more than pseudo progress bars are progress bars that restart when they complete =P

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: I'm not sure about creating an if-chain in the caller, that always turns into "block-of-text" to my eye until I focus on it and parse it out.  This might be clearer if I inverted the check to a negative assertion like:
```
if (!(examQueryParams.getAssessmentType() == null || examQueryParams.getAssessmentType() == ICA)) {
  return;
}
```
or even broke it into a check method:
```
if (shouldNotAppend(ICA)) {
  return;
}
```
thoughts?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/ReportProcessorConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+    public void configure(final WkhtmltopdfWebServiceHtmlToPdfProcessor processor) {
+        processor.setOptions(wkhtmltopdfOptions());
+    }
+
+    @Bean
+    public RetryTemplate retryTemplate() {
+        //Back-off policy from 1 second x2 per attempt
+        final ExponentialBackOffPolicy backOffPolicy = new ExponentialBackOffPolicy();
+        backOffPolicy.setInitialInterval(SECONDS.toMillis(1));
+        backOffPolicy.setMultiplier(2);
```
bgsavage: This retry strategy is quickly becoming your staple.  I like it :)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
esotrope: I think we should always write to positive conditions if we can.
I still would prefer to see this in the caller:
```java
if (includesIcaReports(params)) {
  addIcaReports(...);
}
```

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
bgsavage: Do we need to send back an entire `Report` here?  If not, the only required input is the reportId.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/ReportProcessorConfiguration.java
esotrope: WillRetry (TM)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
```diff
@@ -23,6 +24,7 @@ public DefaultReportService(final ReportRepository reportRepository) {
     }
 
     @Override
+    @Transactional
     public Report insert(final Report report) {
         return reportRepository.insert(report);
```
bgsavage: Is `insert` the java standard naming convention for create / add?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudentsSource.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.springframework.cloud.stream.annotation.Output;
+import org.springframework.messaging.MessageChannel;
+
+/**
+ * This source publishes student report requests.
+ */
+public interface FetchStudentsSource {
+
+    String REPORT_FETCH_STUDENTS = "REPORT_FETCH_STUDENTS";
```
bgsavage: nit: should this be a constant?  Also I thought constants were PascalCase

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
esotrope: I think the correct way to say that is - "There's no I in CRUD!"

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+                .findAllForStudentsByExamFilter(
+                        permissionScope,
+                        studentIds,
+                        examQueryParams);
+
+        iabReports.forEach((key, value) -> compositeReports.computeIfPresent(key, (id, compositeReport) -> StudentCompositeReport.builder()
+                .copy(compositeReport)
+                .iabEla(value.get(ELA))
+                .iabMath(value.get(MATH))
+                .build()
```
bgsavage: This feels odd to me, should you be returning `compositeReports` since you are modifying it and is the objective of this method?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
```diff
@@ -88,7 +91,7 @@ public Report update(final Report report) {
                 new MapSqlParameterSource()
                         .addValue("id", report.getId())
                         .addValue("user_login", report.getUser())
-                        .addValue("job_execution_id", report.getJobExecutionId())
+                        .addValue("total_chunk_count", report.getTotalChunkCount())
```
malaffoon: Not sure how update is used but i would think the generic update method would also set complete_chunk_count. If the update is used only to update the status (my guess) perhaps we should replace this method with `updateStatus`?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/ReportProcessorConfiguration.java
agorina: Should these values come from the configuration (interval, multiplier and retry policy below)?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
malaffoon: Personally i'd simplify the method to take the report id and return the (new) count. But, if the only use case for it likes this signature, i'm fine with it.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
```diff
@@ -23,6 +24,7 @@ public DefaultReportService(final ReportRepository reportRepository) {
     }
 
     @Override
+    @Transactional
```
malaffoon: Why are these transactional if they are making a single call to a repo?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudentsSource.java
malaffoon: Will is what we call and "old dog". Camel casing is what we call a "new trick". You do the math  ;)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
agorina: I know it is debatable, and I know you do not like @Transactional on the repository, but I feel that because we only have one transaction manager and do not plan on having more than one, it would feel safer to add this annotation to the Repository. 

One argument for it that the Repository ITs would function more like the real world.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/GeneratePdf.java
```diff
@@ -0,0 +1,84 @@
...
+        this.reportGenerationTemporaryStorage = reportGenerationTemporaryStorage;
+        this.pdfMerger = pdfMerger;
+        this.reportService = reportService;
+        this.sink = sink;
+    }
+
+    @ServiceActivator(inputChannel = REPORT_COLLATE)
+    public void collateReportPdf(final ReportRequestMessage message) {
+        logger.debug("Received report collation message for report: {}", message.getReportId());
+        final Report report = reportService.findOneById(message.getReportId())
```
malaffoon: see reply to your global comment

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+    public StudentReportProcessor(final Set<ExamQueryParamsParser> queryParamsParsers,
+                                  final IabReportRepository iabReportRepository,
+                                  final IcaReportRepository icaReportRepository,
+                                  final ReportService reportService,
+                                  final ReportGenerationTemporaryStorage temporaryStorage,
+                                  final ReportViewSupport support,
+                                  final TemplateProcessor templateProcessor,
+                                  final HtmlToPdfProcessor htmlToPdfProcessor,
+                                  final RetryTemplate retryTemplate,
+                                  final GenerateStudentsReportProcessor processor) {
```
malaffoon: dang, that's a lot of helpers

report-processor/src/main/resources/application.yml
```diff
@@ -18,11 +18,34 @@ spring:
     type: SIMPLE
     cache-names: internalTranslationsByCode
 
+  cloud:
+    stream:
+      bindings:
+        REPORT_FETCH_STUDENTS:
```
malaffoon: I have to say that CamelCasing would make this look better, i.e. `report-fetch-students` or `reportFetchStudents`. I know the values in ingest are uppercase but that was me being naive early on (and using the ImportContent enum as the driver).

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
esotrope: Concurrency support no? I thought the spring transaction manager registers only @Transactional methods and makes sure to best prevent race conditions between them. Depending on the propagation strategy you chose it will have different degrees of integrity.

report-processor/src/main/resources/application.yml
```diff
@@ -18,11 +18,34 @@ spring:
...
+          group: default
+          producer:
+            requiredGroups: default
+
   jackson:
       default-property-inclusion: non_null
       serialization:
         WRITE_DATES_AS_TIMESTAMPS: false
 
+  rabbitmq:
```
malaffoon: hmm, we don't put defaults in anywhere else and this is the default value for it, so why include it here?

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportServiceIT.java
```diff
@@ -0,0 +1,98 @@
...
+            });
+        }
+        startSemaphore.release(chunkCount);
+        endSemaphore.tryAcquire(chunkCount, 30, TimeUnit.SECONDS);
+
+        assertThat(isLastChunkResults.stream().filter((x) -> x)).hasSize(1);
+        for (int i = 1; i <= chunkCount; i++) {
+            assertThat(completeCount).contains(i);
+        }
+    }
```
agorina: Nice, thank you for this test!

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportServiceIT.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Before
+    public void setup() {
+        executorService = Executors.newFixedThreadPool(chunkCount);
+    }
+
+    @After
+    public void tearDown() {
+        executorService.shutdownNow();
+        template.update("DELETE FROM user_report;", new MapSqlParameterSource());
```
agorina: I know we are running ITs on the test DB, but just in case - I believe we have been trying to make this type of queries as safe as we can if by accident it gets executed on the real one.  Is there anything you can add to the DELETE to make sure it only deletes the ones generated by this test?

malaffoon:
Our philosophy elsewhere has been to make sure the message processing method *never* throws. Unless there is a situation where you think the message could be successfully processed during a retry (which is not the case for any of the ingest messages). I know it is a naive strategy that fails to take advantage of some of the message queue features but it is simple and, so far, effective.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
malaffoon: Don't fall into the trap i always hit: Transactions != Concurrency.

The only purpose of a Transaction is to make sure the data is consistent within the method. If the method makes a single call, there is no consistency to worry about.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportServiceIT.java
jtreuting: doesn't spring roll back for ITs when the @Transactional is present?  is this not the way we have been handling things in general?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
wtritch: This is more of a conceptual debate than a functional one.  Update is used to:
1. (functionally) update the status
2. (functionally) update the total chunk count
3. (conceptually) update any other property

Update is *not* used to update the complete_chunk_count because that is a separate, highly-managed path.  Conceptually, I suppose we could have broken out the complete_chunk_count into a separate join table.  It really only exists as a column on this table for convenience and performance.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
wtritch: Because @agorina told me to =)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
malaffoon: Perhaps just add to the javadoc then. I *know* somebody will get burned in the future when they try to use update and it doesn't update all the fields.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
agorina: To be more precise it depends on the Transaction Isolation Level and in some cases  Transaction = Concurrency, but you usually do not want it.

wtritch: The Repository ITs are already wrapped in @Transactionals, though.  If the repository is making a single call, it's already transactional by default, yes?
In general, everything I've read recommends keeping @Transactional annotations at the business logic layer, where you may make multiple calls to the DB and want them all to succeed or fail together.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: Yeah, this is where the bulk of the logic lives.  This gets a collection of students and does everything required to create and persist a PDF for that collection of students.  I could break it up into multiple services, but wasn't sure if that would actually increase clarity.  Thoughts?

report-processor/src/main/resources/application.yml
wtritch: I meant to call this out.  This service *requires* rabbit to function.  I'd like to let someone coming in blind know that without having to look at the dependency list.  That being said, they'll figure it out pretty quick if they just try to run the app...  Thoughts?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
agorina: @wtritch,  you are right about me asking for @Transaction on the update Repository call because logically it updates and reads the last updated value. And I remember that with the ingest project we had issues without this annotation in some cases. 

I admit that I am not very familiar with that mySQL magic you are doing there, so I could be wrong.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportServiceIT.java
wtritch: Yeah, I'll track the created Report#id and just delete that one.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
wtritch: I like the idea of the service and repository working on model instances rather than having functional methods.  It helps keep the API and expectations clean in my opinion.  Plus, then if the consumer *does* need an updated Report instance, they don't have to manually copy/build the updated value in.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/ReportProcessorConfiguration.java
wtritch: @agorina yes, they should.  I created a ReportGenerationRetryProperties to hold the three configuration values with defaults set to the values here.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepository.java
wtritch: good call, done

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
wtritch: Naming is my bane.  Changed to "create"

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: hmmm... good question.  I generally dislike parameter modification methods and prefer return values, but with the name "appendBlah" I feel like it's fairly clear what this method is doing.  Thoughts?

report-processor/src/main/resources/application.yml
wtritch: I'll remove it, since one of these things is not like the others.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: Done, with a small change:
```
        if (includeReports(IAB, queryParams)) {
            appendIABReports(studentReports, message.getPermissionScope(), studentReports.keySet(), queryParams);
        }
        if (includeReports(ICA, queryParams)) {
            appendICAReports(studentReports, message.getPermissionScope(), studentReports.keySet(), queryParams);
        }
```

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudentsSource.java
wtritch: ok, ok, changed to CamelCase.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
bgsavage: I'm guessing you don't want to just return a map that the caller can then append to `compositeReports`?

Maybe `appendIabReportsToCompositeReports` ?  (Also nit: IAB should be Iab if an acronym is 3 letters or more, right? )

report-processor/src/main/resources/application.yml
wtritch: done

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportServiceIT.java
wtritch: Discussed offline with @jtreuting but this test class doesn't have an @Transactional annotation, so each thread in the executorService can open it's own DB transaction.  The downside of that is that we're forced to clean up after ourselves.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: The caller is already pretty bulky, I'd rather not put the merging logic in there.  
As for the name... that's kind of a mouthful and I don't feel like it really increases readability all that much.  
As for the IAB -> Iab, does that apply to the AssessmentType enum as well?  That sounds like a rule, is that a rule?  Where do all these rules come from??

wtritch:
@malaffoon I hear what you're saying about never throwing.  I can only picture a few exceptions that *may* be recoverable, but they're likely long-lasting and it may be simpler and more user-friendly to fail the report and have the user re-request it:
1. S3 is unreachable due to Amazon system outage/misconfiguration/credential expiration/etc
2. wkhtmltopdf processor is unreachable due to Pod or Node outage/misconfiguration/etc
3. Database is unreachable due to Amazon system outage/misconfiguration/etc
4. RabbitMQ is unreachable due to ""
5. InterruptedException caused by Pod shutdown/kill (graceful SIGTERM shutdown will allow the pods to finish their current task before shutting down)
6. ???

Most of the other exceptions are non-recoverable:
1. Report Id can't be found in the Database
2. Processing failure due to bad data
3. Some other unknown systematic bug

My only real goal is to prevent reports from being stuck in "Pending"-limbo, where they have failed but we haven't updated the Report record to reflect that fact.

After reading a bit, I think I'm going to register a @GlobalChannelInterceptor on all "Report*" channels that will mark the report as failed if any exception is seen during either sending or receiving and add a fast-fail to all of the message handlers that will abort if the Report is flagged as FAILED.  Thoughts?

wtritch:
I added a ReportFailingChannelInterceptor class that flags the Report as FAILED if an exception occurs during processing.

PR #461, opened 2017-09-19
Now returning a translateService.get as a resolve instead of the RdwT…

bgsavage:
…ranslateLoader observable in order to check that the translate service is done loading prior to rendering content.

This fixes dwr-1017 which has the same cause as dwr-555

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/
+out/
 !.mvn/wrapper/maven-wrapper.jar
 src/main/resources/static/
 
--- a/webapp/src/main/webapp/src/app/home/translate.resolve.ts
+++ b/webapp/src/main/webapp/src/app/home/translate.resolve.ts
@@ -2,16 +2,17 @@ import { Injectable } from "@angular/core";
 import { Resolve, ActivatedRouteSnapshot, RouterStateSnapshot } from "@angular/router";
 import { Observable } from "rxjs";
 import { RdwTranslateLoader } from "../shared/rdw-translate-loader";
+import { TranslateService } from "@ngx-translate/core";
 
 /**
  * Resolver that only resolves once the translate getTranslation is complete.
  */
 @Injectable()
 export class TranslateResolve implements Resolve<string> {
-  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<any>|Promise<any>|any {
-    return this.service.observable;
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<string>|Promise<string>|string {
+    return this.service.get("welcome");
   }
 
-  constructor(private service: RdwTranslateLoader) {
+  constructor(private service: TranslateService) {
   }
 }
--- a/webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
+++ b/webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
@@ -7,36 +7,28 @@ import { Injectable } from "@angular/core";
 
 @Injectable()
 export class RdwTranslateLoader {
-  /**
-   * @returns The observable for the getTranslation request.
-   */
-  public get observable() {
-    return this._observable;
-  };
-
   constructor(private http: Http) {
   }
 
   private apiLoader = new TranslateHttpLoader(this.http, '/api/translations/', '');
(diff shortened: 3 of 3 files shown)


webapp/src/main/webapp/src/app/home/translate.resolve.ts
```diff
@@ -2,16 +2,17 @@ import { Injectable } from "@angular/core";
...
 
 /**
  * Resolver that only resolves once the translate getTranslation is complete.
  */
 @Injectable()
 export class TranslateResolve implements Resolve<string> {
-  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<any>|Promise<any>|any {
-    return this.service.observable;
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<string>|Promise<string>|string {
+    return this.service.get("welcome");
```
bgsavage: Instead of returning RdwTranslateLoader's observable, now just trying to get a value out of the translate service which will resolve after RdwTranslateLoader.

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -7,36 +7,28 @@ import { Injectable } from "@angular/core";
 
 @Injectable()
 export class RdwTranslateLoader {
-  /**
-   * @returns The observable for the getTranslation request.
-   */
-  public get observable() {
-    return this._observable;
-  };
```
bgsavage: No longer need to expose this, so removing.

webapp/src/main/webapp/src/app/home/translate.resolve.ts
esotrope: So basically, the change is to initiate a translation request at "resolve-time"? Before i'm guessing it started loading when translate.use() was called?

bgsavage: Yep, which causes an issue with the primeng dropdown when the "Select" placeholder text hasn't been translated yet, and therefore assumes there is no placeholder and preselects the first school in the dropdown.

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -7,36 +7,28 @@ import { Injectable } from "@angular/core";
...
   private uiLoader = new TranslateHttpLoader(this.http, '/assets/i18n/', '.json');
-  private _observable: Observable<any>;
 
   getTranslation(lang: string) {
     let uiObservable = this.uiLoader.getTranslation(lang);
     let apiObservable = this.apiLoader.getTranslation(lang).catch(res => Observable.of({}));
 
     let translateObserver: Observer<any>;
-    this._observable = new Observable<any>(observer => translateObserver = observer).share();
+    let observable = new Observable<any>(observer => translateObserver = observer);
```
esotrope: Should we be adding ```: type``` to all var declarations? I feel like this is another thing we do very inconsistently. If we buy into TS typing we are probably better off embracing it everywhere for consistency maybe.

.gitignore
```diff
@@ -1,4 +1,5 @@
 build/
+out/
```
esotrope: what is this for btw?

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
bgsavage: Declaring the type and instantiating the type do the same thing.  IMO, it only makes it harder to read... 
`let observable: observable = new Observable` <-- Ok I get it, it's an observable.
Even in C# I like to say `var observable = new Observable()`.  One of my pet peeves in Java is that you can't do that.

.gitignore
bgsavage: Ever since I upgraded IntelliJ I've been getting a bunch of build files in the my out folder.

esotrope: must be that cool new feature they added?

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
esotrope: I like that idea.

Only hole i can see is if we do something like:
```let thing = something.thing``` 
would you advocate adding the type here?

PR #463, opened 2017-09-20
Generate a single-student report with unspecified individual/group pe…

wtritch:
…rmissions

This PR adds support for handling a single-student report request submitted via a StudentExamReportRequest.
Most of the changes revolve around security concerns.  Previously we were able to assume a single PermissionScope was enough security information to find the students and their exams for reporting.  Unfortunately, when submitting a single-student report request, the system does not know whether to submit the INDIVIDUAL_PII_READ or GROUP_PII_READ permissions.  Additionally, if using GROUP_PII_READ permissions, we need to ensure that the user has been granted rights to a group that the student belongs to.

The end result is that we have gone from 2 SQL query implementations each for fetching IAB reports, ICA reports, and students to 3 SQL query implementations each for fetching the data.

NOTE: This PR should not be merged directly to "develop"  We should first consolidate this PR with @esotrope's #457 PR to avoid any down-time due to mis-matched implementations between the webapp and report processor.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
@@ -4,8 +4,10 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import org.opentestsystem.rdw.reporting.common.report.ExamQueryParams;
+import org.opentestsystem.rdw.reporting.common.report.UserPermissions;
 import org.opentestsystem.rdw.reporting.common.security.Permission;
 import org.opentestsystem.rdw.reporting.common.security.PermissionScope;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -104,6 +106,7 @@ public class QueryUtils {
         params.put("grade_id", examQueryParams.getGradeId());
         params.put("group_id", examQueryParams.getGroupId());
         params.put("school_year", examQueryParams.getSchoolYear());
+        params.put("student_id", examQueryParams.getStudentId());
         return params;
     }
 
@@ -140,6 +143,39 @@ public class QueryUtils {
             return null;
         }
         return ImmutableList.copyOf(Splitter.on(delimiter).split(delimitedValues));
+    }
+
+    /**
+     * Retrieve the correct security parameters for the given user permissions
+     * and exam query permission type.
+     *
+     * @param userPermissions       The requesting user's permissions
+     * @param queryPermissionType   The query permission type
+     * @return The appropriate security parameters
+     */
+    public static Map<String, Object> getSecurityParameters(final UserPermissions userPermissions,
+                                                            final ExamQueryParams.ExamQueryPermissionType queryPermissionType) {
+        switch (queryPermissionType) {
+            case Unknown:
+                return new MapSqlParameterSource()
+                        .addValues(getSecurityParameters(userPermissions.getPermissionsById()))
+                        .addValue("group_ids", nullOrEmptyToDefault(userPermissions.getGroupsById().keySet(), UNMATCHABLE_IDS))
+                        .getValues();
+            case Individual:
+                final Permission individualPiiPermission = userPermissions.getPermissionsById().get(IndividualPiiRead);
+                final PermissionScope individualScope = individualPiiPermission == null ? PermissionScope.EMPTY : individualPiiPermission.getScope();
+                return new MapSqlParameterSource()
+                        .addValues(getSecurityParameters(individualScope))
+                        .getValues();
+            case Group:
+                final Permission groupPiiPermission = userPermissions.getPermissionsById().get(GroupPiiRead);
+                final PermissionScope groupScope = groupPiiPermission == null ? PermissionScope.EMPTY : groupPiiPermission.getScope();
+                return new MapSqlParameterSource()
+                        .addValues(getSecurityParameters(groupScope))
+                        .getValues();
+            default:
+                throw new IllegalArgumentException("Unknown query permission type: " + queryPermissionType);
+        }
     }
 
 }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamQueryParams.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamQueryParams.java
@@ -13,6 +13,8 @@ public class ExamQueryParams {
     private Integer gradeId;
     private Long groupId;
     private Integer schoolYear;
+    private Long studentId;
+    private ExamQueryPermissionType queryPermissionType;
 
     /**
      * @return The exam subject
@@ -56,6 +58,20 @@ public class ExamQueryParams {
         return schoolYear;
     }
 
+    /**
+     * @return The target student's id
+     */
+    public Long getStudentId() {
+        return studentId;
+    }
+
+    /**
+     * @return The exam query type
+     */
+    public ExamQueryPermissionType getQueryPermissionType() {
+        return queryPermissionType;
+    }
+
     public static Builder builder() {
         return new Builder();
     }
@@ -67,6 +83,8 @@ public class ExamQueryParams {
         private Integer gradeId;
         private Long groupId;
         private Integer schoolYear;
+        private Long studentId;
+        private ExamQueryPermissionType queryPermissionType;
 
         public ExamQueryParams build() {
             final ExamQueryParams examQueryParams = new ExamQueryParams();
@@ -76,6 +94,8 @@ public class ExamQueryParams {
             examQueryParams.gradeId = gradeId;
             examQueryParams.groupId = groupId;
             examQueryParams.schoolYear = schoolYear;
+            examQueryParams.studentId = studentId;
+            examQueryParams.queryPermissionType = queryPermissionType;
             return examQueryParams;
         }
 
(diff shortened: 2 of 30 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
```diff
@@ -142,4 +145,37 @@ private QueryUtils() {
...
+            case Group:
+                final Permission groupPiiPermission = userPermissions.getPermissionsById().get(GroupPiiRead);
+                final PermissionScope groupScope = groupPiiPermission == null ? PermissionScope.EMPTY : groupPiiPermission.getScope();
+                return new MapSqlParameterSource()
+                        .addValues(getSecurityParameters(groupScope))
+                        .getValues();
+            default:
+                throw new IllegalArgumentException("Unknown query permission type: " + queryPermissionType);
+        }
+    }
```
wtritch: This is used to create security parameters for 3 repository implementations.  I decided to move it here for easy re-use.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamQueryParams.java
```diff
@@ -108,5 +128,21 @@ public Builder schoolYear(final Integer schoolYear) {
...
+        public Builder queryPermissionType(final ExamQueryPermissionType queryPermissionType) {
+            this.queryPermissionType = queryPermissionType;
+            return this;
+        }
+    }
+
+    public enum ExamQueryPermissionType {
+        Unknown,
+        Individual,
+        Group
```
wtritch: When I first started out, I had Student, SchoolGrade, and Group as the types.  Then I realized that I was binding the implementation to the usage.  The true difference between the three types is how we do security filtering based upon permissions.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/UserPermissions.java
```diff
@@ -40,8 +40,8 @@ public static Builder builder() {
 
         public UserPermissions build() {
             final UserPermissions permissions = new UserPermissions();
-            permissions.groupsById = ImmutableMap.copyOf(groupsById);
-            permissions.permissionsById = ImmutableMap.copyOf(permissionsById);
+            permissions.groupsById = groupsById == null ? Collections.emptyMap() : ImmutableMap.copyOf(groupsById);
```
esotrope: Why the use of Collections alongside Immutables? I would normally go for ```ImmutableMap.of()``` here

wtritch: Good point.  My go-to shared empty-map instance tends to be the Collections.emptyMap() but we should be consistent here.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
```diff
@@ -175,4 +177,17 @@ public IabReport findOneByStudentIdAndSchoolYear(
         return reportsByStudent;
     }
 
+    private String getExamFilterQuery(final ExamQueryParams.ExamQueryPermissionType queryType) {
```
esotrope: Haven't looked through all the code yet but wondering - is there a way that the query can just be in the permission type? that way we could just do ```querySecurityType.findAllIabs```

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
```diff
@@ -104,18 +107,19 @@ public IabReport findOneByStudentIdAndSchoolYear(
...
-            @NotNull final PermissionScope permissionScope,
+            @NotNull final UserPermissions userPermissions,
             final Collection<Long> studentIds,
             final ExamQueryParams examQueryParams) {
 
         final Map<Long, Map<Subject, IabReport.Builder>> studentReportBuilders = new HashMap<>();
         final Map<Long, Map<Subject, Multimap<Assessment, Exam>>> studentReportAssessments = new HashMap<>();
 
-        final String query = examQueryParams.getGroupId() == null ? findAllIabsByStudentIdsAndQueryParams : findAllIabsByStudentIdsAndQueryParamsWithGroup;
+        final String query = getExamFilterQuery(examQueryParams.getQueryPermissionType());
```
wtritch: When we only had 2 options, a ternary made sense here.  Now that we have 3 I've changed to a switch impl.  The general process for alternating is still:
1. pick the query
2. pick the security parameters

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ReportRequestMessage.java
```diff
@@ -43,13 +52,15 @@ protected ReportRequestMessage createInstance() {
 
         private long reportId;
         private PermissionScope permissionScope;
```
wtritch: There will be some cleanup work once the webapp and report processor changes are merged to go back through and remove the PermissionScope references.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/requesthandler/SchoolGradeExamQueryParamsParser.java
```diff
@@ -26,6 +28,7 @@ public ExamQueryParams parseQueryParams(final AbstractExamReportRequest reportRe
                 .assessmentType(request.getAssessmentType() == null ? null : request.getAssessmentType().getAssessmentType())
                 .gradeId(request.getGradeId())
                 .subject(request.getSubject())
+                .queryPermissionType(Individual)
```
esotrope: I think we could use the strategy pattern here with three @Component impls each holding the queries for the respective security model.

wtritch: I'm not sure I like the idea of query SQL escaping the Repository layer...

esotrope: Hm. Yeah it shouldn't. The request shouldn't tell the repo how to do its job. Shame though because it has potential to make the code simpler.

wtritch: Yeah, there are some conflicting design ideals here:
1. keep SQL at the repository layer
2. thou shalt not switch/if-chain (there should not be an all-knowing overlord aware of all possible implementations)
3. behavior should be injected  (this applies to both creating three SqlProvider @components that know about all 3 repositories as well as having 3 repositories that know about all three sql options, but the thought of having 9 components makes me tired)

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamQueryParams.java
```diff
@@ -108,5 +128,21 @@ public Builder schoolYear(final Integer schoolYear) {
...
+            return this;
+        }
+
+        public Builder queryPermissionType(final ExamQueryPermissionType queryPermissionType) {
+            this.queryPermissionType = queryPermissionType;
+            return this;
+        }
+    }
+
+    public enum ExamQueryPermissionType {
```
malaffoon: nit: perhaps call it just `PermissionType`. Otherwise it will be `ExamQueryParams.ExamQueryPermissionType`.

wtritch: Good call, done.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
bgsavage: The individual case and Group case look identical, with the only difference being the constant passed in.  Is there an opportunity for code reuse rather than code duplication here?

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
```diff
@@ -175,4 +177,17 @@ public IabReport findOneByStudentIdAndSchoolYear(
         return reportsByStudent;
     }
 
+    private String getExamFilterQuery(final ExamQueryParams.ExamQueryPermissionType queryType) {
+        switch (queryType) {
+            case Unknown:
```
bgsavage: Switch / cases have always been kind of a code smell for me; doubly so when you're switching on the same cases in multiple spots.  There might be an opportunity here to abstract some interface with an implementation for each of the types.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/requesthandler/SchoolGradeExamQueryParamsParser.java
bgsavage: I would greatly favor having 9 components that all follow SRP over one component that has 9 responsibilities.

I like @esotrope 's original thought.  Maybe we can have a abstracted repo layer that delegates down to lower repo layers based on your conditions?  As far as the service is concerned, there's still a single repo to call, and is unaware of the underlying implementation of separate repo for each permission type?

wtritch: One component doesn't have 9 responsibilities, each repository knows the 3 SQL queries that it uses.  I feel like abstracting that out into 3 SqlProvider implementations per repository that "accept" based upon the PermissionsType would actually decrease readability and maintainability rather than increase it.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepository.java
wtritch: See comment above

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
wtritch: hmm... we have 3 distinct options, 2 of which vary based upon a constant.  I could abstract out the two, but I don't think it'd be worth the extra level of hierarchy.  It's always a balancing act. =/

PR #464, opened 2017-09-21
Refactor classes out of common that are only used by the report-proce…

wtritch:
…ssor

This PR is just shifting classes around to avoid bloating rdw-common with classes that are only used by the report-processor module.


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/status/WkhtmltopdfStatusIndicator.java
```diff
@@ -1,12 +1,12 @@
-package org.opentestsystem.rdw.reporting.status;
+package org.opentestsystem.rdw.reporting.processor.status;
```
esotrope: do we need to do anything special for our health check docs/config to find this endpoint in the report processor app vs the webapp?

common/src/main/java/org/opentestsystem/rdw/reporting/common/CommonConfiguration.java
```diff
@@ -1,17 +0,0 @@
...
-import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Import;
-
-@Configuration
-@Import({
-        DataSourceConfiguration.class,
-        TranslationConfiguration.class,
-        ReportConfiguration.class
-})
-public class CommonConfiguration {
```
esotrope: 👍 been meaning to do this

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/status/WkhtmltopdfStatusIndicator.java
malaffoon: Shouldn't ... it scans for appropriate beans on startup. So, as long as it gets instantiated it should work.

wtritch: double-checked, it's being instantiated

PR #465, opened 2017-09-21
Remove now-unused PermissionScope property from message models

wtritch:
Final cleanup of the report-processor.  This removes the vestigial PermissionScope from the DTO model instances.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
@@ -2,25 +2,20 @@ package org.opentestsystem.rdw.reporting.report;
 
 import com.google.common.collect.ImmutableSet;
 import org.opentestsystem.rdw.reporting.common.model.Report;
-import org.opentestsystem.rdw.reporting.common.report.AbstractBatchExamReportRequest;
 import org.opentestsystem.rdw.reporting.common.report.AbstractExamReportRequest;
 import org.opentestsystem.rdw.reporting.common.report.BatchExamReportRequestHolder;
 import org.opentestsystem.rdw.reporting.common.report.GroupExamReportRequest;
 import org.opentestsystem.rdw.reporting.common.report.PrintOptions;
 import org.opentestsystem.rdw.reporting.common.report.SchoolGradeExamReportRequest;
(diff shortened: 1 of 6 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/report/DefaultExamReportService.java
```diff
@@ -34,19 +29,17 @@
 
     @Override
     public Report createReport(@NotNull final User user, @NotNull final StudentExamReportRequest request) {
-        return createReport(user, request, null);
+        return createReportInternal(user, request);
     }
```
wtritch: I debated removing the three methods and just making the abstract parameter method public... thoughts?

esotrope: Yeah Id go with that. The method bodies are all the same now.

wtritch: hm... except I like the different authorities on the interface based upon the request type...
```
/**
     * Creates an individual student exam report
     *
     * @param user    the user requesting the report
     * @param request settings which shape the content of the report
     * @return the created report handle
     */
    @PreAuthorize("hasAnyAuthority('PERM_INDIVIDUAL_PII_READ', 'PERM_GROUP_PII_READ')")
    Report createReport(@NotNull User user, @NotNull StudentExamReportRequest request);

    /**
     * Creates a school grade scoped student exam report
     *
     * @param user    the user requesting the report
     * @param request settings which shape the content of the report
     * @return the created report handle
     */
    @PreAuthorize("hasAuthority('PERM_INDIVIDUAL_PII_READ')")
    Report createReport(@NotNull User user, @NotNull SchoolGradeExamReportRequest request);

    /**
     * Creates a group scoped student exam report
     *
     * @param user    the user requesting the report
     * @param request settings which shape the content of the report
     * @return the created report handle
     */
    @PreAuthorize("hasAuthority('PERM_GROUP_PII_READ') and #request.groupGrant != null")
    Report createReport(@NotNull User user, @NotNull GroupExamReportRequest request);
```

esotrope: Yeah that's a deal breaker we'll want to keep that.

PR #470, opened 2017-09-26
Added a Zuul filter which adds content-type to IRIS resource request

bgsavage:
Spring Security, by default, automatically adds the `X-Content-Type-Options: nosniff` header to all responses.  Most browsers have an implementation for this header that apply only for scripts and styles.  IE11's implementation applies to images as well.  Since the Iris Renderer does not return the `Content-Type` in it's resource responses, IE11 is then instructed to block the image content from rendering.  

This Zuul filter attempts to sniff and set the content type if a file name is present in the request params.

Typical observed requests have the format of `resources.axd?path=XXX&file=XXXX` but we will need to do more thorough testing to validate there is no other use cases here.

Tests are still forthcoming and will be added to this same PR.


webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+public class IrisAwareContentTypeResponseFilter extends ZuulFilter {
+
+    @Override
+    public String filterType() {
+        return "post";
+    }
+
+    @Override
+    public int filterOrder() {
+        return 999;
```
malaffoon: Perhaps make the magic number a class const.

Looking in the hierarchy of ZuulFilters, the order is pretty much magic numbers all over the place isn't it?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public boolean shouldFilter() {
+        RequestContext context = getCurrentContext();
+        return isRequestUrlAnIrisResourceUrl(context) && isResponseMissingContentType(context);
+    }
+
+    @Override
+    public Object run() {
+        RequestContext context = getCurrentContext();
+        String[] fileNames = context.getRequest().getParameterValues("file");
```
malaffoon: nit: `final`

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
bgsavage: It's pretty arbitrary at this point since it's our only custom filter.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
bgsavage: One of these days I will `final`ly do this as second nature...

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+
+        if(fileNames != null && fileNames.length > 0) {
```
malaffoon: Is the content always going to be returned as a file? I recall that tomcat will inline it if the content is "small" (default configuration defines small as <2k i think but it is configurable).

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+
+        return null;
+    }
+
+    private boolean isRequestUrlAnIrisResourceUrl(RequestContext context) {
+        return context.getRequest().getRequestURI().toLowerCase().contains("/iris/pages/api/resources.axd");
+    }
+
+    private boolean isResponseMissingContentType(RequestContext context) {
+        return context.getResponse().getHeaderNames().stream().noneMatch(x -> Objects.equals(x.toLowerCase(), "content-type"));
```
wtritch: I'd use the constant HttpHeaders.CONTENT_TYPE here

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+        if(fileNames != null && fileNames.length > 0) {
+            HttpServletResponse response = context.getResponse();
+            response.setHeader("Content-Type", URLConnection.guessContentTypeFromName(fileNames[0]));
+        }
+
+        return null;
+    }
+
+    private boolean isRequestUrlAnIrisResourceUrl(RequestContext context) {
+        return context.getRequest().getRequestURI().toLowerCase().contains("/iris/pages/api/resources.axd");
```
wtritch: I wonder if we want to include "/iris" as part of our tests... theoretically we could map any prefix to iris.  It's a spring configuration property.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
bgsavage: This is a query parameter in the request that I'm looking for.  I'm not following?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
bgsavage: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
bgsavage: The pathing is how the `<img src` makes the call.  I'm not sure that this could be configured any differently?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
malaffoon: Okay, so the request from the client will always have the file form field set?

bgsavage: That's my observation and assumption.  We will need to vet this further in testing.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public Object run() {
+        RequestContext context = getCurrentContext();
+        String[] fileNames = context.getRequest().getParameterValues("file");
+
+        if(fileNames != null && fileNames.length > 0) {
+            HttpServletResponse response = context.getResponse();
+            response.setHeader("Content-Type", URLConnection.guessContentTypeFromName(fileNames[0]));
```
jtreuting: Same content-type constant @wtritch  mentioned here

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * IRiS does not set the content type for resources it loads.  This causes a problem in IE11 when Spring Security sets the
+ * X-Content-Type-Options: nosniff; header by default.  Rather than disabling this best practice header, this filter sets the
+ * content type based on the file name of the resource provided by IRIS.
+ */
+public class IrisAwareContentTypeResponseFilter extends ZuulFilter {
+
+    @Override
+    public String filterType() {
+        return "post";
```
jtreuting: Since there was mention of constants, i'm assuming there is a POST constant somewhere we can utilize as well

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
esotrope: Feel like a constructor param or settable prop would be more appropriate for this. In the case that a later filter needs to go before or after it it can be easily configured that way.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
esotrope: My guess is Will is right here. I think the paths IRIS makes are relative to the iframe's path. So if we change the iframe path to /blah/ it will show /blah/pages/api/... 
Would be safer to make this a configurable property anyway.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
esotrope: Does ```RequestContext``` expose a header name to value map? Would be preferable to lookup the header value by key instead of iterating over all the name value pairs.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/IrisAwareContentTypeResponseFilter.java
bgsavage: I'm still not following....

FYI, here's a sample image tag in the iris iframe:

```
<img id="item_900_graphics1" src="/iris/Pages/API/Resources.axd?path=VEzXUQXI3l0JBkNLRO4TZmlGTW%2Foe20m6D20MxBIL%2F3rVVS%2BT7UbJg9ve%2BWXcAEjDEjlipVqQ9NOnF5fHJtgcw%3D%3D&amp;file=item_900_v20_graphics1_png256.png" width="413" height="156" style="vertical-align:baseline;" lang="en-us" data-width="413" data-height="156">
```

Are you telling me the "/iris/" found in the source can be different?

esotrope: Yes. but only in the sense that if we change the iframe src /iris/ this /iris/ will need to change. If they can be bound to the same config prop it would be nice, but if not i think a comment will suffice.

PR #471, opened 2017-09-26
adjust queries to deal with district/school groups

malaffoon:
I believe this is all that needs to be done. The district/school groups don't seem to be needed in the front-end or report generation. Please speak up if that is wrong.

### DIFF
--- a/common/src/main/resources/common.sql.yml
+++ b/common/src/main/resources/common.sql.yml
@@ -52,7 +52,11 @@ sql:
         and a.subject_id=:subject_id
       and
       ( # This query can run from the context of both GROUP_PII_READ and INDIVIDUAL_PII_READ, so need to evaluate both.
-        (1 = :individual_statewide or s.district_id in (:individual_district_ids) or e.school_id in (:individual_school_ids))
+        (1 = :individual_statewide
+          or s.district_group_id in (:individual_district_group_ids)
+          or s.district_id in (:individual_district_ids)
+          or s.school_group_id in (:individual_school_group_ids)
+          or e.school_id in (:individual_school_ids))
         or (
           exists(
               select 1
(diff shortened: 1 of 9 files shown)


common/src/main/resources/common.sql.yml
```diff
@@ -52,7 +52,11 @@ sql:
         and a.subject_id=:subject_id
       and
       ( # This query can run from the context of both GROUP_PII_READ and INDIVIDUAL_PII_READ, so need to evaluate both.
-        (1 = :individual_statewide or s.district_id in (:individual_district_ids) or e.school_id in (:individual_school_ids))
+        (1 = :individual_statewide
+          or s.district_group_id in (:individual_district_group_ids)
+          or s.district_id in (:individual_district_ids)
+          or s.school_group_id in (:individual_school_group_ids)
+          or e.school_id in (:individual_school_ids))
```
malaffoon: @agorina any reason to think the order of the OR clauses matters a lot? Should i be ordered them with the most-likely-to-match first or anything like that?

agorina: I do not think so. 
If MySQL optimizes it, then it does not matter. And if it does not get optimized, then it probably does not matter either.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -90,7 +90,11 @@ sql:
         and (:grade_id is null or e.grade_id=:grade_id)
         and (:assessment_type is null or a.type_id=:assessment_type)
         and ( # This query can run from the context of both GROUP_PII_READ and INDIVIDUAL_PII_READ, so need to evaluate both.
-              (1 = :individual_statewide OR s.district_id IN (:individual_district_ids) OR e.school_id IN (:individual_school_ids))
+              (1 = :individual_statewide
+                or s.district_group_id in (:individual_district_group_ids)
+                or s.district_id in (:individual_district_ids)
+                or s.school_group_id in (:individual_school_group_ids)
+                or e.school_id in (:individual_school_ids))
```
bgsavage: Would love to see a new pattern here to help us reuse permission sql instead of manually finding all the places that need to be updated.

PR #472, opened 2017-09-26
Bump schema dependency version.

wtritch:
With the recent permission changes, we need to bump the schema dependency version that includes the new required tables.


bgsavage:
Has the sql that requires the schema changes been merged?

wtritch:
@bgsavage yessir

PR #473, opened 2017-09-27
Feature/api get schools

bgsavage:
NOTE:  I don't intend to merge this PR at this time, but I'd like to get eyes on my approach before I move too far forward.  I would be implementing a similar pattern for schoolGroups and districtGroups.  I don't want to merge this just yet as it will break the UI as I'm no longer returning district info (now deprecated) as part of the school model.  I'd like to remove schools from UserContext and return the data from these two new endpoints instead, which should shrink the payload of the data over all (since we are not repeating district data for each school row) and allow for the data to be requested in parallel to user context.

Let me know if you'd like to jump on a call to review this together.

Adding two new endpoints: 

`api/schools`
GET:
```
[{ 
   id: 1, 
   name: "Tern Finch Elem", 
   schoolGroupId: 0, 
   districtId: 1, 
   districtGroupId: 0
}...]
```

`api/districts`
GET:
```
[{
   id: 1, 
   name: "LA Unified District", 
   districtGroupId: 0
}...]
```


webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/School.java
```diff
@@ -1,21 +0,0 @@
-package org.opentestsystem.rdw.reporting.exam;
-
-/**
- * Represents a school
- */
-public class School extends Organization {
```
bgsavage: I thought the school and organization model is misplaced under the `exam` package so I moved to a broader `organization` package sitting at the root of `reporting`

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/JdbcOrganizationRepository.java
```diff
@@ -0,0 +1,62 @@
...
+                findAllSchoolsQuery,
+                getSecurityParameters(permissionScope),
+                (row, index) ->
+                        School
+                                .builder()
+                                .id(row.getLong("id"))
+                                .name(row.getString("name"))
+                                .districtId(row.getLong("district_id"))
+                                .schoolGroupId(row.getLong("school_group_id"))
+                                .districtGroupId(row.getLong("district_group_id"))
```
bgsavage: This was an existing call but added additional metadata for schoolGroupId and districtGroupId.  I removed the districtName.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/JdbcOrganizationRepository.java
```diff
@@ -0,0 +1,62 @@
...
+    public List<District> findAllDistricts(@NotNull final PermissionScope permissionScope) {
+        return template.query(
+                findAllSchoolsQuery,
+                getSecurityParameters(permissionScope),
+                (row, index) ->
+                        District
+                                .builder()
+                                .id(row.getLong("id"))
+                                .name(row.getString("name"))
+                                .districtGroupId(row.getLong("district_group_id"))
```
bgsavage: To get districtName, you must now come here, which will be more normalized.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationRepository.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Gets all of the schools accessible through the given permission scope
+     *
+     * @param permissionScope the scope of access
+     * @return all of the schools accessible through the given permission scope
+     */
+    List<School> findAllSchools(@NotNull PermissionScope permissionScope);
+
+    /**
+     * Gets all of the schools accessible through the given permission scope
```
bgsavage: nit: oops.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/School.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private long schoolGroupId;
+    private long districtId;
+    private long districtGroupId;
+
+    @Deprecated
+    public School(long id, String name, Organization district) {
+        super(id, name);
+        this.district = district;
+    }
```
bgsavage: This constructor is now deprecated and will be removing references to it.  I'm deprecating the district composition to optimize the payload size for the schools request.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/school/DefaultSchoolService.java
```diff
@@ -1,34 +0,0 @@
...
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Service;
-
-import javax.validation.constraints.NotNull;
-import java.util.Set;
-
-import static org.opentestsystem.rdw.reporting.common.security.ReportingRoles.IndividualPiiRead;
-
-@Service("schoolGradeDefaultSchoolService")
-class DefaultSchoolService implements SchoolService {
```
bgsavage: Not really deleted, rather, moved & renamed to DefaultOrganizationService.

webapp/src/main/resources/application.sql.yml
```diff
@@ -487,19 +487,34 @@ sql:
        and exists (select 1 from exam e where e.student_id = s.id)
 
   school:
-    findAll: >-
-      select s.name, s.id, d.id as district_id, d.name as district_name
-      from school s
-        join district d on s.district_id = d.id
-      where (1=:statewide or s.district_group_id in (:district_group_ids) or s.district_id in (:district_ids) or s.school_group_id in (:school_group_ids) or s.id in (:school_ids))
-
```
bgsavage: Moved the sql as well as not really related to only assessment anymore.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/school/DefaultSchoolService.java
esotrope: There are also some org repos in the security package that might benefit from moving to the org package

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/District.java
```diff
@@ -0,0 +1,46 @@
...
+            this.name = name;
+            return this;
+        }
+        public Builder districtGroupId(final long districtGroupId) {
+            this.districtGroupId = districtGroupId;
+            return this;
+        }
+
+        public District build() {
+            final District district = new District(this.id, this.name);
```
esotrope: If possible, we might benefit from just making all the props builder props instead of constructor.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/JdbcOrganizationRepository.java
malaffoon: But wasn't there a place where we show a list of schools with the district name included? So that UI code would change to fetch both these and then merge the results?

webapp/src/main/resources/application.sql.yml
```diff
@@ -487,19 +487,35 @@ sql:
...
+        from school s
+          join district d on s.district_id = d.id
+        where (1=:statewide or s.district_group_id in (:district_group_ids) or s.district_id in (:district_ids) or s.school_group_id in (:school_group_ids) or s.id in (:school_ids))
+
+    district:
+      findAll: >-
+        select distinct d.id,
+          d.name,
+          s.district_group_id
+        from district d
```
malaffoon: Minor concern that is probably not be a real-life issue. If a person is granted district-level permissions and there are no schools for that district, the district won't be returned from this query. I think it is fair to assert that we are school-based reporting system so that's not biggie.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/District.java
```diff
@@ -0,0 +1,46 @@
+package org.opentestsystem.rdw.reporting.organization.model;
+
+public class District extends Organization {
+    private long districtGroupId;
```
malaffoon: A thought. If we used `Long` here it could be left null if there is no value set in the database. And the marshaled (JSON) payload could omit the key, making the payload smaller. Since there aren't very many groups in the system, this could be a non-trivial payload size decrease, especially for schools.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/School.java
```diff
@@ -0,0 +1,100 @@
...
+ * Represents a school
+ */
+public class School extends Organization {
+
+    @Deprecated
+    private final Organization district;
+
+    private long schoolGroupId;
+    private long districtId;
+    private long districtGroupId;
```
malaffoon: see comment above about null values

malaffoon:
The UserContext having all the schools in it ... i know we discussed it and decided that was a good idea because we always needed the list of schools. Is that still true? We could make the UserContext a *LOT* smaller if removed the schools and/or just included the organizations with granted roles (although i'm not sure what purpose that would serve).  Just thinking "out loud".

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/JdbcOrganizationRepository.java
bgsavage: Yeah and this is where I struggle.  We only show district names when we show the typeahead which is when you have access to >= 25 schools.  So in most use cases when it's just a teacher that has access to only their school, then they won't make the call to api/districts.  But you're right, in the case of a statewide user, they will make two calls to api/schools and api/districts.  What do you think?

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/District.java
bgsavage: Yep I'll be updating this to `Long`.  Landon and I discussed this offline.

webapp/src/main/resources/application.sql.yml
bgsavage: I agree with the assertion, if there's no schools then that also means there is no exams.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/JdbcOrganizationRepository.java
malaffoon: Depends on how much i'm allowed to blow things up. It seems like there are two distinct situations (and code paths):
1. User with just a few schools. Put the schools in the UserContext (retrieved using code like you have in this PR).
2. User with lots of schools. Don't put anything in the UserContext except a flag saying there are too many schools in context. Then the type-ahead API end-point would be used and it would return fully annotated school info including district id/name.

But i think less-disruptive would be to have the UI fetch the districts when it goes into type-ahead mode and merge the data on the fly.

bgsavage: I like your thoughts.  Landon and I definitely have tossed around the idea of bringing back the server side typeahead.  

Right now in QA for statewide (13,000 schools) the api/users call is 1.3 MB .

After implementing nullable `Long`s and keeping schools as part of api/users, the payload is 814KB.  The secondary api/districts call that would be required for statewide is 84 KB.

1300 KB - (814KB + 84KB) = 402KB savings.

I'm tempted to just leave schools in the user context and make the call for districts only when the UI requires it and do a merge of district / school data on the client.

esotrope: I like the idea to default to the less disruptive change.

malaffoon: baby steps

bgsavage:
Ok, I've updated this PR with the UI changes and do intend to merge at this time.  Please take a look at the latest.

@malaffoon @esotrope @wtritch @jtreuting @agorina

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -0,0 +1,52 @@
...
+        schoolObserver.complete();
+      });
+
+    return observable;
+  }
+
+  getDistricts() {
+    return this.dataService
+      .get("/districts")
+      .map(apiModels => new Map<number,string>(apiModels.map(x => [x.id, x.name])));
```
bgsavage: Creates a map to look up districts by id to return their name.

bgsavage: I envision schoolGroups and districtGroups would look similar to this.

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -0,0 +1,52 @@
...
+        let districts =response[ 1 ];
+
+        let schools = user.schools
+          .map(school => {
+            school.districtName = districts.get(school.districtId);
+            return school;
+          });
+
+        this.schoolsLoaded = true;
+        this.schoolsWithDistricts = schools;
```
bgsavage: Saving the data to a "cache" local to this service so that the api call and map only has to be done once.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/School.java
```diff
@@ -0,0 +1,81 @@
+package org.opentestsystem.rdw.reporting.organization.model;
```
malaffoon: nit (well, more opinion than nit): i don't see the added value of the nested `model` package. These are part of the interface for the organization services so could live in the same package.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/School.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Represents a school
+ */
+public class School extends Organization {
+
+    private Long schoolGroupId;
+    private long districtId;
```
esotrope: This could also technically be null

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -0,0 +1,52 @@
+import { Injectable } from "@angular/core";
+import { UserService } from "../user/user.service";
+import { School } from "../user/model/school.model";
+import { User } from "../user/model/user.model";
+import { DataService } from "../shared/data/data.service";
+import { Observable } from "rxjs/Observable";
+import { Observer } from "rxjs/Observer";
+
+@Injectable()
+export class OrganizationService {
```
esotrope: Should we create an interface and then have a ```CachingOrganizationService``` impl here?

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  private schoolsWithDistricts: School[] = null;
+  private schoolsLoaded: boolean = false;
```
esotrope: instead of having this flag why not just have the original set of schools in case a later call needs it to merge the groups?

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
esotrope: interesting so the map constructor accepts a 2d array of kvp

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
```diff
@@ -19,14 +21,17 @@ export class SchoolSelectComponent implements OnInit {
...
+  constructor(private organizationService: OrganizationService) {
   }
 
   get showDropdown() {
-    return this.availableSchools && this.availableSchools.length < 25;
+    return isNullOrUndefined(this.availableSchools) || this.availableSchools.length < 25;
   }
 
   ngOnInit() {
+    if(isNullOrUndefined(this.availableSchools))
```
esotrope: will null be the "more" flag?

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
wtritch: Hm... that's pretty darn flat.  Could this instead be x=> [x.id, x] in case we ever want to track more information for a district than just it's name?  I'd hate to have to refactor from scalar to object in the future.

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
wtritch: I'd rather have something a little more explicit, personally.

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
bgsavage: I think in the interest of memory efficiency, I'd like to keep this as flat as I can for now and yes refactor should we need more data later.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/School.java
bgsavage: I really like model directories / packages because a lot of times the only reason I'm scanning directories is I'm looking for the public models that the api exposes.  I like to see all my models in a single place rather than mixed in with services, controllers, and repos.

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.ts
bgsavage: Nope we are still passing down available schools from user context at this point.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/School.java
esotrope: That is what intellij ```command + N``` is for : j 

I think we have been inconsistent since, but early on in the project we did make an effort to go with feature packages instead of spring's default /controller, /repository, /service, /model packages.
Whatever we choose let's try to be consistent package to package.

PR #474, opened 2017-09-28
Add api calls for schoolGroups and districtGroups

bgsavage:
`api/schoolGroups`
GET
```
[{ 
    id: 2, 
    name: "SchoolGroup2", 
    districtId: 1
}, ...]
```
`api/districtGroups`
GET
```
[{ 
    id: 1, 
    name: "DistrictGroup1"
}, ...]
```

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
@@ -28,6 +28,9 @@ class DefaultOrganizationService implements OrganizationService {
     @Override
     public Set<School> getSchools(@NotNull final User user) {
         final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
+
+        // This method is not PreAuthorized because it is pulled at login time.  Need to check
+        // permissions here.
         if (permission == null) {
             return ImmutableSet.of();
         }
@@ -38,30 +41,18 @@ class DefaultOrganizationService implements OrganizationService {
     @Override
     public Set<SchoolGroup> getSchoolGroups(@NotNull final User user) {
         final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
-        if (permission == null) {
-            return ImmutableSet.of();
-        }
-
         return ImmutableSet.copyOf(repository.findAllSchoolGroups(permission.getScope()));
     }
 
     @Override
     public Set<District> getDistricts(@NotNull final User user) {
         final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
-        if (permission == null) {
-            return ImmutableSet.of();
-        }
-
         return ImmutableSet.copyOf(repository.findAllDistricts(permission.getScope()));
     }
 
     @Override
     public Set<DistrictGroup> getDistrictGroups(@NotNull final User user) {
         final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
-        if (permission == null) {
-            return ImmutableSet.of();
-        }
-
         return ImmutableSet.copyOf(repository.findAllDistrictGroups(permission.getScope()));
     }
 
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
@@ -2,7 +2,9 @@ package org.opentestsystem.rdw.reporting.organization;
 
 import com.google.common.collect.ImmutableSet;
 import org.opentestsystem.rdw.reporting.organization.model.District;
+import org.opentestsystem.rdw.reporting.organization.model.DistrictGroup;
 import org.opentestsystem.rdw.reporting.organization.model.School;
+import org.opentestsystem.rdw.reporting.organization.model.SchoolGroup;
 import org.opentestsystem.rdw.reporting.security.User;
 import org.opentestsystem.rdw.security.Permission;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -34,6 +36,16 @@ class DefaultOrganizationService implements OrganizationService {
     }
 
     @Override
+    public Set<SchoolGroup> getSchoolGroups(@NotNull final User user) {
+        final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
+        if (permission == null) {
+            return ImmutableSet.of();
+        }
+
+        return ImmutableSet.copyOf(repository.findAllSchoolGroups(permission.getScope()));
+    }
+
+    @Override
     public Set<District> getDistricts(@NotNull final User user) {
         final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
         if (permission == null) {
@@ -41,6 +53,16 @@ class DefaultOrganizationService implements OrganizationService {
         }
 
         return ImmutableSet.copyOf(repository.findAllDistricts(permission.getScope()));
+    }
+
+    @Override
+    public Set<DistrictGroup> getDistrictGroups(@NotNull final User user) {
+        final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
+        if (permission == null) {
+            return ImmutableSet.of();
+        }
+
+        return ImmutableSet.copyOf(repository.findAllDistrictGroups(permission.getScope()));
     }
 
 }
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
@@ -1,7 +1,9 @@
 package org.opentestsystem.rdw.reporting.organization;
 
 import org.opentestsystem.rdw.reporting.organization.model.District;
+import org.opentestsystem.rdw.reporting.organization.model.DistrictGroup;
 import org.opentestsystem.rdw.reporting.organization.model.School;
+import org.opentestsystem.rdw.reporting.organization.model.SchoolGroup;
 import org.opentestsystem.rdw.reporting.security.User;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.annotation.AuthenticationPrincipal;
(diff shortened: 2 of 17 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/DistrictGroup.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.reporting.organization.model;
+
+/**
+ * Represents a group of districts.
+ */
+public class DistrictGroup extends Organization {
+    public DistrictGroup(final long id, final String name) {
+        super(id, name);
+    }
+}
```
bgsavage: Not extending anything here but for consistency seems right for getDistrictGroup to return `DistrictGroup` instead of `Organization`.

webapp/src/main/resources/application.sql.yml
```diff
@@ -507,6 +507,16 @@ sql:
...
         where (1=:statewide or s.district_group_id in (:district_group_ids) or s.district_id in (:district_ids) or s.school_group_id in (:school_group_ids) or s.id in (:school_ids))
 
+    schoolGroup:
+      findAll: >-
+        select distinct sg.id,
+          sg.name,
+          s.district_group_id,
+          s.district_id
+        from school_group sg
+          inner join school s on s.school_group_id = sg.id
```
bgsavage: Inner join to return only school groups that have accessible schools.

webapp/src/main/resources/application.sql.yml
```diff
@@ -516,6 +526,14 @@ sql:
           join school s on s.district_id = d.id
```
bgsavage: Hm, maybe I need an inner join here as well.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
```diff
@@ -26,9 +28,19 @@ public OrganizationController(@NotNull final OrganizationService service) {
         return service.getSchools(user);
     }
 
+    @GetMapping("/api/schoolGroups")
```
jtreuting: my initial reaction is that this might be confusing with the other groups concept we have.  does groupOfSchools and groupOfDistricts make more sense?

webapp/src/main/resources/application.sql.yml
bgsavage: Oh my bad, inner join is implied when it's not qualified.  I'll remove.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
wtritch: We'll always want all of these?  Or should we have an optional ancestry filter?

wtritch: Hmmm... I wonder if these all belong under an "/organizations/**" parent?

bgsavage: I strongly considered the `/organizations/` prefix and think I'll add that.  Thanks!

RE: groupOfSchools, I don't follow how changing the possession order helps clarify usage and also is slightly more verbose.

bgsavage: @esotrope  thoughts?

jtreuting: i like the organizations part.  When I read schoolGroups, I thought about the groups that are created for a school that represent a classroom.  That is why i was thinking changing it might help clarify.  Maybe putting it under organizations clears it up.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
```diff
@@ -43,4 +55,14 @@ public DefaultOrganizationService(@NotNull final OrganizationRepository reposito
...
     }
 
+    @Override
+    public Set<DistrictGroup> getDistrictGroups(@NotNull final User user) {
+        final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
+        if (permission == null) {
+            return ImmutableSet.of();
+        }
+
+        return ImmutableSet.copyOf(repository.findAllDistrictGroups(permission.getScope()));
```
esotrope: could create an internal method to remove the redundancy here.
```java
private <T> Set<T> getInternal(User user, Function<PermissionScope, T> fn) {
  final Permission permission = user.getPermissionsById().get(IndividualPiiRead);
  return permission == null ? of() : copyOf(fn.apply(permission.getScope()));
}
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationService.java
```diff
@@ -21,11 +23,27 @@
     Set<School> getSchools(@NotNull User user);
 
     /**
+     * Gets all the school groups the user has access to
+     *
+     * @param user the user accessing the schools
+     * @return all the school groups the user has access to
+     */
+    Set<SchoolGroup> getSchoolGroups(@NotNull User user);
```
esotrope: Each method or the class def line  needs ```@PreAuthorized``` annotation

webapp/src/test/resources/org/opentestsystem/rdw/reporting/organization/schools.sql
```diff
@@ -1,9 +1,22 @@
...
   (-10, 'd1', 'district1'),
-  (-20, 'd2', 'district2');
+  (-20, 'd2', 'district2'),
+  (-30, 'd3', 'district3'),
+  (-40, 'd4', '=-=-=-= No schools are under this district so should not return. =-=-=-=');
+
+insert into school_group(id, natural_id, name) values
+  (-10, 'sg1', 'schoolGroup1'),
+  (-20, 'sg2', 'schoolGroup2'),
+  (-30, 'sg3', '=-=-=-= No schools are under this school group so should not return. =-=-=-=');
```
esotrope: not sure i'd put test expectations in the data setup

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/SchoolGroup.java
```diff
@@ -0,0 +1,68 @@
+package org.opentestsystem.rdw.reporting.organization.model;
+
+/**
+ * Represents a group of schools.
+ */
+public class SchoolGroup extends Organization {
+    private long districtId;
```
esotrope: i still think we should go ahead and support null here (e.g. use ```Long```). There can technically be schools that just belong to the state.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/District.java
```diff
@@ -1,9 +1,12 @@
 package org.opentestsystem.rdw.reporting.organization.model;
 
+/**
+ * Represents a school district
+ */
 public class District extends Organization {
     private Long districtGroupId;
 
-    public District(final long id, final String name) {
+    District(final long id, final String name) {
```
esotrope: So other packages can know what a district is but not make one im guessing?

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/SchoolGroup.java
bgsavage: In our schema, `school.district_id` is `not null` so I don't think we support districtId being null.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/District.java
bgsavage: I thought this was our builder pattern best practice, that you have to go through the builder to construct our objects.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationService.java
bgsavage: Good idea.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
bgsavage: If I use @PreAuthorize, then this null check is no longer needed, right?

esotrope: Ah that's right. And i think the user context doesn't want an exception thrown if the user has no schools. It is fine for now since we return empty when they don't have permission but we may want to just add a try/catch to the special case which is the user context generation

bgsavage: I'll `@PreAuthorize` all the methods but the schools one for UserContext

esotrope: 👍 later we can make it the same if we separate out schools to its own endpoint

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
esotrope: not sure if follow the implications.
/schools would return all ancestry or no ancestry? 

Right now the UI needs one of these datasets to get its job done:

1. Flat school with ancestry (has redundant names)
```javascript
{id, name, groupId, groupName, districtId, districtName, districtGroupId, districtGroupName}
```
2. Schools and ancestors (what we are doing but with 4 endpoints)
```
{
  schools: [{id, name, groupId, districtId, districtGroupId}]
  schoolGroups: [{id, name}]
  districts...
}

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/model/District.java
malaffoon: If a builder is used, and nobody extends the class then this should be private.

PR #476, opened 2017-10-02
Feature/custom export form submission

esotrope:
Adds school year to filter and form submit button.

screen shot 2017-10-02 at 4 47 41 pm

screen shot 2017-10-02 at 4 47 59 pm


webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
```diff
@@ -74,6 +89,14 @@ export class CustomExportComponent implements OnInit {
     );
 
     this.selected = [];
+
+    this.filterOptionService.getExamFilterOptions()
```
esotrope: This page only needs to school years from this service, wonder if it warrants breaking a school year service out

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
```diff
@@ -143,4 +178,64 @@ export class CustomExportComponent implements OnInit {
     this.selectedAll ? this.removeAll() : this.addAll();
   }
 
+  submit(): void {
+    let request = this.createExportRequest(this._schoolYear, this._selected, this._organizations);
+    this.service.createExport(request);
+  }
+
+  private createExportRequest(schoolYear: number, schools: Organization[], organizations: UserOrganizations): CustomExportRequest {
```
esotrope: This method is the meat of the change.
Basically the idea here is to make a request for 1 district ID instead of 1000 school IDs for LA and other districts / school groups with large school counts.

webapp/src/main/webapp/src/app/custom-export/custom-export.service.ts
```diff
@@ -0,0 +1,46 @@
...
+      responseType: ResponseContentType.Blob
+    }).catch(ResponseUtils.throwError);
+  }
+
+}
+
+/**
+ * Represents custom export request and holds different exam result filter options
+ */
+export interface CustomExportRequest {
```
esotrope: @bgsavage the request body format

webapp/src/main/resources/static/assets/i18n/en.json
```diff
@@ -0,0 +1,729 @@
+{
```
wtritch: Was this supposed to be checked in?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
```diff
@@ -45,13 +48,25 @@ export class CustomExportComponent implements OnInit {
...
+
+  /**
    * The sort order of the organizations
    */
   private _comparator = (a: Organization, b: Organization) => a.name && b.name ? a.name.localeCompare(b.name) : 0;
 
   constructor(private route: ActivatedRoute,
               private translate: TranslateService,
-              private mapper: OrganizationMapper) {
+              private service: CustomExportService,
```
wtritch: hmmm... is there a better identifier than "Custom" ?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
wtritch: Do we want to get clever with positive and negative payloads?  (District X, except schools A and D?)

webapp/src/main/webapp/src/app/custom-export/organization/organization.mapper.ts
```diff
@@ -64,7 +66,15 @@ export class OrganizationMapper {
     return options;
   }
 
-  organizationTree(schools: Organization[], organizations: UserOrganizations): Tree<Organization> {
+  /**
+   * Creates organization hierarchy for the given schools and master set of organizations.
+   * This method will add placeholder ancestor nodes to the school when then school group or district ID is undefined.
```
wtritch: nit then => the*

webapp/src/main/webapp/src/app/custom-export/organization/organization.mapper.ts
```diff
@@ -75,6 +85,27 @@ export class OrganizationMapper {
...
+    organizations.schools.forEach(school => {
+      let node = root;
+      if (!isUndefined(school.districtId)) {
+        node = node.getOrCreate(x => x.id === school.districtId, organizations.districtsById.get(school.districtId))
+      }
+      if (!isUndefined(school.schoolGroupId)) {
+        node = node.getOrCreate(x => x.id === school.schoolGroupId, organizations.schoolGroupsById.get(school.schoolGroupId))
+      }
+      node.create(school);
+    });
```
wtritch: Looks like we don't need to display district groups?

esotrope: Yeah took it out because we decided we don't want to be exporting whole district groups at a time. The maximum selection is one district for export.

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
esotrope: Probably. RegionScopedExamExport/DistrictExamExport* maybe?

webapp/src/main/resources/static/assets/i18n/en.json
esotrope: Nope thanks for catching

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
wtritch: OrganizationExamExport?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
wtritch: The more I think about this the more potential bugs I can come up with.  I like keeping the payload simple with only sending positive payloads.

webapp/src/main/webapp/src/app/custom-export/custom-export.component.html
```diff
@@ -3,33 +3,42 @@ <h3 class="blue mb-md">{{'labels.custom-export.heading' | translate}}</h3>
...
+            <span search-select-with-button
+                  [options]="options"
+                  (select)="add($event)"
+                  dropdownThreshold="10"
+                  placeholder="{{'prompt.select' | translate}}"
+                  searchPlaceholder="{{'labels.custom-export.form.organization.placeholder' | translate}}"
+                  buttonLabel="{{'labels.custom-export.form.organization.add' | translate}}">
+            </span>
+            <div class="ml-xs mt-xs">
+              <a *ngIf="selectAllDisabled"
```
bgsavage: So we hide select all if it's enabled?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.html
```diff
@@ -3,33 +3,42 @@ <h3 class="blue mb-md">{{'labels.custom-export.heading' | translate}}</h3>
...
+        <div class="col-md-2">
+          <div class="form-group">
+            <label for="select-year">{{'labels.select-year' | translate}}</label>
+            <select id="select-year" name="select-year" class="form-control" [(ngModel)]="schoolYear">
+              <option *ngFor="let year of schoolYears" [ngValue]="year">{{year | schoolYear}}</option>
+            </select>
+          </div>
+        </div>
+        <div class="col-md-4 text-right">
+          <button type="submit" class="btn btn-info aqua" [disabled]="selected.length == 0"><i class="fa fa-cloud-download"></i> {{'labels.custom-export.form.submit.label' | translate}}</button>
```
bgsavage: I've been wondering what this button does now that it queues up a report instead of triggering an immediate download.  I wonder if we should make this a modal now and this submit button closes the modal after the submit.  Otherwise you're still left on the page where you searched schools / districts leaving you potentially confused of "Did I submit that?  I'll just submit again I guess".

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
bgsavage: I like OrganizationExamExport

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
bgsavage: This caches so it saves the call later right?  Do you think it's possible we may want those other filter options later?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
```diff
@@ -120,6 +143,18 @@ export class CustomExportComponent implements OnInit {
     return this._organizations.districts.length < 2;
   }
 
+  get schoolYears(): number[] {
+    return this._schoolYears;
+  }
+
+  get schoolYear(): number {
+    return this._schoolYear;
```
bgsavage: What's the difference between schoolYear and schoolYears

wtritch: I assumed one was the list of options and the other was the selected value

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
bgsavage: I would prefer to put this method in a service to make it easier to unit test and in favor of keeping components light that depend on services for their business logic.

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
```diff
@@ -143,4 +178,64 @@ export class CustomExportComponent implements OnInit {
...
+    // all school IDs selected by the user
+    let selectedSchoolIds = new Set<number>(schools.map(school => school.id));
+
+    // organizational hierarchy of all organizations entitled to the user
+    let organizationTree = this.mapper.organizationTree(organizations);
+
+    // groups organization by type
+    let groupOrganizationByType = (organization: Organization) => {
+      switch (organization.type) {
+        case OrganizationType.District:
```
bgsavage: So we're mapping UserOrganizations to an organization tree and then breaking that tree out by type.  Is there a more straight forward approach here?  (i.e. go from UserOrganizations -> break out by type)

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
```diff
@@ -143,4 +178,64 @@ export class CustomExportComponent implements OnInit {
...
+          request.schoolGroups.push(organization.id);
+          break;
+        case OrganizationType.School:
+          request.schools.push(organization.id);
+          break;
+      }
+    };
+
+    // recursively checks if all descendants are selected
+    let allDescendantsSelected = (node: Tree<Organization>, selectedSchoolIds: Set<number>): boolean => {
```
bgsavage: couldn't this just be a method instead of redefining a lambda to a variable every time this runs?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
```diff
@@ -143,4 +178,64 @@ export class CustomExportComponent implements OnInit {
...
+      if (node.value.type === OrganizationType.School) {
+        return selectedSchoolIds.has(node.value.id);
+      }
+      return node.children.every(child => allDescendantsSelected(child, selectedSchoolIds));
```
bgsavage: Hm, I think I would have taken another approach where I added the parent organization by default when they first click it from the search, and then convert to a collection of children on first removal of a child.

webapp/src/main/webapp/src/app/custom-export/custom-export.service.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * School year to filter on
+   */
+  readonly schoolYear: number;
+
+  /**
+   * District IDs to filter on
+   */
+  readonly districts?: number[];
```
bgsavage: This should be districtIds

bgsavage: (And same for others)

webapp/src/main/webapp/src/app/custom-export/organization/organization.mapper.ts
```diff
@@ -75,6 +85,27 @@ export class OrganizationMapper {
   }
 
   /**
+   * Creates organization hierarchy with the given organizations.
+   *
+   * @param {UserOrganizations} organizations master set of organizations to create the tree from
+   * @returns {Tree<Organization>} organization hierarchy
+   */
+  organizationTree(organizations: UserOrganizations): Tree<Organization> {
```
bgsavage: method names should be verbs to describe the action the method performs, right?

webapp/src/main/webapp/src/styles.less
```diff
@@ -176,3 +176,7 @@ footer {
 .cursor-pointer {
   cursor: pointer;
 }
+
+custom-export form button[type='submit'] {
+  margin-top: 27px;
```
bgsavage: Nothing in the library to give us a large enough margin top?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.html
esotrope: Agreed. I think if we're not streaming an immediate response we should move this to be like the report modal.

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
esotrope: Will is right. Should I call it schoolYearOptions maybe?

bgsavage: schoolYearOptions and selectedSchoolYear just for clarification.

webapp/src/main/webapp/src/app/custom-export/custom-export.component.html
esotrope: let me check the method body. I think this is working ... somehow

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
esotrope: Tree helps do a top to bottom traversal of the hierarchy and understand containership relationships. If we just use the normalized orgs we can only do a bottom to top navigation of the structure. does that address your comment?

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
esotrope: Not like this runs very often. Is this really that memory abusive? I went with the internal lamda to keep the method cohesion. otherwise this little bit probably deserves its own class.

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
esotrope: Whenever possible I prefer not to maintain app state over time because it is generally much more bug prone. instead i like computing the current state at the moment you need it and not before.

webapp/src/main/webapp/src/styles.less
esotrope: @aaronsleeper said this was fine : P 
Other option that worked pretty well was putting a ```<br>``` above the button but that would be harder to find later and extract to the ui kit i thought

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
bgsavage: I would argue that it breaks cohesion because as you're reading the code line by line, you get a method definition and you don't see it's call until later.  Definitely prefer a service class here for SRP and easier unit testing.

webapp/src/main/webapp/src/app/custom-export/custom-export.component.ts
esotrope: Yeah it's cached so I'm not too worried about it.
Here's the model:
```
export class ExamFilterOptions {
  schoolYears: number[] = [];
  ethnicities: string[] = [];
  hasSummative: boolean = false;
  hasInterim: boolean = false;
}
```
Not sure I can imagine needing the others.

esotrope:
@wtritch @bgsavage The comments got buried by the file name changes.
Please re-review and make sure your comments got addressed - thanks!

webapp/src/main/webapp/src/app/organization-export/organization-export.service.ts
```diff
@@ -0,0 +1,157 @@
...
+   * School year to filter on
+   */
+  readonly schoolYear: number;
+
+}
+
+/**
+ * Organization IDs grouped by type
+ */
+interface GroupedOrganizationIds {
```
bgsavage: Would prefer to create these as in `.model.ts` files for consistency with other models, and to treat them as classes rather than interfaces.

bgsavage:
looks good to me too, would also like to see your spec files as well, are those still forthcoming?

webapp/src/main/webapp/src/app/organization-export/organization-export.service.ts
esotrope: I can change to *model for now. Later we should adopt the Angular CLI naming convention for class/interfaces to just have [name].ts though I think.

Not sure what the benefit would be to using class > interface in TS though, would like to know why you like that better.

bgsavage: I prefer to new up objects so you can easily see it's type and navigate to it with cmd+B rather than dynamic creation such as `{ }` , which isn't as obvious that we have the fields defined for it somewhere in an interface.  Does refactor -> rename work when you init objects with `{ }` ?

esotrope: Refactor works 👍  ```ctrl+click``` to jump to field declaration works.
Even if its a class you can dynamically instantiate this way.

I feel doing this the verbose way takes away some of the benefits of using TS  : /

PR #477, opened 2017-10-03
Bugfix/handle large saml payloads

wtritch:
This PR is against a "hotfix" branch off of "master."

The objective is to handle extremely large SAML Response objects where a User has been granted 400+ individual permissions.

The root problem was in the library-provided SAMLObject serialization implementation:
```
    private void writeObject(ObjectOutputStream out) throws IOException {
        try {
            if (serializedObject == null) {
                serializedObject = XMLHelper.nodeToString(SAMLUtil.marshallMessage(getObject()));
            }
            out.writeUTF((String) serializedObject);
        } catch (MessageEncodingException e) {
            log.error("Error serializing SAML object", e);
            throw new IOException("Error serializing SAML object: " + e.getMessage());
        }
    }
```

The call to `out.writeUTF((String) serializedObject);` throws a UTFDataFormatException if the serialized XML is longer than 64k.  This PR replaces the SAMLObject implementation with a LargeSamlObject implementation that writes the serialized payload as a String Object rather than a flat UTF entry, allowing for unlimited size.  
The other classes in this PR are required to convert the SAML framework to use our custom LargeSamlObject rather than default SAMLObject.

NOTE: Once this PR is approved, I'll be creating a nearly-identical PR against RDW_Admin.


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -173,6 +175,11 @@ public SAMLContextProvider contextProvider() throws ServletException {
         return contextProvider;
     }
 
+    @Bean
+    public SAMLMessageStorageFactory storageFactory() {
+        return new SamlHttpSessionStorageFactory();
+    }
```
wtritch: Registering our custom factory as a Bean tells Spring to inject our implementation into the SAMLContextProvider.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SamlHttpSessionStorage.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Session the storage operates on.
+     */
+    private final HttpSession session;
+
+    /**
+     * Internal storage for messages, corresponding to the object in session.
+     */
+    private Hashtable<String, LargeSamlObject<XMLObject>> internalMessages;
```
wtritch: This is a nearly-identical implementation to the library method.  The only change is to use LargeSamlObject instances rather than SAMLObject instances.

wtritch:
Many thanks to @malaffoon for diagnosing the root cause of the issue.

jtreuting:
That seemed more difficult than i would have hoped.  Let's make sure to pass this info along to IAT and TDS folks.

chdietz:
Nice job, and another reason to have some common code in this area.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/LargeSamlObject.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param out output stream
+     * @throws java.io.IOException error performing XMLObject serialization
+     */
+    private void writeObject(final ObjectOutputStream out) throws IOException {
+        try {
+            if (serializedObject == null) {
+                serializedObject = XMLHelper.nodeToString(SAMLUtil.marshallMessage(getObject()));
+            }
+            out.writeObject(serializedObject);
```
chdietz: why do you think the original implementation uses writeUTF and readUTF ?

wtritch: Because 64k ought to be enough for anybody.  And it leads to slightly smaller payloads without the additional object overhead, would be my guess.

chdietz: I was wondering if there was a security reason for it

wtritch: hmm... I doubt it.  If session data isn't secure we have larger issues.

PR #478, opened 2017-10-03
Feature/export exams csv

bgsavage:
This is a progressive merge as I work to not have a super large PR at the end of the week.  I'd like to get some iterative reviews to nudge me in the right direction should I be deviating too far off track.

So this is the "meat" of the export csv.  And if you're a vegetarian, then it's the "eggplant"?

This PR takes advantage of `select into outfile` and aurora `select into outfile s3` to export data directly out of the database without having to pass through the calling code at all.

This is all taking place in the report processor, I don't have any messaging hooked up yet, and I'm stubbing a user for testing purposes. 

This was quite a large deviation from my initial approach of downloading a csv on the fly directly from reporting app, so there might still be some artifacts from that original approach.

No IT / unit tests at this time.


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/StudentContexts.java
```diff
@@ -0,0 +1,24 @@
...
+        builder
+                .iep(row.getBoolean("iep"))
+                .lep(row.getBoolean("lep"))
+                .section504(row.getBoolean("section504"))
+                .economicDisadvantage(row.getBoolean("economic_disadvantage"))
+                .migrantStatus(getNullable(row, row.getBoolean("migrant_status")));
+
+        return builder;
+    }
+}
```
bgsavage: This is an artifact of a previous implementation, but at least sets us up for reusing this code elsewhere, so thought to might as well keep for now.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+
+            return new LocalExamExportService(repository);
+        }
+
+        if (S3ExamExportService.validRoot(archiveRoot)) {
+            if(!exportDestination.equals("S3"))
+                throw new IllegalArgumentException("Export destination must be S3 to export to " + archiveRoot);
+
+            return new S3ExamExportService(repository);
+        }
```
bgsavage: Similar to what we had for ArchiveService, this configuration uses the archive root to determine which ExamExmportService impl to serve up.  Notice the strict configuration requirements for exportDestination, more on that down further.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamExportQuery.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+public class ExamExportQuery {
+    private long schoolYear;
+    private Set<Long> schoolIds;
+    private Set<Long> schoolGroupIds;
+    private Set<Long> districtIds;
+    private Set<Long> districtGroupIds;
```
esotrope: we should probably take this out so as to prevent group of district exports.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/S3ExamExportService.java
```diff
@@ -0,0 +1,46 @@
...
+    private static final String S3SchemePrefix = "s3://";
+    private static final String IntoOutS3Format = "s3-%s://";
+    private static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH:mm:ss");
+    private static final String FileNameFormat = "export-exam-%s.csv";
+
+    @Value("${app.archive.root}")
+    private String archiveRoot;
+
+    @Value("${app.archive.cloud.aws.region.static}")
+    private String region;
```
bgsavage: For Select Into S3, format is `s3-region://bucket-name/file-prefix.manifest`.  

Docs are here: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Aurora.SaveIntoS3.html

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ExportController.java
```diff
@@ -0,0 +1,44 @@
...
+                .builder()
+                .permissionsById(
+                        ImmutableMap.of(IndividualPiiRead,
+                                new Permission(IndividualPiiRead,
+                                        PermissionScope.builder().statewide(true).build())
+                        )
+                )
+                .build();
+
+        return service.exportExam(mockPermissions, query);
```
bgsavage: Please ignore this controller for now, it's only for my own testing.  Eventually this method will be calling code that creates the report record in the db and fires (and forgets) off the spring message.  Right now it executes the export and makes the caller wait in real time.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportService.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.reporting.processor.service.export;
+
+import org.opentestsystem.rdw.reporting.common.report.UserPermissions;
+import org.opentestsystem.rdw.reporting.processor.model.ExamExportQuery;
+
+import javax.validation.constraints.NotNull;
+
+public interface ExamExportService {
+    String exportExam(@NotNull UserPermissions permissions, ExamExportQuery query);
```
esotrope: I think we'll need the pre-auth annotation with PERM_INDIVIDUAL_PII here

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/LocalExamExportService.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Autowired
+    LocalExamExportService(ExamRepository repository){
+        this.repository = repository;
+    }
+
+    @Override
+    public String exportExam(final UserPermissions permissions, final ExamExportQuery query) {
+        String uriLocation = archiveRoot.replace(FileSchemePrefix, "/");
+        String fileNamePrefix = String.format(FileNameFormat, dateFormat.format(new Date()));
```
esotrope: I think jdk default date formats are not thread safe unfortunately so we'll have to new it up every invocation or use another formatter.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -1,4 +1,136 @@
...
+        (case when exists (select 1 from student_ethnicity where student_id = st.id and ethnicity_code = 'NativeHawaiianOrOtherPacificIslander') then 'Yes' else 'No' end ),
+        (case when exists (select 1 from student_ethnicity where student_id = st.id and ethnicity_code = 'TwoOrMoreRaces') then 'Yes' else 'No' end ),
+        (case when exists (select 1 from student_ethnicity where student_id = st.id and ethnicity_code = 'Filipino') then 'Yes' else 'No' end ),
+
+        coalesce((select GROUP_CONCAT(a.code SEPARATOR '|')
+                  from exam_available_accommodation eaa
+                    join accommodation a on a.id = eaa.accommodation_id
+                  where eaa.exam_id = e.id), '')
+
+      into outfile ${app.export.destination} :export_location
```
bgsavage: This is that export destination I referenced earlier.  So in order to have a local / s3 implementation, we need some way to to set the `S3` aurora command dynamically when we're sending to S3.  When it's local file, it should just be `''`.  I'm looking for this configuration at startup so it will fail fast if it's not configured correctly. I  It was suggested I try this as a property, which works, but I'm open to other options as well.

report-processor/src/main/resources/application.yml
```diff
@@ -61,6 +61,9 @@ app:
   state:
     code: CA
 
+  export:
+    destination:
```
bgsavage: Blank for local file system.  S3 for S3.

bgsavage: And I"ll make sure to update other environments' configs as before completing this feature.

webapp/src/main/resources/application.sql.yml
```diff
@@ -569,3 +569,73 @@ sql:
         join item_common_core_standard iccs on iccs.common_core_standard_id = ccs.id
       where iccs.item_id = :item_id
 
+  report:
+    export:
+      exam: >-
```
bgsavage: Oh oops this should go away now.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -1,4 +1,136 @@
...
+        join student st on st.id = e.student_id
+        join asmt a on e.asmt_id = a.id
+        join subject su on su.id = a.subject_id
+        join asmt_type at on at.id = a.type_id
+        join school s on e.school_id = s.id
+        join district d on d.id = s.district_id
+      where e.school_year = :school_year and
+          (
+            1 = :individual_statewide
+            or s.district_group_id in (:individual_district_group_ids)
```
esotrope: same comment here

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/S3ExamExportService.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.beans.factory.annotation.Value;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import static org.opentestsystem.rdw.reporting.common.security.ReportingRoles.IndividualPiiRead;
+
+public class S3ExamExportService implements ExamExportService {
+
+    private final static String BasePath = "/REPORTS/USER/EXAM-EXPORT/";
```
esotrope: one of these things is not like the others : j

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -0,0 +1,69 @@
...
+                        .putAll(getSecurityParameters(permissionScope, "individual_"))
+                        .putAll(getQueryParameters(query))
+                        .put("export_location", uriLocation)
+                        .build(),
+                new PreparedStatementCallback<Boolean>() {
+                    @Override
+                    public Boolean doInPreparedStatement(PreparedStatement ps)
+                            throws SQLException, DataAccessException {
+
+                        return ps.execute();
```
esotrope: nit: no need for abbr

bgsavage: Right I need to make this a lambda again, I had expanded it out to set a break point and see if I could see the fully parameterized sql.  I could not.

esotrope:
Looks good so far. Excited to see how it performs : J

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamExportQuery.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+public class ExamExportQuery {
+    private long schoolYear;
+    private Set<Long> schoolIds;
+    private Set<Long> schoolGroupIds;
+    private Set<Long> districtIds;
```
agorina: Was the decision made to support multiple or one district at a time?

jtreuting: One district is the largest unit

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/LocalExamExportService.java
agorina: Talked offline - date maybe not unique enough. The idea is to follow an approach for PDF generation.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamExportQuery.java
malaffoon: Hmm. I'm not sure i'd remove anything ... unless there is something in the depths of the implementation that will get confused, why remove functionality that may be useful?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -0,0 +1,58 @@
...
+                        .put("export_location", uriLocation)
+                        .build(),
+                PreparedStatement::execute
+        );
+
+        return uriLocation;
+    }
+
+    private Map<String, Object> getQueryParameters(final ExamExportQuery query) {
+        final String prefix = "query_";
```
malaffoon: i'm sure i'll run into it as i dig through the review but ... why the prefix? I understand it is used elsewhere for the group vs. individual security but why is it needed here?

bgsavage: similar, it's security permission vs. query

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamExportQuery.java
esotrope: Shouldn't we not allow a request tinkerer to DL a district group? is what i was thinking anyway.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+                                                      @Autowired final ExamRepository repository) {
+        if (LocalExamExportService.validRoot(archiveRoot)) {
+            if (!exportDestination.equals(""))
+                throw new IllegalArgumentException("Export destination must be '' to export to " + archiveRoot);
+
+            return new LocalExamExportService(repository);
+        }
+
+        if (S3ExamExportService.validRoot(archiveRoot)) {
+            if(!exportDestination.equals("S3"))
```
malaffoon: nit: `if (!ex`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ExamExportConfiguration {
+
+    @Bean
+    public ExamExportService examExportServiceFactory(@Value("${app.archive.root}") final String archiveRoot,
```
malaffoon: The two parameters seem duplicative. But, per your comment, i'll wait until later to see why it's this way.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/S3ExamExportService.java
malaffoon: nit: order is normally `private static final String`

malaffoon: 👍  order is normally `private static final String`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/S3ExamExportService.java
malaffoon: But the region is not required is it?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamExportQuery.java
bgsavage: I can restrict it on the exposed public reporting /api/ which drops the request to report processor, and leave this for the report processor api should we ever need to use this functionality internally?  Remember this api is assumed to be sitting behind the security perimeter so there won't be any outside access to it.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/S3ExamExportService.java
bgsavage: Ahhh that's the solution to @esotrope 's mind game.  I honestly thought it was something bigger that I wasn't seeing :P

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/S3ExamExportService.java
```diff
@@ -0,0 +1,46 @@
...
+import org.opentestsystem.rdw.reporting.processor.repository.ExamRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import static org.opentestsystem.rdw.reporting.common.security.ReportingRoles.IndividualPiiRead;
+
+public class S3ExamExportService implements ExamExportService {
```
malaffoon: I am not quite seeing why there is this ExamExportService specialization. It looks like there could be just one implementation with properties setting the fiddly bits?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+                                                      @Value("${app.export.destination}") final String exportDestination,
+                                                      @Autowired final ExamRepository repository) {
+        if (LocalExamExportService.validRoot(archiveRoot)) {
```
malaffoon: The reason it is this way in ArchiveService is because of bean-management issues. It seems like this could be handled with more conventional condition bean annotations. Of course, i think that there should be a single ExamExportService controlled by properties so that might be a moot point ...

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/StudentContexts.java
```diff
@@ -0,0 +1,24 @@
+package org.opentestsystem.rdw.reporting.common.jdbc;
+
+import org.opentestsystem.rdw.reporting.common.model.StudentContext;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
+
+public class StudentContexts {
```
wtritch: nit: javadocs?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
bgsavage: What were the bean-management issues?  If I combine to one service, I won't be able to get the props using @Value as aws region isn't required for LocalService to run.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+            if (!exportDestination.equals(""))
+                throw new IllegalArgumentException("Export destination must be '' to export to " + archiveRoot);
```
wtritch: Why can't we export to a sub-directory?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportService.java
wtritch: The report processor doesn't have an authentication principal...

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
malaffoon: Given that i don't think region is required, i just don't see a difference in behavior in the two implementations of ExamExportService. They look like they are (or could be) doing exactly the same thing with configurable parameters.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
malaffoon: Ah, thanks.

report-processor/src/main/resources/application.yml
wtritch: Maybe a comment to that effect in here?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/LocalExamExportService.java
wtritch: :+1 you'll need some DB record to allow the requesting user to access the exported CSV anyway.  Makes sense that you could just use the DB record's primary key as part of the export URI.

bgsavage: Yep that's what I'm going for in the next iteration once I've folded in more of the report processes.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
bgsavage: Alright I can merge the two implementations, with if statements within.  I just thought there was a design pattern with a larger separation of concerns to follow here.  You know I'm always a big fan of SRP.

malaffoon: Hmm. I don't mean to push to a bad design decision. Just questioning the necessity of the separation. If we are going to end up with a single implementation that has `switch`/`if` statements then that is bad and you should ignore me.

malaffoon: The bean-management issues: basically, we need multiple archive services and the properties to instantiate them were in multiple locations in the properties hierarchy. So we need to feed the property prefix to the bean creation.

bgsavage: I'm merging now just to see what it looks like, and see what logic I can consolidate / add configurations for.  Will let you know what I come up with.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
bgsavage: Sub-directory is part of the archive.root path?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ExamExportConfiguration.java
bgsavage: The nice thing about separate implementations too is I can fast fail through the `@Configuration`, whereas if I throw the exceptions in the implementations, then you have to actually exercise that piece of code to find out it's misconfigured.

bgsavage: Hm, yeah I think I prefer the separate impl approach.  It just starts to add too many unrelated methods to a single class, I prefer a more focused approach.

report-processor/src/main/resources/application.yml
bgsavage: I'm spending a little more time to see if I can set this property in code based on archive.root's value.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -30,9 +32,12 @@
 
     @Override
     public String export(final PermissionScope permissionScope, ExamExportQuery query, final String uriLocation) {
+        String exportDestinationSqlCommand = uriLocation.toLowerCase().startsWith("s3")
+                ? "S3"
+                : "";
 
         template.execute(
-                exportExam,
+                exportExam.replace(ExportDestination, exportDestinationSqlCommand),
```
bgsavage: Now using string replace instead of relying on a prop to be set correctly.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -107,7 +107,7 @@ sql:
                     join accommodation a on a.id = eaa.accommodation_id
                   where eaa.exam_id = e.id), '')
 
-      into outfile ${app.export.destination} :export_location
+      into outfile ::EXPORT_DESTINATION:: :export_location
```
bgsavage: I'm giving a slightly different annotation for values that will be string replaced vs paramters.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ArchiveExamExportService.java
```diff
@@ -5,36 +5,38 @@
...
+
     private static final String FileSchemePrefix = "file://";
     private static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH:mm:ss");
     private static final String FileNameFormat = "export-exam-%s.csv";
 
     @Value("${app.archive.root}")
     private String archiveRoot;
 
+    @Value("${app.archive.export-root:}")
+    private String exportBasePath;
```
bgsavage: New prop for export base path. Likely would be something like `/REPORTS/USER/EXAM-EXPORT/`.  The reason being is that the base path will be different if it's file system vs. aws.  If it's file system, you have to give mysql explicit access to the directory via a mysql config value.  I added this to the readme.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Map;
+
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.UNMATCHABLE_IDS;
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getSecurityParameters;
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.nullOrEmptyToDefault;
+
+@Repository
+public class JdbcExamRepository implements ExamRepository {
+
+    private static final String ExportDestination = "::EXPORT_DESTINATION::";
```
esotrope: Maybe I'm missing something but cant the property in the yml just have a default and override setup?
```
thing=s3
thing2=hello.${thing:}
```

README.md
```diff
@@ -51,6 +51,7 @@ and add the following lines:
 [mysqld]
 explicit_defaults_for_timestamp=1
 default-time-zone='UTC'
+secure_file_priv=/tmp/
```
bgsavage: Need to give mysql export access to some directory, this is the default local file system directory.

Docs here: https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_secure_file_priv

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
esotrope: Oh nvm. I see - its inferred from the start of the configured uri

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Value("${sql.export.exam}")
```
malaffoon: I *think* if you set this using an Autowired setter, you could do the substitution in that setter and not have to do it in every export() call

malaffoon: Oh nvm, don't think that would work since export() gets the location.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/export/ArchiveExamExportService.java
```diff
@@ -0,0 +1,42 @@
...
+    private final ExamRepository repository;
+
+    @Autowired
+    ArchiveExamExportService(ExamRepository repository) {
+        this.repository = repository;
+    }
+
+    @Override
+    public String exportExam(final UserPermissions permissions, final ExamExportQuery query) {
+        String uriLocation = archiveRoot.replace(FileSchemePrefix, "/") + exportBasePath;
```
malaffoon: Is this right? If the archive.root is set to "s3://bucket/..." what happens?

malaffoon: perhaps a comment in the code that only the `file://` gets replaced while `s3://` values are left alone?

bgsavage: Thanks, done.

PR #481, opened 2017-10-05
fixes school year range display

esotrope:


common/src/main/resources/templates/iab-body.html
```diff
@@ -17,7 +17,7 @@
                             <span class="label" th:text="#{report.iab.short-name}"></span><span class="label" th:text="#{'report.subject.' + ${assessment.subject} + '.name'}"></span>
                         </span>
                     </div>
-                    <div class="h4" th:text="#{('report.iab.title.grade.' + ${report.studentEnrollment.gradeCode})(${#strings.toString(exam.schoolYear)}, ${support.getLastTwoDigits(exam.schoolYear + 1)})}"></div>
+                    <div class="h4" th:text="#{('report.iab.title.grade.' + ${report.studentEnrollment.gradeCode})(${#strings.toString(exam.schoolYear - 1)}, ${support.getLastTwoDigits(exam.schoolYear)})}"></div>
```
esotrope: Logic now matches angular school year pipe:

```typescript
@Pipe({name: 'schoolYear'})
export class SchoolYearPipe implements PipeTransform {

  transform(value: number): string {
    let valString = value.toString();

    if(valString.length !== 4)
      return valString;

    return `${value - 1}-${valString.substring(2)}`;
  }
}
```

bgsavage: Yikes.  Yeah the school year in the database is the ending year.

PR #482, opened 2017-10-05
enable compression by default

malaffoon:


webapp/src/main/resources/application.yml
```diff
@@ -24,12 +24,15 @@ saml:
...
-# tomcat settings for reasonable performance: adjust for tiny/huge configurations
 server:
+  # tomcat settings for reasonable performance: adjust for tiny/huge configurations
   tomcat:
     max-connections: 0
     max-threads: 4
     accept-count: 10
+  compression:
+    enabled: true
+    mime-types: application/json,application/xml,text/html,text/xml,text/plain,text/css,text/javascript,application/javascript
```
malaffoon: just copied this list from what Will did: good?

bgsavage: Just curious if there's a mime-type we don't want to support compression on, or just allow it to do all?

Also do we even serve xml?

wtritch: lgtm.  I don't think we do today, but having it in there doesn't hurt anything.

bgsavage: Why not just wildcard it then?

bgsavage: Actually ignore that, I don't think you can...

esotrope: what about csv?

malaffoon: Noted. I patched the code.

PR #483, opened 2017-10-05
Bugfix/handle large saml response

wtritch:
This PR is a follow-on to #477 which did not solve the issue with serializing large SAML Response payloads.
The issue was a bit more pervasive than just converting the SAML HttpStorageFactory to use a custom SAMLObject implementation.  The SAMLCredential stored in the session also contained a reference to a SAMLObject too large to serialize using spring-saml default serialization.

Rather than extend and re-write most of the spring-saml integration framework I switched to using Jackson for our serialization framework rather than Java serialization.  This required introducing a number of serializers/deserializers/mixins in order to properly de/serialize SAML objects.


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -1,36 +1,38 @@
...
-import org.springframework.security.core.userdetails.UserDetails;
 
 import java.util.Collection;
 import java.util.Map;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 
-public class User implements UserDetails {
+public class User extends org.springframework.security.core.userdetails.User {
```
wtritch: Most of the built-in Spring Security/Jackson de/serializers expect the principal to extend the existing Spring User class.  Since the information in our custom User object was more or less duplicative this was an easy change.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -69,38 +71,11 @@ public PermissionScope getPermissionScopeByPermissionId(final String id) {
...
-
-    public boolean isEnabled() {
-        return enabled;
-    }
-
 
     /**
      * Builder for User
      */
+    @JsonPOJOBuilder(buildMethodName="build", withPrefix="")
```
wtritch: When telling Jackson to use a builder for deserialization it expects industry-standard "with*" prefixes on property names.  Since we don't use "with" I need this annotation here to tell Jackson how to find the builder property setter methods.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -175,11 +173,6 @@ public SAMLContextProvider contextProvider() throws ServletException {
         return contextProvider;
     }
 
-    @Bean
-    public SAMLMessageStorageFactory storageFactory() {
-        return new SamlHttpSessionStorageFactory();
-    }
```
wtritch: Don't need this anymore... we're no longer replacing existing SAML functionality.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/ExpiringUsernameAuthenticationTokenDeserializer.java
```diff
@@ -0,0 +1,78 @@
...
+        final Object principal;
+        if(principalNode.isObject()) {
+            principal = mapper.readValue(principalNode.toString(), new TypeReference<User>() {});
+        } else {
+            principal = principalNode.asText();
+        }
+        final JsonNode expirationNode = readJsonNode(jsonNode, "tokenExpiration");
+        final Date expiration = expirationNode.isMissingNode() ? null : mapper.readValue(expirationNode.toString(), Date.class);
+
+        final Object credentials = mapper.readValue(readJsonNode(jsonNode, "credentials").toString(), Object.class);
```
wtritch: lines 55-58 are the only changes from the UsernamePasswordAuthenticationTokenDeserializer this method was taken from.  We need to de-serialize the token expiration and deserialize the credentials as an object rather than a string.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SAMLCollectionDeserializer.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * This class is responsible for deserializing SAMLCollection instances using
+ * the built-in spring-saml deserialization method.
+ *
+ * @see SAMLCollection#readObject(ObjectInputStream)
+ */
+class SAMLCollectionDeserializer extends StdScalarDeserializer<SAMLCollection> {
```
wtritch: SAMLCollection is an additional class that doesn't handle serialization of large payloads well.  It's one of two wrapper classes that SAML uses to serialize/deserialize it's XMLObject instances. (The other being SAMLObject.)

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SAMLCredentialMixin.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.io.Serializable;
+
+/**
+ * This mixin controls SAMLCredential de/serialization.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY)
+@JsonDeserialize(builder = SAMLCredentialMixin.Builder.class)
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE)
+abstract class SAMLCredentialMixin {
```
wtritch: Since SAMLCredential doesn't have a no-arg constructor, I had to create a builder to use for de-serialization.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SamlModule.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.security.providers.ExpiringUsernameAuthenticationToken;
+import org.springframework.security.saml.SAMLCredential;
+import org.springframework.security.saml.parser.SAMLCollection;
+import org.springframework.security.saml.parser.SAMLObject;
+
+/**
+ * This module registers de/serialization handlers for all classes
+ * serialized by SAML in the session store.
+ */
+public class SamlModule extends SimpleModule {
```
wtritch: This is the Jackson Module that's used to register all of these handlers with an ObjectMapper instance.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SecuredRedisConfiguration.java
```diff
@@ -26,4 +33,21 @@ public static ConfigureRedisAction configureRedisAction() {
...
+    public RedisSerializer<Object> springSessionDefaultRedisSerializer(final ObjectMapper objectMapper) {
+        final ObjectMapper redisObjectMapper = objectMapper.copy();
+        redisObjectMapper.enableDefaultTyping(ObjectMapper.DefaultTyping.NON_FINAL, JsonTypeInfo.As.PROPERTY);
+        redisObjectMapper.registerModule(new WebJackson2Module());
+        redisObjectMapper.registerModule(new GuavaModule());
+        redisObjectMapper.registerModule(new CoreJackson2Module());
+        redisObjectMapper.registerModule(new SamlModule());
+
+        return new GenericJackson2JsonRedisSerializer(redisObjectMapper);
+    }
```
wtritch: This is where we tell Spring Session how to de/serialize session payloads.

webapp/src/test/java/org/opentestsystem/rdw/reporting/report/ExamReportControllerIT.java
```diff
@@ -76,6 +76,8 @@
 
     private User user = User.builder()
             .id("testUser")
+            .username("username")
+            .password("password")
```
wtritch: username and password are now required, in addition to id

webapp/src/test/java/org/opentestsystem/rdw/reporting/security/session/ExpiringUsernameAuthenticationTokenSerializationTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.security.saml.SAMLCredential;
+
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class ExpiringUsernameAuthenticationTokenSerializationTest extends SessionObjectMapperBase {
```
wtritch: I've written serialization round-trip tests for all of the classes I've declared a serializer/deserializer/mixin for.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SAMLXmlDeserializer.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
+
+/**
+ * This helper utility class provides access to the protected
+ * {@link SAMLBase#unmarshallMessage(Reader)} method.
+ */
+class SAMLXmlDeserializer extends SAMLBase<XMLObject, XMLObject> {
```
wtritch: spring-saml hides it's deserialization implementation inside a protected method of the SAMLBase abstract class.  This class just exposes the functionality for utility-style usage.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/User.java
esotrope: Interesting. I thought with* was when you wanted to set a built instances prop and the with* method returns a copy of the instance with the new prop set. At least i think that's how the immutables library uses the word

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/ExpiringUsernameAuthenticationTokenMixin.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.reporting.security.session;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.springframework.security.providers.ExpiringUsernameAuthenticationToken;
+
+/**
+ * This mixing controls de/serialization of
```
esotrope: mixin?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SAMLObjectSerializer.java
```diff
@@ -0,0 +1,37 @@
...
+    public SAMLObjectSerializer() {
+        super(SAMLObject.class);
+    }
+
+    @Override
+    public void serialize(final SAMLObject value, final JsonGenerator gen, final SerializerProvider provider) throws IOException {
+        try {
+            final XMLObject wrappedObj = value.getObject();
+            final String serializedValue = XMLHelper.nodeToString(SAMLUtil.marshallMessage(wrappedObj));
+            gen.writeString(serializedValue);
```
esotrope: the diff in all of these is to use writeString instead of writeUTF right. Just making sure i understand what is different

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/ExpiringUsernameAuthenticationTokenMixin.java
wtritch: doh, fixed

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/session/SAMLObjectSerializer.java
wtritch: Not anymore.  Now we switched the whole serialization framework, so we're no longer writing to an ObjectOutputStream with writeUTF or writeString.  Instead we're writing to a JsonGenerator.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/User.java
```diff
@@ -1,36 +1,38 @@
...
-    private String password;
-    private Collection<? extends GrantedAuthority> authorities;
-    private boolean accountNonExpired;
-    private boolean accountNonLocked;
-    private boolean credentialsNonExpired;
-    private boolean enabled;
 
-    private User() {}
+    private User(final String username,
+                final String password,
```
chdietz: I was concerned when I saw password, why do we need this?

esotrope: when we extend spring's User model / implement UserDetails it needs us to have a password field. we just throw in the same value of ```"REDACTED"``` for all users though

chdietz: ok, I see it's been out there for a while, it just caught my attention.

PR #484, opened 2017-10-05
Feature/export exams csv

bgsavage:
Iterative Merge.

This PR sets up the message stream for Export Exams.  You still hit `api/reports` like you do for the other reports, when the branch being where you send the message request.  This gives me the api call and report write record for free.

This is my first stream setup in spring, so please let me know if I missed anything...


common/src/main/java/org/opentestsystem/rdw/reporting/common/report/BatchExamReportRequestHolder.java
```diff
@@ -4,22 +4,22 @@
...
+public class BatchExamReportRequestHolder<T> {
 
     private String id;
     private UserPermissions userPermissions;
     private String user;
     private String label;
 
     @JsonTypeInfo(use=JsonTypeInfo.Id.CLASS, include=JsonTypeInfo.As.PROPERTY, property="@class")
-    private AbstractExamReportRequest<PrintOptions> request;
+    private AbstractExamReportRequest<T> request;
```
bgsavage: Pushed the generic typing up allowing me to create a `BatchExamReportRequestHolder` which holds my QueryOptions.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExamExportQueryOptions.java
```diff
@@ -1,10 +1,10 @@
-package org.opentestsystem.rdw.reporting.processor.model;
+package org.opentestsystem.rdw.reporting.common.report;
 
 import com.google.common.collect.ImmutableSet;
 
 import java.util.Set;
 
-public class ExamExportQuery {
+public class ExamExportQueryOptions {
```
bgsavage: Rename.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
```diff
@@ -0,0 +1,4 @@
+package org.opentestsystem.rdw.reporting.common.report;
+
+public class ExportExamReportRequest extends AbstractExamReportRequest<ExamExportQueryOptions> {
+}
```
bgsavage: Nothing here to be extended here, mainly here for documenting what the request looks like.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -1,25 +1,29 @@
...
 
     private static final String FileSchemePrefix = "file://";
-    private static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH:mm:ss");
-    private static final String FileNameFormat = "export-exam-%s.csv";
+    private static final String FileNameFormat = "export-exam-%d.csv";
 
     @Value("${app.archive.root}")
+    public void setArchiveRoot(final String value){
+        archiveRoot = value.replace(FileSchemePrefix, "/");
+    }
```
bgsavage: Spring setter as recommended.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -33,12 +37,27 @@
...
+        PermissionScope scope = getPermissionScope(permissions);
+        String uri = getUriLocation(reportId);
+
+        return repository.export(scope, query, uri).replace(archiveRoot, "");
+    }
+
+    private PermissionScope getPermissionScope(final UserPermissions permissions) {
+        Map<String, Permission> permissionMap = permissions.getPermissionsById();
+
+        return (permissionMap.isEmpty() || permissionMap.get(IndividualPiiRead) == null)
```
bgsavage: I don't think we could get here without this permission, but better to code defensively.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -33,12 +37,27 @@
     }
 
     @Override
-    public String exportExam(final UserPermissions permissions, final ExamExportQuery query) {
+    public String exportExam(final UserPermissions permissions, final ExamExportQueryOptions query, final long reportId) {
+        PermissionScope scope = getPermissionScope(permissions);
+        String uri = getUriLocation(reportId);
+
+        return repository.export(scope, query, uri).replace(archiveRoot, "");
```
bgsavage: Removing the archiveRoot from the return result because when we call archiveService to retrieve later, it adds it back for us.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -33,12 +37,27 @@
...
+        Map<String, Permission> permissionMap = permissions.getPermissionsById();
+
+        return (permissionMap.isEmpty() || permissionMap.get(IndividualPiiRead) == null)
+                ? PermissionScope.EMPTY
+                : permissionMap.get(IndividualPiiRead).getScope();
+    }
+
+    private String getUriLocation(final long reportId) {
         // Only FileSchemePrefix needs to get replaced for local MySql.
         // Aurora's SELECT INTO OUTFILE S3 feature shares the same s3:// as archive's ResourceLoader.
```
bgsavage: Oops I need to move this comment to the setter.  See why I don't like comments?? :P

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExams.java
```diff
@@ -0,0 +1,61 @@
...
+    @Autowired
+    public ExportExams(final ExamExportService exportService,
+                       final ReportService reportService) {
+
+        this.exportService = exportService;
+        this.reportService = reportService;
+    }
+
+    @ServiceActivator(inputChannel = ReportExportExams)
+    public void exportExams(final ReportRequestMessage message) {
```
bgsavage: New subscriber to the ReportExportExams channel.  It's pretty similar to FetchStudents except it does everything in one step.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExamsSink.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.messaging.SubscribableChannel;
+
+import static org.opentestsystem.rdw.reporting.processor.stream.ExportExamsSource.ReportExportExams;
+
+public interface ExportExamsSink {
+
+    @Input(ReportExportExams)
+    SubscribableChannel exportExams();
```
bgsavage: This is what I was missing in the demo today, causing half my messages to be ignored.

bgsavage: @chdietz

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ExportController.java
```diff
@@ -1,44 +0,0 @@
...
-public class ExportController {
-
-    private ExamExportService service;
-
-    @Autowired
-    ExportController(ExamExportService service) {
-        this.service = service;
-    }
-
-    @PostMapping("/api/export/exams")
```
bgsavage: I was going to create this as a separate endpoint / controller, but once I got it working through `api/reports`, I couldn't justify it.

report-processor/src/main/resources/application.yml
```diff
@@ -21,6 +21,11 @@ spring:
   cloud:
     stream:
       bindings:
+        ReportExportExams:
+          group: default
+          producer:
+            requiredGroups: default
```
bgsavage: Thoughts?

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/BatchExamReportRequestHolder.java
esotrope: Interesting. Didn't imagine you doing this but i guess its good that we can reuse. This means we can add asmt type & subject filters later more easily.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
esotrope: I thought declarative programming was supposed to relieve us from this x p

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExams.java
```diff
@@ -0,0 +1,61 @@
...
+        final Report report = reportService
+                .findOneById(message.getReportId())
+                .orElseThrow(() -> new IllegalArgumentException("Unknown Report: " + message.getReportId()));
+
+        if (report.getStatus() == ReportStatus.FAILED) {
+            return;
+        }
+
+        final ExportExamReportRequest request = (ExportExamReportRequest) report.getReportRequest();
+        ExamExportQueryOptions options =  request.getOptions();
```
esotrope: var ref is not reused. i would put it inline in your method call below. also final.

report-processor/src/main/resources/application.yml
esotrope: looks right to me

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/BatchExamReportRequestHolder.java
esotrope: @jtreuting can we get rid of PrintOptions.grayscale setting yet ; J

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/ReportProcessorConfiguration.java
```diff
@@ -25,6 +27,8 @@
         ReportGenerationRetryProperties.class
 })
 @EnableBinding({
+        ExportExamsSource.class,
+        ExportExamsSink.class,
```
malaffoon: Hmm, i'll have to see where this is going but i'm surprised to see a single process with both a source and a sink for this work. I thought one process would be posting the request message and another would be processing it.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExams.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @ServiceActivator(inputChannel = ReportExportExams)
+    public void exportExams(final ReportRequestMessage message) {
+        logger.debug("Exporting exams for report request: {}", message.getReportId());
+        final long reportId = message.getReportId();
+
+        final Report report = reportService
+                .findOneById(message.getReportId())
+                .orElseThrow(() -> new IllegalArgumentException("Unknown Report: " + message.getReportId()));
```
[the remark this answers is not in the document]
bgsavage: As I understand it, we have a `ReportFailingChannelInterceptor` which will see that an exception was thrown and will set the Report status to FAILED.

report-processor/src/main/resources/application.yml
malaffoon: I'm still a bit confused why the report processor would be a source.

malaffoon: I'm also a little disappointed in the inconsistency in queue naming but you're just following the pattern so nothing to change.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepositoryIT.java
```diff
@@ -129,6 +132,17 @@ public void exportBySchoolWithNotSchoolAccessShouldExportDistrictExams() throws
         assertThat(count).isEqualTo(HeaderRowCountOnly);
     }
 
+    @After
+    public void CleanUp() throws IOException {
+        try {
+            Files.delete(Paths.get(testCsv));
```
malaffoon: Do you have to use `/tmp/` in your tests. If you use java tmpdir functionality, i bet it would work. (looking for an example to copy)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExams.java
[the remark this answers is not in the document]
malaffoon: Ok, thanks for explaining.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/ReportProcessorConfiguration.java
wtritch: Usually something with both a sink and a source is just called a "*Processor" =)

bgsavage: So the suggestion is to just have a single `ExportExamsProcessor` that has both the `MessageChannel` and the `SubscribableChannel` for exportExams() ?

malaffoon: I would push back against that. A *Processor is something that transforms a message. This really is a separate source and sink that just happen to be in the same application. But it would work either way ... it's just a way of naming and binding the channels.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
```diff
@@ -26,13 +28,16 @@
     private static final Logger logger = LoggerFactory.getLogger(SpringStreamReportGenerator.class);
 
     private final ReportService reportService;
-    private final FetchStudentsSource source;
+    private ExportExamsSource exportReportSource;
```
wtritch: nit: final

report-processor/src/main/resources/application.yml
bgsavage: Ah so the suggestion is to break out the source and the sinks into different projects.  Yeah I was a bit confused as to why we are proxying the webapp calls to report processor instead of just posting the message from the reporting webapp.

malaffoon: Again, "suggestion" may be too strong. I'm just trying to understand the pattern that is established. In ingest, yes, we post from the main API app, then consume from a separate app. I think we should discuss that approach for this. But i know the big student report uses a different pattern.

bgsavage: Sorry, "alternative" ? :) I'll bring it up in parking lot.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExams.java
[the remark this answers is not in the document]
wtritch: @bgsavage nailed it =)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExamsSink.java
```diff
@@ -0,0 +1,12 @@
+package org.opentestsystem.rdw.reporting.processor.stream;
+
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.messaging.SubscribableChannel;
+
+import static org.opentestsystem.rdw.reporting.processor.stream.ExportExamsSource.ReportExportExams;
+
+public interface ExportExamsSink {
```
wtritch: nit: javadocs

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepositoryIT.java
bgsavage: This requires people to update their configuration for my.sql to allow exports to that directory as well.  Is this what we want?  Right now I have local debug / and IT pointing to the same location so you only need to update your my.cnf for the one location.

PR #488, opened 2017-10-06
DWR-1046 If a report request does not find any students, update status to NO_R…

wtritch:
…ESULTS

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
@@ -84,6 +84,10 @@ public class FetchStudents {
         final Collection<Student> students = studentRepository.findStudentsByExamQueryParams(
                 userPermissions,
                 queryParams);
+        if (students.isEmpty())  {
+            setNoContent(report);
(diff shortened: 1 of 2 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
```diff
@@ -84,6 +84,10 @@ public void fetchAndPublishStudents(final ReportRequestMessage message) {
         final Collection<Student> students = studentRepository.findStudentsByExamQueryParams(
                 userPermissions,
                 queryParams);
+        if (students.isEmpty())  {
+            setNoContent(report);
+            return;
+        }
```
wtritch: If there are no students, set the report status to NO_RESULTS and stop processing.

bgsavage: Does this short circuit all the follow up messages that are supposed to happen down the pipe?

wtritch: yup, because nothing gets submitted to the message queue

PR #489, opened 2017-10-06
Create a gradle "coverage" task that reports combined Unit and Integr…

wtritch:
…ation test coverage.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,9 +105,6 @@ subprojects {
         testCompile 'org.mockito:mockito-all'
     }
 
-    // generate coverage report automatically (./build/reports/jacoco/test/html/index.html)
-    test.finalizedBy(jacocoTestReport)
-
     // task to generate build-info.properties with build.version, build.timestamp
     // using this approach instead of resource filtering to avoid modifying source
     task createBuildInfoFile(dependsOn: processResources) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -204,8 +201,20 @@ subprojects {
...
+    }
+
+    jacocoTestReport {
+        reports {
+            xml.enabled = false
+            csv.enabled = false
+            html.enabled = true
+            html.destination = file("${reporting.baseDir}/${name}")
+        }
     }
```
wtritch: With this change you can still generate individual coverage reports on unit tests via:
`gradle clean test jacocoTestReport`

build.gradle
```diff
@@ -204,8 +201,20 @@ subprojects {
...
-    tasks.withType(Test) {
-        reports.html.destination = file("${reporting.baseDir}/${name}")
+    task coverage(type: Task, dependsOn: [test, IT]) {
+        doFirst {
+            jacocoTestReport.executionData(test, IT)
+        }
+        finalizedBy(jacocoTestReport)
+    }
+
+    jacocoTestReport {
```
malaffoon: i know i said leave this in during the call but ... are these the same as the default settings? If so, i guess i'd get rid of it.

chdietz:
Sounds good to me.

agorina:
Thank you for taking care of it!

build.gradle
wtritch: It looks like xml and csv reports are disabled by default.  It looks like the only thing we really want to configure is the html report location.  Before these changes the coverage report was found at:
`build/reports/jacoco/test/html/index.html`
After these changes the "coverage" task coverage report is found at:
`build/reports/coverage/index.html`

PR #490, opened 2017-10-06
Feature/make org typeahead stick to one mode

esotrope:
Made the search-select.ts comp decide to be a dropdown/typeahead once when the options are first set.

Replaced use of school-select with the more generic search-select.
This turned out to be a bigger change than I thought as you'll see.


webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
```diff
@@ -12,16 +12,14 @@ <h3 class="blue mb-md">{{'labels.organization-export.heading' | translate}}</h3>
...
                   (select)="add($event)"
-                  dropdownThreshold="10"
                   placeholder="{{'prompt.select' | translate}}"
                   searchPlaceholder="{{'labels.organization-export.form.organization.placeholder' | translate}}"
                   buttonLabel="{{'labels.organization-export.form.organization.add' | translate}}">
             </span>
             <div class="ml-xs mt-xs">
               <a *ngIf="selectAllEnabled"
-                 class="aqua"
+                 class="link-button aqua"
```
esotrope: @aaronsleeper added this class to remove underline on the anchor, please advise if there is a better way.

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
```diff
@@ -1,12 +1,12 @@
 <div *ngFor="let districtTree of model.children" class="district-tree">
 
-  <a *ngIf="districtTree.value.name" (click)="onClick(districtTree.value)" href="javascript:void(0)">
+  <a class="link-button hover-icon-holder" *ngIf="districtTree.value.name" (click)="onClick(districtTree.value)" href="javascript:void(0)">
```
esotrope: @aaronsleeper ```.hover-icon-holder``` makes it to where the X icon on districts/school groups only shows when you hover over it.

webapp/src/main/webapp/src/app/organization-export/organization/organization.service.ts
```diff
@@ -18,19 +20,15 @@ export class OrganizationService {
...
     return Observable.forkJoin(
-      this.getSchools(),
+      this.userService.getCurrentUser(),
       this.getSchoolGroups(),
       this.getDistricts()
     ).map(response => {
-      let [ schools, schoolGroups, districts ] = response;
-      return this.mapper.createUserOrganizations(schools, schoolGroups, districts);
+      let [ user, schoolGroups, districts ] = response;
+      return this.mapper.createUserOrganizations(user.schools, schoolGroups, districts);
```
esotrope: @bgsavage changed this to use the user schools like we discussed

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -1,52 +1,45 @@
...
 @Injectable()
 export class OrganizationService {
-  private schoolsWithDistricts: School[] = null;
-  private schoolsLoaded: boolean = false;
 
-  constructor(private userService: UserService, private dataService: DataService) {
+  private schoolsWithDistricts: Observable<School[]>;
+
+  constructor(private userService: UserService,
+              private dataService: CachingDataService) {
```
esotrope: @bgsavage district caching works - school-grade/organization.service.ts had to change to use the CacheDataService

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -1,52 +1,45 @@
...
-            school.districtName = districts.get(school.districtId);
-            return school;
-          });
-
-        this.schoolsLoaded = true;
-        this.schoolsWithDistricts = schools;
-
-        schoolObserver.next(schools);
-        schoolObserver.complete();
+    return this.schoolsWithDistricts = Observable
```
esotrope: was able to drastically simplify this method which was nice

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
```diff
@@ -1,57 +1,61 @@
...
-        [school]="currentSchool"
-        [grade]="currentGrade"
-        [schoolYears]="filterOptions.schoolYears"
-        [schoolYear]="currentSchoolYear"
-        (onShow)="initializeDownloader(reportDownloader)"></span>
-  </span>
-</div>
-
-<div *ngIf="!currentSchool" class="alert alert-danger">
+<div *ngIf="!schoolIsAvailable" class="alert alert-danger">
```
esotrope: now following the pattern of the school grades which aren't available on page load. We have to wait till we make the getSchoolsWithDistrictNames() request to know if the route's school ID doesn't match any of your schools.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -91,16 +94,25 @@ export class SchoolResultsComponent implements OnInit {
...
     this.schools = this.route.snapshot.data[ "user" ].schools;
-    this.availableSchools = this.schools;
-
-    let schoolId = this.route.snapshot.params[ "schoolId" ];
-
-    this.currentSchool = this.schools.find(x => x.id == schoolId);
+    let schoolId = Number.parseInt(this.route.snapshot.params[ "schoolId" ]);
+
+    // get schools with districts
+    this.organizationService.getSchoolsWithDistricts()
```
esotrope: Because i changed the org service method to not modify the user's schools in-place we need to call this service on the school result page now (result will be cached from the homepage)

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -109,16 +121,13 @@ export class SchoolResultsComponent implements OnInit {
...
-        .findGradesWithAssessmentsForSchool(this.currentSchool)
-        .subscribe(grades => {
-          this.availableGrades = grades;
-          this.gradesAreUnavailable = this.availableGrades.length == 0;
-
-          this.currentGrade = this.availableGrades.find(grade => grade.id == this.route.snapshot.params[ "gradeId" ]);
-        });
-    }
+    this.schoolService
+      .findGradesWithAssessmentsForSchool(schoolId)
```
esotrope: Made the school service accept only a school ID so that we can make this call with just the route information and don't need to wait on the success of the getSchools() call

webapp/src/main/webapp/src/app/shared/form/search-select.ts
```diff
@@ -99,21 +93,33 @@ export class SearchSelect {
...
-      }
+    this.updateValue();
+
+    // initialize the text if the value is already set
+    if (!this.dropdown && this.value) {
+      this.search = this.value.name;
     }
   }
 
+  private decideMode(): void {
```
esotrope: The method that decides the select input mode. the undefined check makes sure that it only runs once and you can override it by explicitly setting the dropdown boolean ```@Input()```

webapp/src/main/webapp/src/styles.less
```diff
@@ -180,3 +180,42 @@ footer {
...
+
+/*
+  Not sure what to do to get these line-heights right without explicit styles.
+  btn-[size] didn't quite match things up
+ */
+
+organization-export form button[type='submit'],
+search-select-with-button .input-group-btn > .btn,
+[search-select-with-button] .input-group-btn > .btn {
+  line-height: 1.48;
```
esotrope: @aaronsleeper had to add these styles for the input/button heights to match in the organization export page.

webapp/src/main/webapp/src/app/organization-export/organization/organization.service.ts
bgsavage: 👍

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
bgsavage: 👍

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
bgsavage: Is this a combined return and assignment?  This type of return makes me a bit nervous, but I guess I can be ok with it if you are.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
bgsavage: Yeah... this is what I wanted to avoid.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
bgsavage: Hmm, I preferred how this was before with your data being resolved before you even got here...

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.html
```diff
@@ -1,23 +0,0 @@
-<div>
```
bgsavage: Where did this go?

webapp/src/main/webapp/src/app/shared/form/search-select.ts
bgsavage: Maybe `setDefaultDropdownMode`?

webapp/src/main/webapp/src/styles.less
aaronsleeper: that's fine at this point . bootstrap sets heights in px by default for buttons so it's a non-issue to match them. we should use the same units though as 1.48 will be relative to other things that may change independently. so i recommend converting to px

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
aaronsleeper: don't forget `role="alert"`

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
aaronsleeper: it seems like we could abstract this a bit more like `.hidden-until-hover` or something and set the opacity to 0 as well as visibility to hidden

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
esotrope: I'm a Javascript developer. Nothing makes me nervous : J so please keep me in check.

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
esotrope: Cool I'll try and use that.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
esotrope: I'm actually not sure how this worked before.
Because the getSchoolsWithDistricts() modified the session's user schools in place. Meaning if you login and go directly to the school results page you would not have the district names in the dropdown

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
esotrope: Not sure what you mean - either way we leave the grade unresolved until you arrive on the page so we are doing some post arrival async resource loading.

webapp/src/main/webapp/src/app/school-grade/school-select/school-select.component.html
esotrope: app/shared/form/search-select.ts

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
bgsavage: If you try in QA right now (refresh the school results page), you do have the district names.   That's because it's the same component using the same service, it doesn't matter where it gets used first.

esotrope: Ah I remember now. It did do this before but before there was a wrapper component that this was data-res was delegated to was the difference. Do you think that is worth re-introducing to abstract away this one call?

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
esotrope: looks like ```.btn-link``` makes ```border-color:transparent``` only

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
esotrope: I think later we'll replace this with our generic error view in a well anyway so I'll leave it for now

webapp/src/main/webapp/src/styles.less
esotrope: Just adding px didn't work. Also bs3 looks like it uses these unit-less values. That's where i copied from.

PR #491, opened 2017-10-09
Do not execute unit tests when running integration test task

wtritch:
While working on DWR-1048 I found that our integration test task is *also* executing unit tests.  This means that our unit tests are being executed 3 times during the current build process:
'build' task, 'coverage->test' task, 'coverage->IT' task

This PR changes our IT integration test task to *not* run unit tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -179,6 +179,7 @@ subprojects {
     //This task for for manually running the integration tests against schemas being developed simultaneously with this
     // project, versus the schemas that are "fixed" under the RDW_Schema submodule in this project
     task manualIT(type: Test) {
+        exclude '**/*Test.*'
         outputs.upToDateWhen { false }
         doFirst {
             println "Running Manual Integration Tests..."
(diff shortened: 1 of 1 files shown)


bgsavage:
I thought gradle was supposed to be smart enough to organize its tasks to not allow that to happen?

chdietz:
So this prevents build running unit test, unit test, IT.
Is there( or will there be) one command to run all unit and IT tests?
Also does this issue exist in other projects?

build.gradle
```diff
@@ -179,6 +179,7 @@ subprojects {
     //This task for for manually running the integration tests against schemas being developed simultaneously with this
     // project, versus the schemas that are "fixed" under the RDW_Schema submodule in this project
     task manualIT(type: Test) {
+        exclude '**/*Test.*'
```
malaffoon: Should this be a positive match instead of an exclusion. Seems like we're excluding `**/*IT.*` from unit tests so can this be `include '**/*IT.*'`?

wtritch:
@chdietz this isn’t so much about the same task being run multiple times as the same tests being run in two independent tasks

build.gradle
wtritch: Hrm... i’ll see if include is includeOnly or includeAnd

wtritch: Looks like include is includeOnly.  Will switch to a positive assertion using include rather than a negative.

PR #492, opened 2017-10-09
Feature/export exams csv

bgsavage:
Iterative PR.

This PR now allows for CSV exports posted through report processor to be downloaded in the UI.  A couple of minor tweaks made to the URI formatting to be consistent with other archiveService usages.


build.gradle
```diff
@@ -31,7 +31,7 @@ ext['thymeleaf-layout-dialect.version'] = '2.1.1'
 apply plugin: 'io.spring.dependency-management'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.0.0-SNAPSHOT`
-String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.1.0-43"
+String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.1.0-44"
```
bgsavage: Using latest RDW-Common from my previous updates to LocalArchiveService.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -39,25 +40,37 @@ public void setArchiveRoot(final String value){
...
+        return URI.create(uriLocation + fileNamePrefix);
+    }
+
+    private URI formatUri(String uri) {
+        final String uriPath = uri.replace(archiveRoot, "");
+
+        return URI.create(
+                uriPath.startsWith("/")
+                        ? uriPath.substring(1)
+                        : uriPath
```
bgsavage: ArchiveService doesn't handle leading slashes in the path so making sure to remove them here.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -39,25 +40,37 @@ public void setArchiveRoot(final String value){
...
                 ? PermissionScope.EMPTY
                 : permissionMap.get(IndividualPiiRead).getScope();
     }
 
-    private String getUriLocation(final long reportId) {
-        String uriLocation = archiveRoot + exportBasePath;
-        String fileNamePrefix = String.format(FileNameFormat, reportId);
+    private URI getUriLocation(final long reportId) {
+        final String uriLocation = archiveRoot
+                + (!archiveRoot.endsWith("/") && !exportBasePath.startsWith("/") ? "/" : "")
```
bgsavage: Want to make sure we don't end up with an `archiveRootexportBase/filename.csv` scenario when it should be `archiveRoot/exportBase/filename.csv`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -39,25 +40,37 @@ public void setArchiveRoot(final String value){
...
+                + exportBasePath;
 
-        return uriLocation + fileNamePrefix;
+        final String fileNamePrefix = String.format(FileNameFormat, reportId);
+        return URI.create(uriLocation + fileNamePrefix);
+    }
+
+    private URI formatUri(String uri) {
```
esotrope: I think spring might have some util methods for tailing and leading forward slashes. If not, this might be a good candidate for breaking out to a utility class that can be tested independently.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -46,7 +44,7 @@ public String export(final PermissionScope permissionScope, ExamExportQueryOptio
                 PreparedStatement::execute
         );
 
-        return uriLocation;
+        return isDestinationS3 ? uriLocation + ".part_00000" : uriLocation;
```
wtritch: Why is the ".part_00000" needed?

bgsavage: Something Amazon adds... I know, its fugly. :( 

http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Aurora.SaveIntoS3.html
`s3-region://bucket-name/file-prefix.part_00000`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
wtritch: Interesting, why are you including the archive root in the URI?  Won't you have to trim it back out again before accessing the resource via the archive service?

wtritch: Oh, because the full URI needs to be passed to the repository since mysql is outputting the file... Hm, I'd be temped to ask the archive service to give me the full URI for a location rather than hand-jam it outside of the archive service.  That way you only have one class that knows how to combine the archive root with a location.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
wtritch: This also feels a bit like usurping archive service behavior.  Should the ArchiveService implementation be the only thing that knows how to turn a "location" into a full URI?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
bgsavage: Hm, yeah but it is it worth it to add a service dependency just to read a configuration?  Either way, this abstraction is leaking.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
wtritch: Oh, that's gross... Is that javadoc'ed or commented in a few places?  Coming in blind, I'd have no idea that telling the repository to export a CSV in an aurora context would rename my filenames.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
bgsavage: Actually this is turning a full URI into a "location" so that it can be consumed by archiveService later.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
malaffoon: I'm not keeping up with where things are with this but getting at this information is why the archive service sets the "rawURI" property.

Also, there is a FilenameUtils thing that has concat for doing path separator manipulation.

bgsavage: Hm yeah but is there a way to pass that same rawUri back to archiveService when reading the resource?  I think it expects a certain format.

malaffoon: I was just commenting that the raw URI is available from the archive service. Not really talking about using it to talk to the archive service. But i guess we could put something like that in ... "ArchiveService, here's a rawURI that i think you should know about, tell me what your 'location' for that would be".

wtritch: It feels closer to "Hey ArchiveService, here's a location to a file I'm going to inject in underneath you. Give me the rawURI for that location so I can write some bits without your knowledge."

bgsavage: I guess another idea I was thinking was to pass the SQL to ArchiveService and let ArchiveService fill in the location param and execute the export.

wtritch: ehhh, that's bleeding a bit too much outside knowledge into the ArchiveService, IMHO.  I do like the ArchiveService being the only thing that knows about pathing and path concatenation.  I could also see it having a "getProtocol" method that returns either "file" or "s3" so that only the ArchiveService knows about the "app.archive.root" property.

report-processor/src/main/resources/application.yml
```diff
@@ -1,12 +1,12 @@
...
 # since those are the only end-points, switch to a less standard port using the main server.port config
 management:
-  port: 8008
+  port: 8005
   security:
     enabled: false
 
 server:
-  port: 8080
+  port: 8085
```
bgsavage: unrelated, but I feel like the both reporting and report processor apps should have a default configuration to allow them to run in conjunction without any param changes.  Thoughts?

webapp/src/main/webapp/src/app/organization-export/organization-export.service.ts
```diff
@@ -28,7 +28,7 @@ export class OrganizationExportService {
   }
 
   private createExportInternal(request: OrganizationExportRequest): Observable<void> {
-    return this.dataService.post('/organizations/examExport', request)
+    return this.dataService.post('/exams/export', { name: 'export-exams',  options: request })
```
bgsavage: placeholder report name until @esotrope writes the logic for what name to pass in.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
bgsavage: The `FilenameUtils.concat` method does things that can potentially mess us up, just looking at the docs, but here are a couple of examples:

```
* /foo + /bar          -->   /bar
* /foo/ + /bar         -->   /bar
```

In both cases, I would want foo/bar to be the result.

bgsavage: However, it does support the happy path I was trying to solve for: 

```
* /foo + bar           -->   /foo/bar
```

because archive.root will be `s3://bucket-name` and the archive.exam-export-root will be `REPORTS/USER/EXAM-EXPORT/`

Thoughts?

report-processor/src/main/resources/application.yml
wtritch: I thought the same thing until @malaffoon pointed out that we want deployment to be as easy, with as few configuration changes as possible.  Which means that when each of our services runs in its own image, it should be running at 8080.  We can modify the port easily as developers setting a -Dserver.port=8085 in our runtime configs or via docker-compose port mappings.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
wtritch: I'm thinking we could just have two additional methods on the ArchiveService interface that removes the need for all this logic inside a consumer:
```
  String getRootURI(final String location);
  String getProtocol();
```
And we can probably re-use existing concatenation/etc methods that already exist on the archive service.

report-processor/src/main/resources/application.yml
bgsavage: I figured there was a reason I wasn't seeing.  Alright I'll revert.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
bgsavage: There will still be some concatenation as I've introduced the exam-export property so it will save one concat, but we will still need the other.

bgsavage: So after offline discussion, I'm going to keep this PR "as is"  and merge, then create two more PRs for the suggestion of updating ArchiveService and injecting it as a service.

PR #495, opened 2017-10-10
DWR-1072 Trim student search input value before searching for a student

wtritch:


webapp/src/main/webapp/src/app/student/student-exam-history.service.ts
```diff
@@ -45,8 +45,9 @@ export class StudentExamHistoryService {
   existsBySsid(ssid: string): Observable<Student> {
     let params: URLSearchParams = new URLSearchParams();
     params.set('hasExams', 'true');
+    let trimmedSsid: string = ssid.trim();
 
-    return this.dataService.get(`/students/${ssid}`, {params: params})
+    return this.dataService.get(`/students/${trimmedSsid}`, {params: params})
```
bgsavage: didn't want to inline it?

wtritch: I tried it, but it felt strange having a function call inside an interpolated string =/

PR #496, opened 2017-10-10
changes "print" and "download" to "create"

esotrope:
screen shot 2017-10-10 at 1 25 06 pm


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -445,20 +445,19 @@
...
+          "multiple": "Create {{name}} Student Reports"
         },
-        "regenerate": "Regenerate",
         "assessment-type": "Assessment Type",
         "subject": "Subject",
         "school-year": "School Year",
         "accommodations": {
           "title": "Student Accommodations",
-          "info": "TODO"
+          "info": ""
```
esotrope: @jtreuting realized we still don't have text for the info button above student accommodations on the report form. Until we have the text i'm tempted to take the info button out.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -514,7 +513,7 @@
         "submit": {
           "label": "Export Results",
           "success-html": "Creating export. See <a href=\"/reports\">status</a>.",
-          "failure": "Error creating report. Please try again."
+          "failure": "Error creating export. Please try again."
```
esotrope: unrelated - this is to change the org level export fail message.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -445,20 +445,19 @@
...
       "form": {
         "title": {
-          "single": "Print {{firstName}}'s Report",
-          "single-prepopulated": "Print {{name}}'s {{schoolYear}} {{subject}} {{assessmentType}} Report",
-          "multiple": "Print {{name}} Student Reports"
+          "single": "Create {{firstName}}'s Report",
+          "single-prepopulated": "Create {{name}}'s {{schoolYear}} {{subject}} {{assessmentType}} Report",
+          "multiple": "Create {{name}} Student Reports"
         },
-        "regenerate": "Regenerate",
```
bgsavage: This isn't used when a report fails?

esotrope: Right now it just shows an ```fa-refresh``` icon instead

PR #497, opened 2017-10-10
Add additional details to Report payload.

wtritch:
This PR adds additional properties to the Report payload sent to the front-end via a ReportResource assembled from a saturated Report model instance.

This is the first step in updating the "My Reports" page to contain additional information when displaying the user's report history.

On a good note, the front-end currently only requires/uses three properties "label," "status," and "created," so these changes can be merged without requiring any associated front-end changes.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
@@ -1,6 +1,7 @@
 package org.opentestsystem.rdw.reporting.common.model;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import org.opentestsystem.rdw.reporting.common.report.AbstractExamReportRequest;
 
 import java.net.URI;
@@ -60,7 +61,7 @@ public class Report {
     /**
      * @return The report request parameters
      */
-    @JsonIgnore
+    @JsonTypeInfo(use=JsonTypeInfo.Id.CLASS, include=JsonTypeInfo.As.PROPERTY, property="@class")
     public AbstractExamReportRequest getReportRequest() {
         return reportRequest;
     }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
@@ -18,6 +18,8 @@ public abstract class AbstractExamReportRequest<T> {
         return assessmentType;
     }
 
+    public abstract ReportType getReportType();
+
     public Subject getSubject() {
         return subject;
     }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
@@ -12,6 +12,5 @@ public class ExportExamReportRequest extends AbstractExamReportRequest<ExamExpor
     }
 
     public static class Builder extends AbstractExamReportRequest.Builder<ExamExportQueryOptions, ExportExamReportRequest, ExportExamReportRequest.Builder> {
-
     }
 }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
@@ -1,4 +1,17 @@
 package org.opentestsystem.rdw.reporting.common.report;
 
 public class ExportExamReportRequest extends AbstractExamReportRequest<ExamExportQueryOptions> {
+
+    @Override
+    public ReportType getReportType() {
+        return ReportType.ExamExport;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder extends AbstractExamReportRequest.Builder<ExamExportQueryOptions, ExportExamReportRequest, ExportExamReportRequest.Builder> {
+
+    }
 }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportType.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportType.java
@@ -1,0 +1,11 @@
+package org.opentestsystem.rdw.reporting.common.report;
+
+/**
+ * This enum represents a Report type.
+ */
+public enum ReportType {
+    SchoolGrade,
+    Group,
+    Student,
+    ExamExport
+}
--- a/common/src/test/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequestTest.java
+++ b/common/src/test/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequestTest.java
@@ -1,0 +1,47 @@
+package org.opentestsystem.rdw.reporting.common.report;
+
+import org.junit.Test;
+import org.opentestsystem.rdw.common.model.Subject;
+
+import java.util.Locale;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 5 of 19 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Report.java
```diff
@@ -60,7 +61,7 @@ public String getLabel() {
     /**
      * @return The report request parameters
      */
-    @JsonIgnore
+    @JsonTypeInfo(use=JsonTypeInfo.Id.CLASS, include=JsonTypeInfo.As.PROPERTY, property="@class")
     public AbstractExamReportRequest getReportRequest() {
```
wtritch: We need to return this polymorphic property from the report-processor to the reporting webapp in order to provide the required Report details.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
```diff
@@ -18,6 +18,8 @@ public ExamReportAssessmentType getAssessmentType() {
         return assessmentType;
     }
 
+    public abstract ReportType getReportType();
```
wtritch: Rather than do class parsing to determine the type of report request (student, school/grade, group, exam export) I added an enum getter to the implementations.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
```diff
@@ -1,4 +1,17 @@
...
+        return ReportType.ExamExport;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder extends AbstractExamReportRequest.Builder<ExamExportQueryOptions, ExportExamReportRequest, ExportExamReportRequest.Builder> {
+
+    }
```
wtritch: I needed a Builder for testing since I want to be able to create saturated instances outside of JSON deserialization.

common/src/test/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequestTest.java
```diff
@@ -0,0 +1,47 @@
+package org.opentestsystem.rdw.reporting.common.report;
+
+import org.junit.Test;
+import org.opentestsystem.rdw.common.model.Subject;
+
+import java.util.Locale;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public abstract class AbstractExamReportRequestTest<T, A extends AbstractExamReportRequest<T>, B extends AbstractExamReportRequest.Builder<T, A, B>> {
```
wtritch: Since none of the report request implementations were being tested, I created an abstract test hierarchy to fully test the implementations without re-writing the same tests multiple times.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
```diff
@@ -26,20 +25,22 @@
 public class ExamReportController {
 
     private final ExamReportService service;
+    private final ReportResourceAssembler assembler;
 
     @Autowired
     ExamReportController(final ExamReportService service) {
         this.service = service;
+        assembler = new ReportResourceAssembler();
```
wtritch: I'm using an assembler to turn Reports into ReportResources so the front-end gets the information it needs in an easy-to-parse, flat format.

webapp/src/test/java/org/opentestsystem/rdw/reporting/report/ExamReportControllerIT.java
```diff
@@ -97,14 +101,33 @@ public void getStudentExamReportShouldSend404WhenServiceThrowsException() throws
...
-                .andExpect(jsonPath("$.id").value(1L));
+                .andExpect(jsonPath("$.*", hasSize(8)))
+                .andExpect(jsonPath("$.id").value(1L))
+                .andExpect(jsonPath("$.status").value("PENDING"))
+                .andExpect(jsonPath("$.label").value("my-label"))
+                .andExpect(jsonPath("$.reportType").value("Student"))
+                .andExpect(jsonPath("$.assessmentType").value("ICA"))
+                .andExpect(jsonPath("$.subject").value("MATH"))
+                .andExpect(jsonPath("$.schoolYear").value(2))
+                .andExpect(jsonPath("$.created").isNotEmpty());
```
wtritch: I added some tests to make sure the values I expect are being serialized as I expect.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ReportResource.java
```diff
@@ -0,0 +1,88 @@
...
+public class ReportResource {
+
+    private Long id;
+    private ReportStatus status;
+    private String label;
+    private ReportType reportType;
+    private AssessmentType assessmentType;
+    private Subject subject;
+    private Integer schoolYear;
+    private Instant created;
```
wtritch: This is the set of properties that will be displayed in the new "My Reports" table.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ReportType.java
```diff
@@ -0,0 +1,11 @@
+package org.opentestsystem.rdw.reporting.common.report;
+
+/**
+ * This enum represents a Report type.
+ */
+public enum ReportType {
+    SchoolGrade,
+    Group,
+    Student,
+    ExamExport
```
malaffoon: Is there going to be correlation to a database field for these? Are you using the string values or an integer mapping?

wtritch: No DB changes required.  It isn’t a property, each implementation returns a hard coded value.

malaffoon: So the records in the database are only discriminated by the type buried in the json blob that has the request params?

wtritch: yessir

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
```diff
@@ -1,4 +1,16 @@
 package org.opentestsystem.rdw.reporting.common.report;
 
 public class ExportExamReportRequest extends AbstractExamReportRequest<ExamExportQueryOptions> {
+
+    @Override
+    public ReportType getReportType() {
```
esotrope: not getType()?

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
esotrope: why not make it an injected component?

malaffoon: It could be but it is a pretty standard thing for a Controller to be intimately aware of its assembler(s).

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/ExportExamReportRequest.java
wtritch: I debated that, but there's also getAssessmentType() so I wanted it to be distinct.

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/ExamReportController.java
wtritch: Traditionally, using Spring Hateoas, the assemblers are injected because they require injection themselves.  Here, there wasn't really any need and it's one less bean cluttering up the spring environment.

bgsavage: So what are the criteria for injecting vs. initializing?

wtritch: In general, if it's a tightly-coupled 1-to-1 without external dependencies I'll initialize.  If two or more things depend upon it, the dependency tree is more than 1 deep (e.g. if the assembler had dependencies,) or there are external/configurable dependencies I'll inject.

PR #499, opened 2017-10-10
Add additional details to UI report table.

wtritch:
Add additional details to the "My Reports" UI table to help users distinguish between their reports.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report.model.ts
+++ b/webapp/src/main/webapp/src/app/report/report.model.ts
@@ -3,11 +3,17 @@
  * This model serves as a handle on a persisted report that can be downloaded,
  * or report request that can be reissued to regenerate a historic report for download again
  */
+import { AssessmentType } from "../shared/enum/assessment-type.enum";
+
 export class Report {
 
   public id: number;
   public label: string;
   public status: string;
+  public reportType: string;
+  public assessmentType: AssessmentType;
+  public subjectId: number;
+  public schoolYear: number;
   public created: Date;
 
   public get completed(): boolean {
--- a/webapp/src/main/webapp/src/app/report/report.service.spec.ts
+++ b/webapp/src/main/webapp/src/app/report/report.service.spec.ts
@@ -1,0 +1,113 @@
+import { MockDataService } from "../../test/mock.data.service";
+import { inject, TestBed } from "@angular/core/testing";
+import { DataService } from "../shared/data/data.service";
+import { ReportService } from "./report.service";
+import { ReportNamingService } from "./report-naming.service";
+import { Student } from "../student/model/student.model";
+import { ReportOptions } from "./report-options.model";
+import { Group } from "../user/model/group.model";
+import { Grade } from "../school-grade/grade.model";
+import { School } from "../user/model/school.model";
+import { Report } from "./report.model";
+import { Observable } from "rxjs/Observable";
+import { AssessmentType } from "../shared/enum/assessment-type.enum";
+import { AssessmentSubjectType } from "../shared/enum/assessment-subject-type.enum";
+import Spy = jasmine.Spy;
+import createSpy = jasmine.createSpy;
+
+describe('ReportService', () => {
+  let dataService: MockDataService;
+  let namingService: MockNamingService;
+  let service: ReportService;
+
+  beforeEach(() => {
+    dataService = new MockDataService();
+
+
+    TestBed.configureTestingModule({
+      providers: [
+        ReportService, {
+          provide: DataService,
+          useValue: dataService
+        }, {
+          provide: ReportNamingService,
+          useValue: namingService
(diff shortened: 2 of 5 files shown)


webapp/src/main/webapp/src/app/report/report.model.ts
```diff
@@ -3,11 +3,17 @@
...
+import { AssessmentType } from "../shared/enum/assessment-type.enum";
+
 export class Report {
 
   public id: number;
   public label: string;
   public status: string;
+  public reportType: string;
+  public assessmentType: AssessmentType;
+  public subjectId: number;
```
esotrope: not AssessmentSubjectType?

webapp/src/main/webapp/src/app/report/report.service.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      expect(reports[0].schoolYear).toBe(1234);
+      expect(reports[0].created.getTime()).toBeLessThan(new Date().getTime());
+      expect(reports[0].assessmentType).toBeUndefined();
+      expect(reports[0].subjectId).toBe(0);
+
+      expect(reports[1].assessmentType).toBe(AssessmentType.ICA);
+      expect(reports[1].subjectId).toBe(AssessmentSubjectType.MATH);
+      done();
+    });
+  });
```
wtritch: I added a test to make sure the new model properties were being properly mapped from the API

webapp/src/main/webapp/src/app/report/report.model.ts
```diff
@@ -3,11 +3,17 @@
...
  * or report request that can be reissued to regenerate a historic report for download again
  */
+import { AssessmentType } from "../shared/enum/assessment-type.enum";
+
 export class Report {
 
   public id: number;
   public label: string;
   public status: string;
+  public reportType: string;
```
esotrope: Should we add an enum for this too?

webapp/src/main/webapp/src/app/report/report.service.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  constructor() {
+    this.nameStudentExamReport.and.callFake((student: Student, options: ReportOptions) => {
+      return student.ssid;
+    });
+
+    this.nameGroupExamReport.and.callFake((group: Group, options: ReportOptions) => {
+      return group.name;
+    });
+
+    this.nameSchoolGradeExamReport.and.callFake((school: School, grade: Grade, options: ReportOptions) => {
```
esotrope: is there an easy way in jasmine to make per-test mock responses? Seems like we always use one mock class per test file

webapp/src/main/webapp/src/app/report/report.model.ts
esotrope: Ah I see -> need to represent all with 0.
We might want to make an AssessmentSubjectFilterOption enum which extends AssessmentSubjectType.
That way we don't use 0 without explanation in the code.

webapp/src/main/webapp/src/app/report/report.service.spec.ts
bgsavage: I like tests :)

webapp/src/main/webapp/src/app/report/report.model.ts
bgsavage: I second

webapp/src/main/webapp/src/app/report/report.service.spec.ts
wtritch: We can configure the spy in the individual tests.  I just like making a default impl of the mock methods to catch executions I'm unaware of.  
If more than one test uses/needs the same mock I'll move it into the test/ directory for re-use where it makes sense to have standard mock impls for the API methods.

webapp/src/main/webapp/src/app/report/report.model.ts
wtritch: hmm... AssessmentSubjectType is used to determine the id, here and elsewhere.  I'd rather not have 2 separate enum representations of the "same" value floating around.  I wonder if we just want to add "All" as enum value 0 to the existing AssessmentSubjectType?

webapp/src/main/webapp/src/app/report/report.model.ts
wtritch: hmm... I was following the same pattern as the "status" property.  Should they both be enums with associated display pipes, like AssessmentType?  Although, now that I look at it, the AssessmentTypePipe just returns the string value of the enum without going through translation...  Do we have a good example of enum-to-translated-display somewhere?

wtritch: After discussion we've decided to veer away from enums towards data-driven options.

PR #500, opened 2017-10-11
Feature/allow custom report names

wtritch:
This PR suggests a report name to the user when displaying the modal report configuration popup but allows the user to enter any name they like.

image


webapp/src/main/webapp/src/app/report/group-report-download.component.ts
```diff
@@ -29,4 +27,8 @@ export class GroupReportDownloadComponent extends ReportDownloadComponent {
     return this.service.createGroupExamReport(this.group, this.options);
   }
 
+  generateName(): string {
+    return this.group.name;
+  }
```
wtritch: I removed the name generation service, since we no longer generate the name *after* the user has configured their report.  That means we can't append things like sort-order, assessment-type filters, or other shared values to the name, so there's no shared functionality between generating a name for a group vs a school/grade report.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.spec.ts
```diff
@@ -17,34 +16,39 @@ import { User } from "../../user/model/user.model";
...
 import { MockActivatedRoute } from "../../../test/mock.activated-route";
 import { UserModule } from "../../user/user.module";
-import { RouterTestingModule } from "@angular/router/testing";
 import { OrganizationService } from "../organization.service";
+import { MockDataService } from "../../../test/mock.data.service";
+import { MockRouter } from "../../../test/mock.router";
 
 let availableGrades = [];
 
 describe('SchoolResultsComponent', () => {
```
wtritch: Unrelated test fix, I noticed a lot of test failure spam due to no handling of failed promises.  After some digging I found it was because this component uses a Router to navigate whenever filter values are updated and the previous mock router implementation couldn't resolve the routes.

webapp/src/main/webapp/src/app/report/school-grade-report-download.component.ts
```diff
@@ -25,12 +24,18 @@ export class SchoolGradeDownloadComponent extends ReportDownloadComponent {
...
     super(notificationService);
   }
 
   createReport(): Observable<Report> {
     return this.service.createSchoolGradeExamReport(this.school, this.grade, this.options);
   }
 
+  generateName(): string {
+    let gradeLabel: string = this.translate.instant(`labels.grades.${this.grade.code}.short-name`);
+    return `${this.school.name} ${gradeLabel}`;
```
esotrope: Nice. So we took the complexity of all the params out of the names.
I still think this may be better put in a service layer for testability though no?

wtritch: I debated that.  It felt more like a sub-class implementation detail than a global service, though.  (The thing that knows about groups uses groups to generate the name, the thing that knows about a school and grade uses the school and grade to generate the name, etc)

PR #502, opened 2017-10-11
Removed ExportExamsQueryOptions in favor of placing directly on Expor…

bgsavage:
…tReportRequest to fall more in line other report requests


wtritch:
Thank you for doing that!  We should go remove any old user_report values from the QA and Dev DBs as well...  but I can handle that.

bgsavage:
Sounds good 👍

PR #503, opened 2017-10-11
Updating AssessmentType to just be [Interim, Summative] and added Ass…

bgsavage:
…essmentSubtype.

### DIFF
--- a/report-processor/src/main/resources/application.sql.yml
+++ b/report-processor/src/main/resources/application.sql.yml
@@ -18,6 +18,7 @@ sql:
         'TestSessionId',
 
         'AssessmentType',
+        'AssessmentSubType',
         'AssessmentName',
         'Subject',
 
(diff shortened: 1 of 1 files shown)


report-processor/src/main/resources/application.sql.yml
```diff
@@ -68,7 +69,8 @@ sql:
         e.school_year,
         e.session_id,
 
-        at.code as asmt_code,
+        case when at.code = 'sum' then 'Summative' else 'Interim' end,
+        upper(at.code),
```
bgsavage: So during the demo, we mentioned that AssessmentType is just [ Summative, Interim ], so this is what that looks like.  I't be nicer if the asmt_type table had the "type" field to reconcile with the data dictionary, but for now I'm just doing it as a case statement.

image

PR #505, opened 2017-10-12
Feature/organization level exam export naming

esotrope:
Added OrganizationExportNamingService responsible for giving exports a default name.
Moved the Organization ID reducing logic to a new OrganizationGroupingService.
Added tests for both of these services.

This PR does not add an export name input form field.


webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.ts
```diff
@@ -0,0 +1,102 @@
...
+      return selection.firstSchool.name;
+    }
+    if (selection.schoolCount > 1) {
+      if (!isUndefined(selection.firstSchool.schoolGroupId)
+        && selection.schools.every(x => x.schoolGroupId === selection.firstSchool.schoolGroupId)) {
+        return selection.firstSchool.schoolGroup.name;
+      }
+      return selection.firstSchool.district.name;
+    }
+    return '';
```
esotrope: debated on whether to use an exception here or not. This will only happen if you give the naming service no organizations from which to base the name. Either way this behavior is asserted in the test class.

webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Support method for getSelectedOrganizationOrFirstCommonAncestorName()
+   * This method maps each organization ID to an organization object with district and schoolGroup properties to help
+   * navigate the organization ancestry chain.
+   *
+   * @param {GroupedOrganizationIds} request
+   * @param {UserOrganizations} organizations
+   * @returns {any}
+   */
+  private getSelectionWithAncestry(request: GroupedOrganizationIds, organizations: UserOrganizations) {
```
esotrope: This method puts the data in a format that makes the getSelectedOrganizationOrFirstCommonAncestorName method much more readable.

webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * If the user has selected one organization this method will return that organization's name.
+   * If the user has multiple selections it will return the name of their first common ancestor organization.
+   * This method assumes that each school has a district and that only up to one district may be selected.
+   *
+   * @param {GroupedOrganizationIds} ids
+   * @param {UserOrganizations} organizations
+   * @returns {string}
+   */
+  private getSelectedOrganizationOrFirstCommonAncestorName(ids: GroupedOrganizationIds, organizations: UserOrganizations): string {
```
esotrope: This could have been an elegant recursive method but given that the root node, middle nodes and leaf nodes would all have different behaviors anyway I just went with an explicit iterative approach.

webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  it('should return school name when one school', () => {
+    let options = {
+      schoolYear: 1000,
+      schoolIds: [ 1 ],
+      schoolGroupIds: [],
+      districtIds: []
+    };
+    expect(service.name(options, organizations))
+      .toBe(`${organizations.schoolsById.get(1).name} ${options.schoolYear}`);
```
bgsavage: Personally, I like expectations to be hardcoded strings so that way it's easy to see what the expected result is, and also eliminates any potential bugs in the calculated value.

i.e.

```
    expect(service.name(options, organizations))
       .toBe(`School 1 1000`);
```

webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.ts
bgsavage: Hm, it's also inefficient when `(selection.districtCount == 1) `.  In the spirit of efficiency and SRP, it might be better to have individual helper methods that only `getSchools()` or `getDistricts() that are only called as needed.

esotrope: I could. I'm not too concerned with it though. In that case schoolIds/schoolGroupIds.map() would be no-ops anyway.

webapp/src/main/webapp/src/app/organization-export/organization-export-options.ts
```diff
@@ -0,0 +1,35 @@
+
+/**
+ * Represents the filter options for an organization export request
+ */
+export interface OrganizationExportOptions extends GroupedOrganizationIds {
```
bgsavage: I'm still not a fan of these interfaces as models... It breaks the existing model pattern... :)

We should discuss & standardize.

PR #506, opened 2017-10-12
Added JdbcExamRepositoryIT which validates each individual field of t…

bgsavage:
…he exported CSV.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,6 +86,9 @@ subprojects {
             // pdf tooling
             dependency 'org.apache.pdfbox:pdfbox:2.0.7'
 
+            // csv tooling (used by IT)
+            dependency 'com.opencsv:opencsv:4.0'
+
             // commonly used test libraries
             dependency 'org.assertj:assertj-core:3.6.2'
             dependency 'org.mockito:mockito-all:1.10.19'
--- a/build.gradle
+++ b/build.gradle
@@ -86,9 +86,6 @@ subprojects {
             // pdf tooling
             dependency 'org.apache.pdfbox:pdfbox:2.0.7'
 
-            // csv tooling (used by IT)
-            dependency 'com.opencsv:opencsv:4.0'
-
             // commonly used test libraries
             dependency 'org.assertj:assertj-core:3.6.2'
             dependency 'org.mockito:mockito-all:1.10.19'
--- a/report-processor/build.gradle
+++ b/report-processor/build.gradle
@@ -24,7 +24,6 @@ dependencies {
     compile 'org.springframework.cloud:spring-cloud-starter-config'
     compile 'org.springframework.cloud:spring-cloud-starter-stream-rabbit'
     compile 'org.springframework.retry:spring-retry'
-    compile 'com.opencsv:opencsv:4.0'
 
     testCompile project(path: ':rdw-reporting-common', configuration: 'tests')
     testCompile 'org.assertj:assertj-core'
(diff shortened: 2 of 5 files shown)


report-processor/src/main/resources/application.sql.yml
```diff
@@ -89,8 +89,8 @@ sql:
 
         st.gender_code as student_gender_code,
 
-        (case when e.migrant_status then 'Yes' else 'No' end),
-        (case when e.section504 then 'Yes' else 'No' end),
+        (case when e.migrant_status = 1 then 'Yes' when e.migrant_status = 0 then 'No' else '' end),
+        (case when e.section504 = 1 then 'Yes' when e.section504 = 0 then 'No' else '' end),
```
bgsavage: Found a bug while writing my IT.  Since `migrant_status` and `section504` are `NULLABLE`, we need to explicitly check for `1` and `0` so that we can return `''` when it is `null`.

build.gradle
```diff
@@ -86,6 +86,9 @@ subprojects {
             // pdf tooling
             dependency 'org.apache.pdfbox:pdfbox:2.0.7'
 
+            // csv tooling (used by IT)
+            dependency 'com.opencsv:opencsv:4.0'
```
malaffoon: It doesn't matter a whole lot but the idea of this dependency section is to put things that are (or should be) used in multiple places. If this is just a dependency for a test, i'd recommend not putting this here and just including the version in the dependency.

report-processor/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
     compile 'org.springframework.cloud:spring-cloud-starter-config'
     compile 'org.springframework.cloud:spring-cloud-starter-stream-rabbit'
     compile 'org.springframework.retry:spring-retry'
+    compile 'com.opencsv:opencsv:4.0'
```
malaffoon: testCompile?

malaffoon: Oh, and you have the version here ... so you really don't need it in the dependencyManagement section of the root build file.

report-processor/src/main/resources/application.sql.yml
malaffoon: IT ftw!

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepositoryIT.java
```diff
@@ -51,106 +54,219 @@ public void Setup() {
...
+        final ExportExamReportRequest query = ExportExamReportRequest
+                .builder()
+                .schoolIds(ImmutableSet.of(-11L))
+                .schoolYear(1997)
+                .build();
+
+        repository.export(statewide(), query, testCsv);
+        final String[] row = getCsvData().get(1);
+
+        assertThat(getField(row, "DistrictId")).isEqualTo("districtNat1");
```
malaffoon: This works and no changed needed, but i would've been tempted to have a collection that is a mapping of column name to expected value, then all this could be a loop going through the entries.
```
    final Map<String, String> expectedColumnValues = builder...;
    for (final Map.Entry entry : expectedColumnValues.entrySet()) {
        assertThat(getField(row, entry.getKey())).isEqualTo(entry.getValue());
    }
```
(give or take the right calls and stuff)

bgsavage: I thought about something like that, was debating which would be easier to read.

report-processor/build.gradle
bgsavage: Ah ok got it.  Thanks.

PR #508, opened 2017-10-12
Report names now have a maxlength of 255 (matching what is in the dat…

bgsavage:
…abase) now styling to overflow with ellipsis on the My Reports page.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/report/reports.component.html
+++ b/webapp/src/main/webapp/src/app/report/reports.component.html
@@ -12,9 +12,9 @@
 
     <p-column field="label" header="{{'labels.reports.column.label.header' | translate}}" sortable="true">
       <ng-template let-report="rowData" pTemplate="body">
-        <span *ngIf="report.processing">{{report.label}}</span>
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -12,9 +12,9 @@ <h3 class="blue mb-md">{{'labels.reports.heading' | translate}}</h3>
 
     <p-column field="label" header="{{'labels.reports.column.label.header' | translate}}" sortable="true">
       <ng-template let-report="rowData" pTemplate="body">
-        <span *ngIf="report.processing">{{report.label}}</span>
-        <button *ngIf="report.completed" class="btn btn-info btn-xs" (click)="download(report)"><i class="fa fa-cloud-download mr-xs" aria-hidden="true"></i> {{report.label}}</button>
-        <button *ngIf="report.regenerable" class="btn btn-default btn-xs" (click)="regenerate(report)"><i class="fa fa-refresh mr-xs" aria-hidden="true"></i> {{report.label}}</button>
+        <div *ngIf="report.processing" [title]="report.label"  class="label-max-width">{{report.label}}</div>
```
bgsavage: Changed the span to div so we can set a max-length on it.  Also note that I'm setting the title to the label so a user can hover over it and get the full title.

PR #512, opened 2017-10-13
Consistent Naming for CSV Export

bgsavage:
Updated some of the headers of the front end csv export to more closely align with the data dictionary and what the backend csv export does.  I still have open questions when the data fields are represented entirely different...


report-processor/src/main/resources/application.sql.yml
```diff
@@ -35,12 +35,12 @@ sql:
...
-        'Migrant Status',
+        'Sex',
+        'MigrantStatus',
         'Section 504',
-        'Iep',
-        'Economic Disadvantage',
-        'Lep',
+        'IDEAIndicator',
+        'EconomicDisadvantageStatus',
+        'LEPStatus',
```
bgsavage: unrelated bugfix for dwr-1090 I noticed while I was here.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -104,7 +104,7 @@ sql:
         (case when exists (select 1 from student_ethnicity where student_id = st.id and ethnicity_code = 'TwoOrMoreRaces') then 'Yes' else 'No' end ),
         (case when exists (select 1 from student_ethnicity where student_id = st.id and ethnicity_code = 'Filipino') then 'Yes' else 'No' end ),
 
-        e.available_accommodation_codes
+        coalesce(e.available_accommodation_codes, '')
```
bgsavage: unrelated fix since I moved to using the denormalized field, I forgot to `coalesce` so that nulls show as empty strings instead of `\N` in the csv.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.spec.ts
```diff
@@ -112,6 +113,29 @@ describe('CsvBuilder', () => {
...
+  it('should add a column for each ethnicity',
+    inject([ CsvBuilder ], (builder: CsvBuilder) => {
+      let exams = [ { codes: [ "ethnicity1", "ethnicity2", "ethnicity3" ]}, { codes: [ "ethnicity7" ] }].map(x => {
+        let exam = new Exam();
+        exam.student = new Student();
+        exam.student.ethnicityCodes = x.codes;
+        return exam;
+      });
+
+      let ethnicityOptions = [ "ethnicity1", "ethnicity2", "ethnicity3", "ethnicity4", "ethnicity5", "ethnicity6", "ethnicity7" ];
```
bgsavage: This unit test demonstrates that our ethnicities are still at least data driven in the front end since the codes we have in the database are pretty close to the fields in the data dictionary for ethnicities.  The only difference being `TwoOrMoreRaces` has the code of `DemographicRaceTwoOrMoreRaces` instead, which I'm hoping we can let slide in the spirit of keeping the front end data driven and not having to introduce a translate layer that will have all the same key/values except this one.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -92,66 +94,67 @@ export class CsvBuilder {
...
-      this.translateService.instant('labels.export.cols.student-id'),
+      this.translateHeader('student-id'),
       (item) => getStudent(item).ssid
     );
   }
 
   withStudentName(getStudent: (item: any) => Student) {
     return this.withColumn(
-      this.translateService.instant('labels.export.cols.student-first-name'),
+      this.translateHeader('student-first-name'),
```
bgsavage: Minor cleanup for readability.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -306,23 +310,25 @@ export class CsvBuilder {
...
-    )
+  withEthnicity(getExam: (item: any) => Exam, ethnicities: string[]) {
+    for(let ethnicity of ethnicities) {
+      this.withColumn(
+        ethnicity,
+        (item) => {
+          let polarEnum = getExam(item).student.ethnicityCodes.some(code => code == ethnicity) ? 1 : 2;
+          return this.translateService.instant(`enum.polar.${polarEnum}`);
+        });
+    }
```
bgsavage: Now showing a column for each ethnicity with Yes / No as the value, like the district export.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -393,49 +399,53 @@ export class CsvBuilder {
...
-    this.withEthnicity(getExam);
-    return this;
+  withStudentContext(getExam: (item: any) => Exam, ethnicities) {
+    return this
+      .withMigrantStatus(getExam)
+      .with504Plan(getExam)
+      .withIep(getExam)
+      .withEconomicDisadvantage(getExam)
+      .withLimitedEnglish(getExam)
+      .withEthnicity(getExam, ethnicities);
```
bgsavage: unrelated minor cleanups.

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -58,7 +59,7 @@ export class CsvExportService {
       .withMathClaimScores(getNonIABMathExam)
       .withELAClaimScores(getNonIABElaExam)
       .withGender(getStudent)
-      .withStudentContext(getExam)
+      .withStudentContext(getExam, ethnicities)
```
bgsavage: So I debated whether or not to pass ethnicities all the way down or to inject the ExamOptionService into CsvBuilder.  I landed here because both times we call it we already have this info handy, and we don't have to introduce an observable where there currently is none.

PR #513, opened 2017-10-13
Now passing the student data to the StudentReportDownloadComponent so…

bgsavage:
… that the suggested report name can populate correctly

image

is now:
image


bgsavage:
No worries, it happens.  :)

PR #514, opened 2017-10-16
makes typeahead behavior consistent and adds distinction between schools and districts in organization dropdown

esotrope:
screen shot 2017-10-16 at 10 33 01 am

screen shot 2017-10-16 at 10 30 09 am


webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
```diff
@@ -8,11 +8,10 @@ <h3 class="blue mb-md">{{'labels.organization-export.heading' | translate}}</h3>
         <div class="col-md-6" *ngIf="!editingDisabled">
           <div class="form-group">
             <label>{{'labels.organization-export.form.organization.label' | translate}}</label>
-            <span search-select-with-button
+            <span sb-button-typeahead
```
esotrope: went ahead and renamed the typeahead to sb-* so we can bundle it with common bootstrap/angular components later.

webapp/src/main/webapp/src/app/shared/common.module.ts
```diff
@@ -85,7 +82,10 @@ import { CommonFormModule } from "app/shared/form/common-form.module";
     LoaderComponent,
     DataTableRowExpanderComponent,
     InformationLabelComponent,
-    CommonFormModule
+    CommonFormModule,
+    SBRadioButtonComponent,
+    SBCheckboxList,
+    SBToggleComponent
```
esotrope: Tried moving these components to common/forms but the angular/google analytics integration prevented me. Hopefully there is a way to decouple that more from our ui comps and possibly decorate the UI comps themselves.

webapp/src/main/webapp/src/app/shared/form/sb-button-typeahead.component.ts
```diff
@@ -0,0 +1,49 @@
...
+        <button type="button"
+                class="btn btn-default"
+                (click)="onButtonClickInternal()"
+                [disabled]="buttonDisabledInternal"><span [ngClass]="{'gray': buttonDisabledInternal}"><i class="fa fa-plus"></i> {{buttonLabel}}</span>
+        </button>
+      </span>
+    </span>
+  `
+})
+export class SBButtonTypeahead extends SBTypeahead {
```
bgsavage: Might be nice to have some jsdocs to easily be able to tell the difference between SbTypeahead and SbButtonTypeahead and to know when to use each.

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -0,0 +1,101 @@
...
+           [typeaheadMinLength]="0"
+           typeaheadOptionField="label"
+           typeaheadGroupField="group"
+           (typeaheadOnSelect)="onSelectInternal($event.item)"
+           (ngModelChange)="onChangeInternal()"
+           [(ngModel)]="search"
+           [placeholder]="placeholder">
+  `
+})
+export class SBTypeahead implements OnInit {
```
bgsavage: Curious why this couldn't be used in SbButtonTypeahead.

bgsavage:
It might be less confusing for a user if we had some wait to indicate there are more results than what is being shown.

webapp/src/main/webapp/src/app/shared/form/sb-button-typeahead.component.ts
esotrope: sounds good - will add

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
esotrope: the ```input``` has to be a direct child of ```.input-group``` to have the correct BS styles

bgsavage: could you do `<span sb-typeahead` ?

esotrope:
@bgsavage re: indicating more results.
I did test keyboard crawling the list and it doesn't let you navigate all the results when there are more. It loops to the start of the displayed list even when there are more to navigate.

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
esotrope: think it literally has to be an ```<input>``` element. only other thing i can think to do is using a directive instead of a component and have the directive extend the ngx typeahead. but that didn't seem like the best solution either.

PR #515, opened 2017-10-17
Fix reports to filter on assessment grade rather than enrolled grade

wtritch:


report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcStudentRepositoryIT.java
```diff
@@ -63,6 +64,7 @@ public void itShouldFindStudentIdsByGroupId() throws Exception {
         final ExamQueryParams g2Params = ExamQueryParams.builder()
                 .groupId(-20L)
                 .schoolYear(1997)
+                .gradeId(-1)
```
bgsavage: Would be good to get an IT around the bug scenario, but I'm guessing you'll add that in develop?

wtritch: yeah, it's already being tested with this change.  It's just really, really hard to see because we dissociate our test data from our tests, which makes these tests very hard to parse.

wtritch:
@esotrope they may already exist, the problem was only found in the reporting impl

common/src/main/resources/common.sql.yml
```diff
@@ -110,7 +110,7 @@ sql:
         and (:school_year is null or e.school_year=:school_year)
         and (:school_id is null or s.id=:school_id)
         and (:subject_id is null or a.subject_id=:subject_id)
-        and (:grade_id is null or e.grade_id=:grade_id)
+        and (:grade_id is null or a.grade_id=:grade_id)
```
malaffoon: This is an example of what i'd like to change, `:grade_id` should be `:asmt_grade_id` for clarity. But that is for a general sweep thru the code, not a hotfix.

common/src/test/java/org/opentestsystem/rdw/reporting/common/report/iab/JdbcIabReportRepositoryIT.java
```diff
@@ -217,6 +217,7 @@ public void itShouldFindAllIabReportsForStudentsBySchoolAndYear() throws Excepti
         final ExamQueryParams examQueryParams = ExamQueryParams.builder()
                 .schoolId(-10L)
                 .schoolYear(1997)
+                .gradeId(-1)
```
esotrope: should also change this and to be ```assessmentGradeId```

PR #517, opened 2017-10-17
Hotfix/1.0.4

malaffoon:
testing changes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ String dockerPrefix = "smarterbalanced"
 
 allprojects {
     group = 'org.opentestsystem.rdw.reporting'
-    version = '1.0.3-SNAPSHOT'
+    version = '1.0.4-SNAPSHOT'
 
     apply plugin: 'idea'
     apply plugin: 'propdeps'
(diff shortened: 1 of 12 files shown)


webapp/src/main/webapp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "reporting",
-  "version": "1.0.2",
+  "version": "1.0.4",
```
esotrope: 👍

build.gradle
```diff
@@ -38,7 +38,7 @@ String dockerPrefix = "smarterbalanced"
 
 allprojects {
     group = 'org.opentestsystem.rdw.reporting'
-    version = '1.0.3-SNAPSHOT'
+    version = '1.0.4-SNAPSHOT'
```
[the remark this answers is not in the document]
malaffoon: Yes. The idea is that a local developer build will always be a SNAPSHOT. The CI system replaces the placeholder with the build number, or a RELEASE/HOTFIX label.

webapp/src/main/webapp/package.json
malaffoon: Yeah, i think i just missed it last time.

PR #518, opened 2017-10-17
DWR-1089 Defect/go back or go home from org export submission broken

esotrope:


webapp/src/main/webapp/src/app/organization-export/organization-export.component.ts
```diff
@@ -201,21 +204,10 @@ export class OrganizationExportComponent implements OnInit {
...
         },
         () => this.notificationService.error({ id: 'labels.organization-export.form.submit.failure' })
       );
   }
 
-  /**
-   * Navigates user to the previous page or the home page if the current page is their first page.
-   */
-  private goBackOrHome(): void {
-    if (window.history.length == 1) {
```
esotrope: Was under the influence that this would be reliably 1 but it wasn't because of the domain of the login page matches the webapp I believe.

webapp/src/main/webapp/src/app/shared/location-support.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * True when a user clicks on a link in the app after their initial navigation to the app.
+   */
+  private hasInternalNavigationHistory;
+
+  constructor(private router: Router) {
+    router.events
+      .filter(event => event instanceof NavigationStart)
+      .subscribe(() => {
+        this.hasInternalNavigationHistory = true;
```
bgsavage: We can probably unsubscribe at this point as we got what we need.

webapp/src/main/webapp/src/app/shared/location-support.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    // This will be true when a user clicks on a link in the app after their initial navigation to the page.
+    return this.hasInternalNavigationHistory
+      // In the event that a user has internal navigation history but performed a page refresh,
+      // we can still gather this information by checking if the document referrer matches the current domain.
+      // In the case that the browser does not support document.referrer this will compute false
+      || this.referrerIsInternal;
+  }
+
+  private get referrerIsInternal(): boolean {
+    return this.getDomain(document.referrer) === this.getDomain(location.href);
```
bgsavage: It might make it easier to unit test this to pass reference wrappers for `document` and `location`

webapp/src/main/webapp/src/app/shared/location-support.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * @returns True if the previous page the user visited is within the application
+   */
+  private get previousLocationIsInternal(): boolean {
+    // This will be true when a user clicks on a link in the app after their initial navigation to the page.
+    return this.hasInternalNavigationHistory
+      // In the event that a user has internal navigation history but performed a page refresh,
+      // we can still gather this information by checking if the document referrer matches the current domain.
+      // In the case that the browser does not support document.referrer this will compute false
+      || this.referrerIsInternal;
```
bgsavage: What happens someone is sitting on the landing page, is already logged in, and then navigates directly to `/custom-reports` via bookmark or url typing.  Corner case, I know, but just a thought..

esotrope: @bgsavage @wtritch @jtreuting 

This approach is not without a loop hole.
Basically if a user navigates directly to the export page, clicks another page link, then clicks the browser back button, the app will be tricked into thinking the last page was an internal page and route the user back to the previous external url.

This could be mitigated by tracking the previous URL in a app session var and navigating directly to it. However, is that the level of complexity we want to maintain for this navigation feature?

This exercise leads me to believe that we may not want this go back or home behavior but instead
(1) have this form as a modal on the page it is launched (like student reports button)
(2) navigate to a determined page after form submission (probably the "My Reports" page)

Thoughts?

webapp/src/main/webapp/src/app/shared/location-support.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    return this.getDomain(document.referrer) === this.getDomain(location.href);
+  }
+
+  private getDomain(url: string): string {
+    if (url == null) {
+      return null;
+    }
+    return url.replace('http://','')
+      .replace('https://','')
+      .split(/[/?#]/)[0];
```
bgsavage: hmmmm, nothing angular can provide for us here?  Seems like this is something common that should already exist in some lib or framework.

webapp/src/main/webapp/src/app/shared/location-support.service.ts
bgsavage: Don't we already track the previous URL for `/session-expired` ?  Should be able to hook into what that page is doing instead.

esotrope: yes exactly. we'd have to carry a whitelist of acceptable previous urls to mitigate that which makes me think what we are doing is a little strange.

esotrope:
As discussed on DWR-1089. removed go back or home logic in favor of go to My Reports page flow.

PR #521, opened 2017-10-17
fixes enrolled grade display to include all possible values

esotrope:


common/src/main/resources/messages.properties
```diff
@@ -12,6 +12,17 @@ report.student.state.label=State:
...
+report.enrollment-grade.03=3
+report.enrollment-grade.04=4
+report.enrollment-grade.05=5
+report.enrollment-grade.06=6
+report.enrollment-grade.07=7
+report.enrollment-grade.08=8
+report.enrollment-grade.09=9
+report.enrollment-grade.10=10
+report.enrollment-grade.11=11
+report.enrollment-grade.12=12
```
wtritch: Should these be the same values from the reporting app for consistency? "G3" "G4" etc?

esotrope: Don't think so - report would read "Grade: GX" then

wtritch: Good point

wtritch:
Should the assessment grade for G11 show as HS rather than G11?  My vote is “yes” :-)

esotrope:
@wtritch only issue with that is the inconsistency in report titles that will cause. Right now G11 titles are:

```properties
#  IAB assessment label
report.grade.11.label=G11

# IAB / ICA titles
report.iab.title.grade.11=Grade 11 Interim Assessment Blocks {0}-{1}
report.ica.title.grade.11=Grade 11 Interim Comprehensive Assessment {0}-{1}
```

wtritch:
@esotrope I like that better.  The problem is that when I (and I would assume teachers) see Grade 11, they think of Juniors, not High School.  And since assessments are only scoped to High School showing Grade 11 is a little misleading.

common/src/main/resources/messages.properties
```diff
@@ -12,26 +12,30 @@ report.student.state.label=State:
...
 report.grade.06.color=blue-dark
-report.grade.07=7
 report.grade.07.label=G7
 report.grade.07.color=teal
-report.grade.08=8
 report.grade.08.label=G8
 report.grade.08.color=green
-report.grade.11=11
-report.grade.11.label=G11
+report.grade.11.label=HS
```
esotrope: @jtreuting @malaffoon @bgsavage any objections to changing the "G11" IAB label to "HS" and "Grade 11 ..." assessment titles to "High School ..."
This is how assessment grades are labelled in en.json.
```
"11": {
        "long-name": "High School",
        "short-name": "HS",
        "enrolled-name": "G11"
      }
```

jtreuting: I believe that lines up with how the assessments are named in production.  They use High School in the title of the assessment so I think that is consistent.  Surprised no one called that out previously during UAT.

PR #523, opened 2017-10-17
adds all enrolled grade translations outside of 3-12 to student reports

esotrope:


common/src/main/resources/messages.properties
```diff
@@ -22,6 +25,14 @@ report.enrollment-grade.09=9
 report.enrollment-grade.10=10
 report.enrollment-grade.11=11
 report.enrollment-grade.12=12
+report.enrollment-grade.13=13
+report.enrollment-grade.PS=PS
+report.enrollment-grade.UG=UG
+report.enrollment-grade.IT=IT
+report.enrollment-grade.PR=PR
+report.enrollment-grade.PK=PK
+report.enrollment-grade.TK=TK
```
esotrope: adds all other enrollment grades for safe measure.
This will show up at the top of the report after "Grade:"

wtritch: Should we also add these to the report.grade.* messages below? Could an assessment ever be "KG" or "PS" ?

wtritch: (Or Grade 2?)

esotrope: Don't think so but I'll go ahead and add it for another safe measure : j

bgsavage: Any safe default we could put as well in case there's something else that is missing / gets added in the future?

esotrope: Actually, problem with adding that is we don't have messages for the report body for those grades provided by SB.

All the claim messages and some other messages are asmt grade dependent.

wtritch: ahhh, well if SBAC hasn't given us the info for additional grades... /shrug

jtreuting: currently for smarter balanced they only generate 3-8 and 11 for assessments

PR #525, opened 2017-10-19
Bugfix/update section 504

bgsavage:
Changed Section504 -> Section504Status and now coalescing other nullable fields to export as an empty string instead of `/N`

### DIFF
--- a/report-processor/src/main/resources/application.sql.yml
+++ b/report-processor/src/main/resources/application.sql.yml
@@ -62,8 +62,8 @@ sql:
         s.name as school_name,
 
         st.ssid as student_ssid,
-        st.first_name as student_first_name,
-        st.last_or_surname as student_last_name,
+        coalesce(st.first_name, '') as student_first_name,
+        coalesce(st.last_or_surname, '') as student_last_name,
 
         e.completed_at,
         e.school_year,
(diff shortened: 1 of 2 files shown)


report-processor/src/main/resources/application.sql.yml
```diff
@@ -37,7 +37,7 @@ sql:
 
         'Sex',
         'MigrantStatus',
-        'Section 504',
+        'Section504Status',
```
wtritch: No test updates required?

bgsavage: I'm not validating the headers in IT, you think I should?

wtritch: ...I'm honestly not sure.  I could go either way, since it's not likely to change/fail unintentionally.

jtreuting: I think right now it doesn't break anything for us if those change, but with exports and others relying on the CSV headers for automation potentially it is the contract for the file so seems like it would be nice to have that.

bgsavage: I can certainly whip up something rather quickly

bgsavage: Done :)

PR #528, opened 2017-10-23
Feature/dwr 1112 refactor asmt result by component

bgsavage:
Pulled the Items By Points Earned table out into it's own component.

Also changed the way items csv data is exported.  It is now done through the assessmentProvider interface, similar to how data is pulled, instead of having to pass an @Output event 3 levels up.

Also added consistency in the google analytic tracking events.


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.spec.ts
```diff
@@ -78,12 +80,14 @@ describe('AssessmentResultsComponent', () => {
...
         ClaimTargetComponent,
-        ResultsByStudentComponent
+        ResultsByStudentComponent,
+        ResultsByItemComponent
       ],
       providers: [ { provide: APP_BASE_HREF, useValue: '/' },
         { provide: Angulartics2, useValue: mockAngulartics2 },
         ExamStatisticsCalculator,
         ExamFilterService,
+
```
bgsavage: Oops, I'll remove this whitespace.

webapp/src/main/webapp/src/app/assessments/results/view/results-by-item/results-by-item.component.ts
```diff
@@ -1,14 +1,120 @@
...
+    exportRequest.assessmentItems = this.filteredAssessmentItems;
+    exportRequest.pointColumns = this.pointColumns;
+    exportRequest.showAsPercent = this.showValuesAsPercent;
+
+    this.angulartics2.eventTrack.next({
+      action: 'Export ItemsByPointsEarned',
+      properties: {
+        category: 'Export'
+      }
+    });
```
bgsavage: Hm, actually this is a redundant event now.  I'll remove this.  

Unless we don't care if it was a group or school result, in that case, I could remove the tracking in the AssessmentProvider impls.

@jtreuting  thoughts?

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
```diff
@@ -48,6 +58,22 @@ export class GroupAssessmentService implements AssessmentProvider {
...
+
+    this.angulartics2.eventTrack.next({
+      action: 'Export Group Items By Points Earned',
+      properties: {
+        category: 'Export'
+      }
+    });
+
+    this.csvExportService.exportItemsByPointsEarned(exportRequest, filename);
+  }
```
bgsavage: An added benefit of putting this in the two separate impls of AssessmentProvider is it moves code out of the individual components and into the services.  We should always strive to keep our components as light as possible.

webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -34,6 +33,7 @@ export class GroupResultsComponent implements OnInit {
     this._currentGroup = value;
     if(this._currentGroup) {
       this.assessmentProvider.groupId = this._currentGroup.id;
+      this.assessmentProvider.groupName = this._currentGroup.name;
```
esotrope: Should these be set at resolve time? Also, seems odd for a component to have to do this. I would think the provider would be telling the component what it is not the other way. Maybe the word provider is just throwing me though.

bgsavage: We're actually not resolving group info on individual page load, rather, just searching for the group in the user data.

PR #529, opened 2017-10-23
sets up admin build. 

esotrope:
1. moves `admin.common.model.Organization.java` to `admin.model`
2. removes admin-common (no longer needed)
3. adds admin build to settings.gradle
4. adds .gitignore for /static dirs

Moved webapp to reporting-webapp but then reverted because there are pending PRs on it.


.gitignore
```diff
@@ -1,7 +1,6 @@
 build/
 out/
 !.mvn/wrapper/maven-wrapper.jar
-src/main/resources/static/
```
esotrope: moved this to module specific gitignore

docker/docker-compose.yml
```diff
@@ -80,3 +80,36 @@ services:
     image: smarterbalanced/wkhtmltopdf-image:latest
     ports:
       - 8082:80
+
+  admin-webapp:
+    image: smarterbalanced/rdw-admin-webapp
+    ports:
+      - 8088:8080
```
esotrope: not sure what we think about having admin in the same docker compose spec. this new port will require an IDP port config change for the redirect.

settings.gradle
```diff
@@ -1,9 +1,14 @@
...
 include ':webapp'
 include ':report-processor'
+include ':admin-webapp'
 
 rootProject.name = "rdw-reporting"
 
 rootProject.children.each {
-    it.name = rootProject.name + "-" + it.name
+    if (it.name == 'admin-webapp') {
+        it.name = "rdw-${it.name}"
```
esotrope: preserved `rdw-admin-webapp` naming instead of making it `rdw-reporting-admin-webapp`

docker/docker-compose.yml
malaffoon: I would be tempted to configure both as if they were the only webapp, then comment them both out. Add a note that you can uncomment one or the other; that running both at the same time through local-docker is not an expected thing.

settings.gradle
malaffoon: Maybe we should make it `rdw-reporting-admin`? Then we can leave the main one as `rdw-reporting-webapp` and you wouldn't need an exception here. Or change the main reporting webapp to `rdw-reporting-main` or something. Or even let admin be `rdw-reporting-admin-webapp` (i.e. remove the exception). I'm open to changing the naming convention but the goal was to have them group together using the prefix:
```
rdw
  ingest
    import
    exam-processor
    ...
  reporting
    webapp
    admin-webapp
    report-processor
```

docker/docker-compose.yml
wtritch: hmm... wonder how problematic this will be w.r.t. cookies being shared across the "localhost" domain regardless of port...
I also wonder if we want to register a second localhost Entity Provider rather than overwrite the existing on at 8080 so we can develop rdw-admin against either 8080 (like any other webapp) OR 8088 to account for the docker-compose profile...

esotrope: @wtritch seems laborious x P
I'm tempted to just maintain two separate docker compose files for now and maybe put them under their respective webapp dir. thoughts?

settings.gradle
esotrope: Yeah I can remove the exception and just make it `rdw-reporting-admin-webapp`.
Will the name change not cause a continuity issues with our current build artifacts?

malaffoon: Yes, we'll need to have Alex create a new docker hub repo and change the deployment yml's to refer to the new image name.

admin-webapp/CONTRIBUTING.md
```diff
@@ -1,119 +0,0 @@
-## RDW_Reporting for Developers
```
esotrope: removed the admin contributing and readme docs. they were ports from reporting-webapp but were never changed.

settings.gradle
```diff
@@ -1,9 +1,10 @@
 include ':common'
 include ':webapp'
 include ':report-processor'
+include ':admin-webapp'
 
 rootProject.name = "rdw-reporting"
 
 rootProject.children.each {
-    it.name = rootProject.name + "-" + it.name
+    it.name = "${rootProject.name}-${it.name}"
```
esotrope: removed special case naming for admin webapp

admin-webapp/docker-compose.yml
```diff
@@ -0,0 +1,74 @@
+# docker-compose specs for rdw admin service with the configuration server
```
esotrope: gave admin-webapp and reporting-webapp each their own docker compose file and added docs to readme saying they are not meant to be run at the same time with the current IDP configuration

settings.gradle
jtreuting: @malaffoon do you want me to reach out to Alex or will you?

README.md
```diff
@@ -86,84 +86,62 @@ brew install Caskroom/cask/wkhtmltopdf
...
-$ docker images
-REPOSITORY                          TAG                 IMAGE ID            CREATED             SIZE
-smarterbalanced/rdw-reporting-service        latest              da5207b421c0        30 seconds ago      150 MB
-<none>                              <none>              13b96a973d59        About an hour ago   140 MB
-<none>                              <none>              cb5063cbcc56        2 hours ago         140 MB
-<none>                              <none>              2236259b73f0        3 hours ago         140 MB
-```
-These can be quickly cleaned up:
-```bash
-$ docker rmi $(docker images --filter "dangling=true" -q --no-trunc)
```
esotrope: pruned the readme to try to stay relevant. I felt this kind of dips into tutorializing on docker instead of our app.

settings.gradle
malaffoon: @jtreuting i can ... i have regular docker hub email exchanges with Alex.

PR #532, opened 2017-10-24
ignore webapp/src/main/resources/static/*

dsnyder-fairway:
noticed a bunch of auto-generated changes getting picked up by git


webapp/src/main/resources/.gitignore
```diff
@@ -0,0 +1 @@
+/static/
```
esotrope: I have a pending PR with this change I'm about to merge. only diff is i added the .gitignore to the module root.

dsnyder-fairway: Thanks @esotrope, I'll decline this

PR #533, opened 2017-10-24
Bugfix/bug 1110 try sorting

bgsavage:
Now sorting the options collection when it comes in by group then by label so that the results don't flip flop while you're typing.

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
@@ -1,5 +1,7 @@
 import { Component, EventEmitter, Input, OnInit, Output, ViewChild } from "@angular/core";
 import { TypeaheadDirective } from "ngx-bootstrap";
+import { ordering } from "@kourge/ordering";
+import { byNumber, byString, join } from "@kourge/ordering/comparator";
 
 /**
  * This typeahead decorates the ngx-bootstrap typeahead directive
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -61,7 +62,14 @@ export class SBTypeahead implements OnInit {
   @Input()
   set options(options: Option[]) {
     if (this._options !== options) {
-      this._options = options ? options.concat() : [];
+      this._options = options
+        ? options
+          .sort(
+            join(
+              ordering(byString).on<Option>(o => o.group).compare,
+              ordering(byString).on<Option>(o => o.label).compare
```
esotrope: Curious what the observation was.
Does this have to do with how ngx bootstrap groups?

Also, should the typeahead be assuming we want alphabetical order? I guess it's a pretty safe assumption.
Only stipulation is that it enforces that order and there is no way to override it with this setup. thoughts?

bgsavage: I could be wrong but what I saw was ngx-bootstrap will already sort items alphabetically already within each particular, so we already enforce that order by the underlying tool we chose.

Only thing was it wasn't sorting the groups in any particular order, but by providing an ordering which puts items in the same group together ensures a consistent group order in the typeahead results.  

Let me know if that doesn't make sense and I can walk you through it.

esotrope: Only reason it was sorted was because we presorted it before passing it in i think. I don't see any sort calls in the ngx bs code.
My guess is the ngx code groups the items as it iterates down them instead of surveying all the options for available groups and placing the options in presorted group buckets, hence the flip flopping.
K good to be aware of. I'll approve.

PR #534, opened 2017-10-25
moves security components to common

esotrope:
Consolidates each webapp's `security` packages into `common.security`


admin-docker/docker-compose.yml
```diff
@@ -1,74 +0,0 @@
-# docker-compose specs for rdw admin service with the configuration server
```
esotrope: moved this to `/admin-webapp/docker-compose.yml`

admin-webapp/build.gradle
```diff
@@ -34,11 +35,6 @@ dependencies {
     compile 'com.google.guava:guava'
     compile 'mysql:mysql-connector-java'
     compile 'commons-io:commons-io'
-    compile 'org.opentestsystem.rdw.common:rdw-common-archive'
```
esotrope: removed redundant deps

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/Application.java
```diff
@@ -26,6 +27,7 @@
         StatusConfiguration.class,
         YamlPropertiesConfigurator.class,
         DataSourceConfiguration.class,
+        SecurityConfiguration.class,
```
esotrope: now that security is in common it is not auto scanned

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/security/WebSecurityConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public AuthenticationFailureHandler authenticationFailureHandler() {
+        final SimpleUrlAuthenticationFailureHandler failureHandler = new SimpleUrlAuthenticationFailureHandler();
+        failureHandler.setUseForward(true);
+        failureHandler.setDefaultFailureUrl("/error");
+        return failureHandler;
+    }
+
+    @Bean
```
esotrope: This bean and the beans below only need to be defined explicitely in admin.
The admin webapp and reporting webapp are working differently because the difference in their base packages. I plan on bringing this up in standup.

admin-webapp/src/main/resources/application.yml
```diff
@@ -97,14 +97,14 @@ cloud:
 archive:
   root: file:///tmp/
 
-application:
```
esotrope: found a would be bug here. our remote config repo was setting app.state.code instead of this prop that <i>was</i> being used by the auth service.

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/ReportingPermission.java
```diff
@@ -1,9 +1,9 @@
 package org.opentestsystem.rdw.reporting.common.security;
 
 /**
- * Security roles used by the reporting application suite.
+ * Security permissions used by the reporting application suite.
  */
-public class ReportingRoles {
+public class ReportingPermission {
```
esotrope: Renamed these so we don't confuse anyone later

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.context.annotation.Profile;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.web.authentication.AuthenticationFailureHandler;
+import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
+import org.springframework.security.web.authentication.SavedRequestAwareAuthenticationSuccessHandler;
+import org.springframework.security.web.authentication.SimpleUrlAuthenticationFailureHandler;
+
+@Profile("!test")
+@Configuration
+public class WebSecurityConfiguration extends SamlWebSecurityConfiguration {
```
esotrope: This class has all the reporting-webapp specific web sec configs.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/security/WebSecurityConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.web.authentication.AuthenticationFailureHandler;
+import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
+import org.springframework.security.web.authentication.SavedRequestAwareAuthenticationSuccessHandler;
+import org.springframework.security.web.authentication.SimpleUrlAuthenticationFailureHandler;
+
+@Profile("!test")
+@Configuration
+public class WebSecurityConfiguration extends SamlWebSecurityConfiguration {
```
esotrope: This class has all the admin-webapp specific web sec configs.

When making this, I wasn't sure why one of these has `failureHandler.setUseForward(true)` and the other didn't.

Also I noticed the admin app had the success handler routing to `"/landing"` which couldn't have been right. Now it is the default `"/"`

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -91,30 +90,28 @@
...
 
-@Profile("!test")
-@Configuration
 @EnableWebSecurity
-public class SamlWebSecurityConfiguration extends WebSecurityConfigurerAdapter {
+@Import({
+        SamlSettings.class,
+        SamlUserDetailsService.class
+})
+public abstract class SamlWebSecurityConfiguration extends WebSecurityConfigurerAdapter {
```
esotrope: made many code style changes to this class and made it abstract so the parts defining app-specific url paths could be overridden.
I noticed this class was pretty annoying to move things around in because of the way it invokes bean methods instead of requiring them as method params so i might change that down the line.

PR #535, opened 2017-10-25
excludes auto configs and disables security on report processor

esotrope:


report-processor/src/main/resources/application.yml
```diff
@@ -5,6 +5,10 @@ management:
   security:
     enabled: false
 
+security:
```
esotrope: excluding session auto config wasnt enough. for some reason spring still wanted to apply security to our report controller.

PR #536, opened 2017-10-25
Add transferAccessEnabled tenant configuration properties.

wtritch:
Inject transfer access value into SQL security param provider.

This PR is more for gathering thoughts than a ready-for-production patch.

*The problem:*
The issue is that we want to conditionally enable "transfer access" for an installation/environment.  This means that when querying for a student's exams we allow a user that has access to the student's *most recent school* to see the students entire exam history.
For example, a high school teacher with a student in their class has access to that student's entire exam history down to kindergarten.

This solution implements that configuration value as a standard spring property.  Since this is a security concern I want it injected as a standard SQL "security parameter."  We have a static utility class that already provides SQL security parameters, so I'd like to inject it there.  Unfortunately, we're using public static methods rather than an injected provider which means there's some nastiness with regards to DI.
This option is to inject the TenantProperties configuration bean into a static property on the utility class at application startup.
Another option would be to pull the three "getSecurityParameters" methods out of the QueryUtils class and into an injectable, singleton SecurityParameterProvider class that would then need to be injected to all repositories.
Another option would be to inject the TenantProperties into all repositories and set the "transfer_access" manually in each of them...

Thoughts?

(Just as an aside this has happened to me on every single project where we've used public static utility methods...)

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
@@ -1,0 +1,26 @@
+package org.opentestsystem.rdw.reporting.common.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Tenant/state configuration properties.
+ */
+@ConfigurationProperties(prefix = "tenant")
+public class TenantProperties {
+
+    private boolean transferAccessEnabled;
+
+    /**
+     * True if users with access to the students' most recent schools grant access
+     * to the entire student exam history.
+     *
+     * @return True if transfer access is enabled
+     */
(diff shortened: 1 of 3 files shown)


esotrope:
@wtritch once a utility class maintains state it no longer qualifies as a utility class I think. So like you said, I think it would be appropriate to add an `@Bean QueryHelper` instance we inject instead of using the utility class 👍

bgsavage:
I like the `SecurityParameterProvider` option as we definitely don't want to enforce state in a static class.  

I second the second option of extracting the appropriate methods out of QueryUtils and into a new @Bean that can follow our standard DI pattern.

wtritch:
It sounds like extracting the three getSecurityParameters method implementations into an injected SecurityParameterProvider is the preferred option.  I'll get to work.

PR #537, opened 2017-10-25
fixes reporting webapp landing page logic

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -450,6 +451,11 @@ protected void configure(final AuthenticationManagerBuilder auth) throws Excepti
     }
 
     @Override
+    public void configure(final WebSecurity web) throws Exception {
+        web.ignoring().antMatchers("/*.ico", "/assets/public/**");
```
esotrope: optimization to not run security filters on these static asset paths (i think actually we may not need this with boot though as everything under /static/ should already be ignored by this)

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -27,10 +27,10 @@ public AuthenticationFailureHandler authenticationFailureHandler() {
         return handler;
     }
 
-
     @Override
     protected void configure(final HttpSecurity http) throws Exception {
-        super.configure(http);
         http.authorizeRequests().antMatchers("/").permitAll();
```
esotrope: this had to come before the rest of the path matchers

jtreuting:
was this a previous issue, or an issue that arose from combining admin and reporting and security into common?

esotrope:
@jtreuting issue introduced by my change to make these share the same base conf.

PR #538, opened 2017-10-25
Exclude redis auto configurations in report-processor service.

wtritch:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ReportProcessorApplication.java
```diff
@@ -27,7 +28,9 @@
         ElastiCacheAutoConfiguration.class,
         ContextStackAutoConfiguration.class,
         SessionAutoConfiguration.class,
-        SecurityAutoConfiguration.class
+        SecurityAutoConfiguration.class,
+        RedisAutoConfiguration.class,
+        RedisRepositoriesAutoConfiguration.class
```
esotrope: As a part of this Will and I discussed breaking common out into common (common java) and common-web (common web security/session/controller) related stuff. That way the report processor wont be burdened with having to manually exclude these web related configs

PR #539, opened 2017-10-25
DWR-1109 Leave visible the "X" button to remove Districts

dsnyder-fairway:


webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
```diff
@@ -1,6 +1,6 @@
 <div *ngFor="let districtTree of model.children" class="district-tree">
 
-  <a *ngIf="districtTree.value.name && !disabled" class="link-button hover-icon-holder" (click)="onClick(districtTree.value)" href="javascript:void(0)">
+  <a *ngIf="districtTree.value.name && !disabled" class="link-button" (click)="onClick(districtTree.value)" href="javascript:void(0)">
```
esotrope: if this is the behavior we want we can probably remove this class from the school group node below and the css class altogether. i think this is its only use.

bgsavage: Do we need to remove this style from line 10 as well?

Also, can this style now be deleted?

dsnyder-fairway: ok, I'll go ahead and remove those too

PR #541, opened 2017-10-26
Feature/common web

esotrope:
Moved common.web and security to common-web module.
Be prepared to look at an endless list of import line changes : J


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/ExceptionHandlerControllerAdvice.java
```diff
@@ -1,67 +0,0 @@
-package org.opentestsystem.rdw.admin.web;
```
esotrope: moved to common-web

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/support/WebRequests.java
```diff
@@ -1,21 +0,0 @@
-package org.opentestsystem.rdw.admin.web.support;
```
esotrope: moved to common-web

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/ViewController.java
```diff
@@ -6,7 +6,7 @@
 @Controller
 public class ViewController {
 
-	@RequestMapping(value = {"/"})
+	@RequestMapping(value = {"/", "/error"})
```
esotrope: making way for addition of angular error view when things go awry on the backend and we redirect to /error

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ReportProcessorApplication.java
```diff
@@ -26,11 +22,7 @@
 
 @SpringBootApplication(exclude = {
         ElastiCacheAutoConfiguration.class,
-        ContextStackAutoConfiguration.class,
-        SessionAutoConfiguration.class,
-        SecurityAutoConfiguration.class,
-        RedisAutoConfiguration.class,
-        RedisRepositoriesAutoConfiguration.class
```
esotrope: auto conf excludes no longer needed because these jars aren't on the classpath : j

report-processor/src/main/resources/application.yml
```diff
@@ -5,10 +5,6 @@ management:
   security:
     enabled: false
 
-security:
-  basic:
-    enabled: false
```
esotrope: also no longer needed with the security/redis jars off the classpath

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/ViewController.java
bgsavage: Do you have to be authenticated to hit /index.html?

bgsavage: (If yes, this will cause infinite loop if there's an error in SSO)

SSO throws error -> forward to error -> forward to index.html -> oh this requires auth, forward to SSO -> ...

esotrope: Great point. Though we only explicitly throw an error during login for permission service not running we should probably introduce a no-auth static error page like reporting.

PR #543, opened 2017-10-26
rename admin webapp

malaffoon:
Add `skipDocker` technique (same as RDW_Ingest).

Along with this change, i need to make changes in sbac-config-repo and RDW_Deployment to use the new admin webapp with its new name.

@esotrope i don't want to mess up any PR merge you have in play so let me know so we can coordinate the checkin. It's a lot easier for me to redo my stuff.

### DIFF
--- a/admin-webapp/docker-compose.yml
+++ b/admin-webapp/docker-compose.yml
@@ -41,7 +41,7 @@ services:
       - SPRING_RABBITMQ_PASSWORD=guest
 
   webapp:
-    image: smarterbalanced/rdw-admin-webapp
+    image: smarterbalanced/rdw-reporting-admin-webapp
     ports:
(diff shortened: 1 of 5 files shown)


admin-webapp/docker-compose.yml
```diff
@@ -41,7 +41,7 @@ services:
       - SPRING_RABBITMQ_PASSWORD=guest
 
   webapp:
-    image: smarterbalanced/rdw-admin-webapp
+    image: smarterbalanced/rdw-reporting-admin-webapp
```
esotrope: does our remote config repo yaml file name need to change at all because of this?

malaffoon: yup

PR #546, opened 2017-10-26
Feature/add school to exam

wtritch:
In preparation for enabling transfer student access, we need to display an exam's associated school when listing exams in the student-grade search results, group results, etc.

In support of that, we will need the School associated with each Exam in nearly all exam query situations.  This PR embeds the associated School within the Exam model.  I've also removed the now-redundant School model from the StudentExamHistory model.


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/Application.java
```diff
@@ -2,9 +2,9 @@
 
 import org.opentestsystem.rdw.admin.repository.DataSourceConfiguration;
 import org.opentestsystem.rdw.common.status.StatusConfiguration;
-import org.opentestsystem.rdw.reporting.common.web.security.SecurityConfiguration;
-import org.opentestsystem.rdw.reporting.common.web.WebConfiguration;
 import org.opentestsystem.rdw.reporting.common.jdbc.SecurityParameterProvider;
+import org.opentestsystem.rdw.reporting.common.web.WebConfiguration;
+import org.opentestsystem.rdw.reporting.common.web.security.SecurityConfiguration;
```
wtritch: My IntelliJ is set to optimize imports alphabetically... there are a lot of these non-changes.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
```diff
@@ -30,6 +31,8 @@ private Exams(){
 
     public static <A extends Exam.Builder> A map(final ResultSet row, final A builder) throws SQLException {
 
+        final Organization school = new Organization(row.getLong("school_id"), row.getString("school_name"));
```
wtritch: Now when turning a JDBC result row into an Exam, also populate the "school" as an Organization.  I don't think we need the full School model with the districtId, group_of_schools_id, etc... especially since these models tend to be returned with a high number of instances.  Thoughts?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
```diff
@@ -1,4 +1,4 @@
-package org.opentestsystem.rdw.reporting.organization.model;
+package org.opentestsystem.rdw.reporting.common.model;
```
wtritch: Since Exam exists in common and now references an Organization I've moved the Organization model implementations into common as well.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
esotrope: there is no need for that hierarchy info outside of security checks i think. and the way we have our sql work is that if you have permission it comes back, if not, it doesn't. so we should be good

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/StudentHistoryExamWrapper.java
```diff
@@ -62,10 +51,5 @@ public Builder assessment(final Assessment assessment) {
             this.assessment = assessment;
             return this;
         }
-
-        public Builder school(final Organization school) {
-            this.school = school;
-            return this;
-        }
```
wtritch: Associated change to the front-end StudentExamHistory mapper.

webapp/src/main/webapp/src/app/assessments/model/exam.model.ts
```diff
@@ -20,4 +21,5 @@ export class Exam {
   schoolYear: number;
   claimScores: ClaimScore[];
   accommodationCodes: string[];
+  school: School;
```
wtritch: Associated change to the front-end Exam model.

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -113,6 +114,11 @@ export class AssessmentExamMapper {
     uiModel.completeness = apiModel.completenessCode;
     uiModel.schoolYear = apiModel.schoolYear;
 
+    let school: School = new School();
+    school.name = apiModel.school.name;
+    school.id = apiModel.school.id;
+    uiModel.school = school;
```
wtritch: Populate the "school" property on the front-end Exam instances.

PR #547, opened 2017-10-26
DWR-608 We should set the page title based upon the route

dsnyder-fairway:
screen shot 2017-10-26 at 1 49 41 pm
Breadcrumbs will be the modifier to the title. If no breadcrumb is present, then the title remains unchanged

### DIFF
--- a/webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
+++ b/webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
@@ -1,8 +1,10 @@
 import {Component, OnInit} from "@angular/core";
+import { Title } from "@angular/platform-browser";
 import { Router, ActivatedRoute, NavigationEnd, PRIMARY_OUTLET, UrlSegment } from "@angular/router";
 import "rxjs/add/operator/filter";
 import {Utils} from "../shared/Utils";
 import * as _ from "lodash";
+import { TranslateService } from "@ngx-translate/core";
 
 @Component({
   selector: 'breadcrumbs',
@@ -11,13 +13,18 @@ import * as _ from "lodash";
 export class BreadcrumbsComponent implements OnInit{
   breadcrumbs : Array<any> = [];
 
-  constructor(private router: Router, private activatedRoute : ActivatedRoute ) {
+  constructor(private router: Router, private activatedRoute : ActivatedRoute, private title: Title, private translateService: TranslateService ) {
   }
 
   ngOnInit(): void {
     this.router.events.filter(event => event instanceof NavigationEnd).subscribe(() => {
       let root: ActivatedRoute = this.activatedRoute.root;
+      let translatedBreadCrumbs = "";
       this.breadcrumbs = this.getBreadcrumbs(root);
+      for (let breadcrumb of this.breadcrumbs) {
+        translatedBreadCrumbs = translatedBreadCrumbs.concat(this.translateService.instant(breadcrumb.label, breadcrumb.translateParams), " > ");
+      }
+      this.title.setTitle(translatedBreadCrumbs + "Smarter Balanced | Reporting System");
     });
   }
 
--- a/webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
+++ b/webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
@@ -21,8 +21,12 @@ export class BreadcrumbsComponent implements OnInit{
       let root: ActivatedRoute = this.activatedRoute.root;
       let translatedBreadCrumbs = "";
       this.breadcrumbs = this.getBreadcrumbs(root);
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -11,13 +13,18 @@ import * as _ from "lodash";
...
+  constructor(private router: Router, private activatedRoute : ActivatedRoute, private title: Title, private translateService: TranslateService ) {
   }
 
   ngOnInit(): void {
     this.router.events.filter(event => event instanceof NavigationEnd).subscribe(() => {
       let root: ActivatedRoute = this.activatedRoute.root;
+      let translatedBreadCrumbs = "";
       this.breadcrumbs = this.getBreadcrumbs(root);
+      for (let breadcrumb of this.breadcrumbs) {
+        translatedBreadCrumbs = translatedBreadCrumbs.concat(this.translateService.instant(breadcrumb.label, breadcrumb.translateParams), " > ");
```
wtritch: Before you translate, you may want to check the breadcrumb.requiresTranslate boolean property

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -11,13 +13,18 @@ import * as _ from "lodash";
...
+      }
+      this.title.setTitle(translatedBreadCrumbs + "Smarter Balanced | Reporting System");
```
wtritch: hm... I think it might be cleaner as 
`Smarter Balanced | Reporting System | blah > blah > blah` 
rather than
`blah > blah > blah > Smarter Balanced | Reporting System`
Thoughts?

wtritch: I also wonder if the full breadcrumb might be too much text to show comfortably in the browser:
image

esotrope: I would guess the first thing someone reads in the title bar should be the current page title, but I'll defer to whatever the community convention is for this.

dsnyder-fairway: My initial choice was the former, but @aaronsleeper suggested the latter. I'd be fine with whatever the consensus is

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
esotrope: This might read better as a this.breadcrumbs.reduce() call. that way we don't need the explicit for-loop

dsnyder-fairway: if its false, just get the untranslated text?

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
wtritch: hmmm... what if we invert it all, and the last crumb is the first bit of text?
So if your breadcrumb is "major > minor > micro" in the browser, it shows as:
"micro < minor < major < Reporting System | Smarter App" in the history/title?

wtritch: @esotrope this whole shabang is the current page title... feels a little bulky, but more informative than just the last crumb...

esotrope: there is no way to separate bookmark name from page title? i feel like when i'm in the app i just need the whole app title displayed in the tab but when i create a bookmark i want the whole sshabang or just the current page appended to the app name.

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
wtritch: yeah.  Looking at app.routes.ts it looks like only one crumb isn't being translated:
`breadcrumb: { resolve: 'school.name' }`
I don't think pushing it through translation *should* cause any issues...  But it makes my eye twitch pushing school names through the translation pipe. =)

esotrope: that would be funny if one of the names matched one of our message keys though wouldn't it. definitely don't want to open that door x J

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
wtritch: Not that I'm aware of...  @aaronsleeper ?

dsnyder-fairway:
screen shot 2017-10-26 at 2 54 23 pm
latest implementation

PR #552, opened 2017-10-26
Bugfix/fix test imports

wtritch:
Our controller ITs use the Application class to initialize the test application context.  We had too many @Imports on our application classes, leading to standing up nearly the entire application and causing issues with missing production configuration values when attempting to run the controller integration tests.
This fix moves our @Imports off of the Application classes and into associated ApplicationConfiguration classes.


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/Application.java
```diff
@@ -23,17 +13,6 @@
 })
 @EnableCaching
```
esotrope: wondering if all but @SpringBootApplication belong in the AppConfig.java as well

wtritch: I debated it, but felt that these annotations were safe enough to leave

PR #554, opened 2017-10-27
Implement school grade transfer student access

wtritch:


webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/JdbcSchoolGradeAssessmentRepositoryIT.java
```diff
@@ -24,6 +27,9 @@
 public class JdbcSchoolGradeAssessmentRepositoryIT {
 
     @Autowired
+    private TenantProperties tenantProperties;
```
wtritch: @bgsavage this is another argument for using configuration properties.  It allows fiddling with the values mid-test without requiring a bunch of different test application.yml files.

webapp/src/main/resources/application.sql.yml
```diff
@@ -321,11 +321,23 @@ sql:
         where exists (
             select e.asmt_id
             from exam e
-              join school s on e.school_id=s.id
+              join school s on e.school_id = s.id
+              join student st on e.student_id = st.id
+              join school isch on st.inferred_school_id = isch.id
```
wtritch: This is what breaks current data.  We're now joining into the inferred school in order to return inferred school results.  The st.inferred_school_id *will* always be populated in "real life" so I'd prefer not to do a LEFT JOIN here.  Current developer test data does not have this field populated, however...

webapp/src/main/resources/application.sql.yml
```diff
@@ -321,11 +321,23 @@ sql:
...
+              )
+              or
+              (
+                :allow_transfer_access is true
+                and isch.id = :school_id
+                and (1=:statewide or isch.district_group_id in (:district_group_ids) or isch.district_id in (:district_ids) or isch.school_group_id in (:school_group_ids) or isch.id in (:school_ids))
+              )
+            )
         )
         and a.grade_id=:grade_id
```
wtritch: Note that I didn't update the `sql.schoolgrade.exam.findLatest` query to support student transfer because the "latest" exam for students are what defines their inferred school ids.

webapp/src/main/resources/application.sql.yml
```diff
@@ -321,11 +321,23 @@ sql:
...
+            and (
+              (
+                e.school_id=:school_id
+                and (1=:statewide or s.district_group_id in (:district_group_ids) or s.district_id in (:district_ids) or s.school_group_id in (:school_group_ids) or e.school_id in (:school_ids))
+              )
+              or
+              (
+                :allow_transfer_access is true
+                and isch.id = :school_id
+                and (1=:statewide or isch.district_group_id in (:district_group_ids) or isch.district_id in (:district_ids) or isch.school_group_id in (:school_group_ids) or isch.id in (:school_ids))
```
agorina: I am just trying to understand: when we have `school_id` in here, do we know that a teacher has this school in the tenancy chain? If the answer is yes, then I do not think we need the last big AND.  Am I missing something?

webapp/src/main/resources/application.sql.yml
agorina: I think it deserves a comment in this file for whoever comes next.

wtritch: Good call, added a comment:
```
      # NOTE that findLatest does not sample the students' inferred schools
      # because a student's inferred school is determined by their latest exam.
```

jtreuting: i haven't thought it all the way through but doesn't the findLatest get the most recent single exam for the group or school/grad, determine what assessment it is for and then return all the results for that assessment.  The results returned aren't necessarily the latest for each student.  The latest for a student could have been taken at a different school i guess.  Does that impact the logic needed?

webapp/src/main/resources/application.sql.yml
```diff
@@ -321,11 +321,23 @@ sql:
...
-            and (1=:statewide or s.district_group_id in (:district_group_ids) or s.district_id in (:district_ids) or s.school_group_id in (:school_group_ids) or e.school_id in (:school_ids))
             and a.id = e.asmt_id
+            and (
+              (
+                e.school_id=:school_id
+                and (1=:statewide or s.district_group_id in (:district_group_ids) or s.district_id in (:district_ids) or s.school_group_id in (:school_group_ids) or e.school_id in (:school_ids))
+              )
+              or
+              (
+                :allow_transfer_access is true
```
esotrope: is this synonymous with `:allow_transfer_access=1` ?

wtritch: Yup, because mysql uses tinyint as a boolean.

webapp/src/main/resources/application.sql.yml
wtritch: You are correct, but it doesn't impact the logic because the school_id of the most recent exam for a student is always equal to the student's inferred_student_id, so there's no need to check them both.

webapp/src/main/resources/application.sql.yml
jtreuting: yeah that was new to me as well

bgsavage: Should we favor consistency here whereas everywhere else is `:statewide = 1` ?

wtritch: Personally, I think "is true" reads better, but I like consistency more.  Will convert this to int comparison in the next PR.

webapp/src/main/resources/application.sql.yml
chdietz: @wtritch do you mean inferred_school_id?

wtritch: @chdietz doh, yes.  Updated my comment

PR #556, opened 2017-10-27
Support transfer student access when querying by student

wtritch:
This PR adds support for student transfer access when querying by student.

This is slightly different from the two previous PRs, since we can query by student using both individual PII rights as well as group PII rights.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
@@ -54,7 +54,9 @@ public class SecurityParameterProvider {
                     securityParameters.putAll(getSecurityParameters(scope, value));
                 });
 
-        return securityParameters.build();
+        return securityParameters
+                .put("allow_transfer_access", tenantProperties.isTransferAccessEnabled())
+                .build();
     }
 
     /**
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -56,17 +56,24 @@ sql:
       from exam e
         join student st on st.id = e.student_id
         join asmt a on e.asmt_id=a.id
-        left join instructional_resource ir on ir.name = a.name
         join school s on e.school_id=s.id
+        join school isch on st.inferred_school_id = isch.id
+        left join instructional_resource ir on ir.name = a.name
       where e.student_id=:student_id
         and
         (
-          (1 = :individual_statewide
-            or s.district_group_id in (:individual_district_group_ids)
-            or s.district_id in (:individual_district_ids)
-            or s.school_group_id in (:individual_school_group_ids)
-            or e.school_id in (:individual_school_ids))
-          or (
+          # Individual permissions test
+          (
+            (1 = :individual_statewide or s.district_group_id in (:individual_district_group_ids) or s.district_id in (:individual_district_ids) or s.school_group_id in (:individual_school_group_ids) or e.school_id in (:individual_school_ids))
+            or
+            (
+              1 = :allow_transfer_access
+              and (isch.district_group_id in (:individual_district_group_ids) or isch.district_id in (:individual_district_ids) or isch.school_group_id in (:individual_school_group_ids) or isch.id in (:individual_school_ids))
+            )
+          )
+          or
+          # Group permissions test
+          (
             exists(
                 select 1
                 from student_group_membership sgm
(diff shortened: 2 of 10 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
```diff
@@ -54,7 +54,9 @@ public SecurityParameterProvider(final TenantProperties tenantProperties) {
                     securityParameters.putAll(getSecurityParameters(scope, value));
                 });
 
-        return securityParameters.build();
+        return securityParameters
+                .put("allow_transfer_access", tenantProperties.isTransferAccessEnabled())
+                .build();
```
wtritch: When querying by *either* individual or group rights, we create two security "parameter sets:"
individual_allow_transfer_access, individual_statewide, individual_district_ids, etc
group_allow_transfer_access, group_statewide, group_district_ids, etc

Since both individual_allow_transfer_access and group_allow_transfer_access will always be the same value I wanted to keep the SQL clean and consistent with a single allow_transfer_access.

webapp/src/main/resources/application.sql.yml
```diff
@@ -56,17 +56,24 @@ sql:
...
-          or (
+          # Individual permissions test
+          (
+            (1 = :individual_statewide or s.district_group_id in (:individual_district_group_ids) or s.district_id in (:individual_district_ids) or s.school_group_id in (:individual_school_group_ids) or e.school_id in (:individual_school_ids))
+            or
+            (
+              1 = :allow_transfer_access
+              and (isch.district_group_id in (:individual_district_group_ids) or isch.district_id in (:individual_district_ids) or isch.school_group_id in (:individual_school_group_ids) or isch.id in (:individual_school_ids))
+            )
+          )
```
wtritch: I felt like this read better with the single long permission comparator rather than the multi-line version...

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/group/JdbcGroupAssessmentRepositoryIT.java
```diff
@@ -32,6 +33,11 @@
     @Autowired
     private JdbcGroupAssessmentRepository repository;
 
+    @After
+    public void clearState() {
+        tenantProperties.setTransferAccessEnabled(false);
+    }
```
wtritch: Now that I'm running multiple tests within the same test class that are dependent upon the tenantProperties.isTransferAccessEnabled value, I realized that I need to clean up between tests... oops.

webapp/src/test/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcExamItemRepositoryIT.java
```diff
@@ -74,4 +86,33 @@ public void itShouldRetrieveExamItemsForStudentAndExam() throws Exception {
...
+                .build();
+
+        final List<ExamItem> withoutTransfer = repository.findAllByExamId(permissions(groupOf(school4Scope)), ImmutableSet.of(-100L), -100);
+        assertThat(withoutTransfer).isEmpty();
+
+        tenantProperties.setTransferAccessEnabled(true);
+        final List<ExamItem> withTransfer = repository.findAllByExamId(permissions(groupOf(school4Scope)), ImmutableSet.of(-100L), -100);
+        assertThat(withTransfer.stream().map(ExamItem::getExamId))
+                .containsOnly(-100L);
+    }
```
wtritch: I added tests for both the individual rights access and group rights access.

webapp/src/main/resources/application.sql.yml
esotrope: same. it will also be easier to find/replace/modify later too if it's always the same blurb

webapp/src/main/resources/application.sql.yml
```diff
@@ -131,15 +140,22 @@ sql:
...
-            (1 = :individual_statewide
-              or s.district_group_id in (:individual_district_group_ids)
-              or s.district_id in (:individual_district_ids)
-              or s.school_group_id in (:individual_school_group_ids)
-              or e.school_id in (:individual_school_ids))
-            or (
+            # Individual permissions test
+            (
+              (1 = :individual_statewide or s.district_group_id in (:individual_district_group_ids) or s.district_id in (:individual_district_ids) or s.school_group_id in (:individual_school_group_ids) or e.school_id in (:individual_school_ids))
+              or
```
esotrope: wondering, logically can't we bring this nested or up a level and remove the redundant one nested below in the group check?

wtritch: not really, because we're only allowed to check the inferred school against group permissions if the student is actually a member of a group that you have rights to (the exists check)

PR #557, opened 2017-10-27
Add School to CSV Exports.

wtritch:


webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -90,6 +90,7 @@ export class CsvExportService {
       .withFilename(filename)
       .withStudent(getStudent)
       .withExamDateAndSession(getExam)
+      .withSchool(getExam)
```
jtreuting: i'm surprised this wasn't already here as it was in the UI for student test history if i am remembering correctly.  oops.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -551,6 +551,7 @@
         "student-first-name": "FirstName",
         "student-last-name": "LastOrSurname",
         "submit-date-time": "SubmitDateTime",
+        "school": "School",
```
jtreuting: let's use `SchoolName` so it matches the data model.  http://www.smarterapp.org/documents/TestResults-DataModel.pdf

wtritch: sounds good, done

PR #559, opened 2017-10-30
DWR-1097 corrected aXe audits except color contrast corrections

dsnyder-fairway:
note: this address *most* of DWR-1097. I spoke with Nohemi to get some additional details how to access another bullet I had trouble finding. I'll either include the changes here or issue follow up PR, depending on if I get an approval or two (or if anyone says to hold off until then :))

### DIFF
--- a/webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
+++ b/webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
@@ -7,8 +7,9 @@
       <div class="row">
         <div class="col-md-6" *ngIf="!editingDisabled">
           <div class="form-group">
-            <label>{{'labels.organization-export.form.organization.label' | translate}}</label>
+            <label for="search">{{'labels.organization-export.form.organization.label' | translate}}</label>
             <span sb-button-typeahead
+                  forid="search"
                   [options]="organizationOptions"
                   (buttonClick)="add($event)"
                   placeholder="{{'prompt.select' | translate}}"
--- a/webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
+++ b/webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
@@ -9,7 +9,7 @@
           <div class="form-group">
             <label for="search">{{'labels.organization-export.form.organization.label' | translate}}</label>
             <span sb-button-typeahead
-                  forid="search"
+                  inputId="search"
                   [options]="organizationOptions"
                   (buttonClick)="add($event)"
                   placeholder="{{'prompt.select' | translate}}"
--- a/webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
@@ -11,7 +11,7 @@ import { byString, join } from "@kourge/ordering/comparator";
   selector: 'sb-typeahead,[sb-typeahead]',
   template: `
     <input class="form-control"
-           [id]="forid"
+           [id]="inputId"
            [disabled]="disabledInternal"
            [typeahead]="options"
            [typeaheadMinLength]="0"
@@ -49,7 +49,7 @@ export class SBTypeahead implements OnInit {
   search: string;
 
   @Input()
-  forid: string;
+  inputId: string;
 
   private _options: Option[] = [];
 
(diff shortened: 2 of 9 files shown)


webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -47,6 +48,9 @@ export class SBTypeahead implements OnInit {
   @Input()
   search: string;
 
+  @Input()
+  forid: string;
```
bgsavage: I think naming should be camelCasing so `forId` would be better, and you'd set it using `[for-id]`.  That being said I think we should prefer a more descriptive name.  Since this is the id for the `<input>`, how about `inputId`?

dsnyder-fairway: I like that. `inputId` it is

jtreuting: Isn't this really just the `id` or the `elementId`?  i think that it's only related to the `for` passively, meaning that it is needed because of the `for` on the label but it has other purposes so I think the for part confused me as to what it actually did.

dsnyder-fairway: it is the `id`. Part of the reasons why I give this a different name (`forid`) is that we would end up with duplicate ids here https://github.com/SmarterApp/RDW_Reporting/pull/559/files#diff-4afe9de9a4c1cc3c1a8a9b83fe99a825. I would be fine with giving this a new name (such as `inputId`), but I do not know how to simply refer to this as `id` without duplicates

jtreuting: `inputId` is great

dsnyder-fairway: changed to `inputId` just a couple moments ago

webapp/src/main/webapp/src/assets/public/landing.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
wtritch: I haven't read up on best-practices, but should this be set to the current translation language?

dsnyder-fairway: I set this as a default (en) value. We could probably create an additional ticket to address landing.html as it is static html and doesn't currently support any translations

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -47,6 +48,9 @@ export class SBTypeahead implements OnInit {
   @Input()
   search: string;
 
+  @Input()
+  inputId: string;
```
esotrope: Might want to just call this `id`. the "input" tells the consumer what is being used in the implementation and could change.

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
```diff
@@ -7,8 +7,9 @@ <h3 class="blue mb-md">{{'labels.organization-export.heading' | translate}}</h3>
       <div class="row">
         <div class="col-md-6" *ngIf="!editingDisabled">
           <div class="form-group">
-            <label>{{'labels.organization-export.form.organization.label' | translate}}</label>
+            <label for="search">{{'labels.organization-export.form.organization.label' | translate}}</label>
```
esotrope: Just curious - Is this for compliance or testability?

jtreuting: compliance.  i think it relates to being able to navigate and screen readers.  when the focus goes to an input the screen reader will know specifically what label to read to the user.

esotrope: Cool thanks. Wonder if there is tooling to spew out warnings about this kind of thing when we build/test the angular app.

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
dsnyder-fairway: part of the reasons why I give this a different name (`inputId`) is that we would end up with duplicate ids here https://github.com/SmarterApp/RDW_Reporting/pull/559/files#diff-4afe9de9a4c1cc3c1a8a9b83fe99a825

esotrope: Ah ok. Yeah, because angular will think you are talking about the native element ID attribute.

PR #561, opened 2017-10-30
DWR-1097 Set iframe title for irisframe

dsnyder-fairway:
I think other items listed under "School and Grade results page when expanding Item Viewer" are issues with IRIS

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -304,6 +304,9 @@
               "full-credit-info": "Percentage of students in the group who earned the maximum number of points available for a particular test item.",
               "x-points": "{{id}}"
             },
+            "item-viewer": {
+              "title": "Item Viewer"
+            },
             "target": "Target {{target}}"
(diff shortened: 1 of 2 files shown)


dsnyder-fairway:
changes here are pretty small. Merging

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -304,6 +304,9 @@
               "full-credit-info": "Percentage of students in the group who earned the maximum number of points available for a particular test item.",
               "x-points": "{{id}}"
             },
+            "item-viewer": {
+              "title": "Item Viewer"
```
jtreuting: @dsnyder-fairway is there an existing name that could be used for this?  Isn't there a name for the tab that maybe could be used?

dsnyder-fairway: It looks like you are right. There is `labels.assessments.items.tabs.item-viewer`. It looks like that name follows a more logical path too. I'll correct this

PR #563, opened 2017-10-31
DWR-1098 changes to meet WCAG 2 AA contrast ratio thresholds

dsnyder-fairway:
Let me know if you wants some screen shots of anything. This meets the contrast ratio thresholds for Reporting

### DIFF
--- a/report-processor/src/main/resources/templates/iab-body.html
+++ b/report-processor/src/main/resources/templates/iab-body.html
@@ -76,7 +76,7 @@
                                     </div>
                                 </div>
                                 <div th:if="${exam.scaleScore == null}" class="text-center mt-sm ml-xl">
-                                    <span class="gray-darker h2" th:text="#{report.not-scored}"></span>
+                                    <span class="black h2" th:text="#{report.not-scored}"></span>
                                 </div>
                             </div>
                         </div>
--- a/report-processor/src/main/resources/templates/iab-body.html
+++ b/report-processor/src/main/resources/templates/iab-body.html
@@ -76,7 +76,7 @@
                                     </div>
                                 </div>
                                 <div th:if="${exam.scaleScore == null}" class="text-center mt-sm ml-xl">
-                                    <span class="black h2" th:text="#{report.not-scored}"></span>
+                                    <span class="gray-darker h2" th:text="#{report.not-scored}"></span>
                                 </div>
                             </div>
                         </div>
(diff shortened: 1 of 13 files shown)


report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -76,7 +76,7 @@
                                     </div>
                                 </div>
                                 <div th:if="${exam.scaleScore == null}" class="text-center mt-sm ml-xl">
-                                    <span class="gray-darker h2" th:text="#{report.not-scored}"></span>
+                                    <span class="black h2" th:text="#{report.not-scored}"></span>
```
wtritch: I don't think we need to specify "black" as it should be the default?

dsnyder-fairway: I'll remove these

dsnyder-fairway:
> jtreuting reviewed 17 hours ago View changes 
> overall looks good. couple questions. has @aaronsleeper looked at the new colors and given the okay?
> 
> also seems like we may want to move the changes to styles.less to the sbac-ui-kit. @aaronsleeper thoughts? we could always move them there later if needed i guess.

I haven't shared screens with @aaronsleeper yet. I'll schedule something today to go over that

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -15,11 +15,11 @@
       </label>
       <div id="btn-group-off-grade-assessment" class="btn-group btn-group-sm toggle-group" data-toggle="buttons">
         <label class="btn btn-primary" [ngClass]="{'active': !filterBy.offGradeAssessment}">
-          <input [value]="false" [(ngModel)]="filterBy.offGradeAssessment" type="radio"
+          <input [value]="false" [(ngModel)]="filterBy.offGradeAssessment" type="radio" name="boolean-toggle"
```
dsnyder-fairway: I removed boolean-toggle from these in #559, but I noticed this caused the toggle in the upper right (X) to do nothing when clicked

jtreuting: does it cause any accessibility warnings with adding it back in?  i'm guessing no since it was probably around duplicate names before

report-processor/src/main/resources/templates/iab-body.html
dsnyder-fairway: removed changes to report-processor. Not sure how we are supposed to validate the changes to this yet

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
dsnyder-fairway: hmmm, it did a problem back. I experimented with this more and found it was on a page that my original testing wasn't even against (I added this back thinking it was against a different page, but it wasn't). I was able to remove this and it had no impact on either page

dsnyder-fairway:
merging this

PR #564, opened 2017-11-01
Send current language as report request language

wtritch:
While working on installing a new language, I ran into a small issue with the reporting webapp not sending the current language as the report generation language.

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-download.component.ts
+++ b/webapp/src/main/webapp/src/app/report/report-download.component.ts
@@ -7,6 +7,7 @@ import { ReportOrder } from "./report-order.enum";
 import { ModalDirective } from "ngx-bootstrap";
 import { Observable } from "rxjs";
 import { Report } from "./report.model";
+import { TranslateService } from "@ngx-translate/core";
 
 /**
  * Abstract class used to carry the common logic between all exam report download components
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -54,19 +55,19 @@ export abstract class ReportDownloadComponent implements OnInit {
   AssessmentType: any = AssessmentType;
   assessmentTypes: AssessmentType[] = [ null, AssessmentType.IAB, AssessmentType.ICA ];
   subjectTypes: AssessmentSubjectType[] = [ null, AssessmentSubjectType.MATH, AssessmentSubjectType.ELA ];
-  languages: string[] = [ 'eng', 'spa', 'vie' ];
```
wtritch: No longer recording available languages here.  As long as the translation service is set to a language, it must be available.

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -54,19 +55,19 @@ export abstract class ReportDownloadComponent implements OnInit {
...
+              protected translateService: TranslateService) {
   }
 
   ngOnInit(): void {
     let defaultOptions: ReportOptions = new ReportOptions();
     defaultOptions.assessmentType = this.assessmentType != null ? this.assessmentType : this.assessmentTypes[ 0 ];
     defaultOptions.subject = this.subject != null ? this.getSubjectFromString(this.subject) : this.subjectTypes[ 0 ];
     defaultOptions.schoolYear = this.schoolYear != null ? this.schoolYear : this.schoolYears[ 0 ];
-    defaultOptions.language = this.languages[ 0 ];
+    defaultOptions.language = this.translateService.currentLang;
```
wtritch: Setting the report language to the current webapp language.

malaffoon:
Just checking: should the current browser language be used as the reporting language?

jtreuting:
@wtritch I believe we can close this as the language of the current user is not the same as the language of the student report.  An interesting question is, do we want to default to the current language of the user when the language dropdown is displayed on this form?  But changes will need to be made to bring the language option dropdown back when there are multiple languages installed for PDF generation.

wtritch:
After investigation and discussion, it turns out we *don't* want the PDFs to be in the same language as the webapp.
The available languages for a Report may be different from the available languages for the Reporting Webapp, which may be differ from the available languages for the Admin Webapp.

PR #565, opened 2017-11-01
DWR-1134 - added changes from DWR-608 to admin site

dsnyder-fairway:


jtreuting:
This is just making the same change on the admin app side of things?

dsnyder-fairway:
> jtreuting commented just now
> This is just making the same change on the admin app side of things?

yup

jtreuting:
Someone that is stronger at Angular than me should answer this (@esotrope or @bgsavage maybe), is it possible to pull this code up a level and share it across the 2 applications we have?  Since I do think it is the same code.  If it is simple that might be nice, otherwise I say leave it as-is since it's only shared in 2 places.

esotrope:
@jtreuting working on the common-ngx lib for that now.
I think for now we can put it in admin app and delegate to the common-ngx lib later 👍

PR #567, opened 2017-11-03
DWR-1070 export distractor analysis as csv

dsnyder-fairway:


webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
```diff
@@ -75,6 +72,20 @@ export class GroupAssessmentService implements AssessmentProvider {
     this.csvExportService.exportItemsByPointsEarned(exportRequest, filename);
   }
 
+  private getFilename(exportRequest: ExportRequest) {
```
dsnyder-fairway: I would have preferred extracting this to a different class and handling slightly different for school assessments, but ultimately decided against that

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
```diff
@@ -75,6 +72,20 @@ export class GroupAssessmentService implements AssessmentProvider {
...
+    let assessment: Assessment = exportRequest.assessment;
+    let filename: string = this.groupName +
+      "-" + assessment.name;
+    if (exportRequest.type) {
+      filename = filename + "-" + exportRequest.type.toString() + "-" + new Date().toDateString();
+    } else {
+      filename = filename +
```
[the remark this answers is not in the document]
dsnyder-fairway: it shouldn't ever get hit. I did this to avoid the possibility of someone not passing in a type. Thinking about this again it probably makes more sense to remove the if-else blocks and just let developers get an error if they forgot to set the type since this ends up potentially hiding a mistake they might have made

dsnyder-fairway:
> @jtreuting approved these changes 5 minutes ago View changes 
> One question semi-related to this PR :)
> 
> Items By Points Earned is the old name and the UI has shifted to "Results by Items" . I know that it doesn't impact anything but is now a bit confusing since they don't match. Just curious if you could give a a rough idea of what it would take to update the naming to be consistent. Might be worth considering if it isn't a heavy lift.

I don't think it should be difficult at all. I'll update this

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
```diff
@@ -75,6 +72,20 @@ export class GroupAssessmentService implements AssessmentProvider {
...
+        "-ItemsByPoints" +
```
esotrope: Anything with non symbolic text should go through the translate service.
Also, you can take advantage of the es6 string interpolation using ``` `${var}-${var2}` ```

esotrope: Also, `exportRequest.type` should probably go through the translate service.

dsnyder-fairway: I'll make the `exportRequest.type` go through the translate service. I'll also be removing the if-else block too and just create the filename on one line

esotrope: 👍

PR #568, opened 2017-11-03
Support generating a report in multiple languages.

wtritch:
This PR adds a language selector to the PDF report generator modal dialog IFF more than english is available as a language.

I've also added a configurable message-source for providing additional languages/english overrides to the Reporting Common translation message source.  This was tested using a local file-system path but should allow access to remote config-server properties files as well.

Sample changes to the reporting webapp configuration:
```
  reportservice:
    url: http://localhost:8085
    languages:
      - es
      - vi
```

Sample changes to the report-processor configuration:
```
tenant:
  translation-location: file:///my/awesome/language/location/
```

Note that I've changed the reporting webapp configuration properties from @Value annotations to a ConfigurationProperties implementation.  @Value does *not* deal well with parsing collections out of configuration properties.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
@@ -10,6 +10,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 public class TenantProperties {
 
     private boolean transferAccessEnabled;
+    private String translationLocation;
 
     /**
      * True if users with access to the students' most recent schools grant access
@@ -24,5 +25,18 @@ public class TenantProperties {
 
     public void setTransferAccessEnabled(final boolean transferAccessEnabled) {
         this.transferAccessEnabled = transferAccessEnabled;
+    }
+
+    /**
+     * This is the location/prefix of translation files (messages_es.properties, es.json, etc)
+     * (ex: 'binary-${spring.cloud.config.uri}/*&#47;*&#47;master/')
+     * @return The prefix for translation resources.
+     */
+    public String getTranslationLocation() {
+        return translationLocation;
+    }
+
+    public void setTranslationLocation(final String translationLocation) {
+        this.translationLocation = translationLocation;
     }
 }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.reporting.common.i18n;
 
+import org.opentestsystem.rdw.reporting.common.configuration.TenantProperties;
 import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.context.MessageSource;
@@ -7,7 +8,11 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.support.AbstractMessageSource;
+import org.springframework.context.support.ReloadableResourceBundleMessageSource;
 import org.springframework.context.support.ResourceBundleMessageSource;
+import org.springframework.core.io.ResourceLoader;
+
+import static org.springframework.util.StringUtils.isEmpty;
 
 @Configuration
 @EnableCaching
@@ -19,7 +24,9 @@ import org.springframework.context.support.ResourceBundleMessageSource;
 public class TranslationConfiguration {
 
     @Bean
-    public MessageSource messageSource(final InternalTranslationService translationService) {
+    public MessageSource messageSource(final InternalTranslationService translationService,
+                                       final ResourceLoader resourceLoader,
+                                       final TenantProperties tenantProperties) {
 
         final ResourceBundleMessageSource baseMessages = new ResourceBundleMessageSource();
         baseMessages.setBasenames("messages");
@@ -27,7 +34,16 @@ public class TranslationConfiguration {
 
         final AbstractMessageSource databaseMessages = new TranslationServiceMessageSource(translationService);
         databaseMessages.setParentMessageSource(baseMessages);
-        return databaseMessages;
+
+        if (isEmpty(tenantProperties.getTranslationLocation())) {
+            return databaseMessages;
+        }
+
+        final ReloadableResourceBundleMessageSource configMessages = new ReloadableResourceBundleMessageSource();
+        configMessages.setResourceLoader(resourceLoader);
+        configMessages.setBasename(tenantProperties.getTranslationLocation() + "messages");
+        configMessages.setParentMessageSource(databaseMessages);
+        return configMessages;
     }
 
 }
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/configuration/ApplicationProperties.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/configuration/ApplicationProperties.java
@@ -1,0 +1,170 @@
+package org.opentestsystem.rdw.reporting.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ *  Reporting Web Application Properties
+ */
+@ConfigurationProperties(prefix = "app")
+public class ApplicationProperties {
+
+    private IrisProperties iris = new IrisProperties();
+    private AnalyticsProperties analytics = new AnalyticsProperties();
+    private ReportServiceProperties reportservice = new ReportServiceProperties();
+    private String adminWebappUrl;
+    private ExternalLinksProperties externalLinks = new ExternalLinksProperties();
+    private int minItemDataYear;
+    private List<String> uiLanguages = new ArrayList<>();
+
+    /**
+     * @return Iris configuration properties
+     */
+    public IrisProperties getIris() {
+        return iris;
+    }
+
+    /**
+     * @return Google Analytics properties
+     */
+    public AnalyticsProperties getAnalytics() {
+        return analytics;
+    }
+
+    /**
+     * @return Report processor properties
+     */
+    public ReportServiceProperties getReportservice() {
+        return reportservice;
+    }
+
+    /**
+     * @return The reporting admin url
+     */
+    public String getAdminWebappUrl() {
+        return adminWebappUrl;
+    }
+
+    public void setAdminWebappUrl(final String adminWebappUrl) {
+        this.adminWebappUrl = adminWebappUrl;
+    }
+
+    /**
+     * @return External link properties
+     */
+    public ExternalLinksProperties getExternalLinks() {
+        return externalLinks;
+    }
+
+    /**
+     * @return The minimum year with data
+     */
+    public int getMinItemDataYear() {
+        return minItemDataYear;
+    }
+
+    public void setMinItemDataYear(final int minItemDataYear) {
+        this.minItemDataYear = minItemDataYear;
+    }
+
+    /**
+     * @return The available webapp UI languages in addition to English ("en")
+     */
+    public List<String> getUiLanguages() {
+        return uiLanguages;
+    }
+
+    public static class IrisProperties {
+        private String url;
+        private String vendorId;
+
+        /**
+         * @return The iris server URL
+         */
+        public String getUrl() {
+            return url;
+        }
+
+        public void setUrl(final String url) {
+            this.url = url;
+        }
+
+        /**
+         * @return The iris vendor id
+         */
+        public String getVendorId() {
+            return vendorId;
+        }
+
+        public void setVendorId(final String vendorId) {
+            this.vendorId = vendorId;
+        }
+    }
+
+    public static class AnalyticsProperties {
+        private String trackingId;
+
+        /**
+         * @return The Google Analytics tracking id
+         */
+        public String getTrackingId() {
+            return trackingId;
+        }
+
+        public void setTrackingId(final String trackingId) {
+            this.trackingId = trackingId;
+        }
+    }
+
+    public static class ReportServiceProperties {
+        private String url;
+        private List<String> languages = new ArrayList<>();
+
+        /**
+         * @return The report processor url
+         */
+        public String getUrl() {
+            return url;
+        }
+
+        public void setUrl(final String url) {
+            this.url = url;
+        }
+
+        /**
+         * @return The available reporting languages
+         */
+        public List<String> getLanguages() {
+            return languages;
+        }
+    }
+
+    public static class ExternalLinksProperties {
+        private String interpretiveGuide;
+        private String userGuide;
+
+        /**
+         * @return The interpretive guide url
+         */
+        public String getInterpretiveGuide() {
+            return interpretiveGuide;
+        }
+
+        public void setInterpretiveGuide(final String interpretiveGuide) {
+            this.interpretiveGuide = interpretiveGuide;
+        }
+
+        /**
+         * @return The user guide url
+         */
+        public String getUserGuide() {
+            return userGuide;
+        }
+
+        public void setUserGuide(final String userGuide) {
+            this.userGuide = userGuide;
+        }
+    }
+}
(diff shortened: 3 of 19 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
```diff
@@ -1,72 +1,75 @@
...
-    @Value("${app.min-item-data-year}")
-    private int minItemDataYear;
-
-    @Value("${app.admin-webapp-url}")
-    private String adminWebappUrl;
-
-    private AppSettings() {
+    @Autowired
+    private AppSettings(final ApplicationProperties applicationProperties) {
+        this.applicationProperties = applicationProperties;
```
wtritch: I ended up adding the additional "languages" configuration values to the existing AppSettings payload which comes from the api/user UserContext... I'd prefer to move AppSettings out from under UserContext and into it's own API endpoint, but that feels out-of-scope for this task.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
```diff
@@ -19,15 +24,26 @@
...
         databaseMessages.setParentMessageSource(baseMessages);
-        return databaseMessages;
+
+        if (isEmpty(tenantProperties.getTranslationLocation())) {
+            return databaseMessages;
+        }
+
+        final ReloadableResourceBundleMessageSource configMessages = new ReloadableResourceBundleMessageSource();
+        configMessages.setResourceLoader(resourceLoader);
+        configMessages.setBasename(tenantProperties.getTranslationLocation() + "messages");
```
esotrope: should this be smart and conditionally (or not conditionally) prepend a `/` ?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
```diff
@@ -19,15 +24,26 @@
...
+        configMessages.setParentMessageSource(databaseMessages);
+        return configMessages;
```
wtritch: This just optionally adds the tenant.translationLocation as a message source to the java translations used for PDF report generation. (Currently this is *only* used for PDF report generation.)

webapp/src/main/java/org/opentestsystem/rdw/reporting/configuration/ApplicationProperties.java
```diff
@@ -0,0 +1,170 @@
...
+@ConfigurationProperties(prefix = "app")
+public class ApplicationProperties {
+
+    private IrisProperties iris = new IrisProperties();
+    private AnalyticsProperties analytics = new AnalyticsProperties();
+    private ReportServiceProperties reportservice = new ReportServiceProperties();
+    private String adminWebappUrl;
+    private ExternalLinksProperties externalLinks = new ExternalLinksProperties();
+    private int minItemDataYear;
+    private List<String> uiLanguages = new ArrayList<>();
```
wtritch: Most of this is referencing existing configuration values.  There are two new configurable properties:
app.uiLanguages which accepts a collection of UI languages in addition to English
app.reportservice.languages which accepts a collection of Report languages in addition to English

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
```diff
@@ -22,29 +22,28 @@
...
 class DefaultReportWebServiceClient implements ReportWebServiceClient {
 
-    private final String url;
+    private final ApplicationProperties applicationProperties;
     private final RestTemplate template;
 
     @Autowired
     DefaultReportWebServiceClient(
-            @Value("${app.reportservice.url}") final String url,
+            final ApplicationProperties applicationProperties,
```
esotrope: Hm. not sure how i feel about this. Shouldn't the constructor only accept what the component needs and not more?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
```diff
@@ -19,15 +24,26 @@
...
         final ResourceBundleMessageSource baseMessages = new ResourceBundleMessageSource();
         baseMessages.setBasenames("messages");
         baseMessages.setDefaultEncoding("UTF-8");
 
         final AbstractMessageSource databaseMessages = new TranslationServiceMessageSource(translationService);
         databaseMessages.setParentMessageSource(baseMessages);
-        return databaseMessages;
+
+        if (isEmpty(tenantProperties.getTranslationLocation())) {
+            return databaseMessages;
```
esotrope: 👍

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.spec.ts
```diff
@@ -56,48 +28,23 @@ describe('AssessmentResultsComponent', () => {
...
-        { provide: DataService, useValue: dataService },
-        { provide: NotificationService, useValue: service },
-        ItemInfoService,
-        UserService,
-        UserMapper,
-        { provide: CachingDataService, useValue: dataService }
-      ]
+        ColorService
+      ],
+      schemas: [ NO_ERRORS_SCHEMA ]
```
wtritch: This test was loading *a lot* of additional components by loading the entire child hierarchy underneath the assessment results component.  I changed the test initialization to just stand up the assessment results component and ignore any unresolvable child components.  
We should probably make a sweep through all of our front-end component tests to make sure we aren't loading a large hierarchy of child components every time we test a root component.

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -27,14 +25,13 @@ export class AppComponent {
...
-              private location: Location,
-              private angulartics2GoogleAnalytics: Angulartics2GoogleAnalytics) {
+              private location: Location) {
 
-    let languages = [ 'en', 'ja' ];
+    let languages = [ 'en' ];
     let defaultLanguage = languages[ 0 ];
     translate.addLangs(languages);
     translate.setDefaultLang(defaultLanguage);
-    translate.use(languages.indexOf(translate.getBrowserLang()) != -1 ? translate.getBrowserLang() : defaultLanguage);
```
esotrope: good catch 👍

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -41,6 +41,13 @@ <h4 class="modal-title pull-left">{{title}}</h4>
               </select>
             </div>
 
+            <div class="form-group" *ngIf="reportLanguages.length > 1">
+              <label for="language-select">{{'labels.reports.form.language' | translate}}</label>
+              <select id="language-select" class="form-control input-sm" name="language" [(ngModel)]="options.language">
+                <option *ngFor="let language of reportLanguages" [ngValue]="language">{{'labels.locale.' + language | translate}}</option>
+              </select>
+            </div>
```
wtritch: This conditionally adds a language selector to the report generation modal dialog if more than just english is available as a report language.

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -54,23 +56,30 @@ export abstract class ReportDownloadComponent implements OnInit {
   AssessmentType: any = AssessmentType;
   assessmentTypes: AssessmentType[] = [ null, AssessmentType.IAB, AssessmentType.ICA ];
   subjectTypes: AssessmentSubjectType[] = [ null, AssessmentSubjectType.MATH, AssessmentSubjectType.ELA ];
-  languages: string[] = [ 'eng', 'spa', 'vie' ];
   orders: ReportOrder[] = [ ReportOrder.STUDENT_NAME, ReportOrder.STUDENT_SSID ];
   options: ReportOptions;
+  reportLanguages: string[] = ['en'];
```
wtritch: english is always available.

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -54,23 +56,30 @@ export abstract class ReportDownloadComponent implements OnInit {
...
     defaultOptions.accommodationsVisible = false;
     defaultOptions.order = this.orders[ 0 ];
     defaultOptions.grayscale = false;
     this.options = defaultOptions;
+
+    this.userService.getCurrentUser().subscribe(user => {
+      if (!isNullOrUndefined(user)) {
+        this.reportLanguages = this.reportLanguages.concat(user.configuration.reportLanguages);
+      }
+    });
```
wtritch: This adds additional report languages as selectable options.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -44,19 +44,10 @@
...
-        "default": "Spanish"
-      },
-      "vie": {
-        "native": "Tiếng Việt",
-        "default": "Vietnamese"
-      }
+    "locale": {
+      "en": "English",
+      "es": "Spanish",
+      "vi": "Vietnamese"
```
wtritch: I didn't see any current use of the native names, and all of our front-end libraries seem to run on the two-letter-code locales, while the back-end runs on Locale instances which can be initialized from any ISO-standard locale identifier (2- or 3-letter)

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
wtritch: nope, because this is more of a "prefix" than a "path."  For instance, when it comes to the .json files, the admin webapp will likely reference "binary-http://config:8888/*/*/master/admin_" to separate admin_es.json from the webapp's "es.json" (or "webapp_es.json")

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
wtritch: Well, I have two responses to that:
1. I really dislike @Value.  It's clunky, you could have different "defaults" if the value is injected into more than one location, it stinks at type coercion for anything other than simple primitives, and it's locked at launch time.
2. Passing the application configuration to multiple beans makes total sense to me.  It's a central reference to the way things should be, which all consumers should probably agree on.

webapp/src/main/webapp/src/app/app.component.ts
wtritch: Thanks =)  It's a little strange setting a language in the constructor, then conditionally changing it in the onInit method... but I'm not sure we want to wait until onInit to initialize our translation framework... thoughts?

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
```diff
@@ -25,4 +26,17 @@ public boolean isTransferAccessEnabled() {
     public void setTransferAccessEnabled(final boolean transferAccessEnabled) {
         this.transferAccessEnabled = transferAccessEnabled;
     }
+
+    /**
+     * This is the location/prefix of translation files (messages_es.properties, es.json, etc)
+     * (ex: 'binary-${spring.cloud.config.uri}/*&#47;*&#47;master/')
```
malaffoon: is that encoding for `/*/*/`?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
malaffoon: I guess i'm not reading this properly ... translations are going to come from *either* the database or the configured location but not both?

webapp/src/main/java/org/opentestsystem/rdw/reporting/configuration/ApplicationProperties.java
malaffoon: nit: those properties are inconsistent, `app.ui.languages` and `app.reportservice.languages` would be more consistent?

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
```diff
@@ -22,29 +22,28 @@
...
             @Qualifier("pdfAcceptingRestTemplate") final RestTemplate template) {
-        this.url = url;
+        this.applicationProperties = applicationProperties;
```
malaffoon: You could, at least, extract only what you need in the constructor since *everywhere* it does `applicationProperties.getReportservice().getUrl()`.

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: Did i miss the spot where loading from the database deals with this difference?

webapp/src/main/java/org/opentestsystem/rdw/reporting/configuration/ApplicationProperties.java
malaffoon: After seeing the rest of the review i'd lean to `app.uiLanguages` and `app.reportLanguages`.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
wtritch: Not exactly, IFF a translation location is defined, we add an additional MessageSource that uses the database as it's fallback (via .setParent)

webapp/src/main/webapp/src/assets/i18n/en.json
wtritch: The Controllers deal with the difference, translating the string property into a Locale instance when parsing the payload.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
wtritch: yeah... `*/*/` doesn't work well as a javadoc comment =/

webapp/src/main/java/org/opentestsystem/rdw/reporting/report/client/DefaultReportWebServiceClient.java
wtritch: I don't want to extract the value, because that locks it at launch time.  (I still have a small hope that we may allow live-refresh of configurations at some point...)

webapp/src/main/java/org/opentestsystem/rdw/reporting/configuration/ApplicationProperties.java
wtritch: I'll change it to app.ui-languages and app.report-languages.  I'll leave app.reportservice.url alone.

wtritch: Done

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
jtreuting: i feel like this is the third time that has been said, so we should probably create a JIRA to track it.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.spec.ts
jtreuting: It might be worth talking thru in a parking lot discussion about how to see when that is happening.  I know I'm not familiar with it and maybe others aren't either.

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: so each language file will have the languages presented in that language.  does it make more sense to have the english language file say Español as if you are looking to change the UI to Spanish then that would make more sense to you? 

now that I say that, i think the use cases are different.  for choosing your own language for the UI I think that makes sense having it say Español, but for choosing the reporting printout language as a user with english as the UI it saying Spanish there makes sense.

so i think having the native and default would work for those 2 scenarios.

wtritch: hmmm... that's a good point.  The UI language selector can show the language in the native tongue and the report selector can show language in the current language... hrm

esotrope: 👍  that would be the best user experience i think

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.spec.ts
wtritch: Any time you don't see "schemas: [ NO_ERRORS_SCHEMA ]" in a component test, and a huge list of "declarations" it means the component is standing up a full instance of every component it and it's sub-components are referencing.  I'll mention it at the standup.

webapp/src/main/webapp/src/assets/i18n/en.json
wtritch: added native spellings back in in

PR #570, opened 2017-11-06
Support configurable languages for the reporting webapp UI

wtritch:
This PR allows multiple reporting webapp UI languages by accepting .json files from an arbitrary source, such as our configuration server.


common-web/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
     compile 'org.springframework.security.extensions:spring-security-saml2-core'
     compile 'org.springframework.session:spring-session'
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-guava'
+    compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-properties'
```
wtritch: I'm using the jackson library for converting flat properties into hierarchical JSON:
`parent.child.grandchild=blah`
becomes
```
parent: {
  child: {
    grandchild: blah
  }
}
```

esotrope: i was going to ask what this was for

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/FrontendTranslationService.java
```diff
@@ -1,4 +1,4 @@
-package org.opentestsystem.rdw.reporting.search.i18n;
+package org.opentestsystem.rdw.reporting.common.service;
```
wtritch: I moved the FrontendTranslationService stuff into common to supply DB translations to the new TranslationMergeService below.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    //TODO Consider application-level caching scheme - this should almost never change.
+    @Override
+    public ObjectNode getLocaleMessages(final Locale locale) {
+        final ObjectNode messages = new ObjectNode(objectMapper.getNodeFactory());
+
+        final Properties translationProperties = new Properties();
+        translationProperties.putAll(translationService.getTranslationsByCode(locale));
+        appendProperties(messages, translationProperties);
```
wtritch: Read database key/value translations and convert them into JSON.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
```diff
@@ -0,0 +1,130 @@
...
+        appendResource(messages, externalMessages);
+
+        return messages;
+    }
+
+    private String getExternalLocation(final Locale locale) {
+        return tenantProperties.getTranslationLocation() + locale.getLanguage() + ".json";
+    }
+
+    private void appendProperties(final ObjectNode messages, final Properties properties) {
```
esotrope: these private methods probably belong in a service

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
```diff
@@ -0,0 +1,130 @@
...
+        final Properties translationProperties = new Properties();
+        translationProperties.putAll(translationService.getTranslationsByCode(locale));
+        appendProperties(messages, translationProperties);
+
+        if (isEmpty(tenantProperties.getTranslationLocation())) {
+            return messages;
+        }
+
+        final Resource externalMessages = resourceLoader.getResource(getExternalLocation(locale));
+        appendResource(messages, externalMessages);
```
wtritch: If we have an external translation location configured, try to find the external translation json file and merge/overwrite the DB values with the JSON properties.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
```diff
@@ -0,0 +1,130 @@
...
+            final JsonNode resourceNode =  objectMapper.readTree(in);
+            if (resourceNode.isObject()) {
+                deepMerge(messages, (ObjectNode) resourceNode);
+            }
+        } catch (final IOException e) {
+            logger.warn("Unable to open translation resource: {}", resource);
+        }
+    }
+
+    private void deepMerge(final ObjectNode target, final ObjectNode source) {
```
wtritch: Simple ObjectNode deep merge impl.  This comes "for free" in jackson 2.9, but I didn't want to bump our jackson dependency version beyond what Spring expects.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/TranslationController.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Locale;
+
+/**
+ * This controller is responsible for providing translation message values
+ * from both the database and external xx.json resources.
+ */
+@RestController
+public class TranslationController {
```
wtritch: This common TranslationController replaces the reporting webapp TranslationController and provides the merged DB and external translation messages rather than just the DB translation messages.

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.spec.ts
```diff
@@ -1,43 +1,63 @@
...
-  };
-
-  apiResponse = new Response(new ResponseOptions({
-    body: JSON.stringify(apiMockData)
-  }));
-
-  uiResponse = new Response(new ResponseOptions({
-    body: JSON.stringify(uiMockData)
-  }));
+describe('RdwTranslateLoader', () => {
```
wtritch: Restructured this test into a single "describe" suite and converted the test to use a spied Http instance rather than an extended MockHttp instance.

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -7,14 +7,19 @@ import { Injectable } from "@angular/core";
 
 @Injectable()
 export class RdwTranslateLoader {
+
+  private embeddedLanguages: string[] = ["en"];
```
wtritch: We'll only attempt to load embedded "assets/en.json" and rely on the DB/external translations for non-embedded languages.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
wtritch: I'm not sure what you mean?  They are in a service?

esotrope: Oops. Skimming too fast i guess. Thought i was looking in a `@Configuration`

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
```diff
@@ -0,0 +1,130 @@
...
+                                          final ResourceLoader resourceLoader,
+                                          final ObjectMapper objectMapper,
+                                          final FrontendTranslationService translationService) {
+        this.tenantProperties = tenantProperties;
+        this.resourceLoader = resourceLoader;
+        this.objectMapper = objectMapper;
+        this.translationService = translationService;
+    }
+
+    //TODO Consider application-level caching scheme - this should almost never change.
```
wtritch: This was a big question I had.  Have we considered what the RDW-standard is for caching?  Do we want to use Redis?  Do we want to have short-/medium-/long-term caches like TDS?

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
```diff
@@ -7,14 +7,19 @@ import { Injectable } from "@angular/core";
...
+
   constructor(private http: Http) {
   }
 
   private apiLoader = new TranslateHttpLoader(this.http, '/api/translations/', '');
   private uiLoader = new TranslateHttpLoader(this.http, '/assets/i18n/', '.json');
 
   getTranslation(lang: string) {
-    let uiObservable = this.uiLoader.getTranslation(lang);
+    let uiObservable = this.embeddedLanguages.indexOf(lang) >= 0 ?
```
bgsavage: Do you need this check if you're hardcoding this value?

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
bgsavage: I wonder if there's a way to generate this collection based on the contents of a directory rather than hardcoding what we know is there right now.  This could allow us to add additional languages in the deployed directory later without requiring a code change.

bgsavage: Ah, but I guess that's what you're already doing api side in this pr... :)

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
bgsavage: Seems weird to me that we now have a merge api side, and ui side.  This could become confusing to try and figure out what the hierarchy of overwrites is.  I think I would prefer to just see all the merging happen on the ui.  We should document what the hierarchy is now too.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/TranslationController.java
bgsavage: I'm wondering if it might be better to have a 2nd api endpoint to get the external xx.json resources so that it can be merged the same way the api and ui translations are merged today.  It would not be great if the different merge implementations produced inconsistent results.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
jtreuting: This is called once per user (unless they do a hard refresh of the browser that is).  Is that right?

And to answer your question, no we haven't really thought that through as we haven't had a big need yet.  There is a story to think about generic caching solution.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
wtritch: All the merging can't happen on the UI without three separate API calls: embedded, DB, external.  If anything, I'd rather all the merging happen on the back-end with an optional merge of embedded language resources as the final fallback.  But I wonder if that's worth the dev cost...

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/TranslationController.java
wtritch: It's consistent, but I'd prefer to reduce the chatter and front-end processing where possible.  Right now, the front-end only requests en.json + api

webapp/src/main/webapp/src/app/shared/rdw-translate-loader.ts
wtritch: right =)  The idea is that installing another language should be as easy as dropping a xx.properties and/or xx.json file in the config server and adding xx as an available language for reports, ui, or both.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
malaffoon: The legacy RDW system had time-based caches as well. I guess i just don't get it: i don't care if a cache is tagged as "short-term", "long-term" or whatever. It is responsible for having fresh data whenever i call it. So *every* cache should be "just-as-long-as-needed-term". So, if we want to do an application-level cache for this, then it should monitor the message sources for changes and cache as appropriate.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
```diff
@@ -0,0 +1,130 @@
...
+
+        final Properties translationProperties = new Properties();
+        translationProperties.putAll(translationService.getTranslationsByCode(locale));
+        appendProperties(messages, translationProperties);
+
+        if (isEmpty(tenantProperties.getTranslationLocation())) {
+            return messages;
+        }
+
+        final Resource externalMessages = resourceLoader.getResource(getExternalLocation(locale));
```
malaffoon: this will never return null?

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
wtritch: Yeah, it's called once per browser session atm.  (Hard refresh reloads the browser session.)  My problem is that every time a user logs in, we check to see if an external en.json file exists in the external location via an http call... I'd much rather make the call once-per-day or even once-per-hour.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
wtritch: nope, it returns a UrlResource that will return true/false from Resource#exists()

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
bgsavage: Other options could be:
1.  Have the api return two separate api objects that get merged on the ui along with the ui / api merge.
2.  Treat the ui "embedded language" as resource location, possibly as a hardcoded one, and so everything is merged on the api and the ui only makes the one api call (no resource load of en.json on the ui)

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
wtritch: @malaffoon The problem there is that we're caching on a non-CRUD resource that's modified outside of the scope of the cache, so we can't just invalidate on a Create/Update.  
In fact, nothing in our reporting webapp is modified through the app, so timed caches and/or a scheduled task that invalidates are the only way to invalidate/check if we should invalidate.  An expiring cache is an easier implementation, but both options feel roughly equivalent to me.  Thoughts?

jtreuting: Are we able to know when these values would be stale to clear the cache?  If this is a config value that it is pulling what would be the trigger for removing the cache?

wtritch: If we want to use the Spring @Cacheable annotation we could have a public cached service method and a package-local non-cached service method that's hit by a task every X minutes.  If the cached value and the non-cached value are different, invalidate the cache?  It feels pretty hokey though...

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
wtritch: Right, #2 would be moving all the merging to the back-end, but I'm not sure it's worth the dev cost...

wtritch:
I believe I've responded to all the comments, although there were some concerns (especially regarding caching) that I don't believe were answered.  I'll merge this EOD today if there are no objections and we can continue discussion the larger issues elsewhere.

jtreuting:
agreed

PR #571, opened 2017-11-06
Feature/common ngx additions

esotrope:
Moves CachingDataService, DataService, Download, DataTableRowExpander, and Utils into common-ngx

Prunes admin-webapp & webapp of unused .ts files as well.


webapp/src/main/webapp/src/tsconfig.json
```diff
@@ -16,9 +16,8 @@
       "../node_modules/@types"
     ],
     "paths": {
-      "@angular/*": [
-        "../node_modules/@angular/*"
-      ]
+      "@angular/*": [ "../node_modules/@angular/*" ],
+      "rxjs/*": [ "../node_modules/rxjs/*" ]
```
esotrope: A fun intricacy of consuming the common lib is that you need to tell the consuming app to not use the lib's node_modules when they are installed by your app. This is the node/npm way of preventing the equivalent of a `ClassCastException`

common-ngx/src/data/rdw-data.module.ts
```diff
@@ -0,0 +1,45 @@
...
+   */
+  static forRoot(): ModuleWithProviders {
+
+    // Unable to pass options into the forRoot() function.
+    // Providers that are dynamic in value are not supported by the AoT (ngc) compiler
+    // We may want to consider moving to use tsc/nsc compiler instead.
+    // See limitations & issues:
+    // https://github.com/rangle/angular-2-aot-sandbox#func-in-providers-usevalue-top
+    // https://github.com/angular/angular-cli/issues/3706
+    // https://github.com/angular/angular/issues/13892
```
esotrope: Important comments in this file.

common-ngx/src/data/data.service.ts
```diff
@@ -35,13 +38,27 @@ export class DataService {
    * @param options parameters to communicate to the API
    * @returns {Observable<R>}
    */
-  public post(url: string, body: any, options?: RequestOptionsArgs): Observable<any> {
+  post(url: string, body: any, options?: RequestOptionsArgs): Observable<any> {
     return this.http
-      .post(`/api${url}`, body, options)
+      .post(`${this.contextUrl}${url}`, body, options)
```
esotrope: Gives us a way to either remove or change the default `/api` url prefix.

admin-webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -51,7 +51,7 @@ export class BreadcrumbsComponent implements OnInit{
       }
 
       // Parse the route commands for this route
-      let crumbData = child.snapshot.data[BreadcrumbsKeyword];
+      let crumbData = child.snapshot.data[ BreadcrumbsKeyword ];
```
wtritch: Just a humble request to run the formatter after the PR =)

esotrope: Well, if the file was already formatted... ; p

common-ngx/src/data/data.service.ts
wtritch: I wonder how we might be able to do that via configuration... like if we're running the app behind a path-based reverse-proxy?  (ex: http://some.shared.domain.com/reporting/api/blah AND http://some.shared.domain.com/another-reporting/api/blah)

common-ngx/src/data/data.service.ts
```diff
@@ -35,13 +38,27 @@ export class DataService {
...
+   *
+   * @param url the API endpoint
+   * @param options parameters to communicate to the API
+   * @returns {Observable<any>}
+   */
+  delete(url: string, options?: RequestOptionsArgs): Observable<any> {
+    return this.http
+      .delete(`${this.contextUrl}${url}`, options)
+      .map(this.getMapper(options));
+  }
```
wtritch: Is this just to round out the functionality?

esotrope: Admin used the get & delete methods. Reporting just used get & post. Just bundling them up.

webapp/src/main/webapp/src/tsconfig.json
wtritch: neat, so we'll have to specifically call out any shared libs/resources here to avoid having multiple versions at runtime?

esotrope: Correct. Anything that both the linked lib and the app use will have to be called out here. This is specifically a concern when using `npm link`

wtritch: I wonder what we're doing differently such that we don't have to do that with other 3rd-party libs?  Or do we?

esotrope: Other 3rd party libs like @ngx-bootstrap seem to be using a self-rolled wrapper of nsc/tsc which creates JIT & AoT compatible JS. nsc (part of angular-cli) generates only AoT compatible JS. Another think these libs do is publish to npm (making it unnecessary to use `npm link`) though they are compatible with using `npm link` for parallel lib-app development.

PR #572, opened 2017-11-07
Feature/common ngx core module

esotrope:
1. Moves StorageService, WindowRefService to common-ngx/core2. 
2. Moves ScaleScoreService to assessment module since there is no other use in another module
3. Removes now unused standalone mode components


common-ngx/src/core/rdw-core.module.ts
```diff
@@ -0,0 +1,17 @@
...
+@NgModule({
+  imports: [
+    BrowserModule
+  ],
+  providers: [
+    StorageService,
+    WindowRefService
+  ]
+})
+export class RdwCoreModule {
```
bgsavage: Hm, what is "core"?  I'd challenge you to come up with a more descriptive name which describes what it currently contains, and how you envision it growing.  Right now, it seems to just be holding two services that interact directly with the browser, so maybe something to indicate that?

webapp/src/main/webapp/src/app/shared/common.module.ts
```diff
@@ -87,11 +82,8 @@ import { RdwDataModule, RdwDataTableModule, RdwFormatModule, RdwFormModule } fro
     DatePipe,
     ColorService,
     DecimalPipe,
-    ScaleScoreService,
     NotificationService,
-    RdwTranslateLoader,
-    StorageService,
-    WindowRefService
+    RdwTranslateLoader
   ]
```
bgsavage: nit: would be nice to alphabetize these while making these changes.  And I know, I helped contribute to this unorganized list...  Sorry for that... :)

common-ngx/src/core/rdw-core.module.ts
wtritch: I know "common" is slightly overused, but it seems applicable =)

esotrope: I think I might actually split this into two packages in the future. for now i'm tempted just to leave it.

webapp/src/main/webapp/README.md
```diff
@@ -13,11 +13,6 @@ ng test
 #!bash
 ng serve --proxy-config /opt/rdw-reporting/config/proxy.conf.json
 ```
-To run in standalone mode with static data, add the below arguments:
-```
-#!bash
-ng serve --target=development --environment=standalone
```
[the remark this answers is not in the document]
esotrope: That was beautiful. [My tear](http://gph.is/15zlCY3)

PR #573, opened 2017-11-07
updated API to handle specific examItem types and return more info

dsnyder-fairway:
changes are for DWR-1068


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamSearchArgumentResolver.java
```diff
@@ -26,8 +28,10 @@ public Object resolveArgument(
             final WebDataBinderFactory binderFactory) throws Exception {
 
         final Map<String, String> pathVariables = getPathVariables(request);
+        final String itemTypes = StringUtils.defaultString(request.getParameter("types"));
         return GroupExamSearch.builder()
                 .schoolYear(Integer.valueOf(request.getParameter("schoolYear")))
+                .itemTypes(ImmutableList.copyOf(StringUtils.split(itemTypes, ',')))
```
bgsavage: Might be better to treat this as an array instead of a string of the query params.  In other places in reporting, we will do something like `itemType=Item1&itemType=Item2` instead.

webapp/src/main/resources/application.sql.yml
```diff
@@ -541,9 +541,32 @@ sql:
...
+          i.dok_level_subject_id,
+          i.math_practice_code,
+          i.allow_calc,
+          i.common_core_standard_ids,
+          i.answer_key,
+          i.options_count,
+          i.type
         from item i
         where i.asmt_id=:assessment_id
+        and i.type in (:types)
```
bgsavage: Hmm, instead of duping the query, should we consider a conditional in the where clause similar to how we do permissions?  

Example: 
`where i.asmt_id = :assessment_id and (1=:includeAllItemTypes or i.type in (:types)`

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/AssessmentItemRepository.java
```diff
@@ -17,6 +18,8 @@
      */
     List<AssessmentItem> findAllForAssessment(long assessmentId);
 
+    List<AssessmentItem> findAllForAssessmentAndTypes(long assessmentId, Collection<String> types);
```
jtreuting: seems like this is setup as a separate API endpoint that gets the data for distractor analysis.  for the Results by Item screen we use the data that is already client-side to build that view.  the original idea was that we needed additional info (item type, answer key and # of options) to be provided to the existing API so that the client-side has that additional data and then we don't need to query the server again.  since we are doing our other filtering client side i think that is the approach that makes sense.

was there something you saw or ran into where that appraoch won't work?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamSearchArgumentResolver.java
dsnyder-fairway: I had some trouble using that from the client side (didn't realize there was a separate client story for a while). If this is being handled elsewhere in reporting I'll follow that example

webapp/src/main/resources/application.sql.yml
dsnyder-fairway: thanks! I'll go ahead and do that

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamSearchArgumentResolver.java
bgsavage: I'm pretty sure I did this on admin for searching by group type, but can't check right now bc it's down... :S

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/AssessmentItemRepository.java
bgsavage: Actually we need different data for the Results by Item screen than what is already presented in the default Student Results, so we do make an API call when 'Results By Item' is selected.  Distractor Analysis also makes it's own API call as well since it will only need a filtered set of this item data.

bgsavage:
Did you get an opportunity to look at the rubric exemplar api call and remove the answerKey data it was getting from the xml?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/AssessmentItemRepository.java
dsnyder-fairway: I'm not really sure

@bgsavage, when testing this out locally I changed a few of the angular calls. Does distractor-analysis or results by item make a single call and then apply different client side displays of that data?

I may have ran off on this PR a little bit

dsnyder-fairway:
> @bgsavage commented 2 minutes ago
> Did you get an opportunity to look at the rubric exemplar api call and remove the answerKey data it was getting from the xml?

I did look at that briefly. I will have to address that

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/AssessmentItemRepository.java
bgsavage: So each will make it's own call, and then filter the data based on the user's advanced filters selection + selected sessions.  

In chrome, if you open your dev tools, and select the network tab, you can see these calls being made.

image

jtreuting: couldn't we have it use the same examitems API as results by items as long as that API returns the additional data about the items (right answer, num options, item type) and then if the user views Results by Item and then distractor analysis it is 1 call and not 2?  it would mean client-side filtering on the item type for Distractor Analysis.

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItem.java
```diff
@@ -20,6 +20,9 @@
     private int maximumPoints;
     private String mathPracticeCode;
     private Boolean allowCalculator;
+    private String itemType;
+    private int numberOfChoices;
```
malaffoon: nit: can we use `type` instead of `itemType` since this is the type field on the item class?
nit: can we use `optionsCount` instead of `numberOfChoices` to be consistent with ingest?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/AssessmentItemRepository.java
bgsavage: Yeah it's an efficiency trade-off either way and depends on the what the most common use case.  

In your example, when distractor analysis is selected and the user never goes to Items By Points earned, then we are returning too large of a payload since we're chopping data off as soon as we get it.

My thoughts around the current approach is it's a simpler implementation to not share the data and the components are more portable and don't have some cross dependency to share the data.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/JdbcAssessmentItemRepository.java
```diff
@@ -19,6 +22,9 @@
     @Value("${sql.assessment.item.findAllForAssessment}")
     private String findAllForAssessmentQuery;
 
+    @Value("${sql.assessment.item.findAllForAssessmentAndTypes}")
+    private String findAllForAssessmentAndTypes;
```
malaffoon: nit: to be consistent this should have `Query` on the end of it

webapp/src/main/java/org/opentestsystem/rdw/reporting/exam/AssessmentItem.java
dsnyder-fairway: not a problem. I'll include those changes.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/AssessmentItemRepository.java
jtreuting: is the results by item payload much bigger than it would be for distractor analysis?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/JdbcAssessmentItemRepository.java
dsnyder-fairway: good catch. I'll update

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/AssessmentItemRepository.java
bgsavage: Yes, and by how much depends on how many non-multiple choice / non multiple select options there are on the assessment multiplied by the number of students.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/assessment/JdbcAssessmentItemRepository.java
```diff
@@ -37,6 +43,18 @@ public JdbcAssessmentItemRepository(@NotNull final NamedParameterJdbcTemplate te
     }
 
     @Override
+    public List<AssessmentItem> findAllForAssessmentAndTypes(final long assessmentId, final Collection<String> types) {
+        final Map<String, Object> mapSqlParameters = new HashMap<>();
+        mapSqlParameters.put("assessment_id", assessmentId);
+        mapSqlParameters.put("types", types);
+        return template.query(
+                findAllForAssessmentAndTypes,
+                mapSqlParameters,
```
malaffoon: nit: consider using a map builder, something like ...
```
        ImmutableMap.<String, Object>builder.put("assessment_id", assessmentId).put("types", types).build()
```

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/DefaultGroupExamItemService.java
```diff
@@ -28,7 +29,12 @@
     }
 
     public GroupExamItemScores getExamItemScores(@NotNull final User user, @NotNull final GroupExamSearch search) {
-        final List<AssessmentItem> assessmentItems = assessmentItemRepository.findAllForAssessment(search.getAssessmentId());
+        final List<AssessmentItem> assessmentItems;
+        if (CollectionUtils.isEmpty(search.getItemTypes())) {
```
malaffoon: can't you just call `search.getItemTypes().isEmpty()` ?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamSearch.java
```diff
@@ -18,18 +20,29 @@ public GroupGrant getGroupGrant() {
...
 
         private GroupGrant groupGrant;
+        private ImmutableCollection<String> itemTypes;
 
         public Builder groupGrant(final GroupGrant groupGrant) {
             this.groupGrant = groupGrant;
             return this;
         }
 
+        public Builder itemTypes(final ImmutableCollection<String> itemTypes) {
```
malaffoon: nit: requiring the client to provide an immutable collection is annoying ... this should just take a Collection (better would be an Iterable), then construct the ImmutableCollection internally.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/DefaultGroupExamItemService.java
dsnyder-fairway: I wanted to avoid possible NPEs

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/SchoolGradeExamSearch.java
```diff
@@ -37,6 +44,11 @@ public Builder gradeId(final int gradeId) {
             return this;
         }
 
+        public Builder itemTypes(final ImmutableCollection<String> itemTypes) {
```
malaffoon: ditto

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/SchoolGradeExamSearchArgumentResolver.java
```diff
@@ -25,10 +27,12 @@ public Object resolveArgument(
             final WebDataBinderFactory binderFactory) throws Exception {
 
         final Map<String, String> pathVariables = getPathVariables(request);
+        final String itemTypes = StringUtils.defaultString(request.getParameter("types"));
         return SchoolGradeExamSearch.builder()
                 .schoolYear(Integer.valueOf(request.getParameter("schoolYear")))
                 .schoolId(Long.valueOf(pathVariables.get("schoolId")))
                 .gradeId(Integer.valueOf(pathVariables.get("gradeId")))
+                .itemTypes(ImmutableList.copyOf(StringUtils.split(itemTypes, ',')))
```
malaffoon: nit: itemTypes.split(",")?

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/DefaultGroupExamItemService.java
malaffoon: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/SchoolGradeExamSearchArgumentResolver.java
dsnyder-fairway: to avoid possible NPEs. It is an optional requestParam

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamSearchArgumentResolver.java
dsnyder-fairway: done

dsnyder-fairway:
@malaffoon @bgsavage @jtreuting last commit implemented all the requested changes and the UI portion for this as well

dsnyder-fairway:
Forgot to include IT updates for this. Working on this now

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepositoryIT.java
```diff
@@ -165,7 +165,7 @@ public void exportSingleSchoolExportExamRowData() throws IOException {
         assertThat(getField(row, "FirstName")).isEqualTo("student2_firstName");
         assertThat(getField(row, "LastOrSurname")).isEqualTo("student2_lastName");
 
-        assertThat(getField(row, "SubmitDateTime")).isEqualTo("1997-01-01 00:00:00");
+        assertThat(getField(row, "SubmitDateTime")).isEqualTo("1997-01-01 00:00:00.000000");
```
dsnyder-fairway: looks like latest schema version added more time detail

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamSearch.java
```diff
@@ -1,11 +1,16 @@
...
+import com.google.common.collect.ImmutableSet;
 import org.opentestsystem.rdw.reporting.search.exam.AbstractExamSearch;
+import org.opentestsystem.rdw.security.GroupGrant;
+
+import java.util.Collections;
 
 class GroupExamSearch extends AbstractExamSearch {
 
     private GroupGrant groupGrant;
+    private ImmutableCollection<String> itemTypes;
```
esotrope: This is semantically a Set so i would suggest we make the signatures for it a Set

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/group/GroupExamSearchArgumentResolver.java
```diff
@@ -26,8 +29,10 @@ public Object resolveArgument(
             final WebDataBinderFactory binderFactory) throws Exception {
 
         final Map<String, String> pathVariables = getPathVariables(request);
+        final String[] itemTypes = ArrayUtils.nullToEmpty(request.getParameterValues("types"));
         return GroupExamSearch.builder()
                 .schoolYear(Integer.valueOf(request.getParameter("schoolYear")))
+                .itemTypes(ImmutableList.copyOf(itemTypes))
```
esotrope: no need to do this twice if the model always makes a copy in the setter

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/DefaultSchoolGradeExamItemService.java
```diff
@@ -28,7 +29,12 @@
     }
 
     public GroupExamItemScores getExamItemScores(@NotNull final User user, @NotNull final SchoolGradeExamSearch search) {
-        final List<AssessmentItem> assessmentItems = assessmentItemRepository.findAllForAssessment(search.getAssessmentId());
+        final List<AssessmentItem> assessmentItems;
+        if (CollectionUtils.isEmpty(search.getItemTypes())) {
```
esotrope: I think if we always make sure our models don't return null collections (which so far we do) we don't need the collection utils

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/SchoolGradeExamSearch.java
```diff
@@ -37,10 +47,16 @@ public Builder gradeId(final int gradeId) {
...
+        public Builder itemTypes(final Iterable<String> itemTypes) {
+            this.itemTypes = itemTypes;
+            return this;
+        }
+
         public SchoolGradeExamSearch build() {
             final SchoolGradeExamSearch search = super.build(new SchoolGradeExamSearch());
             search.schoolId = schoolId;
             search.gradeId = gradeId;
+            search.itemTypes = ImmutableSet.copyOf(itemTypes == null ? Collections.EMPTY_LIST : itemTypes);
```
esotrope: To follow the rest of the app i would use:
```java
search.itemTypes = itemTypes != null ? ImmutableSet.copyOf(itemTypes) : ImmutableSet.of()
```

webapp/src/main/resources/application.sql.yml
```diff
@@ -541,9 +541,13 @@ sql:
...
           i.math_practice_code,
           i.allow_calc,
-          i.common_core_standard_ids
+          i.common_core_standard_ids,
+          i.answer_key,
+          i.options_count,
+          i.type
         from item i
         where i.asmt_id=:assessment_id
+        and (1=:includeAllItemTypes or i.type in (:types))
```
esotrope: Hm. wondering if we shouldn't just make the types nullable and do:
`and (:types is null or i.type in (:types))`
@wtritch isn't that the pattern we use in the reports filter?

webapp/src/main/webapp/src/app/groups/results/group-assessment.service.ts
```diff
@@ -54,13 +53,27 @@ export class GroupAssessmentService implements AssessmentProvider {
...
+        params: {
+          types: [ 'MC', 'MS' ],
+          schoolYear: this.schoolYear.toString()
+        }
+      })
+        .catch(ResponseUtils.badResponseToNull)
+        .map(x => {
+          return this.mapper.mapAssessmentItemsFromApi(x);
+        });
+    } else {
```
esotrope: since the first if performs a return the else block isn't needed here

dsnyder-fairway: I wasn't able to successfully pass an empty types, or have the types removed all together without this else block. When we do not have types in there, we are performing a "Results by Item". If we do have it there, it is for "Distractor Analysis"

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/schoolgrade/DefaultSchoolGradeExamItemService.java
dsnyder-fairway: yup, I can remove this

dsnyder-fairway:
Any other comments? I think this is good to merge. Just wanted to make sure that would be fine

PR #574, opened 2017-11-08
Fix the admin webapp by not attempting to load reporting database tra…

wtritch:
…nslations.

This PR converts our merging translation service to accept an ordered set of translation providers so that we can retrieve external translations in the admin webapp without requiring access to the reporting database.


common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/MessageSourceConfiguration.java
```diff
@@ -21,7 +21,7 @@
         JdbcTranslationRepository.class,
         DefaultInternalTranslationService.class
 })
-public class TranslationConfiguration {
+public class MessageSourceConfiguration {
```
wtritch: I would really like if our MessageSource used the same TranslationProviders to provide translations, but it is difficult to justify re-writing the existing ResourceBundle implementation with in-memory caching, natural fallback of message provider, converting our existing messages.properties to json, etc at the moment.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
```diff
@@ -25,9 +25,9 @@
         DataSourceConfiguration.class,
         SecurityConfiguration.class,
         JdbcOrganizationRepository.class,
-        WebConfiguration.class,
+        WebExecptionHandlerConfiguration.class,
```
wtritch: I renamed the common WebConfiguration to WebExceptionHandlerConfiguration to specify exactly what's being included.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/FrontendTranslationService.java
```diff
@@ -1,9 +0,0 @@
-package org.opentestsystem.rdw.reporting.common.service;
-
-import org.opentestsystem.rdw.reporting.common.i18n.TranslationService;
-
-/**
- * Translation service for servicing the frontend
- */
-public interface FrontendTranslationService extends TranslationService {
```
wtritch: This has been replaced with a DefaultFrontendRepositoryTranslationProvider.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultExternalTranslationProvider.java
```diff
@@ -0,0 +1,61 @@
...
+            return objectMapper.createObjectNode();
+        }
+
+        try (final InputStream in = externalMessages.getInputStream()) {
+            return (ObjectNode) objectMapper.readTree(in);
+        } catch (final IOException e) {
+            logger.warn("Unable to open translation resource: {}", externalMessages);
+            return objectMapper.createObjectNode();
+        }
+    }
```
wtritch: This provider loads external Json for a given locale and returns the messages as a Json ObjectNode.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
```diff
@@ -3,99 +3,50 @@
...
 
-    public DefaultTranslationMergeService(final TenantProperties tenantProperties,
-                                          final ResourceLoader resourceLoader,
-                                          final ObjectMapper objectMapper,
-                                          final FrontendTranslationService translationService) {
-        this.tenantProperties = tenantProperties;
-        this.resourceLoader = resourceLoader;
+    @Autowired
+    public DefaultTranslationMergeService(final Set<TranslationProvider> translationProviders,
+                                          final ObjectMapper objectMapper) {
```
wtritch: This class now just accepts the registered TranslationProvider instances, orders them, then merges their messages together.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultFrontendRepositoryTranslationProvider.java
```diff
@@ -0,0 +1,26 @@
...
+public class DefaultFrontendRepositoryTranslationProvider extends AbstractRepositoryTranslationProvider {
+
+    @Autowired
+    public DefaultFrontendRepositoryTranslationProvider(final TranslationRepository translationRepository) {
+        super(FRONTEND, translationRepository);
+    }
+
+    @Override
+    public int getOrder() {
+        return 20;
```
esotrope: don't think it matters but if this were a 3rd party lib i would expect this to be configurable with a default.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultTranslationMergeService.java
esotrope: so messages from provider with order=20 comes before 30 right? might be good to clarify that lower numbers are higher priority because usually it is the reverse.

wtritch: It's in the javadoc for the getOrder method =)  Higher numbers overwrite lower numbers.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultFrontendRepositoryTranslationProvider.java
wtritch: hmm... we could look into that with registration via config... but that seems a little overkill.

PR #575, opened 2017-11-08
Feature/common ngx security module

esotrope:
Created common-ngx security module
Now both admin & reporting webapps have a session-expired route/view


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/ViewController.java
```diff
@@ -1,14 +1,20 @@
 package org.opentestsystem.rdw.admin.web;
 
+import org.springframework.boot.autoconfigure.web.ErrorController;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.RequestMapping;
 
 @Controller
-public class ViewController {
+public class ViewController implements ErrorController {
```
esotrope: Necessary for spring to ignore the default "/error" route mapping to the whitelabel error page

admin-webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1,4 +1,11 @@
 {
+  "common-ngx": {
+    "session-expired": {
+      "title": "Authentication Expired",
+      "message": "Click Ok to re-authenticate.",
+      "ok": "Ok"
+    }
+  },
```
esotrope: admin and reporting en.json now have a common-ngx context where messages for common lib components will go.

admin-webapp/src/main/webapp/src/tsconfig.json
```diff
@@ -17,6 +17,7 @@
     ],
     "paths": {
       "@angular/*": [ "../node_modules/@angular/*" ],
+      "@ngx-translate/*": [ "../node_modules/@ngx-translate/*" ],
```
esotrope: Important to note for people developing common-ngx. whenever both the consuming app and common-ngx lib share a dependency the consuming apps need to add this to the tsconfig to avoid "class path exceptions"

common-ngx/src/data/caching-data.service.ts
```diff
@@ -1,8 +1,10 @@
 import { Injectable } from "@angular/core";
 import { RequestOptionsArgs } from "@angular/http";
-import { Observable } from "rxjs/Rx";
```
esotrope: AoT compilation didn't like this.
I guess the proper way to import Observable is either:
```import { Observable } from "rxjs/Observable";```
or
```import { Observable } from "rxjs";```

File: common-ngx/src/form/sb-button-typeahead.component.ts [hunk 5]
```diff
@@ -1,6 +1,6 @@
 import { Component, EventEmitter, Input, Output } from "@angular/core";
 import { SBTypeahead } from "./sb-typeahead.component";
-import { isUndefined } from "util";
```
esotrope: this was accidently pointing to some compile time dev dependency reference to angular CLI. So I added `isUndefined` impl to the `Utils` class

common-ngx/src/security/rdw-security.module.ts
```diff
@@ -0,0 +1,33 @@
...
+@NgModule({
+  declarations: [
+    SessionExpiredComponent
+  ],
+  imports: [
+    BrowserModule,
+    RdwCoreModule,
+    // These can be overriden in the consuming apps
+    RouterModule.forRoot([]),
+    TranslateModule.forRoot()
```
esotrope: Like it says, these are placeholder modules meant to be overriden with app-specific route and translation configs.

common-ngx/src/security/session-expired.component.ts
```diff
@@ -1,26 +1,20 @@
...
    * If a previous route is not available, navigate the user to the home page.
    */
   onOk(): void {
-    let location = this.authenticationService.getReauthenticationLocation();
-    if (!location) {
-      location = this.location.prepareExternalUrl("/home");
-    }
-
-    window.location.href = location;
+    this.authenticationService.authenticate();
```
esotrope: Delegated this logic to the auth service which already knows about the /home route

webapp/src/main/webapp/src/app/app.module.ts
```diff
@@ -48,7 +47,6 @@ import { OrganizationExportModule } from "./organization-export/organization-exp
   ],
   providers: [
     TranslateResolve,
-    { provide: Http, useClass: AuthenticatedHttpService },
```
esotrope: This is now added as a provider by the RdwSecurityModule (which is imported by the app's CommonModule)

webapp/src/main/webapp/src/app/shared/common.module.ts
```diff
@@ -72,19 +72,19 @@ import {
...
     SBCheckboxList,
     SBRadioButtonComponent,
     SBToggleComponent,
     TranslateModule
   ],
   providers: [
-    AuthenticatedHttpService,
-    AuthenticationService,
+    { provide: AuthenticationServiceAuthenticationExpiredRoute, useValue: 'session-expired' },
+    { provide: AuthenticationServiceDefaultAuthenticationRoute, useValue: 'home' },
```
esotrope: This is how we make the security module aware of the apps session expired and home routes.
Notice the admin app does not configure the 'home' route

common-ngx/package.json
```diff
@@ -48,7 +48,9 @@
     "@angular/common": "^4.0.0",
     "@angular/core": "^4.0.0",
     "@angular/forms": "^4.0.0",
-    "@angular/http": "^4.0.0"
+    "@angular/http": "^4.0.0",
+    "@angular/router": "^4.0.0",
+    "@ngx-translate/core": "^6.0.1"
```
wtritch: I think rxjs needs to be in here too?

common-ngx/src/data/caching-data.service.ts
wtritch: After running into the same issue the One True Way to import Observable is:
`import { Observable } from "rxjs/Observable";`

common-ngx/src/security/session-expired.component.spec.ts
```diff
@@ -16,13 +17,11 @@ describe('SessionExpiredComponent', () => {
...
       ],
       declarations: [ SessionExpiredComponent ],
-      providers: [{
-        provide: AuthenticationService,
-        useClass: MockAuthenticationService
-      }, {
-        provide: Location,
-        useClass: MockLocation
-      }]
+      schemas: [ NO_ERRORS_SCHEMA ],
```
wtritch: +1

common-ngx/package.json
esotrope: looks like not i guess...

PR #576, opened 2017-11-08
validate year when loading student groups

malaffoon:
As part of this, i changed the validation to continue past any row with an error. Previously it would break out as soon as it hit a row with an error. And, as part of that, i made it so it wouldn't log duplicate messages (i.e. if the same unknown school is seen in 20 rows, there will be only one error message regarding that school being unknown).

### DIFF
--- a/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/SchoolYearRepository.java
+++ b/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/SchoolYearRepository.java
@@ -1,0 +1,11 @@
+package org.opentestsystem.rdw.admin.repository;
+
+import java.util.List;
+
+public interface SchoolYearRepository {
+
+    /**
+     * @return list of valid school years
+     */
+    List<Integer> findAll();
+}
--- a/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcSchoolYearRepository.java
+++ b/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcSchoolYearRepository.java
@@ -1,0 +1,29 @@
+package org.opentestsystem.rdw.admin.repository.impl;
+
+import org.opentestsystem.rdw.admin.repository.SchoolYearRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import java.util.HashMap;
+import java.util.List;
+
+@Repository
+class JdbcSchoolYearRepository implements SchoolYearRepository {
+
+    private final NamedParameterJdbcTemplate template;
+
+    @Value("${sql.schoolYear.findAll}")
+    private String findAllQuery;
+
+    @Autowired
+    JdbcSchoolYearRepository(final NamedParameterJdbcTemplate template) {
+        this.template = template;
+    }
+
+    @Override
+    public List<Integer> findAll() {
+        return template.queryForList(findAllQuery, new HashMap<>(), Integer.class);
+    }
+}
--- a/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
+++ b/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
@@ -4,6 +4,7 @@ import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import org.apache.commons.csv.CSVRecord;
 import org.opentestsystem.rdw.admin.model.Organization;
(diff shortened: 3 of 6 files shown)


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/SchoolYearRepository.java
```diff
@@ -0,0 +1,11 @@
+package org.opentestsystem.rdw.admin.repository;
```
malaffoon: There is already a SchoolYearRepository in the reporting webapp project but i don't want to share it because they are pointing at different databases.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcSchoolYearRepository.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import java.util.HashMap;
+import java.util.List;
+
+@Repository
+class JdbcSchoolYearRepository implements SchoolYearRepository {
+
+    private final NamedParameterJdbcTemplate template;
```
malaffoon: Don't really need the parameter template but the database configuration in admin doesn't have a plain ol' JdbcTemplate bean available.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -98,8 +103,15 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
             final CsvSubject subject = CsvSubject.valueOfCaseInsensitive(record.get(HeaderSubject));
             recordNumber = record.getRecordNumber();
 
-            validateKeyValues(failures, school, group, schoolYear, recordNumber);
-            if (!failures.isEmpty()) continue;
```
malaffoon: This line was the culprit for skipping all remaining rows once an error was discovered.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -159,27 +176,6 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
         return ImmutableList.copyOf(failures);
     }
 
-    private void validateStudentCount(final List<CsvValidationResult> failures,
```
malaffoon: didn't delete ... moved

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -214,41 +210,82 @@ private Integer getAsInt(final CSVRecord row, final String header) {
...
+    private static void validateSchoolGroupYear(final List<CsvValidationResult> failures,
+                                                final Set<String> completedSchoolGroupYears,
+                                                final String schoolGroupYear,
+                                                final long recordNumber) {
+        if (completedSchoolGroupYears.contains(schoolGroupYear)) {
+            addUniqueFailure(failures, recordNumber, "Record is out-of-order: sort CSV by school, group, year");
+        }
+    }
+
+    private static void addUniqueFailure(final List<CsvValidationResult> failures, final long recordNumber, final String message) {
```
esotrope: should we just be using a `Set` or does order matter?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -214,41 +210,82 @@ private Integer getAsInt(final CSVRecord row, final String header) {
...
+        boolean failure = false;
+
         if (availableSchools.stream().noneMatch(x -> Objects.equals(x.getNaturalId(), school))) {
-            failures.add(failure(recordNumber, "School [" + school + "] can not be found"));
+            addUniqueFailure(failures, recordNumber, "School [" + school + "] can not be found");
+            failure = true;
         }
         if (completedSchools.contains(school)) {
-            failures.add(failure(recordNumber, "School [" + school + "] is out-of-order: sort CSV by school, group, year"));
+            addUniqueFailure(failures, recordNumber, "School [" + school + "] is out-of-order: sort CSV by school, group, year");
```
jtreuting: i'm assuming we aren't doing the sorting ourselves because of the case for very large CSVs, which is why the onus is on the user.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
malaffoon: I like having it ordered: that way the order of the failures will roughly correspond to the order of the rows in the file.

PR #577, opened 2017-11-08
Support multiple languages in the admin webapp

wtritch:
There was some missing functionality required to support multiple configurable languages in the admin webapp.  Now that our common-ngx module is able to run karma/jasmine tests I moved the front-end translation provider into the common module so the admin webapp can also request translations via the API.


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/configuration/ApplicationProperties.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class holds configurable application properties.
+ */
+@ConfigurationProperties(prefix = "app")
+public class ApplicationProperties {
```
wtritch: I needed an ApplicationProperties impl to hold the configured available UI languages.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/AppSettings.java
```diff
@@ -1,32 +1,40 @@
...
-        return homeUrl;
+        return applicationProperties.getHomeUrl();
+    }
+
+    /**
+     * @return The supported UI languages
+     */
+    public List<String> getUiLanguages() {
+        return applicationProperties.getUiLanguages();
     }
```
wtritch: Now sending the available languages to the UI.

admin-webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -44,6 +44,12 @@ export class AppComponent {
         this.router.navigate([ 'access-denied' ]);
       }
 
+      let languages = user.configuration.uiLanguages;
+      this.translate.addLangs(languages);
+      if (languages.indexOf(this.translate.getBrowserLang()) != -1) {
+        this.translate.use(this.translate.getBrowserLang());
+      }
```
wtritch: After getting the list of available languages, if the user is connecting from a browser advertising a preference for an available language, switch to it.

common-ngx/src/i18n/rdw-translate-loader.ts
```diff
@@ -1,12 +1,14 @@
-import { Observable, Observer } from "rxjs";
 import { JsonUnFlat } from "json-unflat";
 import { TranslateHttpLoader } from "@ngx-translate/http-loader";
 import { Http } from "@angular/http";
 import * as _ from "lodash";
 import { Injectable } from "@angular/core";
+import { TranslateLoader } from "@ngx-translate/core";
+import { Observable } from "rxjs/Observable";
+import { Observer } from "rxjs/Observer";
```
wtritch: I needed to import Observable from "rxjs/Observable" rather than from "rxjs" to avoid class cast exceptions.

wtritch:
Any objection to my merging this?

jtreuting:
Sorry, i don't have any but i only had time to skim it briefly.  You can always merge and respond to comments later if they come in.

PR #578, opened 2017-11-08
Back-end support for filtering exams based whether they are "transfer…

wtritch:
… exams" from external schools.

This PR is the back-end component for:
1. optionally filtering out transfer exams in the UI using an Advanced Filter
2. optionally filtering out transfer exams from a PDF report request

For the first requirement I've added a "transfer" boolean property to Exam that will be set to True if the exam's school is not within the permission scope of the user.  The front-end can then implement a simple filter based upon that property value.

For the second requirement I've added a "disableTransferAccess" option to the report request payload and updated the report SQL to filter exams on both the "disableTransferAccess" request property as well as the "transferAccessEnabled" system configuration property.


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
```diff
@@ -70,6 +105,21 @@ private Exams(){
...
+     * @param permissionScope   A permission scope
+     * @return  True if the permissions allow access to the given school
+     */
+    public static boolean hasSchoolAccess(final School school, final PermissionScope permissionScope) {
+        return permissionScope.isStatewide() ||
+                permissionScope.getDistrictIds().contains(school.getDistrictId()) ||
+                permissionScope.getInstitutionIds().contains(school.getId()) ||
+                (school.getDistrictGroupId() != null && permissionScope.getDistrictGroupIds().contains(school.getDistrictGroupId())) ||
+                (school.getSchoolGroupId() != null && permissionScope.getInstitutionGroupIds().contains(school.getSchoolGroupId()));
+    }
```
wtritch: This determines if the user has normal permission access to the exam's school.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Exam.java
```diff
@@ -118,10 +119,17 @@ public String getAdministrativeConditionCode() {
     /**
      * @return The school this exam was taken at
      */
-    public Organization getSchool() {
+    public School getSchool() {
         return school;
     }
```
wtritch: Since I now need district_group, school_group, etc information I've had to change this to a full School instance.

common/src/main/java/org/opentestsystem/rdw/reporting/common/report/AbstractExamReportRequest.java
```diff
@@ -13,6 +13,7 @@
     protected T options;
     protected String name;
     protected boolean accommodationsVisible;
+    protected boolean disableTransferAccess;
```
wtritch: I made this a "boolean" rather than a "Boolean" because I want it to default to false without having to do null-checking elsewhere

report-processor/src/main/resources/application.sql.yml
```diff
@@ -54,7 +57,11 @@ sql:
           (
             s.id=:school_id
             or
-            (1 = :allow_transfer_access and isch.id = :school_id)
+            (
+              (1 = :allow_transfer_access and 0 = :disable_transfer_access)
```
wtritch: There's a *whole* lot of this change in this SQL, copy-pasted the whole way through.

webapp/src/main/java/org/opentestsystem/rdw/reporting/search/exam/student/JdbcStudentExamRepository.java
```diff
@@ -55,19 +65,21 @@
...
+                    Exam exam = Exams.map(row, Exam.builder(), groupPii)
                             .build();
 
+                    if (exam.isTransfer() && Exams.hasSchoolAccess(exam.getSchool(), individualPii)) {
+                        //Individual PII allows access, this is not a transfer exam
+                        exam = Exam.builder()
+                                .copy(exam)
+                                .transfer(false)
+                                .build();
+                    }
```
wtritch: This is a little fugly.  I create the Exam from the ResultSet first using groupPii permissions to determine transfer status, then need to double-check whether individualPii permissions allow access to the Exam's schools and flip the "transfer" boolean back to false. =/  Thoughts?

wtritch:
I'm merging this, will continue to resolve any comments as they come in.

PR #579, opened 2017-11-09
Feature/remove subject pipe

esotrope:
Originally planned on moving subject pipe to common but found that they had two different impls.
This PR removes them in favor of just using translation.
This PR also makes a step toward data driven group subject codes.


admin-webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -66,7 +66,7 @@
         </p-column>
         <p-column field="subject" header="{{'labels.groups.cols.subjects' | translate}}" [sortable]="true">
           <ng-template let-group="rowData" pTemplate="body">
-            {{ group.subject | subject }}
+            {{ 'labels.groups.subjects.' + (group.subject || 'ALL') | translate }}
```
esotrope: turned undefined into 'ALL', thought this might read better in the en.json

webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -30,7 +30,7 @@ <h3 class="blue mb-md">{{'labels.reports.heading' | translate}}</h3>
     </p-column>
     <p-column field="subjectId" header="{{'labels.reports.column.subject.header' | translate}}" sortable="true">
       <ng-template let-report="rowData" pTemplate="body">
-        {{ report.subjectId | subject }}
+        {{ 'labels.reports.subjects.' + report.subjectId | translate }}
```
esotrope: reports are still using IDs instead of codes. Will try to address this in later PR to make all subject refs data driven.

webapp/src/main/webapp/src/app/user/model/group.model.ts
```diff
@@ -2,5 +2,5 @@ export class Group {
   id: number;
   name: string;
   schoolName: string;
-  subjectId: string[];
```
esotrope: this was just plain wrong x J

webapp/src/main/webapp/src/app/shared/subject.pipe.ts
```diff
@@ -1,15 +0,0 @@
...
-
-  transform(value: number): string {
-    let result = "";
-
-    if(value >= 0 && value < this.subjects.length)
-      return this.subjects[value];
-
-    return result;
-  }
-}
```
esotrope: This hardcoded text in the reporting webapp was a fun find.
BTW this is in production : J

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupQueryArgumentResolver.java
```diff
@@ -31,17 +32,16 @@ public Object resolveArgument(final MethodParameter parameter,
         return GroupQuery.builder()
                 .schoolId(parseInt(webRequest.getParameter("schoolId")))
                 .schoolYear(parseInt(webRequest.getParameter("schoolYear")))
-                .subjects(Stream
-                        .of(webRequest.getParameterValues("subject"))
-                        .map(x-> {
+                // TODO: Subjects should be data driven codes
+                .subjects(Stream.of(webRequest.getParameterValues("subject"))
```
esotrope: @chdietz @bgsavage @wtritch @jtreuting The UI sends one or two subjects. Other search filters send none to represent ALL instead of all subjects.
Should we make this the same as everywhere else or do we intend on having like a multiselect in the UI where they can select 2 of 3 for example?

bgsavage:
I think this PR addresses DWR-739 , can you confirm?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupQueryArgumentResolver.java
chdietz: I think this depends on what the UI needs and admin may be a bit different than reporting.

esotrope:
@bgsavage this only partially addresses DWR-739

PR #580, opened 2017-11-09
WIP/thoughts for packaging common-ngx as a .tgz dependency.

wtritch:
@esotrope what do you think?


admin-webapp/src/main/webapp/package.json
```diff
@@ -45,7 +45,7 @@
   "devDependencies": {
     "@angular/cli": "^1.2.7",
     "@angular/compiler-cli": "4.1.1",
-    "@sbac/rdw-reporting-common-ngx": "file:../../../../common-ngx",
+    "@sbac/rdw-reporting-common-ngx": "file:../../../../common-ngx/sbac-rdw-reporting-common-ngx-0.0.0.tgz",
```
malaffoon: Apart from the obvious file-size difference, what's the advantage to using the tgz over letting it expand it?

wtritch: The big difference is that it copies the resources into the node_modules directory rather than sym-linking the common-ngx directory (complete with it's own node_modules directory.)  This gives us the power to remove the node_modules directory during a `gradle clean`  It also opens up the possibility of versioning every time we change something in the common-ngx lib which means other developers wouldn't need to manually wipe their node_directories when getting latest.

malaffoon: Right now i am seeing it copy stuff into src/main/webapp/node_modules/@sbac/rdw-reporting-common-ngx ... it isn't symlinking anything?

common-ngx/package.json
```diff
@@ -2,9 +2,12 @@
   "name": "@sbac/rdw-reporting-common-ngx",
   "version": "0.0.0",
   "description": "",
-  "main": "./dist/index.umd.js",
-  "module": "./dist/index.js",
-  "typings": "./dist/index.d.ts",
+  "directories": {},
+  "files": [
+    "lib"
```
esotrope: wondering the reason of using this dir as opposed to dist

common-ngx/package.json
```diff
@@ -20,29 +23,13 @@
...
-    "preversion": "npm run ci",
-    "test": "karma start",
-    "test:watch": "karma start --auto-watch --no-single-run",
-    "tslint": "tslint"
+    "test": "karma start karma.conf.ts",
+    "tslint": "tslint",
+    "tsc": "tsc",
+    "tsc:w": "tsc -w",
+    "build": "npm run clear-build && ngc -p ./tsconfig-ngc.json && copyfiles -u 1 ./src/**/*.html lib && npm pack",
+    "clear-build": "rimraf lib"
```
esotrope: do we want to do this when the build is done? looks like this is a temp build dir no?

wtritch: yeah, after the .tgz is packaged the lib directory isn't used any longer.

admin-webapp/src/main/webapp/package.json
wtritch: hmm... mine appears to be copying now as well.  However, it still includes more resources than the lib needs (karma config files, a full node-7.10.0 installation, etc)

common-ngx/package.json
wtritch: No reason in particular other than possibly wanting to use dist for tsc compilation (for testing) and lib for ngc compilation (for distribution)

PR #581, opened 2017-11-09
Feature/re introduce ngc compilation for common ngx

esotrope:
1. Re-added ngc compilation for common-ngx
2. Fixed admin/webapp tests by inlining common lib html templates in the component `ts` files
2. Deleted common-ngx/demo dir as we were not using it.
3. Deleted common-ngx/tsconfig-aot.json, webpack.umd.js and other build comps that aren't needed anymore


admin-webapp/src/main/webapp/package.json
```diff
@@ -45,7 +45,7 @@
   "devDependencies": {
     "@angular/cli": "^1.2.7",
     "@angular/compiler-cli": "4.1.1",
-    "@sbac/rdw-reporting-common-ngx": "file:../../../../common-ngx",
+    "@sbac/rdw-reporting-common-ngx": "file:../../../../common-ngx/dist/sbac-rdw-reporting-common-ngx-0.0.0.tgz",
```
bgsavage: so I'm guessing npm will then explode this intot he @sbac/rdw-reporting-common-ngx directory?

common-ngx/demo/package.json
```diff
@@ -1,46 +0,0 @@
-{
-  "name": "@sbac/rdw-reporting-common-ngx-demo",
```
bgsavage: Guessing this was just a prototype?

common-ngx/package.json
```diff
@@ -20,29 +20,9 @@
...
-    "prebuild": "rimraf dist tmp",
-    "prebuild:watch": "rimraf dist tmp",
-    "prepublishOnly": "npm run ci",
-    "preversion": "npm run ci",
-    "test": "karma start",
-    "test:watch": "karma start --auto-watch --no-single-run",
-    "tslint": "tslint"
+    "build": "rimraf dist && ngc -p ./tsconfig-ngc.json && copyfiles package.json dist && cd dist && npm pack",
+    "test": "karma start karma.conf.ts",
+    "lint": "tslint 'src/**/*.ts'"
```
bgsavage: 👍

common-ngx/src/security/session-expired.component.ts
```diff
@@ -3,7 +3,17 @@ import { AuthenticationService } from "./authentication.service";
...
+  template: `
+    <div>
+      <h3 class="blue mb-md">{{'common-ngx.session-expired.title' | translate}}</h3>
+      <div class="well">
+        <p>{{'common-ngx.session-expired.message' | translate}}</p>
+        <button class="btn btn-primary" (click)="onOk()">
+          <span>{{'common-ngx.session-expired.ok' | translate}}</span>
+        </button>
+      </div>
+    </div>
```
bgsavage: We should probably have a rule of when to use template vs. templateUrl.

For this, I'd prefer to keep it as a templateUrl as it's more than a few lines and provides a nice separation of concern.

common-ngx/tsconfig-ngc.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "es2015",
```
bgsavage: not es6?

common-ngx/tsconfig-ngc.json
```diff
@@ -0,0 +1,26 @@
...
+    "stripInternal": true
+  },
+  "exclude": [
+    "node_modules"
+  ],
+  "files": [
+    "./src/index.ts"
+  ],
+  "angularCompilerOptions": {
+    "skipTemplateCodegen": true
```
bgsavage: what's the impact of this?

common-ngx/demo/package.json
esotrope: yup

admin-webapp/src/main/webapp/package.json
esotrope: correctamundo

common-ngx/src/security/session-expired.component.ts
esotrope: ngc requires templates & stylesheets be inlined. originally the gulp.js was taking care of this but for simplicity i took it out and figured we could do that ourselves. so for common lib let's stick to inline templates.
thoughts? concerns?

common-ngx/tsconfig-ngc.json
esotrope: es6 === es2015
`target: es5` means we compile our es6/ts down to es5 (what all browsers have implemented atm)
`libs: [es6, dom]` the libs we need to compile our es6

common-ngx/src/security/session-expired.component.ts
bgsavage: Hm well that's an interesting constraint.  Isn't that how React works?  I think that would not be ideal if we start having larger templates here.  Also creates an inconsistency in here vs. the dependent apps that don't have that constraint.

common-ngx/tsconfig-ngc.json
esotrope: > By default ngc generates ngfactories for the components and modules. By using skipTemplateCodegen flag we can skip this and only get *.metadata.json files.
http://blog.mgechev.com/2017/01/21/distributing-an-angular-library-aot-ngc-types/

PR #582, opened 2017-11-09
Enable simple translation caching

wtritch:
In lieu of a more robust solution, enable simple translation caching for interim work and awsqa testing.

### DIFF
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -50,7 +50,7 @@ spring:
 
   cache:
     type: SIMPLE
-    cache-names: permissionsByRole, internalTranslationsByCode
+    cache-names: permissionsByRole, internalTranslationsByCode, mergeTranslations
 
   thymeleaf:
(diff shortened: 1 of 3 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -111,12 +111,6 @@ app:
   analytics:
     trackingId: ${rdw.reporting.analytics.trackingId:}
 
-  wkhtmltopdf:
-    url: http://localhost:8082
-    options:
-      page-size: Letter
-      zoom: 0.75
```
wtritch: Unrelated, reporting webapp is no longer directly communicating with wkhtmltopdf

esotrope: 👍

malaffoon: Can you please add a note to the CHANGELOG for this ... we need to keep track of stuff that will affect configuration when we deploy. Thanks.

PR #585, opened 2017-11-10
Fix static imports in library ts class

wtritch:
Modify static imports in common-ngx.

### DIFF
--- a/common-ngx/src/security/authenticated-http.service.ts
+++ b/common-ngx/src/security/authenticated-http.service.ts
@@ -1,6 +1,8 @@
 import { Http, Request, RequestOptions, RequestOptionsArgs, Response, XHRBackend } from "@angular/http";
 import { Injectable } from "@angular/core";
 import { Observable } from "rxjs/Observable";
+import { of } from "rxjs/observable/of";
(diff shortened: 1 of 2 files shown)


common-ngx/src/security/authenticated-http.service.ts
```diff
@@ -1,6 +1,8 @@
 import { Http, Request, RequestOptions, RequestOptionsArgs, Response, XHRBackend } from "@angular/http";
 import { Injectable } from "@angular/core";
 import { Observable } from "rxjs/Observable";
```
wtritch: The Observable class from "rxjs/Observable" does not contain the static methods "of()" or "throw()"
I'm including them down below.
The Observable class from "rxjs" *does* include those static methods.  Unfortunately, since we're overriding Angular's Http class' `request(url: string | Request, options?: RequestOptionsArgs): Observable<Response>` method, we *must* have the same method signature, and they expect "Observable" to be imported from "rxjs/Observable"

esotrope: Man what a nasty gotcha

PR #587, opened 2017-11-10
Added dependency on common-ngx from the deleteProjectDependencies tas…

bgsavage:
…k.  This ensures that when gradle buildFe is called, that common-ngx is also built.

### DIFF
--- a/admin-webapp/build.gradle
+++ b/admin-webapp/build.gradle
@@ -69,6 +69,11 @@ bootRun {
 // generate coverage report automatically (./build/reports/jacoco/test/html/index.html)
 test.finalizedBy(jacocoTestReport)
 
+task deleteProjectDependencies(type: Delete) {
+    delete "src/main/webapp/node_modules/@sbac/rdw-reporting-common-ngx"
+    dependsOn project(':rdw-reporting-common-ngx').tasks.build
+}
+
 //This is the build task for the Front End (Fe) and only compiles when the inputs change
 task buildFe(type: NpmTask) {
     inputs.files(fileTree("src/main/webapp/src"))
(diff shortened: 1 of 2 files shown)


webapp/build.gradle
```diff
@@ -63,6 +63,7 @@ test.finalizedBy(jacocoTestReport)
 
 task deleteProjectDependencies(type: Delete) {
     delete "src/main/webapp/node_modules/@sbac/rdw-reporting-common-ngx"
+    dependsOn project(':rdw-reporting-common-ngx').tasks.buildFe
 }
```
bgsavage: I created this as a separate PR as I didn't want this to be buried in all my other feature changes in the previous pr.

gradle buildFe will now trigger a common-ngx buildFe.

chdietz: every time or only if it has changed?

esotrope: the addition of `inputs/outputs` to the common-ngx `build.gradle` config supposedly allow caching/change watching to work so gradle can reuse builds if the contents of those haven't changed

bgsavage: This is what I've observed as well.

bgsavage: What's crazy is this same exact line of code in webapp-admin throws this error:  
```
 Could not get unknown property 'buildFe' for task set of type org.gradle.api.internal.tasks.DefaultTaskContainer.
```
Any thoughts?

admin-webapp/build.gradle
```diff
@@ -69,6 +69,11 @@ bootRun {
 // generate coverage report automatically (./build/reports/jacoco/test/html/index.html)
 test.finalizedBy(jacocoTestReport)
 
+task deleteProjectDependencies(type: Delete) {
+    delete "src/main/webapp/node_modules/@sbac/rdw-reporting-common-ngx"
+    dependsOn project(':rdw-reporting-common-ngx').tasks.build
```
bgsavage: `build` seems to work in admin, though I still don't understand why `buildFe` doesn't work.

Any gradle experts can enlighten me here?  @wtritch @chdietz @malaffoon  ?

Also, should I update the webapp gradle to use `build` as well for consistency?  `buildFe` in theory should be slightly more efficient and as it won't try to compile java code that's not present in common-ngx.

PR #588, opened 2017-11-10
Feature/source icons from ui kit

esotrope:
Removes redundant icon images from admin and reporting webapps by moving them to the ui kit.
The build now sources these assets from the UI kit.
Also, I cleaned up lingering standalone mode artifacts


admin-webapp/src/main/webapp/.angular-cli.json
```diff
@@ -9,11 +9,15 @@
       "outDir": "../resources/static",
       "assets": [
         "assets",
-        "favicon.ico",
```
esotrope: before we were only including favicon.ico instead of all the app icon iterations

jtreuting:
These are just the SBAC logo right that could be shared by the IAT project, so moving to the UI Kit would allow for that.  Is that right?

esotrope:
@jtreuting correct - also tagged Alex and Brett

PR #589, opened 2017-11-13
correcting MySQL queries and adding integration test case

dsnyder-fairway:
This is to correct the MySQL queries for distractor analysis


webapp/src/main/resources/application.sql.yml
```diff
@@ -561,7 +561,7 @@ sql:
           i.type
         from item i
         where i.asmt_id=:assessment_id
-        and (:types is null or i.type in (:types))
```
jtreuting: how come the null check didn't work?  is it passing in an empty array instead maybe?

dsnyder-fairway: Null check didn't work for two reasons:
1. Say types = 'MC', 'MS' -- this would result in ('MC', 'MS' is null or type in ('MC', 'MS')). The first half would cause an error
2. Say types is null -- this would result in (null is null or type in (null)). MySQL didn't like the latter half

jtreuting: ahh right, makes sense.  thanks.

esotrope: Interesting. Sounds like optional collections will need flags/repo logic to accompany them then.

PR #590, opened 2017-11-13
Add a transfer-access display toggle to both the Advanced Filters and…

wtritch:
… report creation dialog.

image

image

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.reporting.search.user;
 
+import org.opentestsystem.rdw.reporting.common.configuration.TenantProperties;
 import org.opentestsystem.rdw.reporting.configuration.ApplicationProperties;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -10,10 +11,13 @@ import java.util.List;
 class AppSettings {
 
     private final ApplicationProperties applicationProperties;
+    private final TenantProperties tenantProperties;
 
     @Autowired
-    private AppSettings(final ApplicationProperties applicationProperties) {
+    private AppSettings(final ApplicationProperties applicationProperties,
+                        final TenantProperties tenantProperties) {
         this.applicationProperties = applicationProperties;
+        this.tenantProperties = tenantProperties;
     }
 
     /**
@@ -70,6 +74,13 @@ class AppSettings {
      */
     public List<String> getUiLanguages() {
         return applicationProperties.getUiLanguages();
+    }
+
+    /**
+     * @return True if transfer access is enabled
+     */
+    public boolean isTransferAccess() {
+        return tenantProperties.isTransferAccessEnabled();
     }
 
 }
--- a/webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
@@ -81,6 +81,7 @@ export class AssessmentExamMapper {
     uiModel.administrativeCondition = apiModel.administrativeConditionCode;
     uiModel.completeness = apiModel.completenessCode;
     uiModel.schoolYear = apiModel.schoolYear;
+    uiModel.transfer = apiModel.transfer;
 
     let school: School = new School();
     school.name = apiModel.school.name;
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.spec.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.spec.ts
@@ -1,7 +1,5 @@
 import { async, TestBed } from "@angular/core/testing";
 import { Observable, Observer } from "rxjs";
-import { HttpModule } from "@angular/http";
-import { FormsModule } from "@angular/forms";
 import { ActivatedRoute } from "@angular/router";
 import { APP_BASE_HREF } from "@angular/common";
 import { AssessmentsComponent } from "./assessments.component";
@@ -11,11 +9,9 @@ import { GroupAssessmentService } from "../groups/results/group-assessment.servi
 import { Assessment } from "./model/assessment.model";
 import { AssessmentExam } from "./model/assessment-exam.model";
 import { ExamFilterOptions } from "./model/exam-filter-options.model";
-import { Component } from "@angular/core";
+import { NO_ERRORS_SCHEMA } from "@angular/core";
 import { CommonModule } from "../shared/common.module";
-import { AssessmentsModule } from "./assessments.module";
-import { Angulartics2Module, Angulartics2 } from 'angulartics2';
-import { PopoverModule } from "ngx-bootstrap";
+import { Angulartics2 } from 'angulartics2';
 import { UserService } from "../user/user.service";
 import { MockUserService } from "../../test/mock.user.service";
 
@@ -33,38 +29,34 @@ describe('AssessmentsComponent', () => {
     mockRouteSnapshot = getRouteSnapshot();
     TestBed.configureTestingModule({
       imports: [
-        HttpModule,
-        FormsModule,
-        CommonModule,
-        AssessmentsModule,
-        Angulartics2Module,
-        PopoverModule.forRoot()
+        CommonModule
       ],
-      declarations: [TestComponentWrapper],
-      providers: [ { provide: APP_BASE_HREF, useValue: '/' }, {
-        provide: ActivatedRoute,
-        useValue: { snapshot: mockRouteSnapshot }
-      }, {
-        provide: ExamFilterOptionsService,
-        useClass: MockExamFilterOptionsService
-      }, {
-        provide: GroupAssessmentService,
-        useClass: MockAssessmentService
-      }, {
-        provide: Angulartics2,
-        useValue: mockAngulartics2
-      }, {
-        provide: UserService,
-        useClass: MockUserService
-      }]
+      declarations: [
+        AssessmentsComponent
+      ],
+      providers: [
+        { provide: APP_BASE_HREF, useValue: '/' },
+        { provide: ActivatedRoute, useValue: { snapshot: mockRouteSnapshot } },
+        { provide: ExamFilterOptionsService, useClass: MockExamFilterOptionsService },
+        { provide: GroupAssessmentService, useClass: MockAssessmentService },
+        { provide: Angulartics2, useValue: mockAngulartics2 },
+        { provide: UserService, useClass: MockUserService }
+      ],
+      schemas: [ NO_ERRORS_SCHEMA ]
     }).compileComponents();
 
   }));
 
-  beforeEach(() => {
-    fixture = TestBed.createComponent(TestComponentWrapper);
-    component = fixture.debugElement.children[ 0 ].componentInstance;
+  beforeEach((done) => {
+    fixture = TestBed.createComponent(AssessmentsComponent);
+    component = fixture.componentInstance;
+    component.assessmentExams = [ new AssessmentExam(), new AssessmentExam() ];
+    component.assessmentProvider = new MockAssessmentService() as any;
+
     fixture.detectChanges();
+    fixture.whenStable().then(() => {
+      done();
+    });
   });
 
   it('should create component', () => {
@@ -181,14 +173,3 @@ class MockAssessmentService {
     return observable;
   }
 }
-
-@Component({
-  selector: 'test-component-wrapper',
-  template: '<assessments [assessmentExams]="assessments" [assessmentProvider]="assessmentProvider"></assessments>'
-})
-class TestComponentWrapper {
-  assessments = [ new AssessmentExam(), new AssessmentExam() ];
-  assessmentProvider = new MockAssessmentService();
-}
-
-
--- a/webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
@@ -8,20 +8,54 @@
   <div id="filter-by-test" class="flex-child">
     <p class="border-bottom">{{ translateRoot + 'test.label' | translate }}</p>
     <div class="block-children">
-      <label>
-          <span info-label
-                title="{{ translateRoot + 'test.off-grade-assessment' | translate }}"
-                content="{{ translateRoot + 'test.off-grade-assessment-info' | translate }}"></span>
-      </label>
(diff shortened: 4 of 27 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/search/user/AppSettings.java
```diff
@@ -72,4 +76,11 @@ public String getAdminWebappUrl() {
         return applicationProperties.getUiLanguages();
     }
 
+    /**
+     * @return True if transfer access is enabled
+     */
+    public boolean isTransferAccess() {
+        return tenantProperties.isTransferAccessEnabled();
+    }
```
wtritch: This tells the front-end whether or not transfer-access is enabled in the system's configuration.

webapp/src/main/webapp/src/app/assessments/assessments.component.spec.ts
```diff
@@ -33,38 +29,34 @@ describe('AssessmentsComponent', () => {
     mockRouteSnapshot = getRouteSnapshot();
     TestBed.configureTestingModule({
       imports: [
-        HttpModule,
-        FormsModule,
-        CommonModule,
-        AssessmentsModule,
-        Angulartics2Module,
-        PopoverModule.forRoot()
```
wtritch: unrelated, cleaned up some of the worst-performing component tests

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -8,20 +8,54 @@
...
         </label>
-        <label class="btn btn-primary" [ngClass]="{'active': filterBy.offGradeAssessment}">
-          <input [value]="true" [(ngModel)]="filterBy.offGradeAssessment" type="radio"
-                 angulartics2On="click" angularticsEvent="FilterClick" angularticsCategory="AdvancedFilters" [angularticsProperties]="{label: 'Off-Grade: Hide'}">{{ 'buttons.visibility.hide' | translate }}
+        <sb-toggle name="off-grade-assessment-toggle"
+                   [(ngModel)]="filterBy.offGradeAssessment"
+                   [options]="[
+                           {value: false, text: 'buttons.visibility.show' | translate, angularticsProperties: {label: 'Off-Grade: Show'}},
+                           {value: true, text: 'buttons.visibility.hide' | translate, angularticsProperties: {label: 'Off-Grade: Hide'}}
+                         ]"></sb-toggle>
```
wtritch: I converted this to use the sb-toggle component.

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -8,20 +8,54 @@
...
+            <span info-label
+                  title="{{ translateRoot + 'test.transfer-assessment' | translate }}"
+                  content="{{ translateRoot + 'test.transfer-assessment-info' | translate }}"></span>
         </label>
+        <sb-toggle name="transfer-assessment-toggle"
+                   [(ngModel)]="filterBy.transferAssessment"
+                   [options]="[
+                           {value: false, text: 'buttons.visibility.show' | translate, angularticsProperties: {label: 'Transfer-Assessment: Show'}},
+                           {value: true, text: 'buttons.visibility.hide' | translate, angularticsProperties: {label: 'Transfer-Assessment: Hide'}}
+                         ]"></sb-toggle>
```
wtritch: Added the new Transfer Assessment show/hide toggle IFF transfer access is enabled.  I'm also disabling the ability to turn off transfer assessments when looking at Student History exams, since it doesn't affect any scores or other aggregate data.

jtreuting:
Do the 2 screens have different defaults?  Shouldn't those be the same, and I think the want it to be on by default.

Also seems like the (i) icon would be useful in the PDF dialog as well since it might be not clear what that is

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -48,6 +48,17 @@ <h4 class="modal-title pull-left">{{title}}</h4>
...
+            <div class="form-group" *ngIf="transferAccess">
+              <label for="transfer-assessment-toggle">{{'labels.groups.results.adv-filters.test.transfer-assessment' | translate}}</label>
+              <sb-toggle name="transfer-assessment-toggle"
+                         class="clearfix"
+                         [(ngModel)]="options.disableTransferAccess"
+                         [options]="[
+                           {value: false, text: 'buttons.visibility.show' | translate},
+                           {value: true, text: 'buttons.visibility.hide' | translate}
+                         ]"></sb-toggle>
+            </div>
```
wtritch: Add the Transfer Assessment toggle to the report creation component IFF transfer access is enabled in the system.

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -122,7 +121,8 @@ export class ReportService {
       language: options.language,
       grayscale: options.grayscale,
       accommodationsVisible: options.accommodationsVisible,
-      order: ReportOrder[ options.order ]
+      order: ReportOrder[ options.order ],
+      disableTransferAccess: options.disableTransferAccess
```
wtritch: Send the enable/disable transfer access property in the report request payload.

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -19,10 +18,14 @@ const CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR = {
   template: `
     <div class="btn-group btn-group-sm toggle-group" data-toggle="buttons">
       <label class="btn btn-primary" [ngClass]="{active: options[0].value === value, disabled: disabled}">
-        <input type="radio" [name]="name" [disabled]="disabled" [value]="options[0].value" [(ngModel)]="value">{{options[0].text}}
+        <input type="radio" [id]="name" [name]="name" [disabled]="disabled" [value]="options[0].value" [(ngModel)]="value"
+               angulartics2On="click" [angularticsEvent]="analyticsEvent"
+               [angularticsCategory]="analyticsCategory" [angularticsProperties]="options[0].angularticsProperties || {label: options[0].text}">{{options[0].text}}
```
wtritch: Unlike the other sb-* controls, the toggle didn't include angulartics... I've added it in here but since I'm not sure how to test angulartics locally I can't be sure this will function as expected.  How do we test angulartics behavior?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -219,7 +219,9 @@
           "test": {
             "label": "Test",
             "off-grade-assessment": "Off-Grade Assessments",
-            "off-grade-assessment-info": "Schools may administer interim assessments for any grade.  This filter indicates that a student took an assessment for a grade other than their assigned grade."
+            "off-grade-assessment-info": "Schools may administer interim assessments for any grade.  This filter indicates that a student took an assessment for a grade other than their assigned grade.",
+            "transfer-assessment": "Transfer Assessments",
+            "transfer-assessment-info": "Indicates whether to show or hide assessments administered outside of the User's standard institutions."
```
wtritch: @jtreuting I shoved some placeholder text in here, but we will probably want to get "official" text from SBAC at some point.

wtritch:
@jtreuting they have the same "Show" default, I just showed the Advanced Filters in the user-activated state to show that the toggle also appears and is properly translated =)

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -81,6 +81,7 @@ export class AssessmentExamMapper {
     uiModel.administrativeCondition = apiModel.administrativeConditionCode;
     uiModel.completeness = apiModel.completenessCode;
     uiModel.schoolYear = apiModel.schoolYear;
+    uiModel.transfer = apiModel.transfer;
```
bgsavage: nit: more descriptive naming.  `isStudentTransferred` maybe?

wtritch: I’m not a fan of putting “is” on property names, from a java bean perspective, its like putting “get” on the property name.  Maybe transferEnabled?

webapp/src/main/webapp/src/app/assessments/assessments.component.spec.ts
```diff
@@ -33,38 +29,34 @@ describe('AssessmentsComponent', () => {
...
   }));
 
-  beforeEach(() => {
-    fixture = TestBed.createComponent(TestComponentWrapper);
-    component = fixture.debugElement.children[ 0 ].componentInstance;
+  beforeEach((done) => {
+    fixture = TestBed.createComponent(AssessmentsComponent);
+    component = fixture.componentInstance;
+    component.assessmentExams = [ new AssessmentExam(), new AssessmentExam() ];
+    component.assessmentProvider = new MockAssessmentService() as any;
```
bgsavage: The reason for the testwrapper before was to call and use the component the same as before.  By setting it this way, ngInit is being tested in a different way then it's actually being used.  

Difference being with this approach, `ngInit` was called without `assessmentExams` and `assessmentProvider` being set.  With the other approach, `ngInit` was called with those guys set already.

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
bgsavage: 👍

webapp/src/main/webapp/src/app/assessments/assessments.component.spec.ts
wtritch: hmm... I was having issues getting an instance of the AssessmentsComponent out of the TestWrapper, and it felt a little awkward.  Maybe I could call ngInit() manually before setting those values?

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
bgsavage: So this is a boolean and it's true if the exam was transferred?

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: Perfect, I'll make a note of asking for official content for that.

webapp/src/main/webapp/src/app/assessments/assessments.component.spec.ts
bgsavage: It should just be `component = fixture.debugElement.children[ 0 ].componentInstance;`

bgsavage: Also, you'd have to call ngInit() after manually setting those values.  Though, ngInit will already have been run once from when the component was created, so I'm not sure what that buys us...

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
wtritch: yessir

webapp/src/main/webapp/src/app/assessments/assessments.component.spec.ts
bgsavage: Nvm on these commnents, we demo'd offline and looks good!  I didn't realize that `.detectChanges` in the end is when `ngOnInit` gets invoked, so we are good here and this is a much better approach.

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -48,6 +48,21 @@ <h4 class="modal-title pull-left">{{title}}</h4>
...
             </div>
 
+            <div class="form-group" *ngIf="transferAccess">
+              <label>
+                <span info-label
+                      title="{{'labels.groups.results.adv-filters.test.transfer-assessment' | translate}}"
+                      content="{{'labels.groups.results.adv-filters.test.transfer-assessment-info' | translate}}"></span>
+              </label>
+              <sb-toggle name="transfer-assessment-toggle"
+                         class="clearfix"
```
bgsavage: Curious as what caused you to add clearfix here, I don't think we had to this on any of the others?

wtritch: The other labels are *just* long enough that their "inline-block" style drops them to the next level.  Without this, I end up with the toggle nestled to the right of the "Transfer Assessment" label in the picture above =/

wtritch: or... at least I did before I put the (i) icon in there.  Let me see if that's a "block" or "inline-block" element.

wtritch: Without the clearfix I end up with:
image
...which maybe people prefer?  Although I dislike when one control doesn't match any of the others...

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
bgsavage: Ask @jtreuting to check google analytics and see if it's getting new data :P

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -33,9 +36,19 @@ export class SBToggleComponent implements ControlValueAccessor, OnInit {
   @Input()
   disabled: boolean = false;
 
+  @Input()
+  name: string;
+
+  // The analytics event to send when clicked
```
bgsavage: nit: should be /**

wtritch: good call, changed.

webapp/src/main/webapp/src/app/report/report-download.component.html
bgsavage: Ah so would display:block fix this?  I would think we'd want to do this for the others and not rely on chance by how long the labels are, especially since these are provided by translate values.

bgsavage: (or just maybe wrap in a <div></div>)

wtritch: I'm really struggling with that.  The idea behind bootstrap is that things just flow to where they're supposed to go, so that you don't have to manually futz with stuff depending upon how wide the display is.  I'm not sure that I'm willing to pick up that responsibility, so maybe I should just remove the clearfix and if we decide we want all of our label-and-controls to display on two levels we configure it globally.  Thoughts @aaronsleeper ?

wtritch: For now I'm going to remove the clearfix.  If we want different behavior for our labels and controls it should be a global change.

bgsavage: I agree with that, should be consistent either way.

PR #591, opened 2017-11-13
Added label-max-width to datatable-row-expander to trim long labels w…

bgsavage:
…ith ellipses (including long file names).

Note:  This also fixes in other places that use the same common component, i.e. the student name in Items By Results.

### DIFF
--- a/admin-webapp/src/main/webapp/src/styles.less
+++ b/admin-webapp/src/main/webapp/src/styles.less
@@ -195,3 +195,9 @@ footer {
 .font-weight-light {
   font-weight: lighter;
 }
+
+.label-max-width {
+  text-overflow: ellipsis;
+  overflow: hidden;
+  max-width: 330px;
+}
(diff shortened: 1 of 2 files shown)


admin-webapp/src/main/webapp/src/styles.less
```diff
@@ -195,3 +195,9 @@ footer {
 .font-weight-light {
   font-weight: lighter;
 }
+
+.label-max-width {
+  text-overflow: ellipsis;
+  overflow: hidden;
+  max-width: 330px;
+}
```
bgsavage: I think we'll need to consider a longer term strategy for when common-ngx has style dependencies.

jtreuting:
There was already one place that we handled this (was it group name or filename on the My Reports, i forget).  So does this change align with that previous change and if not, should we change the existing way it is handled and consolidate on a single approach?

bgsavage:
This is the same approach I did for My Reports.  It was the file name.  The difference though is that it wasn't on a row expander, just on a button label.  But it applied this same css class that used to live just in the reporting webapp.

admin-webapp/src/main/webapp/src/styles.less
esotrope: Hm. i would hesitate to apply this style to the expander. Not sure that is very flexible.
I would rather have a selector like `.my-container datatable-row-expander` that is page/view specific.

PR #592, opened 2017-11-13
Fix Build

wtritch:
I believe I was importing the wrong version of Observable.

### DIFF
--- a/webapp/src/main/webapp/src/test/mock.router.ts
+++ b/webapp/src/main/webapp/src/test/mock.router.ts
@@ -1,5 +1,5 @@
 import Spy = jasmine.Spy;
-import { Observable } from "rxjs/Observable";
+import { Observable } from "rxjs";
 import { UrlTree } from "@angular/router";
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/test/mock.router.ts
```diff
@@ -1,5 +1,5 @@
 import Spy = jasmine.Spy;
-import { Observable } from "rxjs/Observable";
+import { Observable } from "rxjs";
```
wtritch: The Observable from "rxjs/Observable" doesn't contain the static methods empty() or of()...  It's annoying that I can't reproduce the build breakage locally, however...

bgsavage:
LGTM

webapp/src/main/webapp/src/test/mock.router.ts
esotrope: rxjs is the cause of so many build failures now

PR #593, opened 2017-11-14
added task to periodically clean up user reports

dsnyder-fairway:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
```diff
@@ -0,0 +1,48 @@
...
+    @Autowired
+    RemoveStaleReportsTask(final ReportContentService reportContentService, final DefaultReportService reportService) {
+        this.reportContentService = reportContentService;
+        this.reportService = reportService;
+    }
+
+    @Scheduled(cron = "${task.remove-stale-reports.cron}")
+    public void removeStaleReports() {
+        try {
+            Thread.sleep((long) (Math.random() * aMinute * maxSleepMinutes));
```
dsnyder-fairway: add some random delay to avoid collisions with other report processors

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -67,6 +71,26 @@ public ReportContentResource openContent(final Report report) {
...
+
+    @Override
+    public void deleteContent(final Report report) {
+        final URI uri = report.getReportResourceUri();
+        if (uri != null) {
+            final String location = uri.toASCIIString();
+            try {
+                archiveService.delete(location);
+            } catch (RuntimeException e) {
+                logger.warn("failed to delete location {}", location);
```
dsnyder-fairway: catch and log since this is grabbing all items older than 30 days. If we fail to delete this from S3 just log that we were unable to remove it

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -67,6 +71,26 @@ public ReportContentResource openContent(final Report report) {
         return new ReportContentResource(archiveService.openResource(location), reportLength, mediaType);
     }
 
+    @Override
+    public void deleteContents(final Collection<Report> reports) {
```
bgsavage: Couldn't you have a single method with signature `Report... reports` instead?  And then possibly make the below method private?

dsnyder-fairway: yep, I'll change this

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportContentService.java
```diff
@@ -34,4 +35,16 @@ Report writeContent(Report report,
...
     ReportContentResource openContent(Report report);
+
+    /**
+     * Delete a collection of report contents from persistent storage.
+     * @param reports The collection of report contents to delete
+     */
+    void deleteContents(Collection<Report> reports);
+
+    /**
+     * Delete a reprot from persistent storage
```
bgsavage: nit: typo

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
```diff
@@ -35,6 +35,11 @@ public Report create(final Report report) {
     }
 
     @Override
+    public List<Report> findAllOlderThanAmountOfDays(final int numberOfDays) {
+        return reportRepository.findAllOlderThanAmountOfDays(numberOfDays);
```
bgsavage: nit: prefer consistency in terminology of amount vs. number.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
bgsavage: We couldn't randomize the schedule instead?

dsnyder-fairway: @Scheduled only takes cron, fixedDelay, or fixedRate

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
```diff
@@ -46,6 +51,11 @@ public Report update(final Report report) {
     }
 
     @Override
+    public void delete(final Collection<Long> ids) {
+        if (!ids.isEmpty()) reportRepository.deleteIds(ids);
```
esotrope: This guard should be logic in the service layer

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
esotrope: `@Scheduled` doesn't take a function? I'm surprised.
What collision are we specifically avoiding btw?

bgsavage: How about a property getter?

dsnyder-fairway: @esotrope avoiding collisions where we would try to delete something that has already been deleted in S3. I suppose it might be ok to do this as I do not fail in this case -- but it would make the logs pretty noisy

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultReportService.java
dsnyder-fairway: isn't this already in the service layer?

esotrope: Sorry. I'm sloppily skimming again. I'll rephrase - thanks for putting this in the service layer : J 👍

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
dsnyder-fairway: @bgsavage I don't know if I could use a property getter to get a random number

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -67,6 +71,25 @@ public ReportContentResource openContent(final Report report) {
...
+        }
+    }
+
+    private void deleteContent(final Report report) {
+        final URI uri = report.getReportResourceUri();
+        if (uri != null) {
+            final String location = uri.toASCIIString();
+            try {
+                archiveService.delete(location);
+            } catch (RuntimeException e) {
```
malaffoon: nit: exception can be final

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -67,6 +71,25 @@ public ReportContentResource openContent(final Report report) {
...
+                logger.warn("failed to delete location {}", location);
```
malaffoon: Please include `e.getMessage()` in the logger output.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
```diff
@@ -0,0 +1,48 @@
...
+public class RemoveStaleReportsTask {
+
+    private final ReportContentService reportContentService;
+    private final DefaultReportService reportService;
+    private static final long aMinute = 60000L;
+
+    @Value("${task.remove-stale-reports.number-of-days}")
+    private int numberOfDays;
+
+    @Value("${task.remove-stale-reports.maximum-sleep-minutes}")
```
malaffoon: Can we come up with a better name for this property? The fact that we happen to be using `sleep` is an implementation detail. Perhaps `max-delay-minutes` or `max-random-minutes`?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
```diff
@@ -0,0 +1,48 @@
...
+        this.reportService = reportService;
+    }
+
+    @Scheduled(cron = "${task.remove-stale-reports.cron}")
+    public void removeStaleReports() {
+        try {
+            Thread.sleep((long) (Math.random() * aMinute * maxSleepMinutes));
+        } catch (InterruptedException e) {
+        }
+        final Collection<Report> reports = reportService.findAllOlderThanNumberOfDays(numberOfDays);
```
malaffoon: The task should be calling a single method on a single service to do its work. Much like controllers, you want to avoid putting any sort of business/implementation logic in the task itself. Imagine that we add an actuator end-point that manually triggers the cleanup work: we wouldn't want to duplicate these lines of code there: we'd want to inject the same service and make the same single call.

report-processor/src/main/resources/application.yml
```diff
@@ -89,3 +89,28 @@ app:
...
       page-size: Letter
       zoom: 0.75
+
+task:
+# by default, all tasks are disabled for local dev environment; uncomment and set secrets/hosts to test.
+#  remove-stale-reports:
+#    cron: 0 0 * * * *
+#    number-of-days: 30
+#    maximum-sleep-minutes: 20
+#    query: ???
```
malaffoon: I realize you're copying and pasting from a more complex task; lines 99-116 can be deleted.

report-processor/src/main/resources/application.yml
```diff
@@ -89,3 +89,28 @@ app:
     options:
       page-size: Letter
       zoom: 0.75
+
+task:
+# by default, all tasks are disabled for local dev environment; uncomment and set secrets/hosts to test.
+#  remove-stale-reports:
+#    cron: 0 0 * * * *
```
malaffoon: I'm not sure we need to run this every hour. And it's always nice to document what  a cron expression is trying to do, something like:
```
#    # run at 8 (GMT) every day
#    cron: 0 0 8 * * *
```

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
malaffoon: We shouldn't delete the corresponding database record if this fails, so somehow this should be communicated to the main delete loop. Imagine a connectivity problem with S3: none of the content will be removed yet all the database records will be wiped, destroying the necessary information to try again later.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
malaffoon: If we were feeling adventurous we could make @ScheduledRandom that extends @Scheduled and takes a max-random option. Not saying to change anything, just suggesting it.

report-processor/src/main/resources/application.yml
malaffoon: Oh, i just noticed @Scheduled takes a `zone` option, so we *could* use that to enforce the GMT part of that comment. I'll have to retrofit that into the other tasks.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
bgsavage: I guess when we talked about this in the parking lot the other day, I thought that we'd have the schedule be random but we'd always know what the random schedule was.  Like maybe generating the random schedule should be more of a configuration concern rather than an application concern, because we wouldn't even need to do this if we were running a single instance, and would cause confusion / complication when the task wasn't run at it's configured scheduled time.  Maybe we can discuss again in the parking lot?  This feels off to me...

report-processor/src/main/resources/application.yml
dsnyder-fairway: Yeah I should have left a comment. The intent was every day at midnight to run. According to https://crontab.guru/every-night-at-midnight it should be 0 0 * * *. I accidentally left another * which I will remove

report-processor/src/main/resources/application.yml
dsnyder-fairway: How will the S3 configurations be picked up? From running this locally it looks like LocalArchiveService gets picked up which I imagine is just because I am missing the S3 configurations since the ArchiveConfiguration can't find the configurations

report-processor/src/main/resources/application.yml
```diff
@@ -89,3 +89,28 @@ app:
...
+#    number-of-days: 30
```
wtritch: maybe something a little more descriptive here like "max-report-lifetime-days" or something along those lines?  "number-of-days" doesn't really tell me what this property is used for. =)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
dsnyder-fairway: changed to `max-random-minutes`

report-processor/src/main/resources/application.yml
dsnyder-fairway: removed a * and left a comment

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
```diff
@@ -0,0 +1,48 @@
...
+    @Autowired
+    RemoveStaleReportsTask(final ReportContentService reportContentService, final DefaultReportService reportService) {
+        this.reportContentService = reportContentService;
+        this.reportService = reportService;
+    }
+
+    @Scheduled(cron = "${task.remove-stale-reports.cron}")
+    public void removeStaleReports() {
+        try {
+            Thread.sleep((long) (Math.random() * aMinute * maxRandomMinutes));
```
wtritch: hmm... rather than sleeping is is strange that I'd rather shove it back into a ScheduledTaskExecutor with a randomized delay?  Sleeping always makes me cringe a bit.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
dsnyder-fairway: went ahead with Will's suggestion

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
dsnyder-fairway: done

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
dsnyder-fairway: me too 😄 . I'll look into ScheduledTaskExecutor

report-processor/src/main/resources/application.yml
malaffoon: Spring @Scheduled cron is *not* the same as linux cron. I believe the example i gave is more correct.

report-processor/src/main/resources/application.yml
malaffoon: Well, there will be something here, just not `senders`. You are correct, you'll need to add the configuration for the archive service that is being used.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
malaffoon: I'm not going to fight too hard, but I don't like the standard exception handling when the exception is "handled". That will put a stack trace in the log. I guess in this case it might be warranted but in my opinion any stack trace in the log indicates that the code is not handling something properly.

report-processor/src/main/resources/application.yml
dsnyder-fairway: @malaffoon, my bad! Thanks for bringing this up. It looks like they include the first one as seconds. I'll follow your example for an 8am run

report-processor/src/main/resources/application.yml
dsnyder-fairway: changed

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
dsnyder-fairway: there is a new implementation

report-processor/src/main/resources/application.yml
dsnyder-fairway: updated

report-processor/src/main/resources/application.yml
dsnyder-fairway: updated

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
dsnyder-fairway: done. There was a slight complication. I ran into a cycle when moving this to the report service between the DefaultReportService and the ArchiveBackedReportContentService. For this reason I moved this to a new class and interface (RemoveReportService and DefaultRemoveReportService)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
```diff
@@ -1,48 +1,48 @@
...
+@EnableScheduling
+@ConditionalOnProperty(prefix = "task.remove-stale-reports", name = {"cron", "max-report-lifetime-days", "max-random-minutes"})
 public class RemoveStaleReportsTask {
 
-    private final ReportContentService reportContentService;
-    private final DefaultReportService reportService;
+    private static final Logger logger = LoggerFactory.getLogger(RemoveStaleReportsTask.class);
+    private final RemoveReportService removeReportService;
+    private final TaskScheduler taskScheduler;
     private static final long aMinute = 60000L;
```
esotrope: would prefer calling this `minuteInMilliseconds`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportContentService.java
```diff
@@ -41,6 +41,6 @@ Report writeContent(Report report,
      *
      * @param reports The collection of report contents to delete
      */
-    void deleteContents(Collection<Report> reports);
+    Collection<Report> deleteContents(Collection<Report> reports);
```
esotrope: I think we should say here whether we return a set or list

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/RemoveStaleReportsApplication.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+@Configuration
+public class RemoveStaleReportsApplication {
+
+    @Bean
+    public ThreadPoolTaskScheduler threadPoolTaskScheduler() {
+        ThreadPoolTaskScheduler threadPoolTaskScheduler = new ThreadPoolTaskScheduler();
```
esotrope: nit: `final`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
```diff
@@ -0,0 +1,48 @@
...
+    @Value("${task.remove-stale-reports.max-random-minutes}")
+    int maxRandomMinutes;
+
+    @Autowired
+    RemoveStaleReportsTask(final RemoveReportService removeReportService, final TaskScheduler taskScheduler) {
+        this.removeReportService = removeReportService;
+        this.taskScheduler = taskScheduler;
+    }
+
+    @Scheduled(cron = "${task.remove-stale-reports.cron}", zone = "GMT")
```
dsnyder-fairway: @malaffoon this addresses DWR-1159. The reason I did not include a UT for this is the only way I knew how to test the annotation was found here https://stackoverflow.com/questions/17327956/testing-scheduled-in-spring, which acted more of a demonstration than a test of the implementation

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultRemoveReportService.java
```diff
@@ -0,0 +1,30 @@
...
+import org.opentestsystem.rdw.reporting.common.model.Report;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Service
+public class DefaultRemoveReportService implements RemoveReportService {
```
esotrope: ReportRemovalService comes off the tongue a little better. or maybe ReportDisposalService with a dump truck ASCII art pic in the comments : J

dsnyder-fairway: I'll rename

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/RemoveStaleReportsApplication.java
dsnyder-fairway: changed

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
dsnyder-fairway: renamed

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportContentService.java
dsnyder-fairway: I guess this should really be a Set since every report should be unique...I can make this change

dsnyder-fairway: done

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/RemoveStaleReportsApplication.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.reporting.processor;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+@Configuration
+public class RemoveStaleReportsApplication {
```
malaffoon: Curious about the naming ... is this an Application or a Configuration?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/tasking/RemoveStaleReportsTask.java
```diff
@@ -0,0 +1,48 @@
...
+    public void removeStaleReports() {
+        taskScheduler.schedule(new Runnable() {
+            @Override
+            public void run() {
+                logger.info("Deleting stale reports");
+                reportRemovalService.deleteOldReports(maxReportLifetimeDays);
```
malaffoon: 👍

report-processor/src/main/resources/application.yml
```diff
@@ -89,3 +97,12 @@ app:
     options:
       page-size: Letter
       zoom: 0.75
+
+task:
+# by default, all tasks are disabled for local dev environment; uncomment and set secrets/hosts to test.
+#  remove-stale-reports:
+##### at seconds: (0), minutes: (0), hour: (8), every day of month (*), every month (*), every day of week (*).
+##### This runs at 8am
```
malaffoon: nit: please mention that it is GMT

report-processor/src/main/resources/application.yml
```diff
@@ -68,6 +68,14 @@ app:
 
   archive:
     root: file:///tmp/
+#   root: s3:///bucket
```
malaffoon: Are you putting these commented lines in to pseudo-document?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportContentService.java
malaffoon: In general, parameters should be a loosely typed as possible (Iterable > Collection > Set) while return values should be more typed.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/RemoveStaleReportsApplication.java
dsnyder-fairway: should be a configuration. I'll change this

report-processor/src/main/resources/application.yml
dsnyder-fairway: yeah it was to document. I'll remove these

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportContentService.java
dsnyder-fairway: I'll update this again

report-processor/src/main/resources/application.yml
dsnyder-fairway: done

dsnyder-fairway:
✅ ?

PR #594, opened 2017-11-14
Add missing Observable.* static methods as imports.

wtritch:


common-ngx/src/data/caching-data.service.ts
```diff
@@ -5,6 +5,7 @@ import "rxjs/add/operator/map";
 import "rxjs/add/operator/catch";
 import "rxjs/add/operator/share";
 import "rxjs/add/observable/of";
+import "rxjs/add/observable/throw";
```
wtritch: I must have missed the PR where these were added.  Looks like some static methods were just missed in the original import sets.

esotrope: Do these methods always need to be independently imported? seems like sometimes they don't and sometimes they do. Maybe when we import from `"rxjs"` it comes loaded and not when imported from `"rxjs/Observable"? just guessing.

wtritch: That's my thought, importing from "rxjs" gets the full Observable complete with static methods.  Importing from "rxjs/Observable" just gets the instance-level methods.

PR #595, opened 2017-11-14
Feature/language selection

esotrope:


admin-webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -21,38 +22,30 @@ export class AppComponent {
...
-      this.translate.addLangs(languages);
-      if (languages.indexOf(this.translate.getBrowserLang()) != -1) {
-        this.translate.use(this.translate.getBrowserLang());
-      }
-
-      if (window[ 'ga' ] && user.configuration && user.configuration.analyticsTrackingId) {
-        window[ 'ga' ]('create', user.configuration.analyticsTrackingId, 'auto');
-      }
+      this._user = user;
+      this.languageStore.configuredLanguages = user.configuration.uiLanguages;
```
esotrope: translate service initialization and updates are now tucked into the `language store` provider. We call this setter to tell the service about the server-side configured languages

common-ngx/src/i18n/language.store.ts
```diff
@@ -0,0 +1,74 @@
+import { EventEmitter, Injectable, Output } from "@angular/core";
+import { UserPreferenceService } from "../preference/user-preference.service";
+import { TranslateService } from "@ngx-translate/core";
+import { EmbeddedLanguage, EmbeddedLanguages } from "./language-settings";
+
+/**
+ * This repository stores the language setting for the user
+ */
+@Injectable()
+export class LanguageStore {
```
esotrope: Facade and central model to store and retrieve the languages/current language/language preference.

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -22,38 +22,33 @@ export class AppComponent {
...
-        if (window[ 'ga' ] && user.configuration && user.configuration.analyticsTrackingId) {
-          window[ 'ga' ]('create', user.configuration.analyticsTrackingId, 'auto');
-        }
+        this.languageStore.configuredLanguages = user.configuration.uiLanguages;
+        this.initializeAnalytics(user.configuration.analyticsTrackingId);
       } else {
         this.router.navigate([ 'error' ]);
       }
     });
+    this.initializeNavigationScrollReset();
```
esotrope: we probably want this in admin too right?

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -1,10 +1,10 @@
 import { Component } from "@angular/core";
-import { TranslateService } from "@ngx-translate/core";
 import { UserService } from "./user/user.service";
 import { NavigationEnd, Router } from "@angular/router";
 import { Location, PopStateEvent } from "@angular/common";
 import { isNullOrUndefined } from "util";
 import { User } from "./user/model/user.model";
+import { LanguageStore } from "@sbac/rdw-reporting-common-ngx/i18n";
```
esotrope: with the new way we build common-ngx we can now reference packages of the components.

common-ngx/src/preference/user-preference.service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  getLanguage(): string {
+    return this.cookieService.get(languageCookieKey);
+  }
+
+  setLanguage(languageCode: string): void {
+
+    // Sets cookie to expire in the far future
+    const date: Date = new Date();
+    date.setFullYear(date.getFullYear() + languageCookieYearsToExpire);
```
esotrope: sets the cookie expiry to 10 years out.

admin-webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -21,38 +22,30 @@ export class AppComponent {
...
+      this.initializeAnalytics(user.configuration.analyticsTrackingId);
     });
   }
+
+  private initializeAnalytics(trackingId: string): void {
```
esotrope: this might more appropriately be thrown into a module.forRoot() or provider constructor i'm thinking. also it's common code so I'll probably be moving it to common

common-ngx/src/i18n/language.store.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Initializes the repository with all remotely configured language bundles
+   * Make sure to call this at app init time when knowledge of the remotely available languages becomes available
+   *
+   * @param {string[]} remotelyStoredLanguages
+   */
+  set configuredLanguages(values: string[]) {
+    this.translateService.addLangs(values);
+    this.translateService.use(this.getLanguagePreference());
```
malaffoon: Where is the logic that checks that the preferred language is a valid choice?

common-ngx/src/preference/user-preference.service.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /**
+   * Gets the parent domain of the current domain so as to share the preference across domains.
+   *
+   * @returns {string}
+   */
+  private getParentDomain(): string {
+    const domainParts: string[] = this.window.location.hostname.split('.');
+    return (domainParts.length < 3 ? domainParts : domainParts.slice(1)).join('.');
```
wtritch: Do we want to just remove the top-most subdomain? Or do we want to trim to the lowest two?  What if we're running at "staging.reporting.somewhere.com" "staging.admin.somewhere.com" and "production.reporting.somewhere.com" or some other naming scheme we haven't considered?

webapp/src/main/webapp/src/app/app.component.ts
wtritch: I'd guess so, although admin currently has very little navigation.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -4,6 +4,11 @@
       "title": "Authentication Expired",
       "message": "Click Ok to re-authenticate.",
       "ok": "Ok"
+    },
+    "languages": {
+      "en": "English",
+      "es": "Español",
+      "vi": "Tiếng Việt"
```
wtritch: Hmm... I think these may already exist in here.  Maybe remove the dupes?

common-ngx/src/i18n/language.store.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @returns {string}
+   */
+  get language(): string {
+    return this.translateService.currentLang;
+  }
+
+  set language(value: string) {
+    if (this.language !== value) {
+      this.translateService.use(value);
+      this.userPreferenceService.setLanguage(value);
```
bgsavage: I'm not sure that updating the user preferences should be a concern of the language store.  Setting a language may not always set by a user?

common-ngx/src/i18n/language.store.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.translateService.setDefaultLang(EmbeddedLanguage);
+    this.translateService.use(EmbeddedLanguage);
+  }
+
+  /**
+   * All available languages
+   *
+   * @returns {string[]}
+   */
+  get languages(): string[] {
```
bgsavage: nit: why not just name as availableLanguages?  This can eliminate any confusion between `language` and `languages`

common-ngx/src/i18n/language.store.ts
```diff
@@ -0,0 +1,74 @@
...
+  set languages(values: string[]) {
+    this.translateService.langs = values;
+  }
+
+  /**
+   * The current language selection
+   *
+   * @returns {string}
+   */
+  get language(): string {
```
bgsavage: Same comment here, `currentLanguage`

common-ngx/src/i18n/language.store.ts
esotrope: I think they are pretty tightly coupled in our app.
Are you suggesting there is a case that we change the language without changing the user pref? not sure when that would be - also if we do we can talk to `translate.use` directly

common-ngx/src/i18n/language.store.ts
esotrope: 👍

common-ngx/src/i18n/language.store.ts
bgsavage: I'm just thinking we might not want to make that assumption.   This class should serve as an abstraction to `translate` so we shouldn't have to ever access it directly right?

esotrope: Mainly the goal was to keep all the language change/preference store binding logic that is common to admin/reporting together. This way each app doesn't need to do more than import this component and it's good to go - no assembly required.

bgsavage: 👍

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: 👍 will do

admin-webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -11,7 +11,7 @@ const AdminPermissions = [ 'GROUP_WRITE' ];
   selector: 'app-component',
   templateUrl: './app.component.html'
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
```
esotrope: I guess angular will find and call `ngOnInit` without this - wow 8 .

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -488,7 +474,14 @@
         },
         "report-name": "Report Name",
         "assessment-type": "Assessment Type",
-        "language": "Report Language",
+        "language": {
```
esotrope: report language options have a new home under `labels.reports.form.language.options`

common-ngx/src/preference/user-preference.service.ts
```diff
@@ -38,6 +38,11 @@ export class UserPreferenceService {
   /**
    * Gets the parent domain of the current domain so as to share the preference across domains.
    *
+   * WARNING:
+   * This is intended to enable sharing of the preference across admin and reporting webapp domains
+   * Supports domain naming like: admin.xyz.com, reporting.xyz.com
+   * Does not support naming like: staging.admin.xyz.com, staging.reporting.xyz.com
```
esotrope: Added warning about potential domain sharing issues

admin-webapp/src/main/webapp/src/app/app.component.ts
bgsavage: uhh...... wow lol

PR #596, opened 2017-11-14
Feature/dwr 1140 export demo feedback

bgsavage:
"Minor text fixes".

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -512,10 +512,10 @@
         }
       },
       "reportType": {
-        "Student": "Student",
-        "SchoolGrade": "School/Grade",
-        "Group": "Group",
-        "ExamExport": "Exam Export"
+        "Student": "Student Report",
+        "SchoolGrade": "School/Grade Report",
+        "Group": "Group Report",
+        "ExamExport": "Custom Export"
       },
       "status": {
         "PENDING": "Pending",
@@ -533,8 +533,7 @@
       "remove-district": "Remove District"
     },
     "organization-export": {
-      "link": "Custom Exports",
-      "heading": "Customized Export",
+      "title": "Custom Exports",
       "form": {
         "organization": {
           "label": "Search for a District, School or Institution",
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -527,8 +527,8 @@
       }
     },
     "organization-tree": {
-      "header": "Selected Institutions",
-      "subheader-empty": "No institutions selected",
+      "header": "Selected Organizations",
+      "subheader-empty": "No organizations selected",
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -527,18 +527,17 @@
...
-      "subheader-empty": "No institutions selected",
+      "header": "Selected Organizations",
+      "subheader-empty": "No organizations selected",
       "subheader-not-empty": "Click an item to remove it from the list",
       "remove-district": "Remove District"
     },
     "organization-export": {
-      "link": "Custom Exports",
-      "heading": "Customized Export",
+      "title": "Custom Exports",
```
bgsavage: Renamed link to title for consistency with other en.json feature groups where the link and heading are the same.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -527,18 +527,17 @@
...
       "form": {
         "organization": {
-          "label": "Search for a District, School or Institution",
-          "placeholder": "Start typing an institution name",
+          "label": "Search for a District, School or Organization",
```
bgsavage: I actually googled this because I thought you were supposed to have a comma after the second item (i.e. `District, School, or Organization` but it seems either is acceptable.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -527,18 +527,17 @@
...
+          "placeholder": "Start typing an organization name",
```
bgsavage: I couldn't find that this was being used, the placeholder for the input box now is just "Select".  @esotrope Can this be removed or are supposed to be using this one?

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: Would it be more accurate to say `Search for an Organization`?  Doesn't organization cover district and school as well?  Thinking out loud, maybe it technically covers it but the most common are distrcit and school so calling them out makes it clearer to the user? hmmm

bgsavage: Yeah actually I think "Search for an organization" makes more sense and is consistent with "Selected Organizations".

We also might be overthinking a piece of text that maybe 5% of users will actually read :)

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -525,18 +525,17 @@
       }
     },
     "organization-tree": {
-      "header": "Selected Institutions",
-      "subheader-empty": "No institutions selected",
+      "header": "Selected Organizations",
+      "subheader-empty": "No organizations selected",
```
esotrope: Organizations I think is the correct sbac term here but "Institution(s)" came from the mockups - was this change part of sbac feedback?

bgsavage: Yep, the feedback is that institution is perhaps too technical.

PR #599, opened 2017-11-15
some minor corrections to other items

dsnyder-fairway:


webapp/src/main/webapp/src/styles.less
```diff
@@ -106,6 +106,10 @@ h1 + p.pull-left, .h1 + p.pull-left, h2 + p.pull-left, .h2 + p.pull-left, h3 + p
 .nav-tabs > li.active > a, .nav-tabs > li.active > a:hover, .nav-tabs > li.active > a:focus {
   color: @gray-mid;
 }
+
+.btn.btn-borderless.btn-default, .scale-score-graph .bar-container .bar .bar-label, .table .scale-score .error-band {
+  color: @gray-mid;
+}
```
esotrope: 7 commits and 2 file changes. There must have been some git wrangling that took place here x J

For these styles did it not make sense to change the markup to have class `gray-XYZ`?

dsnyder-fairway: I think the commits might just be because this is an old branch with the commits previously squashed when merged.

Part of the problem before was that there are something like 7 layers of gray and for a good chunk (but not all) we want to use a single level of gray. I found introducing this gray easier.

PR #600, opened 2017-11-15
Fix reporting front-end cookie service.

wtritch:
This looks to be reported in the cookie service project as https://github.com/salemdar/angular2-cookie/issues/37

Since we don't need default CookieOptions, I decided to go with the factory provider solution.

Note that this was reproducible with meaningful error messages by changing `npm start` to run `ng serve --proxy-config proxy.conf.json --aot` with the added `--aot`

### DIFF
--- a/common-ngx/src/preference/rdw-preference.module.ts
+++ b/common-ngx/src/preference/rdw-preference.module.ts
@@ -8,9 +8,13 @@ import { CookieService } from "angular2-cookie/services/cookies.service";
     RdwCoreModule
   ],
   providers: [
-    CookieService,
-    UserPreferenceService
+    UserPreferenceService,
+    { provide: CookieService, useFactory: cookieServiceFactory }
   ]
 })
 export class RdwPreferenceModule {
+}
+
+export function cookieServiceFactory() {
+  return new CookieService();
 }


common-ngx/src/preference/rdw-preference.module.ts
```diff
@@ -8,9 +8,13 @@ import { CookieService } from "angular2-cookie/services/cookies.service";
     RdwCoreModule
   ],
   providers: [
-    CookieService,
-    UserPreferenceService
+    UserPreferenceService,
+    { provide: CookieService, useFactory: cookieServiceFactory }
```
esotrope: Thanks for cleaning this up Will.
Interested to know how the factory method is more AoT friendly.

wtritch: The problem is that CookieService has a constructor with an optional param:
```
constructor(_defaultOptions?: CookieOptions);
```
For some reason Angular's injection framework does the right thing when not running in AoT mode and when the provider isn't found it passes in a null.  When running in AoT mode the injection framework attempts to find a provider for a CookieOptions object and explodes when it can't find one.
Using the factory works around the AoT issue by providing an instantiated CookieService rather than telling the injection framework to instantiate one itself, so the framework never tries to find a CookieOptions provider.

esotrope: oh ok.
Maybe i should submit a PR to add `@Optional()` there for them

PR #601, opened 2017-11-15
Fix missing disableTransferAccess support in report-processor student…

wtritch:
… query

When implementing the transfer-access toggle ability for reports I missed one of the SQL filters.

### DIFF
--- a/report-processor/src/main/resources/application.sql.yml
+++ b/report-processor/src/main/resources/application.sql.yml
@@ -831,7 +831,7 @@ sql:
             (1 = :individual_statewide or s.district_group_id in (:individual_district_group_ids) or s.district_id in (:individual_district_ids) or s.school_group_id in (:individual_school_group_ids) or e.school_id in (:individual_school_ids))
             or
             (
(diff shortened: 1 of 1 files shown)


report-processor/src/main/resources/application.sql.yml
```diff
@@ -831,7 +831,7 @@ sql:
             (1 = :individual_statewide or s.district_group_id in (:individual_district_group_ids) or s.district_id in (:individual_district_ids) or s.school_group_id in (:individual_school_group_ids) or e.school_id in (:individual_school_ids))
             or
             (
-              1 = :allow_transfer_access
+              (1 = :allow_transfer_access and 0 = :disable_transfer_access)
```
malaffoon: oh that's not confusing at all   ;)

esotrope: Can we make both of these positive booleans?
`(1 = :system_transfer_access_allowed and 1 = :include_transfers)`
I'm assuming `disable_transfer_access` is from the student report form right?

PR #602, opened 2017-11-16
corrected pdf reports having 0 width scale score. Minimum value is no…

dsnyder-fairway:
…w 1.

Addresses 
reportscaleissue

Issue was caused by a scale score with a 0 width. Since this is used for display purposes only I believe we can have 1 as our minimum instead of 0

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
@@ -98,7 +98,7 @@ public class ReportViewSupport {
         final int length = maximumScore - minimumScore;
 
         // the whole percent of the scale score position along the length of the graph
-        final int position = getPercentRounded(scaleScore - minimumScore, length);
+        final int position = getPercentRoundedWithinBounds(scaleScore - minimumScore, length);
 
         final List<ScaleScoreView.Level> levels = newArrayList();
         for (int i = 1; i < cutPoints.size(); i++) {
@@ -107,7 +107,7 @@ public class ReportViewSupport {
             final int sumOfWidth = levels.stream().mapToInt(ScaleScoreView.Level::getWidth).sum();
 
             final int width = i < cutPoints.size() - 1
-                    ? getPercentRounded(current - previous, length)
+                    ? getPercentRoundedWithinBounds(current - previous, length)
                     // The last bar should be the remaining percent width of the graph
                     // This will avoid rounding errors and the graph width not adding to 100 percent
                     : 100 - sumOfWidth;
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -108,7 +106,7 @@ h1 + p.pull-left, .h1 + p.pull-left, h2 + p.pull-left, .h2 + p.pull-left, h3 + p
 }
 
 .btn.btn-borderless.btn-default, .scale-score-graph .bar-container .bar .bar-label, .table .scale-score .error-band {
-  color: @gray-mid;
+  color: darken(@gray-mid, 10%);
```
dsnyder-fairway: a small fix for the last remaining item on DWR-1098

dsnyder-fairway: a small fix for the last remaining item on DWR-1098

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
```diff
@@ -138,16 +138,16 @@ private float getPercent(final float position, final float length) {
...
      * If the given value is greater than the length <code>100</code> will be returned
      *
      * @param position the position along the length
      * @param length   the distance
-     * @return the percent distance covered by the given position along the given length in whole percent (0 - 100)
+     * @return the percent distance covered by the given position along the given length in whole percent (1 - 100)
      */
     private int getPercentRounded(final float position, final float length) {
-        return Math.round(getPercent(position, length));
+        return clamp(Math.round(getPercent(position, length)), 1, 100);
```
esotrope: I would move this out of this function into another function that is specific to correcting a percent of 0 to 1. Or add some method name mod.

Also, curious to know if this is specific to wkhtmltopdf rendering or not.
Note 1 is 1%, so this could slightly effect where the score point points on the graph.
I would test this case and make sure its ok:
```
asmt type: iab
cutpoints: [100, null, 400, null, 700]
stderr: 200
scores: 100-105 
```
and see how they scores line up and if they look colored right.

malaffoon: I realize i don't know anything about how these calls are being made but isn't a 0 a 0? Is this just bumping small fractional values up? So 1.2 -> 1 and 0.4 -> 1?

dsnyder-fairway: @esotrope This doesn't work correctly with such a high stderr (DWR-1100 is open to address that). Setting stderr to 100 works and looks fine. I'll opt to rename this as introducing a new method appears to introduce another issue with the score point if we use both methods

@malaffoon These are used to create a bar graph. When we have a width of 0 there is an issue where this is not rendered correctly (the width appears extremely small, but causes a break in the graph...see image above)

PR #603, opened 2017-11-17
apply correct clamps on all scale scores

dsnyder-fairway:
large stderr caused issues with displaying the number graphs. The issue was that a large enough stderr would eventually create a sumOfWidth = 100, which would result in the last calculated width being 0. Now, we ensure the max value for scale scores is at most 100 - number of separators (2).


dsnyder-fairway:
I'll reopen after cleaning this up a bit

PR #604, opened 2017-11-17
 apply correct clamps on all scale scores

dsnyder-fairway:
improved some documentation and extracted methods

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
@@ -69,7 +69,7 @@ public class ReportViewSupport {
     }
 
     /**
-     * Gets the position and width attributes necessary for rendering a scale score graph
+     * Gets the position and non-zero width attributes necessary for rendering a scale score graph
      *
      * @param exam       the exam to use for computing the scale score position
      * @param assessment the assessment to use for computing the cut point bar attributes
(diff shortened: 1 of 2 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
```diff
@@ -69,59 +69,97 @@ public String round(final float value) {
...
+     *
+     * @param cutPoints the cut points
+     * @param length    the maximumScore - minimumScore of the cut points
+     * @return the list of levels
+     */
+    private List<ScaleScoreView.Level> getScaleScoreViewLevels(final List<Integer> cutPoints, final int length) {
         final List<ScaleScoreView.Level> levels = newArrayList();
         for (int i = 1; i < cutPoints.size(); i++) {
-            final int previous = cutPoints.get(i - 1) - minimumScore;
-            final int current = cutPoints.get(i) - minimumScore;
```
esotrope: How does this work without the offsetting here?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
```diff
@@ -69,59 +69,97 @@ public String round(final float value) {
...
+    }
+
+    /**
+     * Gets the list of cut points for IAB assessments
+     *
+     * @param exam       the exam
+     * @param assessment the IAB assessment
+     * @return the list of cut point values
+     */
+    private List<Integer> getIabCutPoints(final Exam exam, final Assessment assessment) {
```
esotrope: 👍

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
dsnyder-fairway: we had this being used as `current - previous` = `cutPoint[i] - minimumScore - (cutPoint[i - 1] - minimumScore)` = `cutPoint[i] - cutPoint[i - 1]`

esotrope: Ah right - because we are just interested in the delta/width here. thanks!

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
```diff
@@ -69,59 +69,97 @@ public String round(final float value) {
...
      * @return the position and width attributes necessary for rendering a scale score graph
      */
     public ScaleScoreView getScaleScoreView(final Exam exam, final Assessment assessment) {
 
-        final int minimumScore = assessment.getCutPoints().get(0);
-        final int maximumScore = assessment.getCutPoints().get(assessment.getCutPoints().size() - 1);
-
         final List<Integer> cutPoints = assessment.getType() == AssessmentType.ICA
-                ? assessment.getCutPoints()
+                ? getIcaCutPoints(assessment)
```
jtreuting: i'm guessing we will need to do more to make this all work with Summative.  But the summative and ICA cutpoints are both driven from the the assessment the same way.

So i think this logic should check if it's IAB and get the IAB cut points from that method and otherwise have the method be getCutPoints and it would work for both ICA and Summative how it is now.

dsnyder-fairway: 👍 , I'll issue a follow up PR

PR #605, opened 2017-11-17
remove box-shadow for distractor analysis

dsnyder-fairway:


webapp/src/main/webapp/src/styles.less
```diff
@@ -148,6 +148,10 @@ h1 + p.pull-left, .h1 + p.pull-left, h2 + p.pull-left, .h2 + p.pull-left, h3 + p
   width: 100% !important;
 }
 
+distractor-analysis .table > tbody > tr > td {
```
dsnyder-fairway: bootstrap has a style for `.table > tbody > tr > td` which sets the box-shadow. Since I am not sure if this is desirable on other elements, I specified distractor-analysis to apply only to this view

PR #606, opened 2017-11-18
small correction to handle summative reports as well (in addition to ICA and IAB)

dsnyder-fairway:


dsnyder-fairway:
merging this -- I'll address any comments afterwards

PR #607, opened 2017-11-20
exclude summative assessments from appearing in results by item and d…

dsnyder-fairway:
…istractor analysis

### DIFF
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -560,7 +560,9 @@ sql:
           i.options_count,
           i.type
         from item i
-        where i.asmt_id=:assessment_id
+          join asmt a on i.asmt_id=a.id
+        where a.type_id != 3
+        and i.asmt_id=:assessment_id
         and (1=:includeAllItemTypes or i.type in (:types))
 
       findAllForExam: >-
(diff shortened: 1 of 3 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -560,7 +560,9 @@ sql:
           i.options_count,
           i.type
         from item i
-        where i.asmt_id=:assessment_id
+          join asmt a on i.asmt_id=a.id
+        where a.type_id != 3
```
dsnyder-fairway: type_id of 3 is a summative assessment. I'm open to either passing the summative assessment types to include/exclude from the methods that call this as well. @esotrope, thoughts?

esotrope: 🤔  An include-style query would be more future-proof if other assessment types are added

esotrope: Oh - If we are selecting an assessment by ID I don't think we need the type filter right?
I don' think there will be like an iab{id:1} and a summative{id:1}.

dsnyder-fairway: type is used for item type (MC, MS, etc)

dsnyder-fairway:
@jtreuting this change prevents viewing Summative test results via the Distractor Analysis or Results by Item

PR #608, opened 2017-11-21
UI changes for removing results by item / distractor analysis from summative assessments

dsnyder-fairway:


dsnyder-fairway:
Something didn't look right

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -87,8 +87,30 @@
...
+                   [ngClass]="{'selected': currentResultsView == state }">
+                  <button [disabled]="state.disabled"
+                          class="btn btn-default btn-borderless"
+                          (click)="setCurrentView(state)">
+                    {{ state.label | translate }}
+                  </button>
+                </a>
+                <a *ngIf="state.isResultsByItem()" class="dropdown-item"
+                   popover="{{ assessmentExam.assessment.isSummative ? 'messages.no-results-by-item-for-summative-exams' : 'messages.no-results-by-item' | translate }}"
+                   triggers="{{ !state.disabled ? '': 'mouseenter:mouseleave'}}"
```
jtreuting: if i'm reading this right, the popover text is always set but whether it is shown or not is triggered by the `state.disabled` flag.  So this disabled flag should be set based on whether this is summative or whether the year is too far in the past to where we don't have item level data.

i think i'm missing where that is being set?

dsnyder-fairway: You are right. I updated hasItemLevelData in assessment-results.component.ts to include the summative information. I should rename that method as well

jtreuting: ahh, so I missed that the code didn't nbeed to change cause it was driven by the existing `hasItemLevelData` method.  got it.

`hasItemLevelData` is sort of but not really accurate as you say because we might have it but not want to show it because it's summative.  instead of tying the name of the method to the specific item level reports (since once we add a third this name will miss one of them), how about something like:

`displayItemLevelData` or `shouldDisplayItemLevelData` so it's not whether we have it or not but whether it should be shown.

dsnyder-fairway: 👍 a sound solution -- I'll make the change

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -87,8 +87,30 @@
             </button>
             <ul *dropdownMenu class="dropdown-menu" role="menu">
               <li *ngFor="let state of viewStateOptions" role="menuitem">
```
wtritch: hm... if you're going to do an *ngIf check for each of the viewStateOptions, I think it might be cleaner to have them identified as separate <li> elements rather than looping over a collection and if-checking each one.

dsnyder-fairway: I kinda agree -- the reason why I left it as looping was just in case we add other exams (but if I went that route, I should have removed the results by student if statement for something like `if not resultsByItem and not distractorAnalysis`). Thoughts?

esotrope: sounds like a job for `Array.filter()`

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -194,7 +194,16 @@ export class AssessmentResultsComponent implements OnInit {
...
-      canExport: canExport
+      canExport: canExport,
+      isResultsByItem(): boolean {
+        return viewState == ResultsViewState.ByItem;
+      },
+      isResultsByStudent(): boolean {
+        return viewState == ResultsViewState.ByStudent;
+      },
+      isDistractorAnalysis(): boolean {
+        return viewState == ResultsViewState.DistractorAnalysis;
```
esotrope: does anyone know if enums in angular templates will ever be better supported? It's sad that we have to make these kind of methods i think.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
wtritch: Maybe I'm missing something.  I'm recommending ditching the `viewStateOptions: ResultsView[]` parameter in favor of 3 specific parameters:
```
resultsByStudent: ResultsView;
resultsByItem: ResultsView;
distractorAnalysis: ResultsView;
```
If we never act upon them as a collection except to filter them to a specific instance, why have them as a collection at all?

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
wtritch: This would go away with my above suggestion =)

esotrope: Then i like the suggesting : j 👍 👍

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
dsnyder-fairway: @wtritch changes have been made

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -86,16 +86,42 @@
...
+              <li role="menuitem">
+                <a class="dropdown-item"
+                   popover="{{ assessmentExam.assessment.isSummative ? 'messages.no-distractor-analysis-for-summative-exams' : 'messages.no-distractor-analysis' | translate }}"
+                   triggers="{{ !distractorAnalysisView.disabled ? '': 'mouseenter:mouseleave'}}"
+                   placement="right"
+                   [ngClass]="{'selected': currentResultsView == distractorAnalysisView }">
+                  <button [disabled]="distractorAnalysisView.disabled"
+                          class="btn btn-default btn-borderless"
+                          (click)="setCurrentView(distractorAnalysisView)">
+                    {{ distractorAnalysisView.label | translate }}
```
wtritch: 👍

PR #610, opened 2017-11-22
updated button text and permit for only specific users

dsnyder-fairway:


webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -12,7 +12,7 @@ <h1 class="blue-dark h1 mb-md">{{'labels.homepage.title' | translate}}</h1>
         <a *sbAuthorize="['INDIVIDUAL_PII_READ']" routerLink="/custom-export" class="btn btn-xs btn-default"><i class="fa fa-bold fa-table"></i> {{'labels.organization-export.title' | translate}}</a>
 
         <!-- Student Groups Admin Link -->
-        <a href="{{user.configuration.adminWebappUrl}}" class="btn btn-xs btn-info ml-xs"><i class="fa fa-gear"></i> {{'labels.groups.manage-group' | translate}}</a>
+        <a *sbAuthorize="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE']"href="{{user.configuration.adminWebappUrl}}" class="btn btn-xs btn-info ml-xs"><i class="fa fa-gear"></i> {{'labels.groups.admin-button' | translate}}</a>
```
dsnyder-fairway: didn't contain any authorizations. From reading DWR-1177 I take it this button should only display for users with these roles

admin-webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -13,7 +13,7 @@
   },
   "welcome": "Welcome {{firstName}} {{lastName}}",
   "application": {
-    "title": "Manage Student Groups"
+    "title": "Application Administration"
```
dsnyder-fairway: labels.group.title is "Manage Student Groups", but it looks like we may not want to change that from looking at the uses of that

webapp/src/main/webapp/src/app/home/home.component.html
esotrope: Weird, i could have sworn that had an auth check before 🤔 
Unrelated, but can we add a space before the `href` attribute on this line?

dsnyder-fairway: oops. Good 👁 . corrected

PR #611, opened 2017-11-22
Introduce reporting datasource in reporting admin

wtritch:
This PR introduces a second "reporting" datasource to the reporting admin application.

Existing repositories have been directed towards the "warehouse" datasource.

*Consequences:*
* All Authorization/Grants/Permissions are calculated against warehouse DB values, not reporting DB values.
* Injecting a NamedParameterTemplate, Datasource, etc now requires a qualifier to specify your target database.  This applies to both functional code and tests.
* Specifying \@Transactional now requires specifying either the warehouse or reporting transaction manager.  NO TRANSACTIONS SHALL CROSS DATABASES.  
* Just to be clear: NO TRANSACTIONS SHALL CROSS DATABASES.
* There is a bit of funkiness with running @Sql annotations in repository tests, you must now specify your datasource and transaction manager.


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/configuration/DataSourceConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+import javax.sql.DataSource;
+
+/**
+ * This configuration is responsible for providing datasources for both reporting and warehouse.
+ */
+@Configuration
+@EnableTransactionManagement
+public class DataSourceConfiguration {
```
wtritch: This is the new DataSourceConfiguration that stands up connections to both reporting and warehouse, provides NamedParameterTemplates, and provides transaction managers.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
```diff
@@ -36,7 +37,7 @@
     private String deleteSql;
 
     @Autowired
-    JdbcGroupRepository(final NamedParameterJdbcTemplate template,
+    JdbcGroupRepository(@Qualifier("warehouseJdbcTemplate") final NamedParameterJdbcTemplate template,
```
wtritch: Qualifiers everywhere.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcStudentGroupBatchRepository.java
```diff
@@ -40,12 +41,12 @@
...
 
     @Autowired
-    JdbcStudentGroupBatchRepository(final NamedParameterJdbcTemplate jdbcTemplate) {
+    JdbcStudentGroupBatchRepository(@Qualifier("warehouseJdbcTemplate") final NamedParameterJdbcTemplate jdbcTemplate) {
         this.jdbcTemplate = jdbcTemplate;
     }
 
     @Override
-    @Transactional
+    @Transactional(transactionManager = "warehouseTxManager")
```
wtritch: Transactional annotations need to specify the transaction manager.

admin-webapp/src/test/java/org/opentestsystem/rdw/admin/ApplicationIT.java
```diff
@@ -7,7 +7,7 @@
 import org.springframework.test.context.junit4.SpringRunner;
 
 @RunWith(SpringRunner.class)
-@ActiveProfiles("test")
+@ActiveProfiles("adminTest")
```
wtritch: rdw-common provides a "test" library jar that we're including for our tests.  Unfortunately, it also provides an "application-test.yml" file that creates a test spring.datasource, which muddies the testing waters quite a bit with a third semi-defined datasource.  I've broken free of the "test" profile and created an "adminTest" profile with associated "application-adminTest.yml" configuration file for separation.

admin-webapp/src/test/java/org/opentestsystem/rdw/admin/repository/RepositoryIT.java
```diff
@@ -26,13 +25,11 @@
 @Inherited
 @ContextConfiguration(classes = TestConfiguration.class)
 @SpringBootTest(classes = {
-        DataSourceAutoConfiguration.class,
```
wtritch: Manually including this actually breaks quite a bit since it tries to create an undefined "spring.datasource".

admin-webapp/src/test/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupImportRepositoryIT.java
```diff
@@ -19,14 +21,18 @@
 @RunWith(SpringRunner.class)
 @RepositoryIT
 @ContextConfiguration(classes = {JdbcGroupImportRepository.class})
-@Sql(scripts = {"classpath:integration-test-data.sql"})
-@ActiveProfiles("test")
+@Sql(
+        config = @SqlConfig(dataSource = "warehouseDatasource", transactionManager = "warehouseTxManager"),
+        scripts = {"classpath:warehouse-integration-test-data.sql"}
+)
+@Transactional(transactionManager = "warehouseTxManager")
```
wtritch: This is an example of scoping SQL and test transactions to a specific database.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
esotrope: It's that or `WarehouseJdbcTemplate extends NamedParameterJdbcTemplate`

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/configuration/DataSourceConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+        return new DataSourceTransactionManager(dataSource);
+    }
+
+    /**
+     * Data source for the data warehouse.
+     *
+     * @return DataSource
+     */
+    @Bean(name = "warehouseDatasource")
+    @ConfigurationProperties(prefix = "spring.warehouse_datasource")
```
esotrope: Is there any automagic happening with the spring prefix?
If not, i would think we should use `app.datasource.warehouse|reporting` or something

malaffoon: Just to weigh in ... when i was documenting the annotated configuration i found it, uhmm, unfortunate that we were inconsistent with the settings for the datasources. All the ingest apps use spring.datasource while the main reporting app uses app.datasource. I know foolish consistency is a hobgoblin but ...

esotrope: Yeah. Ideally we would not use it as anything `spring.*` should be channeling spring auto config. We could be legitimately using it in the other apps - I haven't checked. I think if you use `spring.datasource` it will configure the datasource for you without you needing to explicitly define a datasource bean.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
wtritch: hm... I think qualifiers feel cleaner than subclassing personally.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/configuration/DataSourceConfiguration.java
wtritch: That's a good point about the namespacing... but I'd prefer to err on the side of consistency and we can change the namespacing systematically if we want to in the future.  Thoughts?

esotrope: works for me 👍

PR #612, opened 2017-11-23
 adding instructional-resource to the admin page

dsnyder-fairway:
🦃 🦃 🦃 Enjoy your 🦃 🦃 🦃 

This is adding a new instructional-resource section to the admin page.
GROUP_WRITE now displays the following
screen shot 2017-11-23 at 1 45 17 pm

INSTRUCTIONAL_RESOURCE_WRITE now displays the following
screen shot 2017-11-23 at 1 45 51 pm


admin-webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -6,26 +6,32 @@
...
-                  <button type="button" dropdownToggle class="btn btn-link dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                    <i class="fa fa-user-circle"></i> <span *ngIf="user">{{'welcome' | translate:user }}</span> <span class="caret"></span>
+                <div class="btn-group btn-block" dropdown>
+                  <button type="button" dropdownToggle class="btn btn-link dropdown-toggle" data-toggle="dropdown"
+                          aria-haspopup="true" aria-expanded="false">
+                    <i class="fa fa-user-circle"></i> <span *ngIf="user">{{'welcome' | translate:user }}</span> <span
+                    class="caret"></span>
                   </button>
-                  <ul *dropdownMenu class="dropdown-menu" role="menu">
+                  <ul *dropdownMenu class="dropdown-menu mt-md" role="menu">
```
dsnyder-fairway: this is to correct an issue with the logout from the dropdown not being positioned correctly. I saw a ticket for this not in the sprint. I'll reference this soon

admin-webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -57,7 +63,8 @@
 <!-- Footer -->
 <footer class="mt-lg">
   <div class="container">
-    <p class="text-right black">&copy; The Regents of the University of California &ndash; Smarter Balanced Assessment Consortium</p>
+    <p class="text-right black">&copy; The Regents of the University of California &ndash; Smarter Balanced Assessment
```
esotrope: I know this won't be translated but we should drive this through translation so it is not static

admin-webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -6,26 +6,32 @@
         <!-- Navbar -->
         <nav class="navbar navbar-default">
           <div class="navbar-header">
-            <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#bs-example-navbar-collapse-1" aria-expanded="false">
+            <button type="button" class="navbar-toggle collapsed" data-toggle="collapse"
+                    data-target="#bs-example-navbar-collapse-1" aria-expanded="false">
```
esotrope: We should probably scrub the "example" values and give it our own name

admin-webapp/src/main/webapp/src/app/groups/group-filter-options.resolve.ts
```diff
@@ -1,15 +1,35 @@
...
+    return this._user;
+  }
+
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<GroupFilterOptions> | Promise<GroupFilterOptions> | GroupFilterOptions {
+    let userHasAccess: boolean;
+    this._userService.getCurrentUser().subscribe(user => {
+      userHasAccess = user.permissions.some(permission => AdminPermissions.indexOf(permission) !== -1);
+
+    });
+    if (userHasAccess) {
```
esotrope: the previous `subscribe` call is an async op - I don't think this will return the options. Also, i think we can have the permission check on the route itself - there should be a `[canActivate]` thingy

admin-webapp/src/main/webapp/src/app/groups/group-filter-options.resolve.ts
```diff
@@ -1,15 +1,35 @@
 import { Injectable } from "@angular/core";
-import { Resolve, ActivatedRouteSnapshot, RouterStateSnapshot } from "@angular/router";
+import { ActivatedRouteSnapshot, Resolve, RouterStateSnapshot } from "@angular/router";
 import { Observable } from "rxjs";
 import { GroupFilterOptions } from "./model/group-filter-options.model";
 import { GroupService } from "./groups.service";
+import { UserService } from "../user/user.service";
+import { User } from "../user/model/user.model";
+
+const AdminPermissions = [ 'GROUP_WRITE' ];
```
esotrope: should this also have instructional resources perm?

admin-webapp/src/main/webapp/src/app/app.component.html
dsnyder-fairway: DWR-1021

admin-webapp/src/main/webapp/src/app/groups/group-filter-options.resolve.ts
dsnyder-fairway: I think instructional resource is less permissive than group write. I was under the impression that instructional resource should only be able to see the new content. I could be totally off though

esotrope: Ok - we should call the const `GroupFilterOptionPermissions` or something more component specific then. also see my below comment on the impl

admin-webapp/src/main/webapp/src/app/groups/group-filter-options.resolve.ts
dsnyder-fairway: Still investigating how to use this with  `canActive`

esotrope: Example from reporting webapp:
```typescript
{
        path: 'schools/:schoolId',
        data: { breadcrumb: { resolve: 'school.name' }, permissions: [ 'INDIVIDUAL_PII_READ' ] },
        resolve: { school: CurrentSchoolResolve },
        canActivate: [ AuthorizeCanActivate ],
        ...
      },
```

The AuthorizeCanActivate then uses the data.permissions:
```typescript
@Injectable()
export class AuthorizeCanActivate implements CanActivate {

  constructor(private _service : UserService) {
  }

  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean>|Promise<boolean>|boolean {
    return this._service.doesCurrentUserHaveAtLeastOnePermission(route.data["permissions"]);
  }
}
```

dsnyder-fairway: yeah I've given that a shot but dwtest was unable to load all the data in `GroupFilterOptionsResolve` since somehow the `GroupFilterOptions` was undefined

admin-webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -95,6 +96,9 @@
       },
       "history": {
         "title": "Upload History"
```
malaffoon: Nothing to do with this PR, but i find it confusing to have two buttons beside each other, `Upload Groups` and `Upload History`. In one, `upload` is a verb, in the other it is a noun. Perhaps `Upload Groups` and `View Uploads`?

admin-webapp/src/main/webapp/src/app/groups/group-filter-options.resolve.ts
dsnyder-fairway: I had to copy some changes from #384 to get this working. Changes have been pushed

admin-webapp/src/main/webapp/src/assets/i18n/en.json
dsnyder-fairway: :+1: yeah I agree. It would be pretty neat if we supported `Upload History` as an action though. I'll change this

admin-webapp/src/main/webapp/src/app/groups/group-filter-options.resolve.ts
```diff
@@ -3,33 +3,15 @@ import { ActivatedRouteSnapshot, Resolve, RouterStateSnapshot } from "@angular/r
...
-      userHasAccess = user.permissions.some(permission => GroupFilterOptionPermissions.indexOf(permission) !== -1);
-
-    });
-    if (userHasAccess) {
-      return this.service.getFilterOptions();
+export class GroupFilterOptionsResolve implements Resolve<Resolution<GroupFilterOptions>> {
 
-    }
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Resolution<GroupFilterOptions>> {
+    return this.service.getFilterOptions().catch(Resolution.errorObservable);
```
esotrope: What is the case again where `getFilterOptions` fails?
The Resolution object was created for use in resolves but also specifically for when we want to distinguish 5xx from 404 on the view load. Just wondering if that is the case here or not.
Either way we should probably move the Resolution/NotFoundError into common-ngx

dsnyder-fairway: it fails for someone without GROUP_WRITE permissions (user with only INSTRUCTIONAL_RESOURCE_WRITE)

dsnyder-fairway: Removed `group-filter-options.resolve.ts`

admin-webapp/src/main/webapp/src/app/groups/groups.service.ts
```diff
@@ -19,11 +17,9 @@ export class GroupService {
   constructor(private dataService: DataService) {
   }
 
-  getFilterOptions(): Observable<Resolution<GroupFilterOptions>> {
+  getFilterOptions(): Observable<GroupFilterOptions> {
     return this.dataService
-      .get('/groups/filters')
-      .map(this.mapFilterOptionsFromApi.bind(this))
```
esotrope: think we might still need this part no?

dsnyder-fairway: oops. For some reason it works without this, but yes - it was unintentional

dsnyder-fairway:
theres an issue with changing schools/year/etc. I'll re-open once I fix this

dsnyder-fairway:
correction -- the issue is after changing schools, the upload / view upload buttons no longer work

dsnyder-fairway:
The upload button was loading links like 
> http://localhost:4200/groups;schoolId=13533;schoolYear=2018;subject=ALL/import
It was unclear to me if these buttons were supposed to be "Uploading Group" for a particular group, or "View History" of a particular group. However, in QA there only appears to be one school for "Group History" (now View History)... so it is a little unclear if viewing the history of uploads is for a particular school or not.

So, I am now just going to have these links go to addresses like http://localhost:4200/import 

Re-opening for feedback. If these buttons are general links not for a particular school this should be okay to merge.

attn @esotrope

esotrope:
@dsnyder-fairway this may be a complication with the recent app.routes.ts changes? not sure. Looks like we are attempting to preserve filters through navigation though

admin-webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -1,42 +1,58 @@
...
       {
-        path: 'groups',
-        pathMatch: 'prefix',
-        data: { breadcrumb: { translate: 'labels.groups.title' } },
+        path: '',
         children: [
           {
             path: '',
+            pathMatch: 'full',
+            component: HomeComponent
```
esotrope: I think this needs to be moved to the `path: 'home'` route object def.
I think you can remove this node and just move the `component: ...` part over. can you let me know if that addresses the URL issue?

dsnyder-fairway:
I'll open another PR when I have separated the admin landing page with links

PR #613, opened 2017-11-23
updates angular cli from 1.2.7 to 1.5.4 to close handlebars vulnerability

esotrope:


esotrope:
@wtritch @bgsavage @dsnyder-fairway 
I was able to successfully update angular cli to 1.5.4 however when running `rm -rf node_modules && npm install` which usually will recreate the `package-lock.json` I was unable to get NPM to recreate/update the file. Any suggestions?

admin-webapp/src/main/webapp/.angular-cli.json
```diff
@@ -6,7 +6,7 @@
   "apps": [
     {
       "root": "src",
-      "outDir": "../resources/static",
+      "outDir": "dist",
```
esotrope: angular cli no longer lets us build to somewhere outside of the project (to prevent overwriting system files is the intent)

admin-webapp/src/main/webapp/package.json
```diff
@@ -4,13 +4,14 @@
   "license": "MIT",
   "angular-cli": {},
   "scripts": {
-    "build": "ng build --prod",
+    "build": "ng build --prod && npm run move_build_to_static_resources",
```
esotrope: Workaround for not being able to build directly to `../static/resources`

build.gradle
```diff
@@ -258,7 +258,11 @@ task extractSchema(type: Sync) {
 def taskMap = ["migrateReporting": "Migrates the schemas that are required for running locally",
                "cleanReporting": "Cleans the schemas that are required for running locally",
                "migrateReporting_test": "Migrates the schemas that integration tests depend on",
-               "cleanReporting_test": "Cleans the schemas that integration tests depend on"]
+               "cleanReporting_test": "Cleans the schemas that integration tests depend on",
```
esotrope: added migrate warehouse tasks since admin is in this project now.

admin-webapp/src/main/webapp/package.json
```diff
@@ -4,13 +4,14 @@
...
     "start": "ng serve --proxy-config proxy.conf.json",
     "lint": "tslint \"src/**/*.ts\"",
     "test": "ng test",
     "test-ci": "ng test -sr --reporters teamcity",
     "pree2e": "webdriver-manager update",
-    "e2e": "protractor"
+    "e2e": "protractor",
+    "move_build_to_static_resources": "rimraf ../resources/static && copyfiles -u 1 dist/* dist/assets/**/* ../resources/static"
```
esotrope: not sure why copyfiles didn't copy all files when giving it `dist/*`, `dist/**` or `dist/**/*` I had to give it the assets dir explicitly

dsnyder-fairway:
:man_shrugging: maybe `npm cache clean` between removing the node modules and npm install

admin-webapp/src/main/webapp/package.json
bgsavage: I would think this would be more of the full web application / gradle concern rather than something that the npm build does.

esotrope: true. i'll take a shot at doing this in gradle 👍

esotrope:
Opening another PR for this

PR #614, opened 2017-11-24
Task/update angular cli to 1.5.4

esotrope:


admin-webapp/build.gradle
```diff
@@ -84,7 +86,10 @@ task buildFe(type: NpmTask) {
     args = ['run', 'build']
 }
 
-npm_install.dependsOn(deleteProjectDependencies)
+task moveFeBuild(dependsOn: buildFe) {
+    copy { from "src/main/webapp/dist" into "src/main/resources/static" }
+    delete "src/main/webapp/dist"
+}
```
esotrope: @bgsavage gradle now moves build from dist to static resources dir

bgsavage: Hm, I think I would have `moveFeBuild` depend on `NpmTask` and then have `buildFe` depend on `moveFeBuild`.  Otherwise I don't think this will get called if you just run `gradle buildFe`, though it does still get called if you run `gradle build`.  So maybe it doesn't matter?

esotrope: actually what i should do is move this into buildFe as a `doLast` i think

webapp/build.gradle
```diff
@@ -74,10 +76,13 @@ task buildFe(type: NpmTask) {
     outputs.dir('build/resources/static')
     dependsOn npm_install
     args = ['run', 'build']
+    // move build from dist to static resources
+    doLast {
+        copy { from "src/main/webapp/dist" into "src/main/resources/static" }
+        delete "src/main/webapp/dist"
+    }
```
esotrope: changed to doLast so it is done with the build task and not after it

bgsavage: 👍

PR #615, opened 2017-11-24
updates node and npm. adds package-lock.json back

esotrope:
Updates node and npm to latest stable release.
It's a good idea to update node/npm on your system to the same versions as well
```bash
brew upgrade node
npm install -g npm@5.5.1
```
I think the package-lock.json were not being created because we started using node 4.5.0
An update to 5.5.1 seemed to do the trick.

### DIFF
--- a/admin-webapp/build.gradle
+++ b/admin-webapp/build.gradle
@@ -51,8 +51,8 @@ dependencies {
 }
 
 node {
-    version = '7.10.0'
-    npmVersion = '4.5.0'
+    version = '8.9.1'
+    npmVersion = '5.5.1'
     download = true
     workDir = file("${project.projectDir}/src/main/webapp")
     nodeModulesDir = file("${project.projectDir}/src/main/webapp")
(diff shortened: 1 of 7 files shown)


admin-webapp/src/main/webapp/package.json
```diff
@@ -30,6 +30,7 @@
     "@types/file-saver": "0.0.1",
     "@types/lodash": "^4.14.68",
     "angulartics2": "^2.2.2",
+    "bootstrap": "^3.3.7",
```
esotrope: needed an explicit dependency installation for our `style.less` to reference stylesheets in the `node_modules/bootstrap/less` dir

esotrope:
Any aversions to merging this?

admin-webapp/build.gradle
```diff
@@ -51,8 +51,8 @@ dependencies {
 }
 
 node {
-    version = '7.10.0'
-    npmVersion = '4.5.0'
+    version = '8.9.1'
+    npmVersion = '5.5.1'
```
wtritch: Might want to put out a note that devs should update their node and npm versions?

wtritch:
Thank you for doing this!

admin-webapp/build.gradle
esotrope: it's in the description but I'll echo it here:
It's a good idea to update node/npm on your system to the same versions as well
```bash
brew upgrade node
npm install -g npm@5.5.1
```

PR #620, opened 2017-11-27
 adding instructional-resource to the admin page

dsnyder-fairway:
new pull request. 5f3ae89 and 14e725b are the latest changes. 
These changes : 

1. correct the import and history links that were problematic in the previous PR. 

2. move these tables to appear in links that must be navigated to from the landing page

GROUP_WRITE shows
screen shot 2017-11-27 at 12 55 25 pm

INSTRUCTIONAL_RESOURCE_WRITE shows
screen shot 2017-11-27 at 12 54 57 pm


admin-webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -1,87 +1,132 @@
...
-            path: 'import',
+            path: 'groups',
             pathMatch: 'prefix',
-            data: { breadcrumb: { translate: 'labels.groups.import.title' } },
+            data: {
+              breadcrumb: {
+                translate: 'labels.groups.title',
+              }
+            },
             children: [
```
dsnyder-fairway: @esotrope I think this is what was causing the import / history buttons to not work before. I removed these as children of groups in my previous PR

wtritch:
Just to be clear, you show both links when the user has GROUP_WRITE *and* INSTRUCTIONAL_RESOURCE_WRITE, correct?  GROUP_WRITE alone doesn’t grant access to instructional resource management?

dsnyder-fairway:
Hmmm...let me verify

dsnyder-fairway:
@wtritch I believe so. dwtest also has INTRUSTIONAL_RESOURCE_WRITE permissions. Looking at app.routes.ts shows we allow for GROUP_WRITE _and_ INSTURCTIONAL_RESOURCE_WRITE to view the landing page, but the links and access to those links depend on the individual permissions

aaronsleeper:
now i remember. this is a convo @wtritch and i had. this does need to be a page like this because it allows us to extend user level application settings control in the future. in the dropdown user menu there should just be a link to settings or preferences\

PR #621, opened 2017-11-27
Add a test-mode flag to control conditionally loading configurations …

wtritch:
…while running tests

After having just spent time trying to figure this out I would suggest that we use profiles to control properties which can act as flags rather than as flags themselves.

### DIFF
--- a/webapp/src/test/resources/application-test.yml
+++ b/webapp/src/test/resources/application-test.yml
@@ -21,3 +21,5 @@ app:
   iris:
     vendorId: vendorId
     url: http://my-iris.somewhere.org/iris/
+
+  test-mode: true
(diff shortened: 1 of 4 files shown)


webapp/src/test/resources/application-test.yml
```diff
@@ -21,3 +21,5 @@ app:
   iris:
     vendorId: vendorId
     url: http://my-iris.somewhere.org/iris/
+
+  test-mode: true
```
esotrope: actually now that i think about it doesnt this just mirror spring active profiles : test?

wtritch: Yes, but it's a property that's activated by the profile, rather than testing which profile is active.  If we want a second profile with different properties for a different test we'd have to remember to go add that new profile to the list of excluded profiles.  (And we won't.)  Using properties, we could see from application-test.yml or application-adminTest.yml exactly what's being controlled and the profile just becomes a yml-activator rather than an additional configuration itself.

esotrope: 👍  just wondering... why doesn't saml work in tests again?

wtritch: I'm not sure of the back-story, I'm afraid I didn't work on this part.  However, we aren't providing information like "${rdw.admin.idp}" in our test application.yml.  I'm not sure if the SAML library initialization actually attempts to make contact with the SSO system, but either it does or we didn't want to add 6 saml properties into our test config... ?

PR #624, opened 2017-11-29
Feature/dwr 1055 Dynamic SQL POC

agorina:
I do not like this solution, but I cannot think of a better one. Open for suggestions.

I suggest to start the review from `sample.sql.yml`:
https://github.com/SmarterApp/RDW_Reporting/pull/624/files#diff-cf69b859c000ed30a85c87e3f1108bc4

I am using this third party solution: http://openhms.sourceforge.net/sqlbuilder/example.html

It could be used in two ways: one is to define a schema object with all the table specs, and another is to use it as a fancy String builder (aka "custom SQL"). I am using the later in this review.

I have been asked by some people and question it myself if using this library justifiable for what we are doing. Here is where I see a 'value added':

1. it supports constructing SQL by building clauses in any order, and it then assembles them into one SQL string.
2. it takes care of commas and parentheses.
3. **IFF** we decide to use RSQL, it could be helpful.

NOTE: while this could be used for the current reporting queries, this was not the intend. The main goal it to support dynamic queries for the custom aggregate reports. If we decide to keep it and use for the current reporting SQLs, we could/should enhance/simplify it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,7 +62,6 @@ subprojects {
     repositories {
         mavenLocal()
         mavenCentral()
-        maven { url 'https://mvnrepository.com/artifact/com.healthmarketscience.sqlbuilder/sqlbuilder' }
     }
 
     dependencyManagement {
--- a/build.gradle
+++ b/build.gradle
@@ -62,6 +62,7 @@ subprojects {
     repositories {
         mavenLocal()
         mavenCentral()
+        maven { url 'https://mvnrepository.com/artifact/com.healthmarketscience.sqlbuilder/sqlbuilder' }
     }
 
     dependencyManagement {
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -17,7 +17,7 @@ dependencies {
     compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-properties'
     compile 'org.springframework.boot:spring-boot-starter-jdbc'
     compile 'org.springframework.boot:spring-boot-starter-thymeleaf'
-    compile group: 'com.healthmarketscience.sqlbuilder', name: 'sqlbuilder', version: '2.1.7'
+    compile 'com.healthmarketscience.sqlbuilder:sqlbuilder:2.1.7'
 
     optional 'org.springframework.boot:spring-boot-configuration-processor'
 
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -17,6 +17,8 @@ dependencies {
     compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-properties'
     compile 'org.springframework.boot:spring-boot-starter-jdbc'
     compile 'org.springframework.boot:spring-boot-starter-thymeleaf'
+    compile group: 'com.healthmarketscience.sqlbuilder', name: 'sqlbuilder', version: '2.1.7'
+
     optional 'org.springframework.boot:spring-boot-configuration-processor'
 
     testCompile('org.springframework.boot:spring-boot-starter-test') {
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/QueryProvider.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/QueryProvider.java
@@ -1,0 +1,94 @@
+package org.opentestsystem.rdw.reporting.common.sqlbuilder;
+
+import com.healthmarketscience.sqlbuilder.CustomCondition;
+import com.healthmarketscience.sqlbuilder.CustomSql;
+import com.healthmarketscience.sqlbuilder.SelectQuery;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * This class is responsible for constructing sql queries from {@link SqlTemplate} templates
+ */
+@Component
+public class QueryProvider {
+
+    private final SqlBuilderConfig sqlBuilderConfig;
+    private final String SelectClause = "select";
+    private final String JoinClause = "join";
+    private final String WhereClause = "conditions";
+    private final String GroupByClause = "grouping";
+    final String[] clauseNames = {SelectClause, JoinClause, WhereClause, GroupByClause};
+
+    /**
+     * Constructor
+     *
+     * @param sqlBuilderConfig the {@link SqlBuilderConfig}
+     */
+    @Autowired
+    public QueryProvider(final SqlBuilderConfig sqlBuilderConfig) {
+        this.sqlBuilderConfig = sqlBuilderConfig;
+    }
+
+    /**
+     * Returns a {@link SelectQuery} constructed by combing the {@link SqlTemplate} and applying its{@link SqlTemplate#addons}
+     *
+     * @param templateName the {@link SqlTemplate}'s name
+     * @param addons       the {@link SqlTemplate#addons}'s names
+     * @return constructed {@link SelectQuery}
+     */
+    public SelectQuery newQuery(final String templateName, final Iterable<String> addons) {
+        checkArgument(sqlBuilderConfig.getTemplates() != null && sqlBuilderConfig.getTemplates().containsKey(templateName), "unknown sql builder template name " + templateName);
+
+
+        final SqlTemplate template = sqlBuilderConfig.getTemplates().get(templateName);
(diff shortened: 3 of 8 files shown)


build.gradle
```diff
@@ -62,6 +62,7 @@ subprojects {
     repositories {
         mavenLocal()
         mavenCentral()
+        maven { url 'https://mvnrepository.com/artifact/com.healthmarketscience.sqlbuilder/sqlbuilder' }
     }
```
agorina: If you want to learn more about this package: http://openhms.sourceforge.net/sqlbuilder/example.html

common/src/main/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/QueryProvider.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Returns a {@link SelectQuery} constructed by combing the {@link SqlTemplate} and applying its{@link SqlTemplate#addons}
+     *
+     * @param templateName the {@link SqlTemplate}'s name
+     * @param addons       the {@link SqlTemplate#addons}'s names
+     * @return constructed {@link SelectQuery}
+     */
+    public SelectQuery newQuery(final String templateName, final Iterable<String> addons) {
```
agorina: We may need this if we decide to RSQL (REST Query Language) http://www.baeldung.com/rest-api-search-language-rsql-fiql.

I am including it just for the demonstration purpose. 

Please see the IT for an example.

common/src/test/resources/sample.sql.yml
```diff
@@ -0,0 +1,58 @@
...
+            state:
+              clauses:
+                select:  >-
+                   'state' AS type
+
+      # ------------ achievement level - all districts  -------------------------------------------------------------------
+            allDistricts:
+              clauses:
+                select:  >-
+                      sch.district_id,
```
esotrope: I'm guessing the library takes care of adding a comma to the previous line when appending?

common/src/test/resources/sample.sql.yml
```diff
@@ -0,0 +1,58 @@
...
+               fe.school_year,
+               a.subject_id
+
+          from: >-
+              fact_student_ica_exam fe
+
+          join: >-
+              JOIN ica_asmt a ON a.id = fe.asmt_id
+
+          conditions: >-
```
esotrope: i'd be tempted to call conditions "where" and grouping "groupBy" since the others are following the sql syntax

build.gradle
esotrope: Is the artifact really not at maven central? that's rare.

common/src/test/resources/sample.sql.yml
agorina: I was debating this too. The naming here matches the names in the third party package, but I am open to change it.

common/src/test/resources/sample.sql.yml
agorina: Yes, that is correct.

build.gradle
agorina: You are correct, thank you. Removing this reference.

common/build.gradle
```diff
@@ -17,6 +17,8 @@ dependencies {
     compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-properties'
     compile 'org.springframework.boot:spring-boot-starter-jdbc'
     compile 'org.springframework.boot:spring-boot-starter-thymeleaf'
+    compile group: 'com.healthmarketscience.sqlbuilder', name: 'sqlbuilder', version: '2.1.7'
```
malaffoon: if you commit this, please use a consistent format here, `com.healthmarketscience.sqlbuilder:sqlbuilder:2.1.7`

common/src/test/resources/sample.sql.yml
malaffoon: i agree with Landon, the SQL terminology would be clearer

common/src/main/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/QueryProvider.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+@Component
+public class QueryProvider {
+
+    private final SqlBuilderConfig sqlBuilderConfig;
+    private final String SelectClause = "select";
+    private final String JoinClause = "join";
+    private final String WhereClause = "conditions";
+    private final String GroupByClause = "grouping";
+    final String[] clauseNames = {SelectClause, JoinClause, WhereClause, GroupByClause};
```
malaffoon: all the clause names and this collection can/should be static

common/src/main/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/QueryProvider.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param sqlBuilderConfig the {@link SqlBuilderConfig}
+     */
+    @Autowired
+    public QueryProvider(final SqlBuilderConfig sqlBuilderConfig) {
+        this.sqlBuilderConfig = sqlBuilderConfig;
+    }
+
+    /**
+     * Returns a {@link SelectQuery} constructed by combing the {@link SqlTemplate} and applying its{@link SqlTemplate#addons}
```
malaffoon: `combining`

common/src/test/resources/sample.sql.yml
agorina: Will change.

common/build.gradle
agorina: Will change.

common/src/test/resources/sample.sql.yml
```diff
@@ -0,0 +1,58 @@
...
+               sum(CASE WHEN performance_level = 4 THEN 1 ELSE 0 END) AS level4,
+               fe.asmt_grade_id,
+               fe.school_year,
+               a.subject_id
+
+          from: >-
+              fact_student_ica_exam fe
+
+          join: >-
+              JOIN ica_asmt a ON a.id = fe.asmt_id
```
jtreuting: if i'm reading this right (which i'm guessing i'm not), this JOIN is separated so that it can be optionally included?  If that is the case and this JOIN is not included then the `a.subject_id` from the select will blow up.  Am I understanding this right?

agorina: Let me try to clarify it:

There is a main template and 'add ons'.
A main template is broken into separate clauses, but they all get assembled into one SQL. The reason it is broken is because it is easier to add 'add on's using the third party lib. 

Each 'add on' is also broken into clauses, but if 'add on' is included all clauses from that 'add on' are included. 

The idea is that the main template has the min core SQL scaffolding. 'Add on's could be added to modify the SQL.

jtreuting: got it.  makes sense

chdietz:
You say this library can be used in two ways and this example is the latter 'fancy String builder'.
Did you try the  schema object with all the table specs way? And if yes, why was it rejected.

agorina:
No, I did not like the idea of re-defining the schema. If you look at it, you really have to define each column with its data type and so on. Seems like a maintenance nightmare to me.

common/src/test/resources/sample.sql.yml
```diff
@@ -0,0 +1,58 @@
+sqlbuilder:
+  templates:
+      # ------------ achievement level report t  -----------------------------------------------------------------------------
```
chdietz: Are we moving away from the original intent of having sql in yml?  These are not complete queries but fragments.  If we are constructing queries in Java why not have all the components in Java. I suspect you considered this?

common/src/test/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/QueryProviderTestIT.java
```diff
@@ -0,0 +1,122 @@
...
+@Import(ConfigurationPropertiesAutoConfiguration.class)
+@ActiveProfiles("test")
+public class QueryProviderTestIT {
+    @Autowired
+    private QueryProvider provider;
+
+    @Test
+    public void testQueryAsString() {
+
+        String query = provider.newQueryAsString("achievementLevels", newArrayList());
```
chdietz: One of the things I like about these tests is it provides a place a developer can quickly see a fully formed sample query.

common/src/test/resources/sample.sql.yml
agorina: We did, and for the custom SQL this is the best compromise I could come up with.

PR #625, opened 2017-11-29
Feature/introduce common organization repository

wtritch:
While working on the admin instructional resources API I found a need to validate that user has access to a particular organization (state, district_group, district, or school_group) and the front-end will need an auto-completion by name to an organization for any of those organization levels, much like our custom exports organization input: https://awsqa-reporting.sbacdw.org/custom-export 
However, I don't want to send all organizations to the front-end.

While building out an OrganizationRepository I found that we already had a number of Organization and OrganizationRepository implementations.  This PR introduces a consolidated *reporting* OrganizationRepository into reporting common for use across both our reporting and admin webapps.


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/Organization.java
```diff
@@ -1,39 +0,0 @@
-package org.opentestsystem.rdw.admin.model;
-
-/**
- * Represents a district or school responsible for the student who took an exam
- */
-public class Organization {
```
wtritch: -1 Organization implementation.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupService.java
```diff
@@ -37,7 +39,7 @@
 
     @Override
     public Set<Group> getByQuery(final User user, final GroupQuery query) {
-        return ImmutableSet.copyOf(repository.findAllByQuery(user.getPermissionsById().get("GROUP_WRITE").getScope(), query));
+        return ImmutableSet.copyOf(repository.findAllByQuery(user.getPermissionsById().get(GroupWrite).getScope(), query));
```
wtritch: Unrelated, I prefer constants over heavily-reused strings.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
```diff
@@ -1,46 +1,41 @@
...
+
 /**
  * Represents a school district
  */
 public class District extends Organization {
-    private Long districtGroupId;
 
-    District(final long id, final String name) {
-        super(id, name);
-    }
```
wtritch: I dropped the private constructors in favor of an extended builder pattern.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Organization.java
```diff
@@ -5,26 +5,82 @@
...
-    private final String name;
-
-    public Organization(final long id, final String name) {
-        this.id = id;
-        this.name = name;
-    }
+    private long id;
+    private String name;
+    private String naturalId;
+    private OrganizationType organizationType;
```
wtritch: Since I'm now returning Organizations of multiple types I've added in a type enum here, so that a response payload can contain a mixture of DistrictGroups, Districts, Schools, etc.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
```diff
@@ -0,0 +1,67 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import java.util.EnumSet;
+
+/**
+ * This model represents a query for organizations.
+ */
+public class OrganizationQuery {
```
wtritch: I introduced a query object for searching across multiple organization levels.  By default I'm limiting results to 1k, but by pushing this down to the Controller layer we can allow the client to specify how many results it would like.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Reporting JDBC-backed implementation of an OrganizationRepository.
+ */
+public class ReportingJdbcOrganizationRepository implements OrganizationRepository {
```
wtritch: This implementation and it's queries are targeted towards the "reporting" schema.  We could also create a warehouse implementation if there's a need.  Note that I don't have an \@Repository annotation here since I'm not sure how a consumer is going to provide the template or security parameter provider.  A consumer will need to register this instance as a \@Bean for DI.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
```diff
@@ -0,0 +1,119 @@
...
+                        .id(resultSet.getLong("id"))
+                        .name(resultSet.getString("name"))
+                        .naturalId(resultSet.getString("natural_id"))
+                        .districtGroupId(resultSet.getLong("district_group_id"))
+                        .districtId(resultSet.getLong("district_id"))
+                        .schoolGroupId(resultSet.getLong("school_group_id"))
+                        .build();
+            default:
+                throw new IllegalStateException("Unknown type: " + type);
+        }
```
wtritch: I'm not a huge fan of the "case" statement, but since this repository is bound to SQL that only queries for specific organization types it is already tightly-coupled to this information... thoughts?

common/src/main/resources/common.sql.yml
```diff
@@ -13,3 +13,113 @@ sql:
...
+          school.natural_id,
+          'School' as org_type,
+          school.district_group_id,
+          school.district_id,
+          school.school_group_id
+        FROM school
+        WHERE 1=:allow_school
+          AND (:name_query IS NULL OR school.name LIKE concat('%', :name_query, '%'))
+          AND (1=:statewide OR school.district_group_id IN (:district_group_ids) OR school.district_id IN (:district_ids) OR school.school_group_id IN (:school_group_ids) OR school.id IN (:school_ids))
+        LIMIT :limit
```
wtritch: @agorina I know UNION is a bad word, but polymorphic cross-table SQL is a PITA.  I'm open to suggestions.

common/src/main/resources/common.sql.yml
```diff
@@ -13,3 +13,113 @@ sql:
...
+            AND s.id = sch.id
+        WHERE
+          (dg.id IS NOT NULL AND (1=:statewide OR sch.district_group_id IN (:district_group_ids)))
+          OR
+          (d.id IS NOT NULL AND (1=:statewide OR sch.district_group_id IN (:district_group_ids) OR sch.district_id IN (:district_ids)))
+          OR
+          (sg.id IS NOT NULL AND (1=:statewide OR sch.district_group_id IN (:district_group_ids) OR sch.district_id IN (:district_ids) OR sch.school_group_id IN (:school_group_ids)))
+          OR
+          (s.id IS NOT NULL AND (1=:statewide OR sch.district_group_id IN (:district_group_ids) OR sch.district_id IN (:district_ids) OR sch.school_group_id IN (:school_group_ids) OR sch.id IN (:school_ids)))
+        LIMIT 1
```
wtritch: I originally tried something like this for the findAll query, but ran into issues getting the "sch" school table to vary across organization tables.  (If there were district results, then I wouldn't see school results.)

common/src/test/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepositoryIT.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Configuration
+    public static class ReportingJdbcOrganizationRepositoryITConfiguration {
+
+        @Bean
+        public ReportingJdbcOrganizationRepository repository(final NamedParameterJdbcTemplate template,
+                                                              final SecurityParameterProvider securityParameterProvider) {
+            return new ReportingJdbcOrganizationRepository(template, securityParameterProvider);
+        }
+    }
```
wtritch: Example of registering the repository as a bean to allow the \@Value annotations to properly inject the SQL statements.

common/src/test/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepositoryIT.java
```diff
@@ -0,0 +1,261 @@
...
+                .isNull();
+    }
+
+    @Test
+    public void itShouldFindASchoolById() {
+        assertThat(repository.findOneById(statewide(), School, -101).getName())
+                .isEqualTo("school1");
+        assertThat(repository.findOneById(schools(-102L), School, -101))
+                .isNull();
+    }
```
wtritch: Large number of tests to validate all the use-cases I could think of.

webapp/src/main/java/org/opentestsystem/rdw/reporting/ApplicationConfiguration.java
```diff
@@ -39,5 +43,10 @@
 @EnableConfigurationProperties(ApplicationProperties.class)
 public class ApplicationConfiguration {
 
+    @Bean
+    public OrganizationRepository repository(final NamedParameterJdbcTemplate template,
+                                             final SecurityParameterProvider securityParameterProvider) {
+        return new ReportingJdbcOrganizationRepository(template, securityParameterProvider);
+    }
```
wtritch: I'm registering the repository with the reporting webapp.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
```diff
@@ -35,25 +43,57 @@ public DefaultOrganizationService(@NotNull final OrganizationRepository reposito
...
-        return ImmutableSet.copyOf(repository.findAllSchools(permission.getScope()));
+        final OrganizationQuery schoolQuery = OrganizationQuery.builder()
+                .types(EnumSet.of(OrganizationType.School))
+                .limit(Unlimited)
+                .build();
+
+        final Collection<Organization> organizations = repository.findAll(permission.getScope(), schoolQuery);
+        return ImmutableSet.copyOf(organizations.stream()
+                .map(School.class::cast)
+                .collect(Collectors.toSet()));
```
wtritch: I modified the existing webapp repository service to use the common repository.  Ideally we can now change the API to query on-demand across organizations rather than front-load all the user's districts, schools, etc. That felt pretty far out-of-scope, however.

webapp/src/main/java/org/opentestsystem/rdw/reporting/organization/JdbcOrganizationRepository.java
```diff
@@ -1,98 +0,0 @@
...
-import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
-import org.springframework.stereotype.Repository;
-
-import javax.validation.constraints.NotNull;
-import java.util.List;
-
-import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
-
-@Repository()
-class JdbcOrganizationRepository implements OrganizationRepository {
```
wtritch: -1 OrganizationRepository implementation

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcSchoolRepository.java
```diff
@@ -34,11 +34,11 @@
         return template.query(
                 findAllQuery,
                 securityParameterProvider.getSecurityParameters(permissionScope),
-                (row, index) -> new Organization(
-                        row.getLong("id"),
-                        row.getString("name"),
-                        row.getString("natural_id")
-                )
+                (row, index) -> new Organization.Builder()
```
malaffoon: i prefer the convention of `Organization.builder()` (static method) over having to use `new`

wtritch: That doesn't work with extending =/  So `public static School.builder()`  won't fly because it changes the static method signature.  And if I can't be consistent, I didn't want to put it on Organization and assume people wouldn't call it via School.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/security/AdminPermission.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.admin.security;
+
+/**
+ * Well-known permissions in the Admin application.
+ */
+public class AdminPermission {
```
malaffoon: Should this follow the pattern of other utility classes? I think we lean towards just hiding the default constructor but there is also the idea of making the class final.

wtritch: I'm not sure exactly what good it does, but I'll follow whatever recommendation you make.  I'll also make the change to ReportingPermission that this class is based off of.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
```diff
@@ -1,46 +1,41 @@
...
+    private Long districtGroupId;
```
malaffoon: I wonder if this could be generalized in the Organization class itself as parentId.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
```diff
@@ -1,46 +1,41 @@
...
 
-    public static class Builder {
-        private long id;
-        private String name;
         private Long districtGroupId;
 
-        public Builder id(final long id) {
-            this.id = id;
-            return this;
+        public Builder() {
```
malaffoon: As noted elsewhere i prefer the builder constructor to be hidden as well and have a static method on the buildee class.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/OrganizationRepository.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Find an organization of the given type by id.
+     *
+     * @param permissionScope   A permission scope
+     * @param type              The organization type
+     * @param id                The organization id
+     * @return The organization or null
+     */
+    Organization findOneById(PermissionScope permissionScope, OrganizationType type, int id);
```
malaffoon: nit: findOneByTypeAndId?

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
malaffoon: I think the comment about no stereotype annotation should be in the javadocs for the class.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
wtritch: hm... I'm debating that, because it also exists on SchoolGroup and School.  (The entire ancestry is included rather than just the immediate parent.)

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/OrganizationRepository.java
wtritch: good call, changed

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
wtritch: See above =)

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
```diff
@@ -0,0 +1,119 @@
...
+                        .id(resultSet.getLong("id"))
+                        .name(resultSet.getString("name"))
+                        .naturalId(resultSet.getString("natural_id"))
+                        .build();
+            case District:
+                return new District.Builder()
+                        .id(resultSet.getLong("id"))
+                        .name(resultSet.getString("name"))
+                        .naturalId(resultSet.getString("natural_id"))
+                        .districtGroupId(resultSet.getLong("district_group_id"))
```
malaffoon: Did you test this? i believe resultSet.getLong() returns 0 for null values and i think you don't want that. I know i had to add a helper method like getLongOrNull(ResultSet rs, String field) that calls that resultSet was-null method (sorry, too lazy to look up exact method).

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
wtritch: good call, done

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
malaffoon: Meh.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
wtritch: hmm... good point.  We don't want to return 0's if we don't have to.  I'll add some null protections in here.

agorina: org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable - I just happened to have it open :)

common/src/main/resources/common.sql.yml
```diff
@@ -13,3 +13,113 @@ sql:
...
+          null as district_group_id,
+          null as district_id,
+          null as school_group_id
+        FROM district_group
+          JOIN school sch
+            ON sch.district_group_id = district_group.id
+        WHERE 1=:allow_district_group
+          AND (:name_query IS NULL OR district_group.name LIKE concat('%', :name_query, '%'))
+          AND (1=:statewide OR sch.district_group_id IN (:district_group_ids))
+        UNION
```
malaffoon: does MySQL have the same `UNION ALL` is better thing that postgres does?

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
wtritch: Done, and added a validating test.

common/src/main/resources/common.sql.yml
wtritch: UNION ALL apparently just removes the implicit DISTINCT (which I don't want)

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/security/AdminPermission.java
malaffoon: I guess i'd vote to at least hide the default constructor.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcSchoolRepository.java
malaffoon: Oh. Organization isn't abstract? My example was bad (of course) but i was thinking the static builder() method would be only on the child classes. So yeah, School.builder(), District.builder(), etc.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepository.java
wtritch: I heartily agree with your "meh" =P

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcSchoolRepository.java
wtritch: Nope, we actually instantiate Organizations in a few places... I could look for all those places and try to convert them to more concrete versions if you feel strongly?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/security/AdminPermission.java
wtritch: done

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcSchoolRepository.java
wtritch: Looks like we weren't using a concrete Organization in many places, I made the change.  I'll likely need to create a new State class for the instructional resource API, but that shouldn't cause any issues.

malaffoon: 👍

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
esotrope: service layer will still cap it somewhere right? don't think it should support the client saying `?limit=100000000`

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
esotrope: custom-exports consumes this now i gather?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
wtritch: By default it's capped at 1000, we can cap it at the service layer to XXX, but right now I had to remove the cap (Integer.MAX_VALUE) for the custom exports since the front-end expects all the data.  When I build out the API for the instructional resources auto-complete I'll probably have a cap of 1000 at the service layer and the UI shouldn't be requesting more than 10 or 20 results.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/District.java
wtritch: yessir, custom-exports is consuming this repository now.

PR #626, opened 2017-11-30
moves breadcrumbs to common-ngx

esotrope:
With this change admin and reporting breadcrumbs will behave identically.


common-ngx/src/layout/rdw-layout.module.ts
```diff
@@ -0,0 +1,23 @@
...
+    BrowserModule,
+    FormsModule,
+    RouterModule.forRoot([]),
+    TranslateModule.forRoot()
+  ],
+  exports: [
+    Breadcrumbs
+  ]
+})
+export class RdwLayoutModule {
```
esotrope: was thinking this package could hold the nav bar and footer as well - open to suggestions on a better name

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -1,112 +0,0 @@
-import { Component, OnInit } from "@angular/core";
```
esotrope: @wtritch some more lovely red for you

common-ngx/src/layout/breadcrumbs.component.ts
```diff
@@ -0,0 +1,161 @@
...
+
+      return this.createBreadcrumbs(child, routerLinkParameters, breadcrumbs);
+    }
+  }
+
+  private createBreadcrumb(options: BreadcrumbOptions, routeData: any, routerLinkParameters: any[]): Breadcrumb {
+    if (options.translate) {
+      let text = options.translateResolve
+        ? this.translateService.instant(options.translate, Utils.getPropertyValue(options.translateResolve, routeData))
+        : this.translateService.instant(options.translate);
```
esotrope: Now translating this when creating the crumb at ngOnInit time. Before we did this twice in the create title method and in the template.
This makes it to where we don't have to pass translateParams & requiresTranslate props around.

common-ngx/src/layout/rdw-layout.module.ts
wtritch: Hm...  It'll hold basic page structure/framing components... UIStructureModule? ...just shooting in the dark.

PR #627, opened 2017-11-30
Feature/embargo management ui

esotrope:
First hack at the embargo page

screen shot 2017-11-29 at 5 37 24 pm

State only
screen shot 2017-11-29 at 5 34 43 pm

Districts only
screen shot 2017-11-29 at 5 35 04 pm

Both
screen shot 2017-11-29 at 5 34 19 pm


admin-webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -9,12 +9,16 @@ <h1 class="blue-dark h1 mb-md">{{'application.title' | translate}}</h1>
...
-
+    <div class="well-body">
+      <div>
+        <a *sbAuthorize="['GROUP_WRITE']" routerLink="groups">{{ 'labels.groups.title' | translate}}</a>
+      </div>
+      <div>
+        <a *sbAuthorize="['INSTRUCTIONAL_RESOURCE_WRITE']" routerLink="instructional-resource">{{'labels.instructional-resource.title' | translate}}</a>
+      </div>
+      <div>
+        <a routerLink="embargoes">{{'labels.embargo.title' | translate}}</a>
```
esotrope: remind me to not forget `*sbAuthorize="['EMBARGO_WRITE']"` here

admin-webapp/src/main/webapp/src/app/embargo/embargo.service.ts
```diff
@@ -0,0 +1,85 @@
+import { Observable } from "rxjs/Observable";
+import { Embargo, EmbargoSettings } from "./embargo-settings";
+
+export class EmbargoService {
+
+  getEmbargoSettings(): Observable<EmbargoSettings> {
+    return Observable.of({
+      stateEmbargo: {
```
esotrope: @malaffoon WIP payload format

admin-webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.html
```diff
@@ -1,7 +1,13 @@
 <!-- Breadcrumb -->
-<div [hidden]="breadcrumbs.length <= 1">
+<div [hidden]="breadcrumbs.length == 0">
```
esotrope: temp fix for breadcrumbs on this branch

admin-webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -99,7 +101,24 @@ export const routes: Routes = [
                 ]
               },
             ]
+          },
+          {
+            path: 'embargoes',
+            pathMatch: 'prefix',
+            data: {
+              breadcrumb: { translate: 'labels.embargo.title' },
+              permissions: [ 'EMBARGO_WRITE' ]
```
esotrope: permissions will be enforced when we add the `canActivate` param

admin-webapp/src/main/webapp/src/app/embargo/embargo.service.ts
malaffoon: Yeah, we should chat about API.

nit: i think we should use unmodified `embargo` as much as possible; in this case `individualEmbargo` and `aggregateEmbargo`.

admin-webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -101,6 +101,40 @@
     "instructional-resource": {
       "title": "Manage Instructional Resources"
     },
+    "embargo": {
+      "title": "Manage Embargoes",
```
malaffoon: My knee-jerk is this should be `Manage Embargo`. As in, "the user is managing the embargo of test results". Sure, a single user could be managing the settings for multiple district embargoes but, conceptually, it feels like they are just managing the embargo setting.

admin-webapp/src/main/webapp/src/app/embargo/embargo.service.ts
malaffoon: One question/thought: i think the API needs to return the state information whether or not the user has admin rights for it. The current state setting is needed for UI logic (i.e. user needs to be notified that setting the embargo on for a district will have no effect if the state embargo is off).

esotrope: we'll need a flag in the payload since the ui only knows that the user generally has `EMBARGO_WRITE` but doesn't know for what.

admin-webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: 👍

admin-webapp/src/main/webapp/src/app/embargo/embargo-table.component.html
```diff
@@ -0,0 +1,41 @@
...
+             sortField="organization.name"
+             tableStyleClass="table table-striped table-hover"
+             [tableStyle]="{'table-layout':'auto'}" >
+  <p-column field="organizationName" header="{{organizationNameHeader}}" [sortable]="sortable">
+    <ng-template let-embargo="rowData" pTemplate="body">
+      {{embargo.organization.name}}
+    </ng-template>
+  </p-column>
+
+  <!-- would like these subject columns to be part of a forEach of sorted keys to allow for variable number of subjects -->
```
wtritch: 👍

admin-webapp/src/main/webapp/src/app/embargo/toggle.component.ts
```diff
@@ -0,0 +1,111 @@
...
+ * mirror of sb-toggle without angularitics and this one uses ngx-bootstrap button module.
+ * moving sb-toggle to common-ngx will come later
+ */
+@Component({
+  providers: [ CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR ],
+  selector: 'toggle',
+  template: `
+    <div class="btn-group btn-group-xs toggle-group">
+      <label class="btn btn-primary" [(ngModel)]="value" [btnRadio]="options[0].value" [ngClass]="{disabled: disabled}">{{options[0].text}}</label>
+      <label class="btn btn-primary" [(ngModel)]="value" [btnRadio]="options[1].value" [ngClass]="{disabled: disabled}">{{options[1].text}}</label>
```
wtritch: hm... should we generify this into an ngFor across the options array?

esotrope: 👍

esotrope:
@wtritch @malaffoon @jtreuting @dsnyder-fairway major embargo admin ui update - please re-review PR - thanks!

State embargo admin with districts

screen shot 2017-12-19 at 10 45 50 am

District embargo admin (state is readonly)

screen shot 2017-12-19 at 3 27 28 pm

Some modal examples. All text iterations are in `en.json`

screen shot 2017-12-19 at 3 24 05 pm
screen shot 2017-12-19 at 3 24 14 pm

Note the way the toggle/modal UI works is that when you click an embargo toggle the value doesn't change until the modal accept button is clicked. Otherwise we would awkwardly undo the value change when they click cancel or dismiss the modal.

malaffoon:
Makes sense; simple for user; nice.

admin-webapp/src/main/webapp/src/app/embargo/embargo-confirmation-modal.component.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.embargoService.update(embargo, scope, value)
+      .subscribe(() => {
+
+        // reflect new embargo setting in the UI
+        if (scope === EmbargoScope.Individual) {
+          embargo.individualEnabled = value;
+        } else {
+          embargo.aggregateEnabled = value;
+        }
+        toggle.value = value;
```
esotrope: funkyness todo with only applying the toggle change after the modal is accepted

admin-webapp/src/main/webapp/src/app/embargo/embargo-table.component.html
```diff
@@ -0,0 +1,82 @@
...
+
+    <ng-template let-embargo="rowData" pTemplate="body">
+      {{embargo.organization.name}}
+    </ng-template>
+  </p-column>
+
+  <p-column *ngFor="let subjectCode of subjectCodes" field="examCountsBySubject.{{subjectCode}}" styleClass="text-right" [sortable]="hasMultipleEmbargoes">
+
+    <ng-template pTemplate="header">
+      <span class="pull-right">{{translateContext + '.headers.' + subjectCode | translate}}</span>
```
esotrope: only way i could find to right-align the headers. This does not work however when the column is sortable. primeng wants to reserve the right side for the sorting icons

admin-webapp/src/main/webapp/src/app/embargo/embargo.service.ts
```diff
@@ -0,0 +1,78 @@
...
+   * @param {Embargo} embargo the embargo to update
+   * @param {EmbargoScope} scope the scope of the embargo setting to update
+   * @param {boolean} value the new value of the embargo setting
+   * @returns {Observable<Object>}
+   */
+  update(embargo: Embargo, scope: EmbargoScope, value: boolean): Observable<Object> {
+    return this.http.put(
+      `${ResourceContext}/${embargo.organization.type}/${embargo.organization.id ? embargo.organization.id : -1}/${scope}`,
+      String(value),
+      { responseType: 'text' }
```
esotrope: needed when the response is empty/no_content (controller method is void) otherwise it tries to parse it as json

PR #629, opened 2017-11-30
Modifying SQL queries - a first baby step

agorina:
I am trying to make the SQL more manageable. 
This is the first baby step: I moved all the shared permissions queries into shared properties.

### DIFF
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -61,39 +61,7 @@ sql:
         join school s on e.school_id=s.id
         join school isch on st.inferred_school_id = isch.id
       where e.student_id=:student_id
-        and
-        (
-          # Individual permissions test
-          (
-            (1 = :individual_statewide or s.district_group_id in (:individual_district_group_ids) or s.district_id in (:individual_district_ids) or s.school_group_id in (:individual_school_group_ids) or e.school_id in (:individual_school_ids))
-            or
-            (
-              1 = :allow_transfer_access
-              and (isch.district_group_id in (:individual_district_group_ids) or isch.district_id in (:individual_district_ids) or isch.school_group_id in (:individual_school_group_ids) or isch.id in (:individual_school_ids))
-            )
-          )
-          or
-          # Group permissions test
-          (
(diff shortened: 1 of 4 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -253,11 +182,11 @@ sql:
           from exam e
             join student_group_membership sgm on e.student_id=sgm.student_id
             join asmt a on e.asmt_id=a.id
-            join school s on e.school_id=s.id
+            join school sc on e.school_id=sc.id
           where e.school_year=:school_year
             and sgm.student_group_id=:group_id
             and (0=:group_subject_id or a.subject_id=:group_subject_id)
```
esotrope: you could also factor `(0=:group_subject_id or a.subject_id=:group_subject_id)` out into `groupPermission` here

webapp/src/main/resources/application.sql.yml
```diff
@@ -702,3 +583,63 @@ sql:
           OR (ir.org_level = 'District' AND district.id IS NOT NULL)
           OR (ir.org_level = 'SchoolGroup' AND school_group.id IS NOT NULL)
         )
+
+  reporting:
```
esotrope: if we put these in `common.sql.yml` do these not come through to each app's sql yml?

agorina:
Sorry, will reopen shortly

webapp/src/main/resources/application.sql.yml
agorina: Yeah, maybe later. Baby steps at this point. While these queries look very similar there are some differences and I am trying to make sure I do not break anything.

webapp/src/main/resources/application.sql.yml
agorina: They do, but they are not really common. They are just common for the reporting (meaning this file) so I did not see the value in putting it into `common.sql.yml`. Do you see it differently?

esotrope: Ah i see. i thought there was crossover between reporting and the report processor - thanks

PR #632, opened 2017-12-01
An attempt to simplify SQL maintenance - option 1.

agorina:
This is one approach using placeholders for common parts of SQL. And I am about to submit a PR for another one with the 'dynamic' SQL. Would like to know what people think. 

Please compare with : https://github.com/SmarterApp/RDW_Reporting/pull/633

### DIFF
--- a/common/src/main/resources/common.sql.yml
+++ b/common/src/main/resources/common.sql.yml
@@ -127,3 +127,67 @@ sql:
   snippet:
     basicPermission: >-
         (1=:statewide or sc.district_group_id in (:district_group_ids) or sc.district_id in (:district_ids) or sc.school_group_id in (:school_group_ids) or sc.id in (:school_ids))
+
+    claimData: >-
+          e.claim1_category,
+          e.claim1_scale_score,
+          e.claim1_scale_score_std_err,
+          e.claim2_category,
+          e.claim2_scale_score,
+          e.claim2_scale_score_std_err,
+          e.claim3_category,
+          e.claim3_scale_score,
+          e.claim3_scale_score_std_err,
+          e.claim4_category,
+          e.claim4_scale_score,
+          e.claim4_scale_score_std_err,
+          a.claim1_score_code,
+          a.claim2_score_code,
+          a.claim3_score_code,
+          a.claim4_score_code,
+          a.claim1_score_code as asmt_claim1_score_code,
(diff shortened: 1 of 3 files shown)


common/src/main/resources/common.sql.yml
```diff
@@ -127,3 +127,67 @@ sql:
   snippet:
     basicPermission: >-
         (1=:statewide or sc.district_group_id in (:district_group_ids) or sc.district_id in (:district_ids) or sc.school_group_id in (:school_group_ids) or sc.id in (:school_ids))
+
+    claimData: >-
```
esotrope: isn't everything in the database data? 😉

webapp/src/main/resources/application.sql.yml
```diff
@@ -329,54 +242,14 @@ sql:
...
-          s.district_id as school_district_id,
-          s.district_group_id as school_district_group_id,
-          s.school_group_id
-        from exam e
-          join student st on e.student_id=st.id
-          join asmt a on e.asmt_id=a.id
-          join school s on e.school_id=s.id
-          join school isch on st.inferred_school_id = isch.id
+          ${sql.snippet.claimData},
+          ${sql.snippet.exam}
```
esotrope: since this reappears a lot we could make an examWithClaims snippet too maybe

common/src/main/resources/common.sql.yml
agorina: :)  better name ?

webapp/src/main/resources/application.sql.yml
agorina: I was trying to avoid nesting placeholders. But if people like it, I can do that.

common/src/main/resources/common.sql.yml
```diff
@@ -127,3 +127,67 @@ sql:
   snippet:
```
jtreuting: are you thinking that snippet would hold select clauses as well as possible where clauses or joins?  and if so, should they be separated to easily tell what they are for instead of having snippets show in the select and where.

agorina: The way I see it is that a snippet is just a replacement for a 'copy and paste'. Because of that I was not considering breaking it into smaller parts. If you think it will be easier to read, we can discuss it.

PR #633, opened 2017-12-01
Another attempt to simplify SQL maintenance - option 2

agorina:
This is meant to be compared to https://github.com/SmarterApp/RDW_Reporting/pull/632.


common/src/main/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/QueryProvider.java
```diff
@@ -68,7 +69,7 @@ public String newQueryAsString(final String templateName, final Iterable<String>
 
     private void addClauses(final SqlTemplate template, final SelectQuery query) {
         // FROM clause is applicable to the main template only and is handled separately
-        final SqlClause[] clauses = {SELECT, JOIN, WHERE, GROUP_BY};
+        final SqlClause[] clauses = {SELECT, JOIN, WHERE, GROUP_BY, ORDER_BY};
```
agorina: This is unrelated, I was missing  `order by` support and an example query I choose needed it.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepository.java
```diff
@@ -131,11 +127,11 @@
...
             case Unknown:
-                return findAllIabsByStudentIdsAndQueryParamsUnknownPermissions;
+                return queryProvider.newQueryAsString("findAllExamsByStudentIdsAndQueryParams", ImmutableList.of("iab", "unknownScopePermission"));
             case Individual:
-                return findAllIabsByStudentIdsAndQueryParamsIndividualPermissions;
+                return queryProvider.newQueryAsString("findAllExamsByStudentIdsAndQueryParams", ImmutableList.of("iab", "individualScopePermission"));
             case Group:
-                return findAllIabsByStudentIdsAndQueryParamsGroupPermissions;
+                return queryProvider.newQueryAsString("findAllExamsByStudentIdsAndQueryParams", ImmutableList.of("iab", "groupScopePermission" ,"individualScopePermission"));
             default:
```
agorina: This is how the final SQL is assembled from the `template`.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepository.java
```diff
@@ -131,11 +127,11 @@
...
         switch (queryType) {
             case Unknown:
-                return findAllIabsByStudentIdsAndQueryParamsUnknownPermissions;
+                return queryProvider.newQueryAsString("findAllExamsByStudentIdsAndQueryParams", ImmutableList.of("iab", "unknownScopePermission"));
             case Individual:
-                return findAllIabsByStudentIdsAndQueryParamsIndividualPermissions;
+                return queryProvider.newQueryAsString("findAllExamsByStudentIdsAndQueryParams", ImmutableList.of("iab", "individualScopePermission"));
             case Group:
-                return findAllIabsByStudentIdsAndQueryParamsGroupPermissions;
+                return queryProvider.newQueryAsString("findAllExamsByStudentIdsAndQueryParams", ImmutableList.of("iab", "groupScopePermission" ,"individualScopePermission"));
```
esotrope: are the list of strings building the query?

agorina: Yes, it `chains` the `addon's by the names on top of the main template SQL.

esotrope: Wondering if we should build these on the fly or at repo construction time.
Also, have we come up with some criteria for when we dynamically build queries vs statically construct them with snippet substitution?
Seems like we would use the dynamic construction only when user input can cause a ton of permutations

agorina: Yes, you are correct: we should use the dynamic construction only when the user input can cause a ton of permutations. This is not the case for this query. So I guess you are voting for option 1? Or just leave it as is? 

(I think they are all hard to read,not sure which way to go).

esotrope: yeah for this i would go with snippets since the permutations are small, known and not directly influenced by user input. its good that we're testing out this cool new query toy though : j

PR #634, opened 2017-12-01
Introduce an admin instructional-resource api and organization search…

wtritch:
… api

This PR introduces an Admin CRUD API for Instructional Resources as well as an Organization search API.

GET /api/instructional-resources - returns all instructional resources that the user has permissions to edit/delete.
POST /api/instructional-resources - creates a new instructional resource after validating that the user has permissions for the targeted organization
PUT /api/instructional-resources - updates the resource url for the given instructional resource after validating that the user has permissions for the targeted organization
DELETE /api/instructional-resources?assessmentId=1&organizationType=district&performanceLevel=0&organizationId=2 - deletes the targeted (single) instructional resource after validating that the user has permissions to the targeted organization

GET /api/organizations?permission=instructional_resource_write&name=calif&types=district&types=school&limit=20 - searches for organizations of type District and/or School granted to the user with INSTRUCTIONAL_RESOURCE_WRITE permissions with "calif" in their name limited to 20 results.

### DIFF
--- a/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/InstructionalResourceService.java
+++ b/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/InstructionalResourceService.java
@@ -1,0 +1,51 @@
+package org.opentestsystem.rdw.admin.service;
+
+import org.opentestsystem.rdw.admin.model.InstructionalResource;
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+import org.springframework.security.access.prepost.PreAuthorize;
+
+import java.util.List;
+
+/**
+ * Implementations of this service are responsible for working with InstructionalResource instances.
+ */
+@PreAuthorize("hasAuthority('PERM_INSTRUCTIONAL_RESOURCE_WRITE')")
+public interface InstructionalResourceService {
+
+    /**
+     * Find all instructional resources that the given user has permissions
+     * to modify. Results are ordered by assessment name, organization level,
+     * organization, performance level.
+     *
+     * @param user A user
+     * @return All instructional resources
+     */
+    List<InstructionalResource> findAll(User user);
+
+    /**
+     * Create a new instructional resource.
+     *
+     * @param user                  A user
+     * @param instructionalResource A new instructional resource
+     * @return The created instructional resource
+     */
+    InstructionalResource create(User user, InstructionalResource instructionalResource);
+
+    /**
+     * Update an existing instructional resource.
+     *
+     * @param user                  A user
+     * @param instructionalResource An updated instructional resource
+     * @return The updated instructional resource
+     */
+    InstructionalResource update(User user, InstructionalResource instructionalResource);
+
+    /**
+     * Delete an existing instructional resource.
+     * NOTE: Deleting a non-existent instructional resource behaves as a no-op.
+     *
+     * @param user                  A user
+     * @param instructionalResource An instructional resource
+     */
+    void delete(User user, InstructionalResource instructionalResource);
+}
--- a/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
+++ b/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
@@ -1,0 +1,113 @@
+package org.opentestsystem.rdw.admin.service.impl;
+
+import com.google.common.collect.Ordering;
+import org.opentestsystem.rdw.admin.model.InstructionalResource;
+import org.opentestsystem.rdw.admin.repository.InstructionalResourceRepository;
+import org.opentestsystem.rdw.admin.service.InstructionalResourceService;
+import org.opentestsystem.rdw.reporting.common.model.Organization;
+import org.opentestsystem.rdw.reporting.common.repository.OrganizationRepository;
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+import org.opentestsystem.rdw.security.PermissionScope;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+
+import static org.opentestsystem.rdw.admin.security.AdminPermission.InstructionalResourceWrite;
+import static org.opentestsystem.rdw.reporting.common.model.OrganizationType.State;
+
+/**
+ * Default repository-backed implementation of an InstructionalResourceService.
+ */
+@Service
+class DefaultInstructionalResourceService implements InstructionalResourceService {
+    private static final Ordering<InstructionalResource> ResourceOrdering = Ordering.natural()
+            .onResultOf(InstructionalResource::getAssessmentName)
+            .compound(Ordering.natural()
+                    .onResultOf(InstructionalResource::getOrganizationType))
+            .compound(Ordering.natural()
+                    .nullsFirst()
+                    .onResultOf(InstructionalResource::getOrganizationName))
+            .compound(Ordering.natural()
+                    .onResultOf(InstructionalResource::getPerformanceLevel));
+
+
+    private final InstructionalResourceRepository repository;
+    private final OrganizationRepository organizationRepository;
+
+    @Autowired
+    public DefaultInstructionalResourceService(final InstructionalResourceRepository repository,
+                                               final OrganizationRepository organizationRepository) {
+        this.repository = repository;
+        this.organizationRepository = organizationRepository;
+    }
+
+    @Override
+    public List<InstructionalResource> findAll(final User user) {
+        return ResourceOrdering.sortedCopy(repository.findAll(toScope(user)));
+    }
+
+    @Override
+    @Transactional(transactionManager = "reportingTxManager")
+    public InstructionalResource create(final User user, final InstructionalResource instructionalResource) {
+        final PermissionScope scope = toScope(user);
+
+        //Ensure user has access to the specified organization
+        if (invalidOrganization(scope, instructionalResource)) {
+            throw new NoSuchElementException("Organization: " + instructionalResource.getOrganizationType().name() +
+                    " id: " + instructionalResource.getOrganizationId() + " not found");
+        }
+
+        repository.create(instructionalResource);
+        return repository.findOne(scope, instructionalResource);
+    }
+
+    @Override
+    @Transactional(transactionManager = "reportingTxManager")
+    public InstructionalResource update(final User user, final InstructionalResource instructionalResource) {
+        final PermissionScope scope = toScope(user);
+
+        final InstructionalResource existing = repository.findOne(scope, instructionalResource);
+        if (existing == null) {
+            throw new NoSuchElementException("Instructional Resource for Organization: " + instructionalResource.getOrganizationType().name() +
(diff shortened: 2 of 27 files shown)


admin-webapp/src/main/resources/application.yml
```diff
@@ -123,6 +123,7 @@ permissionservice:
 app:
   state:
     code: CA
+    name: California
```
esotrope: my only q is - would this ever be translated?

wtritch: I decided no, since it's a Pronoun.  It would be like translating Bob.

esotrope: I think we can probably get away with it. However  in Japanese it is written as カリフォルニア

wtritch: I am "so" not supporting searching our organization names in Japanese.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -0,0 +1,113 @@
...
+
+import static org.opentestsystem.rdw.admin.security.AdminPermission.InstructionalResourceWrite;
+import static org.opentestsystem.rdw.reporting.common.model.OrganizationType.State;
+
+/**
+ * Default repository-backed implementation of an InstructionalResourceService.
+ */
+@Service
+class DefaultInstructionalResourceService implements InstructionalResourceService {
+    private static final Ordering<InstructionalResource> ResourceOrdering = Ordering.natural()
```
esotrope: 👍

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        repository.delete(instructionalResource);
+    }
+
+    private PermissionScope toScope(final User user) {
+        return user.getPermissionScopeByPermissionId(InstructionalResourceWrite);
+    }
+
+    private boolean invalidOrganization(final PermissionScope permissionScope, final InstructionalResource instructionalResource) {
```
esotrope: would make this a positive function name `isValidOrganization` `hasPermissionToEditOrganization` or something

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -0,0 +1,113 @@
...
+        repository.update(updated);
+        return updated;
+    }
+
+    @Override
+    @Transactional(transactionManager = "reportingTxManager")
+    public void delete(final User user, final InstructionalResource instructionalResource) {
+        final PermissionScope scope = toScope(user);
+
+        //Ensure user has access to the specified organization
```
esotrope: this if block could be in a function too like `assertAccess()`

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
wtritch: Interesting, I prefer positive assertions:
`if (invalidOrganization(...))`
over
`if (!validOrganization(...))`
much like the apache string checks: "isNotBlank(...)" and "isBlank(...)"

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultOrganizationService.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public List<Organization> findByQuery(final PermissionScope permissionScope, final OrganizationQuery query) {
+        if (permissionScope.equals(PermissionScope.EMPTY)) {
+            return of();
+        }
+
+        final OrganizationQuery limitedQuery = OrganizationQuery.builder()
+                .copy(query)
+                .limit(Math.min(MaxLimit, query.getLimit()))
```
esotrope: would there be an issue if the query had `-1`? wondering if we should Math.clamp here

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceArgumentResolver.java
```diff
@@ -0,0 +1,34 @@
...
+    public boolean supportsParameter(final MethodParameter parameter) {
+        return parameter.getParameterType().equals(InstructionalResource.class);
+    }
+
+    @Override
+    public Object resolveArgument(final MethodParameter parameter, final ModelAndViewContainer mavContainer, final NativeWebRequest webRequest, final WebDataBinderFactory binderFactory) throws Exception {
+        return InstructionalResource.builder()
+                .assessmentId(parseInt(webRequest.getParameter("assessmentId")))
+                .organizationType(OrganizationType.caseInsensitiveValue(webRequest.getParameter("organizationType")))
+                .organizationId(Ints.tryParse(trimToEmpty(webRequest.getParameter("organizationId"))))
```
esotrope: didn't know about this util - interesting

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultOrganizationService.java
wtritch: hm... good point.  You'd probably end up with a 500 backed by a SQL exception.  I'll clamp to 0.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceArgumentResolver.java
wtritch: I just wish it handled and propagated nulls... but I still need that trimToEmpty in there =/

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
wtritch: Good call, done.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/InstructionalResourceService.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Find all instructional resources that the given user has permissions
+     * to modify. Results are ordered by assessment name, organization level,
+     * organization, performance level.
+     *
+     * @param user A user
+     * @return All instructional resources
+     */
+    List<InstructionalResource> findAll(User user);
```
malaffoon: nit: is response collection really ordered?

admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -73,3 +73,155 @@ sql:
...
+
+    update: >-
+      UPDATE instructional_resource ir
+        JOIN asmt a
+          ON a.name = ir.asmt_name
+      SET
+        ir.resource = :resource
+      WHERE ir.org_level != 'System'
+        AND ir.org_level = :org_level
+        AND a.id = :asmt_id
```
malaffoon: This is just making me think ... the update is based on asmt id even though the IR is keyed off of asmt name?
 That doesn't feel right.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceController.java
```diff
@@ -0,0 +1,59 @@
...
+public class InstructionalResourceController {
+
+    private final InstructionalResourceService service;
+
+    @Autowired
+    public InstructionalResourceController(final InstructionalResourceService service) {
+        this.service = service;
+    }
+
+    @GetMapping("/api/instructional-resources")
```
chdietz: Is it better to put /api in @RequestMapping at class level?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceController.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @PostMapping("/api/instructional-resources")
+    @ResponseStatus(CREATED)
+    public InstructionalResource createResource(@AuthenticationPrincipal final User user,
+                                                @RequestBody final InstructionalResource instructionalResource) {
+        return service.create(user, instructionalResource);
+    }
+
+    @PutMapping("/api/instructional-resources")
```
chdietz: I'm curious if you mulled over other names for this resource, as we have '-resources' in the resource name?

admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -73,3 +73,155 @@ sql:
...
+              (
+                school_group.id in (:school_group_ids)
+                OR
+                (school_group.id = sch.school_group_id AND (sch.district_id in (:district_ids) OR sch.district_group_id in (:district_group_ids)))
+              )
+            )
+          )
+        )
+
+    findOne: >-
```
chdietz: instructional_resource has unique (asmt_name, org_level, performance_level, org_id) so why do we need DISTINCT?

admin-webapp/src/main/resources/application.sql.yml
wtritch: ...shoot.  I think you're right.  I don't want the assessment id at all, since a name may have an assessment for every year...

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceController.java
wtritch: I keep fighting with the naming, but "Instructional" feels wrong, since this model represents a link to an external "Instructional Resource."

admin-webapp/src/main/resources/application.sql.yml
wtritch: Because we're joining down to the school table for permissions, and their may be multiple school records that grant access to the same instructional resource.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceController.java
wtritch: Good call, done

admin-webapp/src/main/resources/application.sql.yml
chdietz: Is this because we support down to Group of Schools and permissions for this org type comes from the schools in the Group of Schools?

admin-webapp/src/main/resources/application.sql.yml
wtritch: I removed the "assessmentId" property from InstructionalResource and added an "assessmentLabel" property for more user-friendly display.

admin-webapp/src/main/resources/application.sql.yml
wtritch: Permissions for *all* org types come from the DB school records in reporting.

wtritch:
After discussion I realized I'm missing an API to find assessments by label.  I'll add that API in a follow-up PR.

PR #635, opened 2017-12-01
Task/move footer and nav to common ngx

esotrope:
moved header, nav and footer to common-ngx to reduce duplicate code.


admin-webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1,8 +1,20 @@
...
+      "logo": {
+        "title": "Application Administration Dashboard",
+        "alt": "Smarter Balanced Assessment Consortium"
+      },
+      "title": "Application Administration",
+      "toggle-sr": "Toggle navigation",
+      "user-dropdown": "Welcome {{firstName}} {{lastName}}",
+      "logout": "Logout"
+    },
+    "footer": "\u00A9 The Regents of the University of California \u2013 Smarter Balanced Assessment Consortium",
```
esotrope: debated wether or not to put these under common-ngx or not. ideally the app would pass all these in directly as params so there is no hidden dependencies.

PR #636, opened 2017-12-01
Task/add permission handling to common ngx

esotrope:


admin-webapp/src/main/webapp/src/app/user/user-permission.service.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Observable } from "rxjs/Observable";
+
+@Injectable()
+export class UserPermissionService {
+
+  constructor(private userService: UserService){
+  }
+
+  getPermissions(): Observable<string[]> {
+    return this.userService.getCurrentUser().map(user => user.permissions);
```
esotrope: hook into common security components

admin-webapp/src/main/webapp/src/app/user/user.module.ts
```diff
@@ -1,17 +1,24 @@
...
+  imports: [
+    BrowserModule,
+    CommonModule,
+    RdwSecurityModule
+  ],
+  providers: [
+    UserService,
+    UserMapper,
+    UserResolve,
+    { provide: PermissionService, useClass: UserPermissionService }
```
esotrope: how you hook into common permission dependent components

common-ngx/src/security/authorization.directive.ts
```diff
@@ -0,0 +1,59 @@
+import { Directive, Input, TemplateRef, ViewContainerRef } from "@angular/core";
+import { AuthorizationService } from "./authorization.service";
+
+/**
+ * Structural directive that conditionally renders an element if the user has permission to view it.
+ *
+ * Example usages:
+ * <div *hasPermission="'GROUP_WRITE'"></div>
+ * <div *hasAnyPermission="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE']"></div>
```
esotrope: note: signature change. thought this was more straight forward

common-ngx/src/security/rdw-security.module.ts
```diff
@@ -8,25 +8,46 @@ import { AuthenticatedHttpService } from "./authenticated-http.service";
...
   ],
   providers: [
     { provide: Http, useClass: AuthenticatedHttpService },
+    { provide: APP_BASE_HREF, useValue: '/' },
+    { provide: AccessDeniedRoute, useValue: 'access-denied' },
     AuthenticationService,
-    {provide: APP_BASE_HREF, useValue : '/' }
+    AuthorizationService,
+    AuthorizationCanActivate,
+    RoutingAuthorizationCanActivate,
```
esotrope: probably a better name for these `CanActivate`s but i didn't try too hard to think of one

webapp/src/main/webapp/src/app/user/authorize-at-least-one.can-activate.ts
```diff
@@ -1,28 +0,0 @@
...
- Allows access to a route as long as the permissions are not empty.
- */
-@Injectable()
-export class AuthorizeAtleastOneCanActivate implements CanActivate {
-
-  constructor(private _service: UserService, private router: Router) {
-  }
-
-  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean>|Promise<boolean>|boolean {
-    let observable = this._service.doesCurrentUserHaveAnyPermissions().share();
```
esotrope: not sure if there is an effective diff between using `.share().subscribe(router.navigate())` here and the`.map(router.navigate())` now used in the common class.
Please let me know if you guys think there is

common-ngx/src/security/authorization.directive.ts
jtreuting: i like that

webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -9,10 +9,12 @@ <h1 class="blue-dark h1 mb-md">{{'labels.homepage.title' | translate}}</h1>
       <div class="pull-right">
 
         <!-- Custom Exports Link -->
-        <a *sbAuthorize="['INDIVIDUAL_PII_READ']" routerLink="/custom-export" class="btn btn-xs btn-default"><i class="fa fa-bold fa-table"></i> {{'labels.organization-export.title' | translate}}</a>
+        <a *hasPermission="'INDIVIDUAL_PII_READ'"
+           routerLink="/custom-export" class="btn btn-xs btn-default"><i class="fa fa-bold fa-table"></i> {{'labels.organization-export.title' | translate}}</a>
 
         <!-- Student Groups Admin Link -->
-        <a *sbAuthorize="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE']" href="{{user.configuration.adminWebappUrl}}" class="btn btn-xs btn-info ml-xs"><i class="fa fa-gear"></i> {{'labels.groups.admin-button' | translate}}</a>
+        <a *hasAnyPermission="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE', 'EMBARGO_WRITE']"
```
jtreuting: i was originally going to ask why is `EMBARGO_WRITE` tied to the student group admin section?  but then i realized that the comment on line 15 is just outdated.  while you are in there can you change that to say "Application Admin Link"

webapp/src/main/webapp/src/app/user/authorize-at-least-one.can-activate.ts
esotrope: Current impl for ref:
```typescript
    return this.permissionService.getPermissions()
      .map(permissions => {
        const hasPermission: boolean = (permissions || []).length > 0;
        if (!hasPermission) {
          this.router.navigate([ this.accessDeniedRoute ]);
        }
        return hasPermission;
      });
```
@wtritch @bgsavage thoughts? concerns?

admin-webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -10,10 +10,10 @@ <h1 class="blue-dark h1 mb-md">{{'application.title' | translate}}</h1>
   <!-- Row Content Area -->
   <div class="well">
     <!-- Manage Student Groups -->
-    <a *sbAuthorize="['GROUP_WRITE']" routerLink="groups">{{ 'labels.groups.title' | translate}} <br/></a>
+    <a *hasPermission="'GROUP_WRITE'" routerLink="groups">{{ 'labels.groups.title' | translate}} <br/></a>
```
bgsavage: The idea with the `sb` prefix here was that it's a candidate to be promoted to some larger smarter balanced ecosystem if/when there's one available for front end components.

common-ngx/src/security/authorization.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+@Directive({ selector: '[hasPermission],[hasAnyPermission]' })
+export class AuthorizationDirective {
```
bgsavage: Not a fan of having a selector with a different name than the class.  This will make it harder to find when hitting your favorite keystrokes (command + T)

webapp/src/main/webapp/src/app/user/authorize-at-least-one.can-activate.ts
bgsavage: I prefer an approach where you don't need to expose permissions everywhere, rather, do all your permission checking in a single service.  But I wrote the original so... :)

bgsavage:
As a learning experience for me, I'd be more curious about the details of what this refactoring is trying to address / solve.  i.e. what's the ROI of going this approach versus existing?

webapp/src/main/webapp/src/app/user/authorize-at-least-one.can-activate.ts
bgsavage: Nvm, this is coming from an AuthorizationService 👍

bgsavage: Just watch your network calls...  I can't remember if I did the `.share()` so that it didn't make a network call on each permission check or not.

esotrope: k - ill check. pretty sure it wont because userService uses CachingDataService though

common-ngx/src/security/authorization.directive.ts
esotrope: no prob - i can break this up into `has-permission.directive.ts` and `has-any-permission...`

esotrope:
@bgsavage The buy here is to remove the duplicate code (and associated maintenance cost).
This package moves 
`authorize.directive.ts` `authorize.can-activate.ts` and one more `can-activate` to common-ngx

common-ngx/src/security/authorization.directive.ts
wtritch: Would that lead to duplication?  And isn't IntelliJ smart enough to navigate to source regardless of class name?

PR #637, opened 2017-12-01
Feature/sql

agorina:
A few more changes along the same lines - trying to remove SQL duplication.


common/src/main/resources/common.sql.yml
```diff
@@ -128,66 +128,66 @@ sql:
...
-          e.claim4_scale_score_std_err,
-          a.claim1_score_code,
-          a.claim2_score_code,
-          a.claim3_score_code,
-          a.claim4_score_code,
-          a.claim1_score_code as asmt_claim1_score_code,
-          a.claim2_score_code as asmt_claim2_score_code,
-          a.claim3_score_code as asmt_claim3_score_code,
-          a.claim4_score_code as asmt_claim4_score_code
+    claimColumns: >-
```
agorina: @esotrope, it this better ? :)

common/src/main/resources/common.sql.yml
```diff
@@ -128,66 +128,66 @@ sql:
...
+        e.claim1_category,
+        e.claim1_scale_score,
```
agorina: The rest (in this file) it just formatting.

common/src/main/resources/common.sql.yml
esotrope: haha yeah that works : j

PR #638, opened 2017-12-01
user permission service now maps null user to empty permission set

esotrope:


admin-webapp/src/main/webapp/src/app/user/user-permission.service.ts
```diff
@@ -9,7 +9,7 @@ export class UserPermissionService {
   }
 
   getPermissions(): Observable<string[]> {
-    return this.userService.getCurrentUser().map(user => user.permissions);
+    return this.userService.getCurrentUser().map(user => user ? user.permissions : []);
```
esotrope: unresolved user now maps to no permissions.

PR #639, opened 2017-12-02
Change assessment.name to assessment.label

wtritch:
This PR just changes Assessment#name to Assessment#label to align with the DB property being used.  This is in preparation for re-introducing Assessment#name populated by the associated DB property.


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
```diff
@@ -13,7 +13,7 @@
 public class Assessment {
 
     private long id;
-    private String name;
+    private String label;
```
wtritch: This is the big change, everything else is just fallout.

webapp/src/main/webapp/src/app/assessments/assessment.comparator.ts
```diff
@@ -1,8 +1,8 @@
-import { byString, byNumber, join } from "@kourge/ordering/comparator";
```
jtreuting: assuming this was just never used

wtritch: yessir

PR #640, opened 2017-12-05
Add an assessment query API to the admin application.

wtritch:


admin-webapp/src/test/java/org/opentestsystem/rdw/admin/web/AssessmentControllerIT.java
```diff
@@ -0,0 +1,89 @@
...
+            .build();
+
+    @Test
+    public void itShouldGetAssessments() throws Exception {
+        final Assessment assessmentA = assessment(1, "Assessment A");
+        final Assessment assessmentB = assessment(2, "Assessment B");
+        when(service.findAssessments(queryCaptor.capture()))
+                .thenReturn(ImmutableList.of(assessmentA, assessmentB));
+
+        mvc.perform(get("/api/assessments?limit=21&label=label")
```
wtritch: Sample call to the API

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/ReportingJdbcAssessmentRepository.java
```diff
@@ -0,0 +1,54 @@
...
+    public Collection<Assessment> findAll(final AssessmentQuery query) {
+        return template.query(
+                findAllQuery,
+                new MapSqlParameterSource()
+                        .addValues(queryParams(query)),
+                (row, rowNum) -> mapAssessment(row));
+    }
+
+    private Map<String, Object> queryParams(final AssessmentQuery query) {
+        return new MapSqlParameterSource()
```
esotrope: do we need to make `MapSqlParameterSource` twice?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultAssessmentService.java
```diff
@@ -0,0 +1,34 @@
...
+    @Autowired
+    public DefaultAssessmentService(final AssessmentRepository repository) {
+        this.repository = repository;
+    }
+
+    @Override
+    public Collection<Assessment> findAssessments(final AssessmentQuery query) {
+        final AssessmentQuery limitedQuery = AssessmentQuery.builder()
+                .copy(query)
+                .limit(Math.max(0, Math.min(MaxResults, query.getLimit())))
```
esotrope: may get used only once but it would be more readable if we had a clamp util function

admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -3,6 +3,31 @@ sql:
...
+          a.claim2_score_code,
+          a.claim3_score_code,
+          a.claim4_score_code,
+          a.min_score,
+          a.max_score,
+          a.cut_point_1,
+          a.cut_point_2,
+          a.cut_point_3
+        FROM asmt a
+        WHERE
```
agorina: I am just trying to understand: the reason you need all of the columns is because you are reusing a common object and mapper?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/AssessmentQueryArgumentResolver.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.web.method.support.ModelAndViewContainer;
+
+import static org.apache.commons.lang.StringUtils.trimToEmpty;
+
+/**
+ * Argument resolver for AssessmentQuery instances.
+ */
+public class AssessmentQueryArgumentResolver implements HandlerMethodArgumentResolver {
+
+    static final int DefaultLimit = 20;
```
jtreuting: maybe it's because it's late but i'm having a hard time keeping these static limits straight.  The service and repository have max limits and default limits of 1000 and here we have 20.  Not sure what my question is, just throwing it out there cause it seems like maybe there could be some consolidation or there is a big discrepancy or something :)

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/ReportingJdbcAssessmentRepository.java
malaffoon: +1 - why not just return the parameter source?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/AssessmentController.java
```diff
@@ -0,0 +1,31 @@
...
+ * This controller is responsible for providing Assessments.
+ */
+@RestController
+@RequestMapping("/api")
+public class AssessmentController {
+
+    private final AssessmentService service;
+
+    @Autowired
+    public AssessmentController(final AssessmentService service) {
```
malaffoon: ~~public~~ ?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/ReportingJdbcAssessmentRepository.java
wtritch: doh, copy-paste fail.  Fixed.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/AssessmentController.java
wtritch: removed

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/AssessmentQueryArgumentResolver.java
wtritch: one is a controller default (if you don't specify a limit, you're limited to 20)
one is a service-level maximum (even if you specify 987928798234 results, you'll only get 1000)

admin-webapp/src/main/resources/application.sql.yml
wtritch: Correct, I actually wanted a potentially-reusable Assessment API as opposed to a one-off AssessmentName API which wouldn't be re-usable.  While I like the idea of consolidating all our Assessment requests into a single, common, restful AssessmentRepository I feel like that's a bit out of scope...  thoughts?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultAssessmentService.java
wtritch: While I agree with you, this doesn't really belong in QueryUtils and it feels like a very small case to justify creating a ReportingUtils common class... thoughts?

esotrope: i think actually we use clamp logic in the ViewSupport class for pdf templates. might be worth making that common math util. minimally we should have a private method called clamp here i think

PR #641, opened 2017-12-05
 A new bare-bone spring boot app for olap reporting.

agorina:
All what it has is a connection to Redshift and a diagnostic API that I copied from the olap-migrate project.


olap-webapp/src/main/java/org/opentestsystem/rdw/olapreporting/status/DatabaseStatusIndicator.java
```diff
@@ -0,0 +1,73 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A {@link StatusIndicator} for database status.
+ */
+@Component
+public class DatabaseStatusIndicator extends AbstractStatusIndicator {
+    private final NamedParameterJdbcTemplate olapJdbcTemplate;
```
agorina: I copied this from the olap-migrate project. Maybe at some point it should be moved into common, but I am not doing it yet (unless there is a strong objection).

olap-webapp/src/main/java/org/opentestsystem/rdw/olapreporting/status/DatabaseStatusIndicator.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public String name() {
+        return "OLAP-database-operations";
+    }
+
+    private Status getOlapReadStatus() {
+        final Status.Builder builder = Status.builder()
+                .detail("type", "READ");
+        responseTime(builder, 10000, () ->
```
malaffoon: ugh

olap-webapp/src/main/resources/application.yml
```diff
@@ -0,0 +1,45 @@
...
+  aws:
+    # there must be a default region when running locally
+    region:
+      auto: false
+      static: us-west-2
+    # disable CloudFormation stuff
+    stack:
+      auto: false
+    credentials:
+      instance-profile: false
```
agorina: Do I need the above section: `cloud`?
I think I do for the diagnostic API, but not sure.

olap-webapp/src/main/resources/application.yml
```diff
@@ -0,0 +1,45 @@
...
+
+app:
+  datamart:
+     olap_datasource:
+        url: "\
+           jdbc:redshift://${spring.olap_datasource.url-server:localhost:5432}/dev\
+           ?ApplicationName=rdw-migrate-olap\
+           &currentSchema=${spring.olap_datasource.url-schema:reporting_olap}
+           "
+        username: root
```
agorina: I think it was noted before: the convention seems to be diff. between ingest and reporting for the datasource property name. I am using the reporting convention here since it is part of this project.

olap-webapp/src/main/java/org/opentestsystem/rdw/olapreporting/status/DatabaseStatusIndicator.java
agorina: Could you please clarify your emotions? :)

settings.gradle
```diff
@@ -4,6 +4,7 @@ include ':common-ngx'
 include ':webapp'
 include ':report-processor'
 include ':admin-webapp'
+include ':olap-webapp'
```
malaffoon: ditto (about `webapp` part)

olap-webapp/src/main/java/org/opentestsystem/rdw/olapreporting/Application.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.olapreporting;
```
malaffoon: i don't feel strongly about this but ... is "olapreporting" kind of redundant, could this be just `org.opentestsystem.rdw.olap`?

olap-webapp/src/main/resources/bootstrap.yml
```diff
@@ -0,0 +1,17 @@
+# Spring Cloud applications have a bootstrap context so certain properties
+# should be defined here and not in application.yml/properties.
+# (https://github.com/spring-cloud/spring-cloud-commons/blob/master/docs/src/main/asciidoc/spring-cloud-commons.adoc#the-bootstrap-application-context)
+spring:
+  application:
+    name: rdw-olap-reporting-webapp
```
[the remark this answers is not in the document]
agorina: I fine with renaming it to rdw-reporting-olap.

olap-webapp/src/main/java/org/opentestsystem/rdw/olapreporting/Application.java
agorina: Sure, will change.

olap-webapp/src/main/java/org/opentestsystem/rdw/olapreporting/status/DatabaseStatusIndicator.java
malaffoon: Oh nothing to change, i was only complaining about the fact that the cutoff for the read test is 10s.

olap-webapp/src/main/resources/application.yml
malaffoon: Yes, you'll want this when running in an EC2 instance.

olap-webapp/src/main/resources/application.yml
malaffoon: I really wish we had stayed consistent across apps: it is annoying when editing the configuration files and will eventually lead to confusion for a user. Personally, i don't see the need to have the `app` namespace. Everything that is not a third party property will be in the blank namespace which, by convention, is the application namespace. And these are Spring Boot apps so having a bunch of `spring` configuration doesn't bother me at all. 

All that said, being consistent within the reporting suite probably makes sense: they can be left alone together or cleaned up together.

PR #644, opened 2017-12-07
jwt components for poc

esotrope:
This PR demonstrates how we can use JWT to communicate from ui spring boot wrapper to api services. 

Didn't check this in but I used it to test that it works:

in report-processor:
```java
@RestController
public class PocController {

    @RequestMapping("/poc")
    public User poc(@AuthenticationPrincipal final User user) {
        return user;
    }

}
```

in reporting webapp angular ui:
```typescript
poc(): void {
    this.dataService.get('/poc/poc')
      .map(response => response.json())
      .subscribe(data => {
        console.log('/poc', data)
      })
  }
```


webapp/src/main/resources/application.yml
```diff
@@ -134,3 +134,12 @@ zuul:
       path: /iris/**
       url: ${app.iris.url}
       stripPrefix: true
+    poc:
+      path: /api/poc/**
+      url: http://localhost:8085
```
esotrope: in a container context this would really be a pod service url like /myservice i think

webapp/src/main/resources/application.yml
```diff
@@ -134,3 +134,12 @@ zuul:
...
+      stripPrefix: true
+
+
+jwt:
+  secret: ${rdw.reporting.jwt.secret}
+  producer: true
```
esotrope: this is to distinguish the ui boot wrapper from api services to configure them right

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/JwtZuulFilter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public int filterOrder() {
+        return 0;
+    }
+
+    @Override
+    public boolean shouldFilter() {
+        return true;
+    }
```
wtritch: This should probably test the route to avoid sending our JWT tokens to Iris

esotrope: Hm good point. interesting implications though. we should probably drive the filter matching through the yml since that is where the routes are. otherwise they would be in a spring config file. Thinking i'll try and add a `jwt.filter-paths` prop or something

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/DefaultJwtService.java
```diff
@@ -25,9 +24,6 @@
     @Override
     public String encode(final User user) {
         return Jwts.builder()
-                .setIssuer(settings.getIssuer())
-                .setIssuedAt(new Date())
-                .setSubject(user.getUsername())
```
esotrope: removed these things because we don't really need them - we're not serving tokens for a public api

webapp/src/main/resources/application.yml
```diff
@@ -143,3 +143,5 @@ zuul:
 jwt:
   secret: ${rdw.reporting.jwt.secret}
   producer: true
+  routePaths: ${zuul.routes.poc.path}
```
esotrope: each new zuul api service route will need to be added here.

esotrope:
@wtritch @chdietz @jtreuting @malaffoon @agorina @dsnyder-fairway 
Please re-review this PR - I have made significant additions:
1. The webapp module now forwards to the report processor using zuul.
2. All report code has been moved from webapp + common module into the report-processor module.

I'll do my best to highlight the important details.

admin-webapp/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceServiceTest.java
```diff
@@ -123,7 +127,9 @@ public void itShouldNotCreateAStateResourceIfUserDoesNotHaveStatePermissions() {
                 .id("aUser")
                 .username("user")
                 .password("pass")
-                .permissionsById(instructionalResources(districts(1L)).getPermissionsById())
```
esotrope: UserPermissions and the associated utils were intended as a report/report-processor thing so i moved them to the processor module.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/JwtZuulFilter.java
```diff
@@ -0,0 +1,47 @@
...
+        return 0;
+    }
+
+    @Override
+    public boolean shouldFilter() {
+        if (settings.getRoutePaths().isEmpty()) {
+            return true;
+        }
+        final String requestUri = RequestContext.getCurrentContext().getRequest().getRequestURI();
+        return settings.getRoutePaths().stream().anyMatch(path -> pathMatcher.matchStart(path, requestUri));
```
esotrope: @wtritch now this filter matches routes passed in through `jwt.routes` property

common-web/src/test/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/DefaultJwtServiceTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertThat(serialized).isNotBlank();
+
+        final User deserialized = service.decode(serialized);
+
+        assertThat(deserialized).isEqualToComparingFieldByFieldRecursively(user);
+    }
+
+    private DefaultJwtService createService() {
+        final JwtSettings settings = new JwtSettings();
+        settings.setSecret("aGVsbG90aGVyZQ==");
```
esotrope: @wtritch @malaffoon one interesting detail about the jwt secret is it is expected by the JWT library to be base64 encoded at rest. If we don't want to do this we can (against their recommendation) programmatically encode it before passing it to the JWT lib

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultExamReportService.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.springframework.http.HttpHeaders.CONTENT_DISPOSITION;
+import static org.springframework.http.HttpHeaders.CONTENT_LENGTH;
+import static org.springframework.http.HttpHeaders.CONTENT_TYPE;
+
+@Service
+class DefaultExamReportService implements ExamReportService {
+
+    private final ReportGenerator generator;
+    private final ReportService service;
+    private final ReportContentService contentService;
```
esotrope: I will submit a followup PR to flatten this service and these three contained services

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
```diff
@@ -65,7 +65,7 @@ public FetchStudents(final Set<ExamQueryParamsParser> queryParamsParsers,
         this.processor = processor;
     }
 
-    @ServiceActivator(inputChannel = ReportFetchStudents)
+    @StreamListener(ReportFetchStudents)
```
esotrope: Had to change these to `@StreamListener` and the binding content-type to `application/json` for SCS to use jackson instead of kryo (now deprecated) for message serialization. (was necessary to resolve guava serialization issues with kryo)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -1,113 +1,101 @@
...
-import static org.springframework.http.HttpHeaders.CONTENT_TYPE;
-import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
-
-/**
- * This controller is responsible for providing access to Reports.
- */
-@Controller
-@RequestMapping("/api/reports")
+@RestController
 public class ReportController {
```
esotrope: Changes in this file are actually the `ReportController` being deleted and the `ExamReportController` being renamed to `ReportController`

report-processor/src/main/resources/application.yml
```diff
@@ -22,23 +22,27 @@ spring:
...
         ReportGenerateForStudents:
+          content-type: application/json
           group: default
           consumer:
             concurrency: 2
           producer:
             requiredGroups: default
 
         ReportCollate:
+          content-type: application/json
```
esotrope: Needed to tell SCS to use jackson instead of default kryo message converters

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -13,6 +13,8 @@ import { Group } from "../user/model/group.model";
 import { School } from "../user/model/school.model";
 import { Grade } from "../school-grade/grade.model";
 
+const ServiceRoute = '/report-processor';
```
esotrope: angular service class now prefixes calls with service route.
URLs end up being /api/route/endpoint

webapp/src/main/resources/application.yml
```diff
@@ -134,3 +134,13 @@ zuul:
       path: /iris/**
       url: ${app.iris.url}
       stripPrefix: true
+    report-processor:
+      path: /api/report-processor/**
+      url: http://localhost:8085
```
esotrope: controller methods on the report processor are not prefixed with /api

webapp/src/main/resources/application.yml
```diff
@@ -134,3 +134,13 @@ zuul:
...
+      stripPrefix: true
+
+jwt:
+  secret: ${rdw.reporting.jwt.secret}
+  producer: true
+  routePaths: ${zuul.routes.report-processor.path}
```
esotrope: this is how we tell the `JwtZuulFilter` to only process certain routes (a CDL is expected here)

common-web/src/test/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/DefaultJwtServiceTest.java
wtritch: I would set the non-base64 secret via system/yaml properties then base64 encode it programmatically.  That way we can {cipher} it at the configuration layer.

esotrope: ah good point. will do

webapp/src/main/resources/application.yml
wtritch: Does that play nicely with security and the actuator endpoints?  Or is that a non-issue?

webapp/src/main/resources/application.yml
wtritch: How will this change when we add our second route?

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/DefaultJwtService.java
```diff
@@ -0,0 +1,60 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.jsonwebtoken.CompressionCodecs;
+import io.jsonwebtoken.Jwts;
+import io.jsonwebtoken.SignatureAlgorithm;
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+
+import java.io.IOException;
+
+public class DefaultJwtService implements JwtService {
```
wtritch: nit: javadocs

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/JwtAuthenticationFilter.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class JwtAuthenticationFilter extends AbstractAuthenticationProcessingFilter {
```
wtritch: nit: javadocs

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/JwtZuulFilter.java
wtritch: +1

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
wtritch: huh... that's unexpected.

webapp/src/main/resources/application.yml
esotrope: `routePaths: ${zuul.routes.report-processor.path},${zuul.routes.another-route.path}`

webapp/src/main/resources/application.yml
esotrope: Don't forsee an issue but I'll test it out

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
esotrope: yeah i thought so too - according to the docsy if you just set content-type the service activator will have to manually de-serialize the message - when you use stream listener it will automagically do it using the message converter registered to the content-type

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/DefaultJwtService.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private static final String UserClaim = "_user";
+    private final String secret;
+    private final ObjectMapper objectMapper;
+
+    DefaultJwtService(final JwtSettings settings,
+                      final ObjectMapper objectMapper) {
+        this.secret = Base64.getEncoder().encodeToString(settings.getSecret().getBytes());
```
malaffoon: `getBytes` should always be given an encoding, or else the behavior will change in different JVMs, something like `settings.getSecret().getBytes(Charsets.UTF_8)`.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/JwtAuthenticationProvider.java
```diff
@@ -0,0 +1,28 @@
+package org.opentestsystem.rdw.reporting.common.web.security.jwt;
+
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.core.Authentication;
+
+
+public class JwtAuthenticationProvider implements AuthenticationProvider {
```
malaffoon: agree with general comment about move javadocs, why do these classes exist?

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/JwtWebSecurityConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public void configure(final AuthenticationManagerBuilder auth) throws Exception {
+        auth.authenticationProvider(new JwtAuthenticationProvider(service));
+    }
+
+    @Override
+    protected void configure(final HttpSecurity http) throws Exception {
+        http
+                .addFilterBefore(new JwtAuthenticationFilter("/**", authenticationManager()), AbstractPreAuthenticatedProcessingFilter.class)
```
malaffoon: Sharing concern i saw elsewhere: will actuator end-points be excluded from this filter and the requirement to be authenticated? Perhaps now that we're more Spring security compliant it is taken care of in the framework?

report-processor/docker-compose.yml
```diff
@@ -55,3 +56,8 @@ services:
       - task.remove-stale-reports.max-random-minutes=0
     extra_hosts:
       - "dwhost:$DATAWAREHOUSE_HOST"
+
+  wkhtmltopdf:
+    image: smarterbalanced/wkhtmltopdf-image:latest
+    ports:
+      - 8082:80
```
esotrope: unrelated - added wkhtmltopdf in the report processor docker compose

webapp/docker-compose.yml
```diff
@@ -46,12 +46,9 @@ services:
     ports:
       - 8080:8080
       - 8081:8008
-    volumes:
-      - /opt/rdw-reporting/config:/opt/rdw-reporting/config
     links:
       - config-server
-      - rabbitmq
-      - wkhtmltopdf
+      - report-processor-service
```
esotrope: this needed to be linked for the webapp/zuul to be able to resolve the service ID as an IP

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/jwt/JwtWebSecurityConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public void configure(final WebSecurity web) throws Exception {
+        // Spring management.security.enabled=false is overridden when @WebSecurityConfiguration is used
+        // See here: https://github.com/spring-projects/spring-boot/issues/8646
+        // Also, we cannot prefix the secured paths with /api because zuul cannot route to them:
+        // See here: https://github.com/spring-cloud/spring-cloud-netflix/issues/1069
+        for (final AbstractEndpoint endpoint : actuatorEndpoints) {
+            web.ignoring().antMatchers("/" + endpoint.getId() + "/**");
+        }
```
esotrope: some findings on `management.security.enabled` and zuul routing.

malaffoon: 👍   thanks for capturing the findings this time so we don't have to rediscover this next May.

wtritch: :+1

PR #645, opened 2017-12-08
Introduce admin instructional resource CRUD ux

wtritch:
This PR introduces a general CRUD user experience for managing instructional resources.

The list of instructional resources that the user has permissions to modify is shown on the page:

Instructional resource URL links can be created for a given assessment name, organization, and performance level.

Instructional resource URLs can be updated (they cannot be re-bound to a new assessment name, organization, or performance level.)

Instructional resource URLs can be deleted.

ir_delete
ir_update
ir_action_menu
ir_create_district
ir_create_asmt
ir_create
ir_list


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/AssessmentQuery.java
```diff
@@ -23,24 +24,34 @@ public String getLabel() {
         return label;
     }
 
+    /**
+     * @return A name query
+     */
+    public String getName() {
+        return name;
+    }
```
wtritch: I added the ability to query for Assessments by name as well as label while I was debating whether to require the user to interact with Assessments by name only. This is now unused functionality, but may be useful in the future.  I can see arguments for both leaving and removing it, thoughts?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/InstructionalResource.java
```diff
@@ -14,6 +15,7 @@
     private String resource;
     private String assessmentName;
     private String assessmentLabel;
+    private AssessmentType assessmentType;
```
wtritch: In order to convert a performance level integer into a performance level string, I needed the assessment type that this IR is bound to (ICA, IAB, SUMMATIVE)

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -109,4 +115,16 @@ private void checkAccess(final PermissionScope permissionScope, final Instructio
...
+
+    Collection<InstructionalResource> injectStateName(final Collection<InstructionalResource> resources) {
+        return resources.stream()
+                .map(this::injectStateName)
+                .collect(Collectors.toList());
+    }
+
+    private InstructionalResource injectStateName(final InstructionalResource resource) {
+        return State.equals(resource.getOrganizationType())
+                ? InstructionalResource.builder().copy(resource).organizationName(applicationProperties.getState().getName()).build()
```
malaffoon: Hmm, i have a similar issue with the embargo stuff. You have set a good example and i will remove the hacked in setter and use builder copy instead.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -57,7 +63,7 @@ public InstructionalResource create(final User user, final InstructionalResource
         checkAccess(scope, instructionalResource);
 
         repository.create(instructionalResource);
-        return repository.findOne(scope, instructionalResource);
+        return injectStateName(repository.findOne(scope, instructionalResource));
```
wtritch: I debated with using front-end translations to provide the state name, but would prefer that DevOps only have to configure it in application properties.  Thoughts?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -109,4 +115,16 @@ private void checkAccess(final PermissionScope permissionScope, final Instructio
                     " id: " + instructionalResource.getOrganizationId() + " not found");
         }
     }
+
+    Collection<InstructionalResource> injectStateName(final Collection<InstructionalResource> resources) {
```
esotrope: Don't know that i would call this "inject" since that word is strongly associated with DI

admin-webapp/src/main/webapp/src/app/instructional-resource/assessment.service.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        expect(assessments.length).toBe(2);
+        expect(assessments[0].id).toBe(1);
+        expect(assessments[0].name).toBe("name 1");
+        expect(assessments[0].grade).toBe("grade 1");
+        expect(assessments[0].type).toBe("type 1");
+        expect(assessments[0].subject).toBe("subject 1");
+        expect(assessments[0].claimCodes.length).toBe(1);
+        expect(assessments[0].claimCodes[0]).toBe("claim 1");
+        expect(assessments[1].id).toBe(2);
```
esotrope: I think you can do:
```typescript
expect(assessments[0]).toEqual({
  id: 1,
  name: "name 1"
 ...
})
```

admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -102,7 +102,8 @@ sql:
   instructionalResource:
     findAll: >-
       SELECT DISTINCT
-        a.label as asmt_label,
+        GROUP_CONCAT(DISTINCT a.label SEPARATOR ', ') as asmt_label,
```
wtritch: When finding an instructional resource, I need to display the associated Assessment Label.  Unfortunately the same assessment name may have multiple labels (although this shouldn't happen often, if ever, in production.)  This change groups all of the associated assessment labels using a comma separator.

admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -224,7 +228,7 @@ sql:
 
     create: >-
       INSERT INTO instructional_resource (asmt_name, resource, org_level, performance_level, org_id)
-        SELECT
+        SELECT DISTINCT
```
wtritch: This fixes a small bug where we were trying to insert the same IR twice when binding to multiple assessments with the same name.

admin-webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -55,12 +55,12 @@
 
   <!-- Results Table -->
   <div class="well">
-    <div class="gray-lightest pt-sm pr-sm pb-sm pl-sm my-groups-search-form-container">
+    <div class="gray-lightest pt-sm pr-sm pb-sm pl-sm table-search-form-container">
```
wtritch: more generic class name since I'm using the same CSS class in the IR UI.

admin-webapp/src/main/webapp/src/app/instructional-resource/create-instructional-resource.modal.html
```diff
@@ -0,0 +1,118 @@
...
+
+          <!-- Resource URL input -->
+          <div class="form-group">
+            <label for="resource">{{'labels.instructional-resource.cols.resource' | translate}}</label>
+            <input id="resource"
+                   name="resource"
+                   #resource="ngModel"
+                   class="form-control"
+                   [(ngModel)]="resourceUrl"
+                   pattern="https?://.+"
```
wtritch: I didn't want to write a full URL pattern matter, since they are notoriously finicky.  Instead I just require that the URL starts with http(s):// and trust that the user won't do something silly.

admin-webapp/src/main/webapp/src/app/instructional-resource/instructional-resource.component.ts
```diff
@@ -1,15 +1,125 @@
...
+  private _modalSubscriptions: Subscription[] = [];
+
+  constructor(private modalService: BsModalService,
+              private service: InstructionalResourceService,
+              private translateService: TranslateService) {
+  }
+
+  ngOnInit(): void {
+    this.service.findAll()
+      .toPromise()
```
esotrope: why do we need a promise?

admin-webapp/src/main/webapp/src/app/instructional-resource/create-instructional-resource.modal.html
```diff
@@ -0,0 +1,118 @@
...
+      <button *ngIf="!unableToCreate" class="btn btn-primary" [disabled]="!valid() || resource.invalid" (click)="create()">
+        {{ 'prompt.create' | translate }}
+      </button>
+    </span>
+</div>
+
+<ng-template #assessmentTemplate let-assessment="item" let-index="index" let-query="query">
+  <p class="mb-0">{{assessment.label}}</p>
+  <p class="h6"><span class="label label-default">{{'labels.name' | translate}}</span> {{assessment.name}}</p>
+</ng-template>
```
wtritch: auto-complete option templates

admin-webapp/src/main/webapp/src/app/instructional-resource/create-instructional-resource.modal.html
esotrope: software that relies on trust in users is doomed. Haven't looked yet but i'm hoping we have server-side validation on this as well.

admin-webapp/src/main/webapp/src/styles.less
```diff
@@ -205,3 +205,11 @@ footer {
...
   max-width: 330px;
 }
+
+.text-overflow-ellipsis {
+  text-overflow: ellipsis;
+  overflow-x: hidden;
+  overflow-y: visible;
+  white-space: nowrap;
+  padding: .2em 0 .3em;
+}
```
wtritch: I added a class for displaying labelled text with an ellipsis overflow.  Unfortunately it didn't play well with labels (it cut off the rounded background padding) so I munged in the same padding that labels use...  @aaronsleeper thoughts?

common-ngx/src/data/data.service.ts
```diff
@@ -44,6 +44,20 @@ export class DataService {
...
 
+  /**
+   * Puts data to the API server
+   *
+   * @param url the API endpoint
+   * @param body the request body
+   * @param options parameters to communicate to the API
+   * @returns {Observable<R>}
+   */
+  put(url: string, body: any, options?: RequestOptionsArgs): Observable<any> {
```
wtritch: I had to add PUT requests to our DataService

admin-webapp/src/main/webapp/src/app/instructional-resource/instructional-resource.component.ts
```diff
@@ -1,15 +1,125 @@
...
+    let modalReference: BsModalRef = this.modalService.show(UpdateInstructionalResourceModal);
+    let modal: UpdateInstructionalResourceModal = modalReference.content;
+    modal.resource = resource;
+    this._modalSubscriptions.push(modal.updated.subscribe(updatedResource => {
+      resource.resource = updatedResource.resource;
+      this.updateFilteredResources();
+    }));
+    this._modalSubscriptions.push(this.modalService.onHidden.subscribe(() => {
+      this.unsubscribe();
+    }));
```
esotrope: wonder if there is a better way to launch these modals now in the latest ngx-bootstrap

common-ngx/src/menu/popup-menu.component.ts
```diff
@@ -12,7 +12,33 @@ import { Utils } from "@sbac/rdw-reporting-common-ngx";
  */
 @Component({
```
wtritch: I moved the PopupMenu used for providing popup table row-item menus into common-ngx.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/AssessmentQuery.java
malaffoon: I'd leave it.

webapp/src/main/webapp/src/app/assessments/assessments.module.ts
```diff
@@ -35,6 +34,7 @@ import { ResultsByStudentComponent } from './results/view/results-by-student/res
 import { DistractorAnalysisComponent } from './results/view/distractor-analysis/distractor-analysis.component';
 import { ResultsByItemComponent } from "./results/view/results-by-item/results-by-item.component";
 import { ScaleScoreService } from "./results/scale-score.service";
+import { RdwMenuModule } from "@sbac/rdw-reporting-common-ngx";
```
wtritch: All of the webapp changes are fallout from moving the PopupMenu into common-ngx.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/AssessmentQuery.java
wtritch: done

admin-webapp/src/main/webapp/src/app/instructional-resource/model/organization-query.model.ts
```diff
@@ -0,0 +1,6 @@
+export class OrganizationQuery {
+  permission: string = 'INSTRUCTIONAL_RESOURCE_WRITE';
```
esotrope: how is this used?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
malaffoon: How about `copyWithStateName`

admin-webapp/src/main/webapp/src/app/instructional-resource/create-instructional-resource.modal.html
malaffoon: Oh, flashback to javascript URL validation from SRI.

admin-webapp/src/main/webapp/src/app/instructional-resource/model/organization-query.model.ts
wtritch: The organization service backend doesn't know what permission to filter on by default, since it's a common repository.  You have to tell it which permission to use when searching for organizations.  (It's also used to back the organization search in custom export which uses a different permission.)

wtritch:
@aaronsleeper See images in the PR description.  The UX is functional but doesn't feel very... slick.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
wtritch: Give me an alternative =)

esotrope: `addStateName` `setStateName` `createCopyWithStateName`

admin-webapp/src/main/webapp/src/app/instructional-resource/model/organization-query.model.ts
esotrope: hm... could there be a case where i have another role with more institutions and then i send the backend that role and it uses that instead and i get to change things i shouldn't?

admin-webapp/src/main/webapp/src/app/instructional-resource/assessment.service.spec.ts
wtritch: Nope, it has problems coercing an Assessment into an Object with properties =/
```
ERROR in src/app/instructional-resource/assessment.service.spec.ts(32,40): error TS2345: Argument of type '{ id: number; name: string; grade: string; subject: string; claimCodes: string[]; }' is not assignable to parameter of type 'Expected<Assessment>'.
  Type '{ id: number; name: string; grade: string; subject: string; claimCodes: string[]; }' is not assignable to type 'ObjectContaining<Assessment>'.
```

esotrope: Ah i guess i didn't run into that before because i used interfaces instead of classes for models

admin-webapp/src/main/webapp/src/app/instructional-resource/instructional-resource.component.ts
wtritch: doh, habit.  Changed to subscribe

admin-webapp/src/main/webapp/src/app/instructional-resource/create-instructional-resource.modal.html
wtritch: Yeah... I don't want to do that.

admin-webapp/src/main/webapp/src/app/instructional-resource/model/organization-query.model.ts
wtritch: This is for populating the organization auto-complete (user convenience.)  When actually creating the IR, I validate the user's permissions so there isn't a direct API injection security hole.  I never trust the client =P

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
wtritch: setStateName works for me, changed

admin-webapp/src/main/webapp/src/app/instructional-resource/instructional-resource.component.ts
wtritch: yeah... one of the downsides to Observable over Promises... they last forever.

wtritch: Hmm... just tested this and it doesn't look like we need to do this dance.  The modal will construct a new component instance every time the modal is shown and clean it up when the modal is hidden, which means we don't need to manually manage our subscriptions on the modal's EventEmitter.

admin-webapp/src/main/webapp/src/app/instructional-resource/create-instructional-resource.modal.html
wtritch: @esotrope you do have a good point on requiring a URL on the back-end.  I changed the InstructionalResource model's url property to be of type `URL` which will validate that the value can at least be interpreted as a URL.  (i.e. no `http://@(#*$*(@*(#%&(` values are accepted.)

admin-webapp/src/main/webapp/src/app/instructional-resource/instructional-resource.component.ts
esotrope: k - can you verify that is part of the ngx-bs changelog/tsdocs? that was not the case before

admin-webapp/src/main/webapp/src/app/instructional-resource/create-instructional-resource.modal.html
esotrope: cool. as long as angular strips xss perfectly we should be good. This part of the validation is to prevent users from doing something silly

PR #646, opened 2017-12-08
Adding Instructional Resources to Reporting UI 

dsnyder-fairway:
### Overall view of a group
screen shot 2017-12-08 at 1 11 13 pm

### Showing links (how are we supposed to grab the URL link text?)
screen shot 2017-12-08 at 5 04 43 pm

### Student score distribution (no links)
screen shot 2017-12-08 at 1 11 28 pm

### Datatable (no links)
screen shot 2017-12-08 at 1 11 35 pm

###  Changes to student history
screen shot 2017-12-08 at 1 11 47 pm
screen shot 2017-12-08 at 1 11 54 pm


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -48,26 +47,24 @@
...
                     </span>
-              <span class="icon"><i class="fa" [ngClass]="{'fa-minus':session.filter, 'fa-plus': !session.filter }"></i></span>
             </a>
           </div>
         </div>
       </div>
 
       <!-- Aggregate Scale Score -->
-      <div class="col-md-7">
+      <div class="col-md-9 right">
```
esotrope: what's right? `text-right` `pull-right`?

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -213,6 +221,21 @@ export class AssessmentResultsComponent implements OnInit {
...
+    this.instructionalResourcesService.getInstructionalResources(this.assessmentExam.assessment.id, this.assessmentProvider.getSchoolId()).subscribe((instructionalResources: InstructionalResources) => {
+      let resources = instructionalResources.getResourcesByPerformance(performanceLevel);
+      if (resources.length === 0) {
+        this.content = this.translateService.instant('labels.groups.results.assessment.no-instruct-found');
+      }
+
+      resources.forEach(resource => {
+        this.content = this.content.concat('<p>' + resource.url + '</p>');
+      });
+      // this.content = this.content.concat('<p><a [href]="' + this.sanitizer.bypassSecurityTrustUrl(resource.url) + '/>"</p>');
```
esotrope: `<a [attr.href]="url">` doesn't do it?

webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
```diff
@@ -93,13 +96,29 @@ export class ResultsByStudentComponent implements OnInit {
...
+  loadInstructionalResources(exam: Exam) {
+    this.content = '';
+    this.instructionalResourcesService.getInstructionalResources(this.assessment.id, exam.school.id).subscribe((instructionalResources: InstructionalResources) => {
+      let resources = instructionalResources.getResourcesByPerformance(exam.level);
+      if (resources.length === 0) {
+        this.content = this.translate.instant('labels.groups.results.assessment.no-instruct-found');
+      }
+
+      resources.forEach(resource => {
+        this.content = this.content.concat('<p>' + resource.url + '</p>');
```
esotrope: i think this appears about 3 times - we missing some reuse opportunity?

dsnyder-fairway: yes -- I had trouble figuring out how to extract this method correctly

webapp/src/main/webapp/src/app/shared/button/button.component.html
```diff
@@ -0,0 +1,14 @@
...
+        outsideClick="true">
+  <div *ngIf="icon">
+    <span class="sr-only">{{'labels.more-info' | translate}}</span>
+    <i class="fa {{ icon }}"></i>
+  </div>
+</button>
+
+<ng-template #buttonPopover>
+  <div [innerHTML]="content"></div>
+</ng-template>
```
esotrope: should we call this `popover-button` maybe? seems like more than just a button

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
dsnyder-fairway: chrome devtools displays
`WARNING: sanitizing HTML stripped some content (see http://g.co/ng/security#xss).` after setting that

esotrope: can you debug log the value before it is written? sounds like the value actually had something invalid

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -92,6 +95,9 @@ export class AssessmentResultsComponent implements OnInit {
   @Input()
   minimumItemDataYear: number;
 
+  @Input()
+  content: string;
```
esotrope: this is pretty confusing naming. what is this the content for? usually we don't pass around html content like this.
think you can use <ng-template #templateName> and then pass a template ref or something.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
dsnyder-fairway: screen shot 2017-12-08 at 2 04 34 pm

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
wtritch: I have some similar concerns about the [innerHTML] above.  I'd much rather provide data than HTML.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
wtritch: I think this may be related to using [innerHTML]  Have you tried putting the HTML into the .html and just providing an array of models?

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -1,77 +1,111 @@
 <!-- IAB display -->
-<div *ngIf="showIab && hasAverageScore" class="row">
+<div *ngIf="(showIab || showIcaSummative) && hasAverageScore" class="row">
```
wtritch: I think (showIab || showIcaSummative) is always *true* isn't it?

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -1,77 +1,111 @@
...
+            </div>
+          </td>
+        </tr>
+      </table>
     </div>
   </div>
 </div>
+<ng-template #buttonPopover>
+  <div [innerHTML]="content"></div>
+</ng-template>
```
wtritch: Since the button template is duplicated here, I'd be tempted to create an InstructionalResourcePopoverComponent and just do:
```
<ng-template #buttonPopover>
  <instructional-resource-popover [resources]="instructionalResources" />
</ng-template>
```

webapp/src/main/webapp/src/app/assessments/results/instructional-resources.mapper.ts
```diff
@@ -0,0 +1,41 @@
...
+const organizationLevels: string[] = [
+  "System",
+  "State",
+  "DistrictGroup",
+  "District",
+  "SchoolGroup"
+];
+
+@Injectable()
+export class InstructionalResourcesMapper {
```
wtritch: Since this mapper is only used by one service, I'd be tempted to move the mapping methods into the service itself and get rid of this injectable mapper class

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
dsnyder-fairway: it should be. This was done to get rid of some duplication with setting the `average-scale-score-display`. I'll remove the tautology

webapp/src/main/webapp/src/app/shared/button/information-button.component.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * This component is responsible for displaying a label with
+ * an information popover icon.
+ */
+@Component({
+  selector: 'info-button,[info-button]',
+  templateUrl: './information-button.component.html'
+})
+export class InformationButtonComponent extends ButtonComponent {
```
wtritch: This breakout confuses me a little bit, it doesn't seem to be much of a code-saver?

webapp/src/main/webapp/src/app/shared/button/button.component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * This component is responsible for displaying a button with
+ * a specified popover icon.
+ */
+@Component({
+  selector: 'sb-button,[sb-button]',
+  templateUrl: './button.component.html'
+})
+export class ButtonComponent {
```
wtritch: It doesn't look like this is being used anywhere?

webapp/src/main/webapp/src/app/shared/button/information-button.component.ts
dsnyder-fairway: Yeah, to some extent this is an artifact of a previous PR I made for this where I had a BookButtonComponent extending the ButtonComponent as well. I was unsure if I should have undone all of my changes to these files (was previous `InformationLinkComponent`)

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
dsnyder-fairway: done

webapp/src/main/webapp/src/app/shared/button/information-button.component.ts
dsnyder-fairway: done

webapp/src/main/webapp/src/app/shared/button/button.component.ts
dsnyder-fairway: removed

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
dsnyder-fairway: corrected (`pull-right`)

esotrope: i think with columns its not advised to use `float` (`pull-right`) you probably want to put a div in the column with `pull-right`

dsnyder-fairway: I think it was that there was another `row` appearing within a `row`. I removed the `pull-right` and the additional `row` in c646013

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
dsnyder-fairway: done

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
dsnyder-fairway: removed in 9b98a76. Links are displaying correctly now

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
dsnyder-fairway: this has been resolved in 9b98a76

dsnyder-fairway:
URLs are now displaying correctly 
screen shot 2017-12-08 at 5 04 43 pm

dsnyder-fairway:
@wtritch cceed62 adds changes for display names

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -377,6 +377,18 @@
...
+      "link": {
+        "System": "Smarter Balanced Instructional Resource",
+        "State": " Instructional Resource",
+        "DistrictGroup": "{{organizationName}} Instructional Resource",
+        "District": "{{organizationName}} Instructional Resource",
+        "SchoolGroup": "{{organizationName}} Instructional Resource"
+      }
+    },
+    "states": {
+      "CA": "California"
```
esotrope: @wtritch we already have this info in admin config repo settings right?

webapp/src/main/webapp/src/app/assessments/popover/instructional-resource-popover.component.html
```diff
@@ -1,7 +1,9 @@
 <div *ngIf="resources !== undefined && resources.length > 0; else noResources">
   <p class="popover-link"
      *ngFor="let instructionalResource of resources">
-    <a [attr.href]="instructionalResource.url" target="_blank">{{ instructionalResource.url }}</a>
+    <a *ngIf="instructionalResource.organizationLevel === 'System'" [attr.href]="instructionalResource.url" target="_blank">{{ 'labels.instructional-resources.link.System' | translate }}</a>
+    <a *ngIf="instructionalResource.organizationLevel === 'State'" [attr.href]="instructionalResource.url" target="_blank">{{ ('labels.states.' + instructionalResource.stateCode) | translate }}{{ 'labels.instructional-resources.link.State' | translate }}</a>
+    <a *ngIf="instructionalResource.organizationLevel !== 'System' && instructionalResource.organizationLevel !== 'State'" [attr.href]="instructionalResource.url" target="_blank">{{ ('labels.instructional-resources.link.' + instructionalResource.organizationName) | translate:instructionalResource }}</a>
```
esotrope: i think i'd be tempted to have one line for the anchor and pre-process this stuff before sending it to the view

dsnyder-fairway: 👍 agreed. I extracted this into a `linkText` method

webapp/src/main/webapp/src/app/assessments/popover/instructional-resource-popover.component.html
```diff
@@ -1,9 +1,7 @@
 <div *ngIf="resources !== undefined && resources.length > 0; else noResources">
   <p class="popover-link"
-     *ngFor="let instructionalResource of resources">
-    <a *ngIf="instructionalResource.organizationLevel === 'System'" [attr.href]="instructionalResource.url" target="_blank">{{ 'labels.instructional-resources.link.System' | translate }}</a>
-    <a *ngIf="instructionalResource.organizationLevel === 'State'" [attr.href]="instructionalResource.url" target="_blank">{{ ('labels.states.' + instructionalResource.stateCode) | translate }}{{ 'labels.instructional-resources.link.State' | translate }}</a>
-    <a *ngIf="instructionalResource.organizationLevel !== 'System' && instructionalResource.organizationLevel !== 'State'" [attr.href]="instructionalResource.url" target="_blank">{{ ('labels.instructional-resources.link.' + instructionalResource.organizationName) | translate:instructionalResource }}</a>
+     *ngFor="let instructionalResource of resources; let i = index">
+    <a [attr.href]="instructionalResource.url" target="_blank">{{ linkText(i) }}</a>
```
esotrope: does this have to be `[attr.href]` anymore? i think it can be `<a href="{{instructionalResource.url}}"`

esotrope: also - why not pass the `instructionalResource` variable so you can do away with the lookup by index?

webapp/src/main/webapp/src/assets/i18n/en.json
dsnyder-fairway: @wtritch heads up - I decided on using `stateCode` here instead of `stateName` here as we could desire translation for state names (for instance, NY - Nueva York)

webapp/src/main/webapp/src/app/assessments/popover/instructional-resource-popover.component.html
dsnyder-fairway: done

jtreuting:
looking at the UI screenshots, I see the links listed but not as described in the mockups on the requirements page.  It should be saying something like "California Resources", "District A Resources", etc.

dsnyder-fairway:
@jtreuting sorry those changes were included in PRs since cceed62. Here is an example
screen shot 2017-12-12 at 9 37 54 am

wtritch:
hmm... you should never have two "California" resources displayed in one popup

webapp/src/main/java/org/opentestsystem/rdw/reporting/instructionalresource/InstructionalResource.java
```diff
@@ -31,6 +32,13 @@ public String getOrganizationName() {
         return organizationName;
     }
 
+    /**
+     * @return The state code this resource is for.
+     */
+    public String getStateCode() {
+        return stateCode;
+    }
```
wtritch: I wouldn't add this.  I'd saturate the organizationName property with the "app.state.name" configuration value IFF the organizationLevel == State. (At the service layer.)

webapp/src/main/java/org/opentestsystem/rdw/reporting/instructionalresource/JdbcInstructionalResourceRepository.java
```diff
@@ -18,6 +18,9 @@
     @Value("${sql.instructionalResource.findAllByAssessmentId}")
     private String findAllByAssessmentId;
 
+    @Value("${app.state.code}")
+    private String stateCode;
```
wtritch: I recommend injecting the ApplicationProperties instance rather than an @Value injection

dsnyder-fairway:
@wtritch - I believe this is due to some SQL I ran to generate this example:
`insert into reporting.instructional_resource (asmt_name, resource, performance_level, org_level) (select name,  'https://portal.smarterbalanced.org/library/en/digital-library-connections-high-school-brief-writes.docx', 0, 'System' from reporting.asmt);` and then modified performance level. Probably not realistic

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -1,77 +1,111 @@
...
+            <span class="pull-right">
+              {{ examLevelEnum + performanceLevel.id.toString() | translate }}
+              <button
+                class="btn btn-borderless btn-xs icon-only pull-right"
+                angulartics2On="click"
+                angularticsEvent="InstructionalResource"
+                angularticsCategory="AssessmentResults"
+                [disabled]="false"
+                [popover]="buttonPopover"
+                popoverTitle="{{ 'labels.groups.results.assessment.instruct-button-prefix' | translate }}{{ examLevelEnum + performanceLevel.id.toString() | translate }}"
```
wtritch: I don't recommend breaking a single statement into two translations, it assumes the sentance/statement structure.  Recommend consolidating into a single translation (a little duplication is ok, rather than assuming the value will always be divided into left and right segments)

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -2,6 +2,12 @@ import { Component, Input } from "@angular/core";
 import { AssessmentExam } from "../model/assessment-exam.model";
 import { ExamStatistics, ExamStatisticsLevel } from "../model/exam-statistics.model";
 import { ScaleScoreService } from "./scale-score.service";
+import { InstructionalResource, InstructionalResources } from "../model/instructional-resources.model";
+import { InstructionalResourcesService } from "./instructional-resources.service";
+import { GroupAssessmentService } from "../../groups/results/group-assessment.service";
+
+const icaColors = [ 'maroon', 'gray-darkest', 'green-dark', 'blue-dark' ];
+const iabColors = [ 'blue-dark', 'blue-dark aqua', 'aqua' ];
```
wtritch: Why is this here?  Don't we define these colors elsewhere?  @esotrope ?  Aren't we using an indexed color provider service?

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -63,7 +74,39 @@ export class AverageScaleScoreComponent {
       return this.statistics.levels;
   }
 
-  getLevelPercent(num: number): number {
-    return this.levelPercents[num];
+  levelSum(): number {
+    const values = this.statistics.levels.map(l => l.value);
+    return values.reduce((p, c) => p + c);
+  }
```
wtritch: I feel like there are a lot of orthogonal changes creeping into this PR.  I don't recommend trying to break everything out now, but in the future try to keep a single PR to a single task.  (Sometimes even multiple PRs per task in order to keep things focused/small/readable. =)

webapp/src/main/webapp/src/app/assessments/results/instructional-resources.mapper.ts
wtritch: ...Still recommend removing this mapper =)

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
dsnyder-fairway: These were found [here](https://github.com/SmarterApp/RDW_Reporting/pull/646/files/fa67212a8bb66aeb05ee9664c047b56bb1f28ac8#diff-4061477d62dc8d21f8086960477b2b99). Someone suggested I refactor this with a for-loop

webapp/src/main/webapp/src/assets/i18n/en.json
wtritch: We'll only support a single state per one installation.  I'd argue that it's easier to provide the state code and name together as configuration properties rather than as translation overrides in production when this system is deployed by multiple states.  I'd also argue that the state name should probably not be translated into spanish, vietnamese, etc as a pronoun.

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
esotrope: i think we have a ColorService among the webapp angular components - yes. we can see if that satisfies the desired results. What are these colors for btw?

webapp/src/main/webapp/src/assets/i18n/en.json
wtritch: @esotrope yup, although the property naming should be consistent to allow setting it via a shared application.yml in the configuration repo: "app.state.code and app.state.name"

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -366,10 +367,27 @@
...
       "students": "Students",
-      "student-score-distribution": "Student Score Distribution"
+      "student-score-distribution": {
+        "title": "Student Score Distribution"
+      }
+    },
+    "instructional-resources": {
+      "link": {
+        "System": "Smarter Balanced Instructional Resource",
+        "State": " Instructional Resource",
```
wtritch: This is what I was referring to above, assuming that a statement will have the same structure in different languages has caused me problems on projects in the past.  I'd recommend changing this to:
`"{{organizationName}} Instructional Resource"`

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
esotrope: also, just in general let's stay away from single letter variable names - they don't communicate what is what here. The only exception to this is `i` in a for loop i think

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
dsnyder-fairway: These colors are for providing different score distribution colors

wtritch: My only real goal here is that we don't want these colors defined in multiple ways in multiple locations.  Because colors are something everybody has an opinion on and I could see them changing =P

dsnyder-fairway:
I'll be re-opening this once I've wrapped all of this up

PR #649, opened 2017-12-11
DWR-1205 Added gender drill down by ethnicity support

agorina:
This is just for @wtritch, so he could have more data samples. I will focus on adding proper tests (as promised).

### DIFF
--- a/olap/src/main/resources/reports.sql.yml
+++ b/olap/src/main/resources/reports.sql.yml
@@ -100,5 +100,18 @@ sqlbuilder:
 
                 join:
                      JOIN student s ON fe.student_id = s.id
+
                 groupBy: >-
                      s.gender_id
+
+            ethnicity:
+              clauses:
+                select: >-
+                    'ethnicity' AS dimension,
+                     se.ethnicity_id AS dimension_id
+
+                join:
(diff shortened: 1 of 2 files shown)


olap/src/main/resources/reports.sql.yml
```diff
@@ -100,5 +100,18 @@ sqlbuilder:
...
                      s.gender_id
+
+            ethnicity:
+              clauses:
+                select: >-
+                    'ethnicity' AS dimension,
+                     se.ethnicity_id AS dimension_id
+
+                join:
+                     JOIN student_ethnicity se ON se.student_id = fe.student_id
```
agorina: I realised that this has to be a `LEFT JOIN` because we do not enforce Ethnicity to be a required field. 

This leads me to a question: should we instead have 'Unspecified' and enter it during migrate?

@malaffoon , @jtreuting - what do you think?

malaffoon: Hmm. It seems like this should be a LEFT JOIN which results in a NULL ethnicity value. That would be consistent with other optional fields like migrant status. Then the UI needs to know what to do with NULL.

PR #650, opened 2017-12-12
DWR-1205 Added avg std err and a few more tests.

agorina:
More tests to come.


olap/src/main/resources/reports.sql.yml
```diff
@@ -5,6 +5,7 @@ sqlbuilder:
           select: >-
               count(*) AS count,
               round(avg(scale_score)) AS score,
+              round(sqrt(sum(power(scale_score_std_err, 2)) / count(*))) as std_err,
```
agorina: @malaffoon, is this right?

olap/src/main/resources/reports.sql.yml
```diff
@@ -3,8 +3,8 @@ sqlbuilder:
       customAggregate:
         clauses:
           select: >-
-              count(*) AS count,
```
agorina: This is not used. @wtritch had a good point that we could sum level1-4 counts to get the total.

olap/src/main/resources/application.yml
```diff
@@ -28,9 +28,9 @@ app:
   datamart:
      olap_datasource:
         url: "\
-           jdbc:redshift://${spring.olap_datasource.url-server:localhost:5432}/dev\
-           ?ApplicationName=rdw-migrate-olap\
-           &currentSchema=${spring.olap_datasource.url-schema:reporting_olap}
+           jdbc:redshift://${app.datamart.olap_datasource.url-server:localhost:5439}/dev\
+           ?ApplicationName=rdw-reporting-olap\
+           &currentSchema=${app.datamart.olap_datasource.url-schema:reporting_olap}
            "
```
agorina: Unrelated clean up. Since we do not have it configured in the CI or Configuration service, it really does not matter, but still.

olap/src/test/resources/application-test.yml
```diff
@@ -2,10 +2,10 @@ app:
...
         url: "\
-              jdbc:redshift://${spring.olap_datasource.url-server:localhost:5432}/dev\
-              ?ApplicationName=rdw-migrate-olap\
-              &currentSchema=${spring.olap_datasource.url-schema:reporting_olap}
+            jdbc:redshift://${app.datamart.olap_datasource.url-server:localhost:5439}/dev\
+            ?ApplicationName=rdw-reporting-olap\
+            &currentSchema=${app.datamart.olap_datasource.url-schema:reporting_olap_rst}
            "
-        username: root
+        username: awsrst
```
agorina: I have updated the environments wiki page with the password for this user.

olap/src/main/resources/reports.sql.yml
```diff
@@ -3,8 +3,8 @@ sqlbuilder:
...
               round(avg(scale_score)) AS score,
+              round(sqrt(sum(power(scale_score_std_err, 2)) / count(*))) as std_err,
```
agorina: @malaffoon, is this right?

malaffoon: Yes.

olap/src/test/resources/application-test.yml
```diff
@@ -2,10 +2,10 @@ app:
   datamart:
      olap_datasource:
         url: "\
-              jdbc:redshift://${spring.olap_datasource.url-server:localhost:5432}/dev\
-              ?ApplicationName=rdw-migrate-olap\
-              &currentSchema=${spring.olap_datasource.url-schema:reporting_olap}
+            jdbc:redshift://${app.datamart.olap_datasource.url-server:localhost:5439}/dev\
```
malaffoon: redshift changed its port??

agorina: I think it was always 5439 ( do you use a different one? )

olap/src/main/resources/reports.sql.yml
agorina: Thank you.

olap/src/test/resources/application-test.yml
malaffoon: You're quite right. I didn't know the port off the top of my head and i was surprised to see it change.

PR #651, opened 2017-12-12
Task/add ir management logging

wtritch:
I got some good questions from QA w.r.t. auditing IR changes.  This PR adds update/delete logging when managing Instructional Resources.

### DIFF
--- a/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
+++ b/admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
@@ -9,6 +9,8 @@ import org.opentestsystem.rdw.reporting.common.model.Organization;
 import org.opentestsystem.rdw.reporting.common.repository.OrganizationRepository;
 import org.opentestsystem.rdw.reporting.common.web.security.User;
 import org.opentestsystem.rdw.security.PermissionScope;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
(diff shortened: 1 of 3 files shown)


admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -26,6 +26,7 @@ sql:
         FROM asmt a
         WHERE (:label_query IS NULL OR a.label LIKE concat('%', :label_query, '%'))
           AND (:name_query IS NULL OR a.name LIKE concat('%', :name_query, '%'))
+        ORDER BY a.school_year DESC
```
wtritch: There is an issue with using this generic Assessment searching for unique Assessment.name values.  We have to de-dupe on the front-end, which can cause issues w.r.t. finding a specific count of unique assessment names when combined with a smallish response limit ~20.  It's less likely to have multiple assessments with the same name within the same school year, but there is no guarantee that a limit of 20 responses will always return unique assessment names.  I see this as a small fudge factor that shouldn't affect usability.  Thoughts?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -68,33 +71,51 @@ public InstructionalResource create(final User user, final InstructionalResource
...
         return setStateName(updated);
     }
 
     @Override
     @Transactional(transactionManager = "reportingTxManager")
-    public void delete(final User user, final InstructionalResource instructionalResource) {
+    public void delete(final User user, final InstructionalResource resource) {
         final PermissionScope scope = toScope(user);
 
-        checkAccess(scope, instructionalResource);
```
jtreuting: did you mean to get rid of this check?

wtritch: Yup, I replaced it with a findOne which performs the same security checks but gives me a saturated instance for logging.

jtreuting: nope, sounds good.  just trying to understand the difference.  thanks!

PR #653, opened 2017-12-13
adds default jwt password

esotrope:
makes configuring `RDW_REPORTING_JWT_SECRET` env var optional and negates necessity of configuring `jwt.secret` in config repo

### DIFF
--- a/report-processor/src/main/resources/application.yml
+++ b/report-processor/src/main/resources/application.yml
@@ -108,5 +108,5 @@ task:
 #    max-random-minutes: 20
 
 jwt:
-  secret: ${rdw.reporting.jwt.secret}
+  secret: ${rdw.reporting.jwt.secret:secret}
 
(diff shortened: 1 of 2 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -133,11 +133,11 @@ zuul:
       stripPrefix: true
     report-processor:
       path: /api/report-processor/**
-#      url: http://localhost:8085
+      url: http://localhost:8085
```
esotrope: accidentally committed this commented out - now fixed

report-processor/src/main/resources/application.yml
```diff
@@ -108,5 +108,5 @@ task:
 #    max-random-minutes: 20
 
 jwt:
-  secret: ${rdw.reporting.jwt.secret}
+  secret: ${rdw.reporting.jwt.secret:secret}
```
malaffoon: All those suggestions and `secret` is the best we could come up with??  ;)

esotrope: i thought about being cute but then i thought i might get frowned at x j

PR #654, opened 2017-12-13
Embargo API

malaffoon:


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
```diff
@@ -1,24 +0,0 @@
-package org.opentestsystem.rdw.admin.repository;
```
malaffoon: This isn't really a delete, i renamed it and made it generally useful

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/EmbargoService.java
```diff
@@ -0,0 +1,38 @@
...
+import org.opentestsystem.rdw.reporting.common.model.OrganizationType;
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+import org.springframework.security.access.prepost.PreAuthorize;
+
+import java.util.Collection;
+
+/**
+ * Service to manipulate embargo settings.
+ */
+@PreAuthorize("hasAuthority('PERM_EMBARGO_WRITE')")
```
malaffoon: I didn't really pay attention to security since it is changing; i'll probably want help from @esotrope and/or @wtritch when the time comes to integrate with the UI.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/Embargo.java
```diff
@@ -0,0 +1,190 @@
...
+            return this;
+        }
+
+        public Builder examCount(final Subject subject, final long count) {
+            this.examCounts.put(subject, count);
+            return this;
+        }
+
+        public Builder examCounts(final Map<Subject, Long> counts) {
+            if (counts != null && !counts.isEmpty()) {
```
esotrope: i think inside `.putAll` will be a fast exit for `isEmpty` otherwise the collection lib isn't doing its job

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Collection;
+import java.util.EnumSet;
+
+/**
+ * End-points for administering {@link Embargo}s.
+ *
+ * TODO - should PUT be changed to accept an Embargo payload instead?
+ * TODO   basically, i've postponed the creating of a fake object to the backend
+ *
+ * TODO - should the service accept the raw values instead of a partly populated object?
```
malaffoon: I left these in to get opinions on them. I'm okay with how i did things but these are options, i'll point them out below.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
```diff
@@ -0,0 +1,90 @@
...
+     * </pre>
+     *
+     * @param user user with EMBARGO_WRITE permission for given organization
+     * @param organizationType organization type, "State" or "District"
+     * @param organizationId organization id; use 0 for "State"
+     * @param scope organization scope, "Individual" or "Aggregate"
+     * @param payload setting, "true" or "false"
+     */
+    @PutMapping(value = "/embargos/{organizationType}/{organizationId}/{scope}")
+    public void put(@AuthenticationPrincipal final User user,
```
malaffoon: Option 1 - instead of this it could be `PUT /embargos` and the payload is a (very) partially populated Embargo object.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
```diff
@@ -0,0 +1,90 @@
...
+        switch (EmbargoScope.caseInsensitiveValue(scope)) {
+            case Individual:
+                builder.individualEnabled(enabled);
+                break;
+            case Aggregate:
+                builder.aggregateEnabled(enabled);
+                break;
+        }
+
+        service.save(user, builder.build());
```
malaffoon: Option - service takes the raw fields (type, id, flags) instead of the partially populated Embargo object.

admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -29,44 +29,96 @@ sql:
         ORDER BY a.school_year DESC
         LIMIT :limit
 
+  embargo:
+    # query to find all embargo records given a school year and a list of (user's) districts
+    # this includes counts of summative exams grouped by each org so it has performance implications
+    findAllBySchoolYear: >-
+      SELECT st.*, se.individual, se.aggregate, cnt.subject_id, cnt.count
```
malaffoon: @agorina any performance concerns here?

admin-webapp/src/main/resources/application.sql.yml
```diff
@@ -29,44 +29,96 @@ sql:
...
+      SET individual = :individual, aggregate = :aggregate, updated_by = :updated_by
+      WHERE district_id = :district_id AND school_year = :school_year
+
   group:
     findAllDistinctSchoolYears: >-
-      select distinct sg.school_year
-      from student_group sg
-        join school sc on sc.id = sg.school_id
-      where sg.deleted = 0
+      SELECT distinct sg.school_year
```
malaffoon: changes from here down are just consistent casing of SQL

admin-webapp/src/main/resources/application.yml
```diff
@@ -121,12 +121,13 @@ permissionservice:
...
 
 app:
+  # google analytics configuration
+  analytics:
+    trackingId: ${rdw.admin.analytics.trackingId:}
+  home-url: /
+  school-year: 2018
   state:
     code: CA
     name: California
```
malaffoon: just alphabetized properties

common-web/src/test/java/org/opentestsystem/rdw/reporting/common/web/security/auth/DefaultAuthorizationServiceTest.java
```diff
@@ -99,10 +99,10 @@ public void getPermissionsShouldCorrectlyDerivePermissionsFromGrants() throws Ex
...
                         new Permission("a", PermissionScope.STATEWIDE),
-                        new Permission("b", new PermissionScope(null, ImmutableSet.of(1L), null, null)),
-                        new Permission("c", new PermissionScope(null, null, ImmutableSet.of(1L), ImmutableSet.of(1L))),
-                        new Permission("d", new PermissionScope(null, ImmutableSet.of(1L), null, ImmutableSet.of(1L))),
-                        new Permission("f", new PermissionScope(null, ImmutableSet.of(1L), null, ImmutableSet.of(1L)))
+                        new Permission("b", PermissionScope.builder().addDistrictId(1).build()),
+                        new Permission("c", PermissionScope.builder().addSchoolGroupId(1).addSchoolId(1).build()),
+                        new Permission("d", PermissionScope.builder().addDistrictId(1).addSchoolId(1).build()),
+                        new Permission("f", PermissionScope.builder().addDistrictId(1).addSchoolId(1).build())
                 );
```
malaffoon: I hid the constructors in the latest version of common so had to switch this in a few places

common-web/src/test/java/org/opentestsystem/rdw/reporting/common/web/security/auth/DefaultAuthorizationServiceTest.java
```diff
@@ -158,7 +158,7 @@ public void getPermissionsShouldProduceStateLevelPermissionScopeForUnionWithStat
 
         assertThat(authorizationService.getPermissions(grants))
                 .usingRecursiveFieldByFieldElementComparator()
-                .containsExactlyInAnyOrder(new Permission("a", PermissionScope.STATEWIDE));
+                .containsExactlyInAnyOrder(new Permission("a", PermissionScope.builder().statewide(true).addDistrictId(1).build()));
     }
```
malaffoon: Note: this is a real change in behavior: the statewide flag doesn't "hide" the district values in the permission scope

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/JdbcOrganizationRepository.java
```diff
@@ -0,0 +1,125 @@
...
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
+
+/**
+ * JDBC-backed implementation of an OrganizationRepository.
+ *
+ * This is a base class that the reporting/warehouse-specific implementation can extend.
+ * Currently, there is no difference between the two schemas so everything is here. If things
+ * change, it may be necessary to factor out code.
+ */
+public abstract class JdbcOrganizationRepository implements OrganizationRepository {
```
malaffoon: Common implementation for reporting and warehouse organization repository. For now, i am reusing the queries but doing it this way makes it less impactful if we have to introduce datastore-specific queries.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -0,0 +1,163 @@
...
+                throw new NoSuchElementException();
+            }
+        } else if (embargo.getOrganizationType() == District) {
+            final Set<Long> districtIds = getDistrictIdsForUser(user);
+            if (!districtIds.contains(embargo.getOrganizationId())) {
+                throw new NoSuchElementException();
+            }
+        } else {
+            throw new IllegalArgumentException("Invalid embargo organization type");
+        }
```
esotrope: a takeaway here is that a user cannot be both state & district-scoped for a perm. The state scope will takeover.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * @param user the user
+     * @return list of ids of district that the user has embargo write permission
+     */
+    private Set<Long> getDistrictIdsForUser(final User user) {
+        final PermissionScope permissionScope = user.getPermissionsById().get(EmbargoWrite).getScope();
+
+        // We don't want to get districts inherited because of state permissions (because, for
+        // embargo, state-level permissions do NOT imply permissions for districts) so copy the
+        // permissions with statewide disabled.
```
esotrope: 👍

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/Embargo.java
malaffoon: You're right but whenever i do a null check i usually stick in the empty check, even if it isn't strictly necessary.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
```diff
@@ -0,0 +1,90 @@
...
+                .organizationType(type)
+                .organizationId(organizationId <= 0 ? null : organizationId);
+        switch (EmbargoScope.caseInsensitiveValue(scope)) {
+            case Individual:
+                builder.individualEnabled(enabled);
+                break;
+            case Aggregate:
+                builder.aggregateEnabled(enabled);
+                break;
+        }
```
esotrope: All this building logic probably should go in a arg resolver no?

admin-webapp/src/main/resources/application.yml
```diff
@@ -121,12 +121,13 @@ permissionservice:
   endpoint: ${rdw.permission.service}
 
 app:
+  # google analytics configuration
+  analytics:
+    trackingId: ${rdw.admin.analytics.trackingId:}
+  home-url: /
+  school-year: 2018
```
esotrope: this default will look funny one day

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -0,0 +1,163 @@
...
+        final Collection<Organization> districts = organizationRepository.findAll(
+                permissionScope.copy().statewide(false).build(),
+                OrganizationQuery.builder().types(EnumSet.of(OrganizationType.District)).build());
```
esotrope: this is interesting - are we checking just to make sure that the district IDs are in our DB?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
malaffoon: Not sure why you're saying that. A user can be granted both state and district for a perm (that is part of the change i made in common, to allow that).

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
malaffoon: This will query for all districts that the user can see based on granted district-group and district permissions.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
malaffoon: Hmm. Can i do a resolver that gets the path variables?

Good point about too much logic in here. Perhaps i could make a helper util or move some of the logic into the builder itself.

admin-webapp/src/main/resources/application.yml
malaffoon: Yup, as will "CA" and "California".

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
esotrope: ah that's right - district group 👍

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
esotrope: the permission map is a 1:1 permission to permission scope map.
So if a user has `EMBARGO_WRITE` with scope: `{statewide: true, districtIds: [1,2,3]}` this code block will only look at the statewide part

malaffoon: I really want to make sure i understand this cause i'm still confused. If the *embargo* is state-level then we ask the permission scope for its statewide flag. If the *embargo* is district-level then we ask the permission scope for its district ids.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
wtritch: I went with a partially saturated object, if we want consistency.  I still regret not adding a primary key...

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/Embargo.java
```diff
@@ -0,0 +1,190 @@
...
+
+    public static class Builder {
+        private OrganizationType organizationType;
+        private Long organizationId;
+        private String organizationName;
+        private Integer schoolYear;
+        private Boolean individualEnabled;
+        private Boolean aggregateEnabled;
+        private Boolean readOnly;
+        private Map<Subject, Long> examCounts = newHashMap();
```
wtritch: nit: could be final

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -0,0 +1,151 @@
...
+    private MapSqlParameterSource toSource(final Embargo embargo, final String username) {
+        return new MapSqlParameterSource()
+                .addValue("district_id", embargo.getOrganizationId())
+                .addValue("school_year", embargo.getSchoolYear())
+                .addValue("individual", embargo.isIndividualEnabled())
+                .addValue("aggregate", embargo.isAggregateEnabled())
+                .addValue("updated_by", username);
+    }
+
+    private static ResultSetExtractor<Embargo> EmbargoExtractor = rs -> rs.next() ? Embargo.builder()
```
wtritch: slick

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.stream.Collectors;
+
+import static org.opentestsystem.rdw.admin.security.AdminPermission.EmbargoWrite;
+import static org.opentestsystem.rdw.reporting.common.model.OrganizationType.District;
+import static org.opentestsystem.rdw.reporting.common.model.OrganizationType.State;
+
+/**
+ * Repository-backed implementation of EmbargoService.
+ * <p>
+ * TODO - for state, does the name need to be set or can UI deal with it?<br/>
```
wtritch: I'd prefer to set the state name on the backend so it's controlled via configuration properties rather than force our next client to do translation overrides on the front-end.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -0,0 +1,163 @@
...
+        // this may throw
+        checkAccess(user, embargo);
+
+        final int currentSchoolYear = applicationProperties.getSchoolYear();
+        final Embargo existing = (embargo.getOrganizationType() == State) ?
+                repository.findOneBySchoolYearAndState(currentSchoolYear) :
+                repository.findOneBySchoolYearAndDistrictId(currentSchoolYear, embargo.getOrganizationId());
+
+        if (existing != null
+         && existing.isIndividualEnabled() == embargo.isIndividualEnabled()
```
wtritch: does a submitted value of "null" mean no-change?

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
malaffoon: Yeah, this was an old TODO, it is getting state name from properties.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/model/Embargo.java
malaffoon: fixed

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
malaffoon: Correct.

PR #659, opened 2017-12-14
Task/reporting service

esotrope:
This is just step 1 of the refactor
This copies `webapp/*` (minus `webapp/src/main/webapp`) to a new `reporting-service` module
Next step is to add the zuul gateway

Bad news is it is not possibly to see the history of a file after rename on github.
You have to run `git log --follow [filename]` which is not supported through the github ui
(someone made a chrome extension for it though: https://stackoverflow.com/questions/5646174/how-to-make-github-follow-directory-history-after-renames)

git i guess tracks file content changes (not the files) so the history can actually be seen on the file using this command on either file copy.


wtritch:
I'm not going to review a 269-file copy, but I'll trust you.

reporting-service/build.gradle
```diff
@@ -0,0 +1,39 @@
+apply plugin: 'org.springframework.boot'
+
+createDockerfile {
+    ext.java_opts = ["-Xmx768m"]
```
wtritch: Note to future us, once we have the API and the front-end broken out one or both of these can probably be reduced.

malaffoon:
I'm with Will on this ... reviewing a huge set of copied files doesn't work great so i'll trust you. 

One question: i thought we already had a `report-processor` module. Is this new module going to be the mostly-the-webapp thing?

esotrope:
@malaffoon doh - correcting description

PR #664, opened 2017-12-14
Modify instructional resource creation dialog to match update and del…

wtritch:
…ete dialogs

image


esotrope:
Are the values next to Assessment Name and Organization Type filled in when you select something?

wtritch:
@esotrope yessir

PR #668, opened 2017-12-14
Add an aggregate reporting UI development harness to allow simultaneo…

wtritch:
…us request and result UX development.

This PR just adds a route and sub-route for `/aggregate-reports/reports` with two placeholder components that are ready to be saturated with a UI for both requesting report data from the backend and displaying the report data in the UI.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.module.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.module.ts
@@ -1,0 +1,26 @@
+import { Angulartics2Module } from "angulartics2";
+import { NgModule } from "@angular/core";
+import { CommonModule } from "../shared/common.module";
+import { DataTableModule } from "primeng/primeng";
+import { AggregateReportsComponent } from "./aggregate-reports.component";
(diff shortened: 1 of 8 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.module.ts
```diff
@@ -0,0 +1,26 @@
...
+import { AggregateReportsComponent } from "./aggregate-reports.component";
+import { AggregateReportsResultsComponent } from "./results/aggregate-reports-results.component";
+
+@NgModule({
+  declarations: [
+    AggregateReportsComponent,
+    AggregateReportsResultsComponent
+  ],
+  imports: [
+    Angulartics2Module.forChild(),
```
esotrope: are we tracking to a different app ID for admin, aggregate & reporting?

wtritch: hm... good question.  I doubt it since we can get granularity through labelling, right @jtreuting?

jtreuting: no we aren't and I don't think we need to.

PR #669, opened 2017-12-15
prunes webapp of unneeded components and files

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/ApplicationConfiguration.java
```diff
@@ -1,50 +1,38 @@
...
 @Import({
         ResourceLoaderConfiguration.class,
         EndpointAutoConfiguration.class,
         StatusConfiguration.class,
         YamlPropertiesConfigurator.class,
         DataSourceConfiguration.class,
-        MessageSourceConfiguration.class,
         SecurityConfiguration.class,
+        JdbcOrganizationRepository.class, // needed by SecurityConfiguration
+        SecurityParameterProvider.class, // needed by JdbcOrganizationRepository
```
esotrope: these three above deps can probably be joined and i will try to do that in a followup PR

webapp/src/main/java/org/opentestsystem/rdw/reporting/Application.java
```diff
@@ -2,16 +2,9 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.cache.annotation.EnableCaching;
-import org.springframework.cloud.aws.autoconfigure.cache.ElastiCacheAutoConfiguration;
-import org.springframework.cloud.aws.autoconfigure.context.ContextStackAutoConfiguration;
 import org.springframework.context.annotation.PropertySource;
 
-@SpringBootApplication(exclude = {
-        ElastiCacheAutoConfiguration.class,
```
wtritch: We don't need to exclude this when running in AWS?

esotrope: Ah because it wants to use elasticache instead of embedded (guava) cache? I'll add it back if so

esotrope: Here is the autoconfig:
```java
@Configuration
@Import(ContextCredentialsAutoConfiguration.class)
@EnableElastiCache
@ConditionalOnClass(AmazonElastiCache.class)
@ConditionalOnAwsCloudEnvironment
public class ElastiCacheAutoConfiguration {
}
```
it should only trigger when `org.springframework.cloud:spring-cloud-starter-aws` is on the classpath so i dont think we need it since i removed it from the classpath and it is not in common, common-web

PR #672, opened 2017-12-16
redirect to access-denied when users do not have permissions

dsnyder-fairway:


common-ngx/src/security/authorization.can-activate.ts
```diff
@@ -1,19 +1,27 @@
...
+              private authorizationService: AuthorizationService,
+              private router: Router) {
   }
 
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> | Promise<boolean> | boolean {
-    return this.authorizationService.hasAnyPermission(route.data[ 'permissions' ]);
+    return this.authorizationService.hasAnyPermission(route.data[ 'permissions' ]).map((permission) => {
+        if (!permission) {
+          this.router.navigate([ this.accessDeniedRoute ]);
+        }
```
esotrope: is this different than the `routing-authorization.can-activate.ts` thing in common?

dsnyder-fairway: yeah, that injectable checks that a user has a non-empty set of permissions. This is used to check to see if a user has a permission belonging to whatever the permissions are for a path

esotrope: gotcha cool. my only question is - what did this do before when we didn't route?? x D

dsnyder-fairway: It would show the navbar and the footer, and that was about it

dsnyder-fairway: @esotrope it looks like this change totally busted reporting 😞

PR #673, opened 2017-12-18
Revert "redirect to access-denied when users do not have permissions"

dsnyder-fairway:
Reverts SmarterApp/RDW_Reporting#672

### DIFF
--- a/common-ngx/src/security/authorization.can-activate.ts
+++ b/common-ngx/src/security/authorization.can-activate.ts
@@ -1,8 +1,7 @@
-import { ActivatedRouteSnapshot, CanActivate, Router, RouterStateSnapshot } from "@angular/router";
-import { Inject, Injectable } from "@angular/core";
+import { ActivatedRouteSnapshot, CanActivate, RouterStateSnapshot } from "@angular/router";
+import { Injectable } from "@angular/core";
 import { Observable } from "rxjs/Observable";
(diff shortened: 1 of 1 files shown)


common-ngx/src/security/authorization.can-activate.ts
```diff
@@ -1,27 +1,19 @@
...
 
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> | Promise<boolean> | boolean {
-    return this.authorizationService.hasAnyPermission(route.data[ 'permissions' ]).map((permission) => {
-        if (!permission) {
-          this.router.navigate([ this.accessDeniedRoute ]);
-        }
-        return permission;
-      }
-    );
+    return this.authorizationService.hasAnyPermission(route.data[ 'permissions' ]);
```
esotrope: could the issue have been that we didn't use `route.snapshot.data`?

esotrope: Also, i think only admin has a route for access-denied - could be wrong though

dsnyder-fairway: `route` is already a `snapshot`. Above we have `route: ActivatedRouteSnapshot`. Access-denied appears in both admin and reporting (`webapp/src/main/webapp/src/app/error/access-denied/access-denied.component.ts`)

PR #676, opened 2017-12-18
 safely compare strings to avoid localeCompare on undefined

dsnyder-fairway:


dsnyder-fairway:
small change. Merging

webapp/src/main/webapp/src/app/assessments/filters/select-assessments/select-assessments.component.ts
```diff
@@ -43,7 +45,7 @@ export class SelectAssessmentsComponent {
     let assessmentsByGrade = [];
 
     let grades: string[] = _.uniq(this._assessments.map(assessment => assessment.grade))
-      .sort((a, b) => a.localeCompare(b));
+      .sort(this._comparator);
```
wtritch: hm... I'd be tempted to filter out nulls rather than account for them in sorting.  What happens when the front-end gets a null grade in this array?

esotrope: struggling to see why assessment.grade and subjects would be null ever

wtritch: ...agreed.  Seems unexpected.

wtritch: Oh.  I bet this is also due to the reporting -> reporting-service change?  We were getting back non-json and so deserializing gobbledygook as json may have lead to non-assessment instances?

wtritch: Can this issue be reproduced without this change?

dsnyder-fairway: @esotrope @wtritch, sorry, this change appears to have not addressed the issue. I will be reverting these changes soon

PR #678, opened 2017-12-19
DWR-1205 Update report query to support 'all schools for the given district'.

agorina:


olap/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
```diff
@@ -92,18 +90,62 @@ public void itShouldFindByQueryWithGender() {
...
+                .includeState(true)
+                .includeAllDistricts(true)
+                .dimensions(of(Overall, Gender, Ethnicity))
+                .districtIds(ImmutableMap.of(-9, false, -8, false, -7, false))
+                .schoolIds(of(-9, -8)).build()).size()).isEqualTo(30);
+
+
+        // (1 state + 3 districts + 2 schools(in district 9) + 2 schools) = 8
+        //NOTE: in this case we get duplicate school -9, but this should not happen with the UI controlling the selection
+        assertThat(reportRepository.findByQuery(query
```
agorina: Please let me know if  my note above it not accurate.

@jtreuting , should I create a JIRA to track it?

olap/src/main/java/org/opentestsystem/rdw/olap/model/ReportQuery.java
```diff
@@ -104,7 +110,7 @@ public Builder copy(final ReportQuery query) {
             includeState = query.isIncludeState();
             includeAllDistricts = query.isIncludeAllDistricts();
             schoolIds = query.getSchoolIds();
-            districtIds = query.getDistrictIds();
+            districtIds = query.districtIds;
```
jtreuting: this is standing out.  i see why you need it, but maybe it should be something other than just `districtIds` since it is a map and then the `getDistrictIds` makes sense as just getting the ids and i think you wouldn't then need to access the private variable directly.

olap/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
jtreuting: are you saying that a district is included with all it's schools, but then also the user selects one of those schools in the district directly?

olap/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
```diff
@@ -92,18 +90,62 @@ public void itShouldFindByQueryWithGender() {
...
+                .copy()
+                .includeState(true)
+                .includeAllDistricts(true)
```
jtreuting: i'm confused again :) . does this include all districts within the state?  if so, why are 3 districts added also separately?

olap/src/main/java/org/opentestsystem/rdw/olap/model/ReportQuery.java
agorina: You are absolutely right. I missed it. Cleaning it up.

olap/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
agorina: yes, that is what I am saying.

olap/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
agorina: I just wanted to test that it will work, that is all. Most likely nobody will do this, but still...

olap/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
agorina: https://jira.fairwaytech.com/browse/DWR-1254

olap/src/main/java/org/opentestsystem/rdw/olap/model/ReportQuery.java
```diff
@@ -1,9 +1,12 @@
 package org.opentestsystem.rdw.olap.model;
 
 import javax.validation.constraints.NotNull;
+import java.util.Map;
 import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.stream.Collectors.toSet;
+import static org.hibernate.validator.internal.util.CollectionHelper.newHashMap;
 import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
```
malaffoon: These aren't the correct newHashSet/Map ... we want the ones from guava.

agorina: Oops, thank you!

olap/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
```diff
@@ -30,32 +43,45 @@ public Object resolveArgument(final MethodParameter parameter,
...
-        final String[] asmtGrades = webRequest.getParameterValues("asmtGrades");
-        final String[] schoolIds = webRequest.getParameterValues("schoolIds");
-        final String[] districtIds = webRequest.getParameterValues("districtIds");
-        final String[] dimensions = webRequest.getParameterValues("dimensions");
 
-        checkArgument(subjectIds.length > 0, "report query is missing subjects");
-        checkArgument(schoolYears.length > 0, "report query is missing schoolYears");
-        checkArgument(asmtGrades.length > 0, "report query is missing asmtGrades");
+        final String state = webRequest.getParameter("state");
+        checkArgument(state == null || state.length() == 2, "report query is missing state");
```
malaffoon: This is confusing unless i'm misreading it: if (state == null) this doesn't throw so a missing state does *not* complain that the state is missing? I would've guess this should be `checkArgument(state != null and state.length() == 2, "report query has missing or invalid state");`

olap/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
```diff
@@ -30,32 +43,45 @@ public Object resolveArgument(final MethodParameter parameter,
...
         return ReportQuery.builder()
-                .includeState(Boolean.valueOf(webRequest.getParameter("includeState")))
-                .includeAllDistricts(Boolean.valueOf(webRequest.getParameter("includeAllDistricts")))
-                .subjectIds(Stream.of(subjectIds).map(Integer::valueOf).filter(Objects::nonNull).collect(toSet()))
-                .schoolYears(Stream.of(schoolYears).map(Integer::valueOf).filter(Objects::nonNull).collect(toSet()))
-                .asmtGrades(Stream.of(asmtGrades).map(Integer::valueOf).filter(Objects::nonNull).collect(toSet()))
-                .schoolIds(schoolIds == null ? null : Stream.of(schoolIds).map(Integer::valueOf).filter(Objects::nonNull).collect(toSet()))
-                .districtIds(districtIds == null ? null : Stream.of(districtIds).map(Integer::valueOf).filter(Objects::nonNull).collect(toSet()))
-                .dimensions(dimensions == null ? null : Stream.of(dimensions).map(type -> {
+                .includeState(state != null && state.toUpperCase().startsWith("T"))
```
malaffoon: nit: i think it would be more efficient to have a helper method to check that a character is either a `t` or `T` rather than do all this uppercasing.

olap/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
agorina: Missing state is not an error. I need to change the message.

olap/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -62,6 +62,21 @@ sqlbuilder:
                 where: >-
                     d.id in (:district_ids)
 
+            allSchoolsInDistricts:
+              clauses:
+                select: >-
+                    'school' AS organization_type,
```
malaffoon: `School`?

olap/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
agorina: Ok, I can change it.

olap/src/main/resources/custom-aggregate-report.sql.yml
agorina: Changing. 
But, why do we have a `caseInsensitiveValue` method if we are always case sensitive.

malaffoon: Well, just cause it works doesn't mean it's correct ;)

PR #680, opened 2017-12-19
DWR-1215 Add support for Summative assessments

agorina:


olap/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -57,11 +59,13 @@ public JdbcCustomAggregateReportRowRepository(final NamedParameterJdbcTemplate j
     @Override
     public List<ReportRow> findByQuery(@NotNull final ReportQuery query) {
         checkIsValid(query);
+        checkArgument(!AssessmentType.IAB.equals(query.getAssessmentType()), "custom aggregate report for IAB is not supported yet");
```
agorina: I think that if we put all assessments into one fact table, this may work for IABs without any changes. (just a thought to consider)

jtreuting: that could be nice

olap/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
```diff
@@ -27,7 +28,7 @@
  * District: districts=100F,200T - optional, a comma separated list of district ids and 'True/False' for including all children (schools).
  * School Ids: schoolIds=99,300 - optional, a comma separated list of school ids
  * <p>
- * Example URL: ...?subjectIds=1,2&schoolYears=2017,2018&asmtGrades=4,5&state=FF&districts=15F,17F&dimensions=gender,ethnicity&schoolIds=265,266
+ * Example URL: ...?assessmentType=ICA&subjectIds=1,2&schoolYears=2017,2018&asmtGrades=4,5&state=FF&districts=15F,17F&dimensions=gender,ethnicity&schoolIds=265,266
```
malaffoon: It is odd that we're using AssessmentType but SubjectId. Should/could we use Subject and use enum values instead of ids?

agorina: The reason we use subject ids is to support different subjects. We have talked that maybe subjects should be driven by the database and we could have a subject code. But this feels to me like a part of a different story.

olap/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: On the other hand, @malaffoon had a good point that it could be a simple SQL template change to use a different fact table ( assuming I am not overlooking anything).

olap/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
esotrope: 👍 preemptively positioning ourselves to use codes/ids instead of enum will save us later

PR #681, opened 2017-12-20
redirect to access-denied when users do not have permissions

dsnyder-fairway:


common-ngx/src/security/authorization.can-activate.ts
```diff
@@ -1,19 +1,31 @@
...
+  constructor(@Inject(AccessDeniedRoute)
+              private accessDeniedRoute: string,
+              private authorizationService: AuthorizationService,
+              private router: Router) {
   }
 
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> | Promise<boolean> | boolean {
-    return this.authorizationService.hasAnyPermission(route.data[ 'permissions' ]);
+    return this.authorizationService.hasAnyPermission(route.data[ 'permissions' ]).map((permission) => {
+        if (!permission && route.data[ 'denyAccess' ]) {
```
dsnyder-fairway: requiring denyAccess = true as this would cause issues on the reporting webapp otherwise

esotrope: so later this would go away when reporting has an access denied route?

dsnyder-fairway: This is to prevent users from getting access-denied on the reporting webapp. I noticed that almost all (or maybe all of the users) were getting routed to access-denied on reporting pages

PR #684, opened 2017-12-20
fixes service route for organization level exam export

esotrope:
had the wrong route for org export


webapp/src/main/webapp/src/app/organization-export/organization-export.service.ts
```diff
@@ -7,7 +7,7 @@ import { OrganizationExportOptions } from "./organization-export-options";
 import { OrganizationExportNamingService } from "./organization-export-naming.service";
 import { OrganizationGroupingService } from "./organization-grouping.service";
 
-const ServiceRoute = '/reporting-service';
+const ServiceRoute = '/report-processor';
```
malaffoon: potato ... potahto  ;)

PR #687, opened 2017-12-21
handle schools which are not properly configured or do not contain as…

dsnyder-fairway:
…sessments

screen shot 2017-12-20 at 5 00 17 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
+++ b/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
@@ -46,11 +46,11 @@ export class SchoolResultsComponent implements OnInit {
   /**
    * The currently selected school
    |  */
-  get currentSchool() {
+  get currentSchool(): School {
     return this._currentSchool;
   }
 
-  set currentSchool(value) {
+  set currentSchool(value: School) {
     this._currentSchool = value;
 
     if (!Utils.isNullOrUndefined(value)) {
@@ -63,7 +63,7 @@ export class SchoolResultsComponent implements OnInit {
    * The currently selected grade.
    * @returns {Grade}
    */
-  get currentGrade() {
+  get currentGrade(): Grade {
     return this._currentGrade;
   }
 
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -159,15 +159,21 @@ export class SchoolResultsComponent implements OnInit {
...
               : grade;
+            this.updateRoute('School');
           } else {
             this.currentGrade = undefined;
+            this.assessmentExams = [];
+            this.gradesAreUnavailable = true;
           }
-          this.updateRoute('School');
         });
+    } else {
```
dsnyder-fairway: this block will be entered if we were going to hit our service with undefined. I included this in the else block above as well since we should still handle empty results

jtreuting: instead of having 2 identical blocks, might be cleaner to have `this.currentGrade = undefined;` as the first thing in `schoolSelectChanged` method.  then you could remove the 2 else blocks and replace with one block like:

```
if (this.currentGrade === undefined) {
    this.assessmentExams = [];
    this.gradesAreUnavailable = true;
}
```

just a thought.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -211,7 +217,7 @@ export class SchoolResultsComponent implements OnInit {
     });
   }
 
-  private getAnalyticsEventLabel(changedFilter: string): any {
+  private getAnalyticsEventLabel(changedFilter: string): string | number {
```
jtreuting: i just learned about Union Types, didn't know this was a thing in TypeScript

dsnyder-fairway:
still need to fix up a few things which aren't behaving right. Closing for now to prevent merges

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
dsnyder-fairway: doing this gave me some trouble as we are setting the `currentGrade` within our subscription. This caused two calls to our assessments.component `set assessmentExams`. One with an empty array (on the first pass) and (when we were returning assessment exams) another call with the assessment exams (in our subscription). This caused the "no results" view to flash whenever changing schools

PR #692, opened 2017-12-22
Task/admin service gateway

esotrope:
screen shot 2017-12-21 at 4 19 02 pm

screen shot 2017-12-21 at 4 19 56 pm

This will go side by side with config repo changes to come.

Done in this package:
1. Moved `admin-webapp/src/main/webapp/src/app/*` to `webapp/src/main/webapp/src/app/admin`
2. Added admin routes in `admin-webapp/.../app.routes.ts` under root 'admin' route
3. Added zuul gateway config to admin-webapp
4. Removed admin-webapp controllers' `/api` prefixes
5. Switched admin-webapp to JwtWebSecurityConfiguration
6. Diff'd and selectively copied entries in `admin-webapp/.../en.json` into `webapp/.../en.json`, `admin-webapp/.../styles.less` into `webapp/.../styles.less`


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/security/WebSecurityConfiguration.java
```diff
@@ -1,60 +1,53 @@
...
-import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
-import org.springframework.security.web.authentication.AuthenticationFailureHandler;
-import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
-import org.springframework.security.web.authentication.SavedRequestAwareAuthenticationSuccessHandler;
-import org.springframework.security.web.authentication.SimpleUrlAuthenticationFailureHandler;
+import org.springframework.context.annotation.Import;
 
 @Configuration
 @ConditionalOnProperty(name = "app.test-mode", havingValue = "false", matchIfMissing = true)
-public class WebSecurityConfiguration extends SamlWebSecurityConfiguration {
```
esotrope: bye bye saml hello jwt

admin-webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1,123 +1,207 @@
 {
+  "application": {
```
esotrope: used formatter to alphabetically sort everything for easier comparison with webapp's en.json (this file will go away during a followup pruning sweep)

webapp/src/main/webapp/package.json
```diff
@@ -35,6 +35,7 @@
     "json-unflat": "^1.0.1",
     "lodash": "^4.17.4",
     "mathjs": "^3.15.0",
+    "ng2-file-upload": "^1.3.0",
```
esotrope: the only npm dep that admin-webapp and webapp didnt

webapp/src/main/webapp/src/app/admin/admin.module.ts
```diff
@@ -0,0 +1,34 @@
...
+    GroupImportModule,
+    GroupsModule,
+    InstructionalResourceModule,
+    EmbargoModule,
+    FileFormatModule,
+    HomeComponent
+  ],
+  providers: []
+})
+export class AdminModule {
```
esotrope: new module that just bunches the admin app modules together and exposes them all together. this way the webapp's app.module.ts only need import one module to get all of the admin stuff

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -25,6 +25,124 @@ import { RoutingAuthorizationCanActivate } from "@sbac/rdw-reporting-common-ngx/
...
+import { EmbargoResolve } from "./admin/embargo/embargo.resolve";
+import { ImportHistoryComponent } from "./admin/groups/import/history/import-history.component";
+import { FileFormatComponent } from "./admin/groups/import/fileformat/file-format.component";
+import { ImportHistoryResolve } from "./admin/groups/import/history/import-history.resolve";
+import { GroupImportComponent } from "./admin/groups/import/group-import.component";
+import { GroupImportDeactivateGuard } from "./admin/groups/import/group-import.deactivate";
+import { GroupsComponent } from "./admin/groups/groups.component";
+import { HomeComponent as AdminHomeComponent } from "./admin/home/home.component";
+
+const adminRoute = {
```
esotrope: new route def.

webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -14,7 +14,7 @@ <h1 class="blue-dark h1 mb-md">{{'labels.homepage.title' | translate}}</h1>
 
         <!-- Student Groups Admin Link -->
         <a *hasAnyPermission="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE', 'EMBARGO_WRITE']"
-           href="{{user.configuration.adminWebappUrl}}" class="btn btn-xs btn-info ml-xs"><i class="fa fa-gear"></i> {{'labels.groups.admin-button' | translate}}</a>
```
esotrope: later we can prune out this url

esotrope:
Helps to review these huge packages if you collapse the file diff you are done looking at (and all the green/red-only files)

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/AssessmentController.java
```diff
@@ -14,7 +14,6 @@
  * This controller is responsible for providing Assessments.
```
malaffoon: Just to clarify, where does this (and nearby) files live? Not under `admin-webapp` anymore, right? This would be under something like `admin-service`?

esotrope: right now they live under `admin-webapp`. The idea is that a followup PR will prune `admin-webapp` and rename it `admin-service`

webapp/docker-compose.yml
```diff
@@ -56,6 +56,25 @@ services:
       - CONFIG_SERVICE_URL=http://config-server:8888
       # need this until we have it on config server
       - zuul.routes.reporting-service.url=http://reporting-service:8080
+      - zuul.routes.admin-service.url=http://admin-service:8080
+
+    extra_hosts:
+      - "dwhost:$DATAWAREHOUSE_HOST"
+
+  admin-service:
+    image: smarterbalanced/rdw-reporting-admin-webapp
```
malaffoon: Similar question: isn't this `rdw-reporting-admin-service`?

esotrope: it will be when we rename admin-webapp to admin-service

wtritch:
I'd strongly recommend waiting to merge this until after we have the smarterbalanced dockerhub repo ready to go.

jtreuting:
@esotrope, @wtritch just let me know what dockerhub repos are needed and I can reach out to SB to get those setup.

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/security/WebSecurityConfiguration.java
jtreuting: feels good huh

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceController.java
```diff
@@ -23,7 +23,6 @@
  * This controller is responsible for interacting with instructional resources
  */
 @RestController
-@RequestMapping("/api")
```
jtreuting: seems like we might want to consolidate how we do this.  Some places we put the base route here and in some we don't.  For example, `@RequestMapping("/embargoes")`.

Feels like this should be `@RequestMapping("/instructional-resources")` snd same with groups above.

And maybe that is a follow-up PR to clean them up and get consistent, not sure.  There appear to be both ways of doing it throughout the controllers.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -74,6 +192,7 @@ export const routes: Routes = [
     resolve: { user: UserResolve, translateComplete: TranslateResolve },
     children: [
       { path: '', pathMatch: 'full', component: HomeComponent },
+      adminRoute,
```
jtreuting: should this be a const above since it's only used once?

webapp/src/main/webapp/src/assets/template/groups-template.csv
```diff
@@ -0,0 +1 @@
+group_name,school_natural_id,school_year,subject_code,student_ssid,group_user_login
```
jtreuting: @malaffoon feels like we had some strangeness with this file and the line ending.  Not sure if there is a newline here or not, but just mentioning as we probably want to make sure this is "correct" (meaning what we want it to be) and that git hasn't auto-changed it to something else.

malaffoon:
The docker hub repos have all been created and are ready.  fyi, @wtritch @esotrope

esotrope:
@malaffoon neeto

webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
```diff
@@ -0,0 +1,116 @@
...
+    this._event = event;
+    this.content = this.createContent(event);
+  }
+
+  accept() {
+    const { embargo, scope, value, toggle } = this.event;
+
+    // apply desired embargo setting via the API
+    this.embargoService.update(embargo, scope, value)
+      .subscribe(() => {
```
wtritch: Do you want to handle errors here outside of the global error handler?  What can go wrong?

esotrope: good point - will adjust

PR #695, opened 2017-12-22
adds migate warehouse as task dependency in build script for IT

esotrope:


build.gradle
```diff
@@ -192,7 +192,7 @@ subprojects {
     }
 
     //The omission of the ':' in manualIT is on purpose - we need the relative path to manualIT
-    task IT(type: Test, dependsOn:[":cleanReportingTest", ":migrateReportingTest"]) {
+    task IT(type: Test, dependsOn:[":cleanReportingTest", ":migrateReportingTest", ":cleanWarehouseTest", ":migrateWarehouseTest"]) {
```
wtritch: I might scope that to admin rather than the entire app

esotrope: hm can we fine-tune later?

PR #696, opened 2017-12-22
reverting changes

dsnyder-fairway:
reverting changes. See https://github.com/SmarterApp/SBAC-Global-UI-Kit/pull/18 for the fix.

Thanks @jtreuting


dsnyder-fairway:
since this depends on newest verion of SBAC Global UI Kit (currently in a PR) we should not merge this until that has been merged

PR #697, opened 2017-12-22
fix UI issues with groups/school pages

dsnyder-fairway:
This should resolve two things:

1. unable to completely render pages on first calls to groups/schools where no data is found

2. Select  assessments still visible when no data is found


webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -117,17 +117,19 @@ <h2 class="blue-dark h3 mt-lg mb-md">
...
 
 <!-- Results Container -->
-<div *ngFor="let assessment of assessmentExams; let i = index" class="groups-results-wrapper">
-  <assessment-results [assessmentExam]="assessment"
-                      [showValuesAsPercent]="showValuesAsPercent"
-                      [filterBy]="clientFilterBy"
-                      [allowFilterBySessions]="allowFilterBySessions"
-                      [assessmentProvider]="assessmentProvider"
-                      [minimumItemDataYear]="minimumItemDataYear"></assessment-results>
+<div *ngIf="!hideAssessments">
```
esotrope: been meaning to try it out but in the angular docs it suggests using a `<ng-container>` element for these kind of logical structure-only divs. if we use it it wont actually render a div i guess

PR #700, opened 2017-12-25
handle sb-typeahead select events passing to click listeners

dsnyder-fairway:
sb-typeahead is only used in for handling schoolSelectChanged(events). Somehow typescript allows for Events to be passed to a method expected school as a parameter.


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -428,6 +428,15 @@
       },
       "students": "Students"
     },
+    "instructional-resources": {
```
dsnyder-fairway: looks like these strings were recently dropped from en.json

dsnyder-fairway:
merging for now. I will address feedback in github comments or future pull requests

PR #702, opened 2017-12-26
delete all reports greater than a certain duration, so long as we do …

dsnyder-fairway:
…not get an error deleting these from s3

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
@@ -90,12 +90,12 @@ public class ArchiveBackedReportContentService implements ReportContentService {
             final String location = uri.toASCIIString();
             try {
                 archiveService.delete(location);
-                return report;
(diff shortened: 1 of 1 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -90,12 +90,12 @@ private Report deleteContent(final Report report) {
...
             try {
                 archiveService.delete(location);
-                return report;
             } catch (final RuntimeException e) {
                 logger.warn("failed to delete location {}", location, e);
+                return null;
             }
         }
-        return null;
+        return report;
```
dsnyder-fairway: issue here is that the reports weren't deleting because they did not have a uri, due to being in progress or failing to generate

malaffoon: There should be a change to a test somewhere to go along with this. Looking in ArchiveBackedReportContentServiceTest ... there is a test in there that could be enhanced to check this behavior.

PR #704, opened 2017-12-26
Task/move webapp only common web components into webapp

esotrope:
image

image

moved all the components only used in webapp (saml, redis & more) from common/common-web into webapp. this allowed for some cleanup in `@Configuration` imports and `application.yml` properties.


admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
```diff
@@ -28,7 +28,7 @@
 import java.util.stream.Collectors;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.apache.commons.lang.StringUtils.isNotBlank;
```
malaffoon: lang3 is actually better, why the switch? We really should establish which utility collections we want to use and stick with them. I think guava and lang3 make a good combo.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/MessageSourceConfiguration.java
```diff
@@ -14,6 +14,8 @@
 
 import static org.springframework.util.StringUtils.isEmpty;
```
malaffoon: (just pointing out yet another utility collection)

admin-webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceService.java
esotrope: So the change here was to satisfy the import path after changing from:
`commons-io:commons-io` to `org.apache.commons:commons-io`
So should we be using the `commons-io:commons-io` package? This was only in one of our apps and not in the parent build.gradle - hence the change

PR #710, opened 2017-12-28
Task/move common ngx into webapp

esotrope:
A good PR to look at commit by commit.
This moves all common-ngx into webapp/src/main/webapp/src/app/shared.

A followup PR will remove the `common-ngx` module 😢 😭


webapp/src/main/webapp/src/app/shared/menu/popup-menu-action.model.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * This model represents a menu action.
+ */
+import { Observable } from "rxjs/Observable";
+import "rxjs/add/observable/of";
```
esotrope: somehow i needed to add these to all files using Observable.of()
What can happen is the file build order can change and will succeed or fail depending on whether or not this line has been processed in a previous file or not.
From what I read, we actually only need to import these "rxjs/add" additions once and in one central place - still looking into where that should be in our app.

esotrope:
Merging now since reviewers are out.
Feel free to comment on the merged package so i can address any questions etc.

PR #711, opened 2017-12-28
enables docker for olap module and maps its name to rdw-reporting-agg…

esotrope:
…regate-service

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -11,7 +11,11 @@ rootProject.name = "rdw-reporting"
 
 rootProject.children.each {
     if (it.name == "reporting-service") {
+        // TODO? later rename module to service?
         it.name = "${rootProject.name}-service"
+    } else if (it.name == "olap") {
+        // TODO later rename module to aggregate-service
+        it.name = "${rootProject.name}-aggregate-service"
     } else {
         it.name = "${rootProject.name}-${it.name}"
     }
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -101,6 +101,10 @@ zuul:
       path: /api/admin-service/**
       url: http://localhost:8280
       stripPrefix: true
+    admin-service:
+      path: /api/aggregate-service/**
+      url: http://localhost:8380
+      stripPrefix: true
     report-processor:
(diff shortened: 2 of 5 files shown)


settings.gradle
```diff
@@ -11,7 +11,11 @@ rootProject.name = "rdw-reporting"
 
 rootProject.children.each {
     if (it.name == "reporting-service") {
+        // TODO? later rename module to service?
         it.name = "${rootProject.name}-service"
```
esotrope: should we rename reporting-service to just service?

settings.gradle
```diff
@@ -11,7 +11,11 @@ rootProject.name = "rdw-reporting"
...
+    } else if (it.name == "olap") {
+        // TODO later rename module to aggregate-service
+        it.name = "${rootProject.name}-aggregate-service"
```
esotrope: @agorina let me know when it would be convenient to rename the `olap` module to `aggregate-service`

webapp/src/main/resources/application.yml
```diff
@@ -101,6 +101,10 @@ zuul:
       path: /api/admin-service/**
       url: http://localhost:8280
       stripPrefix: true
+    admin-service:
+      path: /api/aggregate-service/**
+      url: http://localhost:8380
+      stripPrefix: true
```
esotrope: doesn't break anything to preemptively add this so i thought I would. It won't be used however until the config repo yml and k8s pods are in place

settings.gradle
agorina: You can do it at any time. I am not actively working on it now. Thank you for asking.

webapp/src/main/resources/application.yml
```diff
@@ -101,6 +101,10 @@ zuul:
       path: /api/admin-service/**
       url: http://localhost:8280
       stripPrefix: true
+    admin-service:
+      path: /api/aggregate-service/**
```
malaffoon: maybe it's the diff view but this looks like admin-service is being mapped to aggregate-service ... that doesn't seem right?

esotrope: fixed in another package.

PR #714, opened 2017-12-28
Task/rename olap module to aggregate service

esotrope:


settings.gradle
```diff
@@ -1,20 +1,17 @@
+include 'admin-service'
+include 'aggregate-service'
 include 'common'
 include 'common-web'
-include 'webapp'
 include 'report-processor'
 include 'reporting-service'
-include 'admin-service'
-include 'olap'
+include 'webapp'
```
esotrope: changed olap to aggregate-service and alphabetized

jtreuting:
don't the java packages need to be renamed as well?  looks like it was just file renaming at this point

esotrope:
yeah i think i might put that in another PR.
`org.opentestsystem.rdw.admin` -> `.reporting.admin`
`org.opentestsystem.rdw.olap` -> `.rdw.reporting.aggregate`? `aggregates`?

We also need a jira to address adding a qualifying package to the webapp. right now it is under `...rdw.reporting` without a subpackage. This will take a bit of work though.

PR #717, opened 2017-12-29
alphabetizes keys

esotrope:
helps to look commit by commit.
1. added `labels.type|name`
2. sorted

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -333,46 +333,46 @@
     "aggregate-reports": {
       "heading": "Aggregate Reports",
       "results": {
-        "heading": "Aggregate Report Results",
-        "empty-message": "No Results Found",
-        "performance-group-prompt": "Show performance as",
-        "performance-grouped": "Grouped",
-        "performance-separated": "Separate",
         "cols": {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -875,6 +866,15 @@
         "title": "Update Instructional Resource"
       }
     },
+    "instructional-resources": {
+      "link": {
+        "District": "{{organizationName}} Instructional Resource",
+        "DistrictGroup": "{{organizationName}} Instructional Resource",
+        "SchoolGroup": "{{organizationName}} Instructional Resource",
+        "State": "{{organizationName}} Instructional Resource",
+        "System": "Smarter Balanced Instructional Resource"
```
jtreuting: Is there a reason these are duplicated?  is it admin vs. user facing?

esotrope: my guess is that admin had `instructional-resourceS` and webapp had `instructional-resource` my bet is the one under `instructional-resource` can be removed. @dsnyder-fairway ?

dsnyder-fairway: yeah I agree. It doesn't look like `instructional-resource` is used anywhere

PR #718, opened 2017-12-29
Cleanup hard-coded state name in aggregate report results table

wtritch:


webapp/src/main/webapp/src/app/aggregate-report/results/mock-aggregate-reports.service.ts
```diff
@@ -17,37 +19,39 @@ import { HttpClient } from "@angular/common/http";
...
     let uiModel = new AggregateReportItem();
     uiModel.assessmentId = apiModel.assessment.id;
     uiModel.gradeId = apiModel.assessment.gradeId;
     uiModel.subjectId = apiModel.assessment.subjectId;
     uiModel.schoolYear = apiModel.examSchoolYear;
     uiModel.organizationType = apiModel.organization.type;
-    //TODO: "California" should come from the back-end api populated via config-properties
     uiModel.organizationName = (uiModel.organizationType == "State")
-      ? "California"
+      ? user.configuration.stateName
```
jtreuting: my initial reaction is that it seems like this should be populated by the server and this just always uses the organization.name provided.  is there a reason you were thinking client side for this?

agorina: I have to admit that initially I mislead Will by saying that it is hard to do in SQL without reading each field (which I did not want to do). I found a solution that would not require it: since our SQL is in the properties file and State is in the properties file too, I can just use a placeholder substitution in there.  

I think Will had some other concerns about it, but I will let him comment on it :)

esotrope: that's cool. should we be doing that for allow_transfer_access?

agorina: If allow_transfer_access is driven by the properties, then I do not see why we should not. (Sorry, I am not very familiar with it).

wtritch: I’m torn between a standardized payload with a consisten organizationName property and needless data saturation with the state name coming back on multiple rows... but I suppose the extra data isn’t all that much.  I could go either way.

jtreuting: also consider that we are doing aggregate export on the server side, so we would need the same logic on the server and client if we don't populate the name.

wtritch: hm... that's an excellent point.  And if we want to do an output-to-S3 solution like we do with the custom export we'd need it at the SQL layer as well.  I'm going to abandon this PR in favor of @agorina doing my work for me =)

PR #719, opened 2017-12-29
Added state name to the aggregate report result.

agorina:


aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -2,7 +2,7 @@ sql:
   # a tokenized version of the SQL query for the custom aggregate reports
   # the @{template.XXXX} tokens are replaced with the appropriate queries at run time
   customAggregate: >-
-      CASE WHEN dim.organization_id = -1 THEN null else dim.organization_name END,
+      CASE WHEN dim.organization_id = -1 THEN '${app.state.name}' else dim.organization_name END,
```
agorina: There is a view that has this logic for the state:
```sql
 SELECT
    -1          AS id,
    'State'     AS name,
    'State'     AS organization_type,
...
````

PR #721, opened 2018-01-02
DWR-1219 Enhance ‘achievement’ report to support drill down by boolean fields.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/DimensionType.java
```diff
@@ -22,8 +22,4 @@ public static DimensionType caseInsensitiveValue(final String name) {
         }
         return null;
     }
-
-    public String toLowerCase() {
```
agorina: Unrelated, remove since I do not think we need it.

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -41,13 +41,11 @@ sqlbuilder:
             state:
               clauses:
                 select: >-
-                     'State' AS organization_type,
```
agorina: This is a left over from the implementation before we decided to fill in the gaps. Cleaned it up.

jtreuting: Do we need this for sorting?  If we are doing a search with State and a district and schools in it.  The sorting should be state first, then a district and it's children schools.  So is organization_type needed for that type of sort?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -104,7 +104,7 @@ public JdbcCustomAggregateReportRowRepository(final NamedParameterJdbcTemplate j
         final List<SelectQuery> queries = newArrayList();
 
         for (final DimensionType dimension : query.getDimensionTypes()) {
-            final String dimensionName = dimension.toLowerCase();
+            final String dimensionName = dimension.toString();
```
malaffoon: i *think* this should be `dimension.name()`

name = exact name
toString = defaults to name but can be overridden to provide additional/different string

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -123,6 +116,60 @@ sqlbuilder:
                 groupBy: >-
                      se.ethnicity_id
 
+            IEP:
+              clauses:
+                select: >-
+                     sb.id AS dimension_id
```
jtreuting: Where is the data for which type of dimension this is, so we can display something different for IEP Yes and Seciton 504 Yes?

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -93,16 +89,14 @@ sqlbuilder:
                      sch.id
 
       # ---------- dimension parts -----------------------------------------------------------------------
-            overall:
+            Overall:
```
malaffoon: might want to add a comment that these values correspond to exact name of DimensionType enums

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -172,17 +219,42 @@ sqlbuilder:
...
+                  CROSS JOIN (select 'LEP' AS dimension, b.id AS dimension_id, b.code as dimension_code from strict_boolean b) d
+
+            EconomicDisadvantage:
               clauses:
                 join: >-
-                  CROSS JOIN (select 'overall' AS dimension, -1 AS dimension_id, null as dimension_code) d
+                  CROSS JOIN (select 'EconomicDisadvantage' AS dimension, b.id AS dimension_id, b.code as dimension_code from strict_boolean b) d
 
-            gender:
+            Status504:
```
jtreuting: Isn't this `Section504`?

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
agorina: We have it in the other query, this was duplicate data. It is hard to see in this review, but there is an overarching query that has the same info in the application.sql.yml.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: You are right, changing.

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
agorina: Will add.

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
agorina: Yes, I have no idea why I made it Status. Will change.

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
agorina: Line 240 has this info. I know it is confusing to read it.

PR #722, opened 2018-01-03
DWR-1219 Add support to drill down by the student enrolled grade

agorina:
Just to note: I ended up with the version that does not backfill any 'student enrolled grades'.


aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -41,19 +41,24 @@ sqlbuilder:
             state:
               clauses:
                 select: >-
-                     -1 AS organization_id
+                     'State' AS organization_type,
+                     -1 AS organization_id,
+                     NULL AS organization_name
```
agorina: For all other queries it is not needed since the same data is available in the dimension 'customAggregateAllPermutations' query. 

For the student enrolled grade we need this since I would like to reuse the same query builder.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
```diff
@@ -297,36 +364,19 @@ public void itShouldFindByQueryWithEthnicity() {
...
-        final int genderCount = countRowsInTable(template, "gender");
-
-        assertThat(reportRows.size()).isEqualTo(
-                (grade8AsmtSubject1Count * ethnicityCount * genderCount * 2) +
-                        (grade8AsmtSubject2Count * ethnicityCount * genderCount * 2) +
-                        (grade9AsmtSubject1Count * ethnicityCount * genderCount * 2) +
-                        (grade9AsmtSubject2Count * ethnicityCount * genderCount * 2) +
-                        2 * (grade8AsmtSubject1Count + grade8AsmtSubject2Count + grade9AsmtSubject1Count + grade9AsmtSubject2Count)
-        );
+        assertThat(reportRows.size()).isEqualTo(asmtCount * (ethnicityCount + genderCount + 1));
```
agorina: I have no idea what I was thinking when I wrote this test but it seemed complicated and not exactly correct. This is a simplified version.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -107,31 +111,35 @@ public JdbcCustomAggregateReportRowRepository(final NamedParameterJdbcTemplate j
...
         return queries;
     }
 
-    private SelectQuery buildSql(final List<String> addOns) {
+    private SelectQuery buildSql(final List<String> addOns, final DimensionType dimensionType) {
         final Properties properties = new Properties();
-        properties.put(customAggregateAllPermutationsTemplateSql, queryProvider.newQuery(customAggregateAllPermutationsTemplateSql, addOns).toString());
         properties.put(customAggregateTemplateSql, queryProvider.newQuery(customAggregateTemplateSql, addOns).toString());
-        return new SelectQuery().addCustomColumns(new CustomSql(replacePlaceholders(customAggregateSql, properties)));
+        if (dimensionType.equals(DimensionType.StudentEnrolledGrade)) {
```
agorina: I do not like this fork but could come up with a better approach.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
```diff
@@ -217,6 +225,65 @@ public void itShouldFindByQueryWithMigrantStatus() {
...
+
+        assertThat(rows.size()).isEqualTo(2);
+        for (final ReportRow row : rows) {
+            if (row.getDimension().getCode().equalsIgnoreCase("-5")) assertThat(countStudents(row.getMeasures())).isEqualTo(1);
+            else assertThat(countStudents(row.getMeasures())).isEqualTo(2);
+        }
+    }
+
+    @Test
+    public void itShouldBackFillMissingDataForAllDimensionsButStudentEnrolledGrade() {
```
agorina: I added this test to verify that we backfill the data as expected.

malaffoon:
nit: to avoid confusion we should probably avoid "drill down" and use "disaggregate" or some such.

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
malaffoon: I can't remember if i already asked this but ... can organization_id be NULL instead of -1? If it must be -1 perhaps a comment indicating where it is referenced?

agorina: You did ask before :)  (you are very consistent)

I will add a comment: it is needed to support the JOIN in the final overarching query defined in the application.sql.yml

PR #726, opened 2018-01-05
add new style to push report table containing wells out wide with the tables

jtreuting:
a new class is needed since it doesn't work well with all the well's (yes pun was intended!)

**Before**
wide-tables-before

**After**
wide-tables-after

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -498,7 +498,7 @@ body .select-menu-group {
 }
 
 /** used on well's containing wide data tables to push the well out with the table */
-.report-table-container {
+.wide-content-container {
   width: min-content;
   width: -moz-min-content;
   width: -webkit-min-content;
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -496,3 +496,10 @@ body .select-menu-group {
   white-space: nowrap;
   padding: .2em 0 .3em;
 }
+
+/** used on well's containing wide data tables to push the well out with the table */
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -496,3 +496,10 @@ body .select-menu-group {
   white-space: nowrap;
   padding: .2em 0 .3em;
 }
+
+/** used on well's containing wide data tables to push the well out with the table */
+.report-table-container {
```
dsnyder-fairway: can we make this not specific to report tables? The reason I ask is I am doing something similar for the aggregate reports

jtreuting: sure, i'm not sure a better more generic name.  maybe, wide-content-container?

dsnyder-fairway: that sounds good 👍

webapp/src/main/webapp/src/styles.less
```diff
@@ -496,3 +496,10 @@ body .select-menu-group {
   white-space: nowrap;
   padding: .2em 0 .3em;
 }
+
+/** used on well's containing wide data tables to push the well out with the table */
+.wide-content-container {
```
esotrope: I'm thinking we'll probably want all wells to behave this way in the app no? Eventually we might want to extend the bootstrap `.well` class in the ui kit. @aaronsleeper thoughts?

webapp/src/main/webapp/src/styles.less
esotrope: This would have been my name of choice:

PR #728, opened 2018-01-09
basic search filter entity repos and tests

esotrope:
these repos will provide the filter data for the aggregate search page

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
@@ -1,7 +1,7 @@
 package org.opentestsystem.rdw.olap.model;
 
 /**
- * The represents any data entity which has an ID and a translation code
+ * This class represents any data entity which has an ID and a translation code
  */
 public class CodedEntity {
 
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
@@ -1,5 +1,8 @@
 package org.opentestsystem.rdw.olap.model;
 
+/**
+ * The represents any data entity which has an ID and a translation code
+ */
 public class CodedEntity {
 
     private int id;
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
@@ -1,0 +1,47 @@
+package org.opentestsystem.rdw.olap.model;
+
+public class CodedEntity {
+
+    private int id;
+    private String code;
+
+    public int getId() {
+        return id;
+    }
+
+    public String getCode() {
+        return code;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder<A extends CodedEntity, B extends Builder<A, B>> {
+
+        private int id;
+        private String code;
+
+        public B id(final int id) {
+            this.id = id;
+            return (B) this;
+        }
+
+        public B code(final String code) {
+            this.code = code;
+            return (B) this;
+        }
+
+        protected A build(final CodedEntity entity) {
+            entity.id = id;
+            entity.code = code;
+            return (A) entity;
+        }
+
+        public A build() {
+            return build(new CodedEntity());
+        }
+
+    }
+
+}
--- a/aggregate-service/src/main/resources/application.sql.yml
+++ b/aggregate-service/src/main/resources/application.sql.yml
@@ -1,4 +1,37 @@
 sql:
+  administrationCondition:
+    findAll: select id, code from administration_condition
+
+  assessmentGrade:
+    findAll: >-
+      select id,
+        code
+      from grade g where exists (
+          select a.grade_id
+          from fact_student_exam e
+            join asmt a on e.asmt_id=a.id
+          and g.id = a.grade_id
+      )
+
+  assessmentType:
+    findAll: select id, code from asmt_type
+
+  completeness:
+    findAll: select id, code from completeness
+
+  gender:
+    findAll: select id, code from gender
+
+  schoolYear:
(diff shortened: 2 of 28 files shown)


aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -1,4 +1,37 @@
...
+        code
+      from grade g where exists (
+          select a.grade_id
+          from fact_student_exam e
+            join asmt a on e.asmt_id=a.id
+          and g.id = a.grade_id
+      )
+
+  assessmentType:
+    findAll: select id, code from asmt_type
```
agorina: I believe we wanted this to be driven by the exams. @jtreuting , could you please confirm or deny?

aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -1,4 +1,37 @@
...
+  assessmentGrade:
+    findAll: >-
+      select id,
+        code
+      from grade g where exists (
+          select a.grade_id
+          from fact_student_exam e
+            join asmt a on e.asmt_id=a.id
+          and g.id = a.grade_id
+      )
```
agorina: I would go with DISTINCT here: 
```sql
SELECT DISTINCT
  g.id,
  g.code
FROM fact_student_exam f
  JOIN asmt a ON a.id = f.asmt_id
  JOIN grade g ON g.id = a.grade_id
```

Redshift is impressively fast with it.
Aslo, do we want it to be based on the asmt type?

aggregate-service/src/main/resources/application.sql.yml
jtreuting: The idea is that if there are no Summatives it should not be listed as an option.  And the original thinking was that we would drive that off actual exams, but the more I think about it the more I'm not sure if that is needed.  Since we can't use organizaiton or year to narrow this down (since the user hasn't selected it yet), it would only provide value for the very first year using the system before Summatives have been taken.  Cause in year 2 they would be in the system and have to be available as an option so they could look at last years results.

So I think it might be as effective to drive that based on config settings for IsIABEnabled, IsICAEnabled and IsSummativeEnabled.  We've discussed having those flags and I think the might serve the purpose better than the check against exams since it won't hit the DB for that.

Thoughts?

agorina: I am fine with this to be driven by the configuration, but I still think that the grades should be driven by the exams (see my comment below). 

Here are my reasons for it:
1. The sooner we stop the user from running an expensive query with something that does not exists, the less pressure we are putting on Redshift. 
2. We fill in the gaps with the zeros, so I feel that letting them see a large table with something that is not available is not the best user experience.
3. When we do a longitudinal report, this may be even a bigger concern. I believe our model CT web site gives an error if a combination of the selected grade and school year is not available.

aggregate-service/src/main/resources/application.sql.yml
esotrope: ok. And IDK. That is a general question I have about the whole form in general - are any form option values dependent on other previous form value entries? My guess is it was not intended this way as it seems to lay everything out at once but then again I'm not sure it was thought about much yet. @aaronsleeper @jtreuting @dsnyder-fairway was this discussed at all during design?

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAssessmentTypeRepositoryRST.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Autowired
+    private JdbcAssessmentTypeRepository repository;
+
+    @Test
+    public void itShouldFindAll() throws Exception {
+        assertThat(repository.findAll()).containsExactlyInAnyOrder(
+                CodedEntity.builder().id(1).code("ica").build(),
+                CodedEntity.builder().id(2).code("iab").build(),
+                CodedEntity.builder().id(3).code("sum").build()
```
esotrope: In general, is it a good idea to have tests verifying against real data on tables assumed to always be pre-populated? This test will break when we add a new exam type for instance. In other words, should we bother with the repo test when the test doesn't have control of the initial conditions of its data table?

aggregate-service/src/main/resources/application.sql.yml
esotrope: Configuration is fine. I'm thinking it would be best to drive the config through the DB though by adding an `aggregate_enabled` column to `asmt_type`. Will has discusses such an option to alleviate all instances of "isICA" "isSummative" ... etc. littered throughout the code.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
```diff
@@ -0,0 +1,50 @@
+package org.opentestsystem.rdw.olap.model;
+
+/**
+ * The represents any data entity which has an ID and a translation code
```
chdietz: `The` ?

esotrope: You know, _the_ - it represents any data ...

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAssessmentTypeRepositoryRST.java
chdietz: If it is a repository integration type of test test I start with the data the way it is after clean and migrate and then a setup sql in the resources folder to put the database in a known state.

aggregate-service/src/main/resources/application.sql.yml
jtreuting: Before doing that I'd like to think about consistency when it comes to setting these things.  The ops team would need to have SQL scripts for doing this versus doing other things with config settings files.  @malaffoon you know that eco-system much better so would like to hear your thoughts.

@agorina I think this would be an easier conversation to have offline and discuss the flow for that on the query builder page.  Let's sync up this afternoon with some folks to discuss.

agorina: Sounds good.

PR #731, opened 2018-01-09
minor improvements to average scale scores

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -13,13 +13,13 @@
...
               <span class="score label label-primary">{{ statistics.average | number:'1.0-0' | removeComma }}</span>
-              <span class="error-band black pl-xs">± {{ statistics.standardError | number:'1.0-0' }}</span>
+              <span class="error-band black">±{{ statistics.standardError | number:'1.0-0' }}</span>
             </div>
             <span class="text-left gray-darkest">
               <span class="font-large">
                 <span class="score">{{ statistics.total }}</span>
               </span>
-              <span>{{ 'labels.assessments.students' | translate }}</span>
+              <span class="pl-xs">{{ (statistics.total == 1 ? 'labels.assessments.student' : 'labels.assessments.students') | translate }}</span>
```
esotrope: in Japanese this doesn't happen. +1 Japanese

malaffoon: Doesn't angular translate have plural handling code? I vaguely remember from years gone by that some translate mechanism have a hook for it.

dsnyder-fairway: I'm not totally sure. I see ngPlural, but that seems like more than we need

jtreuting: yeah ngplural is a more fancy switch statement it seems.  i don't know of anything that would automatically do plural based on the language.  like pass it 'Student' and it would know to add the 's' for English or not in Japanese (apparently)

PR #732, opened 2018-01-10
redshift tests

malaffoon:


build.gradle
```diff
@@ -260,21 +262,25 @@ task extractSchema(type: Sync) {
...
-               "cleanReporting": "Cleans the schemas that are required for running locally",
-               "migrateReporting_test": "Migrates the schemas that integration tests depend on",
-               "cleanReporting_test": "Cleans the schemas that integration tests depend on",
-               "migrateWarehouse": "Migrates the schemas that are required for running locally",
-               "cleanWarehouse": "Cleans the schemas that are required for running locally",
-               "migrateWarehouse_test": "Migrates the schemas that integration tests depend on",
-               "cleanWarehouse_test": "Cleans the schemas that integration tests depend on"]
+def taskMap = [
+//        "cleanReporting": "Cleans the schemas that are required for running locally",
+//        "migrateReporting": "Migrates the schemas that are required for running locally",
```
malaffoon: Anybody use these? The tasks are `clean/migrateReportingProd`.

aggregate-service/build.gradle
```diff
@@ -33,10 +33,35 @@ dependencies {
...
+// To run these tests locally, you must either manually adjust the application-redshift.yml file
+// (but don't check in your changes!) or set environment variables. If you are using the default
+// CI data resources you should set:
+//   APP_DATAMART_OLAP_DATASOURCE_PASSWORD
+// If you are using different instances the full set of variables can be found in application-redshift.yml.
+// If you are running from the command-line you can use local bash export to do it in one command, e.g.:
+/*
+(export APP_DATAMART_OLAP_DATASOURCE_PASSWORD=secretkey; \
+ gradle rst)
+*/
```
esotrope: Thanks for these comments - very helpful

build.gradle
esotrope: i use them every so often. what are we supposed to use instead?

malaffoon: Oh, i can put them back then. The only real alternative is to run the similar tasks from RDW_Schema, but then the version of RDW_Schema might not match.

PR #733, opened 2018-01-10
fix warnings and typo

malaffoon:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
```diff
@@ -16,6 +18,20 @@ public String getCode() {
         return code;
     }
 
+    @Override
+    public boolean equals(final Object o) {
```
malaffoon: It was either this or `usingFieldByFieldComparator` in all the tests

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCompletenessRepository.java
```diff
@@ -10,7 +10,7 @@
 
     JdbcCompletenessRepository(
             final NamedParameterJdbcTemplate template,
-            @Value("${sql.assessmentType.findAll}") final String findAllQuery) {
+            @Value("${sql.completeness.findAll}") final String findAllQuery) {
```
esotrope: thanks for the catch

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/CodedEntity.java
esotrope: Isn't that preferable? That way you can read in the test that is what constitutes equality.

malaffoon: Heh. I've spent the last year trying to use custom comparators in the tests, instead of adding equals/hash methods. I thought the trend with this codebase was to use equals/hash so i did this. I can change it.

PR #735, opened 2018-01-10
must have at least one assessment selected

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -203,8 +203,10 @@ export class AssessmentsComponent implements OnInit {
   }
 
   removeAssessment(assessment: Assessment) {
-    assessment.selected = false;
-    this.selectedAssessmentsChanged(assessment);
+    if (this.selectedAssessments.length > 1) {
+      assessment.selected = false;
+      this.selectedAssessmentsChanged(assessment);
```
esotrope: so you could remove one that wasn't selected?

dsnyder-fairway: when opening the select assessments you could remove the only selected assessment which would leave the page in an unusable state

esotrope: hm. that might not be straight forward though if someones workflow is to remove everything then add just the one they want no? is this to address a bug?

dsnyder-fairway: yeah it was to address a bug. What happens is if a user removes all the assessments when the "Click an assessment to add or remove it from your list of selections above." is visible, that well disappears and the only options is to (+), but clicking that results in no action. 

My goal was to be consistent with the behavior when not adding an assessment -- if only one is selected, do not allow it to be removed

PR #738, opened 2018-01-10
Changed the 'error band' calculation in the aggregate report.

agorina:


aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -7,7 +7,7 @@ sqlbuilder:
         clauses:
           select: >-
               round(avg(scale_score)) AS score,
-              round(sqrt(sum(power(scale_score_std_err, 2)) / count(*))) as std_err,
+              round(stddev_samp(scale_score)/sqrt(count(*))) as std_err,
```
agorina: With this change we do not need `scale_score_std_err` in the Redshift schema. Should I remove it from there? @jtreuting , @malaffoon ?

jtreuting: it feels like we might want to wait a while before removing it .... just in case ???

malaffoon: Yeah. In theory the std_err calculation should take the individual std_err's into account so it'll be nice to have them.

agorina: Thanks for the feedback. I will keep it then.

PR #740, opened 2018-01-11
adds organization endpoint components and tests

esotrope:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/ApplicationConfiguration.java
```diff
@@ -34,4 +44,16 @@ public OrganizationRepository repository(final NamedParameterJdbcTemplate templa
...
 
+    @Bean
+    public Cache reportOptions() {
+        return new GuavaCache(ReportOptions, CacheBuilder.newBuilder().build());
+    }
+
+    @Bean
+    public Cache reportOrganizations() {
+        return new GuavaCache(ReportOrganizations, CacheBuilder.newBuilder()
+                .expireAfterWrite(1, TimeUnit.DAYS)
```
jtreuting: This means that the cache will be around for 1 day from the first time someone hits up the page.  So if migrate happens at 2am and someone gets to the site at 6am to check on things, this data will be cached from 6am - 6am.  Meaning that tomorrow if someone goes to the site at 5am they won't get the most recently migrated data necessarily.  Just pointing it out since it's not ideal.

Ideally the cache would be triggered by the migrate or a hard expiration time that is in sync with the migrate would be used instead of relative to the writing of the cache.

esotrope: Good point. For now I can choose a cron config for 1am or something. The only way i can imagine expiring the cache on migrate is with a message stream and distributed cache.

jtreuting: @malaffoon or @agorina  Is there a setting that defines when the migrate happens that we could use here to make sure they are in sync?

agorina: I agree with you comment above, the cache cleanup should be triggered by the olap migrate. 

Maybe worth creating a JIRA to come up with the solution.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultOrganizationSearchService.java
```diff
@@ -0,0 +1,71 @@
...
+    public Set<Organization> getOrganizations(final User user, final String nameSearch) {
+
+        // fail fast
+        if (Strings.isNullOrEmpty(nameSearch)) {
+            return ImmutableSet.of();
+        }
+
+        final PermissionScope permissionScope = user.getPermissionScopeByPermissionId(AggregateRead);
+
+        final Predicate<Organization> securityFilter = permissionScope.isStatewide()
```
malaffoon: Not sure why but i find this hard to read. I think a ternary operator returning different lambdas, all in-line is just too much. Perhaps break out the filter into a separate class (can be static in this module to avoid new module). And perhaps have one filter that just exits early if the permission scope is statewide? Although i guess we could have AlwaysTrue predicate.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateServiceCache.java
```diff
@@ -0,0 +1,13 @@
+package org.opentestsystem.rdw.olap.model;
+
+/**
+ * Security permissions used by the aggregate service application
```
malaffoon: c&p?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/ApplicationConfiguration.java
malaffoon: Yes, we've talked about having aggressive caching in the aggregate-service, requiring coordinated flushing of that cache. We need to either have a single property that controls both the migrate and the caches, or have the  migrate process directly trigger the cache flush (probably using an actuator end-point in the service).

wtritch: I think having a 24h expiration is reasonable along with the actuator endpoint.  Just in case.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.scheduling.annotation.Scheduled;
+
+import static org.opentestsystem.rdw.olap.model.AggregateServiceCache.ReportOrganizations;
+
+@Configuration
+@EnableCaching
+@EnableScheduling
+public class CacheConfiguration {
+
+    @Scheduled(cron = "${app.cache.reportOrganizations.eviction-cron}")
```
wtritch: This makes me nervous.  devOps now needs to configure the same kind of cron across multiple environments, making sure that they all line up...

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultOrganizationSearchService.java
```diff
@@ -0,0 +1,71 @@
...
+            // If it's a district make sure the district is in the permission scope
+            if (organization.getOrganizationType() == OrganizationType.District) {
+                return permissionScope.getDistrictIds().contains(organization.getId());
+            }
+
+            // If it's a school make sure the school or school's district is in the permission scope
+            if (organization.getOrganizationType() == OrganizationType.School) {
+                final School school = (School) organization;
+                return permissionScope.getInstitutionIds().contains(school.getId())
+                        || permissionScope.getDistrictIds().contains(school.getDistrictId());
```
wtritch: If the district is in your district group or school is in your school-group, district, or district group you're also allowed.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/ApplicationConfiguration.java
esotrope: Having an endpoint implies we switch to using a distributed cache btw.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultOrganizationSearchService.java
esotrope: The always true predicate was technically more efficient than having one predicate with a fail fast hence the choice to use it - i can go ahead and put this in a class.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateServiceCache.java
esotrope: woops yup

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+    private OrganizationService organizationService;
+
+    /**
+     * Refreshes the contents of the report organizations cache
+     */
+    @Scheduled(cron = "${app.cache.reportOrganizations.refresh-cron}")
+    @CacheEvict(cacheNames = {ReportOrganizations}, beforeInvocation = true, allEntries = true)
+    public void refreshReportOrganizationsCache() {
+        organizationService.getSchoolsAndDistricts();
+    }
```
esotrope: changed the cache to refresh on a cron instead of just evict. this way the user will not have to wait for the data but will (mostly always) get a cached set. Endpoint cache invalidation will have to be a part of a separate PR.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultOrganizationSearchService.java
```diff
@@ -0,0 +1,85 @@
...
+                .filter(securityFilter)
+                .filter(nameContainsTextIgnoreCaseFilter)
+                .collect(toImmutableSet());
+    }
+
+    /**
+     * @param permissionScope the permission scope to test containership with
+     * @return predicate that evaluates true when the tested organization is contained within the organization IDs of the permission scope
+     */
+    private static Predicate<Organization> permissionScopeContainsOrganization(final PermissionScope permissionScope) {
```
esotrope: @malaffoon the predicate to a factory method
@wtritch added the school + district group checks

esotrope:
pushing this through for now so the UI can get some orgs

PR #741, opened 2018-01-11
Add aggregate-report processing to report-processor

wtritch:
This PR adds the ability for the user to request an aggregate report similarly to a PDF or exam export.  

The general process is:
1. user submits an AggregateReportRequest to the report-processor (containing an aggregate ReportQuery)
2. The request gets added to an external processing queue (RabbitMQ)
3. The single queue processing thread chews through the queue, making synchronous calls to the aggregate-service to retrieve the report CSV
4. The CSV is stored for the report in S3, allowing the user to download it at their leisure

The idea is that 1 and 2 can handle a large amount of traffic while 3 and 4 are single-threaded (per report-processor pod) to avoid overwhelming redshift.

NOTE: I'll hold off on merging this until I can add in an actual CSV export endpoint to the aggregate-service.


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
```diff
@@ -1,4 +1,4 @@
-package org.opentestsystem.rdw.olap.model;
+package org.opentestsystem.rdw.reporting.common.model;
```
wtritch: I moved the ReportQuery and DimensionType models into common to allow them to be used in a report-processor report request.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
```diff
@@ -66,7 +66,7 @@ public boolean isIncludeAllDistricts() {
     }
 
     public Set<DimensionType> getDimensionTypes() {
```
wtritch: Jackson demands that field and property names match for easy serialization/deserialization.  I don't particularly care whether we change the field to match the getter or the getter to match the field, but I've changed to field to match the getter here.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public ReportType getReportType() {
+        return ReportType.AggregateReportRequest;
+    }
+
+    /**
+     * Override default behavior to provide the subject from the ReportQuery.
+     * NOTE: ReportQuery instances may contain multiple subject ids.  If multiple
+     * subject ids are declared, return "null" to indicate "all subjects."
+     * TODO Account for multiple subjects if/when we have more than just Math and ELA
```
wtritch: So right now we interpret NULL as MATH or ELA in the reporting service.  I'm not sure if aggregate-service has the same interpretation?  @agorina ?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
```diff
@@ -0,0 +1,74 @@
...
+        return reportQuery.getSubjectIds().size() != 1
+                ? null
+                : Subject.valueOf(Iterables.getOnlyElement(reportQuery.getSubjectIds()));
+    }
+
+    /**
+     * Override default behavior to provide the school year from the ReportQuery.
+     * NOTE: ReportQuery instances may contain multiple school years.  If multiple
+     * school years are declared, return the last school year.
+     * TODO Account for multiple school years in model/API
```
wtritch: Same issue here.  All of our current reports are scoped to a single subject (or all subjects) and a single school year.  The aggregate report is a slightly different animal.  Right now I'm just picking a subject and picking a school year for display in the "My Reports" UI.  Thoughts?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultAggregateReportService.java
```diff
@@ -0,0 +1,104 @@
...
+                .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
+                .setVisibility(PropertyAccessor.GETTER, JsonAutoDetect.Visibility.NONE);
+
+        this.clientHttpRequestFactory = clientHttpRequestFactory;
+        this.reportContentService = reportContentService;
+        this.aggregateReportProperties = aggregateReportProperties;
+    }
+
+    @Override
+    public Report generateReport(final Report report) {
```
wtritch: This is the meat of the processing.  I'm making a simple JWT-authenticated HTTP call from the report-processor to the aggregate-service to fetch the report CSV, stuffing it into the content service, and updating the report status to COMPLETED/FAILED.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
```diff
@@ -78,10 +81,15 @@ private void publishReportRequest(final BatchExamReportRequestHolder requestHold
...
-            pdfReportSource.fetchStudents().send(message);
+        switch (requestHolder.getRequest().getReportType()) {
+            case ExamExport:
+                exportReportSource.exportExams().send(message);
+                break;
+            case AggregateReportRequest:
+                aggregateReportsSource.aggregateReports().send(message);
+                break;
+            default:
+                pdfReportSource.fetchStudents().send(message);
```
wtritch: We now ship report requests to 3 different queues based upon type: pdf, exam export, aggregate report

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
agorina: No, it does not.  Do we need to?

wtritch: I think aggregate-service is doing it the "right way" especially if the trigger is ever pulled on configurable subjects.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
agorina: I am confused...why does it have to be just one? Why couldn't we support multiple years?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultAggregateReportService.java
```diff
@@ -0,0 +1,104 @@
...
+        try {
+            final AggregateReportRequest reportRequest = (AggregateReportRequest) report.getReportRequest();
+            final ReportQuery query = reportRequest.getReportQuery();
+
+            //Send to aggregate-service with JWT
+            final ClientHttpRequest request = clientHttpRequestFactory.createRequest(getUri(), HttpMethod.POST);
```
wtritch: Note that I want to use a POST here to use standard jackson serialization of the ReportQuery rather than write a ReportQuery -> query params transformer in the report-processor project.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultAggregateReportService.java
```diff
@@ -0,0 +1,104 @@
...
+            final HttpHeaders headers = request.getHeaders();
+            final String authorizationHeader = BearerAuthorizationHeaders.create(jwtService.encode(Users.getCurrentUser()));
```
esotrope: is it better to pull the header from the active request? that way we wouldn't have to re-create the header. on the other hand that would make it annoying to test.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
wtritch: We could, but that involves a refactoring from the UI through the API down to the database reporting.user_report.report_request field that stores serialized AbstractReportRequest instances as serialized json text.  If we want to modify our models we'll have to take great care and possibly provide trial deserializations with versioning...

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultAggregateReportService.java
wtritch: Well shoot.  That brings up a great point.  This is being run on a queue handler thread that doesn't have an "authenticated user."  This won't work.  I'll need to pass the User as part of the message that goes through the queue.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
wtritch: @jtreuting This basically means that we have to pick one subject (or all subjects) and one school year to display to the user in the "My Reports" view.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/AggregateReportProperties.java
```diff
@@ -0,0 +1,20 @@
+package org.opentestsystem.rdw.reporting.processor.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Aggregate report properties
+ */
+@ConfigurationProperties(prefix = "app.aggregate-report")
+public class AggregateReportProperties {
```
malaffoon: Are these "report" properties or "service" properties? Seems like the URL is a property of the aggregate-service? I guess you could say it is the URL needed when processing an aggregate-report request.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
malaffoon: Are we trying too hard to extend the base request class as is? Perhaps AbstractExamReportRequest shouldn't have getSubject() and getSchoolYear(), they could be added in an AbstractSimpleExamReportRequest or some such.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
jtreuting: Can the value be a specific single value or maybe "MultipleYears" as the value for example?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultAggregateReportService.java
malaffoon: User or token or both?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
malaffoon: I'm not thrilled posting a message from the `default` block of a switch statement, can we list the explicit report type(s)?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReports.java
```diff
@@ -0,0 +1,41 @@
...
+
+import static org.opentestsystem.rdw.reporting.processor.model.ReportStatus.FAILED;
+import static org.opentestsystem.rdw.reporting.processor.stream.AggregateReportsSource.AggregateReports;
+
+/**
+ * This component is responsible for fetching, ordering,
+ * and chunking students within the context of a report.
+ */
+@Component
+public class AggregateReports {
```
malaffoon: So, in this case, "aggregate" is a verb?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/AggregateReportProperties.java
wtritch: I debated the name a bit, but AggregateServiceProperties felt strange...

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ReportRequestMessage.java
```diff
@@ -1,12 +1,14 @@
...
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+
 /**
  * This message represents a basic report processing message.
  */
 public class ReportRequestMessage {
 
     protected long reportId;
-    protected UserPermissions userPermissions;
+    protected User user;
```
esotrope: 👍 👍

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
wtritch: Possibly, but again, modifying these models will require some jackson acrobatics to still be able to deserialize V1 models from the database.  Maybe we could introduce an API resource that just has getSubjects() and getSchoolYears() that can be populated based upon the underlying processing model?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
wtritch: I don't think either answer is really right.  I think the correct thing to do would be to display all years.  Maybe via a web API resource model from the discussion with @malaffoon above?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/BatchExamReportRequestHolder.java
```diff
@@ -1,52 +0,0 @@
...
-    private String user;
-    private String label;
-
-    @JsonTypeInfo(use=JsonTypeInfo.Id.CLASS, include=JsonTypeInfo.As.PROPERTY, property="@class")
-    private AbstractExamReportRequest<T> request;
-
-    /**
-     * No-arg constructor for serialization
-     */
-    private BatchExamReportRequestHolder() {}
```
wtritch: By passing the actual User instance through the processing flow we no longer need this "poor man's jwt" implementation

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/UserPermissions.java
```diff
@@ -1,59 +0,0 @@
...
-import java.util.Map;
-
-import static com.google.common.collect.ImmutableMap.copyOf;
-import static com.google.common.collect.ImmutableMap.of;
-
-/**
- * This class is responsible for containing the user's granted permissions
- * within the context of the Reporting application suite.
- */
-public class UserPermissions {
```
wtritch: Now that we're passing a User we no longer need this subset-of-a-user object

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReports.java
wtritch: hmm... will rename to AggregateReportGenerator

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/test/support/UserPermissionsTestBuilder.java
```diff
@@ -1,29 +0,0 @@
...
-import org.opentestsystem.rdw.security.Permission;
-import org.opentestsystem.rdw.security.PermissionScope;
-
-import static org.opentestsystem.rdw.reporting.common.security.ReportingPermission.GroupPiiRead;
-import static org.opentestsystem.rdw.reporting.common.security.ReportingPermission.IndividualPiiRead;
-
-/**
- * Test utility for creating UserPermissions instances.
- */
-public class UserPermissionsTestBuilder {
```
wtritch: Renamed to UserTestBuilder

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultAggregateReportService.java
wtritch: I like working with model objects rather than opaque serialized values, personally.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
wtritch: done

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
jtreuting: Okay sounds like this is used for more than jus the name of the file.  So these are displayed on the reports listing as well, right?  I was thinnking if it was just the name of the file, it's not a big deal to say "MultipleYears" instead of the specific ones but that feels stranger on the listing page i guess.

PR #743, opened 2018-01-13
adds basic security to aggregate report services

esotrope:
This is part 1 of adding security to custom aggregate reports
This basically just makes sure the user has CUSTOM_AGGREGATE_READ in order to access any aggregate-service endpoints

Part 2 will make sure the user has rights to make reports based on the orgs associated with the permission

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
@@ -1,7 +1,5 @@
 package org.opentestsystem.rdw.olap.repository.impl;
 
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableMap.Builder;
 import com.healthmarketscience.sqlbuilder.CustomSql;
 import com.healthmarketscience.sqlbuilder.SelectQuery;
 import org.opentestsystem.rdw.common.model.AssessmentType;
@@ -38,6 +36,7 @@ import static org.opentestsystem.rdw.reporting.common.model.ReportQuery.checkIsV
 @SuppressWarnings("unchecked")
 @Repository
 class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRepository {
+
     private static final String customAggregateTemplateSql = "customAggregate";
     private static final String customAggregateAllPermutationsTemplateSql = "customAggregateAllPermutations";
     private static final String stateAddOn = "state";
@@ -45,6 +44,9 @@ class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRep
     private static final String allDistrictsAddOn = "allDistricts";
     private static final String schoolsAddOn = "schools";
     private static final String allSchoolsInDistrictsAddOn = "allSchoolsInDistricts";
+
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+    private final QueryProvider queryProvider;
 
     @Value("${sql.customAggregate}")
     private String customAggregateSql;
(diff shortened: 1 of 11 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -55,43 +57,46 @@
...
     public List<ReportRow> findByQuery(@NotNull final ReportQuery query) {
+
         validate(query);
 
-        final Builder parmBuilder = ImmutableMap.<String, Object>builder()
-                .put("school_years", query.getSchoolYears())
-                .put("asmt_grade_ids", query.getAsmtGrades())
-                .put("subject_ids", query.getSubjectIds())
-                .put("asmt_type_id", query.getAssessmentType().id());
+        final MapSqlParameterSource parameters = new MapSqlParameterSource()
```
esotrope: changed this to be consistent with the other method. also in general it's probably best to use this class since we are using spring rest template and this is a spring dep

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/OrganizationSearchService.java
```diff
@@ -2,13 +2,15 @@
...
 import org.opentestsystem.rdw.reporting.common.web.security.User;
+import org.springframework.security.access.prepost.PreAuthorize;
 
 import javax.validation.constraints.NotNull;
 import java.util.Set;
 
 /**
  * Service responsible for managing organizations
  */
+@PreAuthorize("hasAuthority('PERM_CUSTOM_AGGREGATE_READ')")
```
esotrope: Only added these to the "entry point" services. debating whether or not to add it to all services in the app or not.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
```diff
@@ -14,23 +16,30 @@
  * This controller is responsible for providing custom aggregate report
  */
 @RestController
-@RequestMapping("/api")
```
esotrope: removed `/api` prefix.

webapp/docker-compose.yml
```diff
@@ -49,8 +49,7 @@ services:
     links:
       - config-server
       - admin-service
-#      TODO uncomment when aggregate-service is uncommented
-#      - aggregate-service
+      - aggregate-service
       - report-processor-service
       - reporting-service
```
esotrope: unrelated but i wasn't able to get reporting-service running today in local docker it failed to connect to the config-server. anything in this file jump out as a reason to why that might happen?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Thank you, makes sense.

I am making changes in this class too. To avoid merge conflicts, could you please merge it before doing part 2 of the permissions?

esotrope:
merging now - will address comments in part 2

agorina:
thank you!

webapp/docker-compose.yml
```diff
@@ -78,25 +77,24 @@ services:
...
-#      - /tmp:/tmp
-#    links:
-#      - config-server
-#    environment:
-#      - spring.profiles.active=local-docker
-#      - CONFIG_SERVICE_ENABLED=true
-#      - CONFIG_SERVICE_URL=http://config-server:8888
-#    extra_hosts:
-#      - "dwhost:$DATAWAREHOUSE_HOST"
+  aggregate-service:
```
malaffoon: I'm not sure we want this uncommented by default. Anything that goes to redshift should be disabled by default and the developer needs to explicitly uncomment it and configure it. For example, which redshift instance does this point to be default?

PR #744, opened 2018-01-13
DWR-1220 Add filtering by completeness and admin condition to aggregate report

agorina:
A first baby step to add filters to the aggregate report: add filtering by `completeness` and `admin conditions`to the repo. These two filters are the easiest to add since they do not overlap with the disaggregate options.

Too many people are working on the same files; to minimize the scope of the change I am breaking it into small steps.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
@@ -1,8 +1,5 @@
 package org.opentestsystem.rdw.olap.repository.impl;
 
-import com.google.common.collect.ImmutableSet;
-import com.healthmarketscience.sqlbuilder.CustomSql;
-import com.healthmarketscience.sqlbuilder.SelectQuery;
 import org.opentestsystem.rdw.common.model.AssessmentType;
 import org.opentestsystem.rdw.olap.model.Assessment;
 import org.opentestsystem.rdw.olap.model.Dimension;
@@ -10,29 +7,21 @@ import org.opentestsystem.rdw.olap.model.Measures;
 import org.opentestsystem.rdw.olap.model.Organization;
 import org.opentestsystem.rdw.olap.model.ReportRow;
 import org.opentestsystem.rdw.olap.repository.CustomAggregateReportRepository;
+import org.opentestsystem.rdw.olap.sqlbuilder.ReportQueryProvider;
 import org.opentestsystem.rdw.reporting.common.model.DimensionType;
 import org.opentestsystem.rdw.reporting.common.model.OrganizationType;
 import org.opentestsystem.rdw.reporting.common.model.ReportQuery;
-import org.opentestsystem.rdw.reporting.common.sqlbuilder.QueryProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Repository;
-import org.springframework.util.PropertyPlaceholderHelper;
 
 import javax.validation.constraints.NotNull;
-import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Properties;
 import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.collect.ImmutableSet.copyOf;
-import static com.google.common.collect.Lists.newArrayList;
-import static com.healthmarketscience.sqlbuilder.SetOperationQuery.unionAll;
-import static java.util.Arrays.copyOf;
 import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
 import static org.opentestsystem.rdw.reporting.common.model.ReportQuery.checkIsValid;
 
@@ -40,31 +29,17 @@ import static org.opentestsystem.rdw.reporting.common.model.ReportQuery.checkIsV
 @Repository
 class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRepository {
 
-    private static final String customAggregateTemplateSql = "customAggregate";
-    private static final String customAggregateAllPermutationsTemplateSql = "customAggregateAllPermutations";
-    private static final String stateAddOn = "state";
-    private static final String districtsAddOn = "districts";
-    private static final String allDistrictsAddOn = "allDistricts";
-    private static final String schoolsAddOn = "schools";
-    private static final String allSchoolsInDistrictsAddOn = "allSchoolsInDistricts";
-
     private final NamedParameterJdbcTemplate jdbcTemplate;
-    private final QueryProvider queryProvider;
-
-    @Value("${sql.customAggregate}")
-    private String customAggregateSql;
-
-    @Value("${sql.customAggregateByEnrolledGrade}")
-    private String customAggregateByEnrolledGradeSql;
+    private final ReportQueryProvider customAggregateQueryProvider;
 
     @Value("${sql.estimateReportRowCount}")
     private String estimateReportRowCountSql;
 
     @Autowired
     public JdbcCustomAggregateReportRowRepository(final NamedParameterJdbcTemplate jdbcTemplate,
-                                                  final QueryProvider queryProvider) {
+                                                  final ReportQueryProvider customAggregateQueryProvider) {
         this.jdbcTemplate = jdbcTemplate;
-        this.queryProvider = queryProvider;
+        this.customAggregateQueryProvider = customAggregateQueryProvider;
     }
 
     private static void validate(final ReportQuery query) {
@@ -90,7 +65,7 @@ class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRep
         if (!query.getAdministrativeConditionCodes().isEmpty()) parameters.addValue("administration_condition_codes", query.getAdministrativeConditionCodes());
 
         return jdbcTemplate.query(
-                toQueryString(query),
+                customAggregateQueryProvider.toSqlQueryString(query),
                 parameters,
                 (row, index) -> ReportRow.builder()
                         .assessment(new Assessment(row.getInt("asmt_id"), row.getInt("asmt_grade_id"), row.getInt("subject_id")))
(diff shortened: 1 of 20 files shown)


aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -199,6 +199,23 @@ sqlbuilder:
...
                       g.code
 
+      # ------------ filters -------------------------------------------------------------
+            completeness:
+               clauses:
+                 where: >-
+                     cmp.code = :completeness_code
+
+                 join: >-
+                    JOIN completeness cmp ON cmp.id = fe.completeness_id
```
agorina: I would like to get rid of this JOIN and a similar one below.
In order to do it I need to convert 'code' to db id. There are a few options I see:
1. Similar to the reporting schema, store codes in the fact table. I do not think I like it because some codes are long.
2. Convert codes to IDs in the service layer. In this case I need two ReportQuery objects: one with codes and another with ids.
3. Convert codes to IDs in the web layer (argument resolver). I am not sure if this is the best practice, but if I do this I do not need two different ReportQuery objects.

Would like to know if somebody has a strong opinion about it?

common/src/test/java/org/opentestsystem/rdw/reporting/common/model/ReportQueryTest.java
```diff
@@ -1,10 +1,9 @@
-package org.opentestsystem.rdw.olap.model;
+package org.opentestsystem.rdw.reporting.common.model;
```
agorina: Moved the the same packages as the object being tested.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
```diff
@@ -71,7 +71,9 @@ public void setUp() {
                 .schoolYears(of(1999))
                 .subjectIds(of(1))
                 .assessmentType(AssessmentType.ICA)
-                .dimensionType(Overall);
+                .dimensionType(Overall)
+                .administrativeConditionCode("IN")
+                .completenessCode("Complete");
```
agorina: This is a very minimal testing, I know. Will add more tests with the next change.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAdministrationConditionRepositoryRST.java
```diff
@@ -10,7 +10,6 @@
 
 @Import(JdbcAdministrationConditionRepository.class)
 @Sql(scripts = {"classpath:OlapEntitiesSetup.sql"})
-@SuppressWarnings({"SqlNoDataSourceInspection", "SqlResolve"})
```
agorina: Unrelated, this was probably 'copy and paste', not needed. Removed here and in other test files.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAdministrationConditionRepositoryRST.java
```diff
@@ -19,10 +18,10 @@
...
     public void itShouldFindAll() {
         assertThat(repository.findAll()).usingFieldByFieldElementComparator().containsExactlyInAnyOrder(
-                CodedEntity.builder().id(-9).code("ac_code-9").build(),
-                CodedEntity.builder().id(-8).code("cc_code-8").build(),
-                CodedEntity.builder().id(-97).code("cc_code-97").build(),
-                CodedEntity.builder().id(-96).code("cc_code-96t").build()
+                CodedEntity.builder().id(-9).code("IN").build(),
+                CodedEntity.builder().id(-8).code("SD").build(),
+                CodedEntity.builder().id(-97).code("IN").build(),
+                CodedEntity.builder().id(-96).code("Valid").build()
```
agorina: I modified test values.

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -132,57 +131,42 @@ sqlbuilder:
             IEP:
               clauses:
                 select: >-
-                     sb.id AS dimension_id
-
-                join:
-                     JOIN strict_boolean sb ON sb.id = fe.iep
```
agorina: Unrelated. This JOINS are not needed.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
```diff
@@ -28,7 +28,7 @@
  * District: districts=100F,200T - optional, a comma separated list of district ids and 'True/False' for including all children (schools).
  * School Ids: schoolIds=99,300 - optional, a comma separated list of school ids
  * <p>
- * Example URL: ...?assessmentType=ICA&subjectIds=1,2&schoolYears=2017,2018&asmtGrades=4,5&state=FF&districts=15F,17F&dimensions=gender,ethnicity&schoolIds=265,266
+ * Example URL: ...?assessmentType=ICA&subjectIds=1,2&schoolYears=2017,2018&asmtGrades=4,5&state=FF&districts=15F,17F&dimensionTypes=gender,ethnicity&schoolIds=265,266
  * </p>
```
agorina: This was changes before, just updated the comment

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -21,15 +22,17 @@
...
 import java.util.Properties;
 import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.collect.ImmutableList.of;
+import static com.google.common.collect.ImmutableSet.copyOf;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.healthmarketscience.sqlbuilder.SetOperationQuery.unionAll;
 import static java.util.Arrays.copyOf;
+import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
```
jtreuting: feels like maybe there is a guava version of this instead of bring hibernate in?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
```diff
@@ -24,6 +24,8 @@
     private Map<Integer, Boolean> districts; // district ids and a flag to indicate whether to include all schools for the district
     private Set<Integer> schoolIds;
     private Set<DimensionType> dimensionTypes;
+    private String completenessCode;
```
jtreuting: I know these right now are set on the front-end as All or a single value, but some of the filters will be multiple values possible, or I can see that being requested later if not.  Just thinking out loud to see if this should be a Set to allow for this even if the front-end restricts it right now?

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
jtreuting: What is the downsize of a code being large?  Storage?  It won't really impact performance right.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Oops, I missed it. Yes, will change. Thank you.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
agorina: Good point, will change.

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
agorina: I am concerned about the performance too: comparing strings is slower than comparing integers. Although with our data set this maybe not an issue.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -145,36 +140,50 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+                queries.add(buildSql(newArrayList(allSchoolsInDistrictsAddOn), dimension, query));
             }
 
             if (!query.getSchoolIds().isEmpty()) {
-                queries.add(buildSql(of(schoolsAddOn, dimensionName), dimension));
+                queries.add(buildSql(newArrayList(schoolsAddOn), dimension, query));
             }
         }
-        return queries;
+        return queries.size() == 1 ? queries.get(0).toString() : unionAll(copyOf(queries.toArray(), queries.size(), SelectQuery[].class)).toString();
```
esotrope: this seems pretty low level. i feel like we should have another builder layer between the queries array and the repo which handles this

aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -28,54 +28,6 @@ sql:
   subject:
     findAll: select id, code from subject
 
-
-  # a tokenized version of the SQL query for the custom aggregate reports
```
agorina: I moved it into a separate file.

aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -0,0 +1,292 @@
+sqlbuilder:
```
agorina: This file was renamed.

@wtritch, please note this change (since you are changing this file too)

aggregate-service/src/main/resources/custom-aggregate-report.sql.yml
```diff
@@ -1,291 +1,47 @@
...
-            Section504:
-              clauses:
-                join: >-
-                  CROSS JOIN (select 'Section504' AS dimension, b.id AS dimension_id, b.code as dimension_code from boolean b) d
-
-            MigrantStatus:
-              clauses:
-                join: >-
-                  CROSS JOIN (select 'MigrantStatus' AS dimension, b.id AS dimension_id, b.code as dimension_code from boolean b) d
+sql:
```
agorina: I moved it into a separate file with the goal of making it easier to navigate through the queries.

agorina:
I merging but will address all the comments later.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
```diff
@@ -0,0 +1,114 @@
...
+import static com.healthmarketscience.sqlbuilder.SetOperationQuery.unionAll;
+import static java.util.Arrays.copyOf;
+
+/**
+ * This implementation is responsible for constructing a custom aggregate query.
+ * <p>
+ * It used a tokenized version of the SQL query and replaces tokens with the queries built using {@link QueryProvider}.
+ */
+@Component
+public class CustomAggregateQueryProvider implements ReportQueryProvider {
```
agorina: @esotrope , it this what you had in mind?

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAdministrationConditionRepositoryRST.java
malaffoon: 👍

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
esotrope: Yeah this helps make the repo more readable thanks

PR #745, opened 2018-01-15
DWR-1220 Add filtering by gender the aggregate report.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -110,8 +113,15 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
                 .addValue("isIncludeSchools", !query.getSchoolIds().isEmpty())
-                .addValue("school_ids", !query.getSchoolIds().isEmpty() ? query.getSchoolIds() : -1);
+                .addValue("school_ids", !query.getSchoolIds().isEmpty() ? query.getSchoolIds() : -1), query);
 
         return jdbcTemplate.queryForObject(estimateReportRowCountSql, parameterSource, Integer.class);
     }
+
+    private MapSqlParameterSource addFilterParameters(@NotNull final MapSqlParameterSource parameterSource, @NotNull final ReportQuery query) {
+        //TODO: these parameters are only applicable if gender is either a dimension or a filter. But jdbcTemplate is forgiving if unused parameters are set
+        return parameterSource
```
agorina: I will remove this comment, just wanted to point it out. 
I am debating it I should still consider adding the filters only when it is needed.
Thoughts?

aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -113,6 +113,10 @@ sqlbuilder:
 
                 join: >-
                      JOIN student s ON fe.student_id = s.id
+                     JOIN gender gn on gn.id = s.gender_id
```
agorina: Similar to the other filters, I would like to remove this JOIN, but that would mean converting the codes to DB id. Will chat about it with people offline.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
malaffoon: How hard is it to figure out if gender is either a dimension or a filter? I think we should be more correct if it isn't too hard. You never know when the JDBC driver will decide to add a "strict-parameters" option that defaults true and starts breaking things.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -110,8 +113,15 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+                .addValue("allGenders", query.getGenders().isEmpty())
```
malaffoon: I'm not sure how strongly i feel about this but mixing the naming conventions for the parameters seems wrong. Given all the other queries, i would think this would be `all_genders`.

agorina: Ok, will change.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Thank for the input. I will change it.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -110,7 +118,9 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
                 .addValue("isIncludeAllSchoolsInDistricts", !query.getDistrictIdsWithAllSchools().isEmpty())
                 .addValue("school_district_ids", !query.getDistrictIdsWithAllSchools().isEmpty() ? query.getDistrictIdsWithAllSchools() : -1)
                 .addValue("isIncludeSchools", !query.getSchoolIds().isEmpty())
-                .addValue("school_ids", !query.getSchoolIds().isEmpty() ? query.getSchoolIds() : -1);
+                .addValue("school_ids", !query.getSchoolIds().isEmpty() ? query.getSchoolIds() : -1)
```
esotrope: We have a utility function for this logic in `QueryUtils` i think

agorina: thanks, will check it out.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -24,13 +24,15 @@
...
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.nullOrEmptyToDefault;
 import static org.opentestsystem.rdw.reporting.common.model.ReportQuery.checkIsValid;
 
 @SuppressWarnings("unchecked")
 @Repository
 class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRepository {
 
-    private static final List<String> invalidCodes = ImmutableList.of(" ");
+    private static final List<String> INVALID_CODES = ImmutableList.of(" ");
+    private static final List<Integer> UNMATCHABLE_IDS = ImmutableList.of(-1);
```
agorina: @esotrope , I could not use the one in the QueryUtils since it was 'Long'. This goes back to your comment about changing school ids from Long to Integers in the reporting.

esotrope: ah I see. good to know - thanks

PR #747, opened 2018-01-15
adds organization security checks and moves auth logic into central a…

esotrope:
The approach of the PR is to add the organization security check in the java code itself as opposed to adding it to the custom aggregate SQL.
This PR makes sure that every organization referenced in the report query (school or district) is contained within the permission scope for CUSTOM_AGGREGATE_READ.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -0,0 +1,110 @@
...
+
+    // TODO assessment type entity should carry policy settings like "allow-statewide-user-to-make-aggregate-reports"
+    private static final Set<Integer> InterimAssessmentTypeIds = ImmutableSet.of(1, 2);
+    private final UserOrganizationService userOrganizationService;
+    private final boolean interimAggregateReportsForStatewideUsersEnabled;
+
+    @Autowired
+    DefaultAuthorizationService(
+            final UserOrganizationService userOrganizationService,
+            @Value("${app.interim-aggregate-reports-for-statewide-users-enabled}")
```
esotrope: Made this central auth service so that security related to this flag and assessment types isn't floating around in multiple places

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Checks to make sure the given user has permission to run aggregate reports on the given districts and schools
+     *
+     * @param user        the {@link User} creating the report
+     * @param districtIds the requested district IDs for the report
+     * @param schoolIds   the requested school IDs for the report
+     */
+    private void checkOrganizationPermissions(final User user, final Set<Integer> districtIds, final Set<Integer> schoolIds) {
```
esotrope: FYI This method will run before performing a custom aggregate report and will loop over every org in the system (in memory) once.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param user           the {@link User} creating the report
+     * @param assessmentType the requested assessment type of the report
+     */
+    private void checkAssessmentTypePermission(final User user, final AssessmentType assessmentType) {
+        if (user.getPermissionScopeByPermissionId(AggregateRead).isStatewide()
+        && !interimAggregateReportsForStatewideUsersEnabled
+        && InterimAssessmentTypeIds.contains(assessmentType.id())) {
+            throw new RuntimeException(String.format(
+                    "user (%s) is not allowed to create custom aggregate reports for assessment type (%s)",
```
esotrope: This exception should only be thrown when the requests are tampered with or someone is following a link to a report they cannot themselves generate

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -12,53 +11,63 @@
...
+                .collect(toImmutableSet());
+    }
+
     @Override
     public Set<Organization> getOrganizations(final User user, final String nameSearch) {
 
-        // fail fast
-        if (Strings.isNullOrEmpty(nameSearch)) {
+        // fail fast if search is null, empty or whitespace only
+        if (nameSearch == null || nameSearch.trim().isEmpty()) {
```
esotrope: Unrelated change that will fast-fail whitespace searches for org names.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/service/impl/DefaultAssessmentTypeServiceTest.java
```diff
@@ -1,80 +0,0 @@
...
-import java.util.List;
-
-import static com.google.common.collect.Lists.newArrayList;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-import static org.opentestsystem.rdw.olap.model.AggregateServicePermission.AggregateRead;
-import static org.opentestsystem.rdw.reporting.common.test.support.PermissionScopes.permissions;
-
-public class DefaultAssessmentTypeServiceTest {
```
esotrope: Moved these security related tests to `DefaultAuthorizationServiceTest.java`

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
malaffoon: Can we vote on the property name? I wonder if we could/should break it down, e.g. `app.aggregate-reports.statewide-interim-enabled`, or `app.reports.aggregate.statewide-interim-enabled`.

webapp/docker-compose.yml
```diff
@@ -77,21 +78,22 @@ services:
...
-      - /tmp:/tmp
-    links:
-      - config-server
-    environment:
-      - spring.profiles.active=local-docker
-      - CONFIG_SERVICE_ENABLED=true
-      - CONFIG_SERVICE_URL=http://config-server:8888
-    extra_hosts:
-      - "dwhost:$DATAWAREHOUSE_HOST"
+# uncomment to enable aggregate service
```
esotrope: @malaffoon  aggregate service is now non-default in local-docker
I think it is configured to run against redshift dev in local-docker btw

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
esotrope: not opposed. The `app` namespace is intended to mean "props specifically for this app and not a common component" but I'm not sure it communicates that well

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -0,0 +1,110 @@
...
+        final Set<Integer> userSchoolIds = newLinkedHashSet();
+        for (final Organization organization : userOrganizationService.getOrganizations(user)) {
+            if (organization.getOrganizationType() == OrganizationType.School) {
+                userSchoolIds.add((int) organization.getId());
+            } else if (organization.getOrganizationType() == OrganizationType.District) {
+                userDistrictIds.add((int) organization.getId());
+            }
+        }
+
+        for (final Integer districtId : districtIds) {
```
malaffoon: Perhaps:
```
    if (!userDistrictIds.containsAll(districtIds)) {
        throw new ...
    }
```

esotrope: Yeah that would be good. I forget java collections actually have some useful methods sometimes.

webapp/docker-compose.yml
malaffoon: Thanks.

Perhaps an additional comment explaining (briefly) how to configure redshift? Maybe a link to some documentation that i've failed to yet write?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
esotrope: Actually, only problem with that is i don't get the first ID that isn't in the collection. If that is not important for our logging then i can do that

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param user        the {@link User} creating the report
+     * @param districtIds the requested district IDs for the report
+     * @param schoolIds   the requested school IDs for the report
+     */
+    private void checkOrganizationPermissions(final User user, final Set<Integer> districtIds, final Set<Integer> schoolIds) {
+
+        // reduce user's entitled schools and districts to IDs for security check
+        final Set<Integer> userDistrictIds = newLinkedHashSet();
+        final Set<Integer> userSchoolIds = newLinkedHashSet();
```
agorina: This is unrelated to this story, but based on this morning discussion this makes me think if the UI should pass in db ids for the schools/districts or should this be changed to the natural ids?

@wtritch , @malaffoon, thoughts?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
```diff
@@ -21,19 +21,25 @@
...
+    DefaultCustomAggregateService(
+            final CustomAggregateReportRepository repository,
+            final AuthorizationService authorizationService) {
         this.repository = repository;
+        this.authorizationService = authorizationService;
     }
 
     @Override
     public List<ReportRow> findByQuery(final User user, final ReportQuery query) {
+        authorizationService.checkAuthorization(user, query);
```
agorina: This works for the cases when the individual school or district ids are passed in.
What about the use when a query has: `all schools within a district` or `all districts within a state`.  Would will allow this choice(s) ONLY when a user has access to the entire district or state, or do we need to consider it in the query?

Tagging @jtreuting  into this conversation.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
esotrope: should say the thing that is nice about IDs is that it is the smallest (in memory) handle on the data that can go over the wire - aren't the natural IDs long strings?

malaffoon: Hmm. Good question. How do we handle students in the reporting side?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
esotrope: In the event of all schools within a district -> I think the current impl will work so long as the following statement is true: if the user has district rights then they have the right to see individual school data of that distict

For the `all districts within a state` flag, I think we can add security for this by checking if the user has statewide permission correct?

agorina: About your first comment: in the UI, if a user has access to let's say three schools, can a user search for the District name? And if yes, then can the user then add a district to the query meaning - all schools within a district **that I have access to**?

esotrope: That is a great question. Right now it does not work this way (though we may want it to).
Currently the `/organizations` does not return districts for which the user has schools.

esotrope:
Based on team discussion I think this implementation is good as far as security is concerned so I am going to merge this change.

PR #754, opened 2018-01-17
adds organization repo that works with the olap schema

esotrope:


aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -19,6 +20,31 @@ sql:
...
+          district.id,
+          district.name,
+          district.natural_id,
+          'District' as org_type,
+          sch.district_group_id,
+          null as district_id,
+          null as school_group_id
+        FROM district
+          JOIN school sch
+            ON sch.district_id = district.id
```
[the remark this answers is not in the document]
agorina: Just to make sure I understand it: once we add `district.district_group_id` this needs an adjustment?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
```diff
@@ -33,15 +33,13 @@
...
-        // authorizationService.checkAuthorization(user, query);
+        authorizationService.checkAuthorization(user, query);
         return repository.findByQuery(ensureDimension(query, Overall));
     }
 
     @Override
     public int estimateReportRowCount(final User user, final ReportQuery query) {
-        // TODO re-enable security when organization repository is fixed
-        // authorizationService.checkAuthorization(user, query);
+        authorizationService.checkAuthorization(user, query);
```
agorina: I think DefaultCustomAggregateServiceTest needs to be modified to check for this call (here and above).

aggregate-service/src/main/resources/application.sql.yml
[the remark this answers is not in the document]
esotrope: yeah we would not need the join.
I saw this after our conversation btw, looks like we wont need the schema change if we are fine with this join here too.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
esotrope: good idea

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateExportService.java
```diff
@@ -57,26 +58,33 @@
...
 
         //Report CSV will always be in US English
         this.locale = Locale.US;
     }
 
     @Override
     public ResponseEntity<ByteArrayResource> exportReport(final User user, final ReportQuery query) {
+
+        // Ensure user has authority to export report
+        authorizationService.checkAuthorization(user, query);
```
esotrope: report export now has security

esotrope:
Going to merge so we have the functionality
I'll incorporate any suggestions in a followup PR

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/CachingOrganizationService.java
```diff
@@ -28,18 +25,7 @@
...
-//                PermissionScope.STATEWIDE,
-//                OrganizationQuery.builder()
-//                        .types(ImmutableSet.of(
-//                                OrganizationType.District,
-//                                OrganizationType.School
-//                        ))
-//                        .limit(Integer.MAX_VALUE)
-//                        .build()
-//        ));
+        return ImmutableSet.copyOf(repository.findAllSchoolsAndDistricts());
```
malaffoon: I find it odd that the service layer is making a copy of the repository result. Is this just to switch from List to Set? Perhaps we should be consistent in the layers wrt to the semantics of the collection?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateExportService.java
```diff
@@ -57,26 +58,33 @@
...
+
         final List<ReportRow> results = service.findByQuery(user, query);
 
         final byte[] csvContent = printCsv(results, query);
         final ByteArrayResource resource = new ByteArrayResource(csvContent);
         final MediaType mediaType = MediaTypes.TEXT_CSV;
```
malaffoon: I'd recommend the mediaType, possibly the extension, and the charset all be either static class members or initialized in the constructor. 

Also, it is inconsistent to use `Charset.forName("UTF-8")` here and `Charsets.UTF_8` a few lines down. If it is a shared class member that will fix it; otherwise i'd vote for `Charsets.UTF_8`.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/CachingOrganizationService.java
esotrope: it is to switch from a mutable list to an immutable set : /

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateExportService.java
esotrope: I think I'll let @wtritch address this - i think he's actively working this code.

PR #756, opened 2018-01-17
Feature/dwr 1305 writing trait scores

jtreuting:
This adds the writing trait scores to the ExamItems returned to show student responses on the front-end.  It also adds the performance task writing type to the AssessmentItem returned from the same API call.


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItem.java
```diff
@@ -60,6 +68,7 @@ public static Builder builder() {
         private Integer points;
         private int position;
         private String response;
+        private WritingTraitScores writingTraitScores;
```
jtreuting: went back and forth on making a generic SubScores list of Integers or having 3 properties (one for each writing trait score) but decided that having a specific class felt cleaner and would decrease the data sent over the wire since this can be null most of the time (there aren't many WER items).  but let me know if you see it differently.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItems.java
```diff
@@ -21,6 +21,15 @@ public static ExamItem map(final ResultSet row) throws SQLException {
         if (hasColumn(row, "response")) {
             builder.response(row.getString("response"));
         }
+
+        Integer evidenceTraitScore = getNullable(row, row.getInt("trait_evidence_elaboration_score"));
+        Integer organizationTraitScore = getNullable(row, row.getInt("trait_organization_purpose_score"));
+        Integer conventionsTraitScore = getNullable(row, row.getInt("trait_conventions_score"));
+
+        if (evidenceTraitScore != null || organizationTraitScore != null || conventionsTraitScore != null) {
```
jtreuting: i want the entire object to be null if there are no traits instead of a class with 3 null properties.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -51,18 +51,7 @@ sql:
...
-          ei.score,
-          ei.position,
-          ei.response
-        from exam_item ei
-          join exam e on ei.exam_id=e.id
-          join student st on st.id = e.student_id
-          join asmt a on a.id = e.asmt_id
-          join school s on e.school_id=s.id
-          join school isch on st.inferred_school_id = isch.id
+        ${sql.reporting.snippet.selectFromExamItem}
```
jtreuting: I found this was duplicate and not consolidated into the snippet like it was in other places so made this change since i needed the 3 trait scores in this query as well and i added them to the snippet.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/assessment/JdbcAssessmentItemRepositoryIT.java
```diff
@@ -85,14 +85,30 @@ public void findAllForAssessmentShouldFindAllItems() throws Exception {
...
+                                .targetCode("t3")
+                                .difficultyCode("M")
+                                .depthOfKnowledgeCode("-1_1")
+                                .mathPracticeCode("-1")
+                                .maximumPoints(5)
+                                .commonCoreStandardIds(of("S-ID.4", "S-ID.7"))
+                                .optionsCount(0)
+                                .answerKey("Hand")
+                                .type("WER")
+                                .performanceTaskWritingType("Narrative")
```
jtreuting: new test data to check the performance task writing type

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/group/JdbcGroupExamItemRepositoryIT.java
```diff
@@ -91,6 +92,7 @@ public void findAllExamItemScoresForAssessmentShouldReturnAllItems() throws Exce
                                 .points(1)
                                 .position(2)
                                 .response("D")
+                                .writingTraitScores(new WritingTraitScores(2, 3, 1))
```
jtreuting: new test data was included to check the writing trait scores are populated for some of them and null for others

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItem.java
malaffoon: 👍  for nullable. Perhaps note in the getter javadocs that it can be null.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItems.java
```diff
@@ -21,6 +21,15 @@ public static ExamItem map(final ResultSet row) throws SQLException {
         if (hasColumn(row, "response")) {
             builder.response(row.getString("response"));
         }
+
+        Integer evidenceTraitScore = getNullable(row, row.getInt("trait_evidence_elaboration_score"));
+        Integer organizationTraitScore = getNullable(row, row.getInt("trait_organization_purpose_score"));
+        Integer conventionsTraitScore = getNullable(row, row.getInt("trait_conventions_score"));
```
malaffoon: nit: these can all be final.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItems.java
malaffoon: Not saying it is better/worse, just different: you could put this logic in the builder itself, then all this can be a single inlined call.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/WritingTraitScores.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @return the points earned for the conventions trait or <code>null</code> if this trait was not scored
+     */
+    public Integer getConventions() {
+        return conventions;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
malaffoon: equals/hash really needed?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/WritingTraitScores.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Represents the writing trait scores for an item
+ */
+public class WritingTraitScores {
+
+    private Integer evidence;
+    private Integer organization;
+    private Integer conventions;
+
+    public WritingTraitScores(Integer evidence, Integer organization, Integer conventions) {
```
malaffoon: nit: args can be final

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/WritingTraitScores.java
jtreuting: this was added because of the test (which maybe isn't a great reason, but i didn't see the harm).  the test calls `usingFieldByFieldElementComparator` and so it does an equality check on the 'WritingTraitScores' object.  There is probably a way to change the test to check equality of the results a different way but this seemed harmless to me.

malaffoon: That's actually why i asked -- adding these just for the unit test shouldn't be necessary. OTOH it is, as you say, relatively harmless.

malaffoon: nit: param can be final

jtreuting: good point. i found `usingRecursiveFieldByFieldElementComparator` that seems to do the trick.  so i'll remove the equals.  this will drill into the WritingTraitScores property and check each field it has.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/ExamItem.java
```diff
@@ -86,13 +97,25 @@ public Builder response(final String response) {
             return this;
         }
 
+        public Builder writingTraitScores(final Integer evidence, final Integer organization, final Integer conventions) {
```
esotrope: I find this a little confusing, shouldn't `WritingTraitScores` have a builder and we pass in the built object here?

jtreuting: doesn't a builder for a 3 parameter object not make sense?  we decide the cutoff was 4 or 5 right?

esotrope: a cutoff of 3 inclusive is what i have been going by. either way this setter should accept a `WritingTraitScores` object i think

jtreuting: that is how i originally had it (after going back and forth) but then the logic of whether to create an object or provide null is in the mapper and @malaffoon made the good point that that logic should probably be inside the builder.  I guess there is an argument that if you pass in a builder to this method it could decide whether to return null or an object that is populated based on the data, but that seemed like overkill to me when there are only 3 parameters.  @malaffoon, @wtritch any thoughts?

wtritch: In general we build each object independently and send the built object/null into the parent builder.  We've generally left a full object with null values (the json looks like `{... "writingTraitScores": {} ...}` rather than `{... }`)  
Having a multi-parameter builder method feels a little weird...

malaffoon: My opinion is that the builder should have the methods necessary to make the calling code easier, especially if there are data constraint issues involved. The only time i don't like special methods in the builder is if they involved one-off logic that is specific to an environment, e.g. i wouldn't have a builder method that takes a ResultSet and extracts data from columns.

So, in this case, a convenience method that takes the three writing trait scores and does what is necessary is good because it makes the calling code simpler and may be needed in more than one place.

jtreuting: so we all agree that we disagree.  got it.  ;) . thanks.  i think i'm going to keep it as-is because i like the cleanliness of consuming it after I previously passed in a object.

PR #757, opened 2018-01-17
Task/zuul content type cleaning

wtritch:
This adds a Zuul pre-filter that is a little more forgiving of slightly invalid Content-Type headers with an extra space.


webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.stereotype.Component;
+
+import javax.servlet.http.HttpServletRequest;
+
+import static com.netflix.zuul.context.RequestContext.getCurrentContext;
+
+/**
+ * This filter is responsible for cleaning extra whitespace from Content-Type
+ * header values.
```
esotrope: can we add a link to the zuul spec here?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public int filterOrder() {
+        return 0;
+    }
+
+    @Override
+    public boolean shouldFilter() {
+        final RequestContext context = getCurrentContext();
+        return context.getRequest().getContentType() != null && context.getRequest().getContentType().contains(" ");
+    }
```
wtritch: I only run the filter IFF a content-type header exists with a space in it.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private static class ContentTypeRequestWrapper extends Servlet30RequestWrapper {
+
+        ContentTypeRequestWrapper(final HttpServletRequest request) {
+            super(request);
+        }
+
+        @Override
+        public String getContentType() {
+            return MediaType.valueOf(getRequest().getContentType()).toString();
```
wtritch: Spring's MediaType implementation is much more forgiving of slightly invalid header values.

jtreuting:
Is this only relevant on the load testing side, or would we expect this to impact normal usage?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public int filterOrder() {
+        return 0;
+    }
+
+    @Override
+    public boolean shouldFilter() {
+        final RequestContext context = getCurrentContext();
+        return context.getRequest().getContentType() != null && context.getRequest().getContentType().contains(" ");
```
esotrope: should this be more inclusive?
like `/\s/g.test(contenttype)` ?

wtritch: hmmm for tabs and what-not?  Sounds reasonable.

wtritch:
@jtreuting Since the test on whether or not to actually filter is very fast, and this is the second time I've run into this issue I think it's a reasonable "good thing" to add to the project permanently as opposed to a one-off "just for load-testing" hack.

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    private static class ContentTypeRequestWrapper extends Servlet30RequestWrapper {
+
+        ContentTypeRequestWrapper(final HttpServletRequest request) {
+            super(request);
+        }
+
+        @Override
+        public String getContentType() {
```
malaffoon: Why do you need this helper class? Wouldn't Servlet30RequestWrapper get it done?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
wtritch: I'm not sure exactly what you mean?  What would you link to?

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
malaffoon: I really think we need a code comment in here somewhere. Something like, `This wrapper relies on MediaType to parse and format the content-type which might not strictly comply with the spec, which is what the zuul parser requires."

PR #758, opened 2018-01-17
Enable general runtime exception request logging

wtritch:
This is a fix discussed earlier with @agorina and @esotrope to enable general runtime exception request handler logging in our services.

Currently, if there is an exception during request processing that isn't explicitly handled and logged the request silently fails with no logging.  Examples include SQLExceptions, Request parsing failures, etc.

I'm hitting this currently with Bad Request responses for aggregate reports with no indication of where the error lies: parameters, SQL, random NPE, etc.
This may lead to double-logging of some exceptions, but personally I'd rather double-log than never log at all.

### DIFF
--- a/common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
+++ b/common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
@@ -1,5 +1,7 @@
 package org.opentestsystem.rdw.reporting.common.web;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.boot.context.config.ResourceNotFoundException;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
@@ -17,6 +19,7 @@ import java.util.NoSuchElementException;
  */
 @ControllerAdvice
 public class ExceptionHandlerControllerAdvice extends ResponseEntityExceptionHandler {
+    private static final Logger logger = LoggerFactory.getLogger(ExceptionHandlerControllerAdvice.class);
 
     /**
      * @param exception original exception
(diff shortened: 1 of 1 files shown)


common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
```diff
@@ -54,6 +57,7 @@
      */
     @ExceptionHandler(RuntimeException.class)
     protected ResponseEntity<Object> handleRuntimeExceptions(final RuntimeException exception, final WebRequest request) {
+        logger.warn("Problem handling request", exception);
```
wtritch: I could be talked into dropping this to "debug" and enabling on a per-service basis via config.  Thoughts?

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
```diff
@@ -25,6 +28,7 @@
      */
     @ExceptionHandler(IllegalArgumentException.class)
     protected ResponseEntity<Object> handleIllegalArgumentException(final RuntimeException exception, final WebRequest request) {
+        logger.warn("Problem handling request", exception);
         return opaqueError(exception, request, HttpStatus.BAD_REQUEST);
```
wtritch: If we throw a checkArgument(blah != blah, "some message here") I'd like to see the exception in the logs.

wtritch:
I'm going to merge this for now to help with debugging and will resolve comments as they come up.

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
esotrope: Warn level is something that is supposed to be actionable by TO right? Maybe info or debug would be good.

PR #761, opened 2018-01-18
Feature/aggregate reports form

esotrope:
Piecemeal WIP updates to the aggregate report form based on the recent feedback.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
@@ -1,0 +1,84 @@
+import { AggregateReportOptions } from "./aggregate-report-options";
+import { AggregateReportFormOptions } from "./aggregate-report-form-options";
+import { Injectable } from "@angular/core";
+import { Option as SbToggleOption } from "../shared/sb-toggle.component";
+import { Option as SbCheckboxGroupOption } from "../shared/form/sb-checkbox-group.component";
+import { TranslateService } from "@ngx-translate/core";
+import { SchoolYearPipe } from "../shared/format/school-year.pipe";
+
+const localeCompare = property => (a, b) => a[ property ].localeCompare(b[ property ]);
+const localeCompareText: (a: any, b: any) => number = (a, b) => a.text.localeCompare(b.text);
+
+@Injectable()
+export class AggregateReportFormOptionsMapper {
+
+  constructor(private translate: TranslateService,
+              private schoolYearPipe: SchoolYearPipe) {
+  }
+
+  map(options: AggregateReportOptions): AggregateReportFormOptions {
+
+    return <AggregateReportFormOptions>{
+      assessmentGrades: options.assessmentGrades
+        .map(entity => <SbCheckboxGroupOption>{
+          value: entity,
+          text: this.translate.instant(`common.grade.${entity.code}.form-name`),
+          analyticsProperties: { label: `Assessment Grade: ${entity.code}` }
+        })
+        .sort((a, b) => a.value.id - b.value.id),
+      assessmentTypes: options.assessmentTypes
+        .map(entity => <SbToggleOption>{
+          value: entity,
+          text: this.translate.instant(`common.assessment-type.${entity.code}.short-name`)
+        })
+        .sort(localeCompareText),
+      completenesses: options.completenesses
+        .map(entity => <SbCheckboxGroupOption>{
+          value: entity,
+          text: this.translate.instant(`common.completeness.${entity.code}`),
+          analyticsProperties: { label: `Completeness: ${entity.code}` }
+        }),
+      interimAdministrationConditions: options.interimAdministrationConditions
+        .map(entity => <SbCheckboxGroupOption>{
+          value: entity,
+          text: this.translate.instant(`common.administration-condition.${entity.code}`),
+          analyticsProperties: { label: `Manner of Administration: ${entity.code}` }
+        }),
+      schoolYears: options.schoolYears
+        .map(year => <SbCheckboxGroupOption>{
+          value: year,
+          text: this.schoolYearPipe.transform(year),
+          analyticsProperties: { label: `Assessment Grade: ${year}` }
+        })
+        .sort((a, b) => b.value - a.value),
+      subjects: options.subjects
+        .map(entity => <SbCheckboxGroupOption>{
+          value: entity,
+          text: this.translate.instant(`common.subject.${entity.code}.short-name`),
+          analyticsProperties: { label: `Subject: ${entity.code}` }
+        })
+        .sort((a, b) => a.text.localeCompare(b.text)),
+      summativeAdministrationConditions: options.summativeAdministrationConditions
+        .map(entity => <SbCheckboxGroupOption>{
+          value: entity,
+          text: this.translate.instant(`common.administration-condition.${entity.code}`),
+          analyticsProperties: { label: `Validity: ${entity.code}` }
+        }),
+    };
+
+    // return this.dataService.get(`${ServiceRoute}/reportOptions`)
+    //   .map(options => <AggregateReportOptions>{
+    //     ethnicities: options.ethnicities
+    //       .map(mapper(code => `labels.enum.ethnicity.${code}`))
+    //       .sort(localeCompareText),
+    //     genders: options.genders
+    //       .map(mapper(code => `labels.enum.gender.${code}`))
+    //       .sort(localeCompareText),
+    //     schoolYears: options.schoolYears
+    //       .sort((a, b) => a - b),
+    //   });
+
+  }
+
+
+}
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
@@ -6,9 +6,6 @@ import { Option as SbCheckboxGroupOption } from "../shared/form/sb-checkbox-grou
 import { TranslateService } from "@ngx-translate/core";
 import { SchoolYearPipe } from "../shared/format/school-year.pipe";
 
-const localeCompare = property => (a, b) => a[ property ].localeCompare(b[ property ]);
-const localeCompareText: (a: any, b: any) => number = (a, b) => a.text.localeCompare(b.text);
-
 @Injectable()
 export class AggregateReportFormOptionsMapper {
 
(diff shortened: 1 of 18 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
```diff
@@ -0,0 +1,68 @@
+import { AggregateReportOptions } from "./aggregate-report-options";
+import { AggregateReportFormOptions } from "./aggregate-report-form-options";
+import { Injectable } from "@angular/core";
+import { Option as SbToggleOption } from "../shared/sb-toggle.component";
+import { Option as SbCheckboxGroupOption } from "../shared/form/sb-checkbox-group.component";
+import { TranslateService } from "@ngx-translate/core";
+import { SchoolYearPipe } from "../shared/format/school-year.pipe";
+
+@Injectable()
+export class AggregateReportFormOptionsMapper {
```
esotrope: So it is a little bit convoluted but basically there are report options (backend provided) which are mapped to client-friendly translated and sorted options that can be fed directly into our ui components.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  map(options: AggregateReportOptions): AggregateReportFormOptions {
+
+    return <AggregateReportFormOptions>{
+      assessmentGrades: options.assessmentGrades
+        .map(entity => <SbCheckboxGroupOption>{
+          value: entity,
+          text: this.translate.instant(`common.grade.${entity.code}.form-name`),
+          analyticsProperties: { label: `Assessment Grade: ${entity.code}` }
```
esotrope: @jtreuting  these are guesses on these analytics labels. I noticed we run these through the translate pipe in other places but i'm not sure why as these should probably not change when the language changes right?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  constructor(private dataService: CachingDataService) {
+  }
+
+  getReportOptions(): Observable<AggregateReportOptions> {
+    return this.dataService.get(`${ServiceRoute}/reportOptions`)
+      .map(options => <AggregateReportOptions>{
+        assessmentGrades: options.assessmentGrades,
+        assessmentTypes: options.assessmentTypes,
+        completenesses: [
```
esotrope: A future API will deliver completeness, interimAdministrationConditions and summativeAdministrationConditions sorted presumably?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -1,5 +1,178 @@
-<div class="well">
-  <h3>Sample Aggregate Report Landing Page</h3>
-  <p><a routerLink="results">Sample Results View</a></p>
-  <p><a routerLink="query-builder">Query Builder</a></p>
```
esotrope: overwriting the `/aggregate-reports` page here with the new form

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -1,5 +1,178 @@
...
+        <div class="well">
+          <div class="row">
+            <div class="col-md-6 pr-0">
+              <label info-button
+                     title="{{'labels.aggregate-reports.query-builder.assessment-attributes.subjects.label' | translate}}"
+                     content="{{'labels.aggregate-reports.query-builder.assessment-attributes.subjects.info' | translate}}"></label>
+            </div>
+            <div class="col-md-6">
+              <div class="pull-right">
+                <sb-checkbox-group [horizontal]="true"
```
esotrope: changed a bunch of these things that were radio-groups to checkbox groups so they would all feel consistent. it is also in anticipation that different tenants may have more than 2 values as the radio-group only makes sense when there are two possible values. Maybe we need logic to change the behavior for 2 value button groups vs 3+?

webapp/src/main/webapp/src/app/shared/form/sb-checkbox-group.component.ts
```diff
@@ -0,0 +1,179 @@
...
+                 [angularticsCategory]="analyticsCategory"
+                 [angularticsProperties]="option.analyticsProperties">
+          {{option.text}}
+        </label>
+      </div>
+    </div>
+  `,
+  providers: [ CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR ]
+})
+export class SBCheckboxGroup implements ControlValueAccessor, OnInit {
```
esotrope: This is a rework of the existing `SBCheckboxList` to allow for `[(ngModel)]` binding and more flexible option values. This is necessary to get the flexibility of holding an object as the value of an option selection.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -18,6 +18,164 @@
     }
   },
   "collapse": "Collapse",
+  "common": {
```
esotrope: WIP refactor of universal text into "common" prefix.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -18,6 +18,164 @@
...
+    "completeness": {
+      "Partial": "Partial",
+      "Complete": "Complete"
+    },
+    "grade": {
+      "01": {
+        "enrolled-name": "G1",
+        "long-name": "Grade 1",
+        "short-name": "G1",
+        "form-name": "1"
```
esotrope: This is a new attribute for the assessment grade checkbox-group.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
jtreuting: you mean grade specific?  probably would be a number, but i guess you could have someone want to do 3rd, 4th, 5th, etc.  and then maybe that is different in other languages?

esotrope: I was trying to say - if we want our ananlytics data to be consistent and line up no matter the user-language we should construct labels with hard-coded label prefixes and DB entity codes instead of something that goes through the translation pipeline. in the filters on other result pages we do a translation lookup for the labels.
would you agree or should this be changed to work like the existing analytics labels?

jtreuting: oh sorry.  yeah for Analystics labels they should be the same so that we have it documented as to what the events are when viewing data in Google Analytics.  Yeah no need to translate those.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -1,5 +1,178 @@
...
+                container="body">
+          <i class="fa fa-table"></i> {{'labels.export.csv' | translate }}
+        </button>
+        <button class="btn btn-success btn-sm ml-xs green"
+                analyticsEvent="GenerateReport"
+                analyticsCategory="AggregateQueryBuilder"
+                angulartics2on="click"
+                container="body">
+          <i class="fa fa-gears"></i> {{'labels.aggregate-reports.query-builder.customize.generate-report-button' | translate}}
+        </button>
```
wtritch: Should this be enabled/disabled based upon validity? (If you haven't selected any orgs, for instance)

webapp/src/main/webapp/src/app/shared/form/sb-checkbox-group.component.ts
```diff
@@ -0,0 +1,179 @@
...
+  public analyticsEvent: string;
+
+  @Input()
+  public analyticsCategory: string;
+
+  @Input()
+  public allOptionAnalyticsProperties: any = {};
+
+  private _options: Option[];
+  private _value: any[] = [];
```
wtritch: I'm having a disconnect between _value being an array and Option.value being singular... ?

esotrope: yeah same - however i wanted to stick to html convention of `<input>`-like components having a `value` property

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
esotrope: yes it should for sure.

jtreuting: yeah i think that makes sense.  we know that there are a set of things that have to be selected.  it might just be organization, assessment type since there are defaults for everything else i think (does grade have a default?  don't remember)

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.ts
```diff
@@ -1,9 +1,41 @@
...
+    const request: AggregateReportRequest = new AggregateReportRequest();
+    request.assessmentGrades = valuesOf(options.assessmentGrades);
+    request.assessmentType = firstValueOf(options.assessmentTypes);
+    request.completenesses = valuesOf(options.completenesses);
+    request.interimAdministrationConditions = valuesOf(options.interimAdministrationConditions);
+    request.schoolYears = [ firstValueOf(options.schoolYears) ];
+    request.subjects = valuesOf(options.subjects);
+    request.summativeAdministrationConditions = valuesOf(options.summativeAdministrationConditions);
+    return request;
+  }
```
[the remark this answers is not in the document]
esotrope: not my intention no, this has to do with initializing the state of the form based on the available options.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
esotrope: correct. so in the other forms we use radio for [all, a, b] kind of options (like for subject) however this will kind of break if we introduce another subject or introduce a null/"not stated" option for migrant status for instance - hence the change to use checkbox regardless

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
esotrope: currently grade defaults to all, we probably have to know something about the user to have a default grade right? Or maybe we need to introduce a default state of no-selection and add prompt text says you have to select something.

PR #762, opened 2018-01-18
Replace UNION ALL with java logic.

agorina:
This is based on the latest load testing findings. The goal is to reduce the number of SQL query variations.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryProvider.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryProvider.java
@@ -4,14 +4,14 @@ import org.opentestsystem.rdw.reporting.common.model.ReportQuery;
 
 
 /**
- * Implementation of this interface is responsible for converting a {@link ReportQuery} to a sql query string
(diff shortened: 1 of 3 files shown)


agorina:
Merging, will address comments if needed.

jtreuting:
the idea is that if there are 3 queries they will have the same compiled query under the hood and be much faster?

agorina:
Yup, that is the idea.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryProvider.java
```diff
@@ -4,14 +4,14 @@
 
 
 /**
- * Implementation of this interface is responsible for converting a {@link ReportQuery} to a sql query string
+ * Implementation of this interface is responsible for converting a {@link ReportQuery} to a sql query strings
  */
 public interface ReportQueryProvider {
     /**
      * Converts a given {@link ReportQuery} to a sql query string
```
malaffoon: Perhaps some explanation of what the returned collection of sql is?

agorina: Will do.

PR #763, opened 2018-01-18
Feature/change agg report sqlt to fixed size in clause

agorina:
This change is to test our theory that Redshift requires the same size of the IN clause in order to reuse the query plan.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
@@ -146,8 +146,8 @@ class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRep
         if (!query.isIncludeAllDistricts() && !query.getDistrictIds().isEmpty()) parameters.addValue("district_ids", padToSize(query.getDistrictIds(), 5));
         if (!query.getDistrictIdsWithAllSchools().isEmpty()) parameters.addValue("school_district_ids", padToSize(query.getDistrictIdsWithAllSchools(), 5));
         if (!query.getSchoolIds().isEmpty()) parameters.addValue("school_ids", padToSize(query.getSchoolIds(), 5));
-        if (!query.getCompletenessIds().isEmpty()) parameters.addValue("completeness_ids", padToSize(query.getCompletenessIds(), 3));
-        if (!query.getAdministrativeConditionIds().isEmpty()) parameters.addValue("administration_condition_ids", padToSize(query.getAdministrativeConditionIds(), 3));
+        if (!query.getCompletenessIds().isEmpty()) parameters.addValue("completeness_ids", padToSize(query.getCompletenessIds(), 2));
+        if (!query.getAdministrativeConditionIds().isEmpty()) parameters.addValue("administration_condition_ids", padToSize(query.getAdministrativeConditionIds(), 4));
 
         if (query.contains(DimensionType.Gender) || !query.getGendersId().isEmpty()) {
             parameters
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
@@ -138,52 +138,66 @@ class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRep
 
     private MapSqlParameterSource buildParameters(final ReportQuery query) {
         final MapSqlParameterSource parameters = new MapSqlParameterSource()
-                .addValue("school_years", query.getSchoolYears())
-                .addValue("asmt_grade_ids", query.getAsmtGradeIds())
-                .addValue("subject_ids", query.getSubjectIds())
+                .addValue("school_years", padToSize(query.getSchoolYears(), 5))
+                .addValue("asmt_grade_ids", padToSize(query.getAsmtGradeIds(), 5))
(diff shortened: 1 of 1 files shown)


agorina:
Once again, merging so that we could use it for testing. Will address comments, but this is most likely will need some changes based on the test results.

esotrope:
interesting

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -138,52 +138,66 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+    private static List<Integer> padToSize(final Iterable<Integer> collection, final int size) {
+        return padToSize(newArrayList(collection), size, -1);
+    }
+
+    private static <T> List<T> padToSize(final List<T> collection, final int size, final T defaultValue) {
+        if (collection != null && collection.size() >= size) return collection;
+
+        final List<T> paddedCollection = (collection == null) ? newArrayList() : collection;
+        for (int i = paddedCollection.size(); i < size; i++) {
+            paddedCollection.add(defaultValue);
```
jtreuting: if it feels like it wouldn't be hard for it to be smart enough to compile `IN (1, 2, -1, -1)` to IN (1, 2, -1)`  so if that is the case you might need to do:

`paddedCollection.add(defaultValue*i);`

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -138,52 +138,66 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
 
     private MapSqlParameterSource buildParameters(final ReportQuery query) {
         final MapSqlParameterSource parameters = new MapSqlParameterSource()
-                .addValue("school_years", query.getSchoolYears())
-                .addValue("asmt_grade_ids", query.getAsmtGradeIds())
-                .addValue("subject_ids", query.getSubjectIds())
+                .addValue("school_years", padToSize(query.getSchoolYears(), 5))
+                .addValue("asmt_grade_ids", padToSize(query.getAsmtGradeIds(), 5))
```
wtritch: If the UI selects all grades by default, should this be larger? 3-8 + 11 = 7 options?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: You are absolutely right, from the EXPLAIN plan with the `grade_id in (7,8,5,-1,-1,-1)`:
```
Filter: (((grade_id = 5) OR (grade_id = 7) OR (grade_id = 8) OR (grade_id = -1)) ...
```

This may explain why it needs the same number of IN parameters.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Good point, will change.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -138,52 +138,66 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+                    .addValue("economic_disadvantage_ids", padToSize(query.getEconomicDisadvantageIds(), 2));
         }
         return parameters;
     }
+
+    private static List<Integer> padToSize(final Iterable<Integer> collection, final int size) {
+        return padToSize(newArrayList(collection), size, -1);
+    }
+
+    private static <T> List<T> padToSize(final List<T> collection, final int size, final T defaultValue) {
```
malaffoon: I don't quite understand the need to break out the helper but maybe i'm missing where it is used without -1 as the default value.

It would be more efficient to do the size check on the original collection before creating a new collection. I think it would be more efficient to create a new collection with the default values for every position then replace the values from the original list (rather than adding to the collection multiple times). And there should be a const for the unmatchable id. Overall:
```
    private static Collection<Integer> padToSize(final Collection<Integer> collection, final int size) {
        if (collection != null && collection.size() >= size) return collection;

        final List<Integer> padded = Collections.nCopies(size, UNMATCHABLE_ID);
        if (collection != null) {
            final Iterator<Integer> iterator = collection.iterator();
            for (int i = 0; i < collection.size(); ++i) {
                padded.set(i, iterator.next());
            }
        }
        return padded;
    }
```

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
malaffoon: I don't understand the change. The optimization happens during query compilation not during the trip through JDBC. The query passed to redshift will just have a bunch of question marks, `IN (?, ?, ?, ?)`. Did you really establish that redshift doesn't compile the parameterized query first then do optimization?

agorina: When I ran the SQL bypassing java and put multiple -1, the optimizer removes multiple occurrences of -1 and kept only one of them. Since I do not know where the optimization happens, to be safe, I have changed to pad the IN clause with different values. 

If we have time to research if using the same value still work, we can do it.

PR #766, opened 2018-01-19
Feature/report features aggregate form section

esotrope:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
@@ -109,7 +109,19 @@ export class AggregateReportFormOptionsMapper {
           value: value,
           text: this.translate.instant(`common.form-field-boolean-values.${value}`),
           analyticsProperties: { label: `Economic Disadvantage: ${value}` }
-        })
+        }),
+      achievementLevelDisplayTypes: [ 'Separate', 'Grouped' ]
+        .map(value => <SbCheckboxGroupOption>{
+          value: value,
+          text: this.translate.instant(`labels.aggregate-reports.form.field.achievement-level-display-type.value.${value}`),
+          analyticsProperties: { label: `Achievement Level Display Type: ${value}` }
+        }),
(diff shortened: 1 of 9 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
```diff
@@ -109,7 +109,19 @@ export class AggregateReportFormOptionsMapper {
...
+          value: value,
+          text: this.translate.instant(`labels.aggregate-reports.form.field.achievement-level-display-type.value.${value}`),
+          analyticsProperties: { label: `Achievement Level Display Type: ${value}` }
+        }),
+      valueDisplayTypes: [ 'Percent', 'Number' ]
+        .map(value => <SbCheckboxGroupOption>{
+          value: value,
+          text: this.translate.instant(`labels.aggregate-reports.form.field.value-display-type.value.${value}`),
+          analyticsProperties: { label: `Value Display Type: ${value}` }
+        }),
```
esotrope: used code arrays instead of booleans for these two state options. not sure if it makes a difference but it made defining the analytics properties portion more straight forward.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -105,7 +105,7 @@ <h2 class="h3 gray-darkest">
             </div>
           </div>
         </div>
-        <div class="well">
+        <div class="well" [hidden]="summativeFieldsDisabled">
```
esotrope: went ahead and made the form hide the summative/interim only fields in respect to the currently selected assessment type. not sure if this is what we want or not. we also have the option to disable the form item but that didn't look very good.

jtreuting: yes this is what we want.

There should also be a configuration flag that says summative is not available at all. 
 For example CA won't be using the summative so it should be removed as an option completely, and this field shouldn't show at all in that case either.

esotrope: Ok i'll try and work that in

PR #767, opened 2018-01-19
Changed aggregate report SQL to a fixed size IN clause.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -139,7 +139,7 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
     private MapSqlParameterSource buildParameters(final ReportQuery query) {
         final MapSqlParameterSource parameters = new MapSqlParameterSource()
                 .addValue("school_years", padToSize(query.getSchoolYears(), 5))
-                .addValue("asmt_grade_ids", padToSize(query.getAsmtGradeIds(), 5))
+                .addValue("asmt_grade_ids", padToSize(query.getAsmtGradeIds(), 7))
```
agorina: Changed per code review feedback.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -191,12 +191,13 @@ private MapSqlParameterSource buildParameters(final ReportQuery query) {
...
-    private static <T> List<T> padToSize(final List<T> collection, final int size, final T defaultValue) {
+    private static List<Integer> padToSize(final List<Integer> collection, final int size, final int defaultValue) {
         if (collection != null && collection.size() >= size) return collection;
 
-        final List<T> paddedCollection = (collection == null) ? newArrayList() : collection;
+        final List<Integer> paddedCollection = (collection == null) ? newArrayList() : collection;
         for (int i = paddedCollection.size(); i < size; i++) {
-            paddedCollection.add(defaultValue);
+            //TODO: change this if load test proves that we need it
+            paddedCollection.add(defaultValue*(i+300));
```
agorina: This is a complete hack, but we use negative values in the test data and without this it was failing tests. If this proves to be a permanent solution I will change the test data.

aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -13,7 +13,7 @@ sqlbuilder:
               sum(CASE WHEN performance_level = 3 THEN 1 ELSE 0 END) AS level3,
               sum(CASE WHEN performance_level = 4 THEN 1 ELSE 0 END) AS level4,
               fe.asmt_id,
-              fe.asmt_grade_id,
+              a.grade_id as asmt_grade_id,
```
agorina: Based on @wtritch finding, this change along with the corresponding change to the asmt sort key seems to remove the "full index scan" warning.

I have done a very quick check so I am not confident that it is fully solved, but this should not hurt anything.

PR #768, opened 2018-01-19
Remove CROSS JOINs on reporting dimensions in favor of functionally e…

wtritch:
…quivalent LEFT JOINs

Here's the most recent ~ms queries are after switching from a `CROSS JOIN` to a `LEFT JOIN...ON true=true` on the dimension.

image

### DIFF
--- a/aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
+++ b/aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
@@ -297,53 +297,47 @@ sqlbuilder:
               # the 'magic' number -1 is needed here to support JOIN in the final overarching query defined in the application.sql.yml
               clauses:
                 join: >-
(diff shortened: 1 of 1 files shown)


esotrope:
what is a cross join?

wtritch:
@esotrope It's like a LEFT RIGHT JOIN.  It picks all results from the right for each result from the left.
`(1, 2) CROSS JOIN (3, 4) -> 1, 3; 1, 4; 2, 3; 2, 4;`
It's how we were back-filling zeros across all Dimension values even if there were no results.

wtritch:
I'm going to merge this so that I can test if the speed updates are consistent and real across randomized queries.  I will continue to resolve comments as they come in.

jtreuting:
We will want @agorina to review to make sure we aren't missing something of course.

PR #769, opened 2018-01-19
Feature/aggregate report summative enable flag

esotrope:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateServiceCache.java
```diff
@@ -5,6 +5,7 @@
  */
 public final class AggregateServiceCache {
 
+    public static final String AssessmentTypes = "assessmentTypes";
```
esotrope: added assessment type cache because it felt weird for the auth service to get the whole cached report options object when it only wants to see the assessment types.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/UserAssessmentTypeService.java
```diff
@@ -0,0 +1,22 @@
...
+import org.opentestsystem.rdw.olap.model.CodedEntity;
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+
+import javax.validation.constraints.NotNull;
+import java.util.Set;
+
+/**
+ * Service responsible for managing assessment types
+ */
+public interface UserAssessmentTypeService {
```
esotrope: split `AssessmentTypeService` into `AssessmentTypeService` and `UserAssessmentTypeService` for user-specific assessment types. This matches the rest of the service namings

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/AuthorizationService.java
```diff
@@ -22,13 +19,4 @@
...
     void checkAuthorization(@NotNull User user, @NotNull ReportQuery query);
 
-    /**
-     * Filters the provided assessment types given the user's aggregate report permissions
-     *
-     * @param user            the {@link User}
-     * @param assessmentTypes all assessment types of the system
-     * @return assessment types that the user is entitled to create reports for
-     */
-    Set<CodedEntity> getAuthorizedAssessmentTypes(@NotNull User user, @NotNull Collection<CodedEntity> assessmentTypes);
```
esotrope: removed this awkward method and pushes security into the assessment type services

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserAssessmentTypeService.java
```diff
@@ -0,0 +1,47 @@
...
+    DefaultUserAssessmentTypeService(
+            final AssessmentTypeService service,
+            final AggregateReportSettings settings) {
+        this.service = service;
+        this.settings = settings;
+    }
+
+    @Override
+    @Cacheable(AggregateServiceCache.AssessmentTypes)
+    public Set<CodedEntity> getAssessmentTypes(final User user) {
```
wtritch: Hm... I'm debating recommending against caching on a per-user basis...

aggregate-service/src/main/resources/application.yml
```diff
@@ -47,6 +47,7 @@ app:
       refresh-cron: "0 0 1 * * *"
 
   aggregate-reports:
+    summative-enabled: true
```
wtritch: Should this be "false" by default?

esotrope: meant to point this out - i was thinking at a system level it makes more sense to be enabled by default right? and then California should set it false no?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserAssessmentTypeService.java
esotrope: Thanks - this is copy paste error. *removing

aggregate-service/src/main/resources/application.yml
esotrope: @jtreuting a side effect of having
```
summative-enabled: false
```
is that statewide users will not be able to select any assessment type on the report page and will not be able to run reports. 
Is this what we want? should the system throw some warning if these are both false?

esotrope:
Merging this for now. I will bring up the "summative-enabled" flag nuances in a parking lot I think

PR #770, opened 2018-01-19
Removed logic to backfill missing data in aggregate report.

agorina:


aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
```diff
@@ -43,6 +44,7 @@
 
 @Import({JdbcCustomAggregateReportRowRepository.class})
 @Sql(scripts = {"classpath:OlapEntitiesSetup.sql"})
+@Ignore
```
agorina: Most of the test fail, so it was easier to do this for now.

aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -116,75 +116,108 @@ sqlbuilder:
...
+                     fe.section504,
+                     b.code
 
             MigrantStatus:
               clauses:
                 select: >-
-                     fe.migrant_status AS dimension_id
+                     'MigrantStatus' AS dimension,
+                     fe.migrant_status AS dimension_id,
+                     b.code AS dimension_code
```
wtritch: Would a CASE on the two boolean values make more sense than a join on a "boolean" table?

agorina: Checking...

agorina: Will check later, merging so that we could run another round of tests today.

PR #773, opened 2018-01-21
Feature/comparative subgroups

esotrope:
Makes dimensions API provided
Adds form section for comparative subgroups (dimensions)
"Componentizes" the form summary view

image

image


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/CachingReportOptionService.java
```diff
@@ -39,8 +48,14 @@
     @Override
     @Cacheable(AggregateServiceCache.ReportOptions)
     public ReportOptions getReportOptions() {
+
+        final ImmutableSet<DimensionType> dimensionTypes = Arrays.stream(DimensionType.values())
+                .filter(type -> type != DimensionType.Overall)
+                .collect(toImmutableSet());
```
esotrope: Questions i had when doing this:
Should the enum store whether the dimension is selectable on a form or not? (Overall i presume is not)
Should they also then store their display order?
Should this even be backend driven?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
```diff
@@ -5,6 +5,14 @@ import { Option as SbToggleOption } from "../shared/sb-toggle.component";
...
 import { TranslateService } from "@ngx-translate/core";
 import { SchoolYearPipe } from "../shared/format/school-year.pipe";
+import { ordering } from "@kourge/ordering";
+import { ranking } from "@kourge/ordering/comparator";
+
+const localeCompareText = (a, b) => a.text.localeCompare(b.text);
+
+const dimensionTypeComparator = ordering(ranking(
+  ['Gender', 'Ethnicity', 'LEP', 'MigrantStatus', 'Section504', 'IEP', 'EconomicDisadvantage', 'StudentEnrolledGrade']
+)).on((option: SbCheckboxGroupOption) => option.value).compare;
```
esotrope: sort order is stored client side which kind of defeats the purpose of having the backend provide the dimensions hence my previous qs

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -0,0 +1,40 @@
+import { Component, Input } from "@angular/core";
+
+const valueToOptionText = attribute => value => attribute.options.find(option => option.value === value).text
```
esotrope: A little bit of hackery to resolve what should be shown in the summary table. This assumes that all option objects have a `value` and `text` field.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -0,0 +1,40 @@
...
+@Component({
+  selector: 'aggregate-report-summary',
+  templateUrl: './aggregate-report-summary.component.html',
+})
+export class AggregateReportSummary {
+
+  @Input()
+  columns: Column[];
+
+  summarizeSelection(attribute: Attribute): boolean {
```
esotrope: ts docs to come. this method basically decides whether it should display "All" or list all the selected values.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/CachingReportOptionService.java
```diff
@@ -39,8 +48,14 @@
     @Override
     @Cacheable(AggregateServiceCache.ReportOptions)
     public ReportOptions getReportOptions() {
+
+        final ImmutableSet<DimensionType> dimensionTypes = Arrays.stream(DimensionType.values())
```
malaffoon: I think this should use Enum operators, something like `DimensionType.complementOf(DimensionType.of(DimensionType.Overall))`

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/CachingReportOptionService.java
malaffoon: Feels like it shouldn't be in the enum itself, it is a UI concern so business logic should probably be in the front-end?

wtritch:
I know it's consistent UI, but I would discourage having an "All" option on Dimensions, I would like to think that most queries would be run with "None" unless the user is doing some pointed investigation.

esotrope:
@wtritch that's the goal. I'm hoping to incorporate a flag in our sb-checkbox-group component to allow for a "none" option and then use it for dimensions and school grade. This will necessitate form validation though as school grade is required i think.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.mapper.ts
esotrope: I think i'll go for alpha sort and if they see it and want a specific order we can specify a "natural order" field in the enum.

malaffoon: Careful on the alpha sort ... is it based on the "raw" name of the dimension or the translated name? Do we translate these? 

And, yeah, i know the UI isn't going to be translated on our watch but we should do the right thing.

esotrope: Good point. Alpha sorting the translated text is nice but will likely result in inconsistent ordering across languages. That makes me want to always rely on some natural order or backend-defined order.

esotrope:
Going to merge this for now and add the agreed upon sorting in a followup PR

PR #774, opened 2018-01-22
Changes to test aggregate report performance with dimension specific fact tables.

agorina:


aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -116,12 +113,17 @@ sqlbuilder:
...
+                  fact_student_exam_new fe
+
             Gender:
               clauses:
                 select: >-
                      'Gender' AS dimension,
                       s.gender_id  AS dimension_id,
                       gn.code AS dimension_code
+                from: >-
+                  fact_student_exam_gender fe
```
agorina: I created VIEWs for these tables in the Schema project and will load real tables in AWS Dev Redshift.

PR #776, opened 2018-01-23
adds none state to checkbox group

esotrope:
Styles probably need some work - but at least functionally now we can default to no selection on assessment grade and comparative subgroups (dimensions)

image

image

Note: we will need to add form validation if we go this direction because i think we require at least one assessment grade be selected right?


esotrope:
Some style changes to make checkbox groups without the all-option look better

image

image

PR #777, opened 2018-01-23
DWR-1230 Misc changes based on the load test

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -72,6 +72,7 @@ public JdbcCustomAggregateReportRowRepository(final NamedParameterJdbcTemplate j
         this.jdbcTemplate = jdbcTemplate;
         this.customAggregateQueryProvider = customAggregateQueryProvider;
 
+        //TODO: should this be moved into methods below? and the repository calls are cached?
```
agorina: I will remove this comment, but wanted to start a discussion.

Since migrate could change the data in these tables, if we keep it as is, it would required a service restart to notice the change.

I know that we tend to keep Service calls 'cacheable', this would required a repository calls to be cached. 

Any objections?

aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -20,6 +20,9 @@ sqlbuilder:
               a.label as asmt_label,
               subject.code as subject_code
 
+          from: >-
+              fact_student_exam fe
+
```
agorina: The idea of multiple fact table did not work out.

aggregate-service/src/test/resources/OlapEntitiesSetup.sql
```diff
@@ -53,7 +53,7 @@ INSERT INTO student_ethnicity(student_id, ethnicity_id) values
   (-9,  -29);
 
 -- ------------------------------------------ Exams ---------------------------------------------------------------------------------------------
-INSERT INTO  fact_student_exam (id, school_year, asmt_id, asmt_grade_id, completeness_id,
+INSERT INTO  fact_student_exam (id, school_year, asmt_id, completeness_id,
```
agorina: There is a corresponding schema change.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
malaffoon: I'm mixed on this: i don't like the idea that you'd ever have to restart the service due to data changing. But this is, fundamentally, configuration data so it doesn't bother me as much. 

That said, i think you should remove the TODO and create a JIRA (with a nice clear description) so we can prioritize it.

agorina: Makes sense for this, but I have a follow up question in regards to this subject and the code to backfill the report data: "should the backfilling be done in the repo or service?"

Why the repo:
- we already have miles of tests to verify it. Otherwise all the tests will need to be changed, and considering the timeline I think it is a reason to keep in mind;
- each report is a collection of SQLs, it would be easier to backfill per each SQL, not after they are all executed and it would not required an additional looping through the data (if we do it the service layer).

The data used for backfilling should be cached (to avoid calls to Redshift). And that leads to the same question I have above: how do people feel about caching the repo calls?

Just to clarify : I am open for any alternatives, just want to resolve this before I start coding to avoid refactoring later.

PR #778, opened 2018-01-23
adds info text for aggregate report form. adds checkbox group styles …

esotrope:
…intended for other package

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -591,11 +591,11 @@
           },
           "completeness": {
             "label": "Completeness",
-            "info": "Completeness"
+            "info": "Indicates whether the test was completed or was partially complete."
           },
           "interim-administration-condition": {
             "label": "Manner of Administration",
-            "info": "Manner of Administration",
+            "info": "For interim assessments, a Standardized administration follows the same protocols and security procedures as a summative test administration. For a Non Standardized administration, the teacher may use the assessment during instruction and interact with students as they complete the assessment.",
             "qualifier": "Interim"
           },
           "school-year": {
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -331,6 +331,36 @@ body .select-menu-group {
   padding: 0.5em 0.5em 0.5em 0;
 }
 
+/**
```
esotrope: these styles were supposed to be wrapped in the checkbox-none-state PR but i committed them to the wrong branch so i'm including them here

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -608,36 +608,36 @@
...
-            "info": "Limited English Proficiency"
+            "info": "Students who use a language other than English as their primary language and who have been assessed as not possessing the English language skills necessary to succeed in the regular instructional program."
           },
           "economic-disadvantage": {
             "label": "Economic Disadvantage",
-            "info": "Economic Disadvantage"
+            "info": "Indicates that a student qualifies for free or reduced lunch or meets other state criteria for economic disadvantage."
           },
           "achievement-level-display-type": {
             "label": "Achievement Level",
```
esotrope: @jtreuting The following form items do not have popover info text yet.
I'm thinking they don't all need it but let me know and I'll write something.

```
Assessment Grades
Assessment Type
Academic Years
Subjects
Achievement Level (separate or grouped)
Show Value as (percent or number)
Comparative Subgroups
```

jtreuting: Don't write anything yet.  Once we have a stable UI on this page we will ask SB to provide info.  Though some of these might have context info from other screens.  If there is and they make sense we should reuse them for consistency.

esotrope: Alright - will do

PR #780, opened 2018-01-23
Feature/dwr 1246 - generate preview data for customize aggregate reports

dsnyder-fairway:
## ICA with selected subgroups
screen shot 2018-01-23 at 1 41 51 pm
screen shot 2018-01-23 at 1 41 44 pm

## IABs
screen shot 2018-01-23 at 1 42 14 pm


webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports-preview-table.component.ts
```diff
@@ -0,0 +1,234 @@
+import { Component, Input, OnInit, ViewChild } from "@angular/core";
```
dsnyder-fairway: heavily borrowed from `aggregate-reports-table.component.ts`. However, removed some logic which wasn't applicable on static table

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports-preview-table.component.html
```diff
@@ -0,0 +1,139 @@
+<div class="disabled-table">
```
dsnyder-fairway: heavily borrowed from `aggregate-reports-table.component.html`. However, removed some logic which wasn't applicable on static table

webapp/src/main/webapp/src/app/aggregate-report/results/mock-aggregate-reports-preview.service.ts
```diff
@@ -0,0 +1,122 @@
+import { Injectable } from "@angular/core";
```
dsnyder-fairway: heavily borrowed from `mock-aggregate-reports.service.ts`. However, removed some logic which wasn't applicable on static table

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -48,7 +49,8 @@ <h2 class="h3 gray-darkest">
                                  [(ngModel)]="settings.subjects"
                                  analyticsEvent="FilterClick"
                                  analyticsCategory="AggregateQueryBuilder"
-                                 [allOptionAnalyticsProperties]="{label: 'Subject: All'}"></sb-checkbox-group>
+                                 [allOptionAnalyticsProperties]="{label: 'Subject: All'}"
+                                 (changed)="generateReport()"></sb-checkbox-group>
```
esotrope: i think instead of hooking into the change events we can simply give the table component a getter reference.
e.g.
```html
<preview-table [columns]="previewTableColumns">
```
```typescript
get previewTableColumns() {
  return // compute the columns or whatever
}
```

webapp/src/main/webapp/src/app/aggregate-report/model/aggregate-report-item.model.ts
```diff
@@ -7,16 +7,16 @@ export class AggregateReportItem {
...
-  studentsTested: number;
-  performanceLevelCounts: number[] = [];
-  performanceLevelPercents: number[] = [];
-  groupedPerformanceLevelCounts: number[] = [];
-  groupedPerformanceLevelPercents: number[] = [];
+  avgScaleScore: string | number;
+  avgStdErr: string | number;
+  studentsTested: any;
+  performanceLevelCounts: (number | string)[] = [];
+  performanceLevelPercents: (number | string)[] = [];
```
esotrope: not sure but should this be `number[] | string` ? I assume it's for putting in 'N/A' kind of text right?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports-preview-table.component.ts
```diff
@@ -0,0 +1,234 @@
...
+ *   tree columns - these data identity columns are used to organize data into a
+ *                  hierarchical view based upon the order of the columns
+ *   data columns - these columns display averages/counts/performance levels for the
+ *                  identified subject group
+ */
+@Component({
+  selector: 'aggregate-reports-preview-table',
+  templateUrl: './aggregate-reports-preview-table.component.html',
+})
+export class AggregateReportsPreviewTableComponent implements OnInit {
```
esotrope: Assuming there is common code here - is there a way to use extension or otherwise reuse any common code. I want to avoid having to propagate changes to two places when something changes.

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -19,10 +20,13 @@ const CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR = {
   selector: 'sb-toggle',
   template: `
     <div class="btn-group btn-group-sm toggle-group" data-toggle="buttons">
-      <label *ngFor="let option of options" class="btn btn-primary" [ngClass]="{active: option.value === value, disabled: disabled}">
+      <label *ngFor="let option of options" class="btn btn-primary"
+             [ngClass]="{active: option.value === value, disabled: disabled}">
         <input type="radio" [id]="name" [name]="name" [disabled]="disabled" [value]="option.value" [(ngModel)]="value"
+               (ngModelChange)="changed.emit(value)"
```
esotrope: I would hope this isn't necessary since we are implementing `ControlValueAccessor` but maybe it is - let me know

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports-preview-table.component.ts
dsnyder-fairway: There is. I was going down that route but decided on separating since : 

- some of the logic wasn't applicable on an unsortable table. Not a big deal
- we _may_ need to change the `AggregateReportsTableComponent` when we get rid of the `MockAggregateReportsService`

I can consolidate these though

webapp/src/main/webapp/src/app/aggregate-report/model/aggregate-report-item.model.ts
dsnyder-fairway: we are putting `---` in there. Currently we have 3 of the columns for IABs and 4 for summative/ICA. Is there an advantage to tossing a couple redundant `---` and handling these differently?

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
dsnyder-fairway: ah, yeah, let me look into that

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports-preview-table.component.ts
esotrope: K maybe it's a good candidate for having a common ancestor then.
I did like the idea of just saying `[preview]="true"` and it magically working but yeah that will complicate the component

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
dsnyder-fairway: I think my last commit leverages this

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
dsnyder-fairway: hmmm, this should be for rows rather than columns (with the exception of IAB vs Summative/ICA displaying scores). I'll look into hooking something like this up

esotrope: ah yeah that's what i meant - thanks ```[data]="dataGetter"```

dsnyder-fairway: working this now. It should be in the next pull request.

PR #784, opened 2018-01-24
Feature/web writing trait scores

jtreuting:
**Aggregate Writing Trait Score view**
image

I moved the item table down under the view dropdown which is different from the mockup because then you can still do the expanding of the Item to see the responses of students, etc.  which seemed relevant.

**When non-IAB**
image

**TODO**
- Implement CSV export


webapp/src/main/webapp/src/app/assessments/assessment-provider.interface.ts
```diff
@@ -10,7 +10,7 @@ import { ExportRequest } from "./model/export-request.model";
 export interface AssessmentProvider{
   getAvailableAssessments(): Observable<Assessment[]>;
   getExams(assessmentId: number): Observable<Exam[]>;
-  getAssessmentItems(assessmentId: number, multipleChoiceMultipleSelectItems?: boolean): Observable<AssessmentItem[]>;
+  getAssessmentItems(assessmentId: number, itemTypes?: string[]): Observable<AssessmentItem[]>;
```
jtreuting: Changed to pass in an array of item types `['MS','MC']` or `['WER']` instead of a flag since now we have a need for only getting the writing trait score item as well as the distractor analysis items.

webapp/src/main/webapp/src/app/assessments/model/writing-trait-aggregate.model.ts
```diff
@@ -0,0 +1,20 @@
...
+export class WritingTraitAggregate {
+  trait: WritingTrait;
+  average: number;
+  numbers: number[] = [];
+  percents: number[] = [];
+
+  constructor(trait: WritingTrait) {
+    this.trait = trait;
+
+    for (let i = 0; i <= trait.maxPoints; i++) {
```
jtreuting: not sure i love this, but zero fill the arrays that represent the point breakdowns for this trait based on the max points allowed for it (4 for Evidence and Org, 2 for Conventions and 6 for Total)

webapp/src/main/webapp/src/app/assessments/model/writing-trait-score-summary.model.ts
```diff
@@ -0,0 +1,17 @@
...
+import { WritingTrait } from "./writing-trait.model";
+
+/**
+ * This class represents a writing trait score summary report (e.g. all of the rows of aggregate writing traits)
+ */
+export class WritingTraitScoreSummary {
+  evidence: WritingTraitAggregate = new WritingTraitAggregate(new WritingTrait(WritingTraitType.Evidence, 4));
+  organization: WritingTraitAggregate = new WritingTraitAggregate(new WritingTrait(WritingTraitType.Organization, 4));
+  conventions: WritingTraitAggregate = new WritingTraitAggregate(new WritingTrait(WritingTraitType.Conventions, 2));
+  total: WritingTraitAggregate = new WritingTraitAggregate(new WritingTrait(WritingTraitType.Total, 6));
```
jtreuting: again something i don't really like, doesn't feel right.  my idea was that I want the maximum points for each trait to be defined in one place.  so this sets up the summary object and then the totals for each point breakdown get added later.

webapp/src/main/webapp/src/app/assessments/model/writing-trait-type.model.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * This various writing trait types
+ */
+export enum WritingTraitType {
```
jtreuting: used for determining the labels piped thru translate

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -75,61 +75,79 @@
 
     <!-- Student Results -->
     <div *ngIf="!collapsed" @fadeAnimation>
-      <div class="col-md-4 z-index-over-data-table pl-0 mb-sm">
-        <div class="select-menu-group">
-          <label class="sr-only">{{ 'labels.groups.results.assessment.select-view' | translate }}</label>
-          <div class="btn-group" dropdown>
-            <button dropdownToggle type="button" class="btn btn-default btn-borderless dropdown-toggle">
-              {{ currentResultsView.label | translate }} <i class="fa fa-angle-down"></i>
+      <div class="row">
```
jtreuting: this is formatting of an indent really.  I added the bootstrap row to this so that the results table that is smaller is always displayed below it.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -75,61 +75,79 @@
...
           </div>
-          <button *ngIf="currentResultsView.canExport"
-                  class="btn btn-sm btn-default"
-                  [disabled]="!currentExportResults || !currentExportResults.hasDataToExport()"
-                  (click)="currentExportResults.exportToCsv()">
-            <i class="fa fa-cloud-download"></i> {{ 'labels.export.title' | translate }}
-          </button>
+        </div>
+        <div class="col-md-8 pt-md">
+          <span class="results-view-intro">{{ getCurrentViewIntroductionLabel() | translate }}</span>
```
jtreuting: now that it is a row, added the other column to fill it in.  I am thinking this will be used to add descriptive text for what the view is.  There is a request to add this info for the Distractor Analysis and the Writing Trait Scores, so put some placeholders in place that will need styling later when the text is decided on.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -129,7 +131,15 @@ export class AssessmentResultsComponent implements OnInit {
   }
 
   get displayItemLevelData(): boolean {
-    return this._assessmentExam.exams.some(x => x.schoolYear > this.minimumItemDataYear) && !this._assessmentExam.assessment.isSummative;
+    return !this._assessmentExam.assessment.isSummative && this._assessmentExam.exams.some(x => x.schoolYear > this.minimumItemDataYear);
+  }
+
+  get displayWritingTraitScores(): boolean {
+    return this._assessmentExam.assessment.subject == 'ELA';
```
jtreuting: feel wrong, but i am only seeing an enum for subject IDs and the assessment that we have here has the code for the subject.  the logic here is saying that writing trait scores are only relevant for ELA so don't show the option if it's not ELA

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -188,16 +212,18 @@ export class AssessmentResultsComponent implements OnInit {
...
-    this.resultsByItemView = this.getResultViewState(ResultsViewState.ByItem, this.displayItemLevelData, true)
-    this.distractorAnalysisView = this.getResultViewState(ResultsViewState.DistractorAnalysis, this.displayItemLevelData, true);
+    this.resultsByStudentView = this.getResultViewState(ResultsViewState.ByStudent, true, false, true);
+    this.resultsByItemView = this.getResultViewState(ResultsViewState.ByItem, this.displayItemLevelData, true, true)
+    this.distractorAnalysisView = this.getResultViewState(ResultsViewState.DistractorAnalysis, this.displayItemLevelData, true, true);
+    this.writingTraitScoresView = this.getResultViewState(ResultsViewState.WritingTraitScores, this.enableWritingTraitScores, true, this.displayWritingTraitScores);
   }
 
-  getResultViewState(viewState: ResultsViewState, enabled: boolean, canExport: boolean): ResultsView {
+  getResultViewState(viewState: ResultsViewState, enabled: boolean, canExport: boolean, display: boolean): ResultsView {
```
jtreuting: added a new option that controls whether to include the view in the dropdown at all.  I thought it made more sense to not show the Writing Trait Score option for Math tests, and show a disabled popover when it is ICA or Summative ELA since that won't make sense intuitively to the user like it not being there for Math which should be expected.

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -93,6 +94,41 @@ export class ExamStatisticsCalculator {
...
+    let summary = new WritingTraitScoreSummary();
+
+    if (assessmentItems.length == 0) {
+      return summary;
+    }
+
+    let itemsWithTraitScores = assessmentItems[0].scores.filter(x => x.writingTraitScores != null);
+    let totalAnswers = itemsWithTraitScores.length;
+
+    itemsWithTraitScores.forEach((score, index) => {
```
jtreuting: total how many results have each sub-score, then use these totals to calculate the equivalent percent and the average for each trait

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      this.loading = false
+    });
+  }
+
+  hasDataToExport(): boolean {
+    return this.filteredItems && this.filteredItems.length > 0;
+  }
+
+  // TODO: need to get this working still
```
jtreuting: plan to submit a PR for this after this one

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1542,11 +1560,13 @@
     "coming-soon": "{{featureName}} is coming soon.",
     "loading": "Loading",
     "logout": "Logout",
-    "no-distractor-analysis": "Distractor analysis is not available for tests administered prior to 2017-18 school year.",
-    "no-distractor-analysis-for-summative-exams": "Distractor analysis is not available for summative tests.",
-    "no-results-by-item": "Results by item is not available for tests administered prior to 2017-18 school year.",
-    "no-results-by-item-for-summative-exams": "Results by item is not available for summative tests.",
+    "no-distractor-analysis": "Distractor analysis is not available for assessments administered prior to 2017-18 school year.",
```
jtreuting: they prefer the term assessment to test so changed these when i noticed them

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
malaffoon: What if the ELA doesn't have any WER items? Most ELA IABs do *not* have WER.

malaffoon: Also, wouldn't it be more consistent to add isEla to the assessment class?

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -93,6 +94,41 @@ export class ExamStatisticsCalculator {
...
   }
 
+  aggregateWritingTraitScores(assessmentItems: AssessmentItem[]) : WritingTraitScoreSummary {
+    let summary = new WritingTraitScoreSummary();
+
+    if (assessmentItems.length == 0) {
+      return summary;
+    }
+
+    let itemsWithTraitScores = assessmentItems[0].scores.filter(x => x.writingTraitScores != null);
```
malaffoon: i'm confused about the data model (or maybe my js knowledge); i guess i should just wander off and let Landon/Devin review but ... why are you getting only the first assessment item's scores?

esotrope: Yeah looks suspect 🕵️

webapp/src/main/webapp/src/app/assessments/model/writing-trait-aggregate.model.ts
esotrope: Is this necessary for some summation algo to work later? by default array access of an undefined index is undefined btw if that can be leveraged.
```js
a = [];
a[1]; // undefined
```

webapp/src/main/webapp/src/app/assessments/model/writing-trait-score-summary.model.ts
esotrope: so normally that info would live in an enum but since `ts` enums don't support that you could have factory methods maybe?
```ts
WritingTraits.evidence()

static evidence(): WritingTrait {
  return new WritingTrait(WritingTraitType.Evidence, 4);
}
...
```

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.spec.ts
```diff
@@ -169,6 +169,86 @@ describe('Exam Calculator', () => {
     expect(actual[ 2 ].value).toBe(0);
   });
 
+  it('should aggregate writing trait scores', () =>{
```
esotrope: yay tests 👍

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.spec.ts
```diff
@@ -169,6 +169,86 @@ describe('Exam Calculator', () => {
...
+    expect(summary.total.numbers[2]).toBe(1);
+    expect(summary.total.percents[2]).toBe(20.0);
+    expect(summary.total.numbers[3]).toBe(0);
+    expect(summary.total.percents[3]).toBe(0);
+    expect(summary.total.numbers[4]).toBe(3);
+    expect(summary.total.percents[4]).toBe(60.0);
+    expect(summary.total.numbers[5]).toBe(0);
+    expect(summary.total.percents[5]).toBe(0);
+    expect(summary.total.numbers[6]).toBe(1);
+    expect(summary.total.percents[6]).toBe(20.0);
```
esotrope: i think `toEqual` or a similar method will do equality comparison instead of `===` so these expect statements can be shortened to:
```ts
expect(summary.total.numbers).toEqual([1,2,3,4,5]);
```

webapp/src/main/webapp/src/app/assessments/model/writing-trait-aggregate.model.ts
jtreuting: I'll play with that

webapp/src/main/webapp/src/app/assessments/model/writing-trait-score-summary.model.ts
jtreuting: yeah i like this idea better, seems like a good approach.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
jtreuting: If an ELA assessment doesn't have a WER then they will see a screen telling them that after they select it.  Right now the details for an item don't come back initially and we don't retrieve them until we select Results by Item, Distractor Analysis or Writing Trait Scores.  So I figured the best I could do was make it obvious for Math.  Maybe we should try to do more?

Also, I agree with the isEla approach, will make that change.

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
jtreuting: yeah i went back and forth on this.  right now there is only a single WER item per IAB that will come back and the UI is really only setup to show one aggregate table so if there happen to be 2 items they will get combined together in that table even though the top "mini-table" that has the items shown would show 2 items.  So I thought that we should limit to the first WER (which in reality will be the only one) so that we don't get into a situation where the UI is confusing.

But maybe we don't worry about it and take all the scores across all items into that aggregate display?  Easy enough change.  

Thoughts?

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.spec.ts
jtreuting: i appreciate the plural, but more accurate, "yay test"

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.spec.ts
jtreuting: nice, good idea

esotrope: x D

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
esotrope: Yeah i think i need some pics to follow the nuance here. Let's try to do a screenshare offline at some point

jtreuting: Talked it through and implemented it so that it can handle if there are 2 WER items in a single assessment.  It will produce multiple Point Breakdown summary tables that represent each item.  This is not a use-case that we will hit with SB content but it will handle it gracefully now.

PR #787, opened 2018-01-25
DWR-1333 Changes to olap OrganizationRepository and  more

agorina:
Just to explain why I am doing this:
To backfill the aggregate report data I need Organizations to be cached and accessible in a certain way. While doing the change I have noticed a few things that I am trying to resolve:

1. Aggregate service was using two different Organization models: one from the Common and one from the service. The repository was returning the one from the Common and it is used to populate the organization list in the UI. ReportRow was using another one. It just did not seem right. 
2. In the aggregate service, Organization ids where `int` while in other places `long`. While `int` seems to match the data storage, I went back and forth and ended up with `long`. This is because the RDW_Common has a common permission logic that uses `long` ids and I want not sure why and what else will need to be change.


admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
```diff
@@ -1,10 +1,8 @@
 package org.opentestsystem.rdw.admin.repository;
 
 import org.opentestsystem.rdw.admin.model.Embargo;
-import org.opentestsystem.rdw.common.model.Subject;
```
agorina: No real change, just ran formatting.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
```diff
@@ -1,29 +1,94 @@
...
+            AssessmentTypes,
+            Booleans,
+            StrictBooleans,
+            Organizations,
+            CompletenessCodes,
+            Ethnicities,
+            Genders,
+            SchoolYears,
+            Subjects}, beforeInvocation = true, allEntries = true)
+    public void refreshRepositoryCache() {
```
agorina: I added all the cached repos into here. 

Any objections?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/Organization.java
```diff
@@ -1,88 +0,0 @@
-package org.opentestsystem.rdw.olap.model;
```
agorina: I am using the version from Common.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/OrganizationRepository.java
```diff
@@ -1,19 +1,34 @@
...
+     * @return all {@link District}s; cannot be null
+     */
+    Map<Long, District> findAllDistricts();
+
+    /**
+     * Gets all {@link School}s mapped groupped by their {@link District#getId()}
+     *
+     * @return all {@link School}s; cannot be null
+     */
+    Map<Long, List<School>> findAllSchoolsByDistrictIds();
```
agorina: This is the main change that I need.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -72,14 +73,14 @@ public JdbcCustomAggregateReportRowRepository(final NamedParameterJdbcTemplate j
...
-        booleanMaxCount = max(3, booleanRepository.findAll().size());
-        strictBooleanMaxCount = max(2, strictBooleanRepository.findAll().size());
+        this.genderRepository = genderRepository;
+        this.ethnicityRepository = ethnicityRepository;
+        this.assessmentGradeRepository = assessmentGradeRepository;
+        this.subjectRepository = subjectRepository;
+        this.schoolYearRepository = schoolYearRepository;
+        this.administrationConditionRepository = administrationConditionRepository;
+        this.booleanRepository = booleanRepository;
+        this.strictBooleanRepository = strictBooleanRepository;
```
agorina: I change this because it requires Redshift connection if you need to run reporting even if you are not working on Redshift. Did not seem right. 

This whole area is going to be changed soon.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
```diff
@@ -73,22 +73,26 @@ public Object resolveArgument(final MethodParameter parameter,
...
     }
 
-    private Set<Integer> toSet(final String value) {
+    private Set<Integer> toIntSet(final String value) {
         return value == null ? newHashSet() : Stream.of(value.split(",")).map(Integer::valueOf).filter(Objects::nonNull).collect(Collectors.toSet());
     }
 
+    private Set<Long> toLongSet(final String value) {
+        return value == null ? newHashSet() : Stream.of(value.split(",")).map(Long::valueOf).filter(Objects::nonNull).collect(Collectors.toSet());
+    }
```
agorina: I find it ugly to have two identical methods with different return data type; could not find a better way. Open for suggestions.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
esotrope: Nope - actually we should have a `@PostConstruct` run the same stuff too so the user never gets slammed with loading this stuff

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
```diff
@@ -106,4 +107,21 @@ private QueryUtils() {
         }
         return padded;
     }
+
+    public static Collection<Long> padLongToSize(final Collection<Long> collection, final int size) {
```
agorina: This is another ugly thing, similar to the comment above.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Organization.java
```diff
@@ -3,17 +3,17 @@
 /**
  * Represents a district or school responsible for the student who took an exam
  */
-public abstract class Organization {
+public class Organization {
 
-    private long id;
+    private Long id;
```
agorina: State does not have an id, what is why I change the type.

esotrope: I'm wondering if there shouldn't be two base models.
Organization (no id)
IdentifiableOrganization / InStateOrganization (has id)

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/OrganizationRepository.java
```diff
@@ -1,19 +1,34 @@
...
      */
-    List<Organization> findAllSchoolsAndDistricts();
+    Map<Long, School> findAllSchools();
 
+    /**
+     * Gets all {@link District}s mapped by their ids
+     *
+     * @return all {@link District}s; cannot be null
+     */
+    Map<Long, District> findAllDistricts();
```
malaffoon: It seems very odd that a simple `findAll` would return a map. Seems like the repository should just return a collection (Set or List). Any calling method that needs a map for some reason could generate the map.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
```diff
@@ -90,13 +91,13 @@ private QueryUtils() {
      * @param size       final size of the collection to return
      * @return the given collection padded to the given size
      */
-    public static Collection<Integer> padToSize(final Collection<Integer> collection, final int size) {
+    public static Collection<Integer> padIntToSize(final Collection<Integer> collection, final int size) {
```
malaffoon: This couldn't be parameterized to avoid duplication?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Organization.java
malaffoon: The eternal struggle of (avoiding) parallel data models. If i was motivated, i'd poll people and dig through my past efforts to see which approach is more successful but, until then ... i'm fine with a common class with a nullable (and directly settable) id.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/ReportQueryArgumentResolver.java
malaffoon: Although you could parameterize it, that wouldn't be any nicer looking and you'd still have the issue of coercing the valueOf call somehow, so you'd end up with duplication.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
malaffoon: Unlike the previous instance, i think it is worth merging these two by parameterization, because there is a lot of common type-independent logic. I haven't worked through it though, so i'm not sure how ugly it will get.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/OrganizationRepository.java
agorina: Discussed offline - will rename the method and add comments explaining why I need it.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/QueryUtils.java
agorina: Ah, thank you! Changing.

agorina:
Closing due to merge conflict resolution nightmare. I am going to create another PR and merge it.

PR #788, opened 2018-01-25
writing trait score CSV export implementation

jtreuting:
- Implements writing trait score aggregate view CSV export from client side
- Refactored the existing export into a dedicated interface and implementations separate from assessment provider

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessment-provider.interface.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessment-provider.interface.ts
@@ -2,18 +2,13 @@ import { Observable } from "rxjs/Observable";
 import { Assessment } from "./model/assessment.model";
 import { Exam } from "./model/exam.model";
 import { AssessmentItem } from "./model/assessment-item.model";
-import { ExportRequest } from "./model/export-request.model";
 
 /**
  * Implementations of this interface are responsible for providing context-based assessment and exam data.
  */
-export interface AssessmentProvider{
+export interface AssessmentProvider {
   getAvailableAssessments(): Observable<Assessment[]>;
   getExams(assessmentId: number): Observable<Exam[]>;
   getAssessmentItems(assessmentId: number, itemTypes?: string[]): Observable<AssessmentItem[]>;
   getSchoolId(): number;
-
-  // TODO: Technically not a provider method, but if we add one more export, let's break
-  // TODO: this out to an AssessmentExporter interface.
-  exportItemsToCsv(exportRequest: ExportRequest)
 }
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -44,7 +44,7 @@ export class AssessmentsComponent implements OnInit {
 
   /**
    * The provider which implements the AssessmentExporter interface in order
-   * to export CSV data.
+   * to export data.
    */
   @Input()
   assessmentExporter: AssessmentExporter;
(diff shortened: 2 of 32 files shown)


webapp/src/main/webapp/src/app/assessments/assessment-provider.interface.ts
```diff
@@ -2,18 +2,13 @@ import { Observable } from "rxjs/Observable";
...
  */
-export interface AssessmentProvider{
+export interface AssessmentProvider {
   getAvailableAssessments(): Observable<Assessment[]>;
   getExams(assessmentId: number): Observable<Exam[]>;
   getAssessmentItems(assessmentId: number, itemTypes?: string[]): Observable<AssessmentItem[]>;
   getSchoolId(): number;
-
-  // TODO: Technically not a provider method, but if we add one more export, let's break
-  // TODO: this out to an AssessmentExporter interface.
```
jtreuting: This TODO didn't let me get away with just adding the export writing trait scores here

webapp/src/main/webapp/src/app/assessments/model/export-items-request.model.ts
```diff
@@ -2,11 +2,12 @@ import { AssessmentItem } from "./assessment-item.model";
 import { DynamicItemField } from "./item-point-field.model";
 import { Assessment } from "./assessment.model";
 import { RequestType } from "../../shared/enum/request-type.enum";
+import { ExportRequest } from "./export-request.interface";
 
 /**
  * This model represents an Item by Points Earned or Distractor Analysis table export request.
  */
-export class ExportRequest {
+export class ExportItemsRequest implements ExportRequest {
```
jtreuting: the writing trait score export request is slightly different so made 2 implementations with a base interface that is used for common tasks like generating a filename for the export

webapp/src/main/webapp/src/app/assessments/results/view/distractor-analysis/distractor-analysis.component.ts
```diff
@@ -83,22 +90,14 @@ export class DistractorAnalysisComponent implements OnInit, ExportResults {
...
     exportRequest.pointColumns = this.choiceColumns;
     exportRequest.type = RequestType.DistractorAnalysis;
 
-
-    this.angulartics2.eventTrack.next({
-      action: 'Export DistractorAnalysis',
-      properties: {
-        category: 'Export'
-      }
-    });
```
jtreuting: removed these from the components because the analytics event is called from the exporter class so it was getting double called for each export

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -129,4 +130,42 @@ export class CsvExportService {
...
+
+  exportWritingTraitScores(exportRequest: ExportWritingTraitsRequest,
+                           filename: string) {
+
+    let compositeRows: any[] = [];
+    let maxPoints: number = 0;
+
+    // since there can be multiple items, we need to iterate over multiple summary table rows if there are multiple items that have writing trait scores
+    //  here we flatten the items and summary rows into a single array the CSV builder can iterate over
+    exportRequest.assessmentItems.forEach((item, i) => {
```
jtreuting: even though in reality there is a single WER item per assessment, on the UI we are handling displaying 2 summary writing trait tables if there happen to be 2 WER items on the assessment, so here we need to build up that array for the CSV builder to iterate over.  so this builds a row for each writing trait per WER item.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -41,6 +42,13 @@ export class AssessmentsComponent implements OnInit {
   @Input()
   assessmentProvider: AssessmentProvider;
 
+  /**
+   * The provider which implements the AssessmentExporter interface in order
+   * to export CSV data.
```
malaffoon: We're going to regret putting "CSV" in all the comments when AssessmentExporter gets additional methods that have nothing to do with CSV.

jtreuting: yeah good point

PR #789, opened 2018-01-25
Disable parameter padding for aggregate report queries.

wtritch:
Temporarily disable parameter padding to validate that it significantly improves aggregate report query performance.


wtritch:
Merging this for now.  This PR will be followed up with a cleanup PR to either re-enable parameter padding or remove it cleanly depending upon performance testing.

PR #791, opened 2018-01-26
DWR-1333 Add assessment repository

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
```diff
@@ -90,5 +96,9 @@ public void refreshRepositoryCache() {
         booleanRepository.findAll();
         strictBooleanRepository.findAll();
 
+        //TODO:
```
agorina: I do not like explicitly listing assessments here. I could add a method to `AssessmentType` to return all and loop here. Since it is in the RDW_Common, before I make a change I wanted to see if people have better ideas.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -112,25 +108,25 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
                 .addValue("isIncludeState", query.isIncludeState())
                 .addValue("isIncludeAllDistricts", query.isIncludeAllDistricts())
                 .addValue("isIncludeDistricts", (!query.isIncludeAllDistricts() && !query.getDistrictIds().isEmpty()))
-                .addValue("district_ids", query.getDistrictIds())
+                .addValue("district_ids", nullOrEmptyToDefault(query.getDistrictIds(), UNMATCHABLE_IDS))
```
agorina: Unrelated. This fixes a bug I introduced earlier and missed due to `ignored` tests.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
malaffoon: AssessmentType is an enum. All enums have `values()` that returns an array of all the enums.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/Assessment.java
```diff
@@ -10,11 +12,18 @@
     private int gradeId;
     private String subjectCode;
     private String label;
+    // if there are exams for this assessment, then it is a year in which it was administered
```
malaffoon: i suggest this comment go in the javadocs for the getter

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
agorina: Ah, thank you!

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/ActiveAssessmentRepository.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.olap.repository;
+
+import org.opentestsystem.rdw.common.model.AssessmentType;
+import org.opentestsystem.rdw.olap.model.Assessment;
+
+import java.util.List;
+
+/**
+ * Repository responsible for interfacing with the active assessment data store
```
malaffoon: i think this should define what "active" means?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/ActiveAssessmentRepository.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface ActiveAssessmentRepository {
+
+    /**
+     * Gets all {@link Assessment}s of the given {@link AssessmentType}
+     *
+     * @return all {@link Assessment}s;  cannot be null
```
malaffoon: nit: instead of "cannot be null" perhaps "will not be null, may be empty if there are no test results for the assessments"

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/Assessment.java
esotrope: Do we not want two different fields where both are nullable? not sure how things are setup in sql but this seems a little problematic if we are going to show this in UI or under some column name in csv no?

agorina: Not sure I understand the question. Could you please clarify?

esotrope: I'm wondering how examSchoolYear will be used in the UI/export.
I'm wondering if we put a table/csv header on it it will we be able to accurately describe what the data is?

agorina: Form the UI/export point it is just the `school year`. 
If this does not make sense, maybe we should chat?

PR #792, opened 2018-01-26
Feature/aggregate form organizations

esotrope:
image

image

image
TODO: add logic to hide schools of district that has flag set?

The "include all districts" checkbox is only displayed for statewide reporters and when clicked will disable the typeahead and hide the selected schools/districts
image


agorina:
About multiple checkboxes for 'Include all schools' or 'include district': I thought we want one of each for all schools/districts? (sorry if you mentioned it in your review and I just missed it)

esotrope:
@agorina wanted to throw this out there as a starting point. I would like to make an iteration of what you are suggesting next so we can compare

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-options.ts
```diff
@@ -6,23 +6,24 @@ import { Option as SbCheckboxGroupOption } from "../shared/form/sb-checkbox-grou
  */
 export interface AggregateReportFormOptions {
 
+  achievementLevelDisplayTypes: SbToggleOption[];
```
esotrope: alphabetized - yw @wtritch

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-organization-select.component.html
```diff
@@ -0,0 +1,10 @@
+<div>
+  <label>{{'labels.aggregate-reports.form.field.organization.label' | translate}}</label>
+  <organization-typeahead #typeahead
+                          [disabled]="value && value.stateSelection.includeAllDistricts"
+                          [options]="typeaheadOptions"
+                          (select)="onTypeaheadSelect($event)"></organization-typeahead>
```
esotrope: borrowed the organization typeahead from instructional resources - this typeahead does not have an "Add +" button so when you select an organization from the dropdown it clears your search and puts the organization into the list - not sure how people feel about that - the final impl should be consistent with the org export page whichever way we go

webapp/src/main/webapp/src/app/shared/form/abstract-control-value-accessor.ts
```diff
@@ -0,0 +1,87 @@
+import { ControlValueAccessor } from "@angular/forms";
+import { Input } from "@angular/core";
+import { Utils } from "../support/support";
+
+const NoOperation: () => void = () => {};
+
+/**
+ * Abstract class made to hide away the complexity of integrating with angular form control interface
+ */
+export abstract class AbstractControlValueAccessor<T> implements ControlValueAccessor {
```
esotrope: Made an abstract control value accessor - not sure why angular doesn't include this ootb actually...

webapp/src/main/webapp/src/app/shared/organization/organization.ts
```diff
@@ -0,0 +1,83 @@
+export interface Organization {
```
esotrope: created some common organization classes hopefully we can reuse these throughout the app.
Debated whether or not to add `readonly districtName?:string;` ancestor fields since we use them for some views but I'll revisit this later in a consolidation effort hopefully.

dsnyder-fairway:
> TODO: add logic to hide schools of district that has flag set?

This might have been addressed in a prior discussion. I'm not sure how exactly we want these checkboxes to behave, but additionally I believe when a user checks "Include all schools" the schools should populate with all the schools from that district. When a user removes one of those schools belonging to that district, the checkbox would be unchecked. Similarly when a user checks "Include district" and then later removes the district the checkbox should be unchecked.

esotrope:
@dsnyder-fairway interesting proposed behavior - that behavior would necessitate extra data lookups when the checkboxes are checked as the typeahead only populates with orgs a user has direct access to. flags can select orgs that a user may not have direct access to and only has indirect/transitive access to

esotrope:
@agorina @aaronsleeper @wtritch @jtreuting @dsnyder-fairway another iteration or the org form:
image

Thoughts?

agorina:
I like the layout, looks clean and simple to understand. 

I think "Rollup" is not the right term, since it is an overall state, not the rollup of the included organizations.

aaronsleeper:
agreed. rollup is jargon and we should think of an alternative. maybe just
"compare to state"

esotrope:
Another iteration

image

wtritch:
Yet Another Verbiage Comment:
Change "Include District" to "Include Parent District(s)"
I like @aaronsleeper 's comment about "rollup" being a little funky.  Since we use "Include" everywhere else, "Compare" may throw them as well... maybe "Include State Results" ?
Just thoughts, maybe we can discuss in the parking lot?

esotrope:
@wtritch @aaronsleeper I like "Include State Results"
I think for now I will change 
Include District -> Include Districts of selected Schools
Include All Schools -> Include Schools of selected Districts

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.resolve.ts
```diff
@@ -38,21 +38,27 @@ export class AggregateReportFormResolve implements Resolve<AggregateReportForm>
     return <AggregateReportFormSettings>{
       assessmentGrades: [],
       assessmentType: firstValueOf(options.assessmentTypes),
-      completenesses: valuesOf(options.completenesses),
+      completenesses: [ firstValueOf(options.completenesses) ],
```
esotrope: added defaults to Valid, Standardized and Complete

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -95,16 +174,28 @@ <h2 class="h3 gray-darkest">
...
-                <sb-checkbox-group [options]="options.schoolYears"
-                                   [(ngModel)]="settings.schoolYears"
-                                   [ngModelOptions]="{standalone: true}"
-                                   name="schoolYears"
-                                   analyticsEvent="FilterClick"
-                                   analyticsCategory="AggregateQueryBuilder"
-                                   [allOptionAnalyticsProperties]="{label: 'Academic Year: All'}"
-                                   (change)="generateReport()"></sb-checkbox-group>
+                <div class="form-group">
+                  <sb-checkbox-group [options]="options.schoolYears"
```
esotrope: changed school year control to be horizontal and have no all choice - also added validation

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -114,9 +205,6 @@ <h2 class="h3 gray-darkest">
                 <label info-button
                        title="{{'labels.aggregate-reports.form.field.interim-administration-condition.label' | translate}}"
                        content="{{'labels.aggregate-reports.form.field.interim-administration-condition.info' | translate}}"></label>
-                <p class="small gray-darker">
```
esotrope: Removed "Interim" and "Summative" subtext on admin condition form fields

esotrope:
Org select
image

After touch/failed form submit
image
School year display update:
image

Summary org section changes
image

image

jtreuting:
Is it considered "Multiple" organizations when you select a single school and include the district?

esotrope:
@jtreuting currently it doesn't consider the settings flags. Right now it only counts the listed schools or districts. Probably wont fit but I'm wondering if we shouldn't add little headers for some/each flag like "Include State Results"
image

esotrope:
Another iteration
image

PR #794, opened 2018-01-29
Fix PDF report chunk completion

wtritch:
This fixes an issue with completing student chunks during PDF report generation that was caused by the last PR.  Completing a chunk is a separate repository method that needs to increment the completed_chunk count as a metadata property.

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
@@ -98,6 +98,7 @@ public class FetchStudents {
         final Report updated = Report.builder()
                 .copy(report)
                 .metadata(TotalChunk, String.valueOf(chunks.size()))
+                .metadata(CompletedChunk, "0")
                 .build();
         reportService.update(updated);
 
--- a/report-processor/src/main/resources/application.sql.yml
+++ b/report-processor/src/main/resources/application.sql.yml
@@ -274,10 +274,11 @@ sql:
       delete from `user_report` WHERE `id` in (:ids);
 
     completeChunk: >-
-      update user_report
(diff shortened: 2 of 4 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/FetchStudents.java
```diff
@@ -98,6 +98,7 @@ public void fetchAndPublishStudents(final ReportRequestMessage message) {
         final Report updated = Report.builder()
                 .copy(report)
                 .metadata(TotalChunk, String.valueOf(chunks.size()))
+                .metadata(CompletedChunk, "0")
```
wtritch: Initialize the complete count to 0 when we initialize the total chunk count.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -274,10 +274,11 @@ sql:
...
 
     completeChunk: >-
-      update user_report
-        set complete_chunk_count = LAST_INSERT_ID(complete_chunk_count + 1)
+      update user_report_metadata
+        set value = cast(LAST_INSERT_ID(cast(value as unsigned) + 1) AS CHAR)
       where
-        id = :report_id;
+        report_id = :report_id and
+        name = 'completed_chunk_count';
```
wtritch: This is the fix.  It assumes that a "completed_chunk_count" metadata value exists for the given report.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcReportRepositoryIT.java
```diff
@@ -140,14 +140,23 @@ public void itShouldMarkChunksCompleted() {
                 .build();
 
         report = repository.create(report);
+        report = repository.update(report);
```
wtritch: We only set metadata on "update" not on "create" since they represent processing/result values.

PR #795, opened 2018-01-29
Fix User deserialization in report processor

wtritch:
In QA we're having issues deserializing User instances as part of a ReportRequestMessage.  This PR registers a mixin that tells Jackson how to deserialize User instances.

### DIFF
--- a/common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/User.java
+++ b/common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/User.java
@@ -83,7 +83,7 @@ public class User extends org.springframework.security.core.userdetails.User {
         private String lastName;
         private Map<String, Permission> permissionsById;
         private Map<Long, GroupGrant> groupsById;
-        private String password;
+        private String password = "[REDACTED]";
         private String username;
         private Collection<? extends GrantedAuthority> authorities;
         private boolean accountNonExpired;
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
@@ -5,8 +5,6 @@ import org.opentestsystem.rdw.common.status.StatusConfiguration;
 import org.opentestsystem.rdw.reporting.common.i18n.MessageSourceConfiguration;
 import org.opentestsystem.rdw.reporting.common.jdbc.SecurityParameterProvider;
 import org.opentestsystem.rdw.reporting.common.repository.DataSourceConfiguration;
-import org.opentestsystem.rdw.reporting.common.web.security.User;
-import org.opentestsystem.rdw.reporting.common.web.security.jackson2.UserMixin;
 import org.opentestsystem.rdw.reporting.common.web.security.jwt.JwtWebSecurityConfiguration;
 import org.opentestsystem.rdw.reporting.processor.configuration.RemoveStaleReportsConfiguration;
 import org.opentestsystem.rdw.reporting.processor.service.WkhtmltopdfWebServiceClient;
@@ -41,7 +39,7 @@ public class ApplicationConfiguration {
 
     @Bean
     public Jackson2ObjectMapperBuilderCustomizer customizeObjectMapper() {
-        return builder -> builder.mixIn(User.class, UserMixin.class);
+        return builder -> builder.failOnUnknownProperties(false);
     }
 
     @Bean
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
@@ -5,6 +5,8 @@ import org.opentestsystem.rdw.common.status.StatusConfiguration;
 import org.opentestsystem.rdw.reporting.common.i18n.MessageSourceConfiguration;
 import org.opentestsystem.rdw.reporting.common.jdbc.SecurityParameterProvider;
 import org.opentestsystem.rdw.reporting.common.repository.DataSourceConfiguration;
+import org.opentestsystem.rdw.reporting.common.web.security.User;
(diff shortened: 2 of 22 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
```diff
@@ -36,6 +39,11 @@
 @EnableGlobalMethodSecurity(prePostEnabled = true)
 public class ApplicationConfiguration {
 
+    @Bean
+    public Jackson2ObjectMapperBuilderCustomizer customizeObjectMapper() {
+        return builder -> builder.mixIn(User.class, UserMixin.class);
+    }
```
agorina: Just curious: do we know why it worked before?

esotrope: Yeah curious why this didn't show it's head immediately

wtritch: Good question.  I fear that it may have been broken since we first made the change to embed the entire User object in the message.  Although I would like to think we would have noticed that...

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/security/User.java
```diff
@@ -83,7 +83,7 @@ public PermissionScope getPermissionScopeByPermissionId(final String id) {
         private String lastName;
         private Map<String, Permission> permissionsById;
         private Map<Long, GroupGrant> groupsById;
-        private String password;
+        private String password = "[REDACTED]";
```
wtritch: This is a funny little change required to avoid having a "null" password in the injected User object.  Part of the JWT authentication process is to call org.springframework.security.core.userdetails.User#eraseCredentials() which sets the User password property to null.  Which means when passing it through serialization/deserialization during message processing I kept running into "Cannot create a User with a null password" exceptions.  This just changes the builder to have a placeholder default so that when deserializing a User without a password value I'm still able to instantiate a User.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
```diff
@@ -36,6 +37,11 @@
 @EnableGlobalMethodSecurity(prePostEnabled = true)
 public class ApplicationConfiguration {
 
+    @Bean
+    public Jackson2ObjectMapperBuilderCustomizer customizeObjectMapper() {
+        return builder -> builder.failOnUnknownProperties(false);
+    }
```
wtritch: This should have been added a while ago.  I believe it's a Spring default setting, but since we now rely on it during error handling I decided to make it explicit.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ReportRequestMessage.java
```diff
@@ -1,14 +1,11 @@
...
 
-import org.opentestsystem.rdw.reporting.common.web.security.User;
-
 /**
  * This message represents a basic report processing message.
  */
 public class ReportRequestMessage {
 
     protected long reportId;
-    protected User user;
```
wtritch: The authenticated User is no longer part of the message payload.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultMessageSecurityService.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.stereotype.Service;
+
+import static com.google.common.net.HttpHeaders.AUTHORIZATION;
+
+/**
+ * Default implementation of a MessageSecurityService using JWT and Bearer tokens as a message header.
+ */
+@Service
+public class DefaultMessageSecurityService implements MessageSecurityService {
```
wtritch: This service just takes a User, encodes it as a JWT, and adds a bearer token header to the message.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
```diff
@@ -76,10 +79,11 @@ private void publishReportRequest(final Report report, final User user) {
         logger.debug("Publishing report generation request: {}", report.getId());
         final ReportRequestMessage payload = ReportRequestMessage.builder()
                 .reportId(report.getId())
-                .user(user)
                 .build();
 
-        final Message<ReportRequestMessage> message = MessageBuilder.createMessage(payload, new MessageHeaders(new HashMap<>()));
+        final Message<ReportRequestMessage> message = messageSecurityService.withUser(
+                MessageBuilder.createMessage(payload, new MessageHeaders(new HashMap<>())),
+                user);
```
wtritch: I originally tried using an interceptor that would automagically encode/decode/set the security context principal without needing direct "stuff a user in my message/parse a user out of my message" handling in every consumer.  Unfortunately, the interceptor is executed on a separate thread from the `.send(message)` call below.  Since the "current user" is contained in a threadlocal property, the interceptor has no idea who the current user was when the message was published.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ReportFailingChannelInterceptor.java
```diff
@@ -49,13 +53,21 @@ public void afterReceiveCompletion(final Message<?> message,
     private void failReport(final Exception ex, final Message<?> message) {
         if (ex == null) return;
 
-        if (!ReportRequestMessage.class.isInstance(message.getPayload())) {
```
wtritch: So when we were using kyro for message payload encoding/decoding the interceptors were called with the decoded payloads.  Since we switched to Jackson for payload encoding/decoding the interceptors are now called with the serialized string payload, which is why all failing reports have been left as "pending" recently.

wtritch:
I'll merge this now but will continue to resolve comments as they come up.

PR #797, opened 2018-01-30
left navigation for aggregate reports and some improvements to the sample table

dsnyder-fairway:
Still need to work out

- subsections display and correctly mark as active when scrolling past them, however, clicking on them takes you to the parent on the section

screen shot 2018-01-29 at 4 53 22 pm
screen shot 2018-01-29 at 4 53 30 pm
screen shot 2018-01-29 at 4 57 18 pm
screen shot 2018-01-29 at 4 59 41 pm
screen shot 2018-01-29 at 4 59 58 pm


jtreuting:
Sweet.  Does it simplify things to not have the sub-sections?  If so, I'm not sure they are needed.  But would like to get other opinions on that.  I feel like we were leaning towards just the main sections.

webapp/src/main/webapp/src/app/shared/nav/left-nav.component.ts
```diff
@@ -0,0 +1,64 @@
+import { Component, HostListener, Inject, Input } from "@angular/core";
+import { DOCUMENT } from "@angular/common";
+import { WindowRefService } from "../core/window-ref.service";
+
+@Component({
+  selector: 'left-nav',
+  templateUrl: 'left-nav.component.html'
+})
+export class LeftNavComponent {
```
esotrope: hm. putting left in the title seems limiting. Maybe `ScrollNav`?

dsnyder-fairway:
leaving out subsection would definitely simplify this. I would be fine removing subsections. @esotrope, @wtritch thoughts?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
```diff
@@ -1,3 +1,122 @@
+<left-nav
```
esotrope: I'm thinking we need to move everything right so this never appears over/under anything.
I realize this will push the bottom of the page off to the right but i'm not sure which is worse. either way the nav needs to fit in the <container> width if it doesnt already

webapp/src/main/webapp/src/app/shared/nav/left-nav.component.ts
```diff
@@ -0,0 +1,64 @@
...
+    let number = this.window.pageYOffset || this.document.documentElement.scrollTop || this.document.body.scrollTop || 0;
+
+    let activatedSection = this.activateSection(this.sections, number);
+
+    if (activatedSection) {
+      activatedSection.isActive = true;
+      if (activatedSection.subsections) {
+        let activatedSubsection = this.activateSection(activatedSection.subsections, number - document.getElementById(activatedSection.scrollTo.id).offsetTop);
+        if (activatedSubsection)
+         activatedSubsection.isActive = true;
```
esotrope: I'm guessing there wasn't a magic `ngx-bootstrap-scroll-spy-nav` on npm already?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports.component.html
dsnyder-fairway: I put a max-width of 200px. Otherwise some of the subsections would appear to run into the `well`s. It does get completely ran over by the preview table too... not sure how we want to handle that. Still move that to the right?

esotrope:
@dsnyder-fairway looks pretty cool : j
I think we can remove subsections. Since the nav is to simplify things we don't want it to grow as long as the page if we can avoid it.
Also, i'm thinking we will want to make the sections with invalid fields appear red in the navigation so that the user can easily jump to them and correct them.

webapp/src/main/webapp/src/app/shared/nav/left-nav.component.ts
dsnyder-fairway: closest thing I found was a primeNg sidebar which wasn't really what we were looking for

dsnyder-fairway:
## No submenus
screen shot 2018-01-30 at 9 27 20 am

## centered preview table
screen shot 2018-01-30 at 9 27 09 am

PR #798, opened 2018-01-30
Feature/dwr 1306 trait score responses

jtreuting:
Implement student response writing trait scores in the item tab viewing.  Incorporated feedback from demo and added the Purpose plus Total Points row with info icon.

image


webapp/src/main/webapp/src/app/assessments/items/item-writing-trait-scores/item-writing-trait-scores.component.html
```diff
@@ -0,0 +1,64 @@
...
+          <span info-button title="{{'labels.student.responses.cols.correctness' | translate}}"
+                content="{{'labels.student.responses.cols.correctness-info' | translate}}"></span>
+    </th>
+  </tr>
+  </thead>
+  <tbody>
+  <tr>
+    <td>{{'enum.writing-trait.evidence' | translate}}</td>
+    <td>
+      <display-optional [value]="evidence"></display-optional>
```
jtreuting: i found myself doing the same null check to display either the value or a dash.  moved to a component so we can reuse across multiple places.

not sure I love this name but the idea is that it will display a value or a missing value.  am open to changing the name but was having writer's block (or maybe writers block, or possibly writers' block, not sure)

webapp/src/main/webapp/src/app/assessments/items/item-writing-trait-scores/item-writing-trait-scores.component.html
```diff
@@ -0,0 +1,64 @@
...
+    </td>
+    <td>4</td>
+    <td>
+      <display-optional [value]="evidence"
```
jtreuting: when the value that we check null against is not the same as what we would display, we set a separate displayValue that is used instead.  I took this approach instead of passing in a number format to leave it generic so it can be used for numbers or other data types as well.  this is for the correctness which is a decimal

webapp/src/main/webapp/src/app/assessments/items/item-writing-trait-scores/item-writing-trait-scores.component.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    return this.responsesAssessmentItem.writingTraitScores.evidence;
+  }
+
+  /**
+   * The max points for the evidence writing trait
+   * @returns {number}
+   */
+  get evidenceMaxPoints(): number {
+    return WritingTrait.evidence().maxPoints;
```
jtreuting: get the maximum points from the one place we store it and is the point of truth

webapp/src/main/webapp/src/app/shared/display-optional/display-optional.component.html
```diff
@@ -0,0 +1 @@
+<span [innerHTML]="display"></span>
```
jtreuting: i went back and forth on whether to have 2 spans with ngIf that would handle the logic or put the logic in the component so it always returned the appropriate value and the business logic was in there.  it felt cleaner there, so this becomes very simple.

webapp/src/main/webapp/src/app/student/responses/student-responses.component.html
```diff
@@ -60,8 +60,7 @@ <h3 class="blue mb-md">{{'labels.student.responses.title' | translate}}</h3>
     </p-column>
     <p-column field="score" header="{{'labels.student.responses.cols.student-score' | translate}}" sortable="true">
       <ng-template let-score="rowData.score" pTemplate="body">
-        <span *ngIf="score !== null">{{score}}</span>
-        <span *ngIf="score === null">&mdash;</span>
+        <display-optional [value]="score"></display-optional>
```
jtreuting: replace some other places where the same null check options were done in data tables

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -446,7 +446,8 @@
       "evidence": "Evidence / Elaboration",
       "organization": "Organization / Purpose",
       "conventions": "Conventions",
-      "total": "Total Points"
+      "total": "Total Points",
+      "total-info": "total points more info text here"
```
jtreuting: TODO: SB will be providing this real content, so this is a placeholder

webapp/src/main/webapp/src/styles.less
```diff
@@ -728,3 +728,7 @@ aggregate-reports {
     }
   }
 }
+
+.writing-trait-scores {
+  min-width: 60% !important;
```
jtreuting: don't love the important, but only other way i could get it to work was inline in the HTML and this felt better of those 2 options.  With a small table it felt odd being 100% width, so thought the display looked better with a smaller width since we know the content for the table.  (see screenshot)

webapp/src/main/webapp/src/app/assessments/items/item-writing-trait-scores/item-writing-trait-scores.component.ts
esotrope: Since these factory calls create new objects each call it may be good practice to call these once at component/class creation-time and then hold on to them for reference. Angular tends to call methods multiple times throughout its render process

webapp/src/main/webapp/src/styles.less
esotrope: You may have already tried but usually this can be avoided by taking the selector applying the style `min-width` and adding qualification to it.
e.g.
```css
.path .to > .thing {
  min-width: 100%;
}
.path .to > .thing.my-thing-variant {
  min-width: 60%;
}
```

webapp/src/main/webapp/src/app/shared/display-optional/display-optional.component.ts
```diff
@@ -0,0 +1,41 @@
+import { Component, Input } from "@angular/core";
+
+/**
+ * This component is responsible for displaying values that might be missing and need a default display
+ */
+@Component({
+  selector: 'display-optional',
+  templateUrl: './display-optional.component.html',
+})
+export class DisplayOptionalComponent {
```
esotrope: I think the angular way to do this is to use a pipe to avoid extra dom elements.
Could be something like this maybe:
```html
value | optional:defaultValue:processedValue
```
Or possibly with two pipes:
```html
value | and:processedValue | or:defaultValue
```
```typescript
@Pipe({ name: 'optional' })
class OptionalPipe {
  transform(value: any, defaultValue: any, processedValue: any): any {
    if (typeof value !== 'undefined') {
      return typeof processedValue !== 'undefined' ? processedValue : value;
    }
    return defaultValue;
  }
}
@Pipe({ name: 'and' })
class AndPipe {
  transform(value: any, processedValue: any): any {
    return typeof value !== 'undefined' ? processedValue : value;
  }
}
@Pipe({ name: 'or' })
class OrPipe {
  transform(value: any, defaultValue: any): any {
    return typeof value !== 'undefined' ? value : defaultValue;
  }
}
```

jtreuting: Can we actually do that when the processed value needs to go through a number pipe but the default value doesn't?

webapp/src/main/webapp/src/styles.less
jtreuting: i did not, will give that a shot

webapp/src/main/webapp/src/app/shared/display-optional/display-optional.component.ts
esotrope: Yeah i think so. it would look something like:
```html
value | optional:(value|number):'-'
value | and:(value|number) | or:'-'
```

jtreuting: gonna play with this after i merge.  i think that the code in here might be helpful reference for the PDF work that devin is starting on.  but will do a follow-up this afternoon to address this and the CSS if possible.

jtreuting: @esotrope I got the pipe working (basically).  The limitation is that if we want to use `&mdash;` instead of a "normal" `-` then we will need a component to bind the output to the `innerHTML` or else it will get encoded, which puts us back to where we were before.  If we are fine with the dash then the usage looks like this:

`{{ evidence | optional }}` in the simplest case and

`{{ evidence | optional:((evidence / evidenceMaxPoints) | number:'1.2-2') }}` more complex

esotrope: Not sure i understand the point about inner HTML. I would think the component using the pipe has the option to put the output into innerHtml or the DOM text node.
```html
<span [innerHtml]="value | pipe"></span>
<span>{{value | pipe}}</span>
```

jtreuting: yeah i guess that's the case, but since you don't know whether it is coming back with `&mdash;` or not you would always be adding that span.  so in a datatable where you have a template you couldn't just do `{{value | optional}}`, you'd have to do the span with the innerHTML.  but if it's a component you just drop the component and don't have to worry about that.

moving to a standard dash character makes it so it isn't an issue and i think it is a good resolution.

PR #799, opened 2018-01-30
Feature/dwr 1348 misc cleanup

jtreuting:
1. "Distractor Analysis" shall be replaced with "Key / Distractor Analysis".
2. In the "Average" summary at the top, replace "# Students" with "# Results".
3. "Custom Exports" shall be replaced with "District / School Exports"; both the button and the page title (and anywhere else you see it).
4. Move the "Export CSV" button down to the "Results" header line, perhaps just to the left of "Show value as".

image

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit } from "@angular/core";
+import {Component, EventEmitter, Input, OnInit, Output} from "@angular/core";
 import { ActivatedRoute } from "@angular/router";
 import { ordering } from "@kourge/ordering";
 import { FilterBy } from "./model/filter-by.model";
@@ -58,6 +58,9 @@ export class AssessmentsComponent implements OnInit {
    */
   @Input()
   allowFilterBySessions: boolean = true;
+
+  @Output()
+  export: EventEmitter<any> = new EventEmitter<any>();
 
   showValuesAsPercent: boolean = true;
   filterDisplayOptions: any = {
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -59,6 +59,9 @@ export class AssessmentsComponent implements OnInit {
   @Input()
   allowFilterBySessions: boolean = true;
 
+  @Output()
+  export: EventEmitter<any> = new EventEmitter<any>();
```
jtreuting: since we moved the export CSV button for the page to the Results header area, that is displayed from the "generic" assessment component so it needs to call back to the group or school/grade results component where the logic for the export actual is.

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
```diff
@@ -22,6 +21,7 @@
                [hideAssessments]="gradesAreUnavailable"
                [assessmentProvider]="assessmentProvider"
                [assessmentExporter]="assessmentExporter"
+               (export)="exportCsv()"
```
esotrope: Is this for all exports button clicks or only a certain type of export? may want to qualify if only one.

jtreuting: this is for exporting the assessments on the page.  i thought export was fine since it was scoped to the assessments component.  i didn't want to name it `exportCsv` so it's not specific and that is an implementation detail for exporting.  there are view specific ones within it.  if there is a change maybe `exportAll` or something like that since it is across all assessments shown on the page

esotrope: Ok. I wasn't sure if one of the subviews was doing the "export all" or not - sounds good

PR #800, opened 2018-01-30
Feature/dwr 1299

jtreuting:
Fixed Rubric and Exemplar so that it only uses the answer key from the database for multiple choice and multiple select, and gets the others from the item XML.  This will fix the issue of "Hand" and "qrx" showing in the rubric and exemplar tab.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
@@ -1,66 +1,73 @@
-<div *ngIf="loading" class="loader">
-  <div class="inner">
-    <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'messages.loading' | translate }}</span>
-  </div>
+<div *ngIf="hasAnswerKey(item)">
+  <strong>{{translateRoot + 'answer-key' | translate }}</strong>
+  {{item.answerKey}}
 </div>
-<!-- TODO: Remove custom style of white-space: normal once @asleeper's latest update
-     TODO: updates the white-space:nowrap style implementation-->
-<div *ngIf="model" style="white-space: normal">
-  <div *ngIf="model.answerKeyValue" class="pad-bottom-10">
-    <strong>{{ translateRoot + 'answer-key' | translate }}</strong> {{model.answerKeyValue}}
+
+<div *ngIf="!hasAnswerKey(item)">
+  <div *ngIf="loading" class="loader">
+    <div class="inner">
+      <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'messages.loading' | translate }}</span>
+    </div>
   </div>
-
-  <div *ngIf="model.rubrics.length > 0">
-    <div *ngIf="model.rubrics.length ==1" [innerHtml]="model.rubrics[0].templateHtml"></div>
-
-    <div *ngIf="model.rubrics.length > 1">
-      <h4>{{ translateRoot + 'rubric' | translate }}</h4>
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
```diff
@@ -1,66 +1,73 @@
...
+  <strong>{{translateRoot + 'answer-key' | translate }}</strong>
+  {{item.answerKey}}
 </div>
-<!-- TODO: Remove custom style of white-space: normal once @asleeper's latest update
-     TODO: updates the white-space:nowrap style implementation-->
-<div *ngIf="model" style="white-space: normal">
-  <div *ngIf="model.answerKeyValue" class="pad-bottom-10">
-    <strong>{{ translateRoot + 'answer-key' | translate }}</strong> {{model.answerKeyValue}}
+
+<div *ngIf="!hasAnswerKey(item)">
```
jtreuting: this is mostly formatting.  added the display logic to the item exemplar instead of the item tabs.  so added wrapping div around this t control the display when there is no answer key to show

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
```diff
@@ -10,10 +11,11 @@ import { Utils } from "../../../shared/support/support";
 export class ItemExemplarComponent implements OnInit {
 
   /**
-   * The bank item key of an item.
+   * The item which we want to display the exemplar for.
    */
   @Input()
-  public bankItemKey: string;
+  public item: AssessmentItem;
```
jtreuting: pass the item instead of just the key since we are using it to determine if the item type is appropriate for using the answer key from the DB

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
```diff
@@ -48,4 +50,13 @@ export class ItemExemplarComponent implements OnInit {
...
         });
   }
+
+  /**
+   * Returns true if this item is a MC or MS and has an answer key
+   * @param {AssessmentItem} item
+   * @returns {boolean}
+   */
+  hasAnswerKey(item: AssessmentItem): boolean {
+    return (item.type === 'MC' || item.type === 'MC') && (item.answerKey !== undefined || item.answerKey !== null);
```
esotrope: If we use this elsewhere it might be good to make an `AssessmentItems` utility class hold this logic

malaffoon:
Guess i'm confused about this ... i thought we wanted to show `key=Hand` or whatever if we couldn't get info from the XML? I thought the real problem with this was that it wasn't properly getting the XML.

jtreuting:
The problem was that we had an explicit check that if there was a value in the database we would use it and not try to go to the item XML.  You can see that logic here: https://github.com/SmarterApp/RDW_Reporting/pull/800/files#diff-08a1927ba9aa67e9fcc1a2c158377d66L22

So instead I made it more intelligent to only use the DB value if it exists and is a MC or MS since that has a simple answer key we can display that is more accurate than the Item XML value most likely.

We are not checking to see if the value doesn't exist in the Item XML.  We could add some logic to see what comes back from that service call and if it is empty or it errors out we can display the `Key: Hand`.  I'm up for adding that if we think it is better than saying it couldn't be found.

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
jtreuting: This logic is not used anywhere else as far as I know.

PR #802, opened 2018-01-30
DWR-1350 Log error when duplicate ids are found.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcOrganizationRepository.java
```diff
@@ -36,34 +38,49 @@
...
-                .collect(toMap(School::getId, s -> s));
+        try {
+            return template.query(
+                    findAllSchoolsQuery,
+                    (row, index) -> Organizations.map(row))
+                    .stream()
+                    .map(s -> ((School) s))
+                    .collect(toMap(School::getId, s -> s));
+        } catch (final Exception ex) {
+            logger.warn("Duplicate school id", ex);
```
agorina: Is this the right log level?

esotrope: Maybe, WARN is something that can be taken action on to fix in the system/data OR something that may/will produce undesired behavior in the system i think

malaffoon: Since this will happen only if there is bad data in the system then i think WARN is appropriate.

agorina: Just to note: I have create a JIRA (DWR-1358) to consider adding validation steps to migrate to prevent from bad data being loaded into reporting DBs.

PR #805, opened 2018-01-30
byString ordering does not handle undefined.  protect against this so…

jtreuting:
… sorting can still happen in the case where there is no group provided in the typeahead Option

The underlying cause is that when the user only has access to schools, when the `/api/organizations/districts' endpoint is called it returns nothing, and that is being used to populate the grouping property for the typeahead options.  So this fix cause the typeahead to work and not blow up.

I'll be creating another lower priority story to refactor some things to better solve the underlying issue.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
@@ -70,7 +70,7 @@ export class SBTypeahead implements OnInit {
         ? options
           .sort(
             join(
-              ordering(byString).on<Option>(o => o.group).compare,
+              ordering(byString).on<Option>(o => o.group === undefined ? '' : o.group).compare,
               ordering(byString).on<Option>(o => o.label).compare
             ))
         : [];


webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -70,7 +70,7 @@ export class SBTypeahead implements OnInit {
         ? options
           .sort(
             join(
```
esotrope: Wondering if we shouldn't pull out a boolean param `@Input() group: boolean = true` so we can selectively group.Then this comparator doesn't have to check option group existence at sort-time
```typescript
options.sort(
  this.group ? join(/* existing code */) : ordering(byString).on<Option>(o => o.label).compare
)
```

jtreuting: we could, but it makes me nervous that it could blow up without this and make it so that no one can do anything in the site.

jtreuting: plus we would have to know when this is the case and so we'd have to analyze the objects in the Option array first to set that flag right?

esotrope: I was thinking the component declaring the typeahead could configure it `<typeahead [group]="true">

esotrope: Actually ideally, since is the declaring component already provides the options it should just provide sorted options

PR #807, opened 2018-01-31
Implement optional pipe to replace display-optional component

jtreuting:
I took @esotrope advice and implemented an optional pipe instead of the display-optional component.  The only "downside" is that we can't use `&mdash;` as the missing value because it gets encoded and we would need to bind it to `[innerHTML]` of a span and then would need a component anyways.  So I've changed the missing value to just a "normal" -

If that feels like a deal breaker then we can ignore this PR and close it.


webapp/src/main/webapp/src/app/assessments/items/item-writing-trait-scores/item-writing-trait-scores.component.html
```diff
@@ -7,57 +7,53 @@ <h5>{{'labels.student.responses.writing-trait-purpose' | translate:{traitType: r
...
   </thead>
   <tbody>
   <tr>
     <td>{{'enum.writing-trait.evidence' | translate}}</td>
     <td>
-      <display-optional [value]="evidence"></display-optional>
+      {{ evidence | optional }}
     </td>
-    <td>4</td>
+    <td>{{ evidenceMaxPoints }}</td>
```
jtreuting: noticed that I had left the hard-coded max points in the table so fixing those

webapp/src/main/webapp/src/app/assessments/items/item-writing-trait-scores/item-writing-trait-scores.component.ts
```diff
@@ -31,6 +31,14 @@ export class ItemWritingTraitScoresComponent {
     return WritingTrait.evidence().maxPoints;
   }
 
+  /**
+   * Return the correctness value for the evidence writing trait score
+   * @returns {number}
+   */
+  get evidenceCorrectness(): number {
```
jtreuting: moved the logic for calculating the correctness to the component and not in the markup for all of these.  so not related to the overall PR just cleaning up

webapp/src/main/webapp/src/app/shared/optional.pipe.ts
```diff
@@ -0,0 +1,24 @@
...
+   */
+  transform(value: any, displayValue: any, missingValue: any) {
+    const defaultMissingValue: string = '-';
+
+    if (missingValue == null) missingValue = defaultMissingValue;
+
+    if (value === null) return missingValue;
+
+    return displayValue == null ? value : displayValue;
+  }
```
esotrope: I would be tempted to make the default value `''` for broader use.
Also for default params - in typescript we can simplify the above code by doing this:
```typescript
transform(value: any, displayValue: any, missingValue: any = '-') {
```
instead of:
```typescript
transform(value: any, displayValue: any, missingValue: any) {
  const defaultMissingValue: string = '-';
  if (missingValue == null) missingValue = defaultMissingValue;
```
Also, generally undefined is intended for interpretation as "not present"
we might want to use a util method here like `Utils.isNullOrUndefined` instead of null checks to be more robust

PR #808, opened 2018-01-31
Feature/async agg reports generation

wtritch:
This PR adds handling an AggregateRequestMessage, aggregate report generation, writing to S3, and sending an AggregateResponseMessage to the aggregate service.

See diagram below
aggregate-report-flow 2


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
```diff
@@ -0,0 +1,125 @@
...
+                            final ArchiveService archiveService) {
+        this.processor = processor;
+        this.messageSecurityService = messageSecurityService;
+        this.service = service;
+        this.objectMapper = objectMapper;
+        this.archiveService = archiveService;
+    }
+
+    @StreamListener(AggregateRequest)
+    public void handleAggregateRequest(final Message<AggregateRequestMessage> message) {
```
wtritch: This is the big change.  This is where the aggregate-service receives, processes, and responds to report requests.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private void archiveJson(final File jsonfile, final URI targetLocation) throws IOException {
+        final Properties properties = new Properties();
+        properties.put(Headers.CONTENT_LENGTH, jsonfile.length());
+        properties.put(Headers.CONTENT_TYPE, APPLICATION_JSON.toString());
+
+        try (final FileInputStream input = new FileInputStream(jsonfile)) {
+            archiveService.writeResource(targetLocation.toASCIIString(), input, properties);
+        }
```
wtritch: Since these payloads may be large (100MB) I didn't want to double our List<ReportRow> results in memory.  Instead I'm writing the JSON to a temp file to calculate size before persisting via the ArchiveService.

common/src/test/java/org/opentestsystem/rdw/reporting/common/test/support/ReportQueries.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.google.common.collect.ImmutableSet;
+import org.opentestsystem.rdw.reporting.common.model.ReportQuery;
+
+import static org.opentestsystem.rdw.reporting.common.model.DimensionType.Gender;
+
+/**
+ * Test utility for generating ReportQuery test objects.
+ */
+public class ReportQueries {
```
wtritch: I found myself creating a ReportQuery instance in a few places and decided to create a test utility to create one for me.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/jackson/ReportModule.java
```diff
@@ -10,5 +12,6 @@
 
     public ReportModule() {
         addDeserializer(AbstractExamReportRequest.class, new AbstractExamReportRequestDeserializer());
+        setMixInAnnotation(ReportQuery.class, ReportQueryMixin.class);
```
wtritch: I was having issues round-tripping the ReportQuery objects through Jackson because the properties didn't match the fields.  I added a fields-only mixin to help.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ExamReportService.java
```diff
@@ -58,6 +59,16 @@
...
 
+    /**
+     * Creates an aggregate report
+     *
+     * @param user    the user requesting the report
+     * @param request request for the aggregate report
+     * @return the created report handle
+     */
+    @PreAuthorize("hasAuthority('PERM_CUSTOM_AGGREGATE_READ')")
+    Report createReport(@NotNull User user, @NotNull AggregateReportRequest request);
```
wtritch: I added a permission-gated handler to the report-processor for requesting aggregate reports.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReportRequestProcessor.java
```diff
@@ -15,12 +13,20 @@
...
     String AggregateRequest = "AggregateRequest";
     String AggregateResponse = "AggregateResponse";
+    String AggregateReportsOut = "ReportAggregateReportsOut";
+    String AggregateReportsIn = "ReportAggregateReportsIn";
+
+    /**
+     * @return Internal processing output channel
+     */
+    @Output(AggregateReportsOut)
+    MessageChannel aggregateReportsOut();
```
wtritch: I consolidated all the message channel interfaces into three classes representing the three report flows: pdf, exam export, and aggregate reports.

report-processor/src/main/resources/application.yml
```diff
@@ -15,38 +15,68 @@ spring:
   cloud:
     stream:
       bindings:
-        ReportExportExams:
+        ReportExportExamsOut:
```
wtritch: This was a fun bug to uncover.  Each of these bindings and the associated `@Input` `@Output` MessageChannels above is essentially a spring bean.  Trying to bind both an Input and an Output to the same binding was confusing Spring.  The correct way to associate bindings to MessageChannels is via a 1-to-1 relationship.  An Input and an Output can share the same underlying queue/topic via declaring the same "destination" property in the binding.  See Gary Russel's answer [here](https://stackoverflow.com/questions/39558134/spring-cloud-stream-app-dispatcher-has-no-subscribers-for-channel-unknown-chan) for more info.

wtritch: @malaffoon this may have accounted for some of those Pending-locked reports you were seeing in production.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ExamReportService.java
esotrope: 👍

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReportRequestProcessor.java
esotrope: i like the in/out change - makes it more clear

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
```diff
@@ -0,0 +1,125 @@
...
+
+import static org.opentestsystem.rdw.olap.stream.AggregateRequestProcessor.AggregateRequest;
+import static org.opentestsystem.rdw.reporting.common.model.ReportStatus.COMPLETED;
+import static org.opentestsystem.rdw.reporting.common.model.ReportStatus.FAILED;
+import static org.springframework.http.MediaType.APPLICATION_JSON;
+
+/**
+ * This component is responsible for processing aggregate report requests arriving
+ * via a message.  A report is generated, persisted to the archive service, and
+ * a response message is published to the response message channel.
```
malaffoon: 👍

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/jackson/ReportModule.java
malaffoon: Since ReportQuery is a dumb data container, can't we just fix ReportQuery to make properties and fields match? Or is it because there are no setters?

wtritch: There are no setters, and this *may* no longer be required.  The problem before was we had map properties represented as set fields (districts vs districtIds) and the difference was causing Jackson conniptions.  After @agorina 's latest changes we now may have properties that match fields, I'll check.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReportRequestProcessor.java
malaffoon: Hmm, i lack your perception. I find the channel names confusing. Does processing an AggregateRequest cause an AggregateReportsOut or an AggregateResponse?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
malaffoon: So i just realized you might not have to do this. If the objectMapper can produce an inputstream, you can push that directly into writeResource. That method will, if the length is unknown, write to a temporary file then copy it for you.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReportRequestProcessor.java
wtritch: the report-processor publishes to AggregateReportsOut, the report-processor subscribes to AggregateReportsIn, the report-processor publishes to AggregateRequest, the aggregate-service subscribes to AggregateRequest.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
wtritch: Unfortunately, the objectMapper cannot write to an InputStream, it can only write to an OutputStream.  I could use PipedInputStream/PipedOutputStream to convert from the output to an input, but I'd need a Thread/ThreadExecutor to do the shoveling.  And if the ArchiveService is going to be writing to the file-system to calculate the size anyway...

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/jackson/ReportModule.java
wtritch: Removed!  We now have matching properties and fields on the ReportQuery object.

PR #809, opened 2018-01-31
Feature/aggregate report option improvements

esotrope:
Made a dedicated PR for this instead of wrapping it into form submission.

This PR makes all options provided by the backend.
It also rigs the sorting of completeness and strict booleans with a sort function in UI


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportOptions.java
```diff
@@ -98,35 +184,69 @@ public Builder subjects(final List<CodedEntity> subjects) {
...
+
+        /**
+         * Creates an immutable list from the given values.
+         * If the value is null it returns an empty immutable list.
+         *
+         * @param values the values to create the list from
+         * @param <T> the type of values in the list
+         * @return an immutable list copy of the values or an empty list if the values are null
+         */
+        private <T> ImmutableList<T> createImmutableList(final Iterable<T> values) {
```
malaffoon: nit: don't need to change but these methods often have `safe` (for null) and `of` (for transformation) in their name, e.g. `safeImmutableListOf(...)`

webapp/src/main/webapp/src/app/aggregate-report/results/mock-aggregate-reports-preview.service.ts
```diff
@@ -47,14 +47,15 @@ export class MockAggregateReportsPreviewService {
...
     array = this.generateRowsForOption(options, 'LEP', array, settings.limitedEnglishProficiencies, organizationNames, settings.schoolYears, settings.assessmentGrades);
     array = this.generateRowsForOption(options, 'MigrantStatus', array, settings.migrantStatuses, organizationNames, settings.schoolYears, settings.assessmentGrades);
     array = this.generateRowsForOption(options, 'EconomicDisadvantage', array, settings.economicDisadvantages, organizationNames, settings.schoolYears, settings.assessmentGrades);
-    array = this.generateRowsForOption(options, 'IEP', array, settings.ieps, organizationNames, settings.schoolYears, settings.assessmentGrades);
-    array = this.generateRowsForOption(options, 'Section504', array, settings.plan504s, organizationNames, settings.schoolYears, settings.assessmentGrades);
+    array = this.generateRowsForOption(options, 'IEP', array, settings.individualEducationPlans, organizationNames, settings.schoolYears, settings.assessmentGrades);
+    array = this.generateRowsForOption(options, 'Section504', array, settings.section504s, organizationNames, settings.schoolYears, settings.assessmentGrades);
```
agorina: Just curious: have we decided how we are going to display all these `booleans` in the final report? If the data is disaggregated by let's say section504 and IEP, would it say : `section504-yes`, `section504-no` and so on?

esotrope: I think right now it says things like:
`Section 504: Yes`
`IEP: No`

PR #810, opened 2018-01-31
DWR-1324 OLAP API: modify Report Query object and API - part 2

agorina:
I find this change ...convoluted.

There are three ways to choose districts: 
- includeAllDistricts
- includeAllDistrictsOfSchools
- districtIds

And two ways to select schools:
- includeAllSchoolsOfDistricts
- schoolIds

Since for example 'all districts' are selected with a different query than 'district by ids', and then we combine all results, I need to determine an overlap to avoid duplicate rows.
Additionally, when `includeAllSchoolsOfDistricts` is specified, the district(s) have to be selected (using one of the three above options). And a similar case with 'includeAllDistrictsOfSchools'.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
@@ -1,6 +1,5 @@
 package org.opentestsystem.rdw.olap.repository.impl;
 
-import com.google.common.collect.ImmutableList;
 import org.opentestsystem.rdw.common.model.AssessmentType;
 import org.opentestsystem.rdw.olap.model.Assessment;
 import org.opentestsystem.rdw.olap.model.Dimension;
@@ -25,12 +24,9 @@ import java.util.Set;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Lists.newArrayList;
 import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
-import static org.opentestsystem.rdw.reporting.common.model.ReportQuery.checkIsValid;
 
 @Repository
 class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRepository {
-
-    private static final List<Integer> UNMATCHABLE_INT_IDS = ImmutableList.of(-1);
 
     private final NamedParameterJdbcTemplate jdbcTemplate;
     private final ReportQueryProvider customAggregateQueryProvider;
@@ -46,7 +42,6 @@ class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRep
     }
 
     private static void validate(final ReportQuery query) {
-        checkIsValid(query);
         checkArgument(query.getAssessmentTypeId() != AssessmentType.IAB.id(), "custom aggregate report for IAB is not supported yet");
     }
 
@@ -102,8 +97,7 @@ class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRep
     @Override
     public int estimateReportRowCount(@NotNull final ReportQuery query) {
 
-        validate(query);
-
+//        validate(query);
 //        final Set<DimensionType> dimensionTypes = query.getDimensionTypes();
 //        final MapSqlParameterSource parameterSource = new MapSqlParameterSource()
 //                .addValue("school_years", query.getSchoolYears())
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
@@ -1,17 +1,6 @@
 package org.opentestsystem.rdw.olap.sqlbuilder;
 
 import com.google.common.collect.ImmutableList;
-import org.opentestsystem.rdw.olap.model.Assessment;
-import org.opentestsystem.rdw.olap.model.CodedEntity;
-import org.opentestsystem.rdw.olap.model.Dimension;
-import org.opentestsystem.rdw.olap.repository.ActiveAssessmentRepository;
-import org.opentestsystem.rdw.olap.repository.BooleanRepository;
-import org.opentestsystem.rdw.olap.repository.CodedEntityRepository;
-import org.opentestsystem.rdw.olap.repository.EthnicityRepository;
-import org.opentestsystem.rdw.olap.repository.GenderRepository;
-import org.opentestsystem.rdw.olap.repository.OrganizationRepository;
-import org.opentestsystem.rdw.olap.repository.StrictBooleanRepository;
-import org.opentestsystem.rdw.olap.repository.SubjectRepository;
 import org.opentestsystem.rdw.olap.sqlbuilder.QueryWithReportTemplate.Builder;
 import org.opentestsystem.rdw.reporting.common.model.DimensionType;
 import org.opentestsystem.rdw.reporting.common.model.Organization;
@@ -24,9 +13,10 @@ import org.springframework.stereotype.Component;
 
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import static com.google.common.collect.Lists.newArrayList;
+import static java.util.stream.Collectors.toSet;
+import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
 import static org.opentestsystem.rdw.olap.sqlbuilder.QueryWithReportTemplate.builder;
 import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.nullOrEmptyToDefault;
 import static org.opentestsystem.rdw.reporting.common.model.DimensionType.EconomicDisadvantage;
@@ -35,8 +25,8 @@ import static org.opentestsystem.rdw.reporting.common.model.DimensionType.Gender
 import static org.opentestsystem.rdw.reporting.common.model.DimensionType.IEP;
 import static org.opentestsystem.rdw.reporting.common.model.DimensionType.LEP;
 import static org.opentestsystem.rdw.reporting.common.model.DimensionType.MigrantStatus;
-import static org.opentestsystem.rdw.reporting.common.model.DimensionType.Overall;
 import static org.opentestsystem.rdw.reporting.common.model.DimensionType.Section504;
+import static org.opentestsystem.rdw.reporting.common.model.ReportQuery.checkIsValid;
 
 /**
  * This implementation is responsible for constructing a custom aggregate query.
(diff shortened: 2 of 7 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -25,13 +24,10 @@
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Lists.newArrayList;
 import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
-import static org.opentestsystem.rdw.reporting.common.model.ReportQuery.checkIsValid;
 
 @Repository
 class JdbcCustomAggregateReportRowRepository implements CustomAggregateReportRepository {
 
-    private static final List<Integer> UNMATCHABLE_INT_IDS = ImmutableList.of(-1);
-
```
agorina: Just cleaned up.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
```diff
@@ -54,81 +44,64 @@
...
+                                        final ReportQueryRepositoryHelper repositoryHelper) {
         this.queryProvider = queryProvider;
-        this.organizationRepository = organizationRepository;
-        this.activeAssessmentRepository = activeAssessmentRepository;
-        this.genderRepository = genderRepository;
-        this.ethnicityRepository = ethnicityRepository;
-        this.booleanRepository = booleanRepository;
-        this.strictBooleanRepository = strictBooleanRepository;
-        this.subjectRepository = subjectRepository;
+        this.repositoryHelper = repositoryHelper;
```
agorina: I moved all the data manipulations that should have been done by the repositories (if we were not avoiding calling Redshift) into a separate helper.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
```diff
@@ -54,81 +44,64 @@
...
+        }
+    }
 
-            if (query.isIncludeAllDistrictsOfSchools()) {
-                //TODO:
-            }
+    private void appendDistricts(final ReportQuery query, final List<QueryWithReportTemplate> queries, final Builder dimensionTemplateBuilder) {
+        if (query.isIncludeAllDistricts()) {
+            queries.add(buildQuery(builder().addOn(allDistrictsAddOn).districts(repositoryHelper.findAllDistricts()).copy(dimensionTemplateBuilder)));
+        } else if (!query.getDistrictIds().isEmpty() || query.isIncludeAllDistrictsOfSchools()) {
```
agorina: I am making 'isIncludeAllDistricts' option mutually exclusive with other district selection options. Let me know if you disagree.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
```diff
@@ -54,81 +44,64 @@
...
-                        .copy(dimensionTemplateBuilder)));
+        if (!query.getSchoolIds().isEmpty()) {
+            //to avoid duplicate schools we need to filter those that are already included as part of districts selection
+            Set<Long> schoolIds = query.getSchoolIds();
+            if (query.isIncludeAllSchoolsOfDistricts()) {
+                final Set<Long> schoolsIncludedAsAllOfDistricts = repositoryHelper.findSchoolsIdsByDistrictIds(districtsWithAllSchools);
+                schoolIds = schoolIds.stream().filter(s -> !schoolsIncludedAsAllOfDistricts.contains(s)).collect(toSet());
             }
-
+            if (!schoolIds.isEmpty()) queries.add(buildQuery(toBuilderWithSchoolIds(schoolIds).copy(dimensionTemplateBuilder)));
```
agorina: This method makes my head spin. Does anybody see a way to simplify it?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/QueryWithReportTemplate.java
```diff
@@ -104,31 +103,16 @@ Builder districts(final Collection<District> districts) {
             return this;
         }
 
-        Builder district(final District district) {
-            this.organizations.add(district);
-            return this;
-        }
-
```
agorina: These were not used after the changes.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryRepositoryHelper.java
```diff
@@ -0,0 +1,202 @@
...
+import static org.opentestsystem.rdw.reporting.common.model.DimensionType.Overall;
+import static org.opentestsystem.rdw.reporting.common.model.DimensionType.StudentEnrolledGrade;
+
+/**
+ * Helper to manipulate the data from different repositories into format suitable for the query builders.
+ * This helpers uses multiple repositories that are expected to be `cacheable`.
+ * It is only here to avoid multiple round-trips to Redshift.
+ */
+@Service
+public class ReportQueryRepositoryHelper {
```
agorina: Better name?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
```diff
@@ -10,7 +10,7 @@
  * Represents a query for the custom aggregate report
  */
 public class ReportQuery {
-    private Integer assessmentTypeId;
+    private int assessmentTypeId;
```
agorina: Addressing code review feedback from the previous PR.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
```diff
@@ -135,20 +135,26 @@ public boolean contains(final DimensionType dimension) {
...
         checkArgument(!query.getSchoolYears().isEmpty(), "report query is missing school years");
         checkArgument(!query.getDimensionTypes().isEmpty(), "report query is missing dimensionTypes");
         checkArgument(!(!query.isIncludeState() && !query.isIncludeAllDistricts() && query.getDistrictIds().isEmpty() && query.getSchoolIds().isEmpty()), "report query is missing organization");
+
+        if (query.isIncludeAllDistrictsOfSchools()) {
+            checkArgument(!query.getSchoolIds().isEmpty(), "report query is missing schools");
+        }
+        if (query.isIncludeAllSchoolsOfDistricts()) {
+            checkArgument(!query.getDistrictIds().isEmpty() || query.isIncludeAllDistricts(), "report query is missing districts");
+        }
```
agorina: I will add unit tests for this change.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
esotrope: 👍

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
```diff
@@ -54,81 +44,64 @@
...
+        }
 
-            if (!query.getSchoolIds().isEmpty()) {
-                queries.add(buildQuery(toBuilderWithSchoolIds(query.getSchoolIds())
-                        .copy(dimensionTemplateBuilder)));
+        if (!query.getSchoolIds().isEmpty()) {
+            //to avoid duplicate schools we need to filter those that are already included as part of districts selection
+            Set<Long> schoolIds = query.getSchoolIds();
+            if (query.isIncludeAllSchoolsOfDistricts()) {
+                final Set<Long> schoolsIncludedAsAllOfDistricts = repositoryHelper.findSchoolsIdsByDistrictIds(districtsWithAllSchools);
```
esotrope: Probably missing something but I'm wondering wher these IDs go? don't we need to add them to the query?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
esotrope: All of this query code makes my head spin x l
The comments are helpful though

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
agorina: Yes, they end up as parameters to sql query. 

This was done in other PRs, but just to clarify:
`JdbcCustomAggregateReportRowRepository` uses `CustomAggregateQueryProvider` to build a set of `QueryWithReportTemplate`.

Each `QueryWithReportTemplate` contains a SQL string, MapSqlParameterSource loaded with parameters, and a report template that represents all expected rows with `empty` measures.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
wtritch: Maybe get the set of ids that you'll be operating on separately from including the queries?
```
final Set<Long> districtIds = getDistrictIds(query);
final Set<Long> districtIdsWithAllSchools = query.isIncludeAllSchoolsOfDistricts() ? districtIds : Collections.emptySet();
final Set<Long> schoolIds = query.getSchoolIds().removeAll(findSchoolIdsByDistrictIds(districtIdsWithAllSchools));
```
That might help keep the methods smaller and more intuitive?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryRepositoryHelper.java
wtritch: The calls in here make me pretty nervous.  I don't see a security layer or the User being passed at all... If a district-level user says "give me everything" are we going to pass a list of 1000 district ids to redshift to have 999 filtered out by permissions?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
wtritch: Hm... I'm not sure we *want* to validate this.  I'd prefer to just no-op if a user says "give me the district rollup for no schools"
My thought is that the ReportQuery is a live object in that the user should be able to toggle checkboxes on and off and the only property that should change is the value being toggled.  If the user clears the schools from their query and forgets to uncheck the "include district rollup" I don't think they should be presented with an error, they just don't get any districts added to their results.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProvider.java
agorina: I had this way and changed hoping that this is more readable. Will change it back.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryRepositoryHelper.java
agorina: I think you are raising a few concerns:
1. Security. This is done in the service layer and I believe it will stop the call to the repo if a user does not have appropriate permissions. @esotrope , please correct me if needed.
2. Too may IN parameters. It is applicable in three cases:
- individual schools;
- individual districts;
- 'isIncludeAllSchoolsOfDistricts'.

While I do not have a solution to the first two, the last could be addressed with a different SQL template.  I admit being lazy and reusing the same approach as for the individual schools. Considering the timeline I probably would like to  create a separate JIRA to revisit it later.  Thoughts?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
agorina: Makes sense. Will change.

agorina:
Since @wtritch needs to move some code around, I am merging this and will address any new comments later (if needed).

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryRepositoryHelper.java
agorina: Created: https://jira.fairwaytech.com/browse/DWR-1324

PR #811, opened 2018-02-01
Have the item viewer tab pull the answer key from the DB instead of q…

jtreuting:
…uerying the scoring API endpoint which reads from the Item XML.

Now that we have the answer key in the DB we don't need to query the Item XML to get the data.  We can use the value that we already have on the AssessmentItem to display the answer key for MS and MC on the Item Viewer tab.

For visual context:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
@@ -1,9 +1,9 @@
-<div *ngIf="item.validAnswerKey">
+<div *ngIf="item.answerKey">
   <strong>{{translateRoot + 'answer-key' | translate }}</strong>
-  {{ item.validAnswerKey }}
+  {{ item.answerKey }}
 </div>
 
-<div *ngIf="!item.validAnswerKey">
+<div *ngIf="!item.answerKey">
   <div *ngIf="loading" class="loader">
     <div class="inner">
       <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'messages.loading' | translate }}</span>
--- a/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
@@ -1,9 +1,9 @@
-<div *ngIf="hasAnswerKey(item)">
+<div *ngIf="item.validAnswerKey">
   <strong>{{translateRoot + 'answer-key' | translate }}</strong>
-  {{item.answerKey}}
+  {{ item.validAnswerKey }}
 </div>
 
-<div *ngIf="!hasAnswerKey(item)">
+<div *ngIf="!item.validAnswerKey">
   <div *ngIf="loading" class="loader">
     <div class="inner">
       <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'messages.loading' | translate }}</span>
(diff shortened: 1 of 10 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
```diff
@@ -16,50 +16,46 @@
       <strong>{{ translateRoot + 'answer-key' | translate }}</strong> {{model.answerKeyValue}}
     </div>
 
-    <div *ngIf="model.rubrics.length > 0">
-      <div *ngIf="model.rubrics.length ==1" [innerHtml]="model.rubrics[0].templateHtml"></div>
-
-      <div *ngIf="model.rubrics.length > 1">
```
jtreuting: this whole section is only formatting really.  I removed the first check on `length > 0` because the checks within it are for `== 1` and `> 1` so no need for this first check

this is not related to the main change but i noticed while in the code

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.html
```diff
@@ -1,9 +1,9 @@
-<div *ngIf="hasAnswerKey(item)">
+<div *ngIf="item.validAnswerKey">
```
jtreuting: now that this check is in 2 places, i put the logic on the AssessmentItem instead of in the component.

webapp/src/main/webapp/src/app/assessments/model/assessment-item.model.ts
```diff
@@ -49,4 +49,12 @@ export class AssessmentItem {
       ? this.fullCredit / this.scores.length * 100
       : 0;
   }
+
+  /**
+   * If this item is MC or MS the answer key is returned, otherwise the data is not a valid answer key
+   * @returns {string}
+   */
+  get validAnswerKey(): string {
```
jtreuting: i went back and forth on whether to make this part of a static Utils class or a static method on AssessmentItem and opted to have it be a property of the AssessmentItem itself.  I could be convinced of something else I think if someone feels strongly.

but now this check is used on the Item Viewer and the Rubric/Exemplar components so needed it out of the component.

wtritch:
Yay!

webapp/src/main/webapp/src/app/assessments/model/assessment-item.model.ts
wtritch: I'd be tempted to move the logic into the mapper and store a flat boolean on the model rather than a getter function.

jtreuting: i like that.  i moved the logic to the mapper and realized that there is no need to have a second answer key value, we only need the answer key to be populated when we are going to use it which is when it is MC or MS, otherwise the answer key is junk and we are showing the rubric and exemplar instead.  so i am mapping using that logic to the answerKey property.  then it is not populated when it's not needed and won't be used.  simplifies things i think

webapp/src/main/webapp/src/app/shared/support/support.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { Utils } from "./support";
+
+describe('Utils', () => {
```
jtreuting: @esotrope and @wtritch added some tests to these checks and found that `isNullOrEmpty` throws an error if `undefined` is passed in.  Also I confirmed that `value == null` is the same as checking for `value === null || value === undefined`, so we could change the logic in that static method, but wanted to discuss offline.  Basically I'm wondering what the use case is for checking `isNullOrEmpty` and wanting `undefined` to return false.  Seems to me that if it is null, undefined or empty you would want to do the same thing in code.  Maybe a parking lot topic tomorrow.

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
```diff
@@ -29,34 +29,26 @@ export class ItemExemplarComponent implements OnInit {
...
-  }
-
-  /**
-   * Returns true if this item is a MC or MS and has an answer key
-   * @param {AssessmentItem} item
-   * @returns {boolean}
-   */
-  hasAnswerKey(item: AssessmentItem): boolean {
-    return (item.type === 'MC' || item.type === 'MC') && (item.answerKey !== undefined || item.answerKey !== null);
+    if (Utils.isNullOrUndefined(this.item.answerKey)) {
```
jtreuting: since we are only displaying the rubric and exemplar in the HTML if there is no answerKey, don't waste a call to the API if it will never be shown.

PR #812, opened 2018-02-01
adds form submission to aggregate report form

esotrope:
Part 2: implement result page polling for report resource and display conditionally based on size in resource metadata.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-confirmation.modal.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-confirmation.modal.ts
@@ -1,0 +1,36 @@
+import { Component, Input } from "@angular/core";
+import { BsModalRef } from "ngx-bootstrap";
+
+
+/**
+ * This modal component displays an instructional resource creation form.
+ */
+@Component({
+  selector: 'aggregate-report-confirmation-modal',
+  templateUrl: './aggregate-report-confirmation.modal.html'
+})
+export class AggregateReportConfirmationModal {
+
+  @Input()
+  rowCount: number;
+
+  @Input()
+  accept: Function;
+
+  @Input()
+  decline: Function;
+
+  constructor(private modal: BsModalRef){
+  }
+
+  onAcceptButtonClick(): void {
+    this.modal.hide();
+    this.accept && this.accept();
+  }
+
+  onDeclineButtonClick(): void {
+    this.modal.hide();
+    this.decline && this.decline();
+  }
+
+}
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
@@ -1,0 +1,39 @@
+import { Injectable } from "@angular/core";
+import { DataService } from "../shared/data/data.service";
+import { Observable } from "rxjs/Observable";
+import { AggregateServiceRoute } from "../shared/service-route";
+import 'rxjs/add/operator/timeout';
+import { ReportService } from "../report/report.service";
+import { Report } from "../report/report.model";
+
+/**
+ * Responsible for interfacing with aggregate report server
+ */
+@Injectable()
+export class AggregateReportService {
+
+  constructor(private dataService: DataService,
+              private reportService: ReportService) {
+  }
+
+  /**
+   * Gets the estimated report row count for the provided report request
+   *
+   * @param {AggregateReportRequest} request the report parameters
+   * @returns {Observable<number>} the row count
+   */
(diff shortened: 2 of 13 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-confirmation.modal.ts
```diff
@@ -0,0 +1,36 @@
...
+
+
+/**
+ * This modal component displays an instructional resource creation form.
+ */
+@Component({
+  selector: 'aggregate-report-confirmation-modal',
+  templateUrl: './aggregate-report-confirmation.modal.html'
+})
+export class AggregateReportConfirmationModal {
```
esotrope: writing this made me feel like we should have a generic confirmation modal component but at the same time it would probably be inflexible as there are so many ways to make a modal

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-reports-table.component.ts
```diff
@@ -10,6 +10,8 @@ import * as _ from "lodash";
 import { AssessmentDetailsService } from "./assessment-details.service";
 import { AggregateReportQuery } from "../model/aggregate-report-query.model";
 
+export const SupportedRowCount = 1000;
```
esotrope: What should our first arbitrary size threshold be?

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -78,6 +78,20 @@ export class ReportService {
...
+   *
+   * @param request the parameters to create the report with
+   * @returns {Observable<Report>} the handle used the get status on the download
+   */
+  public createAggregateReport(request: AggregateReportRequest): Observable<Report> {
+    return this.dataService.post(`${ServiceRoute}/aggregate`, request, {
+      headers: new Headers({ 'Content-Type': 'application/json' })
+    })
+    .map(this.toReport)
+    .catch(ResponseUtils.throwError);
```
esotrope: @wtritch this matches the controller you're working on right?

webapp/src/main/webapp/src/app/shared/service-route.ts
```diff
@@ -0,0 +1,5 @@
+export const AdminServiceRoute = '/admin-service';
+export const AggregateServiceRoute = '/aggregate-service';
+export const ReportingServiceRoute = '/reporting-service';
+export const ReportProcessorServiceRoute = '/report-processor';
```
esotrope: moved all service routes into common constants

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-reports-table.component.ts
wtritch: At least 10,000 since we support page sizes of up to 1000.  (And I tested it with 14,000.)

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-confirmation.modal.ts
wtritch: Unless you see 3 or more that could be abstracted out with the exact same properties/transclusion, I'd be tempted to leave it for now...

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    return this.dataService.post(`${AggregateServiceRoute}/aggregate/rowCount`, request);
+  }
+
+  /**
+   * Creates an aggregate report
+   *
+   * @param {AggregateReportRequest} request the report parameters
+   * @returns {Observable<Report>} the report resource handle
+   */
+  createReport(request: AggregateReportRequest): Observable<Report> {
```
wtritch: ...think we should let the user name their requests?  Since they'll be displayed in the "My Reports" page?  It's just one more field on the query builder... =)

webapp/src/main/webapp/src/app/report/report.service.ts
wtritch: yessir.  I imagine the resolver that requests the content is still coming?

esotrope: yup

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
esotrope: No : J

esotrope: j/k let's discuss the best way to do that in the parking lot

PR #813, opened 2018-02-01
prevents issues with left nav when navigating to another route

dsnyder-fairway:


webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -11,20 +11,24 @@ export class ScrollNavComponent {
...
+      let number = this.window.pageYOffset || this.document.documentElement.scrollTop || this.document.body.scrollTop || 0;
 
-    let activatedSection = this.activateSection(this.sections, number);
+      let activatedSection = this.activateSection(this.sections, number);
 
-    if (activatedSection) {
-      activatedSection.isActive = true;
+      if (activatedSection) {
+        activatedSection.isActive = true;
+      }
```
esotrope: unrelated but why doesn't the activateSection method do this?

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -11,20 +11,24 @@ export class ScrollNavComponent {
   @Input()
   sections: Section[] = [];
   private window: Window;
+  private path: string;
 
   constructor(@Inject(DOCUMENT) private document: Document,
               private windowRef: WindowRefService) {
     this.window = windowRef.nativeWindow;
+    this.path = this.window.location.href;
```
esotrope: This seems not very straight forward.
Can we manually listen for the window event and manually unsubscribe the old JS way since `@HostListener`s cant be unsubscribed?
We should be able to take advantage of `ngOnInit` and `ngOnDestroy` to bind and unbind the listener

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
dsnyder-fairway: I can move this into that method

dsnyder-fairway:
closing as writing trait UI has the priority at the moment

PR #815, opened 2018-02-01
Task/simplify aggregate reports translation code paths

esotrope:
`s/labels.aggregate-reports/aggregate-reports/g`

trying to move things out from the labels prefix as i go.
striving for the only prefixes being the specific `[component-selector]` or `common` when we want to share the text among all components


esotrope:
Should I take the resounding silence as approval?

PR #816, opened 2018-02-01
when expanding an item while viewing the writing trait score aggregat…

jtreuting:
…e view, change the columns to include the individual traits

@malaffoon and I were going through some use cases for Writing Trait Scores and realized there wasn't a clean way to see each individual students trait breakdown when viewing a group.  you had to go back and forth to view an individual student response which was less than ideal.  we thought when in the context of viewing the Writing Trait Scores if you expand the item and it shows the "Student Scores and Response" that the writing trait sub scores were more relevant than the max points and the correctness since each item is always the same max score those aren't super useful.

image


malaffoon:
Can we center- or right-justify the points?

jtreuting:
i'm not sure we consistently center the scores, but i'll look around to see.

PR #818, opened 2018-02-01
Move ReportRow model from aggregate-service to common to allow access…

wtritch:
… within the report-processor.

The report-processor needs to deserialize the aggregate-report JSON payload in order to conditionally provide the response as text/csv.


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ActiveAssessment.java
```diff
@@ -1,13 +1,14 @@
...
 import java.sql.SQLException;
 import java.util.Objects;
 
 /**
- * Represents an assessment data in the custom aggregate report row
+ * Represents an assessment's data with an associated exam year in which it
+ * was administered.
  */
-public class Assessment {
+public class ActiveAssessment {
```
wtritch: org.opentestsystem.rdw.reporting.common.model.Assessment already exists.  I renamed the aggregate-service "Assessment" to "ActiveAssessment" before moving into common to distinguish it from the existing model.

wtritch:
I'm committing this now to unblock others, will resolve any comments as they come up.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ActiveAssessment.java
```diff
@@ -1,13 +1,14 @@
...
+package org.opentestsystem.rdw.reporting.common.model;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Objects;
 
 /**
- * Represents an assessment data in the custom aggregate report row
+ * Represents an assessment's data with an associated exam year in which it
+ * was administered.
```
agorina: Nit: to be more precise this represents both: it is a UNION of assessments with the years it was administered at and a year it was designed for.

PR #821, opened 2018-02-02
DWR-1355 Modify custom aggregate report to include a parent organization for Schools

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/Organizations.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
+
+/**
+ * Holds common JDBC row mapping methods
+ */
+public final class Organizations {
```
agorina: There is a similar class in the Common, but it constructs Organization differently:
We do not need district_groups and school_groups in the custom aggregate report. 
Keeping them would mean pulling more data and doing more JOINs in the report and I did not think it is needed.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/Organizations.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public static Organization map(final ResultSet row) throws SQLException {
+        final OrganizationType type = OrganizationType.valueOf(row.getString("organization_type"));
+        switch (type) {
+            case School:
+                return School.builder()
+                        .id(row.getLong("organization_id"))
+                        .name(row.getString("organization_name"))
+                        .naturalId(row.getString("organization_natural_id"))
+                        .districtId(row.getLong("organization_parent_id"))
```
esotrope: should probably call the sql var district_id no?

agorina: Yes, changing.

PR #822, opened 2018-02-02
Feature/aggregate result loading

esotrope:
Sorry for the PR size - here's the result:

image

image

Result page now accepts a "id" url param.
This param is resolved to a report resource.
If the resource is too large we display a friendly error for the wannabe-hackers.
If the resource size is supported we load the report json if completed or poll until completed then load.


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -333,38 +321,74 @@ export class AggregateReportsComponent {
     }
   }
 
+  /**
+   * Creates the default/initial state of the aggregate report form based on the available options
+   *
+   * @param {AggregateReportFormOptions} options the options available for selection
+   * @returns {AggregateReportFormSettings} the initial form state
+   */
+  private createDefaultSettings(options: AggregateReportFormOptions): AggregateReportFormSettings {
```
esotrope: moved initialization of this model from the resolve to the component because i want to reuse the resolve for the results component and just need the API provided options objects - not the form-specific options

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -333,38 +321,74 @@ export class AggregateReportsComponent {
...
   /**
    * Creates an aggregate report request from a
    *
    * @param {AggregateReportFormSettings} settings the form state
    * @returns {AggregateReportRequest} the created request
    */
-  createReportRequest(settings: AggregateReportFormSettings): AggregateReportRequest {
-    const idsOf = values => values.map(entity => entity.id);
+  private createReportRequest(settings: AggregateReportFormSettings): AggregateReportRequest {
+    const codesOf = values => values.map(entity => entity.code);
```
esotrope: @wtritch went ahead and made the change to codes on the UI side as i needed to to fix other things in the PR

webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment-definition.resolve.ts
```diff
@@ -0,0 +1,20 @@
...
+import { AssessmentDefinitionService } from "./assessment-definition.service";
+import { Injectable } from "@angular/core";
+import { AssessmentDefinition } from "./assessment-definition";
+import { Observable } from "rxjs/Observable";
+
+/**
+ * Resolves assessment type definitions and properties
+ */
+@Injectable()
+export class AssessmentDefinitionResolve implements Resolve<Map<string, AssessmentDefinition>> {
```
esotrope: tweaked the existing assessment details service to provide all assessment type defs. Also added this resolve so everything under the /aggregate-reports context can access this config data

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.mapper.ts
```diff
@@ -0,0 +1,68 @@
...
+import { AggregateReportRow } from "../../report/aggregate-report";
+import { AssessmentDefinition } from "../assessment/assessment-definition";
+
+/**
+ * Maps server modeled aggregate report rows into client friendly table rows
+ */
+@Injectable()
+export class AggregateReportItemMapper {
+
+  map(assessmentDefinition: AssessmentDefinition, row: AggregateReportRow, uuid: number): AggregateReportItem {
```
esotrope: logic pulled directly from mock service. only mod was to pass along grade and subject codes instead of ids

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -250,3 +246,11 @@ export class AggregateReportsTableComponent implements OnInit {
     });
   }
 }
+
+
+export interface AggregateReportTable {
```
esotrope: big change here is that now the table accepts this model instead of just rows and the row items no longer need to contain the info about the assessment def.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * This component is responsible for performing the aggregate report query and
+ * displaying the results.  Results are displayed in one table per subjectId.
+ */
+@Component({
+  selector: 'aggregate-report',
+  templateUrl: './aggregate-report.component.html',
+})
+export class AggregateReportComponent implements OnInit {
```
esotrope: the new aggregate report viewer page component

webapp/src/main/webapp/src/app/aggregate-report/results/mock-aggregate-reports-preview.service.ts
```diff
@@ -1,37 +1,30 @@
...
+
 /**
  * This service is used by the aggregate reports preview
  */
 @Injectable()
 export class MockAggregateReportsPreviewService {
 
-  constructor(private http: HttpClient,
-              private assessmentDetailsService: AssessmentDetailsService) {
+  constructor(private assessmentDefinitionService: AssessmentDefinitionService) {
```
esotrope: udpated these mock services and tests to use the new assessment def service

webapp/src/main/webapp/src/app/aggregate-report/results/performance-comparison.component.html
```diff
@@ -1,14 +1,14 @@
...
-         [ngClass]="getBelowColor(levelIdx)"
+         [ngClass]="getBelowColor(i + 1)"
          [attr.data-width]="percentage"></div>
   </div>
   <div *ngIf="hasRollup()" class="bar-container" [attr.data-width]="50">
-    <div *ngFor="let percentage of getAbovePercentages(); let levelIdx = index"
+    <div *ngFor="let percentage of getAbovePercentages(); let i = index"
          class="bar"
-         [ngClass]="getAboveColor(levelIdx)"
+         [ngClass]="getAboveColor(i + 1)"
```
esotrope: changed these to pass along the performance level (1 based) - not index (0 based) - as the color service method i introduced expects the level and not an index

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -70,4 +76,16 @@ export class AppComponent {
...
+    this.router.events.subscribe(event => {
+      if (event instanceof NavigationStart) {
+        this.spinnerModal.loading = true;
+      } else if (event instanceof NavigationEnd
+        || event instanceof NavigationCancel
+        || event instanceof NavigationError) {
+        this.spinnerModal.loading = false;
+      }
+    })
+  }
```
esotrope: the spinner html markup proliferation is out of hand so i made a common "loading" module to now contain a spinner modal and spinner sub component. the app now defines when the spinner modal should display so the modal component itself can remain flexible.

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -18,8 +18,8 @@
         <sb-toggle name="off-grade-assessment-toggle"
                    [(ngModel)]="filterBy.offGradeAssessment"
                    [options]="[
-                           {value: false, text: 'buttons.visibility.show' | translate, angularticsProperties: {label: 'Off-Grade: Show'}},
-                           {value: true, text: 'buttons.visibility.hide' | translate, angularticsProperties: {label: 'Off-Grade: Hide'}}
+                           {value: false, text: 'buttons.visibility.show' | translate, analyticsProperties: {label: 'Off-Grade: Show'}},
+                           {value: true, text: 'buttons.visibility.hide' | translate, analyticsProperties: {label: 'Off-Grade: Hide'}}
```
esotrope: unrelated fixes to how we config sb-toggle options

webapp/src/main/webapp/src/app/report/report.model.ts
```diff
@@ -17,6 +17,7 @@ export class Report {
   public subjectId: number;
   public schoolYear: number;
   public created: Date;
+  public request: any;
```
esotrope: @wtritch added request field to report model

webapp/src/main/webapp/src/app/shared/color.service.ts
```diff
@@ -43,6 +54,11 @@ export class ColorService {
    * @returns {string} the class of the color
    */
   getPerformanceLevelColor(assessment: Assessment, performanceLevel: number): string {
-    return this.performanceLevelColors.get(assessment.type)[performanceLevel];
+    return PerformanceLevelColorsByAssessmentType.get(assessment.type)[ performanceLevel ];
+  }
+
+  getPerformanceLevelColorsByAssessmentTypeCode(code: string, performanceLevel: number): string {
+    return PerformanceLevelColorsByAssessmentTypeCode.get(code)[ performanceLevel - 1 ];
   }
```
esotrope: cleaned up this class and added a method that accepts the assessment type code instead of enum.
These colors are another thing that should be provided by a ui config pipeline along side the comparators.
I also still think there should be a level of indirection here between the actual color names and the code.
e.g.
```less
.wheel-color-1 {
  red;
}
.wheel-color-2 {
  aqua;
}
.colored.assessment-type-iab.performance-level-1 {
  .green;
}
```

webapp/src/main/webapp/src/app/shared/loading/spinner.modal.html
```diff
@@ -0,0 +1,5 @@
+<div class="delayed-fade" [class.delayed-fade-in]="loading">
+  <div class="loader-modal">
+    <spinner text=""></spinner>
```
esotrope: the new way to add a loading spinner : j

webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment-definition.service.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * TODO make this hit backend and cache results.
+   * TODO expand to consider subject type possibly.
+   *
+   * Gets all assessment type related data.
+   *
+   * @returns {Observable<Map<string, AssessmentDefinition>>}
+   */
+  public getDefinitionsByAssessmentTypeCode(): Observable<Map<string, AssessmentDefinition>> {
```
wtritch: Were we going to extend this to accept both assessment type and subject?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.mapper.ts
```diff
@@ -0,0 +1,68 @@
...
+    const item = new AggregateReportItem();
+    item.itemId = uuid;
+    item.assessmentId = row.assessment.id;
+    item.gradeId = row.assessment.gradeId;
+    item.gradeCode = row.assessment.gradeCode;
```
wtritch: hmm... do we need both the gradeId and gradeCode?  Or can we just have gradeCode?  (I'll have to update the ActiveAssessment since right now it only has "gradeId")

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.ts
```diff
@@ -4,8 +4,16 @@
 export class AggregateReportItem {
   itemId: number;
   assessmentId: number;
+  /**
+   * @deprecated
+   */
```
wtritch: ah, so I can remove gradeId.  Ignore comment above =)

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,102 +1,101 @@
...
               [sortable]="previewOnly ? false : 'custom'"
               (sortFunction)="sort($event)"
               header="{{'aggregate-reports.results.cols.assessment-grade' | translate}}">
-      <ng-template let-col let-reportItem="rowData" let-i="rowIndex" pTemplate="body">
+      <ng-template let-col let-row="rowData" let-i="rowIndex" pTemplate="body">
         <div *ngIf="treeColumns[i] <= colIndex(col)"
              class="full-cell pt-0"
              [ngClass]="{'border-top': treeColumns[i] <= colIndex(col) && treeColumns[i] < columnOrdering.length - 1}">
-          {{reportItem.gradeId}}
+          {{row.gradeId}}
```
wtritch: This'll have to be changed to push the gradeCode through translation

webapp/src/main/webapp/src/app/aggregate-report/results/query-builder.component.html
```diff
@@ -486,13 +488,13 @@ <h2 class="h3 gray-darkest">
...
-                                 [previewOnly]="true"
-                                 [groupPerformanceLevels]="groupPerformanceLevels"
-        ></aggregate-reports-table>
+        <!--<aggregate-report-table *ngIf="responsePreview" [reportItems]="responsePreview"-->
+                                 <!--[query]="aggregateReportQuery"-->
+                                 <!--[showValuesAsPercent]="!aggregateReportQuery.showValueAs"-->
+                                 <!--[columnOrdering]="['organizationName', 'gradeId', 'schoolYear', 'dimensionValue']"-->
+                                 <!--[previewOnly]="true"-->
+                                 <!--[groupPerformanceLevels]="groupPerformanceLevels"-->
+        <!--&gt;</aggregate-report-table>-->
```
wtritch: ??

webapp/src/main/webapp/src/app/report/report.model.ts
wtritch: Sweet, now I just need to stuff something in the response =)

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -108,6 +119,36 @@ export class ReportService {
     }).catch(ResponseUtils.throwError);
```
wtritch: We should change the accept above this to "*/*" or even remove it altogether.  The download response may be text/csv for an exam export, application/pdf for a student/group report, or application/zip for a large pdf report.

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -108,6 +119,36 @@ export class ReportService {
...
+   * @returns {Observable<Download>}
+   */
+  public getAggregateReportAsSpreadsheet(reportId: number): Observable<Download> {
+    return this.dataService.get(`${ServiceRoute}/reports/${reportId}`, {
+      headers: new Headers({
+        'Accept': 'text/csv',
+      }),
+      responseType: ResponseContentType.Blob
+    }).catch(ResponseUtils.throwError);
+  }
```
wtritch: We shouldn't need this.  Getting the data as a spreadsheet will be returned if the Accept header is anything but "application/json" (even empty/blank).

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
esotrope: thanks for pointing at it - i'll patch it

webapp/src/main/webapp/src/app/aggregate-report/results/query-builder.component.html
esotrope: Quick fix to get everything we use working - May be time to drop the query builder page since i think we have what we need from it.

PR #824, opened 2018-02-02
Task/convert aggregate report query to codes

wtritch:
This PR changes our ReportQuery to contain codes rather than DB ids.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateExportService.java
```diff
@@ -77,130 +50,132 @@
     @Override
     public ResponseEntity<ByteArrayResource> exportReport(final User user, final ReportQuery query) {
```
wtritch: This will be going away in a future update.  I didn't want to spend the effort converting it.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/ReportQueryRepositoryHelper.java
```diff
@@ -165,28 +182,67 @@ public ReportQueryRepositoryHelper(final OrganizationRepository organizationRepo
...
+        return toIds(ethnicityCodes, ethnicityRepository);
+    }
+
+    public Set<Integer> toStrictBooleanIds(final Collection<String> strictBooleanCodes) {
+        return toIds(strictBooleanCodes, strictBooleanRepository);
+    }
+
+    public Set<Integer> toBooleanIds(final Collection<String> booleanCodes) {
+        return toIds(booleanCodes, booleanRepository);
+    }
```
wtritch: I added specific mappings here with a generic implementation below that accepts a repository reference.

PR #825, opened 2018-02-02
Return the original report request with the report resource response …

wtritch:
…payload


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportResource.java
```diff
@@ -89,6 +91,14 @@ void setCreated(final Instant created) {
         this.created = created;
     }
 
+    public AbstractExamReportRequest<?> getRequest() {
+        return request;
+    }
+
+    void setRequest(final AbstractExamReportRequest<?> request) {
+        this.request = request;
+    }
```
wtritch: Quick change to add the original request into the ReportResource used by the front-end as a handle for displaying/requesting reports.

PR #828, opened 2018-02-02
Sort aggregate report table rows hierarchically by state, district, s…

wtritch:
…chool

This updates our aggregate report result viewer to sort the organization column by:
```
State
District X
School X_A
School X_B
District M
School M_A
School M_B
```

Notice I used X and M because while State will always be first and Schools will be alpha-sorted by name under their districts, (District+School) groups will be sorted by district id.
This is a limitation of Schools not having access to their district names.  To correct the issue, we'd need to scan the (potentially thousands of) result rows to build a districtId->districtName map to query against during sorting.  If we expect to have human-readable multi-district reports that the user expects to read in alphabetical district order I can put it in.

Thoughts?


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -17,6 +18,29 @@ const OrderingDimensionType: Ordering<AggregateReportItem> = ordering(
 const OrderingDimensionValue: Ordering<AggregateReportItem> = ordering(byString)
   .on((item) => item.dimensionValue.toString());
 
+const OrderingOrganizationState: Ordering<AggregateReportItem> = ordering(
+  ranking([OrganizationType.State])
+).on((item) => item.organization.type);
```
wtritch: This pulls State rows to the top of the list.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -17,6 +18,29 @@ const OrderingDimensionType: Ordering<AggregateReportItem> = ordering(
...
+  .on((item) => {
+    switch(item.organization.type) {
+      case OrganizationType.District:
+        return (item.organization as District).id;
+      case OrganizationType.School:
+        return (item.organization as School).districtId;
+      default:
+        return -1;
+    }
+});
```
wtritch: This bundles school rows with their district rows.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -17,6 +18,29 @@ const OrderingDimensionType: Ordering<AggregateReportItem> = ordering(
...
+
+const OrderingOrganizationDistrict: Ordering<AggregateReportItem> = ordering(
+  ranking([OrganizationType.District])
+).on((item) => item.organization.type);
```
wtritch: This pulls district rows to the top of their (district+school) row groups.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -17,6 +18,29 @@ const OrderingDimensionType: Ordering<AggregateReportItem> = ordering(
...
+
+const OrderingOrganizationSchools: Ordering<AggregateReportItem> = ordering(byString)
+  .on((item) => item.organization.name);
```
wtritch: This alpha-sorts schools by name within their district-scoped sub-groups.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -73,7 +97,12 @@ export class AggregateReportTableComponent implements OnInit {
   private previousSort: any;
 
   constructor(public colorService: ColorService) {
-    this.orderingByProperty.organization = ordering(byString).on((item) => item.organizationName);
+    this.orderingByProperty.organization = ordering(join.apply(null, [
+      OrderingOrganizationState.compare,
+      OrderingOrganizationDistrictsWithSchools.compare,
+      OrderingOrganizationDistrict.compare,
+      OrderingOrganizationSchools.compare
+    ]));
```
wtritch: This puts together the organization comparators such that they fall through to each successive comparator when the previous comparator returns an "equal" value.

webapp/src/main/webapp/src/app/aggregate-report/results/mock-aggregate-reports-preview.service.ts
```diff
@@ -80,8 +82,7 @@ export class MockAggregateReportsPreviewService {
     uiModel.assessmentId = apiModel.assessment.id;
     uiModel.gradeId = apiModel.assessment.gradeId;
     uiModel.schoolYear = apiModel.examSchoolYear;
-    uiModel.organizationType = apiModel.organization.type;
-    uiModel.organizationName = apiModel.organization.name;
+    uiModel.organization = this.organizationMapper.map(apiModel.organization);
```
wtritch: I removed the flattened organization properties on the report items in favor of an Organization property so I could leverage the existing common OrganizationMapper.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports-preview-table.component.ts
```diff
@@ -73,7 +73,7 @@ export class AggregateReportsPreviewTableComponent implements OnInit {
 
   constructor(public colorService: ColorService,
               private assessmentDefinitionService: AssessmentDefinitionService) {
-    this.orderingByProperty.organizationName = ordering(byString).on((item) => item.organizationName);
+    this.orderingByProperty.organizationName = ordering(byString).on((item) => item.organization.name);
```
wtritch: It concerns me that we have an aggregate-reports-preview-table.component.ts that's nearly a carbon-copy of the aggregate-report-table.component.ts ... This is going to lead to inconsistency errors...

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
esotrope:
```javascript
fn.apply(this, [a, b, c]) === this.fn(a, b, c) 
```

wtritch: um... good point.  Not sure why I did that.

webapp/src/main/webapp/src/app/aggregate-report/results/mock-aggregate-reports-preview.service.ts
esotrope: yay reuse

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
esotrope: you just wanted to do it the fun way i guess : j

webapp/src/main/webapp/src/app/aggregate-report/aggregate-reports-preview-table.component.ts
esotrope: After updating both in the last PR i share the same sentiment

jtreuting:
What happens when there are just districts and no schools?

wtritch:
@jtreuting The state will be first, and the districts will be sorted by DB id =/

esotrope:
if we wanted to go by district name the backend will have to resolve the district name for each school. like custom org export we cant do it on the front end because the user may not be entitled to the district the school references.

wtritch:
I don't think we'd need to resolve it on the backend because if the district row isn't visible, it doesn't matter what the order we show the grouped-by-district schools in.

esotrope:
@wtritch you're right - i forgot this is in the results view when the rows contain the districts and stuff already 👍

wtritch:
@jtreuting I added sorting (district+school) groups by district name.  I left in sorting (district+school) groups by district id to handle the unlikely case where two districts have the same name.

PR #829, opened 2018-02-03
fixes aggregate report route navigation

esotrope:


webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -285,6 +304,7 @@ export const routes: Routes = [
           {
             path: '',
             pathMatch: 'full',
+            data: { canReuse: true },
```
esotrope: basically breadcrumb linking back and forth from aggregate report form to results broke along the way and is fixed now by moving this can reuse config from the parent component to the child "index" route.
Rest of the changes are just formatting for readability

PR #830, opened 2018-02-03
DWR-1341 Modify estimated count to be executed in Java instead of Redshift

agorina:
This is the quickest (but not the cleanest) way to get it working.

There are a few things I do not like with this approach:
1. it does more computation for the count than is needed;
2. if I start nit-picking, then I can find several more things.

What I **like** is that it takes the same execution path to build the count as it does when building the SQL.

To not hold others, I would like to submit it and try to refactor. I admit that I had made a few quick attempts and it was not easy. I will timebox myself.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -43,15 +37,8 @@ public JdbcCustomAggregateReportRowRepository(final NamedParameterJdbcTemplate j
...
     }
 
-    private static void validate(final ReportQuery query) {
-        checkArgument(!IAB.code().equals(query.getAssessmentTypeCode()), "custom aggregate report for IAB is not supported yet");
-    }
-
     @Override
     public List<ReportRow> findByQuery(@NotNull final ReportQuery query) {
-
-        validate(query);
```
agorina: Moved the validation to one place when the query is built.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
```diff
@@ -38,15 +37,13 @@
...
             @AuthenticationPrincipal final User user,
             @RequestBody final ReportQuery query) {
-
         return service.findByQuery(user, query);
     }
 
-    @GetMapping("/estimateReportRowCount")
-    public int getEstimateReportRowCount(
+    @PostMapping("/estimateReportRowCount")
+    public int postEstimateReportRowCount(
```
esotrope: can we change this to `/aggregate/rowCount`

Also, what about the row count makes it an estimate?
Is this the rough estimate formula?
(stateCount + districtCount + schoolCount) * gradeCount * yearCount * dimensionCount
i guess dimension count gets a little funky with filters though

agorina: `StudentEnrolledGrade` dimension makes it an estimate. 

With everything else (since we backfill the data), it is an exact count. And it does account for filters. But with the enrolled grade we assume that there is one per each asmt/schoolYear and this is not exactly right.

esotrope: gotcha good to know. lets compromise and do `/aggregate/estimatedRowCount`

agorina: I honestly have no strong preference. Will change as you are suggesting unless somebody has a different opinion.

PR #832, opened 2018-02-05
writing trait scores on IAB pdf generation

dsnyder-fairway:


common/src/main/resources/messages.properties
```diff
@@ -413,4 +413,88 @@ report.aggregate.export.header.dimension=Subgroup
...
-report.aggregate.export.header.attributes=Attributes
\ No newline at end of file
+report.aggregate.export.header.attributes=Attributes
+
+### Writing Trait Scores
+report.writing.trait.purpose=Writing Purpose: {0}
+report.writing.trait.Organization/Purpose=Organization / Purpose
+report.writing.trait.Evidence/Elaboration=Evidence / Elaboration
+report.writing.trait.Conventions=Conventions
+report.writing.trait.Narrative.Organization/Purpose.0=TODO: they scored a 0
```
dsnyder-fairway: I guess this is slightly presumptuous as I am using Narrative/Argumentative/Explanatory/Informational/Opinion for the strings I assumed would be in item.performance_task_writing_type, however I do not really know if these are correct

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/IabReportItemRepository.java
```diff
@@ -0,0 +1,14 @@
+package org.opentestsystem.rdw.reporting.processor.repository;
+
+import org.opentestsystem.rdw.reporting.common.model.ExamItem;
+import org.opentestsystem.rdw.reporting.common.model.PerformanceWritingExamItem;
+import org.springframework.stereotype.Repository;
+
+import java.util.Collection;
+import java.util.List;
+
+public interface IabReportItemRepository {
```
wtritch: nit, recommend javadocs on interfaces (services/repositories) at both the class level and on every method.

common/src/main/resources/messages.properties
dsnyder-fairway: @jtreuting, are those the permitted values of item.performance_task_writing_type?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/IabReportItemRepository.java
```diff
@@ -0,0 +1,14 @@
...
+
+    List<PerformanceWritingExamItem> findWritingTraitExamItems(Collection<Long> mostRecentExamIds);
```
wtritch: nit: recommend renaming parameter to just examIds (I'm guessing it's the caller that knows they're the "most recent" but this service and below doesn't know or care?)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
```diff
@@ -179,6 +190,39 @@ public void generateStudentChunkReport(final Message<ProcessStudentChunkMessage>
...
+        List<PerformanceWritingExamItem> examItems = iabReportItemRepository.findWritingTraitExamItems(examIds);
+        if (examItems.size() > 0) {
+            compositeReports.forEach((key, report) -> {
+                report.getIabEla().getExamsByAssessment().forEach((assessment, exams) -> {
+                    // First exam is most recent
+                    Optional<Exam> exam = exams.stream().findFirst();
+                    if (exam.isPresent()) {
+                        Optional<PerformanceWritingExamItem> examItem = examItems.stream().filter(examItem1 -> examItem1.getExamItem().getExamId() == exam.get().getId()).findFirst();
+                        if (examItem.isPresent()) {
+                            exam.get().setWritingTraitScores(examItem.get().getExamItem().getWritingTraitScores());
```
dsnyder-fairway: I know this is probably frowned upon as we are using builders for much of our models, however I need to append this information to the exam. Not really sure how I should do this

common/src/main/resources/messages.properties
jtreuting: The values I see from the real content are:
- Argumentative
- Explanatory
- Narrative
- Opinion
- Informational

Also for the name of the second part since we are using a value we control for the 3 sub-score identifiers I would remove the second part and just do `report.writing.trait.Narrative.Organization.0=....`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
esotrope: Usually we put a `Builder copy()` instance level method on the object - which is `Exam` in this case. Then you can modify the newly created builder and then call build to make a new object so the original is never modified

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Exam.java
```diff
@@ -25,10 +25,27 @@
...
+    private String performanceWritingType;
 
     private Exam() {
     }
 
+    public WritingTraitScores getWritingTraitScores() {
+        return writingTraitScores;
+    }
+
+    public void setWritingTraitScores(final WritingTraitScores writingTraitScores) {
```
malaffoon: In general we do builders and don't have setters; what is the need for it here?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/PerformanceWritingExamItem.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public Builder performanceWritingType(final String performanceWritingType) {
+            this.performanceWritingType = performanceWritingType;
+            return this;
+        }
+
+        public PerformanceWritingExamItem build() {
+            final PerformanceWritingExamItem examWithWritingTraitScore = new PerformanceWritingExamItem();
+            examWithWritingTraitScore.examItem = this.examItem;
```
malaffoon: I don't have a strong preference for using `this.` or not when it isn't needed but i think we should be consistent.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
```diff
@@ -179,6 +190,39 @@ public void generateStudentChunkReport(final Message<ProcessStudentChunkMessage>
...
+                });
+            }
+        });
+
+        List<PerformanceWritingExamItem> examItems = iabReportItemRepository.findWritingTraitExamItems(examIds);
+        if (examItems.size() > 0) {
+            compositeReports.forEach((key, report) -> {
+                report.getIabEla().getExamsByAssessment().forEach((assessment, exams) -> {
+                    // First exam is most recent
+                    Optional<Exam> exam = exams.stream().findFirst();
```
esotrope: can be `final`

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Exam.java
dsnyder-fairway: see https://github.com/SmarterApp/RDW_Reporting/pull/832#discussion_r166074688. I'll go ahead and use a copy as @esotrope suggested

report-processor/src/main/resources/application.sql.yml
```diff
@@ -96,6 +96,26 @@ sql:
...
+          ei.trait_organization_purpose_score,
+          ei.trait_conventions_score,
+          i.performance_task_writing_type
+        from exam_item ei
+        join item i on i.id=ei.item_id
+        join exam e on e.id=ei.exam_id
+        where
+          ei.exam_id in (:exam_ids)
+          and i.type=:type
+
```
esotrope: 👍

esotrope:
@dsnyder-fairway can you attach a screenshot? Would also like to review the look and feel

common/src/main/resources/messages.properties
```diff
@@ -413,4 +413,88 @@ report.aggregate.export.header.dimension=Subgroup
...
+report.writing.trait.Unknown.Organization/Purpose.3=Unknown.Organization/Purpose.3
+report.writing.trait.Unknown.Organization/Purpose.4=Unknown.Organization/Purpose.4
+report.writing.trait.Unknown.Evidence/Elaboration.0=TODO: they scored a 0
+report.writing.trait.Unknown.Evidence/Elaboration.1=Unknown.Evidence/Elaboration.1
+report.writing.trait.Unknown.Evidence/Elaboration.2=Unknown.Evidence/Elaboration.2
+report.writing.trait.Unknown.Evidence/Elaboration.3=Unknown.Evidence/Elaboration.3
+report.writing.trait.Unknown.Evidence/Elaboration.4=Unknown.Evidence/Elaboration.4
+report.writing.trait.Unknown.Conventions.0=Unknown.Conventions.0
+report.writing.trait.Unknown.Conventions.1=Unknown.Conventions.1
+report.writing.trait.Unknown.Conventions.2=Unknown.Conventions.2
```
esotrope: how many of these messages will be redundant i wonder?
Also, will some of the keys really have a forward slash? As long as they don't have spaces or equals i guess we are fine.

common/src/main/resources/messages.properties
```diff
@@ -413,4 +413,88 @@ report.aggregate.export.header.dimension=Subgroup
 report.aggregate.export.header.student-count=Students Tested
 report.aggregate.export.header.score=Average Scale Score
 report.aggregate.export.header.standard-deviation=Average Scale Score Standard Deviation
-report.aggregate.export.header.attributes=Attributes
\ No newline at end of file
+report.aggregate.export.header.attributes=Attributes
+
+### Writing Trait Scores
+report.writing.trait.purpose=Writing Purpose: {0}
+report.writing.trait.Organization/Purpose=Organization / Purpose
```
malaffoon: Interesting, is having capitals and a `/` in the property name acceptable? Desirable? I probably would've used `_` or even left off the second term, e.g. `reporting.writing.trait.organization`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportItemRepository.java
```diff
@@ -0,0 +1,58 @@
...
+    @Autowired
+    JdbcIabReportItemRepository(final NamedParameterJdbcTemplate template,
+                                final SecurityParameterProvider securityParameterProvider) {
+        this.template = template;
+        this.securityParameterProvider = securityParameterProvider;
+    }
+
+    @Override
+    public List<PerformanceWritingExamItem> findWritingTraitExamItems(Collection<Long> mostRecentExamIds) {
+        final List<PerformanceWritingExamItem> examItems = new ArrayList<>();
```
malaffoon: You shouldn't have to create your own list, there is a template method that returns a list give a query and RowMapper.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportItemRepository.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Autowired
+    JdbcIabReportItemRepository(final NamedParameterJdbcTemplate template,
+                                final SecurityParameterProvider securityParameterProvider) {
+        this.template = template;
+        this.securityParameterProvider = securityParameterProvider;
+    }
+
+    @Override
+    public List<PerformanceWritingExamItem> findWritingTraitExamItems(Collection<Long> mostRecentExamIds) {
```
malaffoon: nit: param can be final

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
```diff
@@ -64,8 +70,10 @@
 @Component
 public class StudentReportProcessor {
     private final static Logger logger = LoggerFactory.getLogger(StudentReportProcessor.class);
+    private static Collection writingAssessment = newArrayList("WER");
```
malaffoon: You should type the collection, i.e. `Collection<String>`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: That may be difficult since I think we use builders all the way up the chain and the structure we have is:
```
Map<StudentId, CompositeReport> compositeReports
  IabReport report = compositeReport#getIabEla()
    Map<Assessment, Collection<Exam>> = report#getExamsByAssessment()
```
So we would have to copy all the way up the chain in order to modify the bottom-level Collection<Exam>.  I think it's the right thing to do, but it's going to be more involved than initially obvious.  It also means that we'll need to modify a collection we're iterating over which is a no-no.  I'd be tempted to invert this and loop over the `examItems` and update compositeReports in the loop rather than loop over compositeReports.
This means you may need to modify PerformanceWritingExamItem to include the student id...

malaffoon: Ah, my answer from earlier. It looks like it might be ugly to do it here. I wonder if it could've been done earlier in the life-cycle of the report; then you could use copy-set-build semantics.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
```diff
@@ -179,6 +190,39 @@ public void generateStudentChunkReport(final Message<ProcessStudentChunkMessage>
...
+                    Optional<Exam> exam = exams.stream().findFirst();
+                    if (exam.isPresent()) {
+                        examIds.add(exam.get().getId());
+                    }
+                });
+            }
+        });
+
+        List<PerformanceWritingExamItem> examItems = iabReportItemRepository.findWritingTraitExamItems(examIds);
+        if (examItems.size() > 0) {
```
malaffoon: nit: should use `isEmpty` and may want to flip to guard-style logic to avoid indentation, i.e. 
```
    if (examItems.isEmpty()) return;
    compositeReports.forEach ...
```

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: NIt: From my comment above, if we're relying on ordering in any way (the first exam is the "latest") then report#getExamsByAssessment() should return `Map<Assessment, List<Exam>>` rather than `Map<Assessment, Collection<Exam>>`

dsnyder-fairway:
@esotrope @malaffoon @jtreuting @wtritch 
results from running this

## item.performance_task_writing_type = null
screen shot 2018-02-05 at 10 43 51 am

## item.performance_task_writing_type = Opinion
screen shot 2018-02-05 at 11 13 20 am

common/src/main/resources/messages.properties
dsnyder-fairway: mostly these were chosen as I was unsure of what the proper values for item.performance_task_writing_type would be (assumed something like `Organization/Purpose`. I will be removing everything after the `/` but keeping the capitalizaiton

esotrope:
@dsnyder-fairway thanks for the shots. I think we might want a border-top on that section. also i think we need the max possible points next to the actual scored points like maybe `[ 2 ] /4` where `/4` is kind of small and to the top right or something.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
wtritch: ...This makes me wonder if we instead want to modify our existing report exam fetching SQL to left-join on writing traits, rather than make a second call to saturate writing trait scores...

common/src/main/resources/messages.properties
dsnyder-fairway: oops...scratch that I misspoke.
We can keep the `/` in there. Not sure if that is desirable though

malaffoon:
I agree with Landon's commenting about including the max for each. I would also suggest that the size of the subscores be made smaller, the way it is they draw the eye away from the important score (overall).

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/StudentReportProcessor.java
dsnyder-fairway: I'll be removing this line as I am not using this

common/src/main/resources/messages.properties
malaffoon: Personally, i don't like the `/`. For example, it probably isn't going to happen but it raises the issue of how to set it in an environment variable? `EXPORT REPORT_WRITING_TRAIT_ORGANIZATION/PURPOSE=...`. Might work but confusing.

I don't care about the uppercasing a lot, and i think the system resolves it whether it matches or not.

dsnyder-fairway:
@esotrope @malaffoon how about immediately below, centered, not in any box/button we have bold text saying:
`Maximum 4 points` or `Maximum Score: 4` or `out of 4`?

dsnyder-fairway:
accidentally `Close and comment` 😆

malaffoon:
I always wonder who thought putting "Close and comment" right beside "Comment" was a good idea.

jtreuting:
@malaffoon for context, Aaron thought the score could be like the icon we have on the ICA report.

image

aaronsleeper:
**2** of 4
or
2
out of 4

either of those work for me

dsnyder-fairway:
@esotrope @wtritch here is the updated scoring section on the pdf.
I also added the means to have multiple Writing traits for an exam. The separate score breakdowns for each of those would appear immediately after the previous score as a new row (including a new Writing Purpose: blah)
screen shot 2018-02-09 at 12 33 09 pm

dsnyder-fairway:
I should have closed this an re-opened when I was done, but oh well. I think it is good. Please review @esotrope, @wtritch, @malaffoon

esotrope:
Can we get a screenshot of when there are multiple?
Only nit is the header "Writing Purpose: ..." doesnt stand out very much from the headers below it - it is also a little close to the left border.
Should we have a border-top too? or do we want the sections to look joined?

report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -103,6 +107,46 @@
...
+                             th:if="${exam.sessionId == entry.value[0].sessionId && exam.writingTraitScores != null}">
+                            <h2 th:if="${scored}"
+                                th:text="#{('report.writing.trait.purpose')(${exam.performanceWritingType})}"></h2>
+
+                            <div class="col-md-4 pull-left" data-width="33">
+                                <h3 class="text-center" th:text="#{'report.writing.trait.Evidence'}"></h3>
+                                <h2 class="label-group gray-light bordered text-center"><span class="label"
+                                                                                              th:text="${exam.writingTraitScores.evidence}"></span>
+                                </h2>
+                                <h4 class="text-center" th:text="#{('report.writing.trait.outof')(4)}"></h4>
```
esotrope: the max scores need to be data driven right?

dsnyder-fairway:
screen shot with multiple entries
screen shot 2018-02-09 at 1 32 03 pm

aaronsleeper:
I don't understand what case this is illustrating. Can someone help?

It looks like multiple rows of almost identical data. If this is more than
an edge-case we should consider a separate layout.

dsnyder-fairway:
@aaronsleeper that is to illustrate the very rare case of having multiple writing purposes for a given exam. In practice it looks like this doesn't happen, but in theory it can.

aaronsleeper:
ah ok . if it's not likely to happen then we don't need to spend the hrs on
a custom design / build

this works.

report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -103,6 +107,46 @@
...
+                                <br/>
+                                <p th:text="#{'report.writing.trait.' + (${exam.performanceWritingType != null} ? ${exam.performanceWritingType} : 'Unknown') + '.Evidence.' + ${exam.writingTraitScores.evidence}}"></p>
+                            </div>
+                            <div class="col-md-4 pull-left" data-width="33">
+                                <h3 class="text-center" th:text="#{'report.writing.trait.Organization'}"></h3>
+                                <h2 class="label-group gray-light bordered text-center"><span class="label"
+                                                                                              th:text="${exam.writingTraitScores.organization}"></span>
+                                </h2>
+                                <h4 class="text-center" th:text="#{('report.writing.trait.outof')(4)}"></h4>
+                                <br/>
```
esotrope: `<br>` shouldn't be used for styles -> use css class `mt-sm` on the element below this

esotrope: If the br is to make it not on the same line i suggest wrapping the `<h4>` in a `<div>`

report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -103,6 +107,46 @@
                         </ul>
                     </div>
                 </div>
+
+                <!-- Writing trait scores-->
+                <div class="col-lg-12" th:if="${entry.value[0].writingTraitScores}"
+                     th:with="scored=true,assessment=${entry.key}">
+                    <div th:each="exam : ${entry.value}">
+                        <div class="row"
+                             th:if="${exam.sessionId == entry.value[0].sessionId && exam.writingTraitScores != null}">
```
esotrope: wondering now if we cant do this at the DB query level so we dont have to deal with it here

PR #835, opened 2018-02-05
DWR-1220 Add filters to the custom aggregate report - complete test c…

agorina:
…overage.


agorina:
Mering so that I could pass it to QA. Will address comments if needed.

PR #836, opened 2018-02-06
Make custom aggregate API optionally available.

agorina:
This is per discussion at the stand up today.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
@@ -3,7 +3,6 @@ package org.opentestsystem.rdw.olap.web;
 import org.opentestsystem.rdw.olap.service.CustomAggregateExportService;
 import org.opentestsystem.rdw.olap.service.CustomAggregateService;
 import org.opentestsystem.rdw.reporting.common.model.ReportQuery;
-import org.opentestsystem.rdw.reporting.common.model.ReportRow;
 import org.opentestsystem.rdw.reporting.common.web.security.User;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.io.ByteArrayResource;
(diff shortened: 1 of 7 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
```diff
@@ -33,13 +30,6 @@
         this.exportService = exportService;
     }
 
-    @PostMapping
-    public List<ReportRow> getCustomAggregateReport(
-            @AuthenticationPrincipal final User user,
-            @RequestBody final ReportQuery query) {
-        return service.findByQuery(user, query);
-    }
-
```
agorina: Should the `export` be removed from this controller?

aggregate-service/src/main/resources/application.yml
```diff
@@ -52,6 +52,7 @@ app:
   aggregate-reports:
     summative-enabled: true
     statewide-interim-enabled: false
+    api-access-enabled: true
```
agorina: I made the default as enabled. Any objections?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
esotrope: yeah only thing ui uses is `estimatedRowCount` endpoint now

aggregate-service/src/main/resources/application.yml
esotrope: depends... what does this do?

malaffoon: since there is an embedded setting i don't think i care a lot either way

agorina: @esotrope - we want to have an API that allows for the report to be returned as a response, bypassing the report processor. This is for us to be able to continue testing/researching Redshift. And there was a concern of keeping it open for the production release. This turns it off.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
agorina: Then I will remove it.

aggregate-service/src/main/resources/application.yml
esotrope: gotcha. would rather it not exist for the production release though but this is fine for now

PR #837, opened 2018-02-06
Task/minor tweaks aggregate report ui

wtritch:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -38,8 +38,8 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
     this.assessmentDefinitionsByAssessmentTypeCode = this.route.snapshot.data[ 'assessmentDefinitionsByAssessmentTypeCode' ];
     this.options = this.route.parent.snapshot.data[ 'options' ];
     this.report = this.route.snapshot.data[ 'report' ];
-    this.reportSizeSupported = !Utils.isUndefined(this.report.metadata.row_count)
-      && (Number.parseInt(this.report.metadata.row_count) < SupportedRowCount);
+    this.reportSizeSupported = !Utils.isUndefined(this.report.metadata.totalCount)
+      && (Number.parseInt(this.report.metadata.totalCount) < SupportedRowCount);
```
wtritch: I named the metadata property for aggregate-report row counts "totalCount"

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -87,7 +87,7 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   private initializeReportTables(rows: AggregateReportRow[]): void {
 
     const subjects = this.options.subjects;
-    const assessmentDefinition = this.assessmentDefinitionsByAssessmentTypeCode.get(this.report.request.assessmentTypeCode);
+    const assessmentDefinition = this.assessmentDefinitionsByAssessmentTypeCode.get(this.report.request.reportQuery.assessmentTypeCode);
```
wtritch: The report.request payload is the full AbstractExamExportRequest, not just the ReportQuery in the case of an aggregate report request.

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -130,7 +130,7 @@ export class ReportService {
       headers: new Headers({
         'Accept': 'application/json',
       }),
-      responseType: ResponseContentType.Blob
+      responseType: ResponseContentType.Json
```
wtritch: The front-end wasn't parsing the response as JSON when we told it to use Blob... I suppose that makes sense.

esotrope: ah good catch - think this is default actually - probably can remove the responseType param all together

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -204,6 +204,7 @@ export class ReportService {
     local.subjectCode = remote.subjectCode;
     local.schoolYear = remote.schoolYear;
     local.metadata = remote.metadata || {};
+    local.request = remote.request;
```
wtritch: Stuff the returned request into the Report front-end response object.

PR #838, opened 2018-02-06
Defect/fix agg report requests

esotrope:
Aggregate report requests now go through to the backend from the UI

image

It fails on my local but the request goes through.

### DIFF
--- a/webapp/src/main/webapp/src/app/report/aggregate-report-request.ts
+++ b/webapp/src/main/webapp/src/app/report/aggregate-report-request.ts
@@ -2,6 +2,11 @@
  * Represents an aggregate report request
  */
 export interface AggregateReportRequest {
+  readonly name: string;
+  readonly reportQuery: AggregateReportQuery
+}
+
+export interface AggregateReportQuery {
   readonly achievementLevelDisplayType: string;
   readonly administrationConditionCodes: string[];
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/report/aggregate-report-request.ts
```diff
@@ -20,7 +25,7 @@ export interface AggregateReportRequest {
   readonly lepCodes: string[];
   readonly migrantStatusCodes: string[];
   readonly section504Codes: string[];
-  readonly schoolCodes: string[];
+  readonly schoolIds: number[];
```
esotrope: overzealously named these *codes at some point x D

webapp/src/main/webapp/src/app/shared/organization/organization.mapper.ts
```diff
@@ -8,7 +8,11 @@ import {
 export class OrganizationMapper {
 
   public map(organization: any): Organization {
-    switch (organization.organizationType) {
+
+    // TODO fix models to avoid this bridge
+    const type = organization.type || organization.organizationType;
```
esotrope: appears we have two api server org models being mapped with this same mapper and they do not share signatures

PR #841, opened 2018-02-06
Fix a bug with gender filter.

agorina:


aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -121,7 +121,7 @@ sqlbuilder:
                      JOIN student s ON fe.student_id = s.id
                      JOIN gender gn on gn.id = s.gender_id
                 where: >-
-                     ( true = :all_genders OR gn.code in (:gender_ids) )
+                     ( true = :all_genders OR s.gender_id in (:gender_ids) )
                 groupBy: >-
                      s.gender_id,
```
malaffoon: isn't group by gender_id and code redundant?

agorina: It is (logically), but Redshift does not know it. Since it is in the SELECT (because we want to return codes), it needs to be in a GROUP BY.

PR #844, opened 2018-02-06
Restore padding of IN clause in Redshift SQL

agorina:


agorina:
Merging so that I could test with it, but will address comments if needed.

PR #847, opened 2018-02-07
fixes index out of bound

esotrope:
fixes grouped performance level look ups for sample data table


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -81,7 +81,7 @@ export class AggregateReportTableDataService {
     const groupedPerformanceLevelCounts = [ groupedPerformanceLevelCount, groupedPerformanceLevelCount ];
 
     let uuid = 0;
-    const rows = [];
+    const rows: AggregateReportItem[] = [];
```
esotrope: Needed this for type checking to kick in. i was assuming since i returned the object and the method signature is this type it would get checked but i was wrong : /

PR #848, opened 2018-02-07
adds home link to aggregate reports and related security

esotrope:
Not pretty, but a starting point:

image

Error displayed when the user has CUSTOM_AGGREGATE_REPORTER but cannot create reports for any assessment type.

image


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -1,4 +1,13 @@
-<scroll-nav [sections]="[
+<ng-container *ngIf="accessDenied">
```
esotrope: handled this in the form instead of using a guard. ideally we have route guards that fail with tailored messages and views potentially but I'll have to look over our options for this and change it everywhere potentially.

esotrope:
going to merge for now so qa has a convenient link

PR #849, opened 2018-02-08
optimizes report requests. fixes scrollTop js error defect

esotrope:
1. maps CodedEntities to codes because we are no longer using IDs, this allows us to remove all the code mapping logic running around int the code
2. optimizes report queries to exclude params if ALL is selected - this should help backend performance
3. hotfixes defect with scroll nav


webapp/src/main/webapp/src/app/shared/organization/organization-typeahead.ts
```diff
@@ -54,7 +54,7 @@ export class OrganizationTypeahead extends AbstractControlValueAccessor<string>
   }
 
   onTypeaheadSelectInternal(event: TypeaheadMatch): void {
-    event.item && this.select.emit(event.item);
+    event.item && this.selected.emit(event.item);
```
esotrope: learned its a bad idea to mirror default event names - the input's select event was triggering this event emitter but with the wrong payload - a js event object instead of an org : /

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -9,54 +9,62 @@ import { WindowRefService } from "../core/window-ref.service";
...
-    let activatedSection: Section;
-    for (let section of sections) {
-      if (section.isLink) {
+  private updateActiveLink(): void {
+    const scrollTop = this.getScrollTop();
+    this.items
+    // TODO clean up or use a 3rd party lib if possible
+    // document.getElementById(item.scrollTo.id) is a HOTFIX for offsetTop javascript error
+    // ngOnDestroy is not called in this component because it belongs to a "reusable" view
+    // Angular does not currently support removal of global host listeners
```
esotrope: the ugly underbelly and complexity of angular rears its head

webapp/src/main/webapp/src/app/report/aggregate-report-request.ts
```diff
@@ -8,25 +8,25 @@ export interface AggregateReportRequest {
...
-  readonly migrantStatusCodes: string[];
-  readonly section504Codes: string[];
-  readonly schoolIds: number[];
+  readonly lepCodes?: string[];
+  readonly migrantStatusCodes?: string[];
+  readonly section504Codes?: string[];
+  readonly schoolIds?: number[];
   readonly schoolYears: number[];
-  readonly subjectCodes: string[];
+  readonly subjectCodes?: string[];
```
esotrope: added `?` to the optional ones

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -0,0 +1,81 @@
...
+import { AggregateReportFormOptions } from "./aggregate-report-form-options";
+
+const equalSize = (a: any[], b: any[]) => a.length === b.length;
+const idsOf = values => values.map(value => value.id);
+
+/**
+ * Responsible for creating aggregate report requests from supplied models
+ */
+@Injectable()
+export class AggregateReportRequestMapper {
```
esotrope: @agorina  where the optimization happens

agorina: Yup, saw it. Thank you!

PR #850, opened 2018-02-08
touches up aggregate report form ui

esotrope:
made the links smaller to avoid overlap as much as possible

image

moved header out of well and separated table and summary wells so they dont expand together with the table

image


webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -50,14 +60,17 @@ export class ScrollNavComponent {
...
-        if (itemOffsetTop - 5 <= scrollTop) {
+        if (itemOffsetTop - 20 <= scrollTop) {
           this._activeItem = item;
         } else {
           return;
         }
       });
+    if (!this.activeItem && this.items.length) {
+      this._activeItem = this.items[0];
+    }
```
esotrope: now first item is default selected

PR #851, opened 2018-02-08
improves report creation workflow

esotrope:
Loading message
image

If the report fails
image

If the report is large
image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
@@ -256,80 +256,17 @@ export class AggregateReportFormComponent {
    * Creates a report if the form is valid
    */
   onGenerateButtonClick(): void {
-    console.log('called!')
     this.validate(this.formGroup, () => {
-      const request = this.createReportRequest();
-      this.reportService.getEstimatedRowCount(request.reportQuery)
+      this.reportService.createReport(this.createReportRequest())
         .subscribe(
-          count => {
-            if (count < SupportedRowCount) {
-              this.createViewableReport(request);
-            } else {
-              this.showConfirmationModal(count, () => this.createNonViewableReport(request))
-            }
+          resource => {
+            this.router.navigate([ resource.id ], { relativeTo: this.route });
+          },
+          error => {
+            this.notificationService.error({ id: 'labels.reports.messages.submission-failed.html', html: true });
           }
-        )
+        );
     });
-  }
-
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -256,82 +256,19 @@ export class AggregateReportFormComponent {
    * Creates a report if the form is valid
    */
   onGenerateButtonClick(): void {
-    console.log('called!')
```
esotrope: stripped out logic for launching modal when report is too large. now it sends you to the results page when the report resource is created regardless of other variables

wtritch:
nit: I would make the "Download Data" button on the "too large" page a blue confirm color rather than the default gray cancel color.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -195,7 +195,11 @@ export class AggregateReportTableComponent implements OnInit {
   }
 
   get paginationEnabled(): boolean {
-    return !this.preview;
+    return !this.preview
+      && this.table.rows
+      && this.resultsTable.rowsPerPageOptions
+      && this.resultsTable.rowsPerPageOptions.length
+      && this.table.rows.length > this.resultsTable.rowsPerPageOptions[0];
```
wtritch: 👍

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.html
```diff
@@ -1,40 +1,55 @@
...
-</ng-container>
-
-<ng-container *ngIf="!reportSizeSupported">
-  <div class="well">
-    <p>{{'aggregate-report.unsupported-size-error' | translate}}</p>
-    <ng-container *ngTemplateOutlet="updateRequestButton"></ng-container>
+<div class="row mb-sm">
+  <div class="col-md-12">
+    <button class="btn btn-default btn-sm" (click)="onUpdateRequestButtonClick()">
+      <i class="fa fa-chevron-left mr-xs"></i>{{'aggregate-report.update-request' | translate}}
```
wtritch: I still thing we should change the verbiage from "Update/Modify" to "View"

esotrope: good idea - should be the same everywhere

PR #857, opened 2018-02-09
Feature/saturate aggregate query from report

wtritch:
This PR adds an optional report id parameter to the aggregate report query builder page.  If provided, the form settings are initialized to the same values as the reference report's request.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/UserOrganizationService.java
```diff
@@ -23,12 +24,11 @@
...
-     * @param user       the user accessing the organizations
-     * @param nameSearch the text used to match against
-     * @return all the schools the user has access to
+     * @param user      the user accessing the organizations
+     * @param query     the organization query
+     * @return all the matching organizations the user has access to
      */
-    Set<Organization> getOrganizations(@NotNull User user, String nameSearch);
-
+    Set<Organization> getOrganizations(@NotNull User user, OrganizationQuery query);
```
wtritch: I modified the /organizations endpoint to accept an OrganizationQuery rather than just a name.  This allows searching by name, type, and/or id.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationServiceTest.java
```diff
@@ -48,36 +49,47 @@
...
 
     @Test
-    public void itShouldNotThrowExceptionOnNullSearch() {
-        assertThat(service.getOrganizations(userWithStatewidePermission, null)).isEmpty();
+    public void itShouldReturnAllOrgsOnEmptyQuery() {
+        assertThat(service.getOrganizations(
+                userWithStatewidePermission,
+                OrganizationQuery.builder().build()))
+                .containsOnly(district1, district2, school1, school2);
+    }
```
wtritch: This is a slight functionality change, but it leads to a more RESTful API:
A GET /organizations request without any filters now returns all the organizations that a user has access to.  However, this isn't currently being done by our API (and probably shouldn't be.)

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/OrganizationQueryArgumentResolver.java
```diff
@@ -1,6 +1,7 @@
-package org.opentestsystem.rdw.admin.web;
+package org.opentestsystem.rdw.reporting.common.web;
```
wtritch: I'm re-using the common OrganizationQuery model.  Since we now need the OrganizationQueryResolver in both admin and aggregate-service packages I moved the resolver into common-web.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+        expect(optionMapper.map).toHaveBeenCalledWith(route.parent.data.options);
+        expect(organizationService.getOrganizationsByIdAndType)
+          .toHaveBeenCalledWith(OrganizationType.School, [2]);
+        expect(organizationService.getOrganizationsByIdAndType)
+          .toHaveBeenCalledWith(OrganizationType.District, [1]);
+        done();
+      });
+  });
+
+  function mockReport(): Report {
```
wtritch: Building out all of these mocks for testing was a PITA.  As we start back-filling front-end aggregate report tests (fingers crossed) these should probably be moved into test builders so we don't have these impls in multiple places.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
```diff
@@ -0,0 +1,190 @@
...
+const firstValueOf = values => values[ 0 ].value;
+const hasValues = (value: any[]) => value && value.length;
+const hasOption = (options: Option[], value: any) => options.find(option => option.value === value) != null;
+
+/**
+ * This resolver is responsible for fetching an aggregate report based upon
+ * an optional report id query parameter.
+ */
+@Injectable()
+export class AggregateReportFormSettingsResolve implements Resolve<AggregateReportFormSettings> {
```
wtritch: This resolver is now responsible for initializing the aggregate report form settings.  If a "src" query param is provided, it'll initialize the settings to the referenced aggregate report request query.  If not, it'll return a settings object with default values (based on the available report options.)

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -91,4 +84,38 @@
         };
     }
 
+    private Predicate<Organization> toNamePredicate(final OrganizationQuery query) {
+        //If we're not querying on name, don't filter results
```
esotrope: we're already filtering i would say "filter out" vs "filter in"

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
```diff
@@ -0,0 +1,190 @@
...
+      return Observable.of(settings);
+    }
+
+    //Fetch the report, lookup schools/districts, map to settings
+    return this.service.getReportById(Number.parseInt(reportId))
+      .flatMap(report => this.saturateOrgs(report.request))
+      .map(saturatedRequest => {
+        this.mapOntoSettings(saturatedRequest, options, settings);
+        return settings;
+      })
```
wtritch: First I get the report.  Then I go grab all of the Districts and Schools referenced by the report by id.  Then I jam the whole thing together into the resolved settings.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
```diff
@@ -33,20 +37,29 @@ public String getName() {
         return name;
     }
 
+    /**
+     * @return The organization ids (only useful if a single type is selected)
+     */
+    public Set<Long> getIds() {
+        return ids;
+    }
```
esotrope: so do i have to make two queries for schools and districts?
i would have to if school / district entity IDs collide right?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -105,6 +102,9 @@ export class AggregateReportFormComponent {
 
     this.aggregateReportOptions = route.parent.snapshot.data[ 'options' ];
 
+    this.settings = route.snapshot.data[ 'settings' ];
+    this.organizations = this.organizations.concat(this.settings.districts, this.settings.schools);
```
wtritch: Took me a little while to realize we are double-recording our organizations... I wonder if we should instead have a programmatic "getter" that concats the districts and schools from the settings object...

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.spec.ts
esotrope: tests are for suckers and quality code bases x P

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -1,6 +1,9 @@
 import { Component, Input } from "@angular/core";
 
-const valueToOptionText = attribute => value => attribute.options.find(option => option.value === value).text;
+const valueToOptionText = attribute => value => {
+  const option: any = attribute.options.find(option => option.value === value);
+  return option ? option.text : undefined;
+};
```
wtritch: I ran into a problem with a hand-jammed report request that had values that didn't exist in the options.  This minor change just strips them out if they don't match up.  (e.g. assessmentGradeCode = "01")

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
```diff
@@ -0,0 +1,190 @@
...
+import { AggregateReportFormSettings } from "./aggregate-report-form-settings";
+import { AggregateReportQuery, AggregateReportRequest } from "../report/aggregate-report-request";
+import { Option } from "../shared/form/sb-checkbox-group.component";
+import { District, OrganizationType, School } from "../shared/organization/organization";
+import { AggregateReportOrganizationService } from "./aggregate-report-organization.service";
+import { AggregateReportFormOptionsMapper } from "./aggregate-report-form-options.mapper";
+
+const valuesOf = values => values.map(value => value.value);
+const firstValueOf = values => values[ 0 ].value;
+const hasValues = (value: any[]) => value && value.length;
```
esotrope: `Utils.isNullOrEmpty()` does this

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -194,14 +201,6 @@ export class AggregateReportTableComponent implements OnInit {
     return this.preview ? false : 'custom';
   }
 
-  get paginationEnabled(): boolean {
-    return !this.preview
-      && this.table.rows
-      && this.resultsTable.rowsPerPageOptions
-      && this.resultsTable.rowsPerPageOptions.length
-      && this.table.rows.length > this.resultsTable.rowsPerPageOptions[0];
-  }
```
wtritch: This was causing a console error in the report view component.  We use the paginationEnabled property to initialize the resultsTable, but the act of initializing the resultsTable changed the paginationEnabled property value.  That made Angular choke a bit.  I changed the flow to manually set and update the paginationEnabled property within the expected flow.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
```diff
@@ -0,0 +1,190 @@
...
+      : Observable.of([]);
+
+    const districtIds: number[] = request.reportQuery.districtIds;
+    const districts: Observable<District[]> = hasValues(districtIds)
+      ? this.organizationService.getOrganizationsByIdAndType(OrganizationType.District, districtIds)
+      : Observable.of([]);
+
+    return Observable.forkJoin(schools, districts)
+      .map((results) => {
+        let [ schools, districts ] = results;
```
esotrope: 👍 taking advantage of the nice operators i see

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -92,8 +92,7 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   }
 
   onUpdateRequestButtonClick(): void {
-    // TODO pass report ID for resolution
-    this.router.navigate([ '..' ], { relativeTo: this.route });
+    this.router.navigate([ '..'  ], { relativeTo: this.route, queryParams: {src: this.report.id} });
```
wtritch: This updates the "Modify Query" button on the report view page to provide the report id when navigating "back" to the query builder.  This lets us get back to the saturated query builder even if we navigated to the report view from the "My Reports" page.

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -48,22 +44,6 @@ export class ReportsComponent implements OnInit, OnDestroy {
...
-        },
-        () => {
-          this.notificationService.error({ id: 'labels.reports.messages.download-failed' });
-        }
-      );
-  }
-
-  regenerate(report: Report): void {
-    this.notificationService.info({ id: 'labels.reports.messages.coming-soon.report-regeneration' });
-  }
```
wtritch: unrelated, this is cleanup from an earlier PR.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -1,6 +1,9 @@
 import { Component, Input } from "@angular/core";
 
-const valueToOptionText = attribute => value => attribute.options.find(option => option.value === value).text;
+const valueToOptionText = attribute => value => {
+  const option: any = attribute.options.find(option => option.value === value);
+  return option ? option.text : undefined;
```
esotrope: what case is this covering? the users previous report selected an option that is no longer available?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
wtritch: That was my thought.

webapp/src/main/webapp/src/app/report/reports.component.ts
esotrope: 👍

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
esotrope: i generally try to keep the models dumb and just interfaces but if we must we can. also we can make the org-select component accept two arrays of schools/districts and track them independently

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
wtritch: Ah, good call.  Changed.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
esotrope: could the query signature not be `organizationIdsByOrganizationType: Map` ? and then the backend makes two calls but the front end only makes one?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
wtritch: See comment above =)  (fixing a hopefully-non-real-world case of a report request with values that are no longer viable options)

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
wtritch: Changed.  It's fun that "Filter" and "Predicate" do opposite things in the "true" case.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
wtritch: I really hate having maps in API payloads... two requests to our heavily-cached API shouldn't cause us any issues...

esotrope: fair nuff

wtritch:
@esotrope thanks!  It rabbit-holed a bit, but not as badly as I’d feared.

PR #859, opened 2018-02-10
application administration dropdown

dsnyder-fairway:
## Application administration has moved to a dropdown
screen shot 2018-02-09 at 3 58 35 pm

screen shot 2018-02-09 at 3 58 42 pm

## removed application administration breadcrumb
screen shot 2018-02-09 at 3 58 51 pm

## different user permissions
screen shot 2018-02-09 at 3 59 47 pm

### note
I think this should be available on all of the links provided by this dropdown. Now, to navigate, users must go to the home page in order to go to a different admin page


aaronsleeper:
looks good mang

esotrope:
The fa-angle-down or fa-chevron-down should be right of the text i think. we do that everywhere else at least. 
Actually should this be 3 dots or an arrow? i think we use three dots for menus elsewhere @aaronsleeper

webapp/src/main/webapp/src/app/admin/home/home.component.html
```diff
@@ -1,24 +0,0 @@
...
-      </div>
-      <div>
-        <a *hasPermission="'INSTRUCTIONAL_RESOURCE_WRITE'" routerLink="instructional-resource">{{'labels.instructional-resource.title' | translate}}</a>
-      </div>
-      <div>
-        <a *hasPermission="'EMBARGO_WRITE'" routerLink="embargoes">{{'labels.embargo.title' | translate}}</a>
-      </div>
-    </div>
-  </div>
-</div>
```
esotrope: 👍

webapp/src/main/webapp/src/app/home/admin-dropdown.component.html
```diff
@@ -0,0 +1,21 @@
+<div dropdown style="display: inline-block;" class="btn-group ml-xs">
+  <a dropdownToggle type="button"
+     *hasAnyPermission="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE', 'EMBARGO_WRITE']"
+     class="btn btn-xs btn-info" (click)="toggle()"
+     title="{{'labels.groups.admin-button' | translate}}"><i class="fa mr-xs"
+                                                             [ngClass]="{'fa-chevron-right': !expanded, 'fa-chevron-down': expanded}"></i>
+    {{'labels.groups.admin-button' | translate}}</a>
+  <ul *dropdownMenu class="dropdown-menu" role="menu">
+    <li><a *hasPermission="'GROUP_WRITE'" routerLink="admin-groups">{{ 'labels.admin-groups.title' |
```
esotrope: would prefix router links with `/`
unless they are under /home/admin ... i dont think it will work in `ng build --prod` distribution. - ran into this with the recent aggregate link i added

dsnyder-fairway:
@esotrope I believe we are moving away from the 3 dots. This is to follow a similar pattern found on student responses
screen shot 2018-02-09 at 4 33 54 pm

webapp/src/main/webapp/src/app/home/admin-dropdown.component.html
```diff
@@ -0,0 +1,21 @@
+<div dropdown style="display: inline-block;" class="btn-group ml-xs">
+  <a dropdownToggle type="button"
+     *hasAnyPermission="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE', 'EMBARGO_WRITE']"
```
esotrope: I think this should be on the declared component just because if they dont have permission we shouldnt have an empty div floating around.

home.html:
```html
<ng-container *hasAnyPermission="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE', 'EMBARGO_WRITE']"></ng-container>
```

webapp/src/main/webapp/src/app/home/admin-dropdown.component.html
```diff
@@ -0,0 +1,21 @@
...
+     class="btn btn-xs btn-info" (click)="toggle()"
+     title="{{'labels.groups.admin-button' | translate}}"><i class="fa mr-xs"
+                                                             [ngClass]="{'fa-chevron-right': !expanded, 'fa-chevron-down': expanded}"></i>
+    {{'labels.groups.admin-button' | translate}}</a>
+  <ul *dropdownMenu class="dropdown-menu" role="menu">
+    <li><a *hasPermission="'GROUP_WRITE'" routerLink="admin-groups">{{ 'labels.admin-groups.title' |
+      translate}}</a>
+    </li>
+    <li>
+      <a *hasPermission="'INSTRUCTIONAL_RESOURCE_WRITE'" routerLink="instructional-resource">{{'labels.instructional-resource.title'
```
esotrope: similarly hasPermission should be on the `<li>` element so we dont have empty elements hanging around when the user doesnt have the permission

dsnyder-fairway: done

webapp/src/main/webapp/src/app/home/admin-dropdown.component.html
dsnyder-fairway: done, except I was able to add this to

```
<admin-dropdown *hasAnyPermission="['GROUP_WRITE', 'INSTRUCTIONAL_RESOURCE_WRITE', 'EMBARGO_WRITE']"></admin-dropdown>
```

esotrope:
@dsnyder-fairway i'm fine with the arrows - looks better imo anyway. i think it should be on the right side though to follow the rest of the app. the table row expansion is different than general menu expansion i was thinking.

dsnyder-fairway:
screen shot 2018-02-09 at 4 47 20 pm

screen shot 2018-02-09 at 4 47 24 pm

👍

esotrope:
i think there is also a bootstrap dropdown left option to make it expand left instead of right too.
That will prevent it from going off the page hopefully

malaffoon:
We might have to do a sweep through the app for the `v` vs. `>`, left vs right thing. The SB folks have been commenting about it a lot.

But these changes look good, thanks.

PR #860, opened 2018-02-10
Handle exam exports when the user only has rights to multiple Schools.

wtritch:
This PR fixes an issue with exporting exams when a user only has rights to multiple schools and no districts.

### DIFF
--- a/webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
+++ b/webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
@@ -118,10 +118,14 @@ export class OrganizationMapper {
     organizations.schools.forEach(school => {
       let node = root;
       if (!Utils.isUndefined(school.districtId)) {
-        node = node.getOrCreate(x => x.id === school.districtId, this.getDistrict(organizations, school.districtId));
+        node = node.getOrCreate(
+          x => x.id === school.districtId,
+          this.getOrCreateDistrict(organizations, school.districtId));
       }
       if (!Utils.isUndefined(school.schoolGroupId)) {
-        node = node.getOrCreate(x => x.id === school.schoolGroupId, this.getSchoolGroup(organizations, school.schoolGroupId))
+        node = node.getOrCreate(
+          x => x.id === school.schoolGroupId,
+          this.getOrCreateSchoolGroup(organizations, school.schoolGroupId))
       }
       node.create(school);
     });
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
```diff
@@ -118,10 +118,10 @@ export class OrganizationMapper {
     organizations.schools.forEach(school => {
       let node = root;
       if (!Utils.isUndefined(school.districtId)) {
-        node = node.getOrCreate(x => x.id === school.districtId, organizations.districtsById.get(school.districtId))
+        node = node.getOrCreate(x => x.id === school.districtId, this.getDistrict(organizations, school.districtId));
```
wtritch: The problem was that when executing the "matcher" `x => x.id === school.districtId` x was null, leading to an NPE when de-referencing `x.id`
I changed the call to create a shallow matchable District and SchoolGroup placeholder that just contains the ID if the district or school group can't be found in the user's organizations.

webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
```diff
@@ -132,6 +132,26 @@ export class OrganizationMapper {
     return Utils.isUndefined(a) ? b : a;
   }
 
+  private getDistrict(organizations: UserOrganizations, id: number): District {
```
esotrope: `getDistrictOrCreatePlaceholder()`

webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
```diff
@@ -132,6 +132,26 @@ export class OrganizationMapper {
...
+    if (organizations.districtsById.has(id)) {
+      return organizations.districtsById.get(id);
+    }
+
+    let district: District = new District();
```
esotrope: `const`

esotrope: look i'm the @malaffoon of typescript : j

webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
wtritch: Ugh, i’m going to break these into multiple lines with that rename

webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
wtritch: done

webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
wtritch: done

PR #864, opened 2018-02-10
DWR-1346 OLAP API: implement parallel execution of SQLs in aggregate report

agorina:
I tested it with ASW DEV (configured with Concurrency = 6) and one user running a (cached) previously executed query. My request had a school with its district and state included and 7 dimensions (including overall) - total 21 SQLs. As expected, the results are:

Before change: 9211 ms
After change:  3610 ms

@malaffoon and @wtritch, Please take another look before I merge it. Thank you.


aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
```diff
@@ -52,7 +55,10 @@
         QueryProvider.class,
         QueryWithReportTemplate.class
 })
-@Sql(scripts = {"classpath:OlapEntitiesSetup.sql"})
+@SqlGroup({
+        @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:OlapEntitiesSetup.sql"}),
+        @Sql(executionPhase = AFTER_TEST_METHOD, scripts = {"classpath:OlapEntitiesTeardown.sql"})
+})
```
agorina: New threads did not see data within a transaction. 
This change increases the RST execution time.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcActiveAssessmentRepositoryRST.java
```diff
@@ -14,7 +14,7 @@
 @Import(JdbcActiveAssessmentRepository.class)
 @Sql(scripts = {"classpath:OlapEntitiesSetup.sql"})
 @CachingTest
-public class JdbcActiveAssessmentRepositoryRST extends RepositoryBackedIT {
+public class JdbcActiveAssessmentRepositoryRST extends TransactionalRepositoryBackedIT {
```
agorina: These changes are explained below.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -35,11 +41,12 @@
     private final QueryProviderRepositoryHelper reportQueryRepositoryHelper;
     private final ActiveAssessmentRepository activeAssessmentRepository;
     private final QueryProvider queryProvider;
-
     private final Organization state;
 
     @Value("${app.aggregate-reports.custom-aggregate-report-repository.organization-partition-size}")
     private int organizationPartitionSize;
+    @Value("${app.aggregate-reports.custom-aggregate-report-repository.query-partition-size}")
```
malaffoon: I find this name confusing. Just to verify, this is the max number of simultaneous queries we want to execute? So, in your example, 6, right? Perhaps queryConcurrency? queryPoolSize? 

I'm also not thrilled with the really long property names. These are things that IT folks will be modifying. Yeah, they *should* copy and paste from documentation but the longer the name, the more likely there will be a typo and the harder it is to read and diagnose.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -62,22 +68,55 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+
+            final List<RunQueryTask> runQueryTask = newArrayList();
+            for (final QueryWithReportTemplate queryWithReportTemplate : queries) {
+                runQueryTask.add(new RunQueryTask(queryWithReportTemplate, query.getAssessmentTypeCode()));
+            }
+            final ExecutorService executor = newFixedThreadPool(queries.size());
+            final List<CompletableFuture<List<ReportRow>>> taskFutures = runQueryTask
+                    .stream()
+                    .map(t -> CompletableFuture.supplyAsync(t::run, executor))
+                    .collect(toList());
```
malaffoon: Seems like 83-91 could be combined into a single stream chain. Haven't tried it but something like:
```
    taskFutures = queries.stream()
        .map(q -> new RunQueryTask(q, query.getAssessmentTypeCode)
        .map(t -> CompletableFuture.supplyAsync(t::run, executor))
        .collect(toList());
````
Just trying to avoid multiple stream chains.

File: aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java [hunk 5]
```diff
@@ -62,22 +68,55 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+
+            final List<ReportRow> taskResults = taskFutures
+                    .stream()
+                    .map(CompletableFuture::join)
+                    .collect(ArrayList::new, List::addAll, List::addAll);
```
malaffoon: hmm, haven't done futures since streams were introduced; looks fine just pontificating

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -90,32 +129,29 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
-        for (final ActiveAssessment activeAssessment : activeAssessmentRepository.findAllByTypeCode(assessmentTypeCode)) {
-            if (activeAssessment.getId() == assessmentId && activeAssessment.getExamSchoolYear() == schoolYear)
-                return activeAssessment;
+        private ReportRow removeFromTemplateIfExists(final Set<ReportRow> template, final ReportRow rowToVerify) {
+            ReportRow foundRow = null;
+            for (final ReportRow row : template) {
+                if (row.getDimension().equals(rowToVerify.getDimension()) && row.getOrganization().equals(rowToVerify.getOrganization()) && row.getAssessment().equals(rowToVerify.getAssessment()))
+                    foundRow = row;
+            }
+            if (foundRow != null) template.remove(foundRow);
```
malaffoon: Can't this logic just be in the loop? Or is there a case where the row might be in there more than once and we only want to return the last instance? If not, can't this whole thing be replaced with something like `template.removeIf(predicate that matches if condition);'?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
malaffoon: And then i wonder if that can be combined with lines 93-96?

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
malaffoon: It increases the RST time for just this IT, right?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -62,22 +68,55 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+                    .collect(ArrayList::new, List::addAll, List::addAll);
+
+            reportRows.addAll(taskResults);
+        }
+        return reportRows;
+    }
 
-        for (final QueryWithReportTemplate queryWithReportTemplate : customAggregateQueryProvider.toQuery(query)) {
-            final Set<ReportRow> templateRows = queryWithReportTemplate.getReportTemplate();
+    class RunQueryTask {
```
malaffoon: i think this should be `private static class`

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -62,22 +68,55 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+        private final QueryWithReportTemplate queryWithReportTemplate;
+        private final String asmtTypeCode;
 
+        RunQueryTask(final QueryWithReportTemplate queryWithReportTemplate, final String asmtTypeCode) {
+            this.queryWithReportTemplate = queryWithReportTemplate;
+            this.asmtTypeCode = asmtTypeCode;
+        }
+
+        List<ReportRow> run() {
```
malaffoon: perhaps a short javadoc explaining what this is basically doing?

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepositoryRST.java
agorina: Yes, but there are many tests in here. I think it took 15 min.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Your understanding of the property is correct. I can rename it to anything you like.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
jtreuting: Isn't there a parallelStream() that takes care of some of this?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -62,22 +68,55 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
+        final List<List<QueryWithReportTemplate>> partitionedQueries = partition(customAggregateQueryProvider.toQuery(query), partitionSize);
+
         final List<ReportRow> reportRows = newArrayList();
+        for (final List<QueryWithReportTemplate> queries : partitionedQueries) {
+
+            final List<RunQueryTask> runQueryTask = newArrayList();
+            for (final QueryWithReportTemplate queryWithReportTemplate : queries) {
+                runQueryTask.add(new RunQueryTask(queryWithReportTemplate, query.getAssessmentTypeCode()));
+            }
+            final ExecutorService executor = newFixedThreadPool(queries.size());
```
wtritch: Hmmm... i’d be tempted to have a single executor that gets injected

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
wtritch: hmm... this will block the current thread until all CompletableFutures return.  Which means we won't be saturating the WLM queue if we have multiple users, right?  If we have a request that has 12 queries:
11 of them complete in under a second
1 of them takes 15 seconds
then we'll be blocking the aggregate-service from making more requests until that 15 second query completes?

Is it weird I was expecting to see a `CompletableFuture.allOf()` call with the completion response posted to rabbit on the completion of the CompletableFuture that represents all of the user's queries?  
Although, then we'd end up queuing up requests against the CompletableFuture executor rather than queueing user requests in the rabbit queue... that wouldn't be good...

malaffoon: @wtritch I went through the same thought process but i think this will work just as if we used `CompletableFuture.allOf()`. Yes, it will block on the futures waiting for them to return but the overall effect will be the same: the longest running query in each group will be the gating factor. The issue there is that unlucky distribution of the queries could result in a long wait if each group has a single slow query. Kind of like multiple lines in a grocery store. 

I think a "better" way to do it would be to have a single MTS queue that has all the queries in it. Then create a pool of runners that pull from that queue. The size of the pool would match the WLM concurrency. In that case a single slow query wouldn't block the group of queries it is in. The same unlucky distribution would likely distribute in parallel. More like a single line in a grocery store feeding all the registers.

Is it worth reworking? Meh.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Who is @jtreuting  and why is asking questions here? :)

My understanding is that ParallelStream and CompletableFuture are two solutions to the same problem. I believe that CompletableFuture approach is more flexible in controlling the number of threads in the pool. However, I do not want to pretend to be an expert in this area, please correct me if needed.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: If I read it correctly, I think @malaffoon and @wtritch  are discussing slightly different issues. I suggest that we whiteboard it on Monday (or some time next week), but here is how I see it.  

There are three separate concerns:
1. Improve the system responsiveness with one user.
2. Be able to support multiple concurrent users as fast as we can.
3. We know that running all queries as they come from multiple users overflows Redshift. So we are  'funneling' requests to Redshift be sending them through a report processing pipeline. Before this change, the number of concurrent calls to Redshift (with our architecture) was equal to the number of Aggregate Services running. Each report is built from up to 28 SQL statements that were executed sequentially. This means that we had a very 'narrow' funnel that was limited to one Redshift call at a time.

We have discussed:
1. To improve a single user response time, we want to parallelize calls to Redshift. This is what this PR is about. This allows to widen our funnel to a configured number of threads.
2. To improve multiple user response we need to either have multiple Aggregate Services running or configure multiple channels. Otherwise our funnel allows only one **report** at a time. I still think we need to do this.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Good idea! Changing.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: Ok.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
agorina: You are correct, it could be replaced as you are suggesting. Changing.

webapp/src/main/resources/application.yml
```diff
@@ -37,7 +37,7 @@ server:
 spring:
   datasource:
     url: "\
-      jdbc:mysql://${spring.datasource.url-server:localhost:3306}/$spring.datasource.url-schema:reporting}\
+      jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:reporting}\
```
agorina: Unrelated, fixing a typo.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -62,22 +69,41 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
...
     public List<ReportRow> findByQuery(@NotNull final ReportQuery query) {
-        final CustomAggregateQueryWithReportTemplateProvider customAggregateQueryProvider = new CustomAggregateQueryWithReportTemplateProvider(state, organizationPartitionSize, queryProvider, reportQueryRepositoryHelper);
-        final List<ReportRow> reportRows = newArrayList();
+        return new CustomAggregateQueryWithReportTemplateProvider(state, organizationPartitionSize, queryProvider, reportQueryRepositoryHelper)
+                .toQuery(query)
+                .stream()
+                .map(q -> new RunQueryTask(q, query.getAssessmentTypeCode()))
+                .map(t -> CompletableFuture.supplyAsync(t::run, threadPoolTaskExecutor))
+                .collect(toList())
+                .stream()
```
malaffoon: i think the collect() and stream() can be skipped?

agorina: Correct, thank you.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -46,13 +50,15 @@ public JdbcCustomAggregateReportRowRepository(@Value("${app.state.name}") final
...
-                                                  final QueryProvider queryProvider) {
+                                                  final QueryProvider queryProvider,
+                                                  final TaskExecutor threadPoolTaskExecutor) {
         this.jdbcTemplate = jdbcTemplate;
         this.reportQueryRepositoryHelper = reportQueryRepositoryHelper;
         this.activeAssessmentRepository = activeAssessmentRepository;
         this.queryProvider = queryProvider;
-
         this.state = Organization.builder().name(stateName).organizationType(OrganizationType.State).build();
+        this.threadPoolTaskExecutor = threadPoolTaskExecutor;
```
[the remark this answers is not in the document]
malaffoon: We definitely don't want a separate pool; there are reasons apart from matching concurrency levels for a connection pool. So we want one connection pool that is big enough to satisfy the number of request threads we allow. Just to throw some example numbers out there, suppose we set WLM=12, our query concurrency to 4 ... then we'd want to set our request thread to 3 and our connection pool to >15ish.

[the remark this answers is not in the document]
malaffoon: I think i agree with you ... assuming that a single pod will be handling multiple user requests, then we'd want a query thread pool per request. Might want to squiggle on a whiteboard with Will in the room cause i know dealing with webapp requests is complicated.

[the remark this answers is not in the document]
agorina: Yes, I am not going to make any changes until the three of us whiteboard it.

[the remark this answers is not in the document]
agorina: @wtritch and I chatted about it and have agreed to keep it as is. We also discuss adjusting the configuration in two places:
1. the size of this pool
2. stream concurrency

I will add a note to the open JIRA about testing it under load to consider/test different configurations.

PR #866, opened 2018-02-10
configuration tweaks

malaffoon:
I ran into some confusion and concern when i was setting up UAT. I'll have a couple of these little tweak PR's ...

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateReportController.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateReportController.java
@@ -19,7 +19,7 @@ import java.util.List;
  */
 @RestController
 @RequestMapping("/aggregate")
-@ConditionalOnProperty(prefix = "app.aggregate-reports", name = "api-access-enabled", havingValue = "true")
+@ConditionalOnProperty(prefix = "app.aggregate-reports", name = "api-enabled", havingValue = "true")
 class CustomAggregateReportController {
 
     private final CustomAggregateService service;
--- a/aggregate-service/src/main/resources/application.yml
+++ b/aggregate-service/src/main/resources/application.yml
@@ -54,7 +54,7 @@ app:
     statewide-interim-enabled: false
     custom-aggregate-report-repository:
       organization-partition-size: 10
-    api-access-enabled: true
+    api-enabled: false
 
 spring:
   jackson:
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/ExamRepository.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/ExamRepository.java
@@ -7,5 +7,15 @@ import javax.validation.constraints.NotNull;
 
 public interface ExamRepository {
 
+    /**
+     * Execute the exam report, writing the results to the requested (relative) location
+     * and returning the actual (relative) location (may be different due to behavior of
+     * repository wrt archive service).
+     *
+     * @param permissionScope permissions to apply during exam report
+     * @param query exam report query params
+     * @param uriLocation requested (relative) location, e.g. REPORTS/USER/123/MyReport.csv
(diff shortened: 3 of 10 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateReportController.java
```diff
@@ -19,7 +19,7 @@
  */
 @RestController
 @RequestMapping("/aggregate")
-@ConditionalOnProperty(prefix = "app.aggregate-reports", name = "api-access-enabled", havingValue = "true")
+@ConditionalOnProperty(prefix = "app.aggregate-reports", name = "api-enabled", havingValue = "true")
```
malaffoon: perhaps i'm being too pedantic but it is enabling the API end-point regardless of whether you can access it; wouldn't make this change but i was changing it anyway

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/ExamRepository.java
```diff
@@ -7,5 +7,15 @@
...
+    /**
+     * Execute the exam report, writing the results to the requested (relative) location
+     * and returning the actual (relative) location (may be different due to behavior of
+     * repository wrt archive service).
+     *
+     * @param permissionScope permissions to apply during exam report
+     * @param query exam report query params
+     * @param uriLocation requested (relative) location, e.g. REPORTS/USER/123/MyReport.csv
+     * @return actual (relative) location, e.g. REPORTS/USER/123/MyReport.csv.part_0000
+     */
```
malaffoon: It wasn't obvious to me what this method was doing so, after i figured it out, i added javadocs.

aggregate-service/src/main/resources/application.yml
```diff
@@ -54,7 +54,7 @@ app:
     statewide-interim-enabled: false
     custom-aggregate-report-repository:
       organization-partition-size: 10
-    api-access-enabled: true
+    api-enabled: false
```
malaffoon: Change #1 - default this to false

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportService.java
```diff
@@ -21,18 +20,11 @@
...
-    @Value("${app.archive.export-root:}")
-    public void setExportBasePath(final String value) {
-        if(!exportBasePath.equals(value)) {
-            exportBasePath = value.endsWith("/") ? value : value + "/";
-        }
-    }
-
-    private String exportBasePath = "";
+    // store exam exports near the other user reports
+    static final String ExportBasePath = ArchiveBackedReportContentService.BasePath + "EXAM-EXPORT/";
```
malaffoon: Change #2 - there is no point making this configurable because the other base path is not configurable. Pretty much none of our relative path locations in the archive are configurable

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveExamExportServiceTest.java
```diff
@@ -5,6 +5,7 @@
 import org.junit.runner.RunWith;
```
malaffoon: Had to rework this test suite a lot because it relied on changing the export base path. Pretty sure i didn't didn't mess up the intent of the tests.

aggregate-service/src/main/resources/application.yml
agorina: At least you remember it :) 

@wtritch, we now need to remember to turn it on in DEV **if** we want to load test direct calls to Redshift.

agorina: @malaffoon, I just realised that Nohemi has been this end point in QA to verify if an issue is UI related or back end.  We should either let her know or enable it in QA.

malaffoon: Ok, i enabled it in dev and qa

agorina: Thank you!

PR #867, opened 2018-02-10
change datasource prefix from custom app.datamart to standard spring.

malaffoon:
I know we've gone back and forth on this in the reporting project but Friday i found myself explaining once again to the IT/Ops folks why some datasources are under `app.datamart` and some are under `spring`. I think in this case consistency, however foolish, is a hobgoblin we want, to make configuration easier (remember, *every* service in *every* deployment requires at least one datasource property to be set).

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/DataSourceConfiguration.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/DataSourceConfiguration.java
@@ -21,7 +21,7 @@ public class DataSourceConfiguration {
      * @return DataSource
      */
     @Bean(name = "olapDatasource")
-    @ConfigurationProperties(prefix = "app.datamart.olap_datasource")
+    @ConfigurationProperties(prefix = "spring.olap_datasource")
     public DataSource olapDataSource() {
         return DataSourceBuilder
                 .create()
--- a/aggregate-service/src/main/resources/application.yml
+++ b/aggregate-service/src/main/resources/application.yml
@@ -25,30 +25,6 @@ cloud:
       instance-profile: false
 
 app:
-  state:
-    code: CA
-    name: California
-
-  archive:
-      root: file:///tmp/
-
-  datamart:
-     olap_datasource:
-        url: "\
-           jdbc:redshift://${app.datamart.olap_datasource.url-server:localhost:5432}/${app.datamart.olap_datasource.url-db:dev}\
-           ?ApplicationName=rdw-reporting-aggregate-service\
-           "
-        username: root
-        password:
-        testWhileIdle: true
-        validationQuery: SELECT 1
-        driverClassName: com.amazon.redshift.jdbc42.Driver
-
-  cache:
-    repository:
-      # every day at 1:00 am
-      refresh-cron: "0 0 1 * * *"
-
   aggregate-reports:
     summative-enabled: true
     statewide-interim-enabled: false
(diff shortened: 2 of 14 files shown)


aggregate-service/src/main/resources/application.yml
```diff
@@ -25,61 +25,61 @@ cloud:
       instance-profile: false
```
malaffoon: really not sure why this is showing so many changes, maybe i alphabetized a few things: fundamentally i just moved the datasource section.

aggregate-service/src/main/resources/application.yml
```diff
@@ -25,61 +25,61 @@ cloud:
...
+    default-property-inclusion: non_null
+    serialization:
+      WRITE_DATES_AS_TIMESTAMPS: false
 
-          AggregateResponse:
-            content-type: application/json
-            group: default
-            producer:
-              requiredGroups: default
+  olap_datasource:
```
malaffoon: Since this is the single datasource for this app, i almost changed the name to just `datasource`. But i didn't.

webapp/src/main/resources/application.yml
```diff
@@ -61,27 +78,9 @@ permissionservice:
   endpoint: ${rdw.permission.service}
 
 app:
-  # todo use tenant.code & name
```
malaffoon: @esotrope, @wtritch i assume this comment is no longer valid: seems like we've settled on the whole ApplicationProperties thing.

webapp/src/test/resources/application-test.yml
```diff
@@ -4,16 +4,14 @@ app:
...
 
 permissionservice:
-  endpoint:
\ No newline at end of file
+  endpoint:
+
+spring:
+  datasource:
+    url-schema: reporting_test
+    maxWait: 10000
```
malaffoon: wasn't sure if the maxWait was put in here on purpose so i left it

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/DataSourceConfiguration.java
```diff
@@ -21,7 +21,7 @@
      * @return DataSource
      */
     @Bean(name = "olapDatasource")
-    @ConfigurationProperties(prefix = "app.datamart.olap_datasource")
+    @ConfigurationProperties(prefix = "spring.olap_datasource")
```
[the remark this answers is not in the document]
malaffoon: Mostly because this is the default behavior and we use it in most of our apps. We can probably remove the customized ConfigurationProperties in our apps because most of our apps now use a single datasource (which is probably a sign of better architecture).

It was interesting (amusing?): when i was investigating this that the **only** app in the reporting project that had the properties under `spring` instead of `app.datamart` was admin-service ... which is the only app that has multiple datasources. That was part of what pushed me to do this.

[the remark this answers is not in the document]
esotrope: gotcha - so why not just plug into `spring.datasource` instead of making our own un-autoconfigged prop?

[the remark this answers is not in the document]
malaffoon: Honestly just to minimize risk. I want to make the configuration consistent for the customer but i'm a bit worried about changing too much at this late stage.

PR #868, opened 2018-02-11
DWR-1406 NULL pointer exception trying to generate aggregate report

agorina:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
```diff
@@ -30,7 +30,7 @@ public ReportType getReportType() {
     public Subject getSubject() {
         return reportQuery.getSubjectCodes().size() != 1
                 ? null
-                : Subject.valueOf(Iterables.getOnlyElement(reportQuery.getSubjectCodes()));
+                : Subject.valueOf(Iterables.getOnlyElement(reportQuery.getSubjectCodes()).toUpperCase());
```
agorina: The issue is consistently reproducible with "Math" subject. The enum is "MATH".
Since most of the time both subjects are selected, the issue seems random.

I am not sure if this is the right solution. @wtritch  and @esotrope, what do you think?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
```diff
@@ -70,6 +70,7 @@ public Report generateReport(final AbstractExamReportRequest request, final User
     }
 
     private void failReport(final Report report) {
+        if (report == null) return;
```
agorina: The above failure cause `null` report being passed into an exception.

esotrope: do we know why a failReport got called with null? seems curious. failed report lookup like a 404?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
esotrope: feels like Subject.valueOf() should hold the case insensitivity logic maybe not sure if that is normal behavior of java enum though so maybe not

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
agorina: I think it fails inside of the repository when it is serializing a request as JSON. 

But it now made me think that it is unrelated to the above. So maybe there is another place where it needs to be checked.  I am still testing it locally. Will post an update later.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
agorina: Yeah, I see us doing `caseInsensitiveValue` in other places. I think it would be a better approach.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/SpringStreamReportGenerator.java
malaffoon: i see that this is the quickest fix but i think it leads to a misleading error message. Can we change it so a null report is messaged as `new RuntimeException("Unable to create report")`.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/web/converter/AggregateReportCsvHandlerTest.java
```diff
@@ -61,17 +61,16 @@
     private MockHttpOutputMessage outputMessage;
     private byte[] reportPayload;
     private Report report;
-    private ObjectMapper objectMapper;
     private AggregateReportCsvHandler handler;
 
     @Before
     public void setup() throws Exception {
-        objectMapper = new ObjectMapper();
+        final ObjectMapper objectMapper = new ObjectMapper();
```
agorina: Unrelated, fixed IntelliJ warning.

PR #869, opened 2018-02-11
Feature/aggregate report naming & scroll nav fixes

esotrope:
**Added report name section**
Right now the field is optional and it shows the default name as the placeholder.
If no name is entered the default name is used.
Probably need to add some constraints/validation like no whitespace-only names etc - thoughts on how we want to validate this?

image

**Sticky nav fixes:**
1. Nav now sticks to the top once you scroll to it/below its original position
image

2. Nav now does not move right when the user scrolls right
image


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
```diff
@@ -176,8 +178,7 @@ export class AggregateReportFormSettingsResolve implements Resolve<AggregateRepo
       settings.schools = saturatedRequest.schools;
     }
 
-    //TODO set name once entered by user
-    // settings.name = request.name + translationService.instant("copy");
+    settings.name = `${request.name} ${this.translate.instant('common.copy-suffix')}`;
```
esotrope: @wtritch I didn't see this working in action so I'm going to revisit it.
Also, i don't think the settings resolver is working as i'd expect.
When a user navigates to the page with `src=` the "remembered" settings of the page are used instead of the report settings i'm navigating to the page with.

malaffoon:
I'd think that a basic filename regex would be good. I know it doesn't have to be that restrictive but it seems like a good starting point.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -5,9 +5,10 @@ <h2 class="h3 gray-darkest">{{'aggregate-reports.form.heading' | translate}}</h2
   </div>
 </ng-container>
 
-<ng-container *ngIf="!accessDenied">
+<div *ngIf="!accessDenied">
 
-  <scroll-nav [items]="[
+  <div sticky marginTop="12" [style.position]="'absolute'">
```
esotrope: would put margin-top setting in the stylesheet but its actually needed in the calculations of the sticky directive. I think ill put position absolute in the stylesheet this PR

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -499,117 +527,118 @@ <h3 class="h3 gray-darkest">
...
+          options: options.assessmentTypes,
+          values: [ settings.assessmentType ]
+        }
+      ]
+    },
+    {
+      heading: 'aggregate-reports.form.summary.column.organization.heading' | translate,
+      label: 'aggregate-reports.form.summary.column.organization.label' | translate,
+      labelStyles: 'aqua',
+      attributeStyles: 'white-space-normal',
```
esotrope: fixes words drifting into other columns of the summary table

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.html
```diff
@@ -1,4 +1,4 @@
-<nav class="sidebar affix leftnav">
```
esotrope: delegates position: fixed to `sticky` directive

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -70,33 +66,39 @@ export class ScrollNavComponent {
...
     this.items
       .filter(item => item.scrollTo && document.getElementById(item.scrollTo.id))
       .forEach(item => {
-        const scrollItem = document.getElementById(item.scrollTo.id);
-        const itemOffsetTop = scrollItem.offsetTop;
-
-        // minus small number (20) since clicking on scroll nav sometimes resulted in the link above the one clicked
-        // being highlighted until scrolling down a bit
-        if ((itemOffsetTop - threshold) <= scrollTop) {
+        const itemOffsetTop = Utils.getAbsoluteOffsetTop(item.scrollTo) - Utils.getHeight(item.scrollTo);
```
esotrope: ditched the arbitrary threshold variable in favor of the element's height. I think this achieves what we were looking for.

webapp/src/main/webapp/src/app/shared/nav/sticky.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+const PositionFixed = 'fixed';
+
+/**
+ * Makes a decorated element stick to the top when the user scrolls past its vertical position
+ * Also, the elements original page x-position is maintained when scrolling horizontally
+ */
+@Directive({
+  selector: '[sticky]'
+})
+export class StickyDirective {
```
esotrope: The sticky directive.
There were a couple github repos with similar code out there but they did not look maintained and the first one i tried didn't even work so i felt i had to roll my own

webapp/src/main/webapp/src/app/shared/nav/sticky.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+    this._window = windowReferenceService.nativeWindow;
+  }
+
+  @Input()
+  set marginTop(value: number) {
+    this._marginTop = Utils.integerValueOf(value);
+  }
+
+  ngAfterViewInit(): void {
+    Utils.setImmediate(() => {
```
esotrope: ugly but i had to wait a frame after the component renders to get its actual position on the page. This may be more robust if i actually listen for document ready event, not sure

webapp/src/main/webapp/src/app/shared/nav/sticky.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+      style.left = pixelsOf(elementX - windowX);
+
+    } else {
+
+      // set position back to original settings
+      style.position = this._initialPosition;
+      style.top = pixelsOf(this._initialOffset.y);
+      style.left = pixelsOf(this._initialOffset.x);
+
+    }
```
esotrope: coding x/y positions brings me back to good times coding games in flash : j

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -5,31 +5,54 @@ import { AggregateReportFormSettings } from "./aggregate-report-form-settings";
...
+ * Form control validator that makes sure the organization control is not empty
+ * or includeStateResults or inlcudeAllDistricts is flagged
+ *
+ * @param properties the properties to propagate when the control value is invalid
+ * @return {null|{notEmpty: any}}}
+ */
+const organizationValidator = (properties, settings) => control => {
+  return settings.includeStateResults
+    || settings.includeAllDistricts
+    || control.value.length ? null : { invalid: properties };
```
esotrope: @agorina you can now submit the form if state, all districts or a single org is selected

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.html
```diff
@@ -46,21 +46,21 @@ <h3 class="h3 label-group green pull-left">
             <div class="pull-right">
               <span>
                 <label class="mr-xs">{{'common.performance-level-display-type-input-label' | translate}}</label>
-                <sb-toggle [options]="displayOptions.performanceLevelDisplayTypes"
+                <sb-radio-button-group [options]="displayOptions.performanceLevelDisplayTypes"
```
esotrope: changed to new alias as the component is more than a toggle now

webapp/src/main/webapp/src/app/shared/form/sb-checkbox-group.component.ts
```diff
@@ -84,6 +108,24 @@ export class SBCheckboxGroup extends AbstractControlValueAccessor<any[]> impleme
     return this._options;
   }
 
+  get buttonGroupStyles(): any {
+    return this._buttonGroupStyles;
+  }
+
+  @Input()
+  set buttonGroupStyles(value: any) {
+    this._buttonGroupStyles = value ? value : DefaultButtonGroupStyles;
```
esotrope: added style extensibility to checkbox and toggle group

webapp/src/main/webapp/src/app/shared/nav/sticky.directive.ts
```diff
@@ -0,0 +1,91 @@
...
+      // apply sticky styles
+      style.position = PositionFixed;
+      style.top = pixelsOf(this._marginTop);
+
+      // correct the position: fixed element's left offset
+      // without this the element will move along the x-axis when the page scrolls horizontally
+      if (scrollX > 0) {
+
+        // TODO fix choppyness
+        style.left = `${initialX - scrollX}px`;
```
esotrope: tad choppy but at least it doesnt go under or over the form now

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -3,33 +3,34 @@ import { Utils } from "./support/support";
...
-             [ngClass]="getButtonClassesInternal({active: option.value === value, disabled: disabled})">
+             [ngClass]="computeStylesInternal(buttonStyles, {
+                 active: option.value === value, 
+                 disabled: disabled
+             })">
         <input type="radio"
-               [id]="name"
-               [name]="name"
+               id="{{name}}"
+               name="{{name}}"
```
esotrope: changed a bunch of these where i could - this will make the framework have to do less work when rendering/checking state changes

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -5,31 +5,54 @@ import { AggregateReportFormSettings } from "./aggregate-report-form-settings";
...
 };
 
+/**
+ * Form control validator that makes sure the control value is a valid filename
+ *
+ * @param properties the properties to propagate when the control value is invalid
+ * @return {null|{fileName: any}}}
+ */
+const fileName = (properties: any) => control => {
+  return /^[^\\<>:;,?"*|/]*$/.test((control.value || '').trim()) ? null : { fileName: properties };
```
esotrope: added filename validation to report name (not sure if the backend sanitizes this stuff atm or not)

esotrope:
Pushing this through since the form validation is a bit of a testing holdup

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
wtritch: @esotrope I believe that's because we have `data: { canReuse: true },` in the form route.  Since we're no longer relying on the "back" button populating the query, we can probably remove that?

esotrope: hm. probably not a good user experience though if the user is coming from the form and the form state isn't preserved on the back button.

wtritch: ...not sure how we can have it both ways,  do you have specific use-cases for both options that are detectably different?

PR #871, opened 2018-02-12
Task/admin dropdown stylechecks

esotrope:
Fixed dropdown placement to be right aligned and drop out leftward onto the page.
Fixed arrow to always be down (it would not toggle back when you click out off of the button to close the dropdown)
image


webapp/src/main/webapp/src/app/home/admin-dropdown.component.html
```diff
@@ -1,20 +1,28 @@
-<div dropdown style="display: inline-block;" class="btn-group ml-xs">
```
esotrope: the change here was to remove the style tag, and move ml-xs to a wrapper span as its specific to the home page.

PR #873, opened 2018-02-12
Bugfix/handle exam export multi school permissions

wtritch:
This fixes the naming and grouping services used during organization exam exporting to handle the case when a user has permissions for multiple schools but not for the schools' school-groups or districts.


webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.ts
```diff
@@ -52,10 +52,14 @@ export class OrganizationExportNamingService {
...
       if (!Utils.isUndefined(selection.firstSchool.schoolGroupId)
+        && !Utils.isUndefined(selection.firstSchool.schoolGroup)
         && selection.schools.every(x => x.schoolGroupId === selection.firstSchool.schoolGroupId)) {
         return selection.firstSchool.schoolGroup.name;
       }
-      return selection.firstSchool.district.name;
+      if (!Utils.isUndefined(selection.firstSchool.district)) {
+        return selection.firstSchool.district.name;
+      }
+      return selection.firstSchool.name;
```
esotrope: revisiting this i think we should really just make the user name the thing. coding these smart names is pretty hairy.

webapp/src/main/webapp/src/app/organization-export/organization-grouping.service.ts
```diff
@@ -85,11 +86,23 @@ export class OrganizationGroupingService {
...
+  /**
+   * Test if this is a valid grouping node.  While building the organization tree we create shallow
+   * placeholders for districts and school groups that the user does not have explicit access to.  These
+   * shallow nodes should not be used for grouping.
+   *
+   * @param {Tree<Organization>} node An organizaiton node
+   * @returns {boolean} True if the node is a valid grouping node
+   */
+  private validGroupingNode(node: Tree<Organization>) {
+    return !isNullOrUndefined(node.value.name);
```
esotrope: is this null-safe?
I would expect:
```typescript
return !isNullOrUndefined(node.value) && !isNullOrUndefined(node.value.name);
```

webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.ts
wtritch: +1 but not today.  Heuristics always end poorly IMHO =)

webapp/src/main/webapp/src/app/organization-export/organization-grouping.service.ts
wtritch: yes, the value always exists (even if it's shallow and just has an "id" property)

esotrope: ok cools - thanks

webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.ts
esotrope: ill make a jira for it

PR #874, opened 2018-02-12
fixes assessment type display for aggregate reports

esotrope:
Before:
image

After:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report.service.ts
+++ b/webapp/src/main/webapp/src/app/report/report.service.ts
@@ -13,7 +13,7 @@ import { School } from "../user/model/school.model";
 import { Grade } from "../school-grade/grade.model";
 import { DataService } from "../shared/data/data.service";
 import { Download } from "../shared/data/download.model";
-import { AggregateReportRequest } from "./aggregate-report-request";
+import { AggregateReportRequest, AggregateReportQuery } from "./aggregate-report-request";
 import { AggregateReportRow } from "./aggregate-report";
 
 const ServiceRoute = '/report-processor';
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -180,7 +180,15 @@ export class ReportService {
     local.created = remote.created;
     local.reportType = remote.reportType;
     local.assessmentType = AssessmentType[ remote.assessmentType as string ];
-    local.assessmentTypeCode = remote.assessmentTypeCode;
+
+    // HOTFIX for aggreagte report assessment type display
+    // unable to use ExamReportAssessmentType enum because it does not support summatives
```
esotrope: I was going to add a delegating getAssessmentType() method in AggregateReportRequest.java similarly to how @wtritch added a `getSubject()` method there however the signature is `ExamReportAssessmentType` enum which doesn't support summatives - hence why i put this fix here for now.
I'm thinking we need to redesign those poor report requests that were made just for student pdf reports to begin with.
Since we always use type-conditional introspection i'm thinking all the report requests need to be like:
`{name: string, type: string, query: any}` and the endpoint or whatever can channel them based on type

thoughts?

wtritch: Unfortunately, since we're storing the requests as JSON in the database, we'll need to be very careful about being able to serialize older request instance versions if we ever restructure the request objects.

esotrope: well we expire them right? maybe we can phase out old requests for new request ones somehow?

PR #876, opened 2018-02-12
Update our ReportController report content method to advertize an app…

wtritch:
…lication/octet-stream response content-type to avoid Spring auto-changing all requests requesting to application/json.

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
@@ -22,6 +22,8 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Set;
 
+import static org.springframework.http.MediaType.APPLICATION_OCTET_STREAM_VALUE;
+
 @RestController
 public class ReportController {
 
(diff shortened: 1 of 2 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/ReportController.java
```diff
@@ -43,7 +45,10 @@
                 : service.getReports(user));
     }
 
-    @GetMapping("/reports/{reportId}")
+    @GetMapping(
+            path = "/reports/{reportId}",
+            produces = APPLICATION_OCTET_STREAM_VALUE
+    )
```
wtritch: The problem is that as a `@RestController` Spring was assuming this endpoint could return `application/json` so requests with an `Accept: */*` header were being treated by the underlying message writers as requests for `application/json` This modifies Spring to auto-convert request headers into a more generic MediaType.

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -112,6 +112,9 @@ export class ReportService {
    */
   public getReportContent(reportId: number): Observable<Download> {
     return this.dataService.get(`${ServiceRoute}/reports/${reportId}`, {
+      headers: new Headers({
+        'Accept': '*/*',
+      }),
```
wtritch: By default the angular client sends an Accept header that contains `application/json` as a media type.  This changes our report content download requests to accept any response content type (zip, pdf, csv, etc)

PR #880, opened 2018-02-13
summary view now displays entries when statewide or all districts fla…

esotrope:
…gs are on

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -128,7 +128,7 @@
           </div>
 
           <!-- Selections -->
-          <div [hidden]="!settings.districts.length || settings.includeAllDistricts">
+          <div [hidden]="!settings.districts.length">
             <label>{{'aggregate-reports.form.section.organization.district-list-heading' | translate}}</label>
             <aggregate-report-organization-list [organizations]="settings.districts"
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -92,15 +93,20 @@ export class AggregateReportTableDataService {
     return dimensions;
   }
 
-  private createSampleOrganizations(selectedOrganizationCount: number): Organization[] {
+  private createSampleOrganizations(settings: AggregateReportFormSettings): Organization[] {
+
+    const organizationCount = (settings.includeStateResults ? 1 : 0)
+      + (settings.includeAllDistricts ? MaximumOrganizations : 0)
+      + [ ...settings.districts, ...settings.schools ].length;
```
esotrope: new logic to decide how many orgs to display in the preview table

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -128,7 +128,7 @@ <h3 class="h3 gray-darkest">
           </div>
 
           <!-- Selections -->
-          <div [hidden]="!settings.districts.length || settings.includeAllDistricts">
```
esotrope: removed this - as it looks like a bug when this is on and you select a district

PR #882, opened 2018-02-13
Feature/auto select org on aggregate reports form

esotrope:
if you have only one district it will add that district
or if you have only one school it will add that school automatically


esotrope:
going to merge - still comment if you see anything weird

PR #886, opened 2018-02-13
prevent 400 on select all events

dsnyder-fairway:


webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
```diff
@@ -39,8 +39,10 @@ export class SchoolGradeComponent {
     }
   }
 
-  schoolChanged(value: School) {
-    this.school = value;
+  schoolChanged(value: any) {
+    if (value instanceof School) {
+      this.school = value;
```
esotrope: i bet this is because we are binding to a custom component's native (change) or (select)
I think the real fix is to make our own event names for custom components otherwise it will auto-bind/listen to nested elements' change and select events. This is what i ran into with the organization typeahead on the aggregate report form. Can you add a comment to this extent around the `instanceof` guard?

PR #889, opened 2018-02-13
Add max-width and full-text title popover text to popup menu options

wtritch:


webapp/src/main/webapp/src/app/shared/menu/popup-menu.component.ts
```diff
@@ -37,8 +38,9 @@ import { Utils } from "../support/support";
          triggers="{{ action.tooltip(item) == '' ? '': 'mouseenter:mouseleave'}}"
          placement="right"
          container="body"
+         [attr.title]="action.displayName( item )"
          (click)="action.isDisabled(item) || onMenuClick($event, action)">
-        <button class="btn btn-default btn-borderless"
+        <button class="btn btn-default btn-borderless label-max-width"
```
wtritch: I noticed the max-width style was missing here.  I don't think we want any popup menu options to display more than ~350 pixels wide (although I don't believe we have any atm.)

wtritch:
@esotrope this should play nicely with your existing PR against the same component

webapp/src/main/webapp/src/app/shared/menu/popup-menu.component.ts
esotrope: that would be a huge menu.

PR #891, opened 2018-02-13
spinner while loading instructional resources on admin page

dsnyder-fairway:


webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.component.ts
```diff
@@ -39,8 +40,9 @@ export class InstructionalResourceComponent implements OnInit {
 
   ngOnInit(): void {
     this.service.findAll()
-      .subscribe((resources) => {
-        this.resources = resources
+      .subscribe((resources: InstructionalResource[]) => {
+        this.resources = resources;
+        this.loading = false;
```
esotrope: you could just add a getter that delegates to resources.
```
get loading(): boolean {
  return Utils.isUndefined(this.resources);
}
```

webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.component.html
```diff
@@ -4,9 +4,13 @@
     <span class="btn btn-primary mr-xs" (click)="openCreateResourceModal()"><i class="fa fa-cloud-upload"></i> {{'prompt.create' | translate}}</span>
   </div>
 </div>
-
+<div *ngIf="loading" class="loader">
+  <div class="inner">
+    <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'messages.loading' | translate }}</span>
+  </div>
+</div>
```
esotrope: Use the new `<spinner text="{{'messages.loading' | translate}}"></spinner>`

esotrope: now that i think about it it might need to be extended to support variable fa-[size]

PR #894, opened 2018-02-13
fixes school groups text. fixes typeahead dropdown menu styles

esotrope:


webapp/src/main/webapp/src/styles.less
```diff
@@ -789,7 +789,7 @@ aggregate-report-organization-list {
   }
 }
 
-.dropdown-header, .dropdown-menu > li > a {
+:not(typeahead-container) > .dropdown-menu > li > a {
```
esotrope: fixes bug i introduced that messed with typeahead dropdown styles -> this makes it so that the ngx-bootstrap typeahead dropdowns do not get the new dropdown styles

PR #896, opened 2018-02-14
Item Viewer tab should indicate that student answer is displayed

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
```diff
@@ -12,10 +12,10 @@
...
-    <span *ngIf="!showResponse">
-      {{ 'labels.assessments.items.tabs.item-viewer-response-not-included' | translate }}
-    </span>
+    {{ 'labels.assessments.items.tabs.item-viewer-intro-text-prefix' | translate }}{{ (showResponse ?
+    'labels.assessments.items.tabs.item-viewer-response-included' :
+    'labels.assessments.items.tabs.item-viewer-response-not-included' ) | translate}}
   </p>
-  <iframe title="{{'labels.assessments.items.tabs.item-viewer' | translate }}" class="center" #irisframe src="/iris/" width="950" height="500" frameborder="0" ></iframe>
+  <iframe title="{{'labels.assessments.items.tabs.item-viewer' | translate }}" class="center" #irisframe src="/iris/"
+          width="950" height="500" frameborder="0"></iframe>
```
wtritch: nit: recommend breaking each attribute into a new line

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -983,8 +983,9 @@
             "view-ref": "View reference"
           },
           "item-viewer": "Item Viewer",
-          "item-viewer-intro-text": "This is the view of the item as seen while taking the assessment.  ",
-          "item-viewer-response-not-included": "This does not include student responses.",
+          "item-viewer-intro-text-prefix": "This is the view of the item as seen while taking the assessment",
+          "item-viewer-response-included": ", including the student's response.",
+          "item-viewer-response-not-included": ".  This does not include student responses.",
```
wtritch: I'd recommend breaking this into two full statements rather than a prefix + suffix.

esotrope: confused by why we have a before and after message. i would expect something like an interpolated message.
Mind attaching before/after screenshots?

dsnyder-fairway: 👍 sounds good

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
esotrope: that should be our angular template standard moving forward i think. i've been trying to move everything in that direction anyway

webapp/src/main/webapp/src/assets/i18n/en.json
dsnyder-fairway: Before included the text:
`This is the view of the item as seen while taking the assessment.` when an answer was given on the item viewer and `This is the view of the item as seen while taking the assessment.  This does not include student responses.` when no answer was given

esotrope: Ah yeah - however redundant i would have the prefix in both of them. because if translated we wouldn't want that awkward english language dependent limitation of having a two-part message.

malaffoon: I don't think we need to explicitly say "This does not include student responses." That is kind of assumed.

webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
dsnyder-fairway: done

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -983,9 +983,8 @@
             "view-ref": "View reference"
           },
           "item-viewer": "Item Viewer",
-          "item-viewer-intro-text-prefix": "This is the view of the item as seen while taking the assessment",
-          "item-viewer-response-included": ", including the student's response.",
-          "item-viewer-response-not-included": ".  This does not include student responses.",
+          "item-viewer-response-included": "This is the view of the item as seen while taking the assessment, including the student's response.",
+          "item-viewer-response-not-included": "This is the view of the item as seen while taking the assessment.  This does not include student responses.",
```
malaffoon: If we keep the second sentence, `responses` should be singular, `response`. Yes, a single item may have multiple selections but they are all part of a single response.

dsnyder-fairway: I think this might make sense in this case. The text is displayed when clicking on the `Item Viewer` from the following screen. There are multiple students here
screen shot 2018-02-13 at 5 03 40 pm

dsnyder-fairway: The `response` text is when clicking on the Student and viewing their item response

PR #898, opened 2018-02-14
Task/instructional resources resource field text

esotrope:
Moved the warning text and gave it more margin
image

Added resource URL validation messages to form
image

image

Talking with Will - the whole form needs this kind of validation.

Made table URL an actual link.
Made well container extend with the table body like our other pages
image


webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
```diff
@@ -80,17 +80,27 @@ <h4 class="modal-title pull-left">{{ 'labels.instructional-resource.create.title
             </select>
           </div>
         </div>
-        <div class="form-group row mb-xs">
+        <div class="form-group row mb-xs"
+             [ngClass]="{'has-error': resource.invalid && (resource.dirty || resource.touched)}">
```
esotrope: we really need this logic on all of our form-groups to make the inputs highlight red

wtritch:
That was quick

PR #900, opened 2018-02-14
Improve instructional resource display per customer feedback

wtritch:
Added a border between SBAC instructional resources and user-entered instructional resources if both exist:
image

Re-worded the "not found" instructional resource message:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/popover/instructional-resource-popover.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/popover/instructional-resource-popover.component.html
@@ -6,7 +6,7 @@
 <div *ngIf="!loading">
   <div *ngIf="resources && resources.length; else noResources">
     <div *ngFor="let instructionalResource of resources"
-         [ngClass]="{'border-bottom': instructionalResource.organizationLevel == 'System' && resources.length > 1}"
+         [ngClass]="{'border-bottom mb-xs': instructionalResource.organizationLevel == 'System' && resources.length > 1}"
          class="popover-link">
       <a href="{{instructionalResource.url}}" target="_blank">{{ ('labels.instructional-resources.link.' + instructionalResource.organizationLevel) | translate:instructionalResource }}</a>
     </div>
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1351,7 +1351,7 @@
             "target": "Target {{target}}",
             "title": "Results By Item"
           },
-          "no-instruct-found": "No resources found.",
+          "no-instruct-found": "Resources Not Yet Available",
```
esotrope: do we really need all caps?

webapp/src/main/webapp/src/app/assessments/popover/instructional-resource-popover.component.html
```diff
@@ -5,10 +5,11 @@
 </div>
 <div *ngIf="!loading">
   <div *ngIf="resources && resources.length; else noResources">
-    <p class="popover-link"
-       *ngFor="let instructionalResource of resources">
+    <div *ngFor="let instructionalResource of resources"
+         [ngClass]="{'border-bottom': instructionalResource.organizationLevel == 'System' && resources.length > 1}"
```
esotrope: i would suggest the border be a lighter gray and have more vertical margin

webapp/src/main/webapp/src/assets/i18n/en.json
wtritch: changed to initial caps

webapp/src/main/webapp/src/app/assessments/popover/instructional-resource-popover.component.html
wtritch: Hm... the padding does look a little off.  I'd really rather not create a custom border color for this component, however.  It's one more thing to maintain in the case of a style update.

wtritch: image

PR #903, opened 2018-02-14
correct toggling after user clicks on select option for embargo

dsnyder-fairway:
after clicking embargo they see the following:
screen shot 2018-02-14 at 9 24 27 am

once the user confirms, embargo is selected

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.ts
@@ -50,12 +50,13 @@ export class EmbargoTable {
   private toggleInternal(toggle: Toggle, embargo: Embargo, scope: EmbargoScope, embargoEnabled: boolean, overridingEmbargoEnabled: boolean): void {
     const value = toggle.value;
 
-    // Fixes display by undoing visual state change to toggle when clicked.
-    // It was not sufficient to prevent event propagation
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.ts
```diff
@@ -65,6 +65,8 @@ export class EmbargoTable {
         overridingEmbargo: this.overridingEmbargo,
         overridingEmbargoEnabled: overridingEmbargoEnabled
       });
+    } else {
+      toggle.value = toggle.previousValue;
```
dsnyder-fairway: looks like all I really need is `toggle.value = !toggle.value` here

webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.ts
```diff
@@ -65,6 +65,8 @@ export class EmbargoTable {
         overridingEmbargo: this.overridingEmbargo,
         overridingEmbargoEnabled: overridingEmbargoEnabled
       });
+    } else {
+      toggle.value = !toggle.value;
```
esotrope: Hm i'm not sure this will always work. i think we might need to put the original `toggle.value = !value` within the if clause

dsnyder-fairway: its already set to ` toggle.value = !value;` on line 56

PR #904, opened 2018-02-14
Remove duplicate .csv in aggregate report export filename

wtritch:


wtritch:
Merging now, will resolve comments as they come in.

PR #905, opened 2018-02-14
fixes ellipsis text align

esotrope:
Before

image

After

image


webapp/src/main/webapp/src/app/report/report-action.component.html
```diff
@@ -1,12 +1,21 @@
...
+     aria-hidden="true"></i>
+  {{report.label}}
 </button>
 
 <!-- Multiple Actions -->
 <popup-menu *ngIf="reportActions.length > 1"
-            [item]="report" [actions]="menuActions"
+            [item]="report"
+            [actions]="menuActions"
             [text]="report.label"></popup-menu>
```
esotrope: new-line attr formatting only

PR #907, opened 2018-02-14
IE11 fix student score distribition display

wtritch:


webapp/src/main/webapp/src/styles.less
```diff
@@ -422,12 +422,28 @@ body .select-menu-group {
   }
 
   .score-distribution-container {
+    width: 100%;
+    display: -ms-grid;
     display: grid;
+    -ms-grid-columns: 40% 13% 1fr;
     grid-template-columns: 40% 13% 1fr;
```
wtritch: IE11 supports CSS Grid display, but it requires prefixing, as well as explicit column (and row) identifying below

wtritch:
@esotrope yeah... but bootstrap columns are pretty chunky in terms of available widths.  You have either one column, or two, or ... up to twelve.  CSS grid allows much more arbitrary sizing.

PR #908, opened 2018-02-14
Clean up home page links when screen is at a specific width

wtritch:
When slowly shrinking browser width on the home page, the links are temporarily in an awkward view state:
home-page-links-small-screen-before

This fix uses flex layout to better position the links when the view is too narrow to display them all side-by-side:
home-page-links-small-screen-after


malaffoon:
Thanks Will.

PR #909, opened 2018-02-14
Defect/update query admin conditions not reflecting query

esotrope:
fixes DWR-1434 and more

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.spec.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.spec.ts
@@ -134,7 +134,7 @@ describe('AggregateReportFormSettingsResolve', () => {
   function mockReport(): Report {
     const query: AggregateReportQuery = {
       achievementLevelDisplayType: 'Separate',
-      administrationConditionCodes: ['Valid','SD'],
+      administrativeConditionCodes: ['Valid','SD'],
       assessmentGradeCodes: ['03', '04'],
       assessmentTypeCode: 'ica',
       completenessCodes: ['Complete'],
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
@@ -40,7 +40,7 @@ export class AggregateReportFormSettingsResolve implements Resolve<AggregateRepo
 
     //Fetch the report, lookup schools/districts, map to settings
     return this.service.getReportById(Number.parseInt(reportId))
-      .flatMap(report => this.saturateOrgs(report.request))
+      .flatMap(report => this.addOrganizations(report.request))
       .map(saturatedRequest => {
         this.mapOntoSettings(saturatedRequest, options, settings);
         return settings;
@@ -81,7 +81,7 @@ export class AggregateReportFormSettingsResolve implements Resolve<AggregateRepo
     };
   }
 
-  private saturateOrgs(request: AggregateReportRequest): Observable<SaturatedRequest> {
+  private addOrganizations(request: AggregateReportRequest): Observable<SaturatedRequest> {
     const schoolIds: number[] = request.reportQuery.schoolIds;
     const schools: Observable<School[]> = !Utils.isNullOrEmpty(schoolIds)
       ? this.organizationService.getOrganizationsByIdAndType(OrganizationType.School, schoolIds)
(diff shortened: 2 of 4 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.spec.ts
```diff
@@ -134,7 +134,7 @@ describe('AggregateReportFormSettingsResolve', () => {
   function mockReport(): Report {
     const query: AggregateReportQuery = {
       achievementLevelDisplayType: 'Separate',
-      administrationConditionCodes: ['Valid','SD'],
+      administrativeConditionCodes: ['Valid','SD'],
```
esotrope: ion -> ive

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
```diff
@@ -115,60 +116,54 @@ export class AggregateReportFormSettingsResolve implements Resolve<AggregateRepo
...
     settings.schoolYears = query.schoolYears;
     settings.subjects = query.subjectCodes;
+    settings.valueDisplayType = query.valueDisplayType;
 
-    if (query.achievementLevelDisplayType) {
-      settings.performanceLevelDisplayType = query.achievementLevelDisplayType;
-    }
+    settings.completenesses = Utils.isNullOrEmpty(query.completenessCodes)
+      ? valuesOf(options.completenesses)
+      : query.completenessCodes;
```
esotrope: optional fields will not be in the request query when all are selected so we need to assume that if they are not present in the query that the user selected all of them

webapp/src/main/webapp/src/app/report/report-action.component.html
```diff
@@ -18,4 +18,4 @@
 <popup-menu *ngIf="reportActions.length > 1"
             [item]="report"
             [actions]="menuActions"
-            [text]="report.label"></popup-menu>
+            text="{{report.label}}"></popup-menu>
```
esotrope: unrelated angular optimization.

basically all `string` props that are set once we don't need the `[prop]="val"` syntax we can use `prop="{{val}}"`

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
wtritch: This is modifying the "default" settings created by:
const settings: AggregateReportFormSettings = this.createDefaultSettings(options);
which *should* already have "all" selected, shouldn't it?  We should be able to just override if the values exist?

esotrope: All is not the default for all optionals (e.g. admin conditions and completeness)

wtritch: I guess my question is:
1. The settings contains default values before coming into this method
2. We were overriding the default values only if the request had values (non-default only?)
3. Did that not work for some reason?

Not recommending a change or anything, just trying to understand where the logic broke down before.

esotrope: So it works when the default setting is All.
It does not work otherwise - because - how we optimize report query mapping on the front end. We remove the field from the query if All of the options are selected. See AggregateReportRequestMapper.

We could avoid this complication/confusion by doing the query "simplification" on the backend.
Which is what I'm thinking we should probably be doing. The backend should accept a sub-optimal query and simplify/optimize it for its own performance since backend performance is a backend concern right?

@agorina thoughts?

agorina: I am not sure I follow your idea. Just to make sure I understand it:

There is a query, and there are 'default' query values for UI. 'All' is just a visually way to present user's input : 'do not care about this field'. 

So what are the 'sub-optimal query' and 'optimized'? Are you saying that suboptimal would be - 'All' =' list all values' and then back end will detect 'all' and remove them?

esotrope: Yes, when ALL is selected for an optional filter field on the UI it represents that as an array of all the options.

As discussed, right now the front end "simplifies" the report request before sending it to the backend by removing filter fields from the request if ALL options of that specific field are selected.
This is because we want to ignore the filtering when ALL options of a filter are selected - which makes sense - but i don't think it should be a concern of the front end.

Ideally the backend performance shouldn't hinge on whether the UI sends ALL options of a field or no field in the request.
I think if we put the request "simplification" logic on the backend we can relieve the UI of having to intimately know how to optimize its request for backend performance.

PR #912, opened 2018-02-14
 avoid having footer be broken between pages

dsnyder-fairway:
and if accommodations is displayed, page break afterwards


dsnyder-fairway:
I was except for the accommodations since I am not entirely sure how those get wired in. But from looking at the pdf attached to the ticket this should work out okay for that too

PR #913, opened 2018-02-15
Add an option to exclude transfer access exams to the exam export dia…

wtritch:
…log.

The back-end and queries were all wired for optional transfer access on a per-query basis.  We just forgot to add the toggle to the UI.

When transfer-access is enabled via back-end properties:
image

When transfer-access is not enabled via back-end properties: (no change)
image


malaffoon:
Can we force the info button to stay with the title text?

esotrope:
We'll need to figure out where to put the name field too : /
It would be really nice if both aggregate and school/dist export shared the same org selector too.

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
```diff
@@ -33,7 +33,28 @@ <h3 class="blue mb-md">{{'labels.organization-export.title' | translate}}</h3>
...
+                           {value: true, text: 'buttons.visibility.hide' | translate}
+                         ]">
+            </sb-radio-button-group>
+          </div>
+        </div>
+        <div class="text-right" [ngClass]="{
+            'col-md-2': !editingDisabled && transferAccess,
+            'col-md-4': !editingDisabled && !transferAccess,
+            'col-md-8': editingDisabled && transferAccess,
+            'col-md-10': editingDisabled && !transferAccess
```
esotrope: what is `editingDisabled`? i don't see that defined

webapp/src/main/webapp/src/app/organization-export/organization-export.service.ts
```diff
@@ -23,16 +23,18 @@ export class OrganizationExportService {
...
+    return this.dataService.post(`${ServiceRoute}/exams/export`, this.createExportRequest(schoolYear, schools, disableTransferAccess, organizations))
   }
 
-  private createExportRequest(schoolYear: number, schools: Organization[], organizations: UserOrganizations): OrganizationExportRequest {
+  private createExportRequest(schoolYear: number, schools: Organization[], disableTransferAccess: boolean, organizations: UserOrganizations): OrganizationExportRequest {
     let ids = this.groupingService.groupSelectedOrganizationIdsByType(schools, organizations);
     let options = Object.assign({ schoolYear: schoolYear }, ids);
     let name = this.namingService.name(options, organizations);
+    options = Object.assign({ disableTransferAccess: disableTransferAccess}, options);
     return Object.assign({ name: name }, options);
```
esotrope: Not sure why we are so sequential here.
We could just do this:
```
return Object.assign(ids, {
  schoolYear: schoolYear,
  name: this.namingService.name(...)
  disableTransferAccess: disableTransferAccess
})
```

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
wtritch: I have no idea.  It was here when I got here.  Looks like it's related to hiding the organization selector... is/was this used in a place where the user cannot choose an organization?

webapp/src/main/webapp/src/app/organization-export/organization-export.service.ts
wtritch: Yeah... I think I'm going to change this pattern.

wtritch:
@malaffoon Not really.  This is an unfortunate side effect of using bootstrap gridding.  There isn't enough space to bump the column span from 2 to 3 columns.
Alternatively, I could drop the input to a new row and give it 3 columns.  Thoughts?

esotrope:
@wtritch we should play with where we put them - we'll also need to make way for the the name field we'll introduce too.

malaffoon:
Can we change the verbiage to keep that info button with it? "Transfers (i)"

wtritch:
@malaffoon I know the app is english-centric, but messing with translations for layout just feels like kicking the can.  I'm tempted to drop to a second level that can also host the upcoming naming text input field.

wtritch:
hmmm... dropping it to a new line creates a visual break between the organization entry field and the current organizations list... I'll play with it.  I want the organization entry to be right above the organization list.

wtritch:
image
Thoughts?

esotrope:
Looking good 👍 I would make school year a col-3 and put the export button in a col-3 right of it though to get rid of the extra whitspace.

wtritch:
hmm... looks allright:
image

malaffoon:
It seems like searching/adding organizations, school year and transfer assessments should all be near each other; while report name and export results button should be more associated

wtritch:
@malaffoon Except that I want to organization search and selected organizations visually close together, which means the organization search needs to be at the bottom.  And in (nearly) all our other forms the submit button is in the bottom-right, so I don't want to shift it around...

wtritch:
Here's the final view complete with the default report name as a placeholder:
image

webapp/src/main/webapp/src/app/organization-export/organization-export-naming.service.spec.ts
```diff
@@ -1,121 +0,0 @@
-import { OrganizationMapper } from "./organization/organization.mapper";
-import { OrganizationExportNamingService } from "./organization-export-naming.service";
-
-describe('OrganizationExportNamingService', () => {
```
wtritch: I removed the heuristic-based naming service in favor of providing a default report name with the option for the user to name their export however they like.

webapp/src/main/webapp/src/app/organization-export/organization-export.component.ts
```diff
@@ -52,34 +41,39 @@ export class OrganizationExportComponent implements OnInit {
    */
   private _schoolYearOptions: number[];
 
-  /**
-   * Currently selected school year
-   */
-  private _selectedSchoolYear: number;
```
wtritch: Removed the separate properties in favor of a single OrganizationExport object that gets passed to the service to trigger the export.

wtritch:
Merging for now, will address any additional comments as they arrive.

PR #914, opened 2018-02-15
reset the routes when user re-authenticates

dsnyder-fairway:
so far on testing this locally (certainly takes a while to verify) this works out. However, I did not manage to verify the issue appears without my changes since the period takes so long (set a timer for 20 minutes and refreshed the page...still logged in 😒. Only was able to verify the issue did not appear with these changes while returning from dinner

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/security/session-expired.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/security/session-expired.component.ts
@@ -1,5 +1,7 @@
 import { Component } from "@angular/core";
 import { AuthenticationService } from "./authentication.service";
+import { Router } from "@angular/router";
+import { routes } from "../../app.routes";
 
 @Component({
   selector: 'session-expired',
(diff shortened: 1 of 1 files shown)


dsnyder-fairway:
note

```
WARNING in Circular dependency detected:
src/app/app.routes.ts -> src/app/shared/security/session-expired.component.ts -> src/app/app.routes.ts

WARNING in Circular dependency detected:
src/app/shared/security/session-expired.component.ts -> src/app/app.routes.ts -> src/app/shared/security/session-expired.component.ts
```

dsnyder-fairway:
closing because of circular dependency

esotrope:
Is dwr-1240 even still an issue? i thought i introduced that when i joined admin and webapp en.json and we fixed it. also if you want to speed up the session timeout you can change the timeout duration start the webapp with java arg: `-Dserver.session.timeout=5000` (5s)

PR #915, opened 2018-02-15
POC: Changes to repo/permissions - more changes

agorina:
I think I would like to merge this one, unless somebody objects. Any nitpicking is welcomed :) 

@wtritch , @malaffoon , @esotrope - Please take another look.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
@@ -4,16 +4,21 @@ import org.opentestsystem.rdw.common.model.AssessmentType;
 import org.opentestsystem.rdw.reporting.common.model.Exam;
 import org.opentestsystem.rdw.reporting.common.model.ScaleScore;
 import org.opentestsystem.rdw.reporting.common.model.School;
+import org.opentestsystem.rdw.security.Permission;
 import org.opentestsystem.rdw.security.PermissionScope;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Lists.newArrayList;
 import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
 import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.safeSplit;
+import static org.opentestsystem.rdw.reporting.common.security.ReportingPermission.GroupPiiRead;
+import static org.opentestsystem.rdw.reporting.common.security.ReportingPermission.IndividualPiiRead;
 
 /**
  * Holds utility methods for mapping SQL query results to {@link Exam}
@@ -58,14 +63,44 @@ public final class Exams {
      * @throws SQLException
      */
     public static <A extends Exam.Builder> A map(final ResultSet row, final A builder, final PermissionScope permissionScope) throws SQLException {
+        final School school = buildSchool(row);
+        return (A) map(row, builder.school(school), !hasSchoolAccess(school, permissionScope));
+    }
 
-        final School school = School.builder()
+    /**
+     * Populate an Exam.Builder from a database result set.
+     * NOTE this will determine if an Exam is designated as a "Transfer Exam" based upon the given
+     * permissions.  If the user does not natively have access to the exam's school, it is designated
+     * as a transfer exam.
+     *
+     * @param row         The database result set
+     * @param builder     The builder to populate
+     * @param permissions The map of user's permissions
+     * @param <A>         The exam builder concrete class
+     * @return The populated exam builder
+     * @throws SQLException
+     */
+    public static <A extends Exam.Builder> A map(final ResultSet row, final A builder, final Map<String, Permission> permissions) throws SQLException {
+
+        final Permission individualPii = permissions.get(IndividualPiiRead);
+        final Permission groupPii = permissions.get(GroupPiiRead);
+        checkArgument(individualPii != null || groupPii != null, "valid permission scope is required");
+
+        final School school = buildSchool(row);
+        return (A) map(row, builder.school(school), individualPii != null ? !hasSchoolAccess(school, individualPii.getScope()) : !hasSchoolAccess(school, groupPii.getScope()));
+    }
+
+    private static School buildSchool(final ResultSet row) throws SQLException {
+        return School.builder()
                 .id(row.getLong("school_id"))
                 .name(row.getString("school_name"))
                 .districtId(row.getLong("school_district_id"))
                 .districtGroupId(getNullable(row, row.getLong("school_district_group_id")))
                 .schoolGroupId(getNullable(row, row.getLong("school_group_id")))
                 .build();
+    }
+
+    private static <A extends Exam.Builder> A map(final ResultSet row, final A builder, final boolean transfer) throws SQLException {
 
         builder.id(row.getLong("id"))
                 .sessionId(row.getString("session_id"))
@@ -76,8 +111,7 @@ public final class Exams {
                 .schoolYear(row.getInt("school_year"))
                 .accommodationCodes(safeSplit(row.getString("available_accommodation_codes"), "|"))
                 .scaleScore(ScaleScores.map(row))
-                .school(school)
-                .transfer(!hasSchoolAccess(school, permissionScope));
+                .transfer(transfer);
 
         // Collects claim scores for ICA and Summative exams.
         // If a single claim is unscored a null entry will be put in the list to represent the unscored claim
@@ -139,5 +173,4 @@ public final class Exams {
         }
 
     }
-
 }
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
@@ -2,6 +2,7 @@ package org.opentestsystem.rdw.reporting.common.jdbc;
 
 import com.google.common.collect.ImmutableMap;
 import org.opentestsystem.rdw.reporting.common.configuration.TenantProperties;
+import org.opentestsystem.rdw.reporting.common.security.PermissionSource;
 import org.opentestsystem.rdw.security.Permission;
 import org.opentestsystem.rdw.security.PermissionScope;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -48,14 +49,13 @@ public class SecurityParameterProvider {
     /**
      * Creates map of SQL query parameters for the given user name and permissions based on the defined permission prefixes
      *
-     * @param userName        the name of the user
-     * @param permissionsById the security information to prepare the SQL query parameters for
+     * @param permissionSource the security information to prepare the SQL query parameters for
      * @return map of SQL query parameters for the given permission scope
      */
 
-    public Map<String, Object> getSecurityParameters(final String userName, final Map<String, Permission> permissionsById) {
-        return getSecurityParametersBuilder(permissionsById)
-                .put("user_login", userName)
+    public Map<String, Object> getSecurityParameters(final PermissionSource permissionSource) {
+        return getSecurityParametersBuilder(permissionSource.getPermissionsById())
+                .put("user_login", permissionSource.getUsername())
                 .build();
     }
 
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
@@ -42,20 +42,20 @@ public class SecurityParameterProvider {
      * @return map of SQL query parameters for the given permission scope
      */
     public Map<String, Object> getSecurityParameters(final Map<String, Permission> permissionsById) {
-        final ImmutableMap.Builder<String, Object> securityParameters = ImmutableMap.builder();
+        return getSecurityParametersBuilder(permissionsById).build();
+    }
 
-        PERMISSION_PREFIXES
-                .forEach((key, value) -> {
-                    final Permission permission = permissionsById.get(key);
-                    final PermissionScope scope = permission == null
-                            ? PermissionScope.EMPTY
-                            : permission.getScope();
+    /**
+     * Creates map of SQL query parameters for the given user name and permissions based on the defined permission prefixes
+     *
+     * @param userName        the name of the user
+     * @param permissionsById the security information to prepare the SQL query parameters for
+     * @return map of SQL query parameters for the given permission scope
+     */
 
-                    securityParameters.putAll(getSecurityParameters(scope, value));
-                });
-
-        return securityParameters
-                .put("allow_transfer_access", tenantProperties.isTransferAccessEnabled())
+    public Map<String, Object> getSecurityParameters(final String userName, final Map<String, Permission> permissionsById) {
+        return getSecurityParametersBuilder(permissionsById)
+                .put("user_login", userName)
                 .build();
     }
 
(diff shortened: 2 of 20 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/JdbcStudentExamRepository.java
```diff
@@ -44,6 +48,51 @@
         this.securityParameterProvider = securityParameterProvider;
     }
 
+    @Override
+    public List<StudentHistoryExamWrapper> findAllForStudent(@NotNull final User user, final long studentId) {
```
agorina: This is pretty much a copy of the another method, but is takes a User as an input and used a new SQL.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -16,6 +16,24 @@ sql:
...
+      SELECT
+        e.iep,
+        e.lep,
+        e.section504,
+        e.economic_disadvantage,
+        e.migrant_status,
+        ${sql.snippet.claimColumns},
+        ${sql.snippet.exam}
+      WHERE e.student_id = :student_id AND
+       (  1 = :individual_access OR
```
agorina: I need either this flag or two diff. queries for group PII or just PII. Thoughts?

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -16,6 +16,24 @@ sql:
       where e.student_id=:student_id
         and ${sql.reporting.snippet.unknownScopePermission}
 
+    findAllByStudentIdTest: >-
```
agorina: Just to explain : this finds all student's exams. 
Regardless how user got to view this student's history, we are should all exams.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -351,6 +369,37 @@ sql:
               and (1=:statewide or isch.district_group_id in (:district_group_ids) or isch.district_id in (:district_ids) or isch.school_group_id in (:school_group_ids) or isch.id in (:school_ids))
             )
           )
+
+      scopePermission: >-
```
agorina: This is intended to be 'blindly' added to every SQL that needs permission protection.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/student/JdbcStudentExamRepositoryIT.java
```diff
@@ -61,6 +69,8 @@ public void findAllForStudentShouldReturnNullIfStudentIsNotInGroup() throws Exce
     public void findAllForStudentShouldReturnIfStudentIsInGroup() throws Exception {
         assertThat(repository.findAllForStudent(permissions(groupOf(PermissionScope.STATEWIDE)), ImmutableSet.of(-10L), -1L))
                 .isNotEmpty();
+
+        assertThat(repository.findAllForStudent(userBuilder.permissionsById(permissions(groupOf(PermissionScope.STATEWIDE))).build(), 0L)).isEmpty();
```
agorina: I added a few tests, but I have to say that this change will require miles and miles of new tests. I do not think the test coverage for the repos is extensive enough.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/JdbcStudentExamRepository.java
```diff
@@ -44,6 +48,51 @@
...
+                : permissions.get(IndividualPiiRead).getScope();
+
+        final PermissionScope groupPii = permissions.get(GroupPiiRead) == null
+                ? PermissionScope.EMPTY
+                : permissions.get(GroupPiiRead).getScope();
+
+        return template.query(
+                findAllByStudentIdTestQuery,
+                ImmutableMap.<String, Object>builder()
+                        .putAll(securityParameterProvider.getSecurityParameters(user.getPermissionsById()))
```
malaffoon: nit: last param should be just `permissions`

agorina: There are a lot of 'nit's in here, that was mentioned above. 
I just want the feedback on the approach.

reporting-service/src/main/resources/application.sql.yml
malaffoon: I'm confused. It seems like we have permissions activity in two places. Yeah, i sorta see what is going on but it seems redundant to have to check the group membership twice.

wtritch: Hmm...  Assumed the SQL would just be "get me all student's exams"  And the security portion would be the standard "exam-security-sql" used anywhere we gate exam access by user permissions?  Why do we need this special group checking?

reporting-service/src/main/resources/application.sql.yml
wtritch: So this one gates on "school s" (bound to the request?) and "school isch" (bound to the exam's school in the case of transfer exams).  Are we going to need others bound to more abstract concepts or must every query bind down to the school level if it requires permissions?

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -16,6 +16,24 @@ sql:
...
+        ${sql.snippet.claimColumns},
+        ${sql.snippet.exam}
+      WHERE e.student_id = :student_id AND
+       (  1 = :individual_access OR
+        EXISTS(
+          SELECT 1 FROM student_group_membership gm JOIN user_student_group usg ON usg.student_group_id = gm.student_group_id
+                     WHERE gm.student_id = e.student_id AND usg.user_login = :user_login
+              )
+       )
+       AND ${sql.reporting.snippet.scopePermission}
```
wtritch: Think we may need more generic SQL structuring like we have in the aggregate-service...?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/JdbcStudentExamRepository.java
esotrope: I think he means pass the permission map to the repo instead of the user.

reporting-service/src/main/resources/application.sql.yml
agorina: My thought was to keep it to the school level if it requires permissions. It does not mean that this is right, but I do not see a better option? Open to suggestions.

reporting-service/src/main/resources/application.sql.yml
agorina: Yes, yes - you are both correct. Fixed. Thank you.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/student/JdbcStudentExamRepositoryIT.java
esotrope: isn't the goal to test the snippet somewhat independently? is it possible to test the snippet for what it is responsible for and the queries using the snippet for what they are responsible for?

agorina: Good idea, but so far I have not found a way to do it :)  
Once I fully update this test it may lead to an idea...Let me know if you have some thoughts about it too.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/JdbcStudentExamRepository.java
agorina: I need a user id in the permission SQL to find the groups.

esotrope: Hm. Should we make a nice sub-interface for repos to use that only exposes the user props/methods that the repo needs?
(e.g.)
```java
interface PermissionSource {

  String getId();

  // returns PermissionScope.EMPTY when a miss
  PermissionScope getPermissionScopeByPermissionId(String id);

  // Also returns PermissionScope.EMPTY when an absent key is looked up
  Map<String, PermissionScope> getPermissionScopesByPermissionId();

}
```

agorina: Maybe, later, at some point. How strongly do you feel about it? 

 I really would like to focus on the main objective for now. This should be an easy refactoring later (I think).

esotrope: Yes - the suggestion is for later when you implement not for the POC

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
```diff
@@ -58,14 +63,44 @@ private Exams() {
...
+     */
+    public static <A extends Exam.Builder> A map(final ResultSet row, final A builder, final Map<String, Permission> permissions) throws SQLException {
+
+        final Permission individualPii = permissions.get(IndividualPiiRead);
+        final Permission groupPii = permissions.get(GroupPiiRead);
+        checkArgument(individualPii != null || groupPii != null, "valid permission scope is required");
+
+        final School school = buildSchool(row);
+        return (A) map(row, builder.school(school), individualPii != null ? !hasSchoolAccess(school, individualPii.getScope()) : !hasSchoolAccess(school, groupPii.getScope()));
+    }
```
agorina: Moved this logic from the repo to here.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
```diff
@@ -42,20 +42,20 @@ public SecurityParameterProvider(final TenantProperties tenantProperties) {
...
+     * Creates map of SQL query parameters for the given user name and permissions based on the defined permission prefixes
+     *
+     * @param userName        the name of the user
+     * @param permissionsById the security information to prepare the SQL query parameters for
+     * @return map of SQL query parameters for the given permission scope
+     */
 
-        return securityParameters
-                .put("allow_transfer_access", tenantProperties.isTransferAccessEnabled())
+    public Map<String, Object> getSecurityParameters(final String userName, final Map<String, Permission> permissionsById) {
```
agorina: I would like to pass in a `User` object, but there are dependency issues between modules. 

Maybe something to refactor later?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -35,12 +35,12 @@
...
     @Override
     public String export(final PermissionScope permissionScope, final ExportExamReportRequest request, final String uriLocation) {
-        final boolean isDestinationS3 =  uriLocation.toLowerCase().startsWith("s3");
+        final boolean isDestinationS3 = uriLocation.toLowerCase().startsWith("s3");
 
         template.execute(
                 exportExam.replace(ExportDestination, isDestinationS3 ? "S3" : ""),
                 ImmutableMap.<String, Object>builder()
-                        .putAll(securityParameterProvider.getSecurityParameters(permissionScope, "individual_"))
+                        .putAll(securityParameterProvider.getSecurityParameters(permissionScope))
```
agorina: I am not sure why it was this way, it should not have (I think)

And I think it will be changed again once I get to this repo.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/ExamItemRepository.java
```diff
@@ -1,12 +1,10 @@
 package org.opentestsystem.rdw.reporting.exam.student;
 
-import org.opentestsystem.rdw.security.Permission;
 import org.opentestsystem.rdw.reporting.common.model.ExamItem;
+import org.opentestsystem.rdw.reporting.common.web.security.User;
```
agorina: Am I the only one who thinks that a `Repository` should not depend on anything from the `web`?

I think it is one more reason to find another home for the `User`.  Please let me know your thoughts.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/JdbcStudentExamRepository.java
```diff
@@ -45,60 +37,30 @@
...
-                    }
-
-                    //Parse the Assessment
-                    final Assessment assessment = Assessments.map(row, Assessment.builder(), "asmt_")
-                            .build();
-
                     //Build the wrapper
                     return new StudentHistoryExamWrapper.Builder()
-                            .exam(exam)
-                            .assessment(assessment)
```
agorina: This was moved to the `Exams.map`.

reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -0,0 +1,134 @@
+sql:
```
agorina: I moved all the SQLs that do not require permissions into a separate file. Seemed cleaner to me. But I admit that maybe it is just because I am focused on the permissions for now. Let me know if you do not think it is a good idea.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -1,75 +1,39 @@
...
-            )
-            and (1 = :group_statewide or s.district_group_id in (:group_district_group_ids) or s.district_id in (:group_district_ids) or s.school_group_id in (:group_school_group_ids) or e.school_id in (:group_school_ids))
-          )
-        )
-      limit 1
+      FROM student st
+        JOIN exam e on e.student_id = st.id
+        JOIN asmt a on a.id = e.asmt_id
+        JOIN school s on e.school_id = s.id
+        JOIN school isch on st.inferred_school_id = isch.id
```
agorina: Had to add this `JOIN`. 
A side effect of using a common sql snippets for permissions.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
esotrope: my guess is that that was a hotfix to allow the sql to reuse a snippet

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/ExamItemRepository.java
esotrope: Yeah that's why i was thinking we could have a more basic user interface and then our web.User implements it

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
esotrope: Nice - the idea is that now every exam repo will go through this processing right?

reporting-service/src/main/resources/application-no-permission.sql.yml
esotrope: i like it - that could be helpful when maintaining

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Exams.java
agorina: That is correct.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/ExamItemRepository.java
wtritch: Our user extends the base spring one... but i’m Not sure that’s enough.  Can we just move it to RDW common model?

agorina: That looks like a very simple move. I will put it into `org.opentestsystem.rdw.reporting.common.security`. (just too many files get changed to include into this PR.)

reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -0,0 +1,134 @@
...
+
+  student:
+    findOneById: >-
+      select id,
+        ssid,
+        first_name,
+        last_or_surname
+      from student s
+      where s.id=:studentId
+       and exists (select 1 from exam e where e.student_id = s.id)
```
agorina: @esotrope, @wtritch: Should this call be protected by a permission clause?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/ExamItemRepository.java
malaffoon: I would vote rather emphatically for an interface. Then we don't really care where the implementation lives.

agorina: Ok, will introduce an interface.

agorina:
I am merging and moving forward with these changes. Will address any comments if needed.

reporting-service/src/main/resources/application-no-permission.sql.yml
esotrope: would need more context.
I know some of our sql are executed after a statement that limits results with a security clause.
A fake example:
Get all exams i am permitted to see.
Get all students for each exam i just got back and assemble them together.

wtritch: I'm going to agree with Landon that it's contextual.  I guess my question is if it should be?

PR #916, opened 2018-02-15
add spinner when user clicks "Download Data"

dsnyder-fairway:
also removed query-builder styling and translations


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -115,9 +119,13 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
...
 
   onDownloadDataButtonClick(): void {
+    this.spinnerModal.loading = true;
     this.reportService.getReportContent(this.report.id)
       .subscribe(download => {
+        this.spinnerModal.loading = false;
         saveAs(download.content, download.name);
+      }, error => {
+        this.spinnerModal.loading = false;
       });
```
esotrope: tip: there is a third callback to subscribe method called "completed" which is called after success and error that you could use to put the loading = false in

esotrope:
may want this on the "My Reports" page as well now that i think about it

dsnyder-fairway:
@esotrope good call. I'll add this to downloads there

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
dsnyder-fairway: I wasn't able to add this for completed (specifically, on failures). Seeing the api docs shows for the completed param:
```
A handler for a terminal event resulting from successful completion.
```

dsnyder-fairway:
@esotrope please review my latest changes for the My Reports page

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
esotrope: Looks like you're right : /
I guess there is an RxJS "finally" method that can be used for this functionality - kind of strange

dsnyder-fairway: 👍, added!

PR #918, opened 2018-02-16
Fix homepage button spacing

wtritch:


wtritch:
Merging now, will resolve comments if/when they come in

PR #920, opened 2018-02-16
Feature/request summary view

esotrope:
Minimal

image

Full

image

Default closed

image

Minimal

image

Full

image


malaffoon:
Would the second screen look better with 3 columns instead of 4? Avoid some of the text wrapping?

Also, can we put a visual separator between the four sections? Maybe a horizontal line?

wtritch:
With some of the feedback about being unable to distinguish between subgroups and filters, I wonder if we want to have sections with headers in there?  "Organizations" "Assessments" "Student and Exam Filters" "Comparative Subgroups" ?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -532,131 +536,35 @@ <h3 class="h3 gray-darkest">
...
-      ]
-    }
-    ]"></aggregate-report-summary>
+    <div class="col-md-8 col-md-offset-2">
+      <div class="well">
+
+        <!-- TODO add estimated row count section here -->
+
+        <aggregate-report-summary #summary
+                                  [narrow]="true"
```
wtritch: Can this just be `narrow="true"` ?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -360,11 +364,15 @@ export class AggregateReportFormComponent {
...
     const assessmentDefinition = this.currentAssessmentDefinition;
+
+    // TODO this table should be lazily updated when it is scrolled into view. There is serious lag when changing settings above
     this.previewTable = {
       rows: this.tableDataService.createSampleData(assessmentDefinition, this.settings),
       options: this.aggregateReportOptions,
       assessmentDefinition: assessmentDefinition
     };
+
+    this.summary && this.summary.updateColumns();
```
wtritch: hmm... do we actually need to tell the summary when to update columns?  Can it just update itself when the \@Input changes?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -1,49 +1,203 @@
...
+
+    const orAll = (options, values, codeProvider) => values.length > 0
+      ? (equalSize(options, values) ? [ All ] : values.map(codeProvider))
+      : [ None ];
+
+    const organizations = settings.districts.concat(settings.schools);
+
+    const includes = [];
+    !assessmentDefinition.interim && settings.includeStateResults
+      && includes.push(translate('aggregate-reports-summary.field.include.state-results'));
```
wtritch: I know this is js-standard, but every time I see an `if(...) { ... }` re-written like this it makes me twinge a bit...

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -1,49 +1,203 @@
...
+    settings.includeAllDistricts
+      && includes.push(translate('aggregate-reports-summary.field.include.all-districts'));
+    settings.includeAllSchoolsOfSelectedDistricts
+      && includes.push(translate('aggregate-reports-summary.field.include.all-schools-of-districts'));
+    settings.includeAllDistrictsOfSelectedSchools
+      && includes.push(translate('aggregate-reports-summary.field.include.all-districts-of-schools'));
+
+    const organizationRows = [
```
wtritch: I wonder if we should break these out into providers... ?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
esotrope: yup i think so

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
esotrope: it can if we reorganize the component to accept one binding object that changes when the form is updated like we do the table. Problem is we pass it options  which doesn't change and settings which has it's fields change but doesn't itself change

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
esotrope: this can be done

esotrope:
@malaffoon @wtritch I'll throw up another iteration with 3 columns, column separation and headers

esotrope:
Tried a grid with headers but I'm thinking nah.

image

image

esotrope:
Original but with headers.

image

image

esotrope:
Latest:
image

PR #921, opened 2018-02-16
Add column ordering to the aggregate report results view

wtritch:
This PR adds the ability to order Aggregate Report results table columns.

The order can be modified by selecting a column identifier and pressing the left/right buttons or by drag-and-drop re-ordering.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
@@ -17,6 +17,7 @@ import { saveAs } from "file-saver";
 import { DisplayOptionService } from "../../shared/display-options/display-option.service";
 import { TranslateService } from "@ngx-translate/core";
 import { SpinnerModal } from "../../shared/loading/spinner.modal";
+import { OrderableItem } from "../../shared/order-selector/order-selector.component";
 
 const PollingInterval = 4000;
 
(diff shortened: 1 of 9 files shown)


webapp/src/main/webapp/src/app/shared/order-selector/order-selector.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    this.itemsChange.emit(this.items);
+  }
+
+  public onDrag() {
+    this.selectedIdx = undefined;
+    this.selectedItem = undefined;
+  }
+
+  public onDrop() {
+    console.log("items", this.items);
```
malaffoon: remove?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -175,6 +187,8 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
       const item = this.itemMapper.map(this.assessmentDefinition, row, index);
       const subjectCode = row.assessment.subjectCode;
       const tableWrapper = tableWrappers.find(wrapper => wrapper.subjectCode == subjectCode);
+      //TODO this should come from the query
+      const columnOrder: string[] = ['organization', 'assessmentGrade', 'schoolYear', 'dimension'];
```
malaffoon: i take it we have to add it to the query all the way through the stack?

wtritch: yessir.  And display it on the sample table.  I'll do that in a follow on PR.

webapp/src/main/webapp/src/app/shared/order-selector/order-selector.component.ts
wtritch: doh, nice catch

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
agorina: Just curious: are we allowing users to re-order *after* the report is generated? And if yes, are we planning on saving the changed order as part of their request?

wtritch: We are allowing the user to re-order after the report is generated.
Nothing they do while viewing the report results is "saved."  (Sorting, viewing the next page, displaying Percent or Numbers, etc including column ordering.)
We *should* allow the user to re-order the preview table much they can select whether to display Percent or Numbers.  We should persist their chosen order through the report generation and default their result table column order to the order the user chose at query time.

agorina: Thank you for clarifying it.

wtritch: However, if the user exports a CSV of a table after modifying it, the CSV will respect the currently-displayed column order, sorting, and other display-time options.

PR #925, opened 2018-02-19
Round-trip column ordering from the aggregate report query table to t…

wtritch:
…he results table.

This PR adds the column ordering component to the preview table in the aggregate report query form.  The user-selected column order is round-tripped through the request and initializes the result tables with the column order specified at query time.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-column-order-item.provider.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-column-order-item.provider.ts
@@ -1,0 +1,32 @@
+import { Injectable } from "@angular/core";
+import { OrderableItem } from "../shared/order-selector/order-selector.component";
+
+const ColumnToLabel: {[key: string]: string} = {
+  organization: 'aggregate-reports.results.cols.organization-name',
+  assessmentGrade: 'aggregate-reports.results.cols.assessment-grade',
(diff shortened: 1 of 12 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-column-order-item.provider.ts
```diff
@@ -0,0 +1,32 @@
...
+  schoolYear: 'aggregate-reports.results.cols.school-year',
+  dimension: 'aggregate-reports.results.cols.dimension'
+};
+
+/**
+ * Responsible for providing {@link OrderableItem} representations
+ * of aggregate report table columns.
+ */
+@Injectable()
+export class AggregateReportColumnOrderItemProvider {
```
wtritch: Since I needed this transformation in both the query and results components, I pulled it into a little shared provider.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.ts
```diff
@@ -91,6 +91,11 @@ export interface AggregateReportFormSettings {
    */
   valueDisplayType: string;
 
+  /**
+   * The user-selected result column order
+   */
+  columnOrder?: string[];
```
wtritch: Since we may have reports without a defined column order, I've made the property optional and handled the "missing" case in the consumers.

PR #926, opened 2018-02-19
Handle a session timeout in the embargo admin UX

wtritch:
This PR changes our embargo admin service to use the DataService and AuthenticatedHttpService to properly handle 401 responses.

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
@@ -65,6 +65,9 @@ export class EmbargoConfirmationModal {
 
     // apply desired embargo setting via the API
     this.embargoService.update(embargo, scope, value)
+      .finally(() => {
+        this.modal.hide();
+      })
       .subscribe(() => {
 
         // reflect new embargo setting in the UI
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
```diff
@@ -65,6 +65,9 @@ export class EmbargoConfirmationModal {
 
     // apply desired embargo setting via the API
     this.embargoService.update(embargo, scope, value)
+      .finally(() => {
+        this.modal.hide();
+      })
```
wtritch: After the initial fix I found myself being navigated to the session expired page, but the modal was still showing with the "warning, you're about to release stuff" message.  This closes the confirmation modal on completion regardless of success or failure.

webapp/src/main/webapp/src/app/admin/embargo/embargo.service.ts
```diff
@@ -2,19 +2,20 @@ import { Observable } from "rxjs/Observable";
...
 
 /**
  * Service responsible for managing organization embargo settings
  */
 @Injectable()
 export class EmbargoService {
 
-  constructor(private http: HttpClient) {
+  constructor(private dataService: DataService) {
   }
```
wtritch: The injected HttpClient doesn't have our general 401 handling.

PR #927, opened 2018-02-19
My Reports should display "No Results" when export of district / scho…

dsnyder-fairway:
…ol contains no results

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/FileDetail.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/FileDetail.java
@@ -1,0 +1,67 @@
+package org.opentestsystem.rdw.reporting.processor.model;
+
+import java.net.URI;
+
+public class FileDetail {
+
+    private String relativePath;
+    private boolean hasResults;
+    private URI uri;
+
+    private FileDetail() {
+
+    }
+
+    public String getRelativePath() {
+        return relativePath;
+    }
+
+    public boolean getHasResults() {
+        return hasResults;
+    }
+
+    public URI getUri() {
+        return uri;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+
+        private String relativePath;
+        private boolean hasResults;
+        private URI uri;
+
+        public Builder relativePath(final String relativePath) {
+            this.relativePath = relativePath;
+            return this;
+        }
+
+        public Builder hasResults(final boolean hasResults) {
+            this.hasResults = hasResults;
+            return this;
+        }
+
+        public Builder uri(final URI uri) {
+            this.uri = uri;
+            return this;
+        }
+
+        public Builder copy(final FileDetail fileDetail) {
+            this.relativePath = fileDetail.relativePath;
+            this.hasResults = fileDetail.hasResults;
+            this.uri = fileDetail.uri;
+            return this;
+        }
+
+        public FileDetail build() {
+            final FileDetail fileDetail = new FileDetail();
+            fileDetail.hasResults = hasResults;
+            fileDetail.relativePath = relativePath;
+            fileDetail.uri = uri;
+            return fileDetail;
+        }
+    }
+}
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
@@ -1,15 +1,17 @@
 package org.opentestsystem.rdw.reporting.processor.repository;
 
 import com.google.common.collect.ImmutableMap;
+import com.mysql.jdbc.ResultSetImpl;
 import org.opentestsystem.rdw.reporting.common.jdbc.SecurityParameterProvider;
 import org.opentestsystem.rdw.reporting.processor.model.ExportExamReportRequest;
+import org.opentestsystem.rdw.reporting.processor.model.FileDetail;
 import org.opentestsystem.rdw.security.PermissionScope;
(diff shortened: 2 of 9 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -34,21 +36,28 @@
...
                         .build(),
-                PreparedStatement::execute
+                (ps) -> {
+                    final ResultSet resultSet = ps.executeQuery();
+                    if (resultSet instanceof ResultSetImpl) {
+                        // Headers are one row. So, we have results if we have more than one row.
+                        return ((ResultSetImpl) resultSet).getUpdateCount() > 1;
+                    }
+                    return false;
+                }
```
dsnyder-fairway: this I am not proud of, but I was unable to find a way to get the update count from a "SELECT ... INTO OUTFILE" command

dsnyder-fairway: also not sure how we should handle this outside of MySQL

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/FileDetail.java
```diff
@@ -0,0 +1,67 @@
+package org.opentestsystem.rdw.reporting.processor.model;
+
+import java.net.URI;
+
+public class FileDetail {
```
esotrope: naming nits: FileDetail -> FileDetails, hasResults -> hasContent

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -34,21 +36,28 @@
     }
 
     @Override
-    public String export(final PermissionScope permissionScope, final ExportExamReportRequest request, final String uriLocation) {
+    public FileDetail export(final PermissionScope permissionScope, final ExportExamReportRequest request, final String uriLocation) {
         final boolean isDestinationS3 = uriLocation.toLowerCase().startsWith("s3");
 
-        template.execute(
+        final boolean hasResults = template.execute(
                 exportExam.replace(ExportDestination, isDestinationS3 ? "S3" : ""),
```
esotrope: so the underlying sql writes a file to s3 directly and we never knew if the sql returned results or not. Now we execute the query directly and check how many rows the query returned to know if it had content or not.  - did i gather that correctly?

dsnyder-fairway: yes -- except this shouldn't matter if written to s3 or now

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
agorina: Just curious: have we tested it on Aurora? 

An alternative approach I could see is to run a `SELECT ... LIMIT 1` before exporting and check if it returns result.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/FileDetail.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private String relativePath;
+    private boolean hasResults;
+    private URI uri;
```
malaffoon: Sometimes the relativePath will be set and sometimes the URI? If true, that feels a lot like a union structure from C days and i think we should talk about refactoring.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExamsTest.java
```diff
@@ -71,7 +72,7 @@ public void itShouldUpdateReportLocationUriAfterExport() {
         final URI exportUri = URI.create("EXPORT/EXAMS/test.csv");
 
         when(exportService.exportExam(any(User.class), any(ExportExamReportRequest.class), anyInt()))
-                .thenAnswer(x -> exportUri);
+                .thenAnswer(x -> FileDetail.builder().relativePath("EXPORT/EXAMS/test.csv").hasResults(true).uri(exportUri).build());
```
malaffoon: Having the literal hard-coded here kind of defeats the point of the const variable. I'd suggest either hard-coding both (yech) or adding another const variable, e.g.
```
    final String relativePath = "EXPORT/EXAMS.test.csv";
    final URI exportUri = URI.create(relativePath);
    ...
```
Of course, all this makes me reiterate my earlier comment that the FileDetail structure feels ... messy.

wtritch:
After looking this over I'm *really* tempted to say it's time to get rid of the whole "SELECT INTO OUTFILE" implementation and instead return results from the DB and print them to a CSV via Java...

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/FileDetail.java
dsnyder-fairway: relative path is used first, then the URI is set. I can probably refactor this to just use the URI

dsnyder-fairway:
@wtritch @esotrope I'll go along with that suggestion.

dsnyder-fairway:
closing for now

agorina:
Sorry, I just saw @wtritch's  and @esotrope's suggestion above and I would like to voice my concern about it. If I recall it correctly, the reason COPY options was chosen is because this export could be a few millions of records, and COPY outperforms significantly (we are talking minutes vs hours). Maybe something to test before we do this change. Or just leave it as is..

PR #929, opened 2018-02-19
Fix downloads in the My Reports page

wtritch:
The introduction of the spinner in the My Reports download page prevented users from being able to download aggregate reports.

This PR moves the download saving out of the My Reports component and back into the action handler.  The action handler will now return an observable that the component can use to determine when an arbitrary action has been completed.

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-action.component.ts
+++ b/webapp/src/main/webapp/src/app/report/report-action.component.ts
@@ -3,12 +3,9 @@ import { ReportAction, ReportActionService } from "./report-action.service";
 import { PopupMenuAction } from "../shared/menu/popup-menu-action.model";
 import { Report } from "./report.model";
 import { TranslateService } from "@ngx-translate/core";
-import { Observable } from "rxjs/Observable";
-import { Download } from "../shared/data/download.model";
-import { saveAs } from "file-saver";
 import { SpinnerModal } from "../shared/loading/spinner.modal";
-import { NotificationService } from "../shared/notification/notification.service";
 import 'rxjs/add/operator/finally';
+import { Observable } from "rxjs/Observable";
 
 /**
  * Responsible for providing a UI displaying and performing an action
@@ -32,8 +29,7 @@ export class ReportActionComponent implements OnInit {
   public menuActions: PopupMenuAction[];
 
   constructor(private actionService: ReportActionService,
-              private translateService: TranslateService,
-              private notificationService: NotificationService) {
+              private translateService: TranslateService) {
   }
 
   ngOnInit(): void {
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/report/report-action.component.ts
```diff
@@ -44,25 +40,18 @@ export class ReportActionComponent implements OnInit {
...
-        (download: Download) => {
-          saveAs(download.content, download.name);
-        },
-        () => {
-          this.notificationService.error({ id: 'labels.reports.messages.download-failed' });
-        },
-      );
-    }
+      })
+      .subscribe(() => {});
```
wtritch: The component now sets the spinner visible before performing the action, and sets the spinner invisible after the action has completed.

webapp/src/main/webapp/src/app/report/report-action.component.ts
```diff
@@ -44,25 +40,18 @@ export class ReportActionComponent implements OnInit {
...
   }
 
   private toMenuAction(reportAction: ReportAction): PopupMenuAction {
     const menuAction: PopupMenuAction = new PopupMenuAction();
     menuAction.perform = () => {
-      this.actionService.performAction(reportAction);
+      this.performAction(reportAction);
```
wtritch: This makes the popup menu actions follow the same path as the direct button actions.

webapp/src/main/webapp/src/app/report/report-action.service.ts
```diff
@@ -40,17 +43,33 @@ export class ReportActionService {
...
+    observable
+      .subscribe(
+        (download: Download) => {
+          saveAs(download.content, download.name);
+          console.log("done saving");
+        },
+        (error) => {
+          this.notificationService.error({ id: 'labels.reports.messages.download-failed' });
+        });
+    return observable;
```
wtritch: The action service now performs the action rather than relying upon a consumer to register a subscriber to trigger execution.

webapp/src/main/webapp/src/app/report/report-action.service.ts
```diff
@@ -40,17 +43,33 @@ export class ReportActionService {
...
+   * @param {ReportAction} action
+   * @returns {Observable<any>}
+   */
+  private performDownload(action: ReportAction): Observable<any> {
+    const observable: Observable<any> = this.reportService.getReportContent(action.value);
+    observable
+      .subscribe(
+        (download: Download) => {
+          saveAs(download.content, download.name);
+          console.log("done saving");
```
esotrope: should probably remove this. iris does a good job filling up the console logs for us already : j

dsnyder-fairway:
ugh so sorry for closing the issue :(. Thank you!

webapp/src/main/webapp/src/app/report/report-action.service.ts
wtritch: In general, we only need popup actions as decorators around report actions.  If the report has 1 or 0 actions, we can just use the report actions directly.  I'm using report actions rather than popup menu actions because I don't want to re-purpose PopupMenuActions for use outside of a PopupMenu =)

PR #931, opened 2018-02-20
DWR-1189 Modify reporting permission SQL to include embargo setting check

agorina:


common/src/main/resources/common.sql.yml
```diff
@@ -159,6 +159,14 @@ sql:
             )
           )
         )
+        AND
+        (
+           e.school_year < ${app.school-year}
```
agorina: I did a 'less than' comparison just in case the system is not configured properly for the next year.  I have no strong opinion about it, please let me know if you think an equal is more appropriate.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
```diff
@@ -54,9 +55,12 @@ public SecurityParameterProvider(final TenantProperties tenantProperties) {
      * @return map of SQL query parameters for the given permission scopeo
      */
     public Map<String, Object> getSecurityParameters(final PermissionSource permissionSource, final boolean disableTransferAccess) {
-
-        return getSecurityParametersBuilder(permissionSource.getPermissionsById(), tenantProperties.isTransferAccessEnabled() && !disableTransferAccess)
+        final Map<String, Permission> permissionMap = permissionSource.getPermissionsById();
+        final PermissionScope embargoScope = permissionMap.containsKey(EmbargoRead) ? permissionMap.get(EmbargoRead).getScope() : PermissionScope.builder().build();
```
malaffoon: There is a flyweight for PermissionScope.builder().build(): PermissionScope.EMPTY.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/student/JdbcExamEmbargoPermissionCurrentSchoolYearIT.java
```diff
@@ -0,0 +1,134 @@
...
+        final List<StudentHistoryExamWrapper> exams = repository.findAllForStudent(userBuilder.permissionsById(individualStatewide).build(), -1L);
+        assertThat(repository.findAllForStudent(userBuilder.permissionsById(individualStatewide).build(), -1L).size())
+                .isEqualTo(examCount - sumExamSchool30Count);
+
+        assertThat(exams
+                .stream()
+                .filter(e -> (e.getExam().getSchool().getId() != -30))
+                .filter(e -> (e.getAssessment().getType() == SUMMATIVE))
+                .count())
+                .isGreaterThan(0);
```
malaffoon: What exactly is this testing? I guess i'm asking why the vague count check, can't we know exactly what it should be?

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/SecurityParameterProvider.java
agorina: I will add it to common since it is most likely will be needed in the aggregate report too.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/student/JdbcExamEmbargoPermissionCurrentSchoolYearIT.java
agorina: Yes, we can. I will change it.

common/src/main/resources/common.sql.yml
esotrope: this assumes all previous years' exams are released right? I think this is a safe assumption but could be wrong. What issue is there with using `=`?

agorina: "_this assumes all previous years' exams are released right_" - I thought this was the assumption all along? The `reporting` db has only the current year embargo flag.

"_What issue is there with using =?_" - if everything is done properly, there is no issue.
Here is a use case when things could go wrong: Come 2019, the `app.school-year` is not changed (for some reason), but 2019 exams are embargoed.  An equal in here will not protect the embargoed data.

esotrope: Gotcha - looking at the where clause i see what you're saying. In that case i think we need to make it `<` or otherwise make sure `>` can not be included in the result set.

PR #932, opened 2018-02-20
Remove the html id conflict between the advanced filters transfer tog…

wtritch:
…gle and the report dialog transfer toggle.

After toggling the control in the advanced filters the control in the report dialog was unresponsive and vice-versa.

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-download.component.html
+++ b/webapp/src/main/webapp/src/app/report/report-download.component.html
@@ -49,12 +49,12 @@
             </div>
 
             <div class="form-group" *ngIf="transferAccess">
-              <label>
+              <label for="report-transfer-assessment-toggle">
                 <span info-button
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -49,12 +49,12 @@ <h4 class="modal-title pull-left">{{title}}</h4>
             </div>
 
             <div class="form-group" *ngIf="transferAccess">
-              <label>
+              <label for="report-transfer-assessment-toggle">
```
esotrope: do we even need label `for` and `id`s? In general I tend to omit html ids and label-for in favor of wrapping inputs with labels:
`<label>Label: <input></label>`

wtritch: We need the id to tell the browser which radio buttons are tied together.  The label for is just habit for accessibility =P

PR #934, opened 2018-02-20
My Reports should display "No Results" when export of district / scho…

dsnyder-fairway:
…ol contains no results

this approach uses the `limit 1` suggested in the previous PR

### DIFF
--- a/report-processor/src/main/resources/application.sql.yml
+++ b/report-processor/src/main/resources/application.sql.yml
@@ -129,19 +129,85 @@ sql:
         'AssignedAccommodationList'
       union all
       select
-        ${sql.reportProcessor.snippet.exportExams.select}
+        d.natural_id as district_natural_id,
+        d.name as district_name,
+
+        coalesce(s.external_id, '') as school_ncesid,
+        s.natural_id as school_natural_id,
+        s.name as school_name,
+
+        st.ssid as student_ssid,
+        coalesce(st.first_name, '') as student_first_name,
+        coalesce(st.last_or_surname, '') as student_last_name,
+
+        e.completed_at,
+        e.school_year,
+        e.session_id,
+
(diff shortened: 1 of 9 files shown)


report-processor/src/main/resources/application.sql.yml
```diff
@@ -129,85 +129,19 @@ sql:
...
-          (
-            (1 = :allow_transfer_access and 0 = :disable_transfer_access) and
-            (
-              isch.district_group_id in (:query_district_group_ids) or isch.district_id in (:query_district_ids) or isch.school_group_id in (:query_school_group_ids) or isch.id in (:query_school_ids)
-            )
-          )
-        )
-        and ${sql.snippet.studentExamPermission}
+      ${sql.reportProcessor.snippet.exportExams.from}
+      ${sql.reportProcessor.snippet.exportExams.where}
```
esotrope: is there a reason we cut it into so many pieces? why not just the query + limit 1?

esotrope: Also, let's try to avoid nesting snippets

report-processor/src/main/resources/application.sql.yml
```diff
@@ -129,85 +129,19 @@ sql:
...
+
+    hasExams: >-
+      select
+        ${sql.reportProcessor.snippet.exportExams.select}
+      ${sql.reportProcessor.snippet.exportExams.from}
+      ${sql.reportProcessor.snippet.exportExams.where}
+      limit 1
```
agorina: I suggest to not change the original SQL for export and instead create a new one to test if an exam exists. The new query really does not need all the data fields in the SELECT. And you may even change it to `SELECT 1 IF EXISTS`.
Let me know if you need help with it.

dsnyder-fairway: done

PR #935, opened 2018-02-20
Bugfix/fix group uploads via zuul

wtritch:
The issue was two-fold:
1. The FileUploader front end component wasn't pointing at the correct URI
2. The Zuul proxy was choking on the multipart content type: `Content-Type:multipart/form-data; boundary=----WebKitFormBoundaryv4veCaucN8EcsIOE`

I bumped the Zuul dependency to a version that properly handles Content-Type headers via parsing rather than attempting to hand-process them.


admin-service/build.gradle
```diff
@@ -27,7 +27,7 @@ dependencies {
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-guava'
     compile 'com.google.guava:guava'
     compile 'mysql:mysql-connector-java'
-    compile 'org.apache.commons:commons-io'
+    compile 'commons-io:commons-io'
```
wtritch: As long as I was in here, I changed all of the reporting modules to bind to the same apache commons-io version.

build.gradle
```diff
@@ -94,6 +94,10 @@ subprojects {
 
             // jwt
             dependency 'io.jsonwebtoken:jjwt:0.9.0'
+
+            // override default spring boot zuul version to handle multipart form proxy requests
+            // TODO remove this override after upgrading to Camden.SR7 or above
+            dependency 'org.springframework.cloud:spring-cloud-netflix-core:1.2.7.RELEASE'
```
wtritch: Previously we were bound to 1.2.5.RELEASE

webapp/src/main/java/org/opentestsystem/rdw/reporting/zuul/RequestContentTypeCleaningFilter.java
```diff
@@ -1,57 +0,0 @@
...
-import static com.netflix.zuul.context.RequestContext.getCurrentContext;
-
-/**
- * This filter is responsible for cleaning extra whitespace from Content-Type
- * header values.
- * This wrapper relies on MediaType to parse and format the content-type which
- * might not strictly comply with the spec, which is what the zuul parser requires.
- */
-@Component
-public class RequestContentTypeCleaningFilter extends ZuulFilter {
```
wtritch: The the Zuul dependency bump we no longer need this filter.

webapp/src/main/webapp/src/app/admin/groups/import/group-import.component.ts
```diff
@@ -5,7 +5,7 @@ import { FileUploader } from "ng2-file-upload";
 import { TranslateService } from "@ngx-translate/core";
 import { Utils } from "../../../shared/support/support";
 
-const URL = '/api/studentGroups/';
+const URL = '/api/admin-service/studentGroups';
```
wtritch: This is the url change for the FileUploader used during group CSV uploading.

admin-service/build.gradle
esotrope: I did that before and things blew up. dont remember what though

webapp/src/main/webapp/src/app/admin/groups/import/group-import.component.ts
esotrope: fyi made a common file with the service routes. You should be able to import `
`AdminServiceRoute`

wtritch: Works for me, changed to:
`const URL = '/api' + AdminServiceRoute + '/studentGroups';`

admin-service/build.gradle
wtritch: I did a quick smoke test and everything *appeared* to work fine...  Having multiple versions of the same lib gives me heart-burn =/

esotrope: yeah i remember changing this and i got some runtime exceptions so i begrudgingly reverted

PR #937, opened 2018-02-20
Allow a user to include all aggregate report dimensions

wtritch:
We were previously treating dimensions similarly to filters, where if all were selected the front-end was stripping the parameters from the request payload.

### DIFF
--- a/webapp/docker-compose.yml
+++ b/webapp/docker-compose.yml
@@ -25,109 +25,109 @@ services:
       - 5672:5672
       - 15672:15672
 
-#  config-server:
-#    image: smarterbalanced/configuration-service:3.1.0.RELEASE
-#    container_name: config-server
+  config-server:
+    image: smarterbalanced/configuration-service:3.1.0.RELEASE
+    container_name: config-server
+    ports:
+      - 8888:8888
+    links:
+      - rabbitmq
+    environment:
+      - CONFIG_SERVICE_REPO
+      - ENCRYPT_KEY
+      - GIT_USER
+      - GIT_PASSWORD
+      - SPRING_RABBITMQ_ADDRESSES=rabbitmq:5672
+      - SPRING_RABBITMQ_USERNAME=guest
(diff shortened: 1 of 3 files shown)


webapp/docker-compose.yml
```diff
@@ -25,109 +25,109 @@ services:
       - 5672:5672
       - 15672:15672
 
-#  config-server:
```
wtritch: unrelated, revert what I assume was an unintended change

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -136,7 +134,7 @@ export class AggregateReportRequestMapper {
             ? options.completenesses
             : query.completenessCodes,
           dimensionTypes: Utils.isNullOrEmpty(query.dimensionTypes)
-            ? options.dimensionTypes
+            ? []
             : query.dimensionTypes,
```
wtritch: If the query value is `null` initialize to an empty array.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -65,9 +66,6 @@ export class AggregateReportRequestMapper {
     if (!equalSize(settings.completenesses, options.completenesses)) {
       query.completenessCodes = settings.completenesses;
     }
-    if (!equalSize(settings.dimensionTypes, options.dimensionTypes)) {
-      query.dimensionTypes = settings.dimensionTypes;
-    }
```
wtritch: We just want whatever the settings are, don't strip the values if the user selects them all.

webapp/docker-compose.yml
agorina: If this was me, then yes. Sorry and thank you for fixing it!

PR #939, opened 2018-02-20
Bugfix/delay large agg report download

wtritch:
The problem was that the aggregate report view component was immediately downloading a report if its state was "Completed", even if we displayed a warning the user that it may be too large to view.

This PR converts the access-time determination of the ViewState into a ViewState flow with triggers when the value changes from one state to another.  It was the cleanest way I could see to control when we load the report, start polling, display a size warning, or override the warning and load the report anyway.


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.html
```diff
@@ -44,7 +44,9 @@ <h2 class="h3 blue-dark">{{report.label}}</h2>
         <p>{{'aggregate-report.error.unsupported-size' | translate}}</p>
         <div>
           <button class="btn btn-danger btn-sm" (click)="onDisplayReportButtonClick()">{{'aggregate-report.display-report' | translate}}</button>
-          <button class="btn btn-default btn-sm" (click)="onDownloadDataButtonClick()">{{'aggregate-report.download-data' | translate}}</button>
+          <button class="btn btn-default btn-sm"
+                  [disabled]="spinnerModal.loading"
```
wtritch: small, unrelated change.  I'm disabling the "download" button when the user clicks it before the delayed spinner modal appears.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -77,6 +77,7 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
         options: this.options,
         settings: settings
       });
+    this.updateViewState();
```
wtritch: updateViewState() is the gateway for managing/updating/triggering on the current view state

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -163,25 +144,50 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
     tableView.columnOrdering = items.map(item => item.value);
   }
 
-  private loadOrPollReport(): void {
+  private updateViewState(): void {
+    let targetViewState: ViewState = this.getTargetViewState();
+    this.onViewStateChange(this._viewState, targetViewState);
+    this._viewState = targetViewState;
+  }
```
wtritch: determine what the view state should be, perform any view state change processing, and update the current view state.

webapp/src/main/webapp/src/app/report/report-action.service.ts
```diff
@@ -6,6 +6,7 @@ import { Router } from "@angular/router";
 import { saveAs } from "file-saver";
 import { Observable } from "rxjs/Observable";
 import { NotificationService } from "../shared/notification/notification.service";
+import "rxjs/add/observable/empty";
```
wtritch: unrelated, saw some console errors that `Observable.empty()` wasn't a valid method.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -163,25 +144,50 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
...
+
+    if (oldState == ViewState.ReportProcessing) {
+      this.unsubscribe();
+    }
+    if (newState == ViewState.ReportProcessing) {
       this.pollReport();
     }
+    if (newState == ViewState.ReportView) {
+      this.loadReport();
+    }
```
esotrope: was there a bug here or did you just want to make this more clearly a finite state machine?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -35,7 +34,7 @@ const PollingInterval = 4000;
   selector: 'aggregate-report',
   templateUrl: 'aggregate-report.component.html',
 })
-export class AggregateReportComponent implements OnInit, OnDestroy {
+export class AggregateReportComponent implements OnDestroy {
```
esotrope: learned the other day that these interfaces are technically optional - angular will find the methods whether you put these interfaces here or not

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
wtritch: I love FSMs!  But yes, we were simultaneously telling the user that the report may be too large to view *and* downloading it behind the scenes, rather than waiting for them to click the "View anyway" button.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
wtritch: Neat.  Interfaces are for devs, not the system, huh?  ...makes sense, since you can't detect them at runtime.

webapp/src/main/webapp/src/app/report/report-action.service.ts
esotrope: Just for everyone's awareness - I don't think we are using rxjs imports right - i think we are supposed to pick one central place to add all these observable methods like our root app component or something - ill have to look it up again. if we do that it might improve the build time is what i recall. Also with rxjs 5 we are supposed to move away from these in favor of the static method imports for tree shaking in the aot compilation.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
esotrope: gotcha - great find 👍

PR #940, opened 2018-02-21
Prevent student, group, and school-grade report download modal dialog…

wtritch:
…s from interfering with each other.

The problem was that our report download dialog modals are *always* in the DOM, whether or not they're visible.  This led to interference between two modal dialog's inputs due to sharing the same HTML id attribute.

This fix is to remove the modal dialog's elements from the DOM until/unless the modal dialog is visible.

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-download.component.html
+++ b/webapp/src/main/webapp/src/app/report/report-download.component.html
@@ -6,7 +6,8 @@
      aria-hidden="true"
      (onShow)="onShowInternal($event)">
 
-  <div class="modal-dialog modal-md">
+  <div *ngIf="modal.isShown"
+       class="modal-dialog modal-md">
     <div class="modal-content">
       <div class="modal-header">
         <h4 class="modal-title pull-left">{{title}}</h4>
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1450,7 +1450,7 @@
       },
       "form": {
         "accommodations": {
-          "info": "This feature allows student reports that may be shared with colleges and universities or other external organizations to be printed without reference to accommodations.",
+          "info": "This feature allows student reports that may be shared with colleges and universities or other external organizations to be printed without reference to accommodations made available to the student.",
```
wtritch: semi-related verbiage change

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -6,7 +6,8 @@
      aria-hidden="true"
      (onShow)="onShowInternal($event)">
 
-  <div class="modal-dialog modal-md">
+  <div *ngIf="modal.isShown"
+       class="modal-dialog modal-md">
```
wtritch: Keep the content off the DOM until the user is viewing the modal.

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: 👍

PR #941, opened 2018-02-21
handle empty responses on put operations

dsnyder-fairway:
this prevents our UI from misbehaving on PUT operations while receiving an empty response body

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/data/data.service.ts
+++ b/webapp/src/main/webapp/src/app/shared/data/data.service.ts
@@ -88,7 +88,7 @@ export class DataService {
     }
     return response => {
       const contentLength = response.headers.get("content-length");
-      if (!contentLength)
+      if (contentLength == null || Number.parseInt(contentLength) > 0)
         return response.json();
       return null;
     }
--- a/webapp/src/main/webapp/src/app/shared/data/data.service.ts
+++ b/webapp/src/main/webapp/src/app/shared/data/data.service.ts
@@ -53,9 +53,10 @@ export class DataService {
    * @returns {Observable<R>}
    */
   put(url: string, body: any, options?: RequestOptionsArgs): Observable<any> {
-    return this.http
-      .put(`${this.contextUrl}${url}`, body, options)
-      .map(this.getMapper(options));
+    let observable: Observable<any> = this.http
+      .put(`${this.contextUrl}${url}`, body, options);
+    if (options) observable = observable.map(this.getMapper(options));
+    return observable;
   }
 
   /**
--- a/webapp/src/main/webapp/src/app/shared/data/data.service.ts
+++ b/webapp/src/main/webapp/src/app/shared/data/data.service.ts
@@ -53,10 +53,9 @@ export class DataService {
    * @returns {Observable<R>}
    */
   put(url: string, body: any, options?: RequestOptionsArgs): Observable<any> {
-    let observable: Observable<any> = this.http
-      .put(`${this.contextUrl}${url}`, body, options);
-    if (options) observable = observable.map(this.getMapper(options));
-    return observable;
+    return this.http
+      .put(`${this.contextUrl}${url}`, body, options)
+      .map(this.getMapper(options));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -53,9 +53,10 @@ export class DataService {
    * @returns {Observable<R>}
    */
   put(url: string, body: any, options?: RequestOptionsArgs): Observable<any> {
-    return this.http
-      .put(`${this.contextUrl}${url}`, body, options)
-      .map(this.getMapper(options));
+    let observable: Observable<any> = this.http
+      .put(`${this.contextUrl}${url}`, body, options);
+    if (options) observable = observable.map(this.getMapper(options));
+    return observable;
```
wtritch: Hmm... I don't like the conditional on "If the user provided options, then map the response."  Whether the user provides options is independent of what the response could be, isn't it?

Ideally our mappers would instead be able to handle any type of response, including an empty one.

esotrope: Yeah I'm confused by this - so is `Http` returning `Observable.empty()` for empty responses and we are missing that info? What is the spec for `Http.put()` and empty responses?

dsnyder-fairway:
@esotrope @wtritch no longer making modifications to the `put` method, but to the `getMapper`

webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -87,7 +87,11 @@ export class DataService {
         new Blob([ response.blob() ], { type: this.getContentType(response) })
       );
     }
-    return response => response.json();
+    return response => {
+      const contentType = response.headers.has("content-type");
+      if (contentType)
+        return response.json();
+    }
```
wtritch: This... still doesn't feel quite right.
These are the response headers:
```
HTTP/1.1 200
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: 0
X-Frame-Options: SAMEORIGIN
X-Application-Context: rdw-reporting-webapp
Date: Wed, 21 Feb 2018 18:19:07 GMT
Content-Length: 0
```
I would instead probably `return null;` if the "Content-Length" header value is "0", otherwise `return response.json();`
We're still relying on the back-end properly sending the Content-Length header, but at least it's respecting the HTTP protocol.

dsnyder-fairway: corrected this

webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -88,9 +88,10 @@ export class DataService {
       );
     }
     return response => {
-      const contentType = response.headers.has("content-type");
-      if (contentType)
+      const contentLength = response.headers.get("content-length");
+      if (!contentLength)
```
esotrope: shouldn't this be `if (contentLength != null && contentLength > 0)` ?

dsnyder-fairway: `content-length` is often not a provided header. So instead, we are checking that it has either not been provided, or is not 0

dsnyder-fairway: alternatively, we could do `if (contentLength == null || contentLength > 0)`

esotrope: Maybe i don't understand the problem correctly - we are trying to capture when contentLength is not set by the server or 0 right? and then we return null (empty) to the listener? or what am i missing?

dsnyder-fairway: Really we are just trying to avoid contentLength = 0. If it is null or any other value we are good since it would not throw an exception when `response.json()` is called. When contentLength is 0, `response.json()` throws an exception

esotrope: gotcha - thanks for the explanation. In that case i would go with your suggestion of `if (contentLength == null || contentLength > 0)`

dsnyder-fairway: I updated this. Part of my concern for doing this is the we get a string value. So I thought `!contentLength` was more safe. From reading up on content-length it should always be a number though.

dsnyder-fairway:
merging this one

PR #942, opened 2018-02-21
address uat feedback + misc cleanup

esotrope:
Visual changes:

1. Advanced filters will now come after subgroups
2. Well headers moved into wells to make way for advanced filters expander button.
3. Advanced filters summary section and rows are hidden when they are not different from the defaults

image

image

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -320,11 +320,8 @@
           <div class="well">
             <div class="row">
               <div class="col-md-6">
-                <!--<label info-button-->
-                       <!--title="{{'aggregate-reports.form.field.comparative-subgroup.label' | translate}}"-->
-                       <!--content="{{'aggregate-reports.form.field.comparative-subgroup.info' | translate}}"></label>-->
               </div>
-              <div class="col-md-3">
+              <div class="col-md-6">
                 <sb-checkbox-group [options]="options.dimensionTypes"
                                    [(ngModel)]="settings.dimensionTypes"
                                    [ngModelOptions]="{standalone: true}"
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -18,12 +18,12 @@
         text: 'aggregate-reports.form.section.assessment.heading' | translate
       },
       {
-        scrollTo: subgroupFiltersSection,
-        text: 'aggregate-reports.form.section.subgroup-filters.heading' | translate
+        scrollTo: subgroupSection,
+        text: 'aggregate-reports.form.section.comparative-subgroups.heading' | translate
       },
       {
-        scrollTo: comparativeSubgroupSection,
-        text: 'aggregate-reports.form.section.comparative-subgroups.heading' | translate
+        scrollTo: advancedFiltersSection,
+        text: 'aggregate-reports.form.section.subgroup-filters.heading' | translate
       },
       {
         scrollTo: reportNameSection,
@@ -39,11 +39,12 @@
       }
       ]">
       <button class="btn btn-success btn-sm green ml-sm mt-xs"
+              [disabled]="submissionSubscription"
               (click)="onGenerateButtonClick()">{{'aggregate-reports.form.submit' | translate}}</button>
     </scroll-nav>
   </div>
 
-  <div class="row mt-sm">
+  <div class="row mt-sm mb-sm">
     <div class="col-md-8 col-md-offset-2">
       <h2 class="h3 gray-darkest">
         {{'aggregate-reports.form.heading' | translate}}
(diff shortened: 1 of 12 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -332,7 +349,11 @@ export class AggregateReportFormComponent {
    */
   onGenerateButtonClick(): void {
     this.validate(this.formGroup, () => {
-      this.reportService.createReport(this.createReportRequest())
+      this.submissionSubscription = this.reportService.createReport(this.createReportRequest())
+        .finally(() => {
+          this.submissionSubscription.unsubscribe();
+          this.submissionSubscription = undefined;
+        })
```
esotrope: adds guard so that the report cannot be created twice by double clicking.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -133,6 +145,7 @@ export class AggregateReportFormComponent {
     this.assessmentDefinitionsByTypeCode = route.snapshot.data[ 'assessmentDefinitionsByAssessmentTypeCode' ];
     this.aggregateReportOptions = route.snapshot.data[ 'options' ];
     this.settings = route.snapshot.data[ 'settings' ];
+    this.showAdvancedFilters = this.hasExplicitAdvancedFilters(this.settings, this.aggregateReportOptions);
```
esotrope: advanced filters are hidden by default but if the user is visiting the page based on a previous query this will open the advanced filters section when they are different from the defaults.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -344,6 +365,28 @@ export class AggregateReportFormComponent {
     });
   }
 
+  /**
+   * Reloads the report preview based on current form state
+   */
+  onSettingsChange() {
```
esotrope: just moved where this method is in the file

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.html
```diff
@@ -1,27 +1,53 @@
...
         {{section.label}}
       </div>
-      <table>
-        <tbody>
-        <tr *ngFor="let row of section.rows">
-          <ng-container *ngTemplateOutlet="rowTemplate; context: {$implicit: row}"></ng-container>
-        </tr>
-        </tbody>
-      </table>
+      <div class="row" *ngFor="let row of section.rows">
```
esotrope: changed from table to bs rows/cols so that the columns would line up and not cause visual tension

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -413,8 +418,7 @@ export class AggregateReportTableComponent implements OnInit {
   }
 
   private updatePagination() {
-    this._paginationEnabled = !this.preview
-        && this.table
+    this._paginationEnabled = this.table
```
esotrope: re-enabled pagination for the preview table because the table could get really long and render all at once

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -64,6 +62,11 @@ export class ScrollNavComponent {
 
   private updateActiveLink(): void {
 
+    // Sets the first item to active if the window scroll is zero
+    if (this._window.scrollY <= 0) {
+      this._activeItem = this.items[ 0 ];
+    }
```
esotrope: misc scroll nav fixes for better usability

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -52,7 +52,7 @@
       "empty-message": "No Results Found",
       "performance-group-prompt": "Show performance as",
       "performance-grouped": "Grouped",
-      "performance-separated": "Separate",
+      "performance-separated": "All",
```
esotrope: here are the text changes - I'm sure the wording could be better so please suggest improvements

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -124,59 +126,75 @@ export class AggregateReportRequestMapper {
...
-          ethnicities: Utils.isNullOrEmpty(query.ethnicityCodes)
-            ? options.ethnicities
-            : query.ethnicityCodes,
-          genders: Utils.isNullOrEmpty(query.genderCodes)
-            ? options.genders
-            : query.genderCodes,
+          economicDisadvantages: or(
+            sort(query.economicDisadvantageCodes, options.economicDisadvantages),
+            options.economicDisadvantages
+          ),
```
esotrope: not the most straight forward of changes but what this does is tucks the verbose `Utils.isNullOrUndefined(a) ? b : a` into `or(a, b)` and adds client-side sorting to the report query settings (we lose the natural order when sending the settings to the backend where they are represented as `java.util.Set`s)

esotrope:
Thoughts on moving the report name section to the "summary" well and adding a submit button next to it?

image

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
wtritch: We'll need to remember this if we ever re-enable route reuse.  Won't it preserve state?

wtritch:
I like moving the report name into the summary and sticking a button next to it.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -314,217 +309,221 @@ <h3 class="h3 gray-darkest">
...
           </div>
           <div class="well">
             <div class="row">
               <div class="col-md-6">
-                <label info-button
-                       title="{{'aggregate-reports.form.field.ethnicity.label' | translate}}"
-                       content="{{'aggregate-reports.form.field.ethnicity.info' | translate}}"></label>
+                <!--<label info-button-->
+                       <!--title="{{'aggregate-reports.form.field.comparative-subgroup.label' | translate}}"-->
+                       <!--content="{{'aggregate-reports.form.field.comparative-subgroup.info' | translate}}"></label>-->
```
wtritch: ??

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -344,6 +365,28 @@ export class AggregateReportFormComponent {
...
+
+    const assessmentDefinition = this.currentAssessmentDefinition;
+
+    this.summary = {
+      assessmentDefinition: assessmentDefinition,
+      options: this.aggregateReportOptions,
+      settings: this.settings
+    };
+
+    // TODO this table should be lazily updated when it is scrolled into view. There is serious lag when changing settings above
```
wtritch: I wonder if we could get away with doing something a little silly like a 3-second debounce?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -171,10 +171,12 @@
           "heading": "Assessment Attributes"
         },
         "subgroup-filters": {
-          "heading": "Subgroup Filters"
+          "heading": "Advanced Filters",
+          "subtext": "Refine results by selecting student attributes to filter by."
```
wtritch: Hmmm... I'm looking for a simple sentence that says "You're filtering exams from the Overall and all selected comparative subgroups."  with a dash of "If you filter *and* subgroup on a single attribute, only subgroup values matching your filter will be returned."
My guess is that when a user sees Overall, Gender: Male, Section504: Yes|No after filtering on gender and economic disadvantage status they won't instinctively know that Overall is actually Overall sans their filtered exams...

malaffoon: We talked with SB about this. We decided we'd see what it looks like with the new Review section before deciding further.

malaffoon:
I like these changes.
I like combining the Report Name with the Report Review well and adding the generate button.
For the Subgroups well, should we just left-justify the selectors now?

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: Perhaps "Select student attributes to filter results out of your report."

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -171,10 +171,12 @@
...
         },
         "comparative-subgroups": {
-          "heading": "Comparative Subgroups"
+          "heading": "Subgroups",
+          "subtext": "The results will be broken down by the subgroups you select."
```
malaffoon: SB didn't like "broken down" vs. "break down"; perhaps "The report will break down results by the subgroups you select."

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
esotrope: What state would that be?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
esotrope: Good catch - needs to be deleted. This info text is now redundant with the section description.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
esotrope: Yeah maybe - at least as an intermediate fix

esotrope:
@malaffoon - i hesitate to move the subgroups input left because then the user's mouse will have to move left or right depending on which form field they are editing. Like you, i agree the whitespace is pretty annoying though.

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: Maybe we can throw this text over the wall to the client and see what they say?

malaffoon: My suggestion is based on SB input.

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: 👍 i'll incorporate this

PR #943, opened 2018-02-21
Modify instructional resource management UX according to UAT feedback.

wtritch:
This PR resolves some UAT feedback issues/requests:
The "Organization" title was given an info button listing the types of organizations that can be assigned instructional resources.
The create button icon was changed from a cloud upload to a plus.
The labels and layout of the create modal dialog were changed to provide a better prompt to the user that they are filling out an auto-complete dialog.
The modal dialogs now hide if any route navigations are performed.  This fixes an issue where typing into an auto-complete input after the session has expired navigates to the session-expired route, but did not close the modal dialog.

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
+++ b/webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
@@ -13,94 +13,84 @@
     <div *ngIf="duplicateResource" class="alert alert-danger">
       {{ 'labels.instructional-resource.create.duplicate' | translate }}
     </div>
-    <form class="form-horizontal">
+    <form>
       <fieldset>
-        <div class="form-group row mb-xs">
-          <label for="assessment-label" class="col-xs-4 control-label">{{'labels.instructional-resource.cols.assessment-label' | translate}}</label>
-          <div class="col-xs-8 pl-0">
-            <input class="form-control"
-                   id="assessment-label"
-                   name="assessment-label"
-                   [typeahead]="assessmentSource"
-                   (typeaheadLoading)="assessmentLoading = $event"
-                   (typeaheadNoResults)="assessmentNoResults = $event"
-                   (typeaheadOnSelect)="onAssessmentSelect($event.item)"
-                   [typeaheadMinLength]="3"
-                   [(ngModel)]="assessmentSearch"
-                   [typeaheadItemTemplate]="assessmentTemplate"
-                   typeaheadOptionField="label"
-                   placeholder="{{'labels.instructional-resource.form.field.assessment.placeholder' | translate}}">
-            <div>
-              <span [hidden]="!assessmentLoading"><i class="fa fa-spinner fa-pulse"></i></span>
-              <span [hidden]="!assessmentNoResults" class="text-info">{{'labels.instructional-resource.assessments.no-results' | translate}}</span>
-            </div>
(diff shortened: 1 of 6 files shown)


wtritch:
ir_create_populated
ir_create_unpopulated
ir_edit
ir_list_organization_info

webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
```diff
@@ -13,94 +13,84 @@ <h4 class="modal-title pull-left">{{ 'labels.instructional-resource.create.title
     <div *ngIf="duplicateResource" class="alert alert-danger">
       {{ 'labels.instructional-resource.create.duplicate' | translate }}
     </div>
-    <form class="form-horizontal">
+    <form>
```
wtritch: The primary changes in this class are just layout restructuring from displaying labels and inputs side-by-side to displaying labels above inputs to allow for longer label length.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1310,9 +1310,10 @@
       },
       "cols": {
         "assessment-label": "Assessment Label(s)",
-        "assessment-name": "Assessment Name",
+        "assessment-name": "System Name",
```
wtritch: "Assessment Label" and "Assessment Name" are very confusing when displayed next to each other.  The "Assessment Name" is basically provided for de-duping in the unlikely case that two assessments with different names share the same label.  I thought that a label that implies "you probably shouldn't recognize this value" seemed like a reasonable change.  @malaffoon ?

malaffoon: Agree, thanks.

esotrope:
I'm wondering if the things like `[Type] State` wouldnt look better as a pill like `[Type|State]` where the left side is dark gray bg and the right side is white.

esotrope:
Do we have that same modal dismissal problem elsewhere?

webapp/src/main/webapp/src/app/admin/instructional-resource/update-instructional-resource.modal.ts
```diff
@@ -25,9 +27,18 @@ export class UpdateInstructionalResourceModal {
...
   private _resource: InstructionalResource = new InstructionalResource();
+  private _subscription: Subscription;
 
   constructor(private modal: BsModalRef,
-              private resourceService: InstructionalResourceService) {
+              private resourceService: InstructionalResourceService,
+              private router: Router) {
+    this._subscription = router.events.filter(e => e instanceof NavigationStart).subscribe(() => {
+      this.cancel();
+    });
```
esotrope: wondering if we shouldn't make a modal baseclass that does this for all of our modals

wtritch: I thought about that, but I think these are the only modals with an auto-complete/back-end communication that can fail out and cause a navigation *while* working with them.

wtritch:
@esotrope w.r.t. the full pill: It looks really strange with the super-long assessment name text.  At least in my opinion.  Pills for me are small label-value pairs that look like a pill rather than an oversized hot dog.

esotrope:
@wtritch x D there we go - introducing the new ui component [the | hotdoggggggggggg]

PR #944, opened 2018-02-21
Re-enable PopupMenu sub menus.

wtritch:
It looks like somewhere along the PopupMenu refactoring we lost the ability to display submenus.  I just moved the trigger class off of the list item and onto the child anchor tag.

image


wtritch:
Merging this now since it's a relatively light change.  Will resolve any comments if/when they appear.

esotrope:
Looks like we lost the right arrow icon when there is a submenu too

wtritch:
@esotrope hmm... since the submenu appears on hover do you think we need the arrow or is it just more visual clutter?

PR #945, opened 2018-02-21
In the My Reports page, convert single-action reports to use a popup …

wtritch:
…menu for consistency with multi-action reports.

image


wtritch:
I left the "non-actionable" reports displaying as plain text rather than try to create some kind of disabled non-menu visualization.

esotrope:
coolness 👍

PR #946, opened 2018-02-21
corrected dropdown item text and tooltip text

dsnyder-fairway:
screen shot 2018-02-21 at 10 18 21 am


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1617,6 +1617,7 @@
     "no-results-found": "No test results are available for the current selection.  If a new school year has begun recently, please check the year selected above.",
     "no-writing-trait-scores": "Writing trait scores are not available for assessments administered prior to 2017-18 school year.",
     "no-writing-trait-scores-for-summative-ica-exams": "Writing trait scores are only currently available for IAB assessments.",
+    "no-responses-for-summative-exams": "{{firstName}}'s Responses is not available is not available for summative assessments.",
```
esotrope: grammar/caps look bad.
I think it should probably just be:
`"Summative assessment responses are not made available by this system."`
Alternatively, we should probably not even show them the menu option no?
@malaffoon @wtritch ?

dsnyder-fairway: Yikes, yeah `...is not available is not available...`. I'll push changes to your suggested text for now

dsnyder-fairway: Also, I think some of the difficulty of removing the menu option is we would need to change the styling of 
screen shot 2018-02-21 at 11 51 19 am
in those cases, since it would not longer contain any items

esotrope: Yeah that's another thing we need to make consistent in the app.
Should non-actionables show as buttons? non-actionable reports don't.
Should we show actions and disable them? or not show disabled actions?
ect...

malaffoon: I think the menu with disabled option is consistent and okay.

PR #949, opened 2018-02-21
enables summative student reports and cleans up report downloader cli…

esotrope:
…ent model

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamReportAssessmentType.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamReportAssessmentType.java
@@ -5,7 +5,8 @@ import org.opentestsystem.rdw.common.model.AssessmentType;
 public enum ExamReportAssessmentType {
 
     ICA(AssessmentType.ICA, "ica-report"),
-    IAB(AssessmentType.IAB, "iab-report");
+    IAB(AssessmentType.IAB, "iab-report"),
+    SUMMATIVE(AssessmentType.SUMMATIVE, "ica-report");
 
     private final AssessmentType assessmentType;
     private final String template;
--- a/webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
@@ -9,6 +9,7 @@ import { InstructionalResourcesService } from "../../instructional-resources.ser
 import { InstructionalResource } from "../../../model/instructional-resources.model";
 import { Observable } from "rxjs/Observable";
 import { PopupMenuAction } from "../../../../shared/menu/popup-menu-action.model";
+import { Utils } from "../../../../shared/support/support";
 
 enum ScoreViewState {
   OVERALL = 1,
@@ -98,7 +99,7 @@ export class ResultsByStudentComponent implements OnInit {
   }
 
   private createActions(): PopupMenuAction[] {
-    let builder = this.actionBuilder.newActions();
+    const builder = this.actionBuilder.newActions();
 
     if (this.assessment.isInterim) {
       builder.withResponses(exam => exam.id, exam => exam.student, exam => exam.schoolYear > this.minimumItemDataYear);
@@ -110,12 +111,13 @@ export class ResultsByStudentComponent implements OnInit {
         () => this.assessment.type,
         exam => exam.student,
         exam => {
-          let downloader: StudentReportDownloadComponent = this.reportDownloader;
-          let options: ReportOptions = downloader.options;
-          let subject = this.assessment.assessmentSubjectType;
 
-          options.assessmentType = this.assessment.type;
-          options.subject = subject;
+          const downloader: StudentReportDownloadComponent = this.reportDownloader;
+          const options: ReportOptions = downloader.options;
+          const subject = this.assessment.assessmentSubjectType;
+
+          options.assessmentType = Utils.toAssessmentTypeCode(this.assessment.type);
+          options.subject = Utils.toSubjectCode(subject);
           options.schoolYear = exam.schoolYear;
 
           downloader.student = exam.student;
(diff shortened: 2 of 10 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamReportAssessmentType.java
```diff
@@ -5,7 +5,8 @@
 public enum ExamReportAssessmentType {
 
     ICA(AssessmentType.ICA, "ica-report"),
-    IAB(AssessmentType.IAB, "iab-report");
+    IAB(AssessmentType.IAB, "iab-report"),
+    SUMMATIVE(AssessmentType.SUMMATIVE, "ica-report");
```
esotrope: enables summative reports and tells the app to use the ICA report template (which we should probably rename to ica-summative-report

webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
```diff
@@ -110,12 +111,13 @@ export class ResultsByStudentComponent implements OnInit {
...
-          let subject = this.assessment.assessmentSubjectType;
 
-          options.assessmentType = this.assessment.type;
-          options.subject = subject;
+          const downloader: StudentReportDownloadComponent = this.reportDownloader;
+          const options: ReportOptions = downloader.options;
+          const subject = this.assessment.assessmentSubjectType;
+
+          options.assessmentType = Utils.toAssessmentTypeCode(this.assessment.type);
+          options.subject = Utils.toSubjectCode(subject);
```
esotrope: real change is to pass string codes here instead of the client enums (step away from using the enums)

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -11,7 +11,10 @@
     <div class="modal-content">
       <div class="modal-header">
         <h4 class="modal-title pull-left">{{title}}</h4>
-        <button type="button" class="close pull-right" aria-label="Close" (click)="modal.hide()">
+        <button type="button"
```
esotrope: formatted this file and changed translation codes to new code locations where i could

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -48,29 +46,23 @@ export abstract class ReportDownloadComponent implements OnInit {
...
-  onShowInternal(event: any) {
-    this.onShow.emit(event);
-    this.options.name = this.generateName();
-  }
-
-  AssessmentType: any = AssessmentType;
-  assessmentTypes: AssessmentType[] = [ null, AssessmentType.IAB, AssessmentType.ICA ];
-  subjectTypes: AssessmentSubjectType[] = [ null, AssessmentSubjectType.MATH, AssessmentSubjectType.ELA ];
+  assessmentTypes: string[] = [ null, 'ica', 'iab', 'sum' ];
+  subjectTypes: string[] = [ null, 'Math', 'ELA' ];
```
esotrope: changed to string codes

webapp/src/main/webapp/src/app/report/report-download.component.ts
```diff
@@ -97,6 +89,11 @@ export abstract class ReportDownloadComponent implements OnInit {
       );
   }
 
+  onShowInternal(event: any) {
+    this.onShow.emit(event);
+    this.options.name = this.generateName();
+  }
```
esotrope: just moved this method

webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -181,4 +184,22 @@ export class Utils {
     return false;
   }
 
+  // TODO stop using subject/assessment type enums. use codes instead.
```
esotrope: introduced this intentionally smelly code - idea is to compel us to refactor out the assessment type and subject enums

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamReportAssessmentType.java
agorina: Yeah, we should rename it. If you are afraid of a ripple effect, maybe at least a comment explaining what is going on. Otherwise it is very tempting to 'fix'.

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -11,7 +11,10 @@
...
+                class="close pull-right"
+                aria-label="Close"
```
wtritch: hmm... aria-label should be translated...

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -94,7 +127,7 @@ <h4 class="modal-title pull-left">{{title}}</h4>
               </label>
               <sb-radio-button-group name="accommodations-toggle"
                          [(ngModel)]="options.accommodationsVisible"
-                         [disabled]="options.assessmentType === AssessmentType.IAB"
+                         [disabled]="options.assessmentType === 'iab'"
```
wtritch: If we're basing logic on a particular assessment type, I'd still prefer to use a constant rather than sprinkle 'iab', 'ica', or 'sum' strings all over =/

esotrope: like the other comment - this is simply a step in a direction - we're not there yet

webapp/src/main/webapp/src/app/report/report-download.component.html
esotrope: I'm not going to fret about it until we get do an accessibility/translation prep sweep of the UIl

webapp/src/main/webapp/src/app/report/report-download.component.html
wtritch: That being said, I think the root problem is the existence of this IF on type to being with... ideally we'd instead have a backend-backed service that we could ask: Does this (assessment type, subject) combination support accommodations?

esotrope: agreed : j

PR #950, opened 2018-02-21
updated info for "completeness"

dsnyder-fairway:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -88,7 +88,7 @@
         },
         "completeness": {
           "label": "Completeness",
-          "info": "Indicates whether the test was completed or was partially complete."
+          "info": "A test is considered complete when the student provides an answer for every question."
```
dsnyder-fairway: didn't consolidate these keys since `adv-filters` uses `labels.filters. + whatever` as the text

esotrope: If it is common text lets take the op to move it to `common.description.completeness` or `common.completeness-info`

PR #952, opened 2018-02-21
 consolidated completeness text into `common`

dsnyder-fairway:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -213,6 +209,12 @@
       "all": "All",
       "none": "None"
     },
+    "completeness": {
+      "label": "Completeness",
+      "info": "A test is considered complete when the student provides an answer for every question.",
+      "Partial": "Partial",
+      "Complete": "Complete"
+    },
```
esotrope: the "common.completeness" prefix is meant to accept only the "common.completeness.{{completenessCode}}"
I'd suggest making a new prefix like "common.completeness-form-control." since label/info are associated with form controls

PR #953, opened 2018-02-21
Feature/join report review and name

esotrope:
Moves name field into review section.

image


esotrope:
I'm really tempted to tuck the table preview away behind a tab/button on the report review section - that way the report name and submit button are the bottom of the form as one would expect.

PR #954, opened 2018-02-22
prevent changing how "Show value as" / "Show achievement levels" from…

dsnyder-fairway:
… misaligning when "Grouped" is chosen
a
b
c
d

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -40,7 +40,8 @@
       ]">
       <button class="btn btn-success btn-sm green ml-sm mt-xs"
               [disabled]="submissionSubscription"
-              (click)="onGenerateButtonClick()">{{'aggregate-reports.form.submit' | translate}}</button>
+              (click)="onGenerateButtonClick()">{{'aggregate-reports.form.submit' | translate}}
+      </button>
     </scroll-nav>
   </div>
 
@@ -77,7 +78,8 @@
                        formControlName="organizations"
                        (ngModelChange)="onSettingsChange()">
                 <div *ngIf="showErrors(organizationsControl)">
-                  <p *ngIf="organizationsControl.errors.invalid" class="help-block small red">{{organizationsControl.errors.invalid.messageId | translate}}</p>
+                  <p *ngIf="organizationsControl.errors.invalid" class="help-block small red">
+                    {{organizationsControl.errors.invalid.messageId | translate}}</p>
                 </div>
               </div>
             </div>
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -40,7 +40,8 @@ <h2 class="h3 gray-darkest">{{'aggregate-reports.form.heading' | translate}}</h2
       ]">
       <button class="btn btn-success btn-sm green ml-sm mt-xs"
               [disabled]="submissionSubscription"
-              (click)="onGenerateButtonClick()">{{'aggregate-reports.form.submit' | translate}}</button>
+              (click)="onGenerateButtonClick()">{{'aggregate-reports.form.submit' | translate}}
```
esotrope: really hard to tell what changed in these files - can you point it out?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -568,24 +573,27 @@ <h3>{{'aggregate-reports.form.section.preview.heading' | translate}} </h3>
...
-                <label class="mr-xs">{{'common.performance-level-display-type-input-label' | translate}}</label>
-                <sb-radio-button-group [options]="options.performanceLevelDisplayTypes"
-                                       [(ngModel)]="settings.performanceLevelDisplayType"
-                                       buttonGroupStyles="btn-group-xs"
-                                       buttonStyles="btn-default"
-                                       analyticsCategory="AggregateQueryBuilder"
-                                       analyticsEvent="Change"></sb-radio-button-group>
-              </span>
+              <div class="pull-right flex-child">
+                <span>
```
dsnyder-fairway: introduced a `<div class="pull-right flex-child">` that contains a `<span>` and a `<span class="ml-xs">`. Before this had no such `div` and contained two `<span class="pull-right flex-child ml-xs">`

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.html
```diff
@@ -73,24 +81,26 @@ <h3 class="h3 label-group green pull-left">
...
-              <span class="pull-right flex-child ml-xs">
-                <label class="mr-xs">{{'common.performance-level-display-type-input-label' | translate}}</label>
-                <sb-radio-button-group [options]="displayOptions.performanceLevelDisplayTypes"
-                                       [(ngModel)]="table.performanceLevelDisplayType"
-                                       buttonGroupStyles="btn-group-xs"
-                                       buttonStyles="btn-default"
-                                       analyticsCategory="AggregateReportControl"
-                                       analyticsEvent="Change"></sb-radio-button-group>
-              </span>
+              <div class="pull-right flex-child">
```
dsnyder-fairway: introduced a <div class="pull-right flex-child"> that contains a <span> and a <span class="ml-xs">. Before this had no such div and contained two <span class="pull-right flex-child ml-xs">

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
esotrope: hm. can we try:
```html
<div class="flex-children">
  <div class="flex-child grow">column order form control</div>
  <div class="flex-child">performance level form control</div>
  <div class="flex-child">value form control</div>
</div>
```

wtritch: Careful with display:grid.  IE11 needs special prefixed properties and explicit column/row membership

aaronsleeper:
Please add an autoprefixer to your build to protect against such things.
See me if you have questions.

Good point. Thanks Will.

dsnyder-fairway:
moved out of sprint

PR #956, opened 2018-02-22
fixes build and adds student response page header fixes

esotrope:


webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -1,6 +1,6 @@
-<page-heading>
+<page-heading *ngIf="examHistory">
```
esotrope: the fix

PR #958, opened 2018-02-22
only allow for includeState aggregate queries when assessment type is…

dsnyder-fairway:
… summative

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
@@ -7,6 +7,7 @@ import java.util.Set;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.google.common.collect.Sets.newHashSet;
+import static org.opentestsystem.rdw.common.model.AssessmentType.SUMMATIVE;
 
 /**
  * Represents a query for the custom aggregate report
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
```diff
@@ -184,7 +185,7 @@ public ReportQuery build() {
             query.subjectCodes = subjectCodes;
             query.schoolYears = schoolYears;
             query.assessmentGradeCodes = assessmentGradeCodes;
-            query.includeState = includeState;
+            query.includeState = includeState && SUMMATIVE.code().equals(assessmentTypeCode);
```
agorina: I think our philosophy is that our builders do not have any business logic. 
I would put this check into `checkIsValid` method and make sure UI never calls it with the invalid parameters.

agorina:
Is it correct to assume that UI change is coming too?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportQuery.java
malaffoon: 👍

dsnyder-fairway:
I didn't think UI changes was needed for this, since we the checkbox states its for summative only. My idea was that we would ignore the box being checked and default it to false in the cases where "Include State results (Summative only)" was checked for non-summative aggregate reports.

PR #959, opened 2018-02-22
component hashtag had to be in the same template if-block

esotrope:


webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -115,7 +115,7 @@ export class GroupResultsComponent implements OnInit {
    *
    * @param downloader
    */
-  private initializeDownloader(downloader: GroupReportDownloadComponent): void {
+  initializeDownloader(downloader: GroupReportDownloadComponent): void {
```
esotrope: no idea how these private methods worked before - also no idea how my shuffling of html around made it not work anymore : /

webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -1,82 +1,92 @@
-<page-heading *ngIf="examHistory">
```
esotrope: another fix - the #reportDownloader was in a different if-block so the popover trigger button couldn't find it

PR #960, opened 2018-02-22
Update the instructional resource creation modal to not change height…

wtritch:
… on key press.

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
+++ b/webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
@@ -16,7 +16,10 @@
     <form>
       <fieldset>
         <div class="form-group mb-xs">
-          <label for="assessment-label" class="control-label">{{'labels.instructional-resource.form.assessment-label' | translate}}</label>
+          <label for="assessment-label" class="control-label">
+            {{'labels.instructional-resource.form.assessment-label' | translate}}
+          </label>
+          <span [hidden]="!assessmentLoading"><i class="fa fa-spinner fa-pulse"></i></span>
           <input class="form-control"
                  id="assessment-label"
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
```diff
@@ -16,7 +16,10 @@ <h4 class="modal-title pull-left">{{ 'labels.instructional-resource.create.title
     <form>
       <fieldset>
         <div class="form-group mb-xs">
-          <label for="assessment-label" class="control-label">{{'labels.instructional-resource.form.assessment-label' | translate}}</label>
+          <label for="assessment-label" class="control-label">
+            {{'labels.instructional-resource.form.assessment-label' | translate}}
+          </label>
+          <span [hidden]="!assessmentLoading"><i class="fa fa-spinner fa-pulse"></i></span>
```
wtritch: moved the "loading" spinner that appears and disappears on user key press next to the form label so it's both visible and doesn't cause the display to expand/shrink on every auto-completion input keypress

esotrope: 👍 had a workaround for that in `organization-typeahead.ts`. wonder if the [typeahead] directive has a spinner hook - i think we can now flag an option to put an (x) clear button in the text inputs now if we want btw - i raised the issue on their github and they closed it as fixed

wtritch: Neat.  I wonder if we should create a 'nice to have' jira to go give our visualization controls a once-over for functionality...  I'd love if the autocomplete showed the spinner/no results as a popup and we didn't have to dedicate space for them around the input.

PR #961, opened 2018-02-22
Update the aggregate report component to initialize spinner state on …

wtritch:
…init rather than in constructor.


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -77,7 +77,6 @@ export class AggregateReportComponent implements OnDestroy {
         options: this.options,
         settings: settings
       });
-    this.updateViewState();
```
wtritch: oops.  The fallout of this method call was attempting to set the spinner's loading property before the spinner was initialized.  Just moved to ngOnInit().

PR #963, opened 2018-02-22
disable state results on non-summative assessment types

dsnyder-fairway:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -378,6 +378,9 @@ export class AggregateReportFormComponent {
       settings: this.settings
     };
 
+    // set include state results to false when summative fields are disabled
+    this.summary.settings.includeStateResults = this.summary.settings.includeStateResults && !this.summativeFieldsDisabled;
```
dsnyder-fairway: I could not find a way around this :(. Basically, when disabling and unchecking the checkbox, we want to have the includeStateResults to be false

esotrope: we can include that logic in the request mapper to just ignore the flag if the asmt type is not summative

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -93,6 +93,8 @@ <h3>{{'aggregate-reports.form.section.organization.heading' | translate}}</h3>
                     <div class="col-sm-6">
                       <label class="inline-label">
                         <input type="checkbox"
+                               [disabled]="summativeFieldsDisabled"
+                               [checked]="!summativeFieldsDisabled"
```
esotrope: This is not the behavior we want.
I'd suggest putting this logic in the `aggregate-report-request.mapper.ts` `map` function

dsnyder-fairway: what UI changes should be done? I went ahead and disabled this as a response to https://github.com/SmarterApp/RDW_Reporting/pull/958#issuecomment-367554328

esotrope: I don't think a ui behavior change is necessary. we just need to make sure we don't send the flag when the user has asmt type 'sum' selected. I would have a different opinion if the flag checkbox appeared after the assessment type select in the form though x J

dsnyder-fairway: part of the difficulty with moving this to aggregate `aggregate-report-request.mapper.ts` `map` is the query is fine going through for an IAB. Include state results defaults to false even when the checkbox is selected, however, when going back to modifying the query, the checkbox is unselected.

dsnyder-fairway: while really, that isn't a problem, since both are fine (State results (**Summative Only**)) it is a little weird

dsnyder-fairway:
@esotrope changes have been made, just be aware that when a user goes back after viewing an aggregate report query for interim exams the option to `State results (Summative Only)` will appear unselected, even if it was previously selected.

esotrope:
@dsnyder-fairway good point - when they come back we would want them to see the checkbox checked. I think the best way to do this might be then to implement the "ignoring" of the flag on the backend.
@agorina @wtritch thoughts?

dsnyder-fairway:
@esotrope ignoring the flag on the backend would have been my first pr for this. Which I would be happy to do 😄 
https://github.com/SmarterApp/RDW_Reporting/pull/958

Maybe not in a builder...but somewhere else

dsnyder-fairway:
@esotrope @agorina @wtritch I think we can either:

1. 🎉 Do something like what was done here https://github.com/SmarterApp/RDW_Reporting/pull/963/commits/2ea29504bba4a882202d16a081ce2c12cb1042bd. I think this is more of what @agorina had in mind. Server throws an error if we receive this input and adjust the UI to avoid sending this input. Or, 

2. 😄 Allow them to set this, but only act on it if the value is set AND they are doing this for summative assessments. Effectively, that is https://github.com/SmarterApp/RDW_Reporting/pull/958.

I propose vote by emoji

agorina:
I suggest we chat about it as a group. It looks like there is some complexity that we are overlooking.

dsnyder-fairway:
@esotrope @agorina @wtritch reverted previous commit as discussed during our morning scrum

PR #965, opened 2018-02-23
reversed order of scale scores from exceeded -> did not meet / above …

dsnyder-fairway:
…-> below standard

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
@@ -30,7 +30,7 @@
     <div class="col-md-7">
       <strong>{{ 'labels.assessments.student-score-distribution.title' | translate }}</strong>
       <div class="well">
-        <div *ngFor="let performanceLevel of performanceLevels; let i = index" class="score-distribution-container">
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -30,7 +30,7 @@
     <div class="col-md-7">
       <strong>{{ 'labels.assessments.student-score-distribution.title' | translate }}</strong>
       <div class="well">
-        <div *ngFor="let performanceLevel of performanceLevels; let i = index" class="score-distribution-container">
+        <div *ngFor="let performanceLevel of performanceLevels.reverse(); let i = index" class="score-distribution-container">
```
malaffoon: This messes up the colors. `i` is used to get the color for the levels and this effectively flips them all.

malaffoon:
See comments about colors.

dsnyder-fairway:
you are correct, I'll fix this up

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
esotrope: @dsnyder-fairway while you are in there try and use the new ColorService method i think i deprecated the old one

PR #967, opened 2018-02-23
fixed the colors

dsnyder-fairway:


webapp/src/main/webapp/src/app/shared/color.service.ts
```diff
@@ -13,9 +13,9 @@ const Colors: string[] = [
 ];
 
 const Pallets: any[] = [
-  [ 'blue-dark', 'blue-dark aqua', 'aqua' ],
-  [ 'maroon', 'gray-darkest', 'green-dark', 'blue-dark' ],
-  [ 'maroon', 'gray-darkest', 'green-dark', 'blue-dark' ],
+  [ 'aqua', 'blue-dark aqua', 'blue-dark' ],
+  [ 'blue-dark', 'green-dark', 'gray-darkest', 'maroon' ],
+  [ 'blue-dark', 'green-dark', 'gray-darkest', 'maroon' ],
```
esotrope: this will change the color order for everything everywhere - i think what you want to do is call this method with the right params:
```typescript
getPerformanceLevelColorsByAssessmentTypeCode(code: string, performanceLevel: number): string
```

esotrope: performanceLevel is 1-based * - should add that to the ts doc too

malaffoon: Personally, i would change colorService.getPerformanceLevelColor() to accept the performanceLevel object or something, but i don't know the innards of all that stuff; i.e. i don't know if ExamStatisticsLevel.id is an indexable value or ???

esotrope: "performanceLevel" in my vocab at least means a 1-based number referring to where the score fell. in this chart though i think it term is being bent - it really refers to a performanceLevelAggregate and might better be named that

dsnyder-fairway:
@esotrope we use enums for AssessmentType in the average-scale-score. Looking at were else we have these enum values being used there are a few other services. I'd like to first tackle this and then maybe I can see about converting each of those services individually, finally refactoring this as well

dsnyder-fairway:
alternatively I can put in a wrapper that takes the AssessmentType, and call `getPerformanceLevelColorsByAssessmentTypeCode` after finding the correct string from the enum.

dsnyder-fairway:
and `getPerformanceLevelColor(assessment: Assessment, performanceLevel: number): string` is used only from `average-scale-score.component.html`

webapp/src/main/webapp/src/app/shared/color.service.ts
```diff
@@ -47,16 +47,23 @@ export class ColorService {
   }
 
   /**
-   * Retrieves the color for the performance level (0-based)
+   * Retrieves the color for the performance level (1-based)
    *
    * @param {Assessment} assessment
-   * @param {number} performanceLevel (0-based)
```
esotrope: this method actually accepts a 0-based "performance level"
I suggest we keep the existing code + docs on it and add @deprecated

PR #969, opened 2018-02-23
fixes duplicate calls to /districts

esotrope:


webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -2,16 +2,15 @@ import { Injectable } from "@angular/core";
 import { UserService } from "../user/user.service";
 import { School } from "../user/model/school.model";
 import { Observable } from "rxjs/Observable";
-import "rxjs/add/observable/forkJoin";
 import { CachingDataService } from "../shared/data/caching-data.service";
+import { ReportingServiceRoute } from "../shared/service-route";
+import { forkJoin } from "rxjs/observable/forkJoin";
+import { map, publishReplay, refCount } from "rxjs/operators";
```
esotrope: learned how to do things the rxjs 5 way.
They now recommend importing the operators like this and then using `Observable.let` or `Observable.pipe` instance methods to chain things together.
(e.g. observable.map().catch() now becomes observable.pipe(map(), catch())
This allows us to omit the "rxjs/add/..." imports and will improve build time if implemented everywhere - i plan on making a jira for this

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -23,26 +22,37 @@ export class OrganizationService {
...
-        return user.schools.map(school => {
-          school.districtName = districts.get(school.districtId);
+    return forkJoin(
+      this.getSchools(),
+      this.getDistrictNamesById()
+    ).pipe(
+      map(([ schools, districtNamesById ]) => {
+        return schools.map(school => {
+          school.districtName = districtNamesById.get(school.districtId);
           return school;
```
esotrope: for the record side effects inside of an array map method are a sin.
i'm not going to fix this now though because i don't know if these side effects are being relied on elsewhere where the user.schools are referenced : [

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -2,16 +2,15 @@ import { Injectable } from "@angular/core";
...
+import { ReportingServiceRoute } from "../shared/service-route";
+import { forkJoin } from "rxjs/observable/forkJoin";
+import { map, publishReplay, refCount } from "rxjs/operators";
 
-const ServiceRoute = '/reporting-service';
 
 @Injectable()
 export class OrganizationService {
 
-  private schoolsWithDistricts: Observable<School[]>;
```
esotrope: caching an observable doesn't really work i guess. rookie mistake by a past me x l

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -23,26 +22,37 @@ export class OrganizationService {
...
+      publishReplay(1),
+      refCount()
+    );
+  }
+
+  private getSchools(): Observable<School[]> {
+    // TODO use /reporting-service/organizations/schools instead of user payload to untie user from reporting-service
+    return this.userService.getCurrentUser().pipe(
+      map(user => user.schools)
+    );
```
wtritch: huh... should this be replayed as well?

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
```diff
@@ -23,26 +22,37 @@ export class OrganizationService {
...
+      map(([ schools, districtNamesById ]) => {
+        return schools.map(school => {
+          school.districtName = districtNamesById.get(school.districtId);
           return school;
         });
-      });
+      }),
+      // when combined these operators act as a cache of the first valid result of the observable
+      publishReplay(1),
+      refCount()
```
wtritch: neat.  I'm going to have to go read up on this... what's refCount() do/for?

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
esotrope: for now yeah we could - eventually we will make a call for /schools and that observable result will be cached

webapp/src/main/webapp/src/app/school-grade/organization.service.ts
esotrope: it counts sub references so it can know when to cleanup when subs go to zero

PR #970, opened 2018-02-23
DWR-1328 Added percentile data to IAB/PDF report model

agorina:
Note that there is another JIRA to add it to the report. This just loads the data for it.


report-processor/src/main/resources/application.sql.yml
```diff
@@ -280,3 +285,7 @@ sql:
             )
           )
         )
+
+      percentileLeftJoin: >-
+          LEFT JOIN percentile p ON e.asmt_id = p.asmt_id AND e.completed_at >= p.start_date AND e.completed_at <= p.end_date
+          LEFT JOIN percentile_score ps ON ps.percentile_id = p.id AND e.scale_score >= ps.min_inclusive AND e.scale_score < ps.max_exclusive
```
agorina: I know that this seems too small to qualify for the snippet. 
The only reason I did it this way is to avoid repeating tests for groups vs. non-group SQL. There are quite a few test conditions in here for all the >=, <= .

report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -11,7 +11,7 @@
 
             <div class="row">
                 <div class="col-xs-6"
-                     th:with="assessment=${report.examsByAssessment.firstEntry().key},exam=${report.examsByAssessment.firstEntry().value[0]}">
+                     th:with="assessment=${report.examsByAssessment.firstEntry().key},exam=${report.examsByAssessment.firstEntry().value[0].exam}">
                     <div>
```
agorina: @wtritch , thank you for your help here!

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ExamWithPercentile.java
```diff
@@ -0,0 +1,71 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * Represents a single assessment examination event with percentile data
+ */
+public class ExamWithPercentile {
+    private Exam exam;
+    private Integer percentileMean;
+    private Integer percentileRank;
```
esotrope: i'd be tempted to wrap these related fields into a parent object "ScaleScorePercentile" or something

report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -94,8 +94,8 @@
                 <div class="well gray-lightest assessment-instances" data-width="40">
                     <div class="well-body">
                         <ul class="previous-results">
-                            <li th:with="assessment=${entry.key}" th:each="exam : ${entry.value}"><span
-                                    th:if="${examStat.index} > 0"><span class="previous-result"><span class="date"
+                            <li th:each="examWrapper : ${entry.value}" th:with="assessment=${entry.key}, exam=${examWrapper.exam}"  ><span
+                                    th:if="${examWrapperStat.index} > 0"><span class="previous-result"><span class="date"
```
agorina: This `...Stat` thing is a feature of thymeleaf. It returns iteration status properties.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepositoryIT.java
```diff
@@ -53,14 +58,17 @@
...
+    @Autowired
+    private JdbcTemplate template;
+
     @After
     public void cleanup() {
         tenantProperties.setTransferAccessEnabled(false);
     }
 
     @Test
-    public void itShouldFindAllIabReportsForStudentsBySchoolAndYear() throws Exception {
```
agorina: Unrelated, there is not need for the Exception here.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ExamWithPercentile.java
```diff
@@ -0,0 +1,71 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * Represents a single assessment examination event with percentile data
+ */
+public class ExamWithPercentile {
```
malaffoon: super minor nit: this isn't actually an exam-with-percentile it is exam-and-percentile. ExamWithPercentile would extend Exam while ExamAndPercentile would contain Exam. 

Partly i point this out to ask, did we want to extend exam instead of containing it?

report-processor/src/main/resources/application.sql.yml
malaffoon: Just checking on date matching logic: is it correct that it is >=start and <=end? Just worried that might match two rows depending on how the data is set up and validated.

report-processor/src/main/resources/templates/iab-body.html
malaffoon: if ExamWithPercentile extends Exam you wouldn't need this?

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepositoryIT.java
```diff
@@ -302,13 +310,135 @@ public void itShouldFindAllIabReportsForAStudentByGroupPiiRightsRespectingDisabl
...
+        final Map<Long, ExamWithPercentile> examsByID = exams.stream().collect(Collectors.toMap(e -> (e.getExam().getId()), e -> e));
+        return examsByID;
+    }
+
     private Collection<Long> getExamIds(final Map<Long, Map<Subject, IabReport>> reports) {
         return reports.values().stream()
                 .flatMap(map -> map.values().stream())
                 .map(IabReport::getExamsByAssessment)
                 .flatMap(map -> map.values().stream())
                 .flatMap(Collection::stream)
```
malaffoon: i'm always suspicious when i see streams constructed inside streams; since this is in a test i'm not too worried about it but ...

report-processor/src/main/resources/application.sql.yml
agorina: @malaffoon, I was hoping you would know the answer on this. I copied it from Chris's example assuming that this is what we want.

@chdietz , could you please clarify?

report-processor/src/main/resources/templates/iab-body.html
agorina: No, we would not.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ExamWithPercentile.java
agorina: And I went back and forth on this, and was encouraged by people who are more familiar with this project that this is a preferable way to do it.(  @wtritch , not to point the fingers :) ).

wtritch: Well shoot.  Extension would make this easier to do...  In general I tend to prefer wrappers because it helps keep models cleaner, but not having to risk changing the templates would be a nice win, too...

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepositoryIT.java
wtritch: The problem is that we're reporting on:
multiple students:
  for each student -> multiple subjects:
      for each subject -> multiple assessments:
        for each IAB assessment -> multiple exams

and all the nested one-to-many relationships can make stream processing a little funky =/

report-processor/src/main/resources/application.sql.yml
malaffoon: Poking around in the code a bit ...

It looks like the intent is that the start-date and end-dates are inclusive so this query is correct. However, there is a bug in the overlap detection query that would allow the start/end dates for different entries to be the same. I guess i'd recommend this query stay as is, and i'll write a test that demonstrates the overlap detection bug.

malaffoon: All that said, there may be another problem. `completed_at` is a timestamp and the start/end dates are dates. Let's say i have a percentile with start 1/1/2018 and end 2/28/2018; meant to be all tests completed in January and February. If the completed_at is `2018-02-28T15:30:00`, is that less than or equal to the end date?

agorina: Would it make sense to convert `complated_at` to date in this comparison: `DATE(completed_at)`?

malaffoon: Yes, i think that would be the right change.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepositoryIT.java
```diff
@@ -302,13 +310,135 @@ public void itShouldFindAllIabReportsForAStudentByGroupPiiRightsRespectingDisabl
...
+                .queryPermissionType(Unknown)
+                .gradeId(-2)
+                .build();
+        final Map<Long, ExamWithPercentile> examsByID =
+                toExamById(repository.findAllForStudentsByExamFilter(individual(statewide()), of(-1L), query), -1L);
+
+        //find expected mean score based on the start date inclusion
+        final Integer meanScoreExam4 = template.queryForObject("SELECT round(mean) " +
+                "FROM percentile " +
+                "WHERE start_date = (SELECT DATE(completed_at) FROM exam WHERE id = -4)", Integer.class);
```
agorina: I adjusted tests too.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepositoryIT.java
malaffoon: So the real problem is too many tests. We should just have The One Assessment. Each student takes it once. Don't know what all this fuss is about  ;)

PR #971, opened 2018-02-23
 prevent switching order after changing between percent / number

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -24,6 +24,8 @@ export class AverageScaleScoreComponent {
 
   @Input()
   set statistics(value: ExamStatistics) {
+    value.percents = value.percents.reverse();
+    value.levels = value.levels.reverse();
```
malaffoon: Are the colors reversed as well? I wasn't sure from the last review whether that happened or not.

Is the order of these values assumed to be "ascending" anywhere else?

dsnyder-fairway:
saw a comment to jira. Will have changes shortly

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
dsnyder-fairway: colors were reversed in last review and this one. I just saw your comment to the jira. I'll give that a shot

dsnyder-fairway: the order of this is not used anywhere else

malaffoon:
The pipe would only be used in the template, where the reverse() being called over and over was not appropriate. The way you have it (adjusting the order when setting the values) should work fine.

Again, assuming the order wasn't being relied upon elsewhere; which you confirmed.

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
malaffoon: Might want to add a code comment that indicates why we're reversing them. Something like `// reverse so "good" comes before "bad"; that order is assumed in the UI`

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -24,6 +24,9 @@ export class AverageScaleScoreComponent {
 
   @Input()
   set statistics(value: ExamStatistics) {
+    // reverse percents and levels so scale score statistics appear in descending order ("good" statistics levels comes before "bad")
+    value.percents = value.percents.reverse();
+    value.levels = value.levels.reverse();
```
esotrope: `Array.reverse()` is an in-place op. can you add a `.concat()` in between to make a copy of the array before reversing?

PR #973, opened 2018-02-23
separated styling of indicators between dropdown menus and row expans…

dsnyder-fairway:
…ion. Made the styling consistent with the effect of the row expansion (up means not expanded, down means expanded)

#  unchanged
screen shot 2018-02-23 at 12 33 33 pm
screen shot 2018-02-23 at 12 33 43 pm
screen shot 2018-02-23 at 12 34 17 pm

# changed
screen shot 2018-02-23 at 12 33 59 pm
screen shot 2018-02-23 at 12 34 08 pm
screen shot 2018-02-23 at 12 34 27 pm
screen shot 2018-02-23 at 12 34 35 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -362,7 +362,7 @@
               <div class="flex-child">
                 <button class="btn btn-default btn-sm icon-only pull-right"
                         (click)="onAdvancedFiltersExpanderButtonClick()">
-                  <i class="fa" [ngClass]="showAdvancedFilters ? 'fa-caret-square-o-down' : 'fa-caret-square-o-up'"></i>
+                  <i class="fa" [ngClass]="showAdvancedFilters ? 'fa-caret-square-o-up' : 'fa-caret-square-o-down'"></i>
                 </button>
               </div>
             </div>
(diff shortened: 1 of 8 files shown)


dsnyder-fairway:
justification for why we have the arrows point to where the data lies:
we had a few areas where this happened, but for instance, we had it here
screen shot 2018-02-23 at 12 44 27 pm

malaffoon:
I will yield the remainder of my time and comment space to @wtritch and @esotrope. The goal is to make it at least not wrong, and consistent. My only ask is that, when complete, it will be obvious to any developer in the code which symbols to use for any text/button menu, row expansion, etc.

wtritch:
So it looks like we've decided that icons show current state, not the state that will happen when pressing the button.  I'm ok with either way, just so long as we're consistent.  (e.g. in a web conference the mute button generally shows your current state rather than what will happen when you push the button.)

esotrope:
@wtritch i made a comment before that suggested we make everything show current state but this could not be done easily since there is text tied to the arrow buttons like "expand / collapse" so i think the icons are good showing eventual state and it will be consistent in the app 👍

dsnyder-fairway:
Before we had [Hide ^] which we could argue may hide the above, though you do have to first click on show to see the [Hide ^]

esotrope:
@dsnyder-fairway agreed - this is one of those things that is super subjective unfortunately. i think the reason i feel the way i do is because most things that require an expander in our app default closed.

dsnyder-fairway:
Okay, I'll make those changes

PR #974, opened 2018-02-23
Bugfix/handle all enrolled grades

wtritch:
… non-assessment grades

We were previously ordering Enrolled Grades as Assessment Grades (which wasn't accounting for grade 9, 10, 1, etc.)
This PR attempts to parse the grade code as an integer and order results by number value.
If the grade code cannot be parsed as a number, it is displayed last.  The order of multiple non-numeric enrolled grades is unspecified.
(This is not an issue with the current system that hosts grades 3-12, but should be resolved in the future.)

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -419,10 +419,20 @@ export class AggregateReportTableComponent implements OnInit {
       .on((item: AggregateReportItem) => `${item.dimension.type}.${item.dimension.code}`)
       .compare;
 
+    // Attempt to sort based upon the enrolled grade code as a number ("01", "02", "KG", "UG", etc)
+    // If the code cannot be parsed as a number, the order is undefined
+    // TODO we should have a specific ordering for all grade codes, although the system only currently uses "03" - "12"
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -140,32 +140,11 @@ export class AggregateReportTableComponent implements OnInit {
...
-
-      const dimensionTypeAndCodeRanking = table.options.dimensionTypes.reduce((ranking, dimensionType) => {
-        return ranking.concat(
-          (dimensionOptionsByDimensionType[ dimensionType ] || []).map(dimensionCode => `${dimensionType}.${dimensionCode}`)
-        );
-      }, []);
-
-      const dimensionOrdering = ordering(ranking(
-        [OverallDimensionType, ...dimensionTypeAndCodeRanking]
-      )).on((item: AggregateReportItem) => `${item.dimension.type}.${item.dimension.code}`);
```
wtritch: Moved this into a separate private method

wtritch:
@esotrope I'm really not sure... I'm not convinced that this is something "report-scoped" since we display enrolled grades elsewhere in the app and would like the sorting to be consistent. (e.g. as a column in the school/grade and group results tables)

PR #975, opened 2018-02-23
DWR-1265 API for percentile data

agorina:
@malaffoon , @esotrope  and @wtritch - Please take another look. Thank you!

### DIFF
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
@@ -1,0 +1,25 @@
+package org.opentestsystem.rdw.reporting.assessment;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RestController;
+
+import javax.validation.constraints.NotNull;
+import java.util.List;
+
+@RestController
+class AssessmentPercentileController {
+
+    private final AssessmentPercentileService service;
+
+    @Autowired
+    public AssessmentPercentileController(@NotNull final AssessmentPercentileService service) {
+        this.service = service;
+    }
+
+    @GetMapping("/assessment/{id}/percentile")
+    public List<Percentile> getPercentile(@PathVariable final int id) {
+        return service.findAllForAssessment(id);
+    }
+}
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
@@ -2,13 +2,14 @@ package org.opentestsystem.rdw.reporting.assessment;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.GetMapping;
-import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
 import javax.validation.constraints.NotNull;
 import java.util.List;
 
 @RestController
+@RequestMapping("/assessments-percentiles")
 class AssessmentPercentileController {
 
     private final AssessmentPercentileService service;
@@ -18,8 +19,8 @@ class AssessmentPercentileController {
         this.service = service;
     }
 
-    @GetMapping("/assessment/{id}/percentile")
-    public List<Percentile> getPercentile(@PathVariable final int id) {
-        return service.findAllForAssessment(id);
+    @GetMapping
+    public List<Percentile> getPercentile(final AssessmentPercentileRequest request) {
+        return service.findAllForAssessment(request);
     }
 }
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileRequestArgumentResolver.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileRequestArgumentResolver.java
@@ -21,16 +21,10 @@ public class AssessmentPercentileRequestArgumentResolver implements HandlerMetho
 
     @Override
     public Object resolveArgument(final MethodParameter parameter, final ModelAndViewContainer mavContainer, final NativeWebRequest webRequest, final WebDataBinderFactory binderFactory) throws Exception {
-
-        final LocalDate startDate = LocalDate.parse(webRequest.getParameter("startDate"));
-        final LocalDate endDate = LocalDate.parse(webRequest.getParameter("endDate"));
-
-        if (endDate.isBefore(startDate)) throw new IllegalArgumentException("end date must be after start date");
-
         return AssessmentPercentileRequest.builder()
                 .assessmentId(Ints.tryParse(webRequest.getParameter("assessmentId")))
-                .startDate(startDate)
-                .endDate(endDate)
+                .startDate(LocalDate.parse(webRequest.getParameter("startDate")))
+                .endDate(LocalDate.parse(webRequest.getParameter("endDate")))
                 .build();
     }
 }
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileRequestArgumentResolver.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileRequestArgumentResolver.java
@@ -1,0 +1,36 @@
+package org.opentestsystem.rdw.reporting.assessment;
+
+import com.google.common.primitives.Ints;
+import org.springframework.core.MethodParameter;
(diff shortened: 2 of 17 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
```diff
@@ -0,0 +1,25 @@
...
+    private final AssessmentPercentileService service;
+
+    @Autowired
+    public AssessmentPercentileController(@NotNull final AssessmentPercentileService service) {
+        this.service = service;
+    }
+
+    @GetMapping("/assessment/{id}/percentile")
+    public List<Percentile> getPercentile(@PathVariable final int id) {
+        return service.findAllForAssessment(id);
```
malaffoon: This is returning them all. So the front-end will decide which to display based on dates? It seems like the front-end at least has the school-year to pass in to reduce the payload?

agorina: Oh, I missed that.  So the result should be limited to the school year dates?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/JdbcAssessmentPercentileRepository.java
```diff
@@ -0,0 +1,69 @@
...
+                new MapSqlParameterSource("assessment_id", assessmentId),
+                (resultSet) -> {
+                    return buildPercentiles(resultSet, assessmentId);
+                });
+    }
+
+    private static List<Percentile> buildPercentiles(final ResultSet resultSet, final int assessmentId) throws SQLException {
+        final Map<LocalDate, Percentile.Builder> percentileByStartDate = newHashMap();
+
+        while (resultSet.next()) {
```
malaffoon: if i find myself with a method that is completely wrapped by `while (resultSet.next())` i prefer to use a row mapper and let the framework do the looping for me.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/Percentile.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.time.LocalDate;
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+public class Percentile {
+    private Integer assessmentId;
+    private LocalDate startDate;
+    private LocalDate endDate;
```
malaffoon: not sure of philosophy over here but assessment id will never be null

agorina: Changing.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
wtritch: I might accept a collection of school years just to keep the API open, but in general the UI looks at one year at a time, I believe.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/JdbcAssessmentPercentileRepository.java
```diff
@@ -0,0 +1,69 @@
...
+            final LocalDate startDate = resultSet.getDate("start_date").toLocalDate();
```
wtritch: I'd double-check that we aren't getting messed up between UTC and server date-times here...

wtritch: In fact... don't we almost always use an Instant to represent a datetime elsewhere in java-land?

malaffoon: Yes, but this isn't a datetime, it is just a date. No timezone concerns. Just have to make sure that both ends don't accidentally create a timestamp when dealing with this.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileControllerIT.java
```diff
@@ -0,0 +1,72 @@
...
+public class AssessmentPercentileControllerIT {
+
+    @Autowired
+    private MockMvc mvc;
+
+    @MockBean
+    private AssessmentPercentileService service;
+
+    @Test
+    public void isShouldEstimateReportRowCount() throws Exception {
```
wtritch: copy-paste artifact? =P

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/PercentileScore.java
```diff
@@ -0,0 +1,40 @@
+package org.opentestsystem.rdw.reporting.assessment;
+
+public class PercentileScore {
+    private Integer percentileRank;
```
esotrope: percentileRank -> rank

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/Percentile.java
```diff
@@ -0,0 +1,110 @@
...
+    private Integer count;
+    private Integer mean;
+    private Integer standardDeviation;
+    private List<PercentileScore> percentileScores;
```
esotrope: percentileScores -> scores

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
wtritch: @agorina I don't think it can, because an assessment is bound to a school year, isn't it?

malaffoon: Yes, the percentile data can cross years. For example, they might generate a set of percentiles and want to use them as reference for two full years of results.

Will, an assessment may be used across years. It is the exam date that is important here. If a student takes a 2015 assessment on 3/3/2018, it is very likely that the percentiles will be based on data from either all of 2016-17 (last full year) or the previous semester in 2017-18, e.g. 9/1/2017-12/31/2017. 

For us, the UI is presenting test results from a school year so we "know" all the test results being displayed fall into a single school year. To be more precise, we could inspect the actual test results and get the real range of dates. Ugh.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileControllerIT.java
agorina: yes, thank you.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/PercentileScore.java
agorina: yeah, I copied it from the ingest project...

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileController.java
agorina: Chatted with @esotrope offline and changing to the date range (unless people object).

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileRequestArgumentResolver.java
```diff
@@ -0,0 +1,36 @@
...
+        return parameter.getParameterType().equals(AssessmentPercentileRequest.class);
+    }
+
+    @Override
+    public Object resolveArgument(final MethodParameter parameter, final ModelAndViewContainer mavContainer, final NativeWebRequest webRequest, final WebDataBinderFactory binderFactory) throws Exception {
+
+        final LocalDate startDate = LocalDate.parse(webRequest.getParameter("startDate"));
+        final LocalDate endDate = LocalDate.parse(webRequest.getParameter("endDate"));
+
+        if (endDate.isBefore(startDate)) throw new IllegalArgumentException("end date must be after start date");
```
esotrope: i would put this in the service layer no?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/JdbcAssessmentPercentileRepository.java
agorina: This is the only comment I have not addressed. And it is not because I am trying to be difficult, I just do not know how to use a row mapper to return less rows that in the SQL result?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/assessment/AssessmentPercentileRequestArgumentResolver.java
agorina: I am fine either way. I think I have seen us doing it in the resolver and that is why I did it.
Service layer makes sense too.

agorina:
Merging but will address comments if needed.

PR #978, opened 2018-02-24
append or increment number when re-generating an aggregate report

dsnyder-fairway:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.resolve.ts
```diff
@@ -29,11 +29,26 @@ export class AggregateReportFormSettingsResolve implements Resolve<AggregateRepo
...
+   * @param {string} name the name with an optional "(N)" to increment
+   * @returns {string} the given name suffixed with "(N + 1)" or " (1)" if no "(N)" is provided
+   */
+  static incrementFileNameSuffix(name: string): string {
+    return name.replace(/((\((\d+)\)(\s)?)?$)/, (a) => {
+      if (a == '') {
+        return ` (${Number(a.replace(/[()]/g, '')) + 1})`;
+      }
+      return `(${Number(a.replace(/[()]/g, '')) + 1})`;
+    });
```
esotrope: can you put this in `support.ts` and add some tests?
also not sure why the behavior variation on `if (a == '')` is needed. the whitespace before the (\d+) should not be replaced based on the code.

dsnyder-fairway: I made some modifications to the regex. Originally the regex found items like "My Name (1)", but I made this search for "My Name (1)" or the end of the string if that isn't there -- such as "" at the end of "My Name". That is why I append the whitespace ahead in front of that in those cases. I can remove this, and it would result in names like "My Name(1)".

I'll move and test

esotrope: oh ok. would be interested to know what necessitated the change to the original but maybe your tests will show.

dsnyder-fairway: Just thought most aggregate reports shouldn't come with a numeric suffix on their first generation

malaffoon: +100 for unit tests. That will demonstrate behavior as well as test.

dsnyder-fairway: @esotrope migrated over to `support.ts`, simplified statement after the `if (a == '')` and added UTs

PR #980, opened 2018-02-24
DWR-1494  Add 'may-contain-embargoed-results' flag to aggregate report result

agorina:
This is not done yet, but I want to get feedback before I continue.

In aggregate report we manage organizations in memory, so added `isEmbargoEnabled` to School, District and State. I naively wanted to add it as `boolean`, but it did not work. So I ended up with `Boolean`. Open to another ideas (create another model, expand, wrap and so on).

Background that did not allow me to add this flag as mandatory:
We share organization model objects across all projects (including admin).
We share the same SQL and the same object mapper to load organizations from DB (with the exception of Aggregate report, it has its own mapper).
And we do not have an integration test for the organization repository against warehouse. 
Since warehouse table structure is not the same as reporting I find it wrong to share the same SQL, so we may want to revisit it later.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRowRepository.java
```diff
@@ -103,7 +116,7 @@ public int estimateReportRowCount(@NotNull final ReportQuery query) {
                     (row, index) -> removeFromTemplateIfExists(templateRows, ReportRow.builder()
                             .assessment(findByIdAndSchoolYear(asmtTypeCode, row.getInt("asmt_id"), row.getInt("school_year")))
                             .dimension(Dimension.builder().code(getNullable(row, row.getString("dimension_code"))).type(DimensionType.caseInsensitiveValue(row.getString("dimension"))).build())
-                            .organization(Organizations.map(row))
+                            .organization(findOneByIdAndType(row.getLong("organization_id"), OrganizationType.valueOf(row.getString("organization_type"))))
                             .measures(Measures.builder()
```
agorina: I changed the approach here to be similar how we load Assessments. The report SQL returns an organization type and id and we load the object from the Organization repository.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -69,11 +79,35 @@ private void checkOrganizationPermissions(final User user, final Set<Long> distr
...
+        for (final Organization organization : userOrganizationService.getOrganizations(user)) {
+            if (organization.getOrganizationType() == OrganizationType.School && querySchoolIds.contains(organization.getId())) {
+                if (((School) organization).isEmbargoEnabled()) return true;
+            }
+            if (organization.getOrganizationType() == OrganizationType.District && queryDistrictIds.contains(organization.getId())) {
+                if (((District) organization).isEmbargoEnabled()) return true;
+            }
+        }
+        return false;
+    }
```
agorina: I have to write tests for it, but does it look right?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
```diff
@@ -74,7 +74,7 @@ public void handleAggregateRequest(final Message<AggregateRequestMessage> messag
 
             objectMapper.writeValue(jsonFile.toFile(), results);
             archiveJson(jsonFile.toFile(), payload.getResponseLocation());
-            sendSuccess(payload.getId(), results.size());
+            sendSuccess(payload.getId(), results.size(), service.isMayContainEmbargoedResults(user, payload.getQuery()));
```
agorina: The `result` also has `isEmbargoEnabled` flag for each row. So an alternative would be to loop through it?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -69,11 +79,37 @@ private void checkOrganizationPermissions(final User user, final Set<Long> distr
...
 
+    @Override
+    public boolean isMayContainEmbargoedResults(final User user, final ReportQuery query) {
+        final PermissionScope embargoScope = getScopeOrEmpty(user.getPermissionsById(), EmbargoRead);
+        if (embargoScope.equals(EMPTY)) return false;
+
+        //state trumps all the districts embargo settings
+        if (!organizationService.isStateEmbargoEnabled()) return false;
+
+        if (organizationService.isStateEmbargoEnabled() && query.isIncludeState()) return true;
```
malaffoon: If the user isn't a statewide embargo admin, is this correct? It seems like the report will just not contain any embargoed results because the user's query will filter them out.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
```diff
@@ -69,11 +79,37 @@ private void checkOrganizationPermissions(final User user, final Set<Long> distr
         }
     }
 
+    @Override
+    public boolean isMayContainEmbargoedResults(final User user, final ReportQuery query) {
+        final PermissionScope embargoScope = getScopeOrEmpty(user.getPermissionsById(), EmbargoRead);
+        if (embargoScope.equals(EMPTY)) return false;
+
```
malaffoon: Where did we check that the report is for summatives?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
malaffoon: It may be correct, but i'm having trouble understanding how the user's embargo admin rights are being directly correlated to the query organizations. A user may be an embargo admin for district A but may be querying for results from district B and C. In that case the may-contain-embargo-results should be false.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
```diff
@@ -41,6 +46,15 @@
         return repository.findByQuery(user, ensureDimension(query, Overall));
     }
 
+    @Override
+    public boolean isMayContainEmbargoedResults(final User user, final ReportQuery query) {
+        if (query.getAssessmentTypeCode().equalsIgnoreCase(AssessmentType.SUMMATIVE.code())) return false;
```
malaffoon: Oh, this answers a previous question. 

Do we really want this business logic spread out? Shouldn't it all be in one place?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
```diff
@@ -112,11 +112,12 @@ private void sendFailure(final long requestId, final String infoMessage) {
         processor.aggregateResponse().send(responseMessage);
     }
 
-    private void sendSuccess(final long requestId, final long resultCount) {
+    private void sendSuccess(final long requestId, final long resultCount, final boolean isMayContainEmbargoedResults) {
         final AggregateResponseMessage responsePayload = AggregateResponseMessage.builder()
                 .id(requestId)
                 .status(COMPLETED)
                 .resultCount(resultCount)
+                .isMayContainEmbargoedResults(isMayContainEmbargoedResults)
```
malaffoon: I'm beginning to regret my choice of name for this flag  ;)

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
agorina: Agree, will move it into one place.

aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -51,16 +51,22 @@ sql:
...
-          name AS organization_name,
-          natural_id AS organization_natural_id,
-          'District' as organization_type
-        FROM district
+        SELECT DISTINCT
+          s.district_id AS organization_id,
+          d.name AS organization_name,
+          d.natural_id AS organization_natural_id,
+          'District' as organization_type,
+          s.embargo_enabled
```
malaffoon: Ugh. So when we migrate we join the district embargo setting into the school cause that's what we thought we'd need. Now we're reversing that join to get the district embargo setting? If we need it i think we should change the schema (double ugh) and keep the flag with the district as well. Although then the question becomes: do we keep the district embargo flag as set or the effective embargo flag based on district/state settings?

aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -51,16 +51,22 @@ sql:
...
+        FROM school s
+          JOIN district d on s.district_id = d.id
+
+      isStateEmbargoEnabled: >-
+        SELECT EXISTS(SELECT 1FROM state_embargo WHERE aggregate = 1) as embargo_enabled
```
malaffoon: typo

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultAuthorizationService.java
agorina: _having trouble understanding_ - well, this makes two of us - My head is spinning in this method. But if we want to take this path, I will take a break and re-work it.

aggregate-service/src/main/resources/application.sql.yml
agorina: Interesting, it did not complained in test...thank you.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/School.java
```diff
@@ -10,6 +10,7 @@
     private Long schoolGroupId;
     private long districtId;
     private Long districtGroupId;
+    private Boolean isEmbargoEnabled;
```
malaffoon: Seems like this could go in Organization, if it represents the effective embargo flag. Every organization has an effective embargo flag.

Since this is a common data model, is this the `aggregate` or `individual` embargo flag?

aggregate-service/src/main/resources/application.sql.yml
agorina: I know, I have to admit that this whole approach just does not feel right to me. But I do not have a better solution for now.

malaffoon:
I do remember having the conversation about the common OrganizationRepository: we knew that we'd have to deal with the different data mart versions of it at some point.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/School.java
agorina: I guess we would need two flags. 

Here is what feels wrong about this:  embargo is not just for the organization, it is for Summative and the current school year. But putting this on Organization makes it misleading.

aggregate-service/src/main/resources/application.sql.yml
malaffoon: I know we're under a severe time crunch but i really think we might want to chat about this to see if we can avoid a complete hack job.

agorina: *Completely* agree!

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/School.java
malaffoon: Valid points. I guess we could have two flags: effectiveAggregateEmbargoForCurrentSchoolYearSummativeResults and effectiveIndividualEmbargoForCurrentSchoolYearSummativeResults.

agorina: You know that I am going to use these names :)

agorina:
Just a thought to consider: 
1. remove the flag from the common model and create an aggregate specific wrapper that has an organization and embargo flag.
2. when building a report, for each row check Assessment Type, Organization Embargo Flag and year. If it qualifies as embargoed - add a flag to a report. Yes, the report will be changed to have this flag and a collection of rows. 

This feels cleaner, thoughts?

malaffoon:
I plead can't-focus-well-enough-to-evaluate. Conceptually it sounds better since it keeps the concerns of aggregate reports separate. But i haven't thought it through enough yet.

PR #981, opened 2018-02-25
Bug - failing with exception while checking the estimated row count

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
```diff
@@ -37,30 +37,31 @@
...
         authorizationService.checkAuthorization(user, query);
-        checkIsValid(query);
-        return repository.findByQuery(user, ensureDimension(query, Overall));
+        return repository.findByQuery(user, checkIsValid(query));
     }
 
     @Override
     public int estimateReportRowCount(final User user, final ReportQuery query) {
         authorizationService.checkAuthorization(user, query);
-        checkIsValid(query);
```
agorina: Front end does not sent Overall, and the check for valid was failing.

agorina:
Merging, will address comments if needed.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
malaffoon: Just for my benefit, what is the actual change? It looks like the same calls are being made and the code was just pushed around a bit?

agorina: `ensureDimension` adds Overall dimension to the query if it is missing. Before this change it was done after `checkIsValid`, causing the check to fail.

malaffoon: Ah, thanks.

PR #983, opened 2018-02-26
Modify report downloads to use browser to download rather than ajax.

wtritch:
This PR just converts our report download actions to open a browser tab rather than download via AJAX request.
The benefit is that for large report downloads the user can watch/cancel progress using standard browser download management.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
@@ -13,7 +13,6 @@ import { Utils } from "../../shared/support/support";
 import { Comparator, ranking } from "@kourge/ordering/comparator";
 import { ordering } from "@kourge/ordering";
 import { AggregateReportQuery } from "../../report/aggregate-report-request";
-import { saveAs } from "file-saver";
 import { DisplayOptionService } from "../../shared/display-options/display-option.service";
 import { TranslateService } from "@ngx-translate/core";
 import { AggregateReportRequestMapper } from "../aggregate-report-request.mapper";
@@ -125,14 +124,8 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   }
 
   onDownloadDataButtonClick(): void {
-    this.spinnerModal.loading = true;
-    this.reportService.getReportContent(this.report.id)
-      .finally(() => {
-        this.spinnerModal.loading = false;
-      })
-      .subscribe(download => {
-        saveAs(download.content, download.name);
-      });
+    this.reportService.downloadReportContent(this.report.id);
+    this.router.navigateByUrl("/reports");
   }
 
   getExportName(table: AggregateReportTableView): string {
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -125,14 +124,8 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
...
-    this.spinnerModal.loading = true;
-    this.reportService.getReportContent(this.report.id)
-      .finally(() => {
-        this.spinnerModal.loading = false;
-      })
-      .subscribe(download => {
-        saveAs(download.content, download.name);
-      });
+    this.reportService.downloadReportContent(this.report.id);
+    this.router.navigateByUrl("/reports");
```
wtritch: After triggering a download from the "Your report is very large, view or download?" dialog of the aggregate report result page, navigate the user to the "My Reports" page.

webapp/src/main/webapp/src/app/report/report-action.component.ts
```diff
@@ -39,13 +38,7 @@ export class ReportActionComponent implements OnInit {
...
 
   performAction(reportAction: ReportAction): void {
-    this.spinnerModal.loading = true;
-    const action: Observable<any> = this.actionService.performAction(reportAction);
-    action
-      .finally(() => {
-        this.spinnerModal.loading = false;
-      })
-      .subscribe(() => {});
+    this.actionService.performAction(reportAction);
```
wtritch: Report actions are now instantaneous.  The user has either triggered a download or navigated.  Either way, there is no longer a reason to bind a spinner to performing an action on a report.

webapp/src/main/webapp/src/app/report/report-action.service.ts
```diff
@@ -44,32 +41,17 @@ export class ReportActionService {
...
-      case ActionType.Navigate:
-        this.performNavigate(action);
-        return Observable.empty();
+  public performAction(action: ReportAction): void {
+    if (action.type == ActionType.Download) {
+      this.performDownload(action);
+    }
+    if (action.type == ActionType.Navigate) {
+      this.performNavigate(action);
     }
```
wtritch: Now that we're no longer returning, an if-chain felt cleaner than a CASE

dsnyder-fairway:
you might be able to remove the `<spinner-modal>` from the html files too

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -120,6 +121,15 @@ export class ReportService {
...
   }
 
+  /**
+   * Download the given report by opening a tab to the report content.
+   *
+   * @param {number} reportId The report id
+   */
+  public downloadReportContent(reportId: number): void {
+    window.open(`${this.contextUrl}${ServiceRoute}/reports/${reportId}`, '_blank');
+  }
```
wtritch: Since this (currently) only happens as a direct response to a user action we shouldn't have any issues with popup blockers.

webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -120,6 +121,15 @@ export class ReportService {
     }).catch(ResponseUtils.throwError);
   }
 
+  /**
+   * Download the given report by opening a tab to the report content.
+   *
+   * @param {number} reportId The report id
+   */
+  public downloadReportContent(reportId: number): void {
+    window.open(`${this.contextUrl}${ServiceRoute}/reports/${reportId}`, '_blank');
```
esotrope: neat : j

wtritch:
@dsnyder-fairway good call, removed it from the report action component.

PR #985, opened 2018-02-26
DWR-1496 Moved things around in preparation for a different embargo S…

agorina:
…QL for export/pdf vs view exams

This just moves files around so that I we could have a different SQL and different test for the embargo related permissions.

There are some parts of this review that I am not really happy about. Open to a better alternatives if people have it.


common/src/test/resources/integration-test-data-permissions.sql
```diff
@@ -0,0 +1,171 @@
+insert into school_group (id, natural_id, name) values
```
agorina: Since this is the file currently used for the permissions tests, I moved it into common.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -238,11 +238,22 @@ sql:
...
     snippet:
+      studentExamPermission: >-
+            ${sql.snippet.studentExamPermissionNoEmbargo}
+            AND
+            (
+               e.school_year < ${app.school-year}
+               OR e.type_id <> 3
+               OR s.embargo_enabled = 0
+               OR 1 = :state_embargo_admin
+               OR s.district_id IN (:district_embargo_admin_ids)
```
agorina: This will be changed with the next PR.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamEmbargoPermissionCurrentSchoolYearIT.java
```diff
@@ -0,0 +1,178 @@
...
+                        "  scale_score, scale_score_std_err, performance_level, completed_at," +
+                        "  claim1_category, claim1_scale_score, claim1_scale_score_std_err," +
+                        "  claim2_category, claim2_scale_score, claim2_scale_score_std_err," +
+                        "  claim3_category, claim3_scale_score, claim3_scale_score_std_err," +
+                        "  claim4_category, claim4_scale_score, claim4_scale_score_std_err," +
+                        "  update_import_id, updated, migrate_id) values" +
+                        " (-99, 1, -1, 'g1', -1, -30, 0, 0, 0, 0, 0, 1997, -1, 'v1', 2, 'Complete', 1, 'Valid', 'session1', 2000, 20, 1, '1997-01-01 00:00:00.000000', 1, 100, 10, 2, 200, 20, 3, 300, 30, 4, 400, 40, -1, '1997-07-18 20:14:34.000000', -1)"
+        })
+})
+public class JdbcExamEmbargoPermissionCurrentSchoolYearIT {
```
agorina: This will test for the report processor embargo rules.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcStudentExamPermissionRepository.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import javax.validation.constraints.NotNull;
+import java.util.List;
+
+@Repository
+class JdbcStudentExamPermissionRepository implements StudentExamPermissionRepository {
```
agorina: This is what I do not like:

When I created permission related test in the reporting, I re-used the existing repository since it had all what was needed to test all the permissions angles. This repository is proprietary to the reporting service. Now I am creating a version of it with the report processor embargo SQL clause to test embargo permission in the report processor.

Any better ideas?

common/src/main/resources/common.sql.yml
```diff
@@ -128,7 +128,7 @@ sql:
     basicPermission: >-
         (1=:statewide or sc.district_group_id in (:district_group_ids) or sc.district_id in (:district_ids) or sc.school_group_id in (:school_group_ids) or sc.id in (:school_ids))
 
-    studentExamPermission: >-
+    studentExamPermissionNoEmbargo: >-
```
esotrope: ideally we name the one with embargo *WithEmbargo right? would be slightly more future-proof

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcStudentExamPermissionRepository.java
esotrope: only idea is that we make a "fake repo" with dedicated (not used in the real app) queries that we test.
I think the only issue with this is that it is tied to a "real life" repo that can potentially change and not lend itself to this testing

common/src/test/resources/integration-test-data-permissions.sql
esotrope: should we prefix this file with what schema it is for?

agorina: will do

common/src/main/resources/common.sql.yml
agorina: Good point, will rename.

PR #986, opened 2018-02-26
do not allow printing summative reports

dsnyder-fairway:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1615,6 +1615,7 @@
     "no-writing-trait-scores": "Writing trait scores are not available for assessments administered prior to 2017-18 school year.",
     "no-writing-trait-scores-for-summative-ica-exams": "Writing trait scores are currently available for only ELA Performance Task IABs.",
     "no-responses-for-summative-exams": "Summative assessment responses are not made available by this system.",
+    "no-print-report-for-summative-exams": "Printing summative reports are not made available by this system.",
```
wtritch: Maybe something like: "The system does not currently allow printing summative reports."

malaffoon:
Unless i'm missing something, i don't think we want to do this. Yes, the system won't properly print summative student reports but, at worst, it will be blank or look like an ICA. This is just a known short-coming / bug that will be addressed in Phase 3. No production system will have summatives this year, so it shouldn't be a problem. And we don't want to have to undo this change for testing in QA, UAT or demo.

@esotrope, @agorina please weigh in if you know/feel differently.

webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
```diff
@@ -129,7 +129,9 @@ export class ResultsByStudentComponent implements OnInit {
           });
 
           downloader.modal.show();
-        }
+        },
+        () => this.assessment.isSummative,
+        () => this.assessment.isSummative ? this.translate.instant('messages.no-print-report-for-summative-exams') : ''
```
wtritch: Maybe add a TODO in here to re-enable once summative PDF reports are implemented.

agorina:
I agree with @malaffoon's comment. There are other features that are not summative-tested, but we do not disable them. As long as nothing catastrophical happens, I do not see a reason to disable it.

dsnyder-fairway:
declining. We aren't expecting summative tests for phase 2

PR #987, opened 2018-02-26
require filename for student / group reports

dsnyder-fairway:


webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -25,12 +25,17 @@ <h4 class="modal-title pull-left">{{title}}</h4>
...
               <label for="report-name">{{'labels.reports.form.report-name' | translate}}</label>
               <input id="report-name"
+                     #resourceControl="ngModel"
                      name="report-name"
                      class="form-control input-sm"
                      type="text"
                      maxlength="255"
                      [(ngModel)]="options.name" required>
             </div>
+            <div *ngIf="resourceControl.invalid && (resourceControl.dirty || resourceControl.touched)">
```
esotrope: i made a util method under `shared/forms/Forms` for this if you want to use it - totally optional though because its awkward getting util methods into templates

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -25,12 +25,17 @@ <h4 class="modal-title pull-left">{{title}}</h4>
             <div class="form-group">
```
esotrope: this also needs class `.has-error` when this is true:
```
resourceControl.invalid && (resourceControl.dirty || resourceControl.touched)
```

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -222,6 +221,9 @@
     },
     "copy-suffix": "(copy)",
```
esotrope: `rm` 🙇

dsnyder-fairway: good eye. Removing

webapp/src/main/webapp/src/app/report/report-download.component.html
dsnyder-fairway: 👍

webapp/src/main/webapp/src/app/report/report-download.component.html
dsnyder-fairway: hmm. I'll skip that

PR #988, opened 2018-02-26
avoid page break in the middle of pdf graph

dsnyder-fairway:
screen shot 2018-02-26 at 2 47 14 pm


malaffoon:
Hmph, that was easy. Are there other things that get broken that shouldn't be?

@wtritch any feedback on this one. I know you're in the same general area and had a comment about ICA's being more than one page.

dsnyder-fairway:
No worries @wtritch, you can take it

PR #991, opened 2018-02-27
Bugfix/handle empty aggregate report

wtritch:
This handles both the case where the aggregate report contains no rows and when the report contains only rows with no results.

Results page when report contains no results:
image

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
@@ -30,6 +30,7 @@ import java.util.Properties;
 import static org.opentestsystem.rdw.olap.stream.AggregateRequestProcessor.AggregateRequest;
 import static org.opentestsystem.rdw.reporting.common.model.ReportStatus.COMPLETED;
 import static org.opentestsystem.rdw.reporting.common.model.ReportStatus.FAILED;
+import static org.opentestsystem.rdw.reporting.common.model.ReportStatus.NO_RESULTS;
 import static org.springframework.http.MediaType.APPLICATION_JSON;
 
 /**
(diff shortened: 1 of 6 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
```diff
@@ -69,6 +70,14 @@ public void handleAggregateRequest(final Message<AggregateRequestMessage> messag
         try {
             final User user = messageSecurityService.getUser(message);
             final AggregateReport report = service.findByQuery(user, payload.getQuery());
+
+            final boolean hasResults = report.getRows().stream()
+                    .anyMatch(reportRow -> reportRow.getMeasures().getAvgScaleScore() != null);
+            if (!hasResults) {
+                sendNoResults(payload.getId());
+                return;
+            }
```
wtritch: Best-case is we check the first report row and it has an avg scale score (has results.)
Worst-case is we check x thousand report rows and *none* have results.  (Should be extremely unlikely.)

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.html
```diff
@@ -46,6 +46,12 @@ <h2>{{report.label}}</h2>
       <spinner text="{{'aggregate-report.processing' | translate}}"></spinner>
     </ng-container>
 
+    <ng-container *ngIf="reportEmpty">
+      <div class="well">
+        <p>{{'aggregate-report.empty' | translate}}</p>
+      </div>
+    </ng-container>
```
wtritch: Added a new state to the FSM, @esotrope =)

webapp/src/main/webapp/src/app/report/report.model.ts
```diff
@@ -24,12 +24,12 @@ export class Report {
...
   }
 
-  get processing(): boolean {
-    return this.status === 'PENDING' || this.status === 'RUNNING';
+  get empty(): boolean {
+    return this.status === 'NO_RESULTS';
   }
 
-  get loadable(): boolean {
-    return this.status !== 'FAILED' && this.status !== 'EXPIRED';
```
wtritch: I removed the "loadable" in favor of "completed" since they're synonymous.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -16,6 +16,7 @@
   "aggregate-report": {
     "column-order": "Column Order",
     "column-order-info": "Change your result column order to easily compare your results broken down in different ways.",
+    "empty": "Your report does not contain any results. Please modify your query and generate a new report.",
```
wtritch: Double-check my word-smithing, please.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
malaffoon: woot, thanks

PR #993, opened 2018-02-27
swapped order subgroup and advanced filters on "Report Review"

dsnyder-fairway:
appears consistent with page layout now
screen shot 2018-02-26 at 6 28 34 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
@@ -6,11 +6,11 @@ import { AggregateReportFormSettings } from "./aggregate-report-form-settings";
 import { AssessmentDefinition } from "./assessment/assessment-definition";
 import { Utils } from "../shared/support/support";
 
-const NarrowColumnProvider: ColumnProvider = (organization, assessment, filter, subgroup) =>
-  [[organization, assessment], [filter, subgroup]];
+const NarrowColumnProvider: ColumnProvider = (organization, assessment, subgroup, filter) =>
+  [[organization, assessment], [subgroup, filter]];
 
-const WideColumnProvider: ColumnProvider = (organization, assessment, filter, subgroup) =>
-  [[organization], [assessment], [filter], [subgroup]];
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -6,10 +6,10 @@ import { AggregateReportFormSettings } from "./aggregate-report-form-settings";
 import { AssessmentDefinition } from "./assessment/assessment-definition";
 import { Utils } from "../shared/support/support";
 
-const NarrowColumnProvider: ColumnProvider = (organization, assessment, subgroup, filter) =>
+const NarrowColumnProvider: ColumnProvider = (organization, assessment, filter, subgroup) =>
   [[organization, assessment], [filter, subgroup]];
 
-const WideColumnProvider: ColumnProvider = (organization, assessment, subgroup, filter) =>
+const WideColumnProvider: ColumnProvider = (organization, assessment, filter, subgroup) =>
   [[organization], [assessment], [filter], [subgroup]];
```
esotrope: missing a commit? i only see param order changes.
Dont you need to change the order the args are fed or the order these arrays are built?
would expect to keep params the same but change the array creation here and above

dsnyder-fairway: I thought this was a bit strange too. What we have here in this file below was 
```
 this.columns = this._columnProvider(
      {
        label: translate('aggregate-reports.form.section.organization.heading'),
        rows: organizationRows
      },
      {
        label: translate('aggregate-reports.form.section.assessment.heading'),
        rows: assessmentRows
      },
      {
        label: translate('aggregate-reports.form.section.comparative-subgroups.heading'),
        rows: subgroupRows
      },
      {
        label: translate('aggregate-reports.form.section.subgroup-filters.heading'),
        rows: filterRows
      }
    )
```
Before I swapped the last two rows in there and that gave me what I wanted, but the order there looks okay. I searched around and found some other areas whose names appeared out of order and simply changing those resulted in the correct behavior

PR #995, opened 2018-02-27
An idea on how to disable a front end export for embargoed exams.

agorina:
Just a thought - if the front end people like it, I will complete the tests.

With this, the UI  could disable the export of any records that have `createdWhileDataEmbargoed` set to true.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Exam.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Exam.java
@@ -27,6 +27,7 @@ public class Exam {
     private boolean transfer;
     private WritingTraitScores writingTraitScores;
     private String performanceWritingType;
+    private Boolean createdWhileDataEmbargoed;
 
     protected Exam() {
     }
@@ -140,6 +141,13 @@ public class Exam {
         return transfer;
     }
 
+    /**
+     * @return true if the result may contain embargo enabled data, false otherwise
+     */
+    public Boolean isCreatedWhileDataEmbargoed() {
+        return createdWhileDataEmbargoed;
+    }
+
     public static Builder builder() {
         return new Builder();
     }
@@ -162,6 +170,7 @@ public class Exam {
         private boolean transfer;
         private WritingTraitScores writingTraitScores;
         private String performanceWritingType;
+        private Boolean createdWhileDataEmbargoed;
 
         protected E createElement() {
             return (E) new Exam();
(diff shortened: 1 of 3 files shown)


common/src/main/resources/common.sql.yml
```diff
@@ -217,7 +217,8 @@ sql:
         s.name as school_name,
         s.district_id as school_district_id,
         s.district_group_id as school_district_group_id,
-        s.school_group_id
+        s.school_group_id,
+        CASE WHEN e.school_year = ${app.school-year} AND s.embargo_enabled THEN true ELSE false END AS embargo_enabled
```
agorina: Do we need to care about the inferred school embargo flag here too?

malaffoon: Regarding inferred school. That is one heck of a corner case. It means the student transferred from another district in the same state, and took a summative test there while the embargo setting was different than the current district. My vote: file a JIRA describing the possible hole and move on.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Exam.java
```diff
@@ -27,6 +27,7 @@
     private boolean transfer;
     private WritingTraitScores writingTraitScores;
     private String performanceWritingType;
+    private Boolean createdWhileDataEmbargoed;
```
malaffoon: variable name doesn't quite make sense (the exam itself wasn't created while the data was embargoed) but i get that it is consistent with AggregateReport

malaffoon: Should the flag be on Exams instead of Exam? Or are there places where Exam is returned not in the Exams collection?

agorina: Exams is just a static helper, the return result is a collection of Exam.

And yes, this flag == true means that this Exam cannot be exported on the front end.

malaffoon: Oh duh (about Exams). 

Do the exams come back in a wrapper collection? Because it seems like that is what should have the single flag that controls whether the Export CSV is enabled. But i guess the front-end can just scan the list of exams and note if any have the flag set.

wtritch:
Hm... I'm wondering if this is necessary.  This will affect every exam.  I wonder if it would be cheaper to do this on the front-end at CSV export time... The front-end would need a school's boolean "embargoed" property set and would need to know the "current school year" from the configuration context...

agorina:
@wtritch , I am not insisting at all. If we have a cleaner way of doing it, I am all for it.

PR #996, opened 2018-02-27
Add a debounce to all our typeahead controls

wtritch:
While working with typeaheads I noticed we were sending a request for every character typed, even if the user quickly typed the word "community" we made 7 requests (minimum request length of 3 characters.)  A UX issue comes into play if the search for "com" returns after the search for "community" leading to odd results in the resulting options list.

This PR just adds a 300ms debounce to our typeaheads so that if a user types a name relatively quickly the request will be delayed until 300ms after the last character press.


webapp/src/main/webapp/src/app/shared/organization/organization-typeahead.ts
```diff
@@ -16,6 +16,7 @@ import { TypeaheadMatch } from "ngx-bootstrap";
            (typeaheadNoResults)="noResultsInternal = $event"
            (typeaheadOnSelect)="onTypeaheadSelectInternal($event)"
            [typeaheadMinLength]="3"
+           typeaheadWaitMs="300"
```
esotrope: wow surprised there isn't a builtin default debounce : /

PR #999, opened 2018-02-27
corrected error text

dsnyder-fairway:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -36,7 +36,7 @@
   "aggregate-report": {
     "column-order": "Column Order",
     "column-order-info": "Change your result column order to easily compare your results broken down in different ways.",
-    "empty": "Your report does not contain any results. Please modify your query and generate a new report.",
+    "empty": "No available test results matching your report query. Please modify your query and generate a new report.",
```
malaffoon: It should be "There are no available ..."

Unless you were trying to shorten it to fit?

dsnyder-fairway: I'll update it

PR #1001, opened 2018-02-27
Disable embargoed aggregate report download in the UI.

wtritch:
This PR disables the download actions for an aggregate report when it was generated containing embargoed results.

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
@@ -144,6 +144,10 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
     tableView.columnOrdering = items.map(item => item.value);
   }
 
+  isEmbargoed(): boolean {
+    return this.report.metadata.createdWhileDataEmbargoed == "true";
+  }
+
   private updateViewState(): void {
     let targetViewState: ViewState = this.getTargetViewState();
     this.onViewStateChange(this._viewState, targetViewState);
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
@@ -145,7 +145,7 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   }
 
   isEmbargoed(): boolean {
-    return this.report.metadata.createdWhileDataEmbargoed == "true";
+    return this.report.metadata.createdWhileDataEmbargoed === "true";
   }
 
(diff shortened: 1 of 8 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -911,3 +911,7 @@ body .ui-widget.ui-datatable table .ui-datatable-data > tr > td.level-down,
     margin-bottom: 0;
   }
 }
+
+a.btn.disabled.with-pointer {
+  pointer-events: auto;
+}
```
wtritch: a.btn.disabled was by-default turning off pointer-events, which prevented the popover tooltip from appearing.

malaffoon:
Channeling SB here ... perhaps "Reports with embargoed results may only be viewed online, they may not be downloaded"

wtritch:
@malaffoon sounds good to me, changed

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -144,6 +144,10 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
     tableView.columnOrdering = items.map(item => item.value);
   }
 
+  isEmbargoed(): boolean {
+    return this.report.metadata.createdWhileDataEmbargoed == "true";
```
esotrope: nit: `===` should be used when it can be used

webapp/src/main/webapp/src/app/report/report-action.service.ts
```diff
@@ -138,7 +144,8 @@ class AggregateReportActionProvider extends DefaultActionProvider {
         type: ActionType.Download,
         value: report.id,
         labelKey: 'labels.reports.report-actions.download-report',
-        disabled: disableViewAndDownload
+        disabled: disableViewAndDownload || embargoed,
+        tooltipKey: embargoed ? 'labels.reports.report-actions.embargoed' : undefined
```
esotrope: technically it is a popover in the lingo of our app so i'd call it that

webapp/src/main/webapp/src/styles.less
esotrope: ew x J

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
wtritch: changed

webapp/src/main/webapp/src/app/report/report-action.service.ts
wtritch: sure, I don't want to get into the property name vs display mechanic debate =P

PR #1002, opened 2018-02-27
adds embargo alert frontend

esotrope:
if the criteria is met - the warning is displayed everywhere in the app below the nav bar / breadcrumbs 

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/app.component.html
+++ b/webapp/src/main/webapp/src/app/app.component.html
@@ -12,6 +12,8 @@
   </li>
 </sb-header>
 
+<embargo-alert *hasPermission="'EMBARGO_WRITE'"></embargo-alert>
+
 <div class="container">
   <router-outlet></router-outlet>
 </div>
--- a/webapp/src/main/webapp/src/app/app.component.html
+++ b/webapp/src/main/webapp/src/app/app.component.html
@@ -12,8 +12,6 @@
   </li>
 </sb-header>
 
-<embargo-alert *hasPermission="'EMBARGO_WRITE'"></embargo-alert>
-
 <div class="container">
   <router-outlet></router-outlet>
 </div>
--- a/webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.service.ts
+++ b/webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.service.ts
@@ -1,0 +1,25 @@
+import { Injectable } from "@angular/core";
+import { CachingDataService } from "../data/caching-data.service";
+import { EmbargoAlert } from "./embargo-alert";
+import { Observable } from "rxjs/Observable";
+import { of } from "rxjs/observable/of";
+import { catchError } from "rxjs/operators";
+
+@Injectable()
+export class EmbargoAlertService {
+
+  constructor(private dataService: CachingDataService) {
+  }
+
+  /**
+   * Gets the alert status used for indicating whether or not to display the embargo admin alert or not
+   *
+   * @returns {Observable<EmbargoAlert>}
+   */
+  getAlert(): Observable<EmbargoAlert> {
+    return this.dataService.get('/embargo-alert')
+      // fill for backend - if the backend is not there return true by default
+      .pipe(catchError(response => of({enabled: true})));
+  }
+
+}
--- a/webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.ts
+++ b/webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.ts
@@ -1,0 +1,8 @@
+export class EmbargoAlert {
+
+  /**
+   * Indicates whether or not the embargo admin alert should be displayed or not
+   */
+  enabled: boolean;
+
+}
--- a/webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.ts
+++ b/webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.ts
@@ -1,8 +1,0 @@
-export class EmbargoAlert {
-
-  /**
-   * Indicates whether or not the embargo admin alert should be displayed or not
-   */
-  enabled: boolean;
-
-}
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -39,7 +39,7 @@
   "aggregate-report": {
     "column-order": "Column Order",
     "column-order-info": "Change your result column order to easily compare your results broken down in different ways.",
-    "empty": "Your report does not contain any results. Please modify your query and generate a new report.",
+    "empty": "There are no available test results matching your report query. Please modify your query and generate a new report.",
     "heading": "Aggregate Report",
     "form-link": "View Query",
     "processing": "Your report is being generated. This may take several minutes. ",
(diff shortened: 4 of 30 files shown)


webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.service.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Gets the alert status used for indicating whether or not to display the embargo admin alert or not
+   *
+   * @returns {Observable<EmbargoAlert>}
+   */
+  getAlert(): Observable<EmbargoAlert> {
+    return this.dataService.get('/embargo-alert')
+      // fill for backend - if the backend is not there return true by default
+      .pipe(catchError(response => of({enabled: true})));
```
esotrope: added temporary fill for the backend

webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(private dataService: CachingDataService) {
+  }
+
+  /**
+   * Gets the alert status used for indicating whether or not to display the embargo admin alert or not
+   *
+   * @returns {Observable<EmbargoAlert>}
+   */
+  getAlert(): Observable<EmbargoAlert> {
+    return this.dataService.get('/embargo-alert')
```
esotrope: was not sure what resource name to give this

webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -12,6 +12,8 @@
   </li>
 </sb-header>
 
+<embargo-alert *hasPermission="'EMBARGO_WRITE'"></embargo-alert>
```
agorina: Should this be 'EMBARGO_READ'?

agorina:
Should we at least differentiate between `individual` or `aggregate` embargos?

webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.ts
```diff
@@ -0,0 +1,8 @@
+export class EmbargoAlert {
+
+  /**
+   * Indicates whether or not the embargo admin alert should be displayed or not
+   */
+  enabled: boolean;
```
malaffoon: Shouldn't there be a separate aggregate and individual flag? Or did i miss that somewhere?

webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable } from "@angular/core";
+import { CachingDataService } from "../data/caching-data.service";
+import { EmbargoAlert } from "./embargo-alert";
+import { Observable } from "rxjs/Observable";
+import { of } from "rxjs/observable/of";
+import { catchError } from "rxjs/operators";
+
+@Injectable()
+export class EmbargoAlertService {
```
wtritch: Might want to rename this to just EmbargoService and have an `isEmbargoed(organization: Organization): boolean` method that just returns true right now.  We could use that to disable the export buttons.

wtritch: ...wonder if it needs the organization parameter...

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -5,6 +5,9 @@
   "home": {
     "aggregate-reports-link": "Create Aggregate Report"
   },
+  "embargo-admin-alert": {
+    "message": "Summative test results in your region of administration are not yet released. These results may be viewed online but not downloaded."
```
wtritch: Maybe rename "These results may..." to "Results may..." to indicate that we're disallowing more than just summative results atm?

webapp/src/main/webapp/src/app/app.component.html
esotrope: this will work but good point : j most other places we query "write" for admins is why i did it

webapp/src/main/webapp/src/app/shared/embargo/embargo-alert.ts
wtritch: Did we decide that the reporting app only cared about the individual flag?  and the aggregate-service only cared about the aggregate flag?

esotrope: good point. i'll have to redesign this a bit to lookup the flag in the respective dbs

malaffoon: Pretty sure we decided that the banners in the aggregate and individual parts of the app are triggered separately.

webapp/src/main/webapp/src/app/app.component.html
malaffoon: Really? i thought this should be EMBARGO_WRITE because that's the main permission that corresponds to Embargo Admin.

agorina: What is the purpose of EMBARGO_READ then?

malaffoon: Ugh. Do we have two perms where we should have one?

My interpretation is that EMBARGO_WRITE is necessary to modify the embargo setting. So that is what indicates that a user is an "Embargo Admin". I don't think we use it but EMBARGO_READ would merely imply the user can see the embargo settings. Something that we don't really care about. Is my interpretation consistent with other READ/WRITE perms?

If my interpretation is correct, then any permission check that allows a user to ignore the embargo settings (i.e. see embargoed summatives for the current school year) should be referring to EMBARGO_WRITE. 

An alternative interpretation is that EMBARGO_WRITE is necessary to modify the embargo setting and EMBARGO_READ is a separate permission that allows the user to ignore the embargo settings when querying test results. Perhaps this is the better interpretation since it allows more flexibility.

agorina: I took EMBARGO_READ from this page: https://confluence.fairwaytech.com/display/SBACDW/Roles+and+Permissions

So do we keep it? What is the conclusion?

malaffoon: Sorry i wasn't explicit, i was waiting for @esotrope and/or @wtritch  to weigh in. 

I favor the last interpretation since it is what we've done so far and provides the most flexibility. I will updated that confluence page.

esotrope: i think technically the right course would be to add a permission to the role called EMBARGOED_PII_READ or something but for now i think the right correct perm is _EMBARGO_READ_ since this allows the introduction of a role that can just read the results but not edit the flag

esotrope:
@malaffoon @agorina @wtritch made updates to support multiple backing schema - pls re-review.
@wtritch - i did not yet add an isEmbargoed(org) method signature

malaffoon:
With the latest changes, will there be a warning on the home page? Not saying there should or shouldn't be, just curious.

agorina:
Just to make sure I understand it right: the idea is that later it will we backed up by the Java calls to find out the flag for individual/aggregate respectively?

esotrope:
@malaffoon - nope (we can put one there if we want though and it would probably look at the reporting db)
@agorina - correct

PR #1003, opened 2018-02-27
Enable aggregate state results for summatives only

wtritch:
This PR fixes the "Include State" toggle in the aggregate report query builder.  Previously the value display was disconnected from the value being sent to the back-end.

See a [movie of the behavior.](https://jira.fairwaytech.com/secure/attachment/25072/agg_report_include_state.mov)

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -112,8 +112,7 @@
                       <label class="inline-label">
                         <input type="checkbox"
                                [disabled]="summativeFieldsDisabled"
-                               [checked]="!summativeFieldsDisabled"
-                               [(ngModel)]="settings.includeStateResults"
+                               [(ngModel)]="includeStateResults"
                                [ngModelOptions]="{standalone: true}"
                                (ngModelChange)="onIncludeStateResultsChange(); onSettingsChange()">
                         <span class="ml-xs">{{'aggregate-reports.form.section.organization.include.state-results' | translate}}</span>
@@ -177,7 +176,7 @@
                            [options]="options.assessmentTypes"
                            [(ngModel)]="settings.assessmentType"
                            [ngModelOptions]="{standalone: true}"
-                           (change)="onSettingsChange()"></sb-radio-button-group>
+                           (change)="onIncludeStateResultsChange(); onSettingsChange()"></sb-radio-button-group>
               </div>
             </div>
           </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
@@ -266,10 +266,6 @@ export class AggregateReportFormComponent {
   }
 
   set includeStateResults(value: boolean) {
-    if (this.summativeFieldsDisabled) {
-      return;
-    }
-
     this.settings.includeStateResults = value;
   }
 
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
@@ -49,19 +49,6 @@ const fileName = (properties: any) => control => {
   return /^[^\\<>:;,?"*|/]*$/.test((control.value || '').trim()) ? null : { fileName: properties };
 };
 
-/**
- * Form control validator that makes sure the organization control is not empty
- * or includeStateResults or inlcudeAllDistricts is flagged
- *
- * @param properties the properties to propagate when the control value is invalid
- * @return {null|{notEmpty: any}}}
- */
-const organizationValidator = (properties, settings) => control => {
-  return settings.includeStateResults
-  || settings.includeAllDistricts
-  || control.value.length ? null : { invalid: properties };
-};
-
 const OrganizationComparator = (a: Organization, b: Organization) => a.name.localeCompare(b.name);
 
 /**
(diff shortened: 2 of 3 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -112,8 +112,7 @@ <h3>{{'aggregate-reports.form.section.organization.heading' | translate}}</h3>
                       <label class="inline-label">
                         <input type="checkbox"
                                [disabled]="summativeFieldsDisabled"
-                               [checked]="!summativeFieldsDisabled"
-                               [(ngModel)]="settings.includeStateResults"
+                               [(ngModel)]="includeStateResults"
```
wtritch: Binding to both [checked] and [(ngModel)] confused angular, and allowed the two to become disconnected.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -177,7 +176,7 @@ <h3>{{'aggregate-reports.form.section.assessment.heading' | translate}}</h3>
                            [options]="options.assessmentTypes"
                            [(ngModel)]="settings.assessmentType"
                            [ngModelOptions]="{standalone: true}"
-                           (change)="onSettingsChange()"></sb-radio-button-group>
+                           (change)="onIncludeStateResultsChange(); onSettingsChange()"></sb-radio-button-group>
```
wtritch: Now that the state toggle is bound to this control we need to handle state results changes in the assessment type selector as well.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -49,19 +49,6 @@ const fileName = (properties: any) => control => {
...
- * or includeStateResults or inlcudeAllDistricts is flagged
- *
- * @param properties the properties to propagate when the control value is invalid
- * @return {null|{notEmpty: any}}}
- */
-const organizationValidator = (properties, settings) => control => {
-  return settings.includeStateResults
-  || settings.includeAllDistricts
-  || control.value.length ? null : { invalid: properties };
-};
```
wtritch: Since this needs access to the component state, I embedded it below rather than pass an additional "accessor" function into this stateless validator.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -271,6 +261,18 @@ export class AggregateReportFormComponent {
...
+    return this.settings.includeStateResults && !this.summativeFieldsDisabled;
+  }
+
+  set includeStateResults(value: boolean) {
+    if (this.summativeFieldsDisabled) {
+      return;
+    }
+
+    this.settings.includeStateResults = value;
+  }
```
wtritch: I moved the "anding" of the settings value with the business logic into a getter and setter virtual property.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -403,9 +405,6 @@ export class AggregateReportFormComponent {
       settings: this.settings
     };
 
-    // set include state results to false when summative fields are disabled
-    this.summary.settings.includeStateResults = this.summary.settings.includeStateResults && !this.summativeFieldsDisabled;
```
wtritch: This was setting the value false on initialization.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -47,7 +47,7 @@ export class AggregateReportRequestMapper {
       includeAllDistricts: settings.includeAllDistricts,
       includeAllDistrictsOfSchools: settings.includeAllDistrictsOfSelectedSchools,
       includeAllSchoolsOfDistricts: settings.includeAllSchoolsOfSelectedDistricts,
-      includeState: settings.includeStateResults,
+      includeState: settings.includeStateResults && settings.assessmentType == "sum",
```
wtritch: I also need to include the business logic for only including state results if the assessment type is summative here...  An objections/alternatives?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -271,6 +261,18 @@ export class AggregateReportFormComponent {
     return this.estimatedRowCount > SupportedRowCount;
   }
 
+  get includeStateResults(): boolean {
+    return this.settings.includeStateResults && !this.summativeFieldsDisabled;
+  }
+
+  set includeStateResults(value: boolean) {
+    if (this.summativeFieldsDisabled) {
+      return;
```
esotrope: can you explain how this comes into play?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
esotrope: nope - thought i had it there already actually

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
wtritch: The user *shouldn't* be able to toggle the control if it's disabled.  Since, currently, the [disabled] attribute is bound to the summativeFieldsDisabled property.  Buuuut... it's a cheap protection for the future.  I can remove it if you prefer?

esotrope: yeah i'd rather it not be there - it's presence implies we are relying on it

wtritch: done

PR #1004, opened 2018-02-27
Do not store a resource URI for aggregate reports with no content.

wtritch:
This PR prevents storing a report URI for an aggregate report with no results.  Storing a non-existent URI may cause issues during report expiration cleanup processing where the attempt to delete the resource from the archive will fail.

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReportResponseHandler.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReportResponseHandler.java
@@ -59,19 +59,17 @@ public class AggregateReportResponseHandler {
             return;
         }
 
-        if (report.getStatus() != COMPLETED) {
-            reportService.update(report.copy()
-                    .status(payload.getStatus())
-                    .build());
-            return;
+        final Report.Builder updatedReport = report.copy()
+                .status(payload.getStatus());
+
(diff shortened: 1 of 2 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/AggregateReportResponseHandler.java
```diff
@@ -59,6 +59,13 @@ public void handleAggregateReportResponse(final Message<AggregateResponseMessage
             return;
         }
 
+        if (report.getStatus() != COMPLETED) {
```
esotrope: should we add a code comment to the same effect?

malaffoon: I won't complain too much but i think it would be better to conditionally deal with the builder from report.copy(), then make a single call to reportService.update(builder.build()). Again, no biggie.

wtritch:
@esotrope I liked @malaffoon 's suggestion to conditionally update the Builder.  It feels cleaner this way.  Think it still needs a comment?  (Call me crazy, but I'm wary of over-commenting code as opposed to interfaces.  If the flow isn't obvious, a comment probably won't help.)

wtritch:
Merging this for now, will continue to resolve comments as they appear.

PR #1006, opened 2018-02-27
adds backend for aggregate embargo admin is embargoed endpoint

esotrope:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/UserOrganizationController.java
```diff
@@ -15,12 +15,12 @@
 
 @RestController
 @RequestMapping("/organizations")
-class OrganizationController {
+class UserOrganizationController {
```
esotrope: renamed this to `User*` to follow other things that deliver user-specific data.
any thoughts about adopting this naming practice more broadly? I found it useful when i had to distinguish between a service that gets you all things vs user-specific.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -76,6 +83,24 @@
         return Optional.empty();
     }
 
+    @Override
+    public boolean hasEmbargoedOrganizations(final User user) {
+        if (embargoRepository.isStateEmbargoed()) {
+            return true;
```
agorina: Is this right? A district user may have a released district while the state is embargoed?

esotrope: either way - findAllEmbargoedDistrictIds will return effective embargo so this is just a short-circuit to calling that

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -76,6 +83,24 @@
         return Optional.empty();
     }
 
+    @Override
+    public boolean hasEmbargoedOrganizations(final User user) {
```
agorina: I may have missed it, but I do not see a check that it is for the embargo admins only.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
wtritch: But effective embargo for a district isn't overridden by the state embargo flag, right?

esotrope: per offline call with Mark and Alla - I'll be removing this if block

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -76,6 +83,24 @@
...
+        if (embargoRepository.isStateEmbargoed()) {
+            return true;
+        }
+        final Set<Long> embargoedDistrictIds = ImmutableSet.copyOf(embargoRepository.findAllEmbargoedDistrictIds());
+        return getUserOrganizationsAsStream(user).anyMatch(organization -> {
```
wtritch: Should we be streaming organizations that the user has EMBARGO_READ access to rather than AGGREGATE_REPORT_READ here?

agorina: +1

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
esotrope: if they are state admin then they will be entitled to all districts and the block of code below will evaluate correctly. but you are right - this block could serve as a short-circuit for state wide admin

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
esotrope: hm... what if i have aggregate read for district A that is embargoed and embargo read for district B that is not embargoed? i mean it is user setup error but..

wtritch:
I'm confused by the API, and this affects the PR I have up as well.
My assumptions:
* If a user has no EMBARGO_READ rights, nothing is ever embargoed to them.  Non-existent vs embargoed is a meaningless distinction.
* A user only sees the system as Embargoed if something they have EMBARGO_READ access to is embargoed.
Some examples:
```
User has EMBARGO_READ on District A and AGGREGATE_REPORT_READ on District A and District B. 
State is embargoed, District A is embargoed, District B is embargoed:
`/embargoed` returns true.
```
```
User has EMBARGO_READ on District A and AGGREGATE_REPORT_READ on District A and District B.
State is embargoed, District A is released, District B is embargoed:
`/embargoed` returns false.  They will never see embargoed results.
```

agorina:
My understanding matches @wtritch 's summary.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -52,12 +60,13 @@
 
     @Override
     public Optional<Organization> getDefaultOrganization(final User user) {
-        if (user.getPermissionScopeByPermissionId(AggregateRead).isStatewide()) {
+        final PermissionScope permissionScope = user.getPermissionScopeByPermissionId(AggregateRead);
```
esotrope: unrelated variable extraction

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -76,9 +85,31 @@
         return Optional.empty();
     }
 
-    private Stream<Organization> getUserOrganizationsAsStream(final User user) {
-        final PermissionScope permissionScope = user.getPermissionScopeByPermissionId(AggregateRead);
+    @Override
+    public boolean hasEmbargoedOrganizations(final User user) {
```
esotrope: @wtritch @agorina @malaffoon updated logic - please re-review this and test cases

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -76,9 +85,31 @@
...
+        // we can short-circuit if the user is statewide and state embargo is in place
+        final PermissionScope permissionScope = user.getPermissionScopeByPermissionId(EmbargoRead);
+        if (permissionScope.isStatewide() && embargoRepository.isStateEmbargoed()) {
+            return true;
+        }
+
```
malaffoon: Was this where @wtritch suggested short-circuiting if the user has absolutely no EmbargoRead permissions?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -76,9 +85,31 @@
...
+                    && embargoedDistrictIds.contains(organization.getId())) {
+                return true;
+            }
+            if (type == OrganizationType.School
+                    && embargoedDistrictIds.contains(((School) organization).getDistrictId())) {
+                return true;
+            }
+            return false;
+        });
+    }
```
agorina: Looks good to me.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -76,9 +85,31 @@
...
+            final OrganizationType type = organization.getOrganizationType();
+            if (type == OrganizationType.District
+                    && embargoedDistrictIds.contains(organization.getId())) {
+                return true;
+            }
+            if (type == OrganizationType.School
+                    && embargoedDistrictIds.contains(((School) organization).getDistrictId())) {
+                return true;
+            }
+            return false;
```
malaffoon: Are group permissions expanded to district/school permissions? In other words, if i am the Embargo Admin for a group of schools, will this work?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
malaffoon: Likewise.

wtritch: Same here

PR #1008, opened 2018-02-28
improves value formatting on percentile tables

esotrope:
Before:

image

After:

image


esotrope:
scratch the date format change. This will blow it up

image

agorina:
Looks much better!

agorina:
But..but..don't you still need a year?

malaffoon:
i like the date format change. For the table with lots of columns will it just wrap the effective dates column making it two lines? That would be okay for now i think.

esotrope:
Actually tried it out and it fits:
image

should have done that before reverting : j

PR #1015, opened 2018-02-28
Bugfix/embargo ui defaults individual to released

wtritch:


webapp/src/main/webapp/src/app/admin/embargo/embargo.service.spec.ts
```diff
@@ -0,0 +1,91 @@
+import { EmbargoService } from "./embargo.service";
+import { MockDataService } from "../../../test/mock.data.service";
+import { inject, TestBed } from "@angular/core/testing";
+import { DataService } from "../../shared/data/data.service";
+import { Embargo } from "./embargo";
+import { OrganizationType, State } from "../../shared/organization/organization";
+import { of } from "rxjs/observable/of";
+
+describe('EmbargoService', () => {
```
wtritch: back-filled some tests

webapp/src/main/webapp/src/app/admin/embargo/embargo.service.ts
```diff
@@ -71,7 +72,7 @@ export class EmbargoService {
       schoolYear: source.schoolYear,
       readonly: source.readOnly,
       examCountsBySubject: source.examCounts,
-      individualEnabled: isUndefined(source.individualEnabled) ? true : source.individualEnabled,
+      individualEnabled: isUndefined(source.individualEnabled) ? false : source.individualEnabled,
```
wtritch: The big change, defaulting an undefined individual embargo status as released.

esotrope: smells like this should be backend logic to me but i could be persuaded otherwise

wtritch: I first started down that route, but then thought that "null" is a valid back-end answer if it's undefined...

wtritch: (We'd be double-defining the meaning of "null" in both the migration process and one other place either way.)

PR #1017, opened 2018-03-01
Bugfix/add distractor analysis green background

wtritch:
image

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -915,3 +915,9 @@ body .ui-widget.ui-datatable table .ui-datatable-data > tr > td.level-down,
 a.btn.disabled.with-pointer {
   pointer-events: auto;
 }
+
+//TODO move to SBAC UI kit
+distractor-analysis .ui-widget.ui-datatable table .ui-datatable-data td.green {
+  color: contrast(@green, darken(@green, 100%), lighten(@green, 100%));
+  background-color: @green;
+}


webapp/src/main/webapp/src/styles.less
```diff
@@ -915,3 +915,9 @@ body .ui-widget.ui-datatable table .ui-datatable-data > tr > td.level-down,
 a.btn.disabled.with-pointer {
   pointer-events: auto;
 }
+
+//TODO move to SBAC UI kit
+distractor-analysis .ui-widget.ui-datatable table .ui-datatable-data td.green {
+  color: contrast(@green, darken(@green, 100%), lighten(@green, 100%));
+  background-color: @green;
+}
```
wtritch: I'm using the same process as the SBAC UI kit for contrasting background with text color.

malaffoon:
Just curious: how did this break?

webapp/src/main/webapp/src/styles.less
esotrope: 👍 we'll need to for-loop this through all the `@colorVars` in the kit

wtritch:
Not sure.  Likely as a result of some style refactoring?

esotrope:
@malaffoon i think i accidentally removed the style with the intention to relocate it but never repasted it anywhere x l

webapp/src/main/webapp/src/styles.less
wtritch: Yup, it belongs in the SBAC UI kit in mixins.less eventually.

malaffoon:
Okay thanks. Just trying to understand if this was something like that (no biggie) or some systemic style issue (which would suck).

PR #1018, opened 2018-03-02
Task/widen popover

wtritch:
image


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -126,6 +129,7 @@
                      popover="{{ (!assessmentExam.assessment.isIab ? 'messages.no-writing-trait-scores-for-summative-ica-exams' : 'messages.no-writing-trait-scores') | translate }}"
                      triggers="{{ !writingTraitScoresView.disabled ? '' : 'mouseenter:mouseleave'}}"
                      placement="right"
+                     container="body"
```
wtritch: These popovers weren't bound to the body, so they were being affected by the container's styles.

PR #1019, opened 2018-03-02
Update the default embargo logic to properly role undefined district …

wtritch:
…embargo flags up to the state embargo flags.  If neither exist, default to individual released and aggregate embargoed.

### DIFF
--- a/admin-service/src/main/resources/application.sql.yml
+++ b/admin-service/src/main/resources/application.sql.yml
@@ -33,7 +33,12 @@ sql:
     # query to find all embargo records given a school year and a list of (user's) districts
     # this includes counts of summative exams grouped by each org so it has performance implications
     findAllBySchoolYear: >-
-      SELECT st.*, se.individual, se.aggregate, cnt.subject_id, cnt.count
+      SELECT
+        st.*,
+        coalesce(se.individual, 0) as individual,
+        coalesce(se.aggregate, 1) as aggregate,
+        cnt.subject_id,
+        cnt.count
       FROM (SELECT 'State' AS org_type, NULL AS org_id, NULL AS org_name, :school_year AS school_year) st
         LEFT JOIN state_embargo se ON se.school_year = :school_year
         LEFT JOIN (SELECT a.subject_id, count(e.id) AS count
@@ -41,8 +46,17 @@ sql:
                      WHERE a.type_id = 3 AND e.deleted = 0 AND e.school_year = :school_year
                      GROUP BY a.subject_id) cnt ON 1=1
       UNION ALL
-      SELECT 'District' AS org_type, d.id AS org_id, d.name AS org_name, :school_year AS school_year, de.individual, de.aggregate, cnt.subject_id, cnt.count
+      SELECT
+        'District' AS org_type,
+        d.id AS org_id,
+        d.name AS org_name,
+        :school_year AS school_year,
+        coalesce(de.individual, se.individual, 0) as individual,
+        coalesce(de.aggregate, se.aggregate, 1) as aggregate,
+        cnt.subject_id,
+        cnt.count
       FROM district d
+        LEFT JOIN state_embargo se ON se.school_year = :school_year
         LEFT JOIN district_embargo de ON d.id = de.district_id and de.school_year = :school_year
         LEFT JOIN (SELECT s.district_id, a.subject_id, count(e.id) AS count
                    FROM school s
--- a/admin-service/src/main/resources/application.sql.yml
+++ b/admin-service/src/main/resources/application.sql.yml
@@ -51,8 +51,8 @@ sql:
         d.id AS org_id,
         d.name AS org_name,
         :school_year AS school_year,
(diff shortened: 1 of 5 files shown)


admin-service/src/main/resources/application.sql.yml
```diff
@@ -33,16 +33,30 @@ sql:
     # query to find all embargo records given a school year and a list of (user's) districts
     # this includes counts of summative exams grouped by each org so it has performance implications
     findAllBySchoolYear: >-
-      SELECT st.*, se.individual, se.aggregate, cnt.subject_id, cnt.count
+      SELECT
+        st.*,
+        coalesce(se.individual, 0) as individual,
+        coalesce(se.aggregate, 1) as aggregate,
+        cnt.subject_id,
+        cnt.count
```
wtritch: simple coalese for the state, to use default values if they aren't defined.

admin-service/src/main/resources/application.sql.yml
```diff
@@ -33,16 +33,30 @@ sql:
...
+        'District' AS org_type,
+        d.id AS org_id,
+        d.name AS org_name,
+        :school_year AS school_year,
+        coalesce(de.individual, se.individual, 0) as individual,
+        coalesce(de.aggregate, se.aggregate, 1) as aggregate,
+        cnt.subject_id,
+        cnt.count
       FROM district d
+        LEFT JOIN state_embargo se ON se.school_year = :school_year
```
wtritch: Slightly more complex coalese to first fall back to state values then to default values for districts.

webapp/src/main/webapp/src/app/admin/embargo/embargo.service.ts
```diff
@@ -72,8 +67,8 @@ export class EmbargoService {
       schoolYear: source.schoolYear,
       readonly: source.readOnly,
       examCountsBySubject: source.examCounts,
-      individualEnabled: isUndefined(source.individualEnabled) ? false : source.individualEnabled,
-      aggregateEnabled: isUndefined(source.aggregateEnabled) ? true : source.aggregateEnabled
+      individualEnabled: source.individualEnabled,
+      aggregateEnabled: source.aggregateEnabled
```
wtritch: @esotrope Now that the logic includes rollup-fallbacks I moved it from the front to the back-end, like you recommended the first time =P

admin-service/src/main/resources/application.sql.yml
```diff
@@ -33,16 +33,30 @@ sql:
...
                      GROUP BY a.subject_id) cnt ON 1=1
       UNION ALL
-      SELECT 'District' AS org_type, d.id AS org_id, d.name AS org_name, :school_year AS school_year, de.individual, de.aggregate, cnt.subject_id, cnt.count
+      SELECT
+        'District' AS org_type,
+        d.id AS org_id,
+        d.name AS org_name,
+        :school_year AS school_year,
+        coalesce(de.individual, se.individual, 0) as individual,
+        coalesce(de.aggregate, se.aggregate, 1) as aggregate,
```
malaffoon: Does this work if the state is false and the district is true? In that situation, the result should be false, not true.

wtritch: hmmm...  good point.  This may not cover everything.  Let me write a test case for that: state release overrides district embargo.

wtritch: You're right, there was an issue.  Updated to re-use the same logic as the migration process.

admin-service/src/main/resources/application.sql.yml
```diff
@@ -33,16 +33,30 @@ sql:
...
                      WHERE a.type_id = 3 AND e.deleted = 0 AND e.school_year = :school_year
                      GROUP BY a.subject_id) cnt ON 1=1
       UNION ALL
-      SELECT 'District' AS org_type, d.id AS org_id, d.name AS org_name, :school_year AS school_year, de.individual, de.aggregate, cnt.subject_id, cnt.count
+      SELECT
+        'District' AS org_type,
+        d.id AS org_id,
+        d.name AS org_name,
+        :school_year AS school_year,
+        IF(se.individual = 0 OR de.individual = 0 OR (se.individual IS NULL and de.individual IS NULL), 0, 1) as individual,
```
malaffoon: oh great, now you've broken my brain

wtritch: I had to build a truth table from it to fully grok the logic =P

wtritch: (The coalesce was prettier.  Just say'in.)

agorina: You did say it! :)

webapp/src/main/webapp/src/app/admin/embargo/embargo.service.ts
esotrope: i knew it smelled like a back-end thing
oops that sounded wrong

PR #1020, opened 2018-03-02
Optionally hide percentile information display in the web UI.

wtritch:


common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
```diff
@@ -11,6 +11,7 @@
 
     private boolean transferAccessEnabled;
     private String translationLocation;
+    private boolean enablePercentileDisplay;
```
wtritch: Just added it to existing tenant properties.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/user/AppSettings.java
```diff
@@ -76,4 +76,10 @@ public boolean isTransferAccess() {
         return tenantProperties.isTransferAccessEnabled();
     }
 
+    /**
+     * @return True if UI should display norms percentile information
+     */
+    public boolean isEnablePercentileDisplay() {
+        return tenantProperties.isEnablePercentileDisplay();
+    }
```
wtritch: this is piping it through to the "user" payload returned to the front-end via the "/user" endpoint.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -157,7 +157,7 @@
                 <i class="fa fa-book"></i> {{'labels.groups.results.assessment.instruct' | translate }}
               </button>
             </li>
-            <li *ngIf="assessmentExam.assessment.isIab">
+            <li *ngIf="enablePercentileDisplay && assessmentExam.assessment.isIab">
```
wtritch: Just added it to the display *ngIf

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
```diff
@@ -39,4 +40,18 @@ public String getTranslationLocation() {
...
         this.translationLocation = translationLocation;
     }
+
+    /**
+     * This property controls whether or not to display the norms percentile
+     * information button within the web UI.
+     *
+     * @return True if the UI should display norms percentile information
+     */
+    public boolean isEnablePercentileDisplay() {
```
esotrope: would read better as percentileDisplayEnabled (isPercentileDisplayEnabled)

wtritch: ...I knew there was going to be a naming change, it it's going to touch every single file.  @malaffoon any other names come to mind?  I don't mind renaming once...

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
esotrope: should we have a default value in the yml?

wtritch: I've been debating that for all of our properties... but I think we decided to not include every configurable property in the .yml files... I don't know if this one and/or translation location and/or transfer access and/or etc should be included.  We may want to sit down and consider them all one-by-one?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/user/AppSettings.java
esotrope: hm as the props grow in number i feel like we might benefit from being explicitly listing them all in the prop files upfront

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
wtritch: done

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/TenantProperties.java
malaffoon: My vote is to use the annotated config files in the RDW project. But yeah, i get that there is some distance to that project.

PR #1027, opened 2018-03-02
Removed District/School export button from student history page

dsnyder-fairway:


dsnyder-fairway:
Thanks! checking into that now

PR #1029, opened 2018-03-02
Restructure the ICA report pdf layout to keep a single report to a si…

wtritch:
…ngle page.

MATH ICA with longest English values for every field:
image

ELA ICA with longest English values for every field:
image

### DIFF
--- a/report-processor/src/main/resources/templates/fragments.html
+++ b/report-processor/src/main/resources/templates/fragments.html
@@ -49,6 +49,30 @@
     .well.pad-small {
         padding: 10px 20px;
     }
+    .scale-score-graph .bar-container .bar {
+        border-collapse: separate;
+    }
+    .scale-score-graph.individual {
+        margin-right: 115px;
+    }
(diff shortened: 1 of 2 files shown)


report-processor/src/main/resources/templates/fragments.html
```diff
@@ -49,6 +49,30 @@
...
+        transform: translateX(-50%);
+    }
+    .scale-score-graph.individual .scale-score {
+        margin-left: 0;
+        -webkit-transform: translateX(-50%);
+        -moz-transform: translateX(-50%);
+        -ms-transform: translateX(-50%);
+        -o-transform: translateX(-50%);
+        transform: translateX(-50%);
+    }
```
wtritch: We had some hard-coded `margin-left: [xx]px;` stuff in the CSS.  I'm guess it's because the transform wasn't being respected by wkhtmltopdf.  After adding prefixing it's being respected now.

malaffoon:
Nice

PR #1030, opened 2018-03-03
Task/update rxjs usage

esotrope:
As tedious to review as it was to change : j
Please help make sure i'm not missing any `return` keywords in the `map()` operators - spent an hour trying to track one of those mistakes down x l

Also snuck in some code style updates to older components missing typed-returns and stuff

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
@@ -4,6 +4,7 @@ import { EmbargoService } from "./embargo.service";
 import { EmbargoToggleEvent } from "./embargo-toggle-event";
 import { EmbargoScope } from "./embargo-scope.enum";
 import { OrganizationType } from "./organization-type.enum";
+import { finalize } from 'rxjs/operators';
 
 /**
  * Confirmation modal displayed to confirm whether the user wants to make an embargo setting change or not
@@ -65,9 +66,11 @@ export class EmbargoConfirmationModal {
 
     // apply desired embargo setting via the API
     this.embargoService.update(embargo, scope, value)
-      .finally(() => {
-        this.modal.hide();
-      })
+      .pipe(
+        finalize(() => {
+          this.modal.hide();
+        })
+      )
       .subscribe(() => {
 
         // reflect new embargo setting in the UI
--- a/webapp/src/main/webapp/src/app/admin/groups/import/fileformat/file-format.service.ts
+++ b/webapp/src/main/webapp/src/app/admin/groups/import/fileformat/file-format.service.ts
@@ -1,21 +1,25 @@
 import { Observable } from "rxjs/Observable";
-import { Http, Response } from "@angular/http";
+import { Http } from "@angular/http";
 import { TranslateService } from "@ngx-translate/core";
 import { Injectable } from "@angular/core";
 import { Download } from "../../../../shared/data/download.model";
+import { map } from 'rxjs/operators';
 
 @Injectable()
 export class FileFormatService {
 
-  constructor(private http: Http, private translate: TranslateService) {
+  constructor(private http: Http,
+              private translate: TranslateService) {
   }
 
   public getTemplateFile(): Observable<any> {
     return this.http.get('/assets/template/groups-template.csv')
-      .map((response: Response) => new Download(
-        this.translate.instant('labels.admin-groups.import.file-format.template.file'),
-        new Blob([ response.text() ], { type: 'text/csv; charset=utf-8' })
-      ));
+      .pipe(
+        map(response => new Download(
+          this.translate.instant('labels.admin-groups.import.file-format.template.file'),
+          new Blob([ response.text() ], { type: 'text/csv; charset=utf-8' })
+        ))
+      );
   }
 
 }
--- a/webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.spec.ts
+++ b/webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.spec.ts
@@ -1,8 +1,8 @@
 import { InstructionalResourceService } from "./instructional-resource.service";
-import { Observable } from "rxjs/Observable";
-import "rxjs/add/observable/empty";
 import { InstructionalResource } from "./model/instructional-resource.model";
 import { MockDataService } from "../../../test/mock.data.service";
+import { of } from 'rxjs/observable/of';
+import { empty } from 'rxjs/observable/empty';
 
 describe("Instructional Resource Service", () => {
   let dataService: MockDataService;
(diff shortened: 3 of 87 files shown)


webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.spec.ts
```diff
@@ -1,8 +1,8 @@
 import { InstructionalResourceService } from "./instructional-resource.service";
-import { Observable } from "rxjs/Observable";
-import "rxjs/add/observable/empty";
```
esotrope: `import "rxjs/add/...` should no longer be a thing in the app

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -207,9 +210,11 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   private loadReport(): void {
     this.spinnerModal.loading = true;
     this.reportService.getAggregateReport(this.report.id)
-      .finally(() => {
-        this.spinnerModal.loading = false;
-      })
+      .pipe(
+        finalize(() => {
```
esotrope: in rxjs 5 i guess finally became finalize, catch became catchError

webapp/src/main/webapp/src/app/admin/groups/import/fileformat/file-format.service.ts
```diff
@@ -1,21 +1,25 @@
...
   public getTemplateFile(): Observable<any> {
     return this.http.get('/assets/template/groups-template.csv')
-      .map((response: Response) => new Download(
-        this.translate.instant('labels.admin-groups.import.file-format.template.file'),
-        new Blob([ response.text() ], { type: 'text/csv; charset=utf-8' })
-      ));
+      .pipe(
+        map(response => new Download(
+          this.translate.instant('labels.admin-groups.import.file-format.template.file'),
+          new Blob([ response.text() ], { type: 'text/csv; charset=utf-8' })
```
malaffoon: Just for my own edumacation: in this case it is relying on the side effect of creating the Download object which will do things on its own? So this is similar to a java stream not really collecting anything and relying on side effects in the operations?

webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.spec.ts
```diff
@@ -14,39 +14,39 @@ describe("Instructional Resource Service", () => {
...
 
   it("should find instructional resources", (done) => {
-    dataService.get.and.returnValue(Observable.of([apiResource(1), apiResource(2)]));
+    dataService.get.and.returnValue(of([ apiResource(1), apiResource(2) ]));
 
     service.findAll()
       .subscribe((resources: InstructionalResource[]) => {
         let dataArgs: any[] = dataService.get.calls.first().args;
-        expect(dataArgs[0]).toEqual("/admin-service/instructional-resources");
+        expect(dataArgs[ 0 ]).toEqual("/admin-service/instructional-resources");
```
malaffoon: We like `[ 0 ]` better than `[0]`? Really?

webapp/src/main/webapp/src/app/admin/instructional-resource/update-instructional-resource.modal.ts
```diff
@@ -32,7 +33,9 @@ export class UpdateInstructionalResourceModal implements OnDestroy {
   constructor(private modal: BsModalRef,
               private resourceService: InstructionalResourceService,
               private router: Router) {
-    this._subscription = router.events.filter(e => e instanceof NavigationStart).subscribe(() => {
+    this._subscription = router.events.pipe(
+      filter(e => e instanceof NavigationStart)
+    ).subscribe(() => {
```
malaffoon: I suppose the observable stuff is adding nifty flexibility but i have to say, for examples like this, the new syntax is less readable.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-organization.service.ts
```diff
@@ -19,16 +20,16 @@ export class AggregateReportOrganizationService {
   getOrganizationsMatchingName(nameSearch: string): Observable<Organization[]> {
     return this.dataService
       .get(`${ServiceRoute}/organizations`, { params: { name: nameSearch } })
-      .map(organizations => organizations
-        .map(organization => this.organizationMapper.map(organization))
+      .pipe(
+        map(organizations => organizations.map(organization => this.organizationMapper.map(organization)))
```
malaffoon: still need that first map call? pipe can't deal with the internal map?  i guess i should go read a tutorial on this.

esotrope: the first map transforms the observable payload, the second map transforms the array that is the payload : j

webapp/src/main/webapp/src/app/admin/instructional-resource/update-instructional-resource.modal.ts
esotrope: this new syntax is currently optional but recommended - it allows "tree shaking" by the AoT compiler ~ smaller build sizes and build times

webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.spec.ts
esotrope: intellij thinks it's better. but at the same time doesn't like `method( arg1, arg2 )` spacing : j

webapp/src/main/webapp/src/app/admin/groups/import/fileformat/file-format.service.ts
esotrope: nope - no side effects here. this takes the http response and turns it into a "download" abstraction (a name/byte array pair) which can then be sent to a 3rd party lib for download

webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
```diff
@@ -65,9 +66,11 @@ export class EmbargoConfirmationModal {
...
     this.embargoService.update(embargo, scope, value)
-      .finally(() => {
-        this.modal.hide();
-      })
+      .pipe(
+        finalize(() => {
+          this.modal.hide();
+        })
+      )
       .subscribe(() => {
```
wtritch: Does the subscribe belong in the pipe?  Or is the pipe just a set of transformations/triggers on the observable?

webapp/src/main/webapp/src/app/admin/instructional-resource/update-instructional-resource.modal.ts
wtritch: It's going to take some time to fully wrap my head around it, but I think I like the syntax with `.pipe(` on the line above rather than on it's own line...  By the time I'm 3 or 4 indents in I start to lose context...

webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
esotrope: pipe is the new way to register "processors" to an observable. subscribe actually invokes the observable to give you some data.

PR #1032, opened 2018-03-05
Handle a missing exam item response for an assessment item when viewi…

wtritch:
…ng student responses.

### DIFF
--- a/webapp/src/main/webapp/src/app/student/responses/student-responses.component.ts
+++ b/webapp/src/main/webapp/src/app/student/responses/student-responses.component.ts
@@ -44,15 +44,21 @@ export class StudentResponsesComponent implements OnInit {
   }
 
   private mapAssessmentItem(item: AssessmentItem): StudentResponsesAssessmentItem {
-    let responseItem = new StudentResponsesAssessmentItem();
+    const responseItem = new StudentResponsesAssessmentItem();
     responseItem.assessmentItem = item;
 
-    let score: ExamItemScore = item.scores.length === 1 ?  item.scores[0] : null;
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/student/responses/student-responses.component.ts
```diff
@@ -48,11 +48,11 @@ export class StudentResponsesComponent implements OnInit {
     responseItem.assessmentItem = item;
 
     let score: ExamItemScore = item.scores.length === 1 ?  item.scores[0] : null;
-    responseItem.score = score && item.scores[0].points >= 0 ? item.scores[0].points : null;
+    responseItem.score = score && score.points >= 0 ? score.points : null;
     let maxScore = item.maxPoints;
     responseItem.correctness = responseItem.score !== null ? responseItem.score / maxScore : null;
     responseItem.response = score ? score.response : null;
-    responseItem.writingTraitScores =  score.writingTraitScores;
+    responseItem.writingTraitScores =  score ? score.writingTraitScores : null;
```
wtritch: Quick fix here to add a null check before dereferencing the writingTraitScores property.

esotrope: would prefer we use `undefined` here to represent absence. `null` is generally reserved for something that was present but is now nullified.

wtritch: done, rebuilt to guard-code impl that doesn't set missing properties rather than explicit null setting.

PR #1033, opened 2018-03-05
DWR-1543 Fix to a slow sql

agorina:


report-processor/src/main/resources/application.sql.yml
```diff
@@ -215,7 +215,7 @@ sql:
         AND ${sql.reportProcessor.snippet.studentExamPermissionWithEmbargo}
 
     hasExams: >-
-      SELECT 1
+      SELECT EXISTS (SELECT 1
```
agorina: This is instantaneous ( checked in production).

malaffoon: nice

PR #1036, opened 2018-03-06
Task/remove logging for 404s

esotrope:


admin-service/src/main/java/org/opentestsystem/rdw/admin/web/InstructionalResourceController.java
```diff
@@ -58,4 +61,13 @@ public void deleteResource(@AuthenticationPrincipal final User user,
...
         service.delete(user, instructionalResource);
     }
+
+    @ExceptionHandler({
+            ResourceNotFoundException.class,
+            NoSuchElementException.class
+    })
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    public void ignore() {
+    }
```
esotrope: Added this code to all controllers where 404s are expected as a normal occurrence (the best i could anyway)
Debating whether or not to use class extension to share the code block - thoughts?

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
```diff
@@ -28,24 +31,37 @@
...
             ResourceNotFoundException.class,
-            NoSuchElementException.class,
-            AccessDeniedException.class
+            NoSuchElementException.class
     })
-    protected ResponseEntity<Object> handleNotFound(final RuntimeException exception, final WebRequest request) {
-        return opaqueError(exception, request, HttpStatus.NOT_FOUND);
+    protected ResponseEntity<Object> handleNotFoundException(final RuntimeException exception, final WebRequest request, final HttpServletRequest httpRequest) {
+        logger.info("Unhandled resource not found exception: {}", httpRequest.getRequestURI());
+        return createOpaqueError(exception, request, HttpStatus.NOT_FOUND);
```
esotrope: Added a 1-line info-level log message in the common controller advice. this is is to try and catch unexpected 404s and let us find the ones that we need to ignore.

esotrope:
Despite additional handling the following still gets logged for "ignored" 404s:

image

This can be disabled by setting this:

image

I think this is ok to do since we have a catch-all exception handler that logs in our controller advice - thoughts?

admin-service/src/main/resources/application.yml
```diff
@@ -15,6 +15,9 @@ server:
     mime-types: application/json,application/xml,text/html,text/xml,text/plain,text/css,text/javascript,application/javascript,text/csv
 
 spring:
+  mvc:
+    log-resolved-exception: false
```
malaffoon: Isn't this the default?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateReportController.java
```diff
@@ -31,4 +37,13 @@
...
         return service.findByQuery(user, query);
     }
+
+    @ExceptionHandler({
+            ResourceNotFoundException.class,
+            NoSuchElementException.class
+    })
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    public void ignore() {
+    }
```
malaffoon: Per Will's mention, should this do DEBUG logging. Or TRACE?

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
malaffoon: Perhaps add the gist of this comment in the javadocs for the method, something like `This default handler will log a 1-line INFO message for "not-found" exceptions. If an app wants to suppress this (or change the log-level), it should install its own handler.`   (assuming that is right, of course)

malaffoon: I have to say though, i wonder if we should really be doing this. I think the default behavior should be to not emit anything, and only specific failures should require coding. But i get the concern about missing things.

webapp/src/main/resources/application.yml
```diff
@@ -35,6 +35,9 @@ server:
     mime-types: application/json,application/xml,text/html,text/xml,text/plain,text/css,text/javascript,application/javascript,text/csv
 
 spring:
+  mvc:
+    log-resolved-exception: false
```
malaffoon: Still don't understand this change.

admin-service/src/main/resources/application.yml
esotrope: that's what the docs say but the logs didn't go away until i flipped this bit

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateReportController.java
wtritch: This doesn't feel right...  If we have to add something to every controller we're doing something wrong.  I have to admit, I thought we'd just be modifying the shared ExceptionHandlerControllerAdvice.  Why didn't that work?

webapp/src/main/resources/application.yml
wtritch: yeah... this also feels like we're fiddling something we shouldn't be...

esotrope: I agree that i don't like fiddling with this setting.
Without this setting spring boot will print a the log message at level WARN though : /

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateReportController.java
esotrope: it works. would we prefer just suppressing all 404s? and maybe logging them at debug level? if so i'll go ahead and do that. would make things simpler : j

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/student/StudentExamController.java
```diff
@@ -29,16 +28,14 @@ public StudentExamHistory getStudentExams(
...
-     * @param hasExams  Filter parameter
+     * @param user The authenticated user
+     * @param ssid The student SSID
      * @return The student
      */
     @GetMapping("/students/{ssid}")
     public Student getStudent(
             @AuthenticationPrincipal final User user,
-            @PathVariable("ssid") final String ssid,
-            @RequestParam("hasExams") final String hasExams) {
```
esotrope: removed this unused param

esotrope:
@wtritch @malaffoon please re-review thx

common-web/src/main/java/org/opentestsystem/rdw/reporting/common/web/ExceptionHandlerControllerAdvice.java
```diff
@@ -28,24 +31,37 @@
...
+    /**
+     * Maps {@link AccessDeniedException} to HTTP 404 instead of 403.
+     *
+     * @param exception original exception
+     * @param request   request resulting in exception
+     * @return NOT_FOUND (404) response
+     */
+    @ExceptionHandler(AccessDeniedException.class)
+    protected ResponseEntity<Object> handleAccessDeniedException(final RuntimeException exception, final WebRequest request) {
+        logger.warn(WarningPreamble, exception);
```
wtritch: I feel like this should be INFO or DEBUG.  (Again, script kiddies tend to hammer at public sites.)

esotrope: hm, don't we want to track that kind of hammering? the web server would not know it was a 403 since we mask it. that info would get lost.

wtritch: I don't think we want to track it at a WARN level, which should be firing off alerts.  At most I could see it at an INFO level, since it doesn't require human interaction but should be available if an admin is looking for it.

esotrope: cool - i'll change to info

PR #1037, opened 2018-03-06
fixes order resetting of table columns

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -104,14 +104,26 @@
...
+  <ng-container *ngTemplateOutlet="performanceLevelsTemplate; context: {
+                                   hidden: performanceLevelDisplayType !== 'Grouped',
+                                   performanceLevelDisplayType: 'Grouped'}"></ng-container>
+</p-dataTable>
+
+<ng-template #performanceLevelsTemplate
+             let-hidden="hidden"
+             let-performanceLevelDisplayType="performanceLevelDisplayType">
   <p-column *ngFor="let level of performanceLevels; let levelIndex = index"
+            [hidden]="hidden"
```
esotrope: Doh - before we were adding/removing elements when changing performance level display grouped/separate x l
Changing this to render both sets then just hide the one we don't want to see not only fixed the issue but it made the table more responsive : j

wtritch: Huh... why did adding/removing <p-column> elements revert the column ordering?

esotrope: my guess is that angular's `<ng-container *ngFor>` was bypassing the p-datatable's render process and was probably working on some cached dom state to do its rendering...or something?

PR #1038, opened 2018-03-06
display not found on Rubric and Exemplar when rubric is empty, exempla…

dsnyder-fairway:
…rs is empty and no answer key is found

avoids issue of displaying no information found and the answer key when an answer key is found along with empty rubric and exemplars

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.spec.ts
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.spec.ts
@@ -46,7 +46,7 @@ describe('ItemExemplarComponent', () => {
     expect(component).toBeTruthy();
   });
 
-  it('should report not found when item scoring guide is empty', () => {
+  it('should display not found when item scoring guide is empty', () => {
     component.ngOnInit();
     expect(component.notFound).toBeTruthy();
     expect(component.errorLoading).toBeFalsy();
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.spec.ts
```diff
@@ -38,14 +45,27 @@ describe('ItemExemplarComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
-});
 
+  it('should report not found when item scoring guide is empty', () => {
```
esotrope: nit: test names don't read well

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
```diff
@@ -37,7 +37,8 @@ export class ItemExemplarComponent implements OnInit {
 
             // TODO re-look at this logic
```
esotrope: i knew this logic looked goofy months ago when i made this comment : p

esotrope: Does this change save us any network calls btw?

dsnyder-fairway: I don't think so. Before the issue was that we would display the answer key (on line 35) and the not found message (here). I don't think this impacts traffic though

webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.spec.ts
dsnyder-fairway: suggestions? Jasmine naming is a little unfamiliar to me.

esotrope: Ah i see, "report" is a verb here. I think the real problem is i don't read well : j
I think i'd just replace the word "report" with "display" in these test names 👍

PR #1041, opened 2018-03-06
display not found on Rubric and Exemplar when rubric is empty, exempl…

dsnyder-fairway:
…ars is empty and no answer key is found

same on develop

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
@@ -37,7 +37,8 @@ export class ItemExemplarComponent implements OnInit {
 
             // TODO re-look at this logic
             this.notFound = guide.rubrics.length === 0
-              && guide.exemplars.length === 0;
+              && guide.exemplars.length === 0
+              && Utils.isNullOrEmpty(guide.answerKeyValue);
           },
           (response) => {
 
--- a/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.spec.ts
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.spec.ts
@@ -8,12 +8,19 @@ import { ItemScoringGuide } from "./model/item-scoring-guide.model";
 import { MockDataService } from "../../../../test/mock.data.service";
 import { CachingDataService } from "../../../shared/data/caching-data.service";
 import { DataService } from "../../../shared/data/data.service";
-import {AssessmentItem} from "../../model/assessment-item.model";
-import {Component} from "@angular/core";
+import { AssessmentItem } from "../../model/assessment-item.model";
+import { Component } from "@angular/core";
+
+let mockItemScoringGuide = { itemScoringGuide: new ItemScoringGuide() };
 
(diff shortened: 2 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-exemplar/item-exemplar.component.ts
```diff
@@ -37,7 +37,8 @@ export class ItemExemplarComponent implements OnInit {
 
             // TODO re-look at this logic
             this.notFound = guide.rubrics.length === 0
-              && guide.exemplars.length === 0;
+              && guide.exemplars.length === 0
+              && Utils.isNullOrEmpty(guide.answerKeyValue);
```
agorina: Nohemi was looking for instructions on how to reproduce it. Could you please add it to the JIRA?

dsnyder-fairway: unfortunately reproducing this only involved modifying the data in chrome debugger. In other words, I am not sure what database entries would be required to reproduce this.

agorina: We should be able to trace it. Just to make sure I get it: it is happening now when there is no rubric and no exemplar **AND**  `answerKeyValue` is null?

dsnyder-fairway: we would report notFound when there is no rubric and no exemplar **AND** `answerKeyValue` is null. On line 35 in this method we have
`this.model = guide;`. This sets the answer. 

prior to my changes we would provide the answer and via line 35 **AND** we would report notFound

jtreuting: shouldn't line 32 have stopped this call from happening?  because the bug had a screenshot where the answer key was shown.

agorina: I am afraid it is not the fix to the problem in production.

The use case in production is: item.answerKey is NOT null and item is 'EBSR'. 
And IRIS returns 500 on some calls ( I will add details to the JIRA).

dsnyder-fairway: @jtreuting from looking in the code it looks like this error message is only available here. From looking at this a bit further with @agorina this should only happen when this is an EBSR item. The reason why is because in `assessment-exam.mapper.ts` we have 
```
    // only multiple choice and multiple select have valid answer keys, so ignore the others
    uiModel.answerKey = (apiModel.type === 'MC' || apiModel.type === 'MS') ? apiModel.answerKey : undefined;
```
which would have set `item.answerKey` to `undefined`

agorina: Would it make sense to change that logic to include `EBSR`?

dsnyder-fairway: I am not sure. We were to display answerKey for MC/MS only. It looks like we are using item.answerKey there too. We could include a `includeInDistractorAnalysis` boolean and set it to true for MC/MS.

PR #1047, opened 2018-03-07
allow assessment-types collection to be set empty

malaffoon:
As we all know (and love), with Spring properties there is no way to set an empty value when overriding a non-empty value. This tweaks AggregateReportSettings so we can do that.

Why? Because SmarterBalanced wants to disable summative test results in production because there are no summative tests. But the state user also can't see interim results. So the state user can't see anything. And, yeah, i have no idea how the app will behave when we disable all assessment types and go to aggregate reporting.

### DIFF
--- a/aggregate-service/src/test/resources/application-ars.yml
+++ b/aggregate-service/src/test/resources/application-ars.yml
@@ -1,0 +1,4 @@
+app:
+  aggregate-reports:
+    assessment-types: iab,ica,sum
+    statewide-user-assessment-types: none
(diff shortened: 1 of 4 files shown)


aggregate-service/src/test/resources/application-ars.yml
```diff
@@ -0,0 +1,4 @@
+app:
+  aggregate-reports:
+    assessment-types: iab,ica,sum
```
esotrope: interesting. surprised `assessment-types: ''` doesn't work. I also wonder if we change this to use yaml list notation if we could make an empty list

esotrope: do we also need to use this util for the `ui-languages` setting? maybe it always has 'en' in it i cant remember

malaffoon: Yeah, for languages it is (currently) always additive and there is no default.

wtritch: The default is programmatically defined as "en" so configuration can't remove it.

PR #1052, opened 2018-03-07
removed unused translation keys. 

dsnyder-fairway:
Also

- favoring `-` as opposed to camelCase. 
- Removed some translateRoot methods as they very often did not save space, nor did they allow for easily finding where the use of these keys are.

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
@@ -33,11 +33,6 @@ import { OrganizationType } from "./organization-type.enum";
 export class EmbargoConfirmationModal {
 
   /**
-   * Translation code namespace
-   */
-  private translateContext: string = 'labels.embargo.modal';
-
-  /**
    * The event that triggered the modal
    */
   private _event: EmbargoToggleEvent;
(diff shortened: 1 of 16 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -411,123 +386,63 @@
     },
     "grade": {
       "01": {
-        "enrolled-name": "G1",
-        "long-name": "Grade 1",
-        "short-name": "G1",
```
esotrope: are these under the `common` prefix?
if so are they deleted because we don't use them yet? i put them here because we will use them once we move things to use the `common` stuff.
It's fine to delete them though as part of your process but i was just curious.
Also, I'm thinking to remain flexible, when we add different ways of displaying the same type of thing im thinking it deserves its own prefix.
(e.g.)
```javascript
"enrolled-grade.03": "Grade 3"
"enrolled-grade-short-name.03": "G3"
```
as opposed to how we see it here.

webapp/src/main/webapp/src/app/admin/embargo/embargo-confirmation-modal.component.ts
```diff
@@ -32,11 +32,6 @@ import { OrganizationType } from "./organization-type.enum";
 })
 export class EmbargoConfirmationModal {
 
-  /**
-   * Translation code namespace
-   */
-  private translateContext: string = 'labels.embargo.modal';
```
esotrope: 👍

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1643,10 +1477,7 @@
     "no-writing-trait-scores": "Writing trait scores are not available for assessments administered prior to 2017-18 school year.",
     "no-writing-trait-scores-for-summative-ica-exams": "Writing trait scores are currently available for only ELA Performance Task IABs.",
     "no-responses-for-summative-exams": "Summative assessment responses are not made available by this system.",
-    "noExams": "There are no exams associated with your account.",
-    "noExamsMatchingCriteria": "There are no exams matching this criteria.",
-    "noExamsSelected": "Please select from the assessments above",
-    "notApplicable": "-",
```
esotrope: so are all the reds here things that were not being used? i'm assuming for a lot of the things i see in red here there is a real mapping that is being used somewhere in the file right? just want to make sure we don't over aggressively prune stuff

webapp/src/main/webapp/src/assets/i18n/en.json
dsnyder-fairway: yeah they were under `common` and removed just because they aren't currently in use.

webapp/src/main/webapp/src/assets/i18n/en.json
dsnyder-fairway: the red is for item I did not find a corresponding use of. The only camelCased instance in this block that was renamed was `noExamsMatchingCriteria` -> `no-exams-matching-criteria`

dsnyder-fairway: but yes, this was pruning. Stuff I did not find any use of in our code

PR #1054, opened 2018-03-07
Fix leaking connections to S3 in the report-processor when retrieving…

wtritch:
… a V1 PDF report with a default content-type of application/octet-stream.

The problem was that we were running out of S3 connections in the connection pool.
This was happening because we were opening an InputStream to an S3 resource, then throwing an exception without closing the InputStream.

This fix converts our ReportContentResource to hold a Supplier<InputStream> lambda so that the InputStream isn't opened until it is actually being read within an AutoCloseable try/catch.

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
@@ -95,6 +95,16 @@ public class ArchiveBackedReportContentService implements ReportContentService {
         return URI.create(BasePath + report.getId() + "." + getExtension(mediaType));
     }
 
+    /**
+     * Determine the content media type.
+     * NOTE: Old V1 PDF reports are stored with the S3-default content type of
+     * 'application/octet-stream' This assumes any resource with the S3-default
+     * content type is actually 'application/pdf'
+     * TODO: Remove APPLICATION_PDF assumption once we no longer need to support V1 PDF reports
+     *
+     * @param contentType The resource content-type
+     * @return The parsed media type
+     */
     private MediaType parseMediaType(final String contentType) {
         if (isEmpty(contentType)) {
             return APPLICATION_PDF_UTF8;
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
@@ -18,6 +18,7 @@ import java.util.Set;
 
 import static org.opentestsystem.rdw.reporting.common.util.MediaTypes.APPLICATION_PDF_UTF8;
 import static org.opentestsystem.rdw.reporting.common.util.MediaTypes.getExtension;
+import static org.springframework.http.MediaType.APPLICATION_OCTET_STREAM;
 import static org.springframework.util.StringUtils.isEmpty;
 
 /**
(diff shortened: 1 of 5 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -66,17 +67,16 @@ public ReportContentResource openContent(final Report report) {
         final Properties reportProperties = archiveService.readProperties(location);
         if (!reportProperties.containsKey(Headers.CONTENT_LENGTH)) return null;
 
-        final String contentType = reportProperties.getProperty(Headers.CONTENT_TYPE);
-        final MediaType mediaType = isEmpty(contentType) ? APPLICATION_PDF_UTF8 : MediaType.parseMediaType(contentType);
+        final MediaType mediaType = parseMediaType(reportProperties.getProperty(Headers.CONTENT_TYPE));
```
wtritch: This fixes the exception that was causing the issue:
In V1 we weren't setting the ContentType on our S3 resources, since they were all of type `application/pdf`
When we tried to fetch a V1 PDF report, we were attempting to determine a file-extension for the S3-auto-assigned media type of `application/octet-stream` and throwing an exception.
This change maps both missing and default `application/octet-stream` content types to `application/pdf`

wtritch:
@esotrope either DefaultReportHandler#writeContent or AggregateReportCsvHandler#writeContent

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -95,6 +95,17 @@ public URI createUri(final Report report, final MediaType mediaType) {
         return URI.create(BasePath + report.getId() + "." + getExtension(mediaType));
     }
 
+    private MediaType parseMediaType(final String contentType) {
+        if (isEmpty(contentType)) {
+            return APPLICATION_PDF_UTF8;
```
malaffoon: `// old PDF reports won't have a content type so assume that is the case`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
```diff
@@ -95,6 +95,17 @@ public URI createUri(final Report report, final MediaType mediaType) {
...
+        }
+
+        final MediaType parsedMediaType = MediaType.parseMediaType(contentType);
```
malaffoon: Do we want to catch exceptions here, log and set to the default? Or let the exception out to be handled further up?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
wtritch: Added:
```
/**
     * Determine the content media type.
     * NOTE: Old V1 PDF reports are stored with the S3-default content type of
     * 'application/octet-stream' This assumes any resource with the S3-default
     * content type is actually 'application/pdf'
     * TODO: Remove APPLICATION_PDF assumption once we no longer need to support V1 PDF reports
     * 
     * @param contentType The resource content-type
     * @return The parsed media type
     */
```

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportContentService.java
wtritch: This shouldn't throw any exceptions.

PR #1055, opened 2018-03-07
Allow better production diagnosing via jmap, jstack, etc on our sprin…

wtritch:
…g boot processes.

The benefit is that now I'm able to exec into a running pod and do:
```
$ docker exec -it 18fd65fecd7f /bin/sh
/ # ps
PID   USER     TIME   COMMAND
    1 root       0:00 /sbin/tini -- java -Xmx768m -jar rdw-reporting-webapp-1.2
    7 root       1:19 java -Xmx768m -jar rdw-reporting-webapp-1.2.0-SNAPSHOT.ja
   63 root       0:00 /bin/sh
   68 root       0:00 ps
/ # apk add --no-cache openjdk8
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/x86_64/APKINDEX.tar.gz
(1/1) Installing openjdk8 (8.131.11-r2)
Executing java-common-0.1-r0.trigger
OK: 99 MiB in 52 packages
/ # jstack 7
2018-03-07 22:39:07
Full thread dump OpenJDK 64-Bit Server VM (25.131-b11 mixed mode):

"Attach Listener" #53 daemon prio=9 os_prio=0 tid=0x0000560e1c69f800 nid=0x5b waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"DestroyJavaVM" #52 prio=5 os_prio=0 tid=0x0000560e1f89a800 nid=0x8 waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

...
```


build.gradle
```diff
@@ -153,13 +153,15 @@ subprojects {
             label([maintainer: 'Fairway Technologies'])
             copyFile("${jar.archivePath.name}", "${jar.archivePath.name}")
             exposePort(8008, 8080)
+            runCommand("apk add --no-cache tini")
```
wtritch: More details on tini can be found (on their page.)[https://github.com/krallin/tini]

malaffoon: I'm not thrilled with this idea. I think you may just have to spend a little time convincing me it is really worth it for our default image to do this.

build.gradle
```diff
@@ -153,14 +153,10 @@ subprojects {
...
-                List<String> entryPointArgs = ["java"]
-                entryPointArgs.addAll(ext.java_opts)
-                entryPointArgs.add("-jar")
-                entryPointArgs.add("${jar.archivePath.name}")
-                entryPointArgs.toArray()
-            }
+            runCommand("apk add --no-cache tini")
+            entryPoint("/sbin/tini", "--")
+            environmentVariable("MAX_HEAP_SIZE", "-Xmx384m")
+            instruction('CMD exec java $MAX_HEAP_SIZE $JAVA_OPTS -jar ' + "${jar.archivePath.name}")
```
wtritch: I updated our dockerfile creation process to run the spring boot jar using environment-variable-configurable options.  This is necessary in order to modify heap limits if necessary for larger data sets.  A kubernetes deployment descriptor can now both declare the container memory limits AND set the java process memory limits in a single location.

esotrope: before this was hardcoded in the docker image right?

admin-service/build.gradle
```diff
@@ -1,7 +1,7 @@
 apply plugin: 'org.springframework.boot'
 
 createDockerfile {
-    ext.java_opts = ["-Xmx768m"]
+    environmentVariable("MAX_HEAP_SIZE", "-Xmx768m")
```
esotrope: nvm answered my own q

wtritch:
@esotrope that's the plan after deploying to qa and performing a quick smoke-test.  Docker likes the change, so Kubernetes should as well.

PR #1056, opened 2018-03-07
Update our package-lock.json to fix security vulnerabilities.

wtritch:


jtreuting:
looks like this also changes versions of packages.  for example, angular from 5.0 to 5.2.  Was that intentional or were you just changing the sha of the packages?

PR #1057, opened 2018-03-08
removes use of schools in the user context

esotrope:


esotrope:
@wtritch @agorina @malaffoon if there are no objections i'd like to merge and move on

PR #1063, opened 2018-03-08
 moved some messages.*, labels.export.*, labels.school-grade.*

dsnyder-fairway:


webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -222,7 +222,7 @@ export class CsvBuilder {
 
   withScaleScore(getExam: (item: any) => Exam) {
     return this.withColumn(
-      this.translateService.instant('labels.export.cols.scale-score'),
+      this.translateService.instant('csv-builder.scale-score'),
```
esotrope: for csv - i think each csv may need its own prefix. it kind of made sense actually to have the csv headings reflect the column headings because the intent of the csv is to be kind of a "print" of what you see on the screen so if the screen changes we want the csv to change (i think)

dsnyder-fairway: I renamed these to be in line with:
```
"my-component": {
    "my-sub-component-or-element": {
      "element-attribute-name": //...
    }
  }
```
`labels.export.cols` were not used in any other files

esotrope: gotcha - well in that case i think its a good change : j

dsnyder-fairway:
heads up -- alphabetizing will be the last thing to do for this ticket

PR #1066, opened 2018-03-09
fixes caching data service and home page urls

esotrope:


webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -52,12 +52,11 @@ <h1>{{'labels.homepage.title' | translate}}</h1>
   <div class="col-md-3">
 
     <!-- Welcome Content -->
-    <div class="welcome-container well">
+    <div *ngIf="applicationSettings" class="welcome-container well">
```
esotrope: not sure how this didn't rear its head in my other testing but caught it now.

webapp/src/main/webapp/src/app/home/home.component.ts
```diff
@@ -14,17 +11,13 @@ import { ApplicationSettingsService } from '../app-settings.service';
...
-  user: User;
   applicationSettings: ApplicationSettings;
 
-  constructor(userService: UserService,
-              applicationSettingsService: ApplicationSettingsService) {
+  constructor(private applicationSettingsService: ApplicationSettingsService) {
+  }
 
-    forkJoin(
-      userService.getUser(),
```
esotrope: wasn't actually using the user object in the view anymore

webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
```diff
@@ -2,38 +2,54 @@ import { Injectable } from "@angular/core";
...
-    const observable = this.dataService
-      .get(url, options)
-      .pipe(
-        share()
-      );
-
-    observable.subscribe(
-      response => {
-        this.responseByUrl[ url ] = response;
-      },
```
esotrope: This actually did not cache requests when concurrent requests for a resource were made - when i removed the explicit caching in the user service this became apparent. The fix was to use `tap()` and update the response cache before the `share()` operator. Using subscribe didn't work because `subscribe()` is called after the `share()` operator processes the response

PR #1067, opened 2018-03-09
continued moving away from labels.*

dsnyder-fairway:


webapp/src/main/webapp/src/app/admin/groups/groups.component.html
```diff
@@ -46,7 +46,7 @@ <h2>{{'labels.admin-groups.title' | translate}}</h2>
       </div>
       <!-- School Year Select -->
       <div class="col-md-2">
-        <label for="select-schoolyear">{{'common.school-year-typeahead-label' | translate}}</label>
+        <label for="select-schoolyear">{{'common.school-year-select-label' | translate}}</label>
```
dsnyder-fairway: renamed

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
```diff
@@ -53,38 +53,35 @@ <h2>{{'labels.organization-export.title' | translate}}</h2>
...
       </div>
     </form>
   </div>
 
   <div class="well gray-lightest">
-    <h3 class="gray-darkest mb-0"><i class="fa fa-list-alt"></i> {{'labels.organization-tree.header' | translate}}</h3>
-    <p class="small gray-darkest" [hidden]="selectedSchools.length != 0" [ngClass]="{invisible: editingDisabled}">{{'labels.organization-tree.subheader-empty' | translate}}</p>
-    <p class="small gray-darkest" [hidden]="selectedSchools.length == 0" [ngClass]="{invisible: editingDisabled}">{{'labels.organization-tree.subheader-not-empty' | translate}}</p>
-    <organization-tree [model]="organizationTree" (select)="remove($event)" [disabled]="editingDisabled"></organization-tree>
+    <organization-tree [model]="organizationTree" (select)="remove($event)" [disabled]="editingDisabled" [hasSelectedSchools]="selectedSchools.length != 0"></organization-tree>
```
dsnyder-fairway: moved the above within the `organization-tree`. I can move these back, but if we do I feel we should rename the keys

wtritch: I think it's reasonable to push it down

webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
```diff
@@ -120,10 +120,10 @@ <h4 class="modal-title pull-left">{{ 'labels.instructional-resource.create.title
...
 <ng-template #assessmentTemplate let-assessment="item" let-index="index" let-query="query">
   <p class="mb-0">{{assessment.label}}</p>
-  <p class="h6"><span class="label label-default">{{'labels.name' | translate}}</span> {{assessment.name}}</p>
+  <p class="h6"><span class="label label-default">{{'common.name' | translate}}</span> {{assessment.name}}</p>
 </ng-template>
 
 <ng-template #organizationTemplate let-organization="item" let-index="index" let-query="query">
   <p class="mb-0">{{organization.name}}</p>
-  <p class="h6"><span class="label label-default">{{'labels.type' | translate}}</span> {{ ('labels.instructional-resource.organizations.type.' + organization.organizationType) | translate }}</p>
+  <p class="h6"><span class="label label-default">{{'common.type' | translate}}</span> {{ ('labels.instructional-resource.organizations.type.' + organization.organizationType) | translate }}</p>
```
esotrope: common type and common name probably shouldn't exist. I would prefer these exist under a form specific qualifier so that a change to one does not change others

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -245,7 +243,7 @@ export const routes: Routes = [
         resolve: { examHistory: StudentExamHistoryResolve },
         data: {
           breadcrumb: {
-            translate: 'labels.student.results.crumb',
+            translate: 'student.results.crumb',
```
esotrope: hm don't know if we should but i'm wondering if the component prefixes should have `-component` in them or `component.` before them.  `student.` doesn't read like we are talking about `student-component` 🤔

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
esotrope: the reason i didn't put them in there was to prepare for the event that we use that tree view elsewhere under different headers. don't think we will - but in general it is good practice to set things up in this way

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
```diff
@@ -1,10 +1,13 @@
...
+<p class="small gray-darkest" [hidden]="hasSelectedSchools" [ngClass]="{invisible: disabled}">{{'organization-export.tree.subheader-empty' | translate}}</p>
+<p class="small gray-darkest" [hidden]="!hasSelectedSchools" [ngClass]="{invisible: disabled}">{{'organization-export.tree.subheader-not-empty' | translate}}</p>
 <div *ngFor="let districtTree of model.children" class="district-tree">
 
   <a *ngIf="districtTree.value.name && !disabled" class="link-button" (click)="onClick(districtTree.value)" href="javascript:void(0)">
     <h3>{{districtTree.value.name}}</h3>
   </a>
   <a *ngIf="districtTree.value.name && !disabled" class="tag tag-xs" (click)="onClick(districtTree.value)" href="javascript:void(0)">
-    <span class="pr-xs"></span> {{'labels.organization-tree.remove-district' | translate}}<span class="icon"><i class="fa fa-close"></i></span>
+    <span class="pr-xs"></span> {{'organization-export.tree.remove-district' | translate}}<span class="icon"><i class="fa fa-close"></i></span>
```
esotrope: should be `organization-export-tree` because that is the component name right?

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.ts
```diff
@@ -17,6 +17,9 @@ export class OrganizationTreeComponent {
   @Input()
   disabled: boolean = false;
 
+  @Input()
+  hasSelectedSchools: boolean = false;
```
esotrope: is there a bug fix in here somewhere? what is this for?

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
dsnyder-fairway: this was following:
```
 "my-component": {
    "my-sub-component-or-element": {
      "element-attribute-name": //...
    }
  }
```
where `tree` was the subcomponent. I can rename this to 
`organization-export.organization-tree.remove-district` if you'd like

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.ts
dsnyder-fairway: This was done to support https://github.com/SmarterApp/RDW_Reporting/pull/1067#discussion_r173554800

webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.html
dsnyder-fairway: sounds good. I'll update

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
esotrope: the translation prefix here should be the same value of the selector in the `@Component` annotation as described in the JIRA. These are enforced by angular to be unique across the whole app

webapp/src/main/webapp/src/app/app.routes.ts
dsnyder-fairway: prefer
`student-results-component.crumb` and `student-results-component.title` and then `student.results.cols.*` for usage in `student-history-iab-table.component.ts` and `student-history-ica-sumamtive-table.component.ts`?

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
dsnyder-fairway: gotcha. So `organization-export.organization-tree`

webapp/src/main/webapp/src/app/app.routes.ts
esotrope: same comment - prefix should be the `@Component()` selector whatever that is.
For the tables - we eventually want these to be the same table. so maybe give it a name already in the en.json like `student-exam-history-table` or something

webapp/src/main/webapp/src/app/organization-export/organization/organization-tree.component.html
esotrope: not exactly - no need to include the package - just `organization-tree`.
specific @Component() are under various folders but we can ignore that folder organization in the en.json

PR #1068, opened 2018-03-10
Task/convert agg report to new table

wtritch:
This PR converts our aggregate report result table view from a deprecated PrimeNG DataTable implementation to a PrimeNG Table implementation.  On a good note, the new implementation performs significantly better with large result sets.

I also noticed ExpressionChangedAfterItHasBeenCheckedError errors being thrown while modifying the organizations in the aggregate report query builder while validating the preview display of the new table.  This was mostly due to a delayed binding between the FormControl and the `organizations` property in the component.


webapp/src/main/webapp/package.json
```diff
@@ -39,7 +39,7 @@
     "mathjs": "^3.15.0",
     "ng2-file-upload": "^1.3.0",
     "ngx-bootstrap": "^2.0.0-beta.9",
-    "primeng": "5.2.0",
+    "primeng": "5.2.2",
```
wtritch: I had to bump the PrimeNG version to get access to a [pSortableColumnDisabled] property to be able to disable sorting on the bar-chart column.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-column-order-item.provider.ts
```diff
@@ -2,9 +2,9 @@ import { Injectable } from "@angular/core";
 import { OrderableItem } from "../shared/order-selector/order-selector.component";
 
 const ColumnToLabel: {[key: string]: string} = {
-  organization: 'aggregate-reports.results.cols.organization-name',
-  assessmentGrade: 'aggregate-reports.results.cols.assessment-grade',
-  schoolYear: 'aggregate-reports.results.cols.school-year',
+  organization: 'aggregate-reports.results.cols.organization',
+  assessmentGrade: 'aggregate-reports.results.cols.assessmentGrade',
+  schoolYear: 'aggregate-reports.results.cols.schoolYear',
   dimension: 'aggregate-reports.results.cols.dimension'
```
wtritch: This is fallout from changing our column identifiers to also be the translation keys so that we don't have to double-define them.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -89,8 +89,7 @@ <h3>{{'aggregate-reports.form.section.organization.heading' | translate}}</h3>
               <div class="form-group">
                 <input type="hidden"
                        [(ngModel)]="organizations"
-                       formControlName="organizations"
-                       (ngModelChange)="onSettingsChange()">
+                       formControlName="organizations">
```
wtritch: This was one of the issues in the component: toggling a checkbox changed the `organizations` model which fired this onSettingsChange() in addition to the onSettingsChange() call made by the control modifying the model, leading to some double-processing and asynchronous funkiness.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -122,7 +121,7 @@ <h3>{{'aggregate-reports.form.section.organization.heading' | translate}}</h3>
                                [disabled]="summativeFieldsDisabled"
                                [(ngModel)]="includeStateResults"
                                [ngModelOptions]="{standalone: true}"
-                               (ngModelChange)="onIncludeStateResultsChange(); onSettingsChange()">
+                               (ngModelChange)="onIncludeStateResultsChange()">
```
wtritch: rather than have the template double-call multiple listeners, I rolled secondary effects into the one-true-listener bound to the model change event.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -352,8 +354,9 @@ export class AggregateReportFormComponent {
   }
 
   private markOrganizationsControlTouched(): void {
+    this.organizationsControl.setValue(this.organizations);
     this.organizationsControl.markAsDirty();
-    this.organizations = this.organizations.concat();
+    this.organizationsControl.updateValueAndValidity();
```
wtritch: This was part of the problem, re-binding `this.organizations` to a new value didn't synchronously update the control's value, so validations were being performed against the old value.  We also weren't triggering a re-calculation of the control and form validity.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -379,6 +381,7 @@ export class AggregateReportFormComponent {
       this.settings.schools.push(<School>organization);
       this.settings.schools.sort(OrganizationComparator);
     }
+    this.markOrganizationsControlTouched();
```
wtritch: We need to mark the organizations model as touched *after* modifying it.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,158 +1,128 @@
-<spinner *ngIf="loading"></spinner>
```
wtritch: Since we no longer have any setTimeouts or pre-emptive processing and the new implementation renders quickly, we don't need a rendering timer.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,158 +1,128 @@
...
+         [autoLayout]="true"
+         styleClass="table table-striped table-hover overflow table-row-size-fixed">
+  <ng-template pTemplate="header"
+               let-columns>
+    <tr>
+      <th *ngFor="let column of columns"
+          [hidden]="!column.visible"
+          [pSortableColumn]="column.field"
+          [pSortableColumnDisabled]="!column.sortable"
+          [ngSwitch]="column.id">
```
wtritch: This loops over the columns, creating a <th> element for each one.  I'm using an ngSwitch to control the contents based upon the column type.  (Performance level column headers have special rendering for colors.)

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,158 +1,128 @@
...
+      <td *ngFor="let column of columns; let columnIdx = index;"
+          [ngSwitch]="column.id"
+          [hidden]="!column.visible">
+        <div [ngSwitch]="column.id"
+             class="full-cell"
+             [ngClass]="{
+              'pt-0': column.id === 'achievementComparison',
+              'hidden': treeColumns[i] > columnIdx,
+              'border-top': treeColumns[i] <= columnIdx && treeColumns[i] < columnOrdering.length - 1
+             }">
```
wtritch: This is the basic template for data cells.  I like the columnar approach of the new table a little better.  We can get more re-use of boiler-plate stuff.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -65,20 +66,26 @@ export class AggregateReportTableComponent implements OnInit {
...
-  public loading: boolean = true;
+  public columns: Column[] = [
+    new Column({id: "organization", field: "organization.name"}),
+    new Column({id: "assessmentGrade", field: "assessmentGradeCode"}),
+    new Column({id: "schoolYear"}),
+    new Column({id: "dimension", field: "dimension.id"}),
+    new Column({id: "studentsTested"}),
+    new Column({id: "achievementComparison", sortable: false}),
+    new Column({id: "avgScaleScore"})
+  ];
```
wtritch: Here I'm defining the basic columns for the table.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -229,37 +215,36 @@ export class AggregateReportTableComponent implements OnInit {
...
     const ordering: Comparator<AggregateReportItem>[] = this.getColumnOrdering();
-    if (event && (
-        !this._previousSortEvent ||
+
+    if (!event.field) {
+      //We're not sorting on a field.  Just apply the default column ordering
+      event.data.sort(join(...ordering));
+      this.calculateTreeColumns();
+      return;
+    }
```
wtritch: Something interesting with the new table is that you're supposed to sort the data provided in the event rather than sorting the source data, although I believe they're passed around by reference so sorting the event data also updates the source data.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -229,37 +215,36 @@ export class AggregateReportTableComponent implements OnInit {
...
-   * Retrieve the current display index of the given column.
-   *
-   * @param {Column} column A PrimeNG Column instance
-   * @returns {number}  The current display index of the column
-   */
-  public getColumnIndex(column: Column): number {
-    return this.resultsTable.columns.indexOf(column);
+  public readField(item: AggregateReportItem, field: string) {
+    return _.get(item, field, "");
   }
```
wtritch: Since I need to access the "field" for the table cell as a column property we end up with fields like
`performanceLevelByDisplayTypes.${displayType}.${this.valueDisplayType}.${index}`

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -399,6 +391,33 @@ export class AggregateReportTableComponent implements OnInit {
...
+  private updatePerformanceLevelColumns() {
+    this.columns.forEach((column) => {
+      if (column.id !== "performanceLevel") {
+        return;
+      }
+
+      column.visible = column.displayType === this.performanceLevelDisplayType;
+      column.field = `performanceLevelByDisplayTypes.${column.displayType}.${this.valueDisplayType}.${column.index}`
+    });
+  }
```
wtritch: I'm adding all the performance level columns to the table and toggling visibility/field rather than adding/removing columns to the array because there's a flash of untranslated content when new columns are added to the table.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -483,3 +492,46 @@ export interface AggregateReportTable {
...
+  level?: number;
+  index?: number;
+
+  constructor({ id,
+                field = "",
+                sortable = true,
+                visible = true,
+                displayType = "",
+                level = -1,
+                index = -1}) {
```
wtritch: This is the closest I could get to a builder pattern in typescript...

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
esotrope: oops - looks like a feedback loop there : j

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
esotrope: interesting

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,158 +1,128 @@
-<spinner *ngIf="loading"></spinner>
+<p-table #dataTable
+         [columns]="columns"
+         [rows]="100"
```
esotrope: this is not rowsPerPageOptions[0] ?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,158 +1,128 @@
...
-           [ngClass]="{'border-top': treeColumns[i] < columnOrdering.length - 1}">
-        <span *ngIf="!row.studentsTested">-</span>
-        <span *ngIf="row.studentsTested">
-          <scale-score [score]="row.avgScaleScore"
-                       [standardError]="row.avgStdErr"></scale-score>
+        <div *ngSwitchCase="'performanceLevel'"
+             class="text-center">
+          <div class="label"
+               [ngClass]="colorService.getPerformanceLevelColorsByAssessmentTypeCode(table.assessmentDefinition.typeCode, column.level)">
+            {{ getPerformanceLevelColumnHeaderTranslationCode(column.level, column.index) | translate }}
```
esotrope: can we precompute this?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,158 +1,128 @@
...
+                ('common.value-display-type-format.' + valueDisplayType) | translate:{
+                    value: readField(rowData, column.field)
+                }
+              }}
+            </span>
+          </div>
+
+          <div *ngSwitchDefault
+               class="text-right">
+            {{ readField(rowData, column.field) }}
```
esotrope: should we make a row wrapper class? we seem to have a lot of getters in the html

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
esotrope: hm. i would be tempted to move this to the ngOnInit() and do something like [col1, col2, ...getVariableColumns()] to avoid the two-part construction

webapp/src/main/webapp/src/styles.less
```diff
@@ -556,6 +556,10 @@ body .select-menu-group {
 .full-cell {
   margin: -@table-cell-padding;
   padding: @table-cell-padding;
+
+  &.pt-0 {
+    padding-top: 0;
+  }
```
esotrope: we should call this out as a ui-kit candidate i think

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
wtritch: doh, good call.  Changed.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
wtritch: Hm... yes, I think we can.  These values depend upon 3 independent variables:
* The assessment type (which shouldn't change after ngOnInit)
* The performance levels for the assessment (which shouldn't change after ngOnInit)
* The performance level display type (grouped/separate) which we're already binding a change listener to

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
wtritch: hm... I'd rather not create a wrapper for the performance level data since access will still depend upon three independent variables, two of which can change on-the-fly:
* The performance levels for the assessment (doesn't change)
* The performance level display type (changes)
* The performance value display type (changes)
So we'd still end up with a weird matrix for accessing that would have issues with using a single "performance level" column definition.

But I can get rid of this *ngSwitchDefault column definition, and replace it with a single `*ngSwitchCase="'studentsTested'"` column (which was the only column it ended up being responsible for now that development is complete.)

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
wtritch: hmmm good point, done

PR #1069, opened 2018-03-10
Make an assessment type for the state level aggregate configurable.

agorina:


aggregate-service/src/main/resources/application.yml
```diff
@@ -28,6 +28,7 @@ app:
   aggregate-reports:
     assessment-types: ica,sum
     statewide-user-assessment-types: sum
+    state-aggregate-assessment-types: sum
```
agorina: I have a feeling that some people may have strong opinions about this name :)

esotrope: this is to conditionalize which reports can display statewide results right? seems confusing that that be right next to these other configs. this or the others may need better wording or another prefix eventually.

PR #1072, opened 2018-03-13
Task/move setting endpoint to webapp

esotrope:
1. Moved system settings endpoint to webapp
2. consolidated system level configs into one property holder - ReportingSystemSettings
3. moved system settings under the "reporting" property qualifier in the yaml files


admin-service/src/main/resources/application.yml
```diff
@@ -89,11 +89,11 @@ cloud:
 archive:
   root: file:///tmp/
 
-app:
+jwt:
+  secret: ${rdw.reporting.jwt.secret:secret}
+
+reporting:
```
esotrope: change here is `app` -> `reporting`

admin-service/src/test/resources/application-adminTest.yml
```diff
@@ -11,4 +11,10 @@ permissionservice:
...
 app:
-  test-mode: true
\ No newline at end of file
+  test-mode: true
+
+reporting:
+  school-year: 2018
+  state:
+    code: CA
+    name: California
```
esotrope: i went ahead and included state name and code in each module even if only one is needed in the module

webapp/src/main/webapp/src/app/app-settings.service.ts
```diff
@@ -15,17 +14,21 @@ export class ApplicationSettingsService {
...
         reportLanguages: serverSettings.reportLanguages,
-        uiLanguages: serverSettings.uiLanguages,
+        state: {
+          code: serverSettings.state.code,
+          name: serverSettings.state.name
+        },
         transferAccess: serverSettings.transferAccessEnabled,
-        percentileDisplayEnabled: serverSettings.percentileDisplayEnabled
+        uiLanguages: serverSettings.uiLanguages,
+        userGuideUrl: serverSettings.userGuideUrl
```
esotrope: alphabetized these and added the state config because why not

webapp/src/test/resources/application-test.yml
```diff
@@ -1,11 +1,10 @@
 app:
   test-mode: true
 
+reporting:
   state:
     code: CA
-
-  iris:
-    url: http://my-iris.somewhere.org/iris/
```
esotrope: cleaned up this unused prop

admin-service/src/test/resources/application-adminTest.yml
malaffoon: 👍

webapp/src/test/resources/application-test.yml
malaffoon: is there another change for the zuul iris path, it used to use this property as a placeholder?

esotrope: @malaffoon not sure what you mean but here is the current config in the webapp app yml:
```yml
zuul:
  ...
  routes:
    iris:
      path: /iris/**
      url: ${rdw.reporting.iris.url}
      stripPrefix: true
```

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettings.java
```diff
@@ -0,0 +1,194 @@
...
+    public int getMinItemDataYear() {
+        return minItemDataYear;
+    }
+
+    public void setMinItemDataYear(final int minItemDataYear) {
+        this.minItemDataYear = minItemDataYear;
+    }
+
+    /**
+     * @return The minimum year with data
```
wtritch: ?

esotrope: the task was to move the settings not to give them sensible docs ok : j

PR #1074, opened 2018-03-13
Changed the agg. report SQL to see if it helps with getting a better execution plan.

agorina:
I just want to test and see if this helps Redshift with choosing  `XN GroupAggregate` for school reports.

 https://docs.aws.amazon.com/redshift/latest/dg/c_designing-queries-best-practices.html
_"Use sort keys in the GROUP BY clause so the query planner can use more efficient aggregation.A query might qualify for one-phase aggregation when its GROUP BY list contains only sort key columns, one of which is also the distribution key. The sort key columns in the GROUP BY list must include the first sort key, then other sort keys that you want to use in sort key order. For example, it is valid to use the first sort key, the first and second sort keys, the first, second and third sort keys, and so on.It is not valid to use the first and third sort keys.You can confirm the use of one-phase aggregation by running the EXPLAIN command and looking for XN GroupAggregate in the aggregation step of the query."_

### DIFF
--- a/aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
+++ b/aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
@@ -67,7 +67,7 @@ sqlbuilder:
               clauses:
                 select: >-
                      'School' AS organization_type,
-                     sch.id AS organization_id
+                     fe.school_id AS organization_id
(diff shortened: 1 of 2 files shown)


agorina:
Merging since I need it to continue testing. Will address comments if needed.

PR #1075, opened 2018-03-13
Add better error handling to the group upload component.

wtritch:
While testing handling large-sized group uploads I noticed that we weren't handling communication errors/5XX responses well in the UI.  This PR both increases the file upload limit in the webapp to match the admin-service and improves error handling in the UI.

Note that I've changed the error message slightly from the image below to "Failed to upload groups to server."
group_error_handling

### DIFF
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -60,6 +60,11 @@ spring:
   session:
     store-type: none
 
+  http:
+    multipart:
+      max-file-size: 500MB
+      max-request-size: 500MB
+
   # still need this for translations api
   jackson:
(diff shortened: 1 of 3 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -60,6 +60,11 @@ spring:
   session:
     store-type: none
 
+  http:
+    multipart:
+      max-file-size: 500MB
+      max-request-size: 500MB
```
wtritch: In admin-service we already had a max file size set to 500MB by default.  This increases the webapp API to handle the same size uploads.  It seems a little high to me, but I assume we had a reason for selecting 500MB limits initially?

PR #1076, opened 2018-03-13
fixes autocomplete on homepage typeahaed

esotrope:


webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -15,13 +15,13 @@ import { byString, join } from "@kourge/ordering/comparator";
            [disabled]="disabledInternal"
            [typeahead]="options"
            [typeaheadMinLength]="0"
-           typeaheadWaitMs="300"
```
esotrope: this made the typeahead wait 300ms to display after clicking even though the data was in-hand : /
(this component is currently only used on the home page and school results page where the schools are in-hand after page load)

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -15,13 +15,13 @@ import { byString, join } from "@kourge/ordering/comparator";
...
            typeaheadOptionField="label"
            typeaheadGroupField="group"
            (typeaheadOnSelect)="onSelectInternal($event.item)"
            (ngModelChange)="onChangeInternal()"
            [(ngModel)]="search"
-           [placeholder]="placeholder">
+           placeholder="{{placeholder}}"
+           autocomplete="off">
```
esotrope: the fix

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
wtritch: I guess we'll need to put that back once we connect it to a back-end query? =)

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
wtritch: I wonder if we want to add autocomplete="off" to all of our inputs... I don't suppose we could apply it globally with a CSS selector? (Without using a DOM accessor like jQuery)

PR #1077, opened 2018-03-14
Task/convert assessment student table

wtritch:
This PR converts the assessment "by student" results table from the old PrimeNg DataTable implementation to the new PrimeNg Table implementation.

image

image


webapp/src/main/webapp/src/app/shared/button/information-button.component.html
```diff
@@ -5,7 +5,8 @@
         popoverTitle="{{ title }}"
         [placement]="placement"
         container="body"
-        outsideClick="true">
+        outsideClick="true"
+        (click)="propagateClick($event)">
```
wtritch: While testing the new Table implementation I found that clicks on (i) info icons in headers sorted the table by the clicked header.
Since this is a change in functionality and is unexpected behavior, I changed the information-button component to stop propagating it's click event, then create and fire a document-level click event to trigger any registerd "close on outside click" behaviors.  It feels hacky and dirty to manually create and dispatch a "click" event, but it was the only way @esotrope and myself could think of to avoid changing the UX.  Any objections?

webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.html
```diff
@@ -28,107 +28,120 @@
...
-            [hidden]="!isClaimScoreSelected">
-    <ng-template let-exam="rowData" pTemplate="body">
-      <span *ngIf="exam.claimScores[idx].level">{{ 'common.assessment-type.iab.performance-level.' + exam.claimScores[idx].level.toString() + '.name' | translate }}</span>
-      <span *ngIf="!exam.claimScores[idx].level">-</span>
-    </ng-template>
-  </p-column>
-</p-dataTable>
+<p-table [columns]="columns"
+         [value]="exams"
+         [autoLayout]="true"
```
esotrope: what does this do?

wtritch: It recalculated column widths on data population/modification

PR #1078, opened 2018-03-14
Handle the asynchronous modification of the minimumItemDataYear property

wtritch:
This fixes an issue where the "minimumItemDataYear" setting isn't loaded/asynchronously bound on the assessment results display, leading to disabled options like "View Distractor Analysis."
This is fairly difficult to reproduce since we cache the system settings results.  You have to hard-refresh on the home page then navigate directly to a group's assessments (with results for the current year) then try to display the distractor analysis without changing assessments.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
@@ -108,7 +108,14 @@ export class AssessmentResultsComponent implements OnInit {
    * If there are no exams that are after this school year, then disable the ability to go there and show proper message
    */
   @Input()
-  minimumItemDataYear: number;
+  set minimumItemDataYear(value: number) {
+    this._minimumItemDataYear = value;
+    this.initializeViews();
+  };
+
+  get minimumItemDataYear(): number {
+    return this._minimumItemDataYear;
+  }
 
   /**
    * Exam filters applied, if any.
@@ -214,6 +221,7 @@ export class AssessmentResultsComponent implements OnInit {
   private _filterBy: FilterBy;
   private _assessmentExam: AssessmentExam;
   private _filterBySubscription: Subscription;
+  private _minimumItemDataYear: number;
 
   constructor(private applicationSettingsService: ApplicationSettingsService,
               public colorService: ColorService,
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -108,7 +108,14 @@ export class AssessmentResultsComponent implements OnInit {
...
   @Input()
-  minimumItemDataYear: number;
+  set minimumItemDataYear(value: number) {
+    this._minimumItemDataYear = value;
+    this.initializeViews();
+  };
+
+  get minimumItemDataYear(): number {
+    return this._minimumItemDataYear;
+  }
```
wtritch: We were creating the result view state objects in ngOnInit, then the minimum item year was resolving to a value, but the view state objects and their "enabled/disabled" state were already initialized.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -108,7 +108,14 @@ export class AssessmentResultsComponent implements OnInit {
    * If there are no exams that are after this school year, then disable the ability to go there and show proper message
    */
   @Input()
-  minimumItemDataYear: number;
+  set minimumItemDataYear(value: number) {
+    this._minimumItemDataYear = value;
+    this.initializeViews();
+  };
```
esotrope: Hm i think it would be a bit nicer to make sure that all the necessary variables are set before computing the view states that depend on those variables. alternatively we could lazy compute the disabled aspects of the view states by passing in lambdas instead of the variable value

wtritch: Hmm...  waiting for all required variables to be set feels like it could be a little complicated. I'm debating between a lambda that's run on every digest or having logic that runs this.initializeViews() on all of the \@Input properties that can affect it...  This is slightly higher maintenance, the other is slightly worse performance...  I could go either way, thoughts?

esotrope: hm a 3rd option which i think may be simplest would be to wrap the `<assessment-results>` component with an `<ng-container *ngIf="minimumDataYear">` so it is not loaded at all until this setting is available.

wtritch: I debated that, but I think that every component should deal with changing input values.  Otherwise we could have deep-knowledge bleed of whether this or that [input] on a child component respects data changes/late bindings.  I could see that becoming a mess pretty quickly.

wtritch: Discussed offline, we decided to go with binding `this.updateViews()` to all \@Inputs that can  affect the results.

PR #1080, opened 2018-03-14
fixes org mapper test

esotrope:


webapp/src/main/webapp/src/app/organization-export/organization/organization.mapper.ts
```diff
@@ -145,7 +147,9 @@ export class OrganizationMapper {
     }
 
     const schoolGroup: SchoolGroup = new SchoolGroup();
-    schoolGroup.id = id;
+    if (!Utils.isNullOrUndefined(id)) {
+      schoolGroup.id = id;
+    }
```
esotrope: the annoying thing about `undefined` is
{id: undefined} != {} when doing property by property comparison : [

PR #1081, opened 2018-03-14
fixes modal header overflow

esotrope:


webapp/src/main/webapp/src/styles.less
```diff
@@ -919,11 +919,17 @@ distractor-analysis .ui-widget.ui-datatable table .ui-datatable-data td.green {
...
 .form-control {
   &.btn-primary {
     color: @btn-primary-color;
     background-color: @btn-primary-bg;
     border-color: @btn-primary-border;
   }
 }
+
+.modal-title {
+  max-width: 96%;
```
esotrope: this width choice prevents the `x` button from getting pushed to a new line. i could have changed the markup to wrap the modal `<h>` tag and close button tags with grid stuff do this more robustly but i thought it might be overkill

dsnyder-fairway:
@esotrope are changes to "Search by My Groups" table coming later?

esotrope:
@dsnyder-fairway yeah working on that

webapp/src/main/webapp/src/styles.less
wtritch: % and px widths always make me do a double-take... Because sooner or later something we didn't account for comes up.  But if the alternative is a heavy lift... /shrug

esotrope: yeah i realize this kicks the can down the road a bit

PR #1083, opened 2018-03-14
fixes groups width using shared label-max-width class

esotrope:
Would be nicer if the ellipsis went right up to the edge of the cell but that is easier said than done.

This fix uses the label-max-width css class to be consistent with when we do this in other tables.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/groups.component.html
+++ b/webapp/src/main/webapp/src/app/groups/groups.component.html
@@ -3,56 +3,54 @@
 </ng-container>
 
 <ng-template #loaded>
-  <div class="home-view">
+  <div class="flex-child grow">
+    <label>{{'groups.search-label' | translate }}</label>
+  </div>
 
-    <div class="flex-child grow">
-      <label>{{'groups.search-label' | translate }}</label>
-    </div>
+  <!-- My Groups Search Input -->
+  <div *ngIf="groups.length > 2"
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -3,56 +3,54 @@
 </ng-container>
 
 <ng-template #loaded>
-  <div class="home-view">
```
esotrope: the red is due to removing this curious wrapper

webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -3,56 +3,54 @@
...
+                 emptyMessage="{{ emptyMessageTranslateKey | translate}}"
+                 sortField="name"
+                 tableStyleClass="table table-striped table-hover">
+
+      <p-column field="name"
+                header="{{'groups.columns.group' | translate}}"
+                [sortable]="true">
+
+        <ng-template let-group="rowData" pTemplate="body">
+          <div class="label-max-width">
```
esotrope: the change

webapp/src/main/webapp/src/app/groups/groups.component.html
wtritch: Most curious.

PR #1084, opened 2018-03-14
Re-scope a private property that is accessed by the component hmtl.

wtritch:


webapp/src/main/webapp/src/app/assessments/results/view/distractor-analysis/distractor-analysis.component.ts
```diff
@@ -58,9 +58,9 @@ export class DistractorAnalysisComponent implements OnInit, ExportResults {
 
   loading: boolean = false;
   columns: Column[];
+  filteredMultipleChoiceItems: AssessmentItem[];
 
   private _multipleChoiceItems: AssessmentItem[];
-  private filteredMultipleChoiceItems: AssessmentItem[];
```
wtritch: Hm.  I'm not quite sure how this worked before... I thought properties accessed by the template still need to be public?

esotrope: yeah this happened to me before too where something private worked and then stopped. i think it might work if it is accessed in the root template? but when a nested template tries to use it it fails maybe?

PR #1085, opened 2018-03-15
Task/migrate results by item

wtritch:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/view/distractor-analysis/distractor-analysis.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/view/distractor-analysis/distractor-analysis.component.html
@@ -3,10 +3,12 @@
     <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i><span class="text">{{ 'common.messages.loading' | translate }}</span>
   </div>
 </div>
-<p-table #distractorTable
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/assessments/results/view/distractor-analysis/distractor-analysis.component.html
```diff
@@ -39,7 +41,7 @@
           </span>
         </span>
 
-        <p-sortIcon [field]="column.field"></p-sortIcon>
+        <p-sortIcon [field]="column.sortField"></p-sortIcon>
```
wtritch: semi-related.  Noticed a couple of issues in the distractor table impl while working on the results-by-item table.

malaffoon:
Hmm, unrelated but why does it go to 10? Is this in awsqa, awsdev, local dev snapshot?

wtritch:
@malaffoon It's because of the localhost test data:
image

This is in the Grade 11 ELA ICA exam.  Is there a bug here?

malaffoon:
Yeah, i'll add a TODO to fix asmt labels and WER max scores in the dev snapshot.

wtritch:
Merging this now, will resolve comments as they appear.

PR #1089, opened 2018-03-15
DWR-1218 Add IAB support to the custom aggregate report.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateService.java
```diff
@@ -59,7 +58,6 @@ private ReportQuery checkIsValid(final ReportQuery query) {
         checkArgument(!query.getSchoolYears().isEmpty(), "report query is missing school years");
         checkArgument(!query.getDimensionTypes().isEmpty(), "report query is missing dimensionTypes");
         checkArgument(!(!query.isIncludeState() && !query.isIncludeAllDistricts() && query.getDistrictIds().isEmpty() && query.getSchoolIds().isEmpty()), "report query is missing organization");
-        checkArgument(!IAB.code().equals(query.getAssessmentTypeCode()), "custom aggregate report for IAB is not supported yet");
```
agorina: Oops, need to modify some tests. Working on it.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryProviderRST.java
```diff
@@ -3,6 +3,7 @@
```
agorina: I have not added any repository RST for this change.
Since (at least for now), the only difference is in the FROM clause, I felt that this is good enough for testing.
Please let me know if you disagree.

PR #1090, opened 2018-03-15
truncates long group names on admin table and adds title attribute to…

esotrope:
… both tables for group names

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/groups/groups.component.html
+++ b/webapp/src/main/webapp/src/app/admin/groups/groups.component.html
@@ -5,98 +5,124 @@
   <ng-template #content>
     <ul class="list-unstyled list-inline">
       <li>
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/admin/groups/groups.component.html
```diff
@@ -5,98 +5,124 @@ <h2>{{'admin-groups.title' | translate}}</h2>
...
     </ul>
   </ng-template>
 </page-heading>
 
-<ng-container *ngIf="filterOptions && filterOptions.schools.length == 0">
-  <div class="alert alert-danger">
-    {{ 'admin-groups.no-schools' | translate }}
-  </div>
-</ng-container>
+<ng-container *ngIf="filterOptions">
```
esotrope: re organized these if blocks to be less redundant.
also changed the "no schools" message to be in an `alert-info` instead of `alert-danger` since its not an error

PR #1092, opened 2018-03-16
Fix broken tests

wtritch:


webapp/src/main/webapp/src/app/admin/groups/groups.component.spec.ts
```diff
@@ -44,26 +42,25 @@ describe('GroupsComponent', () => {
...
-          useValue: { snapshot: mockRouteSnapshot, params: mockParams }
-        },
-        { provide: GroupService, useValue: mockGroupService },
-        { provide: Router, useValue: MockRouter },
-        { provide: UserService, useClass: MockUserService }
-      ]
+        { provide: BsModalService, useValue: mockModalService },
+        { provide: GroupService, useValue: mockGroupService }
+      ],
+      schemas: [ NO_ERRORS_SCHEMA ]
```
wtritch: This looked like a pretty old test.  Made some changes to bring it a little closer to the current standards.

PR #1094, opened 2018-03-16
Migrate the group administration tables to the new implementation

wtritch:
This PR migrates the group admin tables to the new table implementation.

image

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/groups/import/import-table/import-table.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/groups/import/import-table/import-table.component.ts
@@ -1,11 +1,11 @@
-import { Component, Input, OnInit } from '@angular/core';
+import { Component, Input } from '@angular/core';
 import { ImportResult } from "../import-result.model";
 
 @Component({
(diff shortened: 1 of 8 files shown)


webapp/src/main/webapp/src/app/admin/groups/import/import-table/import-table.component.ts
```diff
@@ -1,21 +1,36 @@
...
    */
   @Input()
   public imports: ImportResult[] = [];
 
-  constructor() { }
+  columns: Column[] = [
+    new Column({id: 'file-name', field: 'fileName'}),
+    new Column({id: 'created'}),
+    new Column({id: 'status'}),
+    new Column({id: 'id'})
```
esotrope: even if it is redundant i feel like it is more clear to explicitly state the field name here

wtritch: Even if it duplicates the id? ...ehhh, I'm not sure I agree.  Especially since it happens so often.

PR #1095, opened 2018-03-16
Migrate the embargo administration table to the new implementation

wtritch:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.html
+++ b/webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.html
@@ -1,82 +1,97 @@
-<p-dataTable #datatable
-             [value]="embargoes"
-             sortField="organization.name"
-             tableStyleClass="table table-striped table-hover"
-             [tableStyle]="{'table-layout':'auto'}">
+<p-table #embargoTable
+         [columns]="columns"
+         [value]="embargoes"
+         [autoLayout]="true"
+         sortField="organization.name"
+         styleClass="table table-striped table-hover">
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -46,7 +46,6 @@
                let-columns="columns">
     <tr>
       <td *ngFor="let column of columns; let columnIdx = index;"
-          [ngSwitch]="column.id"
```
wtritch: Unrelated cleanup, noticed a double-switch in the aggregate report table

wtritch:
Merging this now, will resolve any comments as they arrive.

webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.html
```diff
@@ -1,82 +1,96 @@
...
+
+  <ng-template pTemplate="header"
+               let-columns>
+    <tr>
+      <th *ngFor="let column of columns"
+          [ngSwitch]="column.id">
+
+        <div *ngSwitchCase="'subject'"
+              class="text-right">
+          {{translateContext + '.headers.' + column.code | translate}}
```
esotrope: we should standardize our column interfaces probably i think i've seen us use code/headerKey etc.

wtritch: I keep debating that, but they're also super-tightly coupled to their individual views... I'm not sure this is something I *want* to have a base-class or interface for...

esotrope: i feel like - base class - _no_, established convention - _ideally_

PR #1096, opened 2018-03-16
Add IAB support to custom aggregate report - round 2.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryWithReportTemplateProvider.java
```diff
@@ -67,53 +69,55 @@ public Builder createBuilderWithCommonAttributes(final ReportQuery query) {
...
-    public Builder createBuilderWithState() {
-        return super
-                .createBuilderWithState()
-                .addOn(stateAddOn)
-                .parameter("state_embargo_admin", getScopeOrEmpty(permissionSource.getPermissionsById(), EmbargoRead).isStatewide());
+    public Builder createBuilderWithState(final String assessmentTypeCode) {
+        final Builder builder = super
+                .createBuilderWithState(assessmentTypeCode)
+                .addOn(stateAddOn);
+        return assessmentTypeCode.equalsIgnoreCase(SUMMATIVE.code()) ?
```
agorina: There is really no reason to add embargo to IAB or ICA.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
```diff
@@ -8,10 +8,10 @@
...
     Integer avgScaleScore;
     Integer avgStdErr;
-    int level1Count;
-    int level2Count;
-    int level3Count;
-    int level4Count;
+    Integer level1Count;
+    Integer level2Count;
+    Integer level3Count;
+    Integer level4Count;
```
agorina: Changed them all. This allows for the backfilled data to be NULL.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
```diff
@@ -21,33 +21,41 @@ public Integer getAvgStdErr() {
...
+    public Integer getLevel3Count() {
         return level3Count;
     }
 
-    public int getLevel4Count() {
+    public Integer getLevel4Count() {
         return level4Count;
     }
 
+    public static int studentCount(final Measures measures) {
```
agorina: This is used in a few places.

aggregate-service/src/main/resources/custom-aggregate-report-template.sql.yml
```diff
@@ -98,15 +94,21 @@ sqlbuilder:
                 groupBy: >-
                      fe.school_id
 
-            embargo:
+            districtEmbargo:
               clauses:
                 where: >-
                     fe.school_year < ${reporting.school-year}
-                    OR 3 <> :asmt_type_id
```
agorina: This is now controlled in Java.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryWithReportTemplateProvider.java
malaffoon: nit: i think this would read better with just a conditional to add embargo info and then return builder, something like:
```
    final builder builder = super
        .createBuilderWithState(assessmentTypeCode)
        .addOn(stateAddOn);
    // apply embargo add-on for summatives
    if (assessmentTypeCode.equalsIgnoreCase(SUMMATIVE.code()) {
        builder.addOn(stateEmbargoAddOn)...;
    }
    return builder;
```

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateReportContextProvider.java
```diff
@@ -36,27 +36,27 @@ public B createBuilderWithCommonAttributes(final ReportQuery query) {
     }
 
     @Override
-    public B createBuilderWithState() {
+    public B createBuilderWithState(final String assessmentTypeCode) {
```
malaffoon: i don't know this code very well but it seems ... unfortunate ... to have to change the base class signature because a particular implementation needs more info; normally this would be handled by some sort of stateful provider.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
malaffoon: might be nice to get a couple saved reports into a unit test to make testing the serialization easier

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
malaffoon: I admit that i don't understand why this is called studentCount?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateReportContextProvider.java
agorina: Discussed offline - I will merge it as is, but open to refactoring later once @malaffoon gets more familiar with this code. Will come back to it on Monday.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
agorina: Clarified offline. Will change it to a non-static methods.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
esotrope: just curious - why isn't this represented as a list?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateProviderTemplate.java
```diff
@@ -99,7 +99,7 @@ private void appendSchools(final ReportQuery query, final List<B> builders, fina
             final Set<Long> schoolsIncludedAsAllOfDistricts = isIncludeAllSchoolsOfDistricts(query) ? repositoryHelper.findSchoolsIdsByDistrictIds(districtsWithAllSchools) : emptySet();
             schoolIds.removeAll(schoolsIncludedAsAllOfDistricts);
 
-            if (!schoolIds.isEmpty()) builders.add((B) createBuilderWithSchoolIds(schoolIds).merge(dimensionTemplateBuilder));
+            if (!schoolIds.isEmpty()) builders.add((B) createBuilderWithSchoolIds(query.getAssessmentTypeCode(), schoolIds).merge(dimensionTemplateBuilder));
```
esotrope: nit: would prefer brackets

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
```diff
@@ -21,33 +21,41 @@ public Integer getAvgStdErr() {
...
+        return toZeroIfNull(measures.getLevel1Count()) + toZeroIfNull(measures.getLevel2Count()) + toZeroIfNull(measures.getLevel3Count()) + toZeroIfNull(measures.getLevel4Count());
+    }
+
+    private static int toZeroIfNull(final Integer integer) {
```
esotrope: `nullToZero` would probably be the java naming convention here. otherwise i would as `orZero`

agorina: Can somebody please point me to the  java naming convention bible? :)

(will change to nullToZero).

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
agorina: About saved report: since they expire in a month and we do not plan on releasing it in a month, I do not think I want to spend time on introducing this test. Unless people strongly feel otherwise.

As for the list, maybe...not now. Too many changes at once and since there is only 4 for now it feels cleaner to deal with them this way.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateQueryWithReportTemplateProvider.java
agorina: Changed.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/CustomAggregateProviderTemplate.java
agorina: Changed.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
esotrope: *points at head* x J

PR #1097, opened 2018-03-16
Migrate the item scores table to the new implementation.

wtritch:
This PR migrates the "Student Scores and Responses" nested table to the new table implementation.

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-scores/item-scores.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-scores/item-scores.component.html
@@ -1,63 +1,133 @@
 <p>{{ 'item-scores.student-scores-intro-text' | translate }}</p>
-<p-dataTable #datatable [value]="scores" tableStyleClass="table table-striped table-hover" [tableStyle]="{'table-layout':'auto'}" sortField="student.lastName" expandableRows="true" rowExpandMode="single">
-  <p-column sortField="student.lastName" header="{{'common.results.assessment-exam-columns.name' | translate}}" [sortable]="true">
-    <ng-template let-row="rowData" let-score="rowData" pTemplate="body">
-      <datatable-row-expander [datatable]="datatable" [row]="row" [text]="'common.person-name' | translate:score.student"></datatable-row-expander>
-    </ng-template>
-  </p-column>
-  <p-column field="date" header="{{'common.results.assessment-exam-columns.date' | translate}}" [sortable]="true">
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-scores/student-score.model.ts
```diff
@@ -3,6 +3,7 @@ import { School } from "../../../school-grade/school";
 import { WritingTraitScores } from "../../model/writing-trait-scores.model";
 
 export class StudentScore {
+  examId: number;
```
wtritch: The new table implementation requires a unique row identifier when using row expansion.  I just added the exam id to the StudentScore class to use as the identifier.

webapp/src/main/webapp/src/styles.less
```diff
@@ -716,7 +716,7 @@ aggregate-report-form {
   }
 }
 
-.writing-trait-scores .ui-table-wrapper {
+.writing-trait-scores > .ui-table-wrapper {
     display: inline-block;
     min-width: 65%;
 }
```
wtritch: I forgot to account for nested tables here.  Now only the top-level writing-trait-scores table gets the min-width treatment.

webapp/src/main/webapp/src/app/assessments/items/item-scores/item-scores.component.html
```diff
@@ -1,63 +1,133 @@
...
+
+        <div *ngSwitchCase="'grade'">
+          {{ 'common.enrolled-grade-label.' + score.enrolledGrade | translate }}
+        </div>
+
+        <div *ngSwitchCase="'school'">
+          {{ score.school.name }}
+        </div>
+
+        <div *ngSwitchCase="'item-score'"
```
esotrope: not sure i understand why we need switch case on item-score and evidence

wtritch: Because I wanted to treat their column headers the same rather than call out each of the 13 columns independently >_<  All "item-score" columns use `{{ ('item-scores.' + column.score) | translate }}` for header text, all "trait" columns use ` {{ ('common.writing-trait.' + column.score) | translate }}` and all other columns use `{{ ('common.results.assessment-exam-columns.' + column.id) | translate }}`
...this is a surprisingly complex table for what it's showing.

esotrope: gotcha - interesting

PR #1098, opened 2018-03-16
Migrate the instructional resource administration table to the new im…

wtritch:
…plementation.

image

### DIFF
--- a/webapp/src/main/webapp/src/primeng-overrides.less
+++ b/webapp/src/main/webapp/src/primeng-overrides.less
@@ -29,6 +29,7 @@ body .ui-table {
   font-size: @font-size-small;
   table {
     border-collapse: separate;
+    border: none;
     width: auto;
     max-width: none;
     min-width: 100%;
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -168,7 +168,8 @@ h1 + p.pull-left, .h1 + p.pull-left, h2 + p.pull-left, .h2 + p.pull-left, h3 + p
   border: 1px solid @gray-lighter;
   border-radius: @border-radius-base @border-radius-base 0 0;
   padding: @padding-base-horizontal @padding-large-horizontal;
-  + .table-list-container .table > thead > tr {
+  + .table-list-container .table > thead > tr,
+  + .table-list-container > p-table > .table > .ui-table-wrapper > table > thead > tr {
```
wtritch: The new table implementations were missing the "don't round the top corners if you're just below a search/filter box" style.

webapp/src/main/webapp/src/primeng-overrides.less
```diff
@@ -29,6 +29,7 @@ body .ui-table {
   font-size: @font-size-small;
   table {
     border-collapse: separate;
+    border: none;
```
wtritch: The new table implementations have been double-bordering.  This turns off the default border.

PR #1099, opened 2018-03-17
Feature/add assessment label column for iab aggregate reports

esotrope:
Almost forgot a pretty picture : j

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -617,7 +617,7 @@
                                        analyticsEvent="Change"></sb-radio-button-group>
               </span>
               <span class="pull-right flex-child ml-xs"
-                    *ngIf="currentAssessmentDefinition.typeCode !== 'iab'">
+                    *ngIf="currentAssessmentDefinition.performanceLevelDisplayTypes.length > 1">
                 <label class="mr-xs">{{'common.performance-level-display-type-input-label' | translate}}</label>
                 <sb-radio-button-group [options]="options.performanceLevelDisplayTypes"
                                        [(ngModel)]="settings.performanceLevelDisplayType"
@@ -634,7 +634,10 @@
                                     [preview]="true"
                                     [table]="previewTable"
                                     [valueDisplayType]="settings.valueDisplayType"
-                                    [performanceLevelDisplayType]="settings.performanceLevelDisplayType"
+                                    [performanceLevelDisplayType]="
+                                      currentAssessmentDefinition.performanceLevelDisplayTypes.includes(settings.performanceLevelDisplayType)
+                                        ? settings.performanceLevelDisplayType
+                                        : currentAssessmentDefinition.performanceLevelDisplayTypes[0]"
                                     [identityColumns]="settings.columnOrder"></aggregate-report-table>
           </div>
         </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -616,7 +616,8 @@
                                        analyticsCategory="AggregateQueryBuilder"
                                        analyticsEvent="Change"></sb-radio-button-group>
               </span>
-              <span class="pull-right flex-child ml-xs">
+              <span class="pull-right flex-child ml-xs"
+                    *ngIf="currentAssessmentDefinition.typeCode !== 'iab'">
                 <label class="mr-xs">{{'common.performance-level-display-type-input-label' | translate}}</label>
                 <sb-radio-button-group [options]="options.performanceLevelDisplayTypes"
                                        [(ngModel)]="settings.performanceLevelDisplayType"
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -635,7 +635,7 @@
                                     [table]="previewTable"
                                     [valueDisplayType]="settings.valueDisplayType"
                                     [performanceLevelDisplayType]="settings.performanceLevelDisplayType"
-                                    [columnOrdering]="settings.columnOrder"></aggregate-report-table>
+                                    [identityColumns]="settings.columnOrder"></aggregate-report-table>
           </div>
         </div>
       </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -184,7 +184,7 @@
                                        [options]="options.assessmentTypes"
                                        [(ngModel)]="settings.assessmentType"
                                        [ngModelOptions]="{standalone: true}"
-                                       (change)="onAssessmentTypeChange(settings.assessmentType)"></sb-radio-button-group>
+                                       (change)="onAssessmentTypeChange()"></sb-radio-button-group>
               </div>
             </div>
           </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -184,7 +184,7 @@
                                        [options]="options.assessmentTypes"
                                        [(ngModel)]="settings.assessmentType"
                                        [ngModelOptions]="{standalone: true}"
-                                       (change)="onAssessmentTypeChange()"></sb-radio-button-group>
+                                       (change)="onAssessmentTypeChange(settings.assessmentType)"></sb-radio-button-group>
               </div>
             </div>
           </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -183,7 +183,7 @@
                                        [options]="options.assessmentTypes"
                                        [(ngModel)]="settings.assessmentType"
                                        [ngModelOptions]="{standalone: true}"
-                                       (change)="onIncludeStateResultsChange()"></sb-radio-button-group>
+                                       (change)="onAssessmentTypeChange()"></sb-radio-button-group>
               </div>
             </div>
           </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -122,7 +122,8 @@
                                [(ngModel)]="includeStateResults"
                                [ngModelOptions]="{standalone: true}"
                                (ngModelChange)="onIncludeStateResultsChange()">
-                        <span class="ml-xs">{{'aggregate-report-form.section.organization.include.state-results' | translate}}</span>
+                        <span class="ml-xs"
+                              [ngClass]="{'text-muted': summativeFieldsDisabled}">{{'aggregate-report-form.section.organization.include.state-results' | translate}}</span>
                       </label>
                     </div>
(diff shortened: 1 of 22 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -122,7 +122,8 @@ <h3>{{'aggregate-report-form.section.organization.heading' | translate}}</h3>
                                [(ngModel)]="includeStateResults"
                                [ngModelOptions]="{standalone: true}"
                                (ngModelChange)="onIncludeStateResultsChange()">
-                        <span class="ml-xs">{{'aggregate-report-form.section.organization.include.state-results' | translate}}</span>
+                        <span class="ml-xs"
+                              [ngClass]="{'text-muted': summativeFieldsDisabled}">{{'aggregate-report-form.section.organization.include.state-results' | translate}}</span>
```
esotrope: unrelated: lightens the disabled checkbox label

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -615,7 +616,8 @@ <h3>{{'aggregate-report-form.section.preview-heading' | translate}}</h3>
                                        analyticsCategory="AggregateQueryBuilder"
                                        analyticsEvent="Change"></sb-radio-button-group>
               </span>
-              <span class="pull-right flex-child ml-xs">
+              <span class="pull-right flex-child ml-xs"
+                    *ngIf="currentAssessmentDefinition.typeCode !== 'iab'">
```
esotrope: dont show grouped vs. separate control for IABs
BTW I littered the code with `type === 'iab'` checks for now instead of adding the info to the assessment definition object. I plan on fixing this in a second pass.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -315,6 +318,23 @@ export class AggregateReportFormComponent {
...
 
+  onAssessmentTypeChange(type: string): void {
+
+    // Preserve column order between changing assessment types
+    let order = this.settings.columnOrder.concat();
+    if (type === 'iab') {
+      order.splice(order.indexOf('assessmentGrade') + 1, 0, 'assessmentLabel');
+    } else {
+      order = order.filter(columnId => columnId !== 'assessmentLabel');
+    }
```
esotrope: need some help making this logic less hideous - basically every time you change to and from IAB we need to
1. the ordered columns for that assessment type are present
2. make sure the order is maintained
3. when going from ica/summative to iab make sure to add back the assessmentlabel column in a sensible place.

Writing this made me feel like the table component should not be bleeding this kind of maintenance out but wasn't sure how not to considering the column list is shared with the column ordering component.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
```diff
@@ -138,7 +136,9 @@ export class AggregateReportOptionsMapper {
       economicDisadvantages: options.economicDisadvantages,
       performanceLevelDisplayType: PerformanceLevelDisplayTypes.Separate,
       valueDisplayType: ValueDisplayTypes.Percent,
-      columnOrder: DefaultColumnOrder,
+      columnOrder: options.assessmentTypes[0] === 'iab'
+        ? DefaultColumnOrder.concat()
+        : DefaultColumnOrder.filter(columnId => columnId !== 'assessmentLabel'),
```
esotrope: i think this logic is repeated 4 times in different locations. not ideal. would appreciate some fresh eyes on how to make this cleaner.
ideally i would like to always have the assessmentLabel in the order array and just show/hide it conditionally somehow : /

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -90,8 +94,8 @@
             <performance-level-distribution-chart [percentages]="rowData.performanceLevelByDisplayTypes.Separate.Percent"
                                                   [cutPoint]="table.assessmentDefinition.performanceLevelGroupingCutPoint"
                                                   [assessmentTypeCode]="table.assessmentDefinition.typeCode"
-                                                  [displayType]="performanceLevelDisplayType"
-                                                  [center]="true"></performance-level-distribution-chart>
+                                                  [displayType]="table.assessmentDefinition.typeCode !== 'iab' ? performanceLevelDisplayType : 'Separate'"
+                                                  [center]="table.assessmentDefinition.typeCode !== 'iab'"></performance-level-distribution-chart>
```
esotrope: new logic to make iab display not support "grouped" and to not display with the center divider line

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -15,8 +15,14 @@ import { Table } from "primeng/table";
...
-
-export const DefaultRowsPerPageOptions = [100, 500, 1000];
+export const DefaultRowsPerPageOptions = [ 100, 500, 1000 ];
+export const DefaultColumnOrder: string[] = [
+  'organization',
+  'assessmentGrade',
+  'assessmentLabel',
+  'schoolYear',
+  'dimension'
+];
```
esotrope: figured default column order belonged in the table - before it was in the mapper i think

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -88,17 +98,7 @@ export class AggregateReportTableComponent implements OnInit {
...
-      new Column({id: "organization", field: "organization.name"}),
-      new Column({id: "assessmentGrade", field: "assessmentGradeCode"}),
-      new Column({id: "schoolYear"}),
-      new Column({id: "dimension", field: "dimension.id"}),
-      new Column({id: "studentsTested"}),
-      new Column({id: "achievementComparison", sortable: false}),
-      new Column({id: "avgScaleScore"}),
-      ...this.getPerformanceLevelColumns()
-    ];
-    this.updateColumnOrder();
```
esotrope: moved to when the table model changes because both the assessment label and performance level columns need to be recomputed

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter-options.mapper.ts
```diff
@@ -3,15 +3,12 @@ import { ExamFilterOptions } from "../../model/exam-filter-options.model";
...
-    uiModel.ethnicities = [];
-    apiModel.ethnicities.forEach(ethnicity => uiModel.ethnicities.push(ethnicity));
-
-    return uiModel;
+  mapFromApi(serverOptions: any): ExamFilterOptions {
+    const options: ExamFilterOptions = new ExamFilterOptions();
+    options.schoolYears = serverOptions.schoolYears.concat();
+    options.ethnicities = serverOptions.ethnicities.concat();
+    return options;
   }
```
esotrope: totally unrelated but accidentally opened the file and couldn't help myself

webapp/src/main/webapp/src/styles.less
```diff
@@ -592,7 +592,7 @@ aggregate-report-table .label {
         border-right: 1px solid #fff;
       }
     }
-    &:last-child {
+    &:nth-child(2) {
```
esotrope: performance distribution chart style fix

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
wtritch: Hm... rather than put this logic here, couldn't we just hide the performanceLevel selector and initialize it to "Separate"?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
wtritch: I feel like the default column order should always be all columns, and the query builder is the only place that has the logic to strip out the ones that don't belong...  I think that's the only place where the user can set/change the assessment type of an aggregate report, yes?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
wtritch: I kind of like the idea that the table and ordering components just accept a list of "identity columns" and only display the columns included in the list.  That way the table and ordering component get to be "dumb" and just match the display to the inputs and this becomes the only place that has to know about IAB/ICA/Summative and which columns to display.

agorina: I concur that the ordering should be based on the "identity columns", while displaying a label.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -632,8 +634,11 @@ <h3>{{'aggregate-report-form.section.preview-heading' | translate}}</h3>
                                     [preview]="true"
                                     [table]="previewTable"
                                     [valueDisplayType]="settings.valueDisplayType"
-                                    [performanceLevelDisplayType]="settings.performanceLevelDisplayType"
-                                    [columnOrdering]="settings.columnOrder"></aggregate-report-table>
+                                    [performanceLevelDisplayType]="
+                                      currentAssessmentDefinition.performanceLevelDisplayTypes.includes(settings.performanceLevelDisplayType)
+                                        ? settings.performanceLevelDisplayType
+                                        : currentAssessmentDefinition.performanceLevelDisplayTypes[0]"
```
esotrope: preserves display type selection when switching to and from iab/ica-summatives

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -315,6 +320,23 @@ export class AggregateReportFormComponent {
     this.showAdvancedFilters = !this.showAdvancedFilters;
   }
 
+  onAssessmentTypeChange(type: string): void {
+
+    // Preserve column order between changing assessment types
+    const currentOrder = this.settings.columnOrder.concat();
+    if (!currentOrder.includes('assessmentLabel')) {
+      currentOrder.splice(currentOrder.indexOf('assessmentGrade') + 1, 0, 'assessmentLabel');
+    }
```
esotrope: there is definitely a generic way to code this but it gets even more complicated.
ideally we dont lose the sort order of the assessment label when switching between assessment types but that is the way it works right now

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
```diff
@@ -120,33 +121,39 @@ export class AggregateReportOptionsMapper {
...
+          subjects: options.subjects,
+          summativeAdministrationConditions: [ options.summativeAdministrationConditions[ 0 ] ],
+          migrantStatuses: options.migrantStatuses,
+          individualEducationPlans: options.individualEducationPlans,
+          section504s: options.section504s,
+          limitedEnglishProficiencies: options.limitedEnglishProficiencies,
+          economicDisadvantages: options.economicDisadvantages,
+          performanceLevelDisplayType: assessmentDefinition.performanceLevelDisplayTypes[ 0 ],
+          valueDisplayType: ValueDisplayTypes.Percent,
+          columnOrder: assessmentDefinition.aggregateReportIdentityColumns,
```
esotrope: we now get default performance level display type and identity columns from the assessment definition

wtritch: hm... I wonder if we're going to want to break up the "assessment definition" concept by vertical...  maybe not.

esotrope: hm. i think ideally we would. but let's start small : j

PR #1103, opened 2018-03-19
Migrate the user's groups table to the new table implementation

wtritch:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/group.service.ts
+++ b/webapp/src/main/webapp/src/app/groups/group.service.ts
@@ -20,7 +20,7 @@ export class GroupService {
         group.name = serverGroup.name;
         group.schoolName = serverGroup.schoolName;
         group.schoolId = serverGroup.schoolId;
-        group.subjectCode = serverGroup.subjectCode;
+        group.subjectCode = serverGroup.subjectCode || 'ALL';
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/groups/group.service.ts
```diff
@@ -20,7 +20,7 @@ export class GroupService {
         group.name = serverGroup.name;
         group.schoolName = serverGroup.schoolName;
         group.schoolId = serverGroup.schoolId;
-        group.subjectCode = serverGroup.subjectCode;
+        group.subjectCode = serverGroup.subjectCode || 'ALL';
```
wtritch: I feel like the mapper should be applying the defaults rather than the individual UI's.  Thoughts?

esotrope: yes. however for this one since we arent mapping a request that was sent i think we will accept an array of subjects eventually right?

wtritch: I'm not sure... this property directly maps to the `reporting.student_group.subject_id tinyint(4)` column value.  If we want a group to be able to have 0..n subjects assigned, we have some work to do on a few levels.

PR #1104, opened 2018-03-19
work-in-progress for implementing a school typeahead

dsnyder-fairway:
I do not plan on merging this. Just up for early feedback

I still need to 

1. Implement this for `school-results.component.ts`
2. Implement a `sb-typeahead` for aggregate reports when we are below the `toggleLimit`
2. Move `toggleLimit` somewhere common to be used by `school-results`, `school-grade` and Aggregate reports


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -70,9 +70,11 @@
             switch (organization.getOrganizationType()) {
                 case District:
                     districts.add(organization);
+                    if (districts.size() > 1) return;
```
dsnyder-fairway: just avoid looping since this method in the end just checks if the size of these arrays are 1

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/OrganizationRepository.java
```diff
@@ -22,6 +22,15 @@
...
     Collection<Organization> findAll(PermissionScope permissionScope, OrganizationQuery query);
 
+    /**
+     * Gets the number of schools matching the query
+     *
+     * @param permissionScope   A permission scope
+     * @param query             The organization query
+     * @return The number of schools matching the query
+     */
+    long getSchoolCount(PermissionScope permissionScope, OrganizationQuery query);
```
dsnyder-fairway: method is used from our front-end to determine if we need to have a `school-typeahead` or an `sb-typeahead`

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
esotrope: I'm confused by this. this is in the olap package - why do we need to change this?

dsnyder-fairway: we don't need to. I stumbled across this early when working on this ticket. This just short-circuits the loop

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/OrganizationRepository.java
esotrope: hm. seems suboptimal. wouldn't a `findSchoolsIfCountUnder(..., count)`
that way we don't check the count and then query again to get the schools

common/src/main/resources/common.sql.yml
```diff
@@ -8,6 +8,13 @@ sql:
 
   organization:
     reporting:
+      findSchoolCount: >-
+        SELECT COUNT(*)
+        FROM school
+        WHERE
+          (1=:statewide OR school.district_group_id IN (:district_group_ids) OR school.district_id IN (:district_ids) OR school.school_group_id IN (:school_group_ids) OR school.id IN (:school_ids))
```
esotrope: we should use the security snippet for this

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/OrganizationRepository.java
dsnyder-fairway: sounds good

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/JdbcOrganizationRepository.java
```diff
@@ -61,6 +64,17 @@
         );
     }
 
+    @Override
+    public long getSchoolCount(final PermissionScope permissionScope, final OrganizationQuery query) {
```
esotrope: i thought all repo methods accepted `PermissionSource`s now. To remain flexible we probably want to have this accept it.

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -83,7 +83,7 @@ export class AssessmentExamMapper {
     uiModel.schoolYear = apiModel.schoolYear;
     uiModel.transfer = apiModel.transfer;
 
-    let school: School = new School();
+    let school: DefaultSchool = new DefaultSchool();
```
esotrope: dont we need the district name in the model? that's why we use `School` here

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/JdbcOrganizationRepository.java
dsnyder-fairway: 👍 I'll update this. Note that there are a bunch of repo's still accepting `PermissionScope`

webapp/src/main/webapp/src/app/school-grade/school-grade.component.ts
```diff
@@ -48,6 +74,14 @@ export class SchoolGradeComponent {
     this.school = value;
   }
 
+  unSelectSchool(value: any) {
```
esotrope: `deselectSchool`

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
dsnyder-fairway: I moved away from `school-grade/school` and moved this to `https://github.com/SmarterApp/RDW_Reporting/pull/1104/files#diff-6e0cd7d7f31fe9f38318efa22ce7aab6`. There were many components using `school-grade/school` which seemed unnecessary (such as https://github.com/SmarterApp/RDW_Reporting/pull/1104/files/47faade20aecddf00b4eff9eb46e720c18f3abdc#diff-9cf2f6e4085e67106f8a6a7bf21dc0c9)

webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
```diff
@@ -31,26 +31,29 @@ export class CachingDataService {
 
   public get(url: string, options?: RequestOptionsArgs): Observable<any> {
 
-    const response = this.responsesByUrl.get(url);
+    const cacheKey = url + '|' + JSON.stringify(options);
```
esotrope: the keys of these options will need to be sorted for this to work reliably.
I would break this key creation into a method.
```typescript
createKey(url, options?) {
  return url + Object.entries(options || {})
    .sort(([a], [b]) => a.localeCompare(b))
    .map(([key, value]) => key + '=' + value)
    .join('')
}
```

webapp/src/main/webapp/src/app/shared/organization/organization.ts
```diff
@@ -86,6 +87,7 @@ export class DefaultSchool extends AbstractIdentifiableOrganization implements S
   schoolGroupId: number;
   districtId: number;
   districtGroupId: number;
+  districtName: string;
```
esotrope: not sure i'm a fan of this exactly. feels like a concern specific to views that want the district name associated with a school but that is theory i guess not what is necessarily practical in practice.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/JdbcOrganizationRepository.java
dsnyder-fairway: `User` implements `PermissionSource`. We still get the `PermissionScope` from the `User`

webapp/src/main/webapp/src/app/shared/organization/school-typeahead.ts
```diff
@@ -0,0 +1,58 @@
...
+           [(ngModel)]="value"
+           [disabled]="disabled"
+           [typeahead]="options"
+           [typeaheadItemTemplate]="schoolTemplate"
+           (typeaheadLoading)="loading = $event"
+           (typeaheadNoResults)="noResults = $event"
+           (typeaheadOnSelect)="onTypeaheadSelectInternal($event)"
+           [typeaheadMinLength]="3"
+           typeaheadOptionField="name"
+           typeaheadWaitMs="300"
```
esotrope: later we'll want this to be 0 if all the schools are known already by the client

dsnyder-fairway: would that be after we have cached all the possible schools for a client?

esotrope: nope i was thinking if we do the fancy thing and do getSchoolsIfUnder(count) and if that has some values we dont need server side lookup and therefore don't need needless debounce

webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
dsnyder-fairway: 🙇 Thank you!

webapp/src/main/webapp/src/app/shared/organization/school-typeahead.ts
dsnyder-fairway: hmm, but if the schools are under the count, then shouldn't we use the `sb-typeahead` and not the `school-typeahead`?

webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
esotrope: Actually we'll need to look at the RequestOptionsArgs interface too. i think it might have a "param" field in it that we might need to reach into like `createKey(url, options ? options.params || {} : {})`

webapp/src/main/webapp/src/app/shared/organization/school-typeahead.ts
dsnyder-fairway: I suppose I could include that sort of stuff in this component

esotrope: ideally we have one component. sb-typeahead is only used by the school form and results page right now

webapp/src/main/webapp/src/app/shared/organization/organization.ts
dsnyder-fairway: I suppose I could create some other ts file in this directory to implement the `school-grade/school`. That way we aren't messing with these classes and also creating a common school to be used by everything including a districtName

common/src/main/resources/common.sql.yml
agorina: Do we need to count? Are we displaying the count somewhere?

PR #1105, opened 2018-03-19
Task/last datatable cleanup

wtritch:
This PR cleans up the datatable style references and ensures that every table has at least a default empty message (as the datatable did.)


webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.html
```diff
@@ -152,6 +160,14 @@ <h4 *ngIf="isWritingTraitItem" class="pt-md">{{'writing-trait-scores.aggregate-h
           </td>
         </tr>
       </ng-template>
+
+      <ng-template pTemplate="emptymessage" let-columns>
```
esotrope: i feel like we need a better way to display no results than a table with a row that says "no results" but that is just my pref.

wtritch: I agree, but right now I'm just mirroring the default DataTable behavior that we used to rely upon.  Ideally we'd have specific "no results" messages for each table.

PR #1110, opened 2018-03-20
implemented a typeahead when number of schools for a user is above a limit

dsnyder-fairway:
no work has been done on the aggregate report pages to make this have a similar behavior. We might want to, but I am unsure if it is required


webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -33,15 +33,6 @@ export class SBTypeahead implements OnInit {
   @Output()
```
dsnyder-fairway: I am not yet completely sure how to consolidate the `sb-typeahead` with the `school-typeahead`. I will continue working on this part I suppose...

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SchoolsWrapper.java
```diff
@@ -0,0 +1,49 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+public class SchoolsWrapper {
```
agorina: Could you please clarify what this is used for? Is this for paging through schools, or getting the top X schools?

dsnyder-fairway: this is used for getting the schools, if the number of schools is within a query limit. If we ask for 10 schools, but we see that there are 100, we say it hasMoreSchools, but the school set is empty

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/JdbcOrganizationRepository.java
```diff
@@ -61,6 +68,23 @@
         );
     }
 
+    @Override
+    public SchoolsWrapper findSchoolsWithinLimit(final PermissionScope permissionScope, final OrganizationQuery query) {
+        final long count = template.queryForObject(
```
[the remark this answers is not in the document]
dsnyder-fairway: We need the count to see if there are more schools than the limit we are trying to query for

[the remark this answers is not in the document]
agorina: If you ask for `limit+1` and the `limit+1` is returned, then you know that there is more than the `limit` without counting.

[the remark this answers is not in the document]
dsnyder-fairway: ah yeah, good point. Yes I'll work this in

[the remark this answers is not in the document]
dsnyder-fairway: done

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SchoolsWrapper.java
wtritch: This feels like re-inventing  the wheel.  I'd recommend returning a `org.springframework.data.domain.Slice<School>` instead.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
```diff
@@ -85,15 +78,56 @@ public DefaultOrganizationService(@NotNull final OrganizationRepository reposito
...
         final Collection<Organization> organizations = repository.findAll(permission.getScope(), districtGroupQuery);
         return ImmutableSet.copyOf(organizations.stream()
                 .map(DistrictGroup.class::cast)
                 .collect(Collectors.toSet()));
     }
 
+    @Override
+    public SchoolsWrapper getSchoolsWithinLimit(@NotNull final User user, final OrganizationQuery query) {
+
+        final OrganizationQuery schoolQuery = buildOrganizationQuery(OrganizationType.School, query.getLimit() + 1);
```
wtritch: This ignores most of the OrganizationQuery parameters.  Instead, I'd do something like:
```
final OrganizationQuery schoolQuery = OrganizationQuery.builder()
  .copy(query)
  .limit(query.getLimit() + 1)
  .build();
```
And change this method signature to:
```
public Slice<School> getSchools(final User user, final OrganizationQuery query)
```
So that it can be used for general school queries as well.  Although as soon as we start using paging/slicing I feel like we should start using the built-in Spring Pageable query parsing... but that's likely out of scope.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
```diff
@@ -45,4 +54,15 @@ public OrganizationController(@NotNull final OrganizationService service) {
         return service.getDistrictGroups(user);
     }
 
+    @GetMapping("/schoolsWrapper")
```
wtritch: I would much rather have a single "/schools" endpoint that accepts an optional OrganizationQuery and returns a Slice<School>.  I'd recommend rolling this endpoint into the "/schools" endpoint, and if the user provides an empty OrganizationQuery then return all the user's schools (to preserve previous functionality.)

dsnyder-fairway:
After speaking with @wtritch I will be leveraging an existing endpoint

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationService.java
```diff
@@ -50,4 +64,22 @@
...
     Set<DistrictGroup> getDistrictGroups(@NotNull User user);
 
+    /**
+     * Gets the schools the user has access to if the number of schools is less than or equal tp the limit
+     *
+     * @param user the user accessing the schools
+     * @param query the query containing the limit
+     * @return the schools if the number of schools are less or equal to the limit
+     */
+    SchoolsWrapper getSchoolsWithinLimit(@NotNull User user, OrganizationQuery query);
```
esotrope: because this method accepts a query and not just a limit, should it just be named getSchools?

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -110,34 +122,73 @@ export class SchoolResultsComponent implements OnInit {
...
+
+          this.filterOptions = filterOptions;
+          this.currentSchoolYear = Number.parseInt(schoolYear) || this.filterOptions.schoolYears[ 0 ];
+
+          this.availableGrades = grades;
+          this.gradesAreUnavailable = this.availableGrades.length == 0;
+          this.currentGrade = this.availableGrades.find(grade => grade.id === gradeIdParam);
+        });
+      } else {
+        this.aboveLimit = true;
```
esotrope: why not just reference the shoolsWrapper/slice variable?

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -110,34 +122,73 @@ export class SchoolResultsComponent implements OnInit {
...
+          this.schoolService.findGradesWithAssessmentsForSchool(schoolIdParam)
+        ).subscribe(([ school, filterOptions, grades ]) => {
+
+          this.schoolOptions = Observable.create(observer => {
+            observer.next(this.schoolTypeahead.value);
+          }).pipe(
+            mergeMap(
+              (search: string) =>
+                this.organizationService.searchSchoolsWithDistrictsBySchoolName(search)
+                  .map(
```
esotrope: should be `.pipe(map())` here i think

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -110,34 +122,73 @@ export class SchoolResultsComponent implements OnInit {
...
+            ));
+          this.currentSchool = school;
+          this.schoolIsAvailable = this.currentSchool !== undefined;
+
+          this.filterOptions = filterOptions;
+          this.currentSchoolYear = Number.parseInt(schoolYear) || this.filterOptions.schoolYears[ 0 ];
+
+          this.availableGrades = grades;
+          this.gradesAreUnavailable = this.availableGrades.length == 0;
+          this.currentGrade = this.availableGrades.find(grade => grade.id === gradeIdParam);
```
esotrope: this is all dup code right? if so let's consolidate

webapp/src/main/webapp/src/app/school-grade/school-grade.component.html
```diff
@@ -10,10 +10,24 @@
...
+              <span [hidden]="!schoolTypeahead.noResults" class="small gray-darker">{{'school-typeahead.no-matches' | translate}}</span>
+            </div>
+          </div>
           <sb-typeahead #schoolTypeahead
+                        *ngIf="!aboveLimit"
                         inputId="search-school-name"
                         [options]="schoolOptions"
                         (selected)="schoolChanged($event)"
+                        (input)="deselectSchool($event)"
                         placeholder="{{'common.action.select' | translate}}"></sb-typeahead>
```
esotrope: is the one school preselect case handled here?

dsnyder-fairway: no

PR #1111, opened 2018-03-20
Convert the aggregate report back-filling to key-based removal.

wtritch:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportRowIdentity.java
```diff
@@ -0,0 +1,32 @@
+package org.opentestsystem.rdw.olap.model;
+
+import org.opentestsystem.rdw.reporting.common.model.ReportRow;
+
+import java.util.Objects;
+
+/**
+ * This class identifies a {@link org.opentestsystem.rdw.reporting.common.model.ReportRow}
+ */
+public class ReportRowIdentity {
```
wtritch: Since I need the same key in both the sql-builder and jdbc layer I decided on an equivalency-identifier class rather than some kind of key-generating static utility method.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportRowIdentity.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        final ReportRowIdentity that = (ReportRowIdentity) o;
+        return Objects.equals(reportRow.getOrganization(), that.reportRow.getOrganization()) &&
+                Objects.equals(reportRow.getAssessment(), that.reportRow.getAssessment()) &&
+                Objects.equals(reportRow.getDimension(), that.reportRow.getDimension());
+    }
```
wtritch: It tests for equivalency using every property of the ReportRow other than `measures`

malaffoon: This comment should be in the class javadocs i think.

wtritch: good point, done

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/QueryWithReportTemplate.java
```diff
@@ -32,8 +33,8 @@ public MapSqlParameterSource getParameterSource() {
         return parameterSource;
     }
 
-    public Set<ReportRow> getReportTemplate() {
-        return newHashSet(reportTemplate);
+    public Map<ReportRowIdentity, ReportRow> getReportTemplate() {
```
malaffoon: Hmm. Feels messy. It seems like this should return the Set (or even an iterator) and it is the caller that cares about the mapping equivalence. OTOH, this exists only to serve the repo.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/QueryWithReportTemplate.java
```diff
@@ -32,8 +33,8 @@ public MapSqlParameterSource getParameterSource() {
...
+        return newHashMap(reportTemplate);
```
wtritch: This concerns me.  I know we like our models to be immutable, but do we really need to create a copy of the back-filled map here?

malaffoon: Yeah, that is part of my messy comment.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/QueryWithReportTemplate.java
wtritch: hmm... I kind of like returning a Set and letting the consumer do with it whatever they will...

wtritch: done

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportRowIdentity.java
agorina: Why not just compare ids of the Assessment and Organization?

wtritch: Because Assessment and Organization already have equality/hashcode methods.  It feels cleaner to use their built-in equality/hashcode methods rather than hand-jam equals tests...

agorina: chatted offline, good for now.

PR #1112, opened 2018-03-20
Lower memory aggregate report backfilling

wtritch:
If we want even lower memory usage of the aggregate report processing we can truly remove and free up the back-filled rows for GC as we process the result set.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/QueryWithReportTemplate.java
```diff
@@ -32,7 +33,7 @@ public MapSqlParameterSource getParameterSource() {
         return parameterSource;
     }
 
-    public Set<ReportRow> getReportTemplate() {
+    public Map<ReportRowIdentity, ReportRow> getReportTemplate() {
         return reportTemplate;
     }
```
wtritch: This will return a reference to the actual map instance, so as we remove ReportRows from the back-filled template they can be GC'd... at the cost of a mutable model.

wtritch:
My thought is that we may have 100k back-filled rows... seems silly to keep them in memory as *real* rows are returned...

PR #1115, opened 2018-03-21
Feature/dwr 1532

dsnyder-fairway:
went ahead and am using /schools to either retrieve the list of schools, or determine if the schools exceeds the limit of my query


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
```diff
@@ -64,9 +64,11 @@ public OrganizationQuery build() {
         }
 
         public Builder copy(final OrganizationQuery query) {
-            limit(query.getLimit());
-            types(query.getTypes());
-            name(query.getName());
+            if (query != null) {
```
dsnyder-fairway: to avoid NPEs

dsnyder-fairway:
just noticed I still need to update documentation

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
```diff
@@ -44,8 +44,8 @@ public DefaultOrganizationService(@NotNull final OrganizationRepository reposito
         }
 
         final OrganizationQuery schoolQuery = OrganizationQuery.builder()
+                .copy(query)
                 .types(EnumSet.of(OrganizationType.School))
-                .limit(Unlimited)
```
dsnyder-fairway: just realized that if we haven't passed the limit on calls to the endpoint that we are not going to have a limit set to anything except 1000. Hmm. This might be fine, as this endpoint is used locally for only `school-grade` and `school-results` where we now query with a limit of 11...thoughts?

esotrope: i would think that default limit should be no limit

agorina: Hmmm...an unbound list in this case is not such a good idea. I think the default should have a limit, maybe a high one, but still a limit.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
```diff
@@ -26,8 +29,13 @@ public OrganizationController(@NotNull final OrganizationService service) {
...
     @GetMapping("/schools")
-    public Set<School> getSchools(@AuthenticationPrincipal final User user) {
-        return service.getSchools(user);
+    public Set<School> getSchools(@AuthenticationPrincipal final User user, final OrganizationQuery query) {
+        return service.getSchools(user, query);
+    }
+
+    @GetMapping("/school/{schoolId}")
+    public School getSchool(@AuthenticationPrincipal final User user, @PathVariable final int schoolId) {
+        return service.getSchool(user, schoolId);
```
esotrope: shouldn't this be `/schools/{schoolId}`? also i just realized this is in the org controller. do we really need a school-specific endpoint here? or can it be generic? if it needs to be specific i might break this out into a separate controller for schools

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
dsnyder-fairway: correction - `OrganizationQueryArgumentResolver` sets the default limit to 20. `OrganizationQuery` has a default of 100.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
dsnyder-fairway: I'll change this to `/schools/{schoolId}`. `OrganizationController` appears to have originally not pertained to to Organizations. Line 57 shows the only response with an Organization.

webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
```diff
@@ -31,26 +32,36 @@ export class CachingDataService {
...
       share()
     );
 
-    this.requestsByUrl.set(url, observable);
+    this.requestsByUrl.set(cacheKey, observable);
     return observable;
   }
 
+  private createKey(url: string, options?: RequestOptionsArgs): string {
+    return url + '?' + Object.entries(options || {})
```
esotrope: looks like RequestOptionsArgs has field `params` or `search` that we need to be using.
Maybe change the method to:
```typescript
if (options == null) {
  return url;
}
const parameters = options.params || options.search || {};
return url + '?' + Object.entries(parameters)
  //...
```

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
dsnyder-fairway: I can break this out into a separate `SchoolsController` though

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
dsnyder-fairway: I'm thinking it makes more sense to leave this up to the caller. We could validate that they have set a limit instead of defaulting to a number

wtritch: I'd rather have a default than demand that callers provide a limit, personally.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
```diff
@@ -34,7 +34,7 @@ public DefaultOrganizationService(@NotNull final OrganizationRepository reposito
     }
 
     @Override
-    public Set<School> getSchools(@NotNull final User user) {
+    public Set<School> getSchools(@NotNull final User user, final OrganizationQuery query) {
```
wtritch: Hmm... the OrganizationQuery has a "type" field.  Why doesn't the controller just copy the OrganizationQuery, hard-code the "type" to School and pass it on to the service.getOrganizations(...) endpoint?

dsnyder-fairway: it sorta does. `service.getSchools` does essentially that. Difference is we do not have a preauth in `service.getSchools`
``` 
        // This method is not PreAuthorized because it is pulled at login time.  Need to check
        // permissions here.
```

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
dsnyder-fairway: where do we document the default limit? Has it been 20 or 1000?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
```diff
@@ -70,9 +70,11 @@
             switch (organization.getOrganizationType()) {
                 case District:
                     districts.add(organization);
+                    if (districts.size() > 1) return;
                     break;
                 case School:
                     schools.add(organization);
+                    if (schools.size() > 1) return;
```
malaffoon: Is this really doing what you think? I'm pretty sure it is going to continue the forEach() processing and this change makes no difference to the behavior.

dsnyder-fairway: hmmm maybe not. I can revert this. Any way out of this? Just avoiding some O(n)

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
dsnyder-fairway: see discussion here https://github.com/SmarterApp/RDW_Reporting/pull/1115#discussion_r176243966

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
malaffoon: Streams aren't great when you want to impose side-effects. To fix it we'd want a different helper that returns an iterator/collection, pseudo-code:
```
    for (final Organization org : getUserOrganizations(permissionScope)) {
        switch (org.getOrganizationType()) {
            ...
        }
        if (districts.size() + schools.size() > 1) return Optional.empty();
    }
```

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
wtritch: I believe the OrganizationQueryArgumentResolver is setting the true default at 20 since all communication goes through the resolver.
I'd be wary of changing it since it's used by other systems, but we could change our front-end requests to provide a limit of 1000 or so

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/OrganizationController.java
dsnyder-fairway: done

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserOrganizationService.java
dsnyder-fairway: reverted the changes

webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
dsnyder-fairway: done

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/OrganizationQuery.java
dsnyder-fairway: reverted these changes

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
dsnyder-fairway: I believe this is fine then as we were not specifying a limit on front end requests before (defaulting to 20). Disregard my original comment https://github.com/SmarterApp/RDW_Reporting/pull/1115#discussion_r176243966

dsnyder-fairway:
@esotrope @wtritch @agorina @malaffoon is this okay to merge?

PR #1119, opened 2018-03-21
Add missing collapse/expand labels to student history table.

wtritch:


webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -55,11 +55,12 @@ <h3>{{'student-results.title' | translate}}</h3>
               [ngClass]="getAssessmentTypeColor(assessmentType)">
           <span class="label">{{'common.assessment-type.' + assessmentType + '.short-name' | translate}}</span><span class="label">{{'common.subject.' + subject + '.short-name' | translate}}</span>
         </span>
-        <button class="btn btn-default btn-sm icon-only pull-right"
+        <button class="btn btn-default pull-right"
```
esotrope: btn-sm didn't look good?

wtritch: I like it better.  But I like being consistent more =)  And the rest of the app doesn't use btn-sm for expand/collapse/show/hide buttons.

PR #1120, opened 2018-03-22
DWR-1627 Refactoring custom aggregate service/repository(step 1)

agorina:
Introduce a different model for the repository.

@wtritch, my hope is that with this change we should be able to work in parallel on the Reporting side changes for the custom analysis as well as refactoring. Please let me know what you think.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
@@ -1,0 +1,313 @@
+package org.opentestsystem.rdw.olap.model;
+
+import org.opentestsystem.rdw.reporting.common.model.DimensionType;
+
+import java.util.List;
+import java.util.Set;
+
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * Represents a query for the custom aggregate report
+ */
+public class AggregateReportQuery {
+    private String assessmentTypeCode;
+    private Set<String> subjectCodes;
+    private Set<Integer> schoolYears;
+    private Set<String> assessmentGradeCodes;
+    private boolean includeState;
+    private boolean includeAllDistricts;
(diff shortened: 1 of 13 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
```diff
@@ -0,0 +1,310 @@
...
+import org.opentestsystem.rdw.reporting.common.model.DimensionType;
+
+import java.util.Set;
+
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * Represents a query for the custom aggregate report
+ */
+public class AggregateReportQuery {
```
agorina: The idea is to have two different models: one for `UI-to-API-to-Service` communication and another for `Service-to-Repository`. This is a new one that is meant for the latter.  For now it is pretty much the copy of another one, but it will change with future changes.

esotrope: can we rename `ReportQuery` to `AggregateReportRequest` along with this change?

agorina: ReportQuery is persisted as part of the user report. I did not want to deal with the json serialization as part of this PR.

wtritch: So this is just introducing the api-to-repository conversion and a repository query object?  Seems reasonable to me, but if we keep trying to make iterative, non-breaking changes into 'develop' this is going to take a while.  I'm still wondering if we just want to coordinate on a long-running feature branch that we can break for a while and validate behavior with tests...

agorina: Maybe, I am not sure. I am still trying to see how much of incremental change I can do for the repository refactoring.

PR #1121, opened 2018-03-22
corrected assessment grades

dsnyder-fairway:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -220,9 +220,9 @@
       "column-order-info": "Change your result column order to easily compare your results broken down in different ways."
     },
     "assessment-grade": {
-      "10": "10",
-      "11": "11",
-      "12": "12",
+      "10": "High School",
```
jtreuting: should be HS instead of the full name so it fits like the numbers

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -232,7 +232,7 @@
       "06": "6",
       "07": "7",
       "08": "8",
-      "09": "9",
+      "09": "High School",
```
esotrope: I think `HS` is what we want here

dsnyder-fairway:
@jtreuting done

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -220,9 +220,9 @@
       "column-order-info": "Change your result column order to easily compare your results broken down in different ways."
     },
     "assessment-grade": {
-      "10": "10",
-      "11": "11",
-      "12": "12",
+      "10": "HS",
+      "11": "HS",
+      "12": "HS",
       "13": "13",
```
malaffoon: I guess the non-11 ones don't matter but 13 would be HS too.

dsnyder-fairway: Hmm, looks like we've been mislabelling this for a while https://github.com/SmarterApp/RDW_Reporting/blob/3fd4c1dcf82eb37565be1877f8bc7dcc8b8b271a/webapp/src/main/webapp/src/assets/i18n/en.json#L482

esotrope: there are HS with a 13th grade?

malaffoon: Ontario (Canada), Oregon, and a few other places have 5-year high schools.

esotrope: 🌈 ⭐️ learned something new today : j

PR #1122, opened 2018-03-22
Defect/aggregate report form breaks on auto select of single org

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -403,7 +404,6 @@ export class AggregateReportFormComponent {
       this.settings.schools.push(<School>organization);
       this.settings.schools.sort(OrganizationComparator);
     }
-    this.markOrganizationsControlTouched();
```
esotrope: this should only be called when when initiated through user input (it was being called before the form object was initialized)

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultUserReportOptionService.java
```diff
@@ -34,7 +35,6 @@
         this.userOrganizationService = userOrganizationService;
     }
 
-    @Cacheable(AggregateServiceCache.ReportOptions)
```
esotrope: @agorina @wtritch so i thought we fixed this but we never did guess?
This means we are caching report options at the user-level in production : /

agorina: I am confused: why wouldn't we want this per user? Some may or may not have access to certain assessment types?

esotrope: i think we were avoiding caching user-specific data in general. Also, i think this might be a security risk, `User` doesn't implement hashCode/equals based on `username` which means this is caching by object reference which can be recycled potentially 8 /

wtritch: Although, side-note, since we're extending org.springframework.security.core.userdetails.User we *do* hashcode/equals on username =)

esotrope: ah good that's a relief ( : J

PR #1123, opened 2018-03-22
changes order of assessment types on student test history and aggrega…

esotrope:
…te report form to sum, ica, iab

image

image


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
```diff
@@ -11,6 +11,7 @@ import { AggregateServiceRoute } from '../shared/service-route';
 const ServiceRoute = AggregateServiceRoute;
 
 // Used to hotfix natural order of completeness and strict booleans not being in "affirmative-first" order
+const assessmentTypeComparator = ordering(ranking([ 'sum', 'ica', 'iab' ])).compare;
```
wtritch: ...I wonder if we should globalize this and use the same order on the student history page...

esotrope: sounds good. i'll put in /shared/ordering dir

esotrope:
going to add central ordering in a followup

PR #1124, opened 2018-03-22
removes test against sort order

esotrope:


webapp/src/main/webapp/src/app/student/results/student-results.component.spec.ts
```diff
@@ -90,11 +90,6 @@ describe('StudentResultsComponent', () => {
     expect(component).toBeTruthy();
   });
 
-  it('should retrieve sorted assessment types', () => {
-    expect(component.assessmentTypes)
-      .toEqual([ 'iab', 'ica', 'sum' ]);
-  });
```
esotrope: i dont feel this test is legit as ideally the order should be backend driven

PR #1125, opened 2018-03-22
Translate dates and numbers that are displayed via the Angular DatePi…

wtritch:
…pe and DecimalPipe using the currently-selected locale.

Update displayed values on locale change.

image


webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -106,4 +108,13 @@ export class AppComponent {
...
   }
 
+  /**
+   * Register locales available to the angular system for
+   * date, number, currency, etc translations.
+   * NOTE: We currently only embed "en" and "es"
+   */
+  private registerLocales(): void {
+    registerLocaleData(localeEs);
+  }
```
wtritch: This embeds "es" as an available locale for the Angular pipes/localization.

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -15,7 +15,7 @@
               </td>
               <td>
                 <div class="scale-score font-large text-left">
-                  <span class="score label label-primary">{{ statistics.average | number:'1.0-0' | removeComma }}</span>
+                  <span class="score label label-primary">{{ averageScore }}</span>
```
wtritch: This threw me for a bit of a loop.  I recall something about it, but it's a little fuzzy.  We were sending the average through the number pipe to round it, then piping it through a "remove comma" pipe to convert `2,345` -> `2345` which felt a little hokey.  I replaced it with a simple Math.round call in the back-end.  If we don't want to display a localized number with grouping we probably shouldn't be piping it through the number display pipe.

jtreuting: won't this now show the score with the commas back in it?

webapp/src/main/webapp/src/app/shared/i18n/rdw-i18n.module.ts
```diff
@@ -25,11 +29,16 @@ import { HttpClientModule } from "@angular/common/http";
...
+    LanguageSelect,
+    TranslateDatePipe,
+    TranslateNumberPipe
   ],
   providers: [
     RdwTranslateLoader,
-    LanguageStore
+    LanguageStore,
+    TranslateDatePipe,
+    TranslateNumberPipe
```
wtritch: This registers and exports our own handlers for `number` and `date` pipes.

webapp/src/main/webapp/src/app/shared/i18n/translate-date.pipe.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * This date pipe proxies requests to the Angular DatePipe
+ * using the currently-selected locale.
+ */
+@Injectable()
+@Pipe({
+  name: 'date',
+  pure: false //Required to update value on translation change
+})
```
wtritch: The magic here is that this is an IMPURE pipe, which means we get one stateful instance per invocation that is queried on every digest.  The ng-translate TranslatePipe works on the same principal but with some additional Observable handling since translations are asynchronous.  This is mostly a stripped-down copy of their implementation.

There are some fast-fails so that we just keep returning the same `dateDisplay` value if the inputs haven't changed and a listener bound to the TranslateService that will update the value if the current locale changes.

webapp/src/main/webapp/src/app/shared/i18n/rdw-i18n.module.ts
esotrope: cant help but feel like ngx-translate should have these plugins written already for us 🤔

webapp/src/main/webapp/src/app/shared/i18n/translate-date.pipe.ts
```diff
@@ -0,0 +1,80 @@
...
+  private lastDate: any;
+  private lastPattern: string;
+
+  private onLangChange: EventEmitter<LangChangeEvent>;
+
+  constructor(private translate: TranslateService) {
+  }
+
+  public transform(date: any, pattern: string = 'mediumDate'): any {
+    if(!date) {
```
esotrope: nit: format

webapp/src/main/webapp/src/app/shared/i18n/translate-date.pipe.ts
```diff
@@ -0,0 +1,80 @@
...
+  name: 'date',
+  pure: false //Required to update value on translation change
+})
+export class TranslateDatePipe implements PipeTransform, OnDestroy {
+
+  private dateDisplay: string = '';
+  private lastDate: any;
+  private lastPattern: string;
+
+  private onLangChange: EventEmitter<LangChangeEvent>;
```
esotrope: shouldn't this be type `Subscription`?

webapp/src/main/webapp/src/app/shared/i18n/translate-date.pipe.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.updateValue(date, pattern);
+
+    // if there is a subscription to onLangChange, clean it
+    this.dispose();
+
+    // subscribe to onLangChange event, in case the language changes
+    if (!this.onLangChange) {
+      this.onLangChange = this.translate.onLangChange.subscribe(() => {
+        if (this.lastDate) {
+          this.updateValue(this.lastDate, this.lastPattern);
```
esotrope: this throws me a little. i think we might want to call it 'currentDate' and 'currentPattern'

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
wtritch: nope, check below.  This is using a component variable named averageScore which is the rounded statistics.average value.

webapp/src/main/webapp/src/app/shared/i18n/translate-date.pipe.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  public ngOnDestroy(): void {
+    this.dispose();
+  }
+
+  private updateValue(date: any, pattern: string): void {
+    this.lastDate = date;
+    this.lastPattern = pattern;
+
+    let ngPipe: DatePipe;
```
esotrope: why declare this at all?

webapp/src/main/webapp/src/app/shared/remove-comma.pipe.ts
```diff
@@ -1,12 +0,0 @@
-import {Pipe, PipeTransform} from "@angular/core";
-
-@Pipe({name: 'removeComma'})
-export class RemoveCommaPipe implements PipeTransform {
```
esotrope: silly us : j

webapp/src/main/webapp/src/app/shared/i18n/rdw-i18n.module.ts
wtritch: You would think: https://github.com/ngx-translate/core/issues/409

webapp/src/main/webapp/src/app/shared/i18n/translate-date.pipe.ts
wtritch: doh, fixed.

webapp/src/main/webapp/src/app/shared/i18n/rdw-i18n.module.ts
esotrope: There have been many times in this project where i am tempted to submit PR to both ngx-translate and ngx-bootstrap

PR #1127, opened 2018-03-23
Defect/disabled manner of administration

esotrope:
Was extracting common sorting and then i found the cause of the disabled manner of admin advanced filter and then one thing led to another...


webapp/src/main/webapp/src/app/student/results/student-results-filter.component.html
```diff
@@ -9,8 +9,8 @@
                 [(ngModel)]="filterState.schoolYear"
                 (change)="onFilterChange('Year')"
                 class="form-control">
-          <option selected value="0">{{ 'common.buttons.all' | translate }}</option>
-          <option *ngFor="let year of filterState.years" [ngValue]="year">{{year | schoolYear}}</option>
+          <option selected [ngValue]="undefined">{{ 'common.buttons.all' | translate }}</option>
+          <option *ngFor="let year of filterState.schoolYears" [ngValue]="year">{{year | schoolYear}}</option>
```
esotrope: fixed these goofy vars we set to 0 and '' to represent unset

webapp/src/main/webapp/src/app/student/results/student-results-filter.component.ts
```diff
@@ -36,31 +40,27 @@ export class StudentResultsFilterComponent implements OnInit {
     return this._showAdvancedFilters;
   }
 
-  ngOnInit(): void {
-    this.filterState.filterBy.onChanges.subscribe(() => this.onFilterChange(''));
```
esotrope: I split the advanced filters component "filterBy" model out of the "filterState" model used by the student history page to avoid some of the complexity here. also because this was bound to onFilterChange we were double reporting the advanced filters and also reporting them incorrectly as either "Subject" or "SchoolYear" changes

webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -39,52 +39,54 @@ <h2>{{'common.person-name' | translate:examHistory.student}}</h2> <div class="st
...
+              angulartics2On="click"
+              angularticsEvent="ToggleCollapseResults"
+              angularticsCategory="StudentHistory">
+        {{'common.action.' + (section.collapsed ? 'expand' : 'collapse') | translate}}
+        <span class="sr-only">{{ 'common.action.collapse' | translate }}</span>
+        <i class="fa fa-bold {{ section.collapsed ? 'fa-caret-square-o-down' : 'fa-caret-square-o-up' }}"></i>
+      </button>
 
-  <ng-container *ngFor="let assessmentType of assessmentTypes">
-    <ng-container *ngFor="let subject of getSubjectsForType(assessmentType)">
```
esotrope: removed the computed on the fly getters in favor of precomputing the data in the structure in which we want to display it.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -51,74 +41,42 @@ export class StudentResultsComponent implements OnInit {
...
-    if (this.examHistory) {
-      this.initializeFilter(this.examHistory.exams, this.route.snapshot.params);
+    const { examHistory } = this.route.snapshot.data;
+    if (examHistory) {
+      this.examHistory = examHistory;
+
+      const { exams } = examHistory;
+      this.filterState = this.createFilterState(exams, this.route.snapshot.params);
+      this.filterOptions.hasSummative = exams.some(wrapper => wrapper.assessment.isSummative);
+      this.filterOptions.hasInterim = exams.some(wrapper => wrapper.assessment.isInterim);
```
esotrope: basically advanced filters will disable controls based on these flags. before these were being updated on the fly based on your filters so if you filter all the summatives out the manner of admin filter would get disabled.

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -127,99 +85,90 @@ export class StudentResultsComponent implements OnInit {
...
   }
 
   getAssessmentTypeColor(assessmentType: string) {
-    const idx = AssessmentTypeColorOrder.indexOf(assessmentType);
-    return this.colorService.getColor(idx >= 0 ? idx + 1 : AssessmentTypes.length);
+    const index = [ 'ica', 'iab', 'sum' ].indexOf(assessmentType);
+    const totalAssessmentTypes = 3;
+    const colorIndex = index >= 0 ? index + 1 : totalAssessmentTypes;
+    return this.colorService.getColor(colorIndex);
   }
```
esotrope: do we really need this much logic behind picking a color for the assessments ?

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -127,99 +85,90 @@ export class StudentResultsComponent implements OnInit {
...
-      let byType: Map<string, StudentHistoryExamWrapper[]> = examsByTypeAndSubject.get(type) || new Map();
-      let bySubject: StudentHistoryExamWrapper[] = byType.get(subject) || [];
-      bySubject.push(wrapper);
-      byType.set(subject, bySubject);
-      examsByTypeAndSubject.set(type, byType);
-
-      // Initialize collapse state if it doesn't exist
-      this.displayState[ type ] = this.displayState[ type ] || {};
-      this.displayState[ type ][ subject ] = this.displayState[ type ][ subject ] || {};
-      this.displayState[ type ][ subject ].collapsed = this.displayState[ type ][ subject ].collapsed || false;
```
esotrope: no just no

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -127,99 +85,90 @@ export class StudentResultsComponent implements OnInit {
...
-
-    return true;
+    this.sections.forEach(section => {
+      section.filteredExams = section.exams.filter(exam => filteredExams.find(x => x.exam.id === exam.exam.id));
+    })
   }
 
   /**
    * Update the current route based upon the current filter state.
-   * TODO Do not re-load the page or re-fetch data.
```
esotrope: need to add this todo back. i was able to get the route to update without refreshing the page using `location.go` however hitting the back button does not change the page when this approach is used so i kept with using `router.navigate` for now

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -228,35 +177,37 @@ export class StudentResultsComponent implements OnInit {
...
-      });
+  private createFilterState(exams: StudentHistoryExamWrapper[], parameters: any): StudentResultsFilterState {
+
+    const filterState: StudentResultsFilterState = exams.reduce((filterState, wrapper: StudentHistoryExamWrapper) => {
+        const { schoolYear } = wrapper.exam;
+        if (filterState.schoolYears.indexOf(schoolYear) === -1) {
+          filterState.schoolYears.push(schoolYear);
+        }
+        const { subjectCode } = wrapper.assessment;
+        if (filterState.subjects.indexOf(subjectCode) === -1) {
```
esotrope: tried to use `Array.includes()` but karma settings didn't allow it, so i'll have to check that out later

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -228,35 +177,37 @@ export class StudentResultsComponent implements OnInit {
...
-    exams
-      .forEach((exam: StudentHistoryExamWrapper) => {
-        years.push(exam.exam.schoolYear);
-        subjects.push(exam.assessment.subject)
-      });
+  private createFilterState(exams: StudentHistoryExamWrapper[], parameters: any): StudentResultsFilterState {
+
+    const filterState: StudentResultsFilterState = exams.reduce((filterState, wrapper: StudentHistoryExamWrapper) => {
+        const { schoolYear } = wrapper.exam;
+        if (filterState.schoolYears.indexOf(schoolYear) === -1) {
```
esotrope: so right now the student history page uses the exam school year and not assessment school year for the school years filter - is this right?

webapp/src/main/webapp/src/app/student/student-exam-history.service.ts
```diff
@@ -48,20 +47,15 @@ export class StudentExamHistoryService {
    * @returns {Observable<boolean>} True if the student exists
    */
   existsBySsid(ssid: string): Observable<Student> {
-    const params: URLSearchParams = new URLSearchParams();
```
esotrope: unrelated: trying to get rid of all references to these deprecated http components

webapp/src/main/webapp/src/app/student/results/student-results-filter.component.ts
```diff
@@ -36,31 +40,27 @@ export class StudentResultsFilterComponent implements OnInit {
...
+    this.filterChange.emit(changedProperty);
 
     // track change event since wiring select boxes on change as HTML attribute is not possible
     this.angulartics2.eventTrack.next({
-      action: 'Change' + changeSource,
+      action: 'Change' + changedProperty,
       properties: {
         category: 'StudentHistoryAdvancedFilters',
-        label: changeSource === 'Subject' ? this.filterState.subject : this.filterState.schoolYear
+        label: changedProperty
```
wtritch: 👍

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: It's annoying that our original colors were based upon the AssessmentType enum order:
ICA, IAB, SUMMATIVE
What I *want* is that if we color the concept of "ICA" it's the same color everywhere.  Just like if we color the concept of "Grade 3" it's the same color everywhere.  When all of our assessment types used the same enum order that wasn't a problem...  Although, in all honesty I don't know if people would notice or care if the colors changed.
And just looking around a bit, we've already failed on that.  In our PDFs IABs are green and ICAs are magenta.  In our student history IABs are orange/brown and ICAs are green...

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: Doing an end-run around the router by messing with location has never ended well for me...  Even in AngularJS...

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
wtritch: I believe so because the table displays the enrolled grade rather than the assessment grade...

PR #1130, opened 2018-03-23
Defect/last option wider than others on checkbox group

esotrope:
Should we roll this into the hotfix too?


jtreuting:
i don't think so.  not that big a deal and that CSS change could impact unexpected things

PR #1135, opened 2018-03-23
accounting for Nonbinary gender on advanced filtering

dsnyder-fairway:
`Nonbinary` will need to be added to the `reporting.gender` table in order for this new value to be viewable

new checkbox input
screen shot 2018-03-23 at 4 15 37 pm

### DIFF
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.reporting.filter;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 
 import java.util.List;
@@ -12,12 +13,12 @@ public class ExamFilterOptions {
 
     private Set<Integer> schoolYears;
     private Set<String> ethnicities;
-    private Set<String> genders;
+    private List<String> genders;
 
     public ExamFilterOptions(final Iterable<Integer> schoolYears, final Iterable<String> ethnicities, final Iterable<String> genders) {
         this.schoolYears = schoolYears != null ? ImmutableSet.copyOf(schoolYears) : ImmutableSet.of();
         this.ethnicities = ethnicities != null ? ImmutableSet.copyOf(ethnicities) : ImmutableSet.of();
-        this.genders = genders != null ? ImmutableSet.copyOf(genders) : ImmutableSet.of();
+        this.genders = genders != null ? ImmutableList.copyOf(genders) : ImmutableList.of();
     }
 
     /**
@@ -34,7 +35,7 @@ public class ExamFilterOptions {
         return ethnicities;
     }
 
-    public Set<String> getGenders() {
+    public List<String> getGenders() {
         return genders;
     }
 
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
@@ -12,10 +12,12 @@ public class ExamFilterOptions {
 
     private Set<Integer> schoolYears;
     private Set<String> ethnicities;
+    private Set<String> genders;
 
-    public ExamFilterOptions(final Iterable<Integer> schoolYears, final Iterable<String> ethnicities) {
+    public ExamFilterOptions(final Iterable<Integer> schoolYears, final Iterable<String> ethnicities, final Iterable<String> genders) {
         this.schoolYears = schoolYears != null ? ImmutableSet.copyOf(schoolYears) : ImmutableSet.of();
         this.ethnicities = ethnicities != null ? ImmutableSet.copyOf(ethnicities) : ImmutableSet.of();
+        this.genders = genders != null ? ImmutableSet.copyOf(genders) : ImmutableSet.of();
     }
 
     /**
@@ -30,6 +32,10 @@ public class ExamFilterOptions {
      */
     public Set<String> getEthnicities() {
         return ethnicities;
+    }
+
+    public Set<String> getGenders() {
+        return genders;
     }
 
 }
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/gender/GenderRepository.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/gender/GenderRepository.java
@@ -1,0 +1,15 @@
+package org.opentestsystem.rdw.reporting.gender;
+
+import java.util.List;
+
+/**
+ * Repository responsible for accessing gender codes
+ */
+public interface GenderRepository {
+
+    /**
+     * @return all system genders
+     */
+    List<String> findAll();
+
+}
--- a/reporting-service/src/main/resources/application-no-permission.sql.yml
+++ b/reporting-service/src/main/resources/application-no-permission.sql.yml
@@ -75,6 +75,11 @@ sql:
         join ethnicity e on se.ethnicity_id=e.id
       where se.student_id=:student_id
 
+  gender:
+    findAll: >-
+      select id, code
+      from gender
+
   status:
     readTest: >-
       SELECT code FROM asmt_type
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -185,13 +185,13 @@ export class AssessmentsComponent {
     return GradeCode.getIndex(gradeCode);
   }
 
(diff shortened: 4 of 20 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/gender/GenderRepository.java
```diff
@@ -0,0 +1,15 @@
+package org.opentestsystem.rdw.reporting.gender;
+
+import java.util.List;
+
+/**
+ * Repository responsible for accessing gender codes
+ */
+public interface GenderRepository {
```
malaffoon: Isn't there a CodedEntityRepository that does this sort of thing?

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters-toggle.component.html
```diff
@@ -21,12 +22,15 @@ <h2 class="h3 gray-darkest pull-left">
...
-                </span>
           <span *ngIf="property.indexOf('ethnicities') != -1">
                   <strong>{{ filters['ethnicities'].enumValue + property.substring(property.indexOf('.')) | translate}}</strong> {{ filters['ethnicities'].label | translate }}
                 </span>
+          <span *ngIf="property.indexOf('genders') != -1">
+                  <strong>{{ filters['genders'].enumValue + property.substring(property.indexOf('.')) | translate}}</strong> {{ filters['genders'].label | translate }}
+                </span>
+          <span *ngIf="property.indexOf('ethnicities') == -1 && property.indexOf('genders') == -1">
+                  <strong>{{ filters[property].enumValue + '.' + filterBy[property] | translate}}</strong> {{ filters[property].label | translate }}
+                </span>
```
dsnyder-fairway: should we remove this logic from the `html`?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/gender/GenderRepository.java
dsnyder-fairway: it looks like that is in the aggregate service

dsnyder-fairway: should I or could I move this to common?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
```diff
@@ -32,4 +34,8 @@ public ExamFilterOptions(final Iterable<Integer> schoolYears, final Iterable<Str
         return ethnicities;
     }
 
+    public Set<String> getGenders() {
```
malaffoon: not quite sure where things are wrt ordering. The database order for this may be reliable so this could be a List?

malaffoon:
The change to make it data-driven is good but I think we may have to clarify some things before completing this. There is supposed to be "NonBinary", a new option like Male/Female, and a "NotSelected" option.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/gender/GenderRepository.java
malaffoon: Ah, okay.

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -88,17 +91,17 @@
     <p class="border-bottom">{{ 'common.filters.student.label' | translate }}</p>
     <div class="flex-children">
       <!-- Column 1 -->
-      <div class="flex-child col1 no-margin">
+      <div class="flex-child col1 grow">
```
dsnyder-fairway: changing from `no-margin` to `grow` allowed for this to display more neatly

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
```diff
@@ -113,11 +117,17 @@ export class ExamFilterService {
   private getFilters(filterBy : FilterBy): string[] {
```
dsnyder-fairway: this method feels like it doesn't really belong here. I'd like to move this into `FilterBy`. Thoughts?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
esotrope: yeah i second that - these should be lists eventually

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/gender/GenderRepository.java
esotrope: would be nice if spring/java supported autowiring based on generics : J that could kill so much boilerplate classes

reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -75,6 +75,11 @@ sql:
         join ethnicity e on se.ethnicity_id=e.id
       where se.student_id=:student_id
 
+  gender:
+    findAll: >-
+      select id, code
```
esotrope: just need to select code right?
Also i would put an explicit order by id here like we do in aggregate stuff to be explicit

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -185,13 +185,13 @@ export class AssessmentsComponent {
...
-    if (this.clientFilterBy.filteredEthnicities.length == 0) {
-      this.clientFilterBy.ethnicities[ 0 ] = true; // None are selected, set all to true.
+  removeFilterFromArray(arr: boolean[], filterArr: any[], gender: string) {
+    arr[ gender ] = false;
+    if (filterArr.length == 0) {
+      arr[ 0 ] = true; // None are selected, set all to true.
     }
 
-    this.clientFilterBy.ethnicities = Object.assign({}, this.clientFilterBy.ethnicities);
+    arr = Object.assign({}, arr);
```
esotrope: : / is `Object.assign(object, array)` something that is done? what is the goal here?

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters-toggle.component.html
esotrope: this whole component needs a serious overhaul. its data modeling is not intuitive or flexible. I would hold off on big changes to it though until we know if we are going to overhaul it with something Aaron suggests though? maybe that is hoping for something that will not come though 😞

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
dsnyder-fairway: good question. I'll have to review what removing this method actually does. Also noticed `gender` hanging around here when it should have some other name

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -505,7 +505,8 @@
     },
     "gender": {
       "Female": "Female",
-      "Male": "Male"
+      "Male": "Male",
+      "Unknown": "Unknown"
```
esotrope: is it really "unknown"? or do we want to say Not Stated?
I would think we would do like we did in aggregate and map code "undefined" to "Not Stated"

dsnyder-fairway: I think "Undeclared" may be more appropriate. Not stated could be interpreted as the person using Smarter Balanced has not stated the gender (which should be "All")

esotrope: possibly yeah - the idea is "not stated" goes hand in hand with "decline to state" which is usually how that option appears on a form

malaffoon: If you look at the clarifications on the requirements page, it will be "Not Selected". And there will also be a "Non Binary".

dsnyder-fairway: In the comments I see yesterday
> Per the discussion with CDE yesterday, please replace references to "other" with "nonbinary" (no hyphen). This should appear in the filters with All, Male, and Female.
"Not Selected" I think may be outdated

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
dsnyder-fairway: updated

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: They didn't go for "Label Transcendent" huh?

PR #1137, opened 2018-03-24
adds autocomplete=off to all input type=text

esotrope:
This goes ahead and adds autocomplete=off to all input type=text which i think is what we want anyway right?


esotrope:
@malaffoon that is probably a good idea - i'll wait for more team input before changing again though.

esotrope:
please re-review, limited autocomplete=off to just typeahead text inputs

PR #1139, opened 2018-03-26
Refactor the ReportQuery object for re-use in both basic and filtered…

wtritch:
… dimension aggregate report requests.

This PR lays the groundwork for supporting two Aggregate Report requests:
1. The existing request method with a single set of filters and 1..n Dimension drill-downs defined as DimensionTypes
2. The upcoming request method with 1..n Dimension drill-downs defined as named sets of filters

Please check out the new models BasicAggregateReportQuery (1.) and FilteredDimensionReportQuery (2.)

Notice that I've deprecated and plan to leave the current ReportQuery representation class in the code-base to help with deserializing soon-to-be-legacy AggregateReportRequest#reportQuery instances as NewAggregateReportRequest#query instances.


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.List;
+import java.util.Set;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * This class provides base functionality for Aggregate report queries.
+ */
+public abstract class AbstractAggregateReportQuery {
```
wtritch: This abstract class contains all of the properties that are shared between BasicAggregateReportQuery and FilteredDimensionReportQuery

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AggregateReportFilters.java
```diff
@@ -0,0 +1,148 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import java.util.Set;
+
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * This class represents student exam result filters for aggregate reports.
+ */
+public class AggregateReportFilters {
```
wtritch: As @agorina recommended, I broke filters out of flat properties on a ReportQuery and into a re-usable filter-set class.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/FilteredDimensionReportQuery.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Map;
+
+import static com.google.common.collect.Maps.newHashMap;
+
+/**
+ * This class represents an aggregate report query specifying the break-downs
+ * for the report as named filter sets.
+ */
+public class FilteredDimensionReportQuery extends AbstractAggregateReportQuery {
```
wtritch: This class has caused me naming heartburn.  I don't like using "custom" because it's a pretty non-descriptive label.  Since this class provides dimensional breakdown with filter sets, this seems appropriate-ish?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/FilteredDimensionReportQuery.java
```diff
@@ -0,0 +1,44 @@
...
+    private Map<String, AggregateReportFilters> dimensions;
```
wtritch: I know we were still discussing whether or not the user gets to name these filter-based dimensions, but we could always auto-name the dimensions for them and provide details on hover and/or in the query details section above the results.  The CSV export will likely need to expand the entire "Dimension" column to a concatenated string value like "Gender: Male | Section504: yes | MigrantStatus: ... | etc" since CSVs don't support value-by-reference very well.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/NewAggregateReportRequest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * This class represents a basic aggregate report request.
+ * TODO: This will replace the existing AggregateReportRequest once the feature is fully implemented.
+ */
+public class NewAggregateReportRequest extends AbstractExamReportRequest<PrintOptions> {
```
wtritch: I intentionally named this awkwardly since it will be overwriting the existing AggregateReportRequest once we've made the query model changes in the UI, report-processor, and aggregate-service.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/jackson/AbstractExamReportRequestDeserializer.java
```diff
@@ -28,35 +24,8 @@ public AbstractExamReportRequestDeserializer() {
...
-
-    private JsonNode getReportType(final JsonNode jsonNode) {
-        return jsonNode.has("reportType") ? jsonNode.get("reportType") : MissingNode.getInstance();
-    }
-
-    private boolean hasNode(final JsonNode jsonNode, final String field) {
-        return jsonNode.has(field);
-    }
-
-    private AbstractExamReportRequest deserializeLegacyJsonByProperty(final JsonNode jsonNode,
```
wtritch: On a good note, we don't need this legacy deserialization support any longer.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/jackson/NewAggregateReportRequestDeserializer.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * This temporary deserializer is responsible for deserializing AggregateReportRequest instances.
+ * It is responsible for handling legacy serialized instances containing a ReportRequest rather
+ * than a BasicAggregateReportRequest.
+ * TODO Remove once we no longer have V1_2 serialized AggregateReportRequests in a production environment
+ *
+ * TODO Rename/rebind this deserializer to "AggregateReportRequest" once the feature is ready.
+ */
+public class NewAggregateReportRequestDeserializer extends DelegatingDeserializer {
```
wtritch: In order to deserialize a legacy AggregateReportRequest containing a ReportQuery property I'm testing for the existence of a "reportQuery" field name.  If it exists, I know I'm dealing with a serialized legacy instance and need to convert from a ReportQuery to a BasicAggregateReportQuery during de-serialization.  Otherwise I delegate to the standard deserializer.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/FilteredDimensionReportQuery.java
jtreuting: i like SubGroup better than Dimension but i'm not sure I like it enough to say it should change.  i can't think of anything much better here.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/FilteredDimensionReportQuery.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public Map<String, AggregateReportFilters> getDimensions() {
+        return dimensions == null ? newHashMap() : dimensions;
```
malaffoon: Intentionally returning the (possibly) mutable collection? just ask cause in the previous class you make a copy to return.

Honestly, i wish we would settle on either immutable collections or return iterables instead of collections. As it is, we always have to copy the collection to be safe, even if the caller is just going to iterate over the values.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/NewAggregateReportRequest.java
malaffoon: So you intentionally made it so this awkward name would live forever? If we end up iterating on our report request again, i guess we can revert to a non-awkward name?

malaffoon:
Somehow i think this should be easier (discriminator-driven deserialization) but all my previous experience relies on the polymorphism having been backed in from the beginning so perhaps this is the best we can do. LGTM.

wtritch:
@malaffoon I felt the same way about this feeling more complex than it should. Deserializing an abstract instance into a concrete instance is easy based off of a discriminator (we're doing it in AbstractExamReportRequestDeserializer). 
Deserializing both current and legacy values into a current instance has a little less baked-in support.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/FilteredDimensionReportQuery.java
wtritch: shoot.  That's a good point.  I dislike copying-on-get, since a value may be accessed many times... how would you feel about creating an immutable copy in the builder at build time?
Think we should change the model getters to return `ImmutableMap<blah, blah>`, `ImmutableList<blah>`, etc?  I'm not sure I'm ready to make the full shift to returning `Iterable<blah>` which could also potentially be mutable (Iterator#remove)?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -0,0 +1,217 @@
...
+    private boolean includeState;
+    private boolean includeAllDistricts;
+    private boolean includeAllSchoolsOfDistricts;
+    private boolean includeAllDistrictsOfSchools;
+    private ImmutableSet<Long> districtIds;
+    private ImmutableSet<Long> schoolIds;
+    //pass through values
+    private String valueDisplayType;
+    private String achievementLevelDisplayType;
+    private ImmutableList<String> columnOrder;
```
wtritch: @malaffoon I changed the properties to be ImmutableFoo<Bar>s but I didn't change the getters to reflect that... I'm still debating whether the getters should advertise that the consumer is getting an immutable instance... thoughts?

esotrope: +1 immutables because it is a little more efficient than defensive copy.

Also, guava does recommend using ImmutableXYZ in interfaces for the stated reasons which i don't think i'm opposed to since the functionality with just XYZ collection interfaces is different. however i wouldn't recommend making that change unless we do it wholesale and i dont think its worth the effort at this point

PR #1140, opened 2018-03-26
accounting for Nonbinary gender on advanced filtering

dsnyder-fairway:
`Nonbinary` will need to be added to the `reporting.gender` table in order for this new value to be viewable

new checkbox input
screen shot 2018-03-23 at 4 15 37 pm

### DIFF
--- a/reporting-service/src/main/resources/application-no-permission.sql.yml
+++ b/reporting-service/src/main/resources/application-no-permission.sql.yml
@@ -75,6 +75,11 @@ sql:
         join ethnicity e on se.ethnicity_id=e.id
       where se.student_id=:student_id
 
+  gender:
+    findAll: >-
+      select id, code
+      from gender
+
   status:
     readTest: >-
       SELECT code FROM asmt_type
--- a/reporting-service/src/main/resources/application-no-permission.sql.yml
+++ b/reporting-service/src/main/resources/application-no-permission.sql.yml
@@ -77,8 +77,8 @@ sql:
 
   gender:
     findAll: >-
-      select id, code
-      from gender
+      select code
+      from gender order by id;
 
   status:
     readTest: >-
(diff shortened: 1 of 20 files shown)


reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -75,6 +75,11 @@ sql:
         join ethnicity e on se.ethnicity_id=e.id
       where se.student_id=:student_id
 
+  gender:
+    findAll: >-
+      select id, code
+      from gender
```
esotrope: - id + explicit order

dsnyder-fairway: done

malaffoon:
I really don't care which but is it going to be "NonBinary" or "Nonbinary"? I think Devin has it correct as "Nonbinary" but just want to make sure so i'm consistent when i make the back-end changes. @jtreuting ?

jtreuting:
@malaffoon I'm guessing you mean the code that is sent over in the TRT?  for that I don't care either way.  what about making the ingest check case insensitive.

for the label, i believe that should be "Non-Binary".  it is hyphenated when i google it.

dsnyder-fairway:
@jtreuting I went with Nonbinary to follow the comment on the sow

> Per the discussion with CDE yesterday, please replace references to "other" with "nonbinary" (no hyphen). This should appear in the filters with All, Male, and Female.

jtreuting:
okay, leave it for now and i'll double check on that since it doesn't fit with what i'm seeing on google.

dsnyder-fairway:
@jtreuting @malaffoon @esotrope is it okay to merge this?

esotrope:
can you wait till i finish merging the hotfix back into develop then i think you're good to merge

PR #1144, opened 2018-03-27
Task/abstract aggregate report request

wtritch:
Additional changes/new models in support of a polymorphic aggregate report flow:
aggregate-report-flow 3


admin-service/build.gradle
```diff
@@ -20,7 +20,6 @@ dependencies {
     compile 'org.springframework.cloud:spring-cloud-starter-stream-rabbit'
     compile 'org.springframework.cloud:spring-cloud-stream'
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
-    compile 'com.fasterxml.jackson.datatype:jackson-datatype-guava'
```
wtritch: Since this existed in every service *except* for aggregate-service (which now needs it) and since we now have Guava-aware models in reporting-common, I've moved the dependency out of the sub-modules and into reporting-common.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/ApplicationConfiguration.java
```diff
@@ -74,6 +75,7 @@ public ArchiveService archiveService(final ArchiveServiceFactory factory, final
     @Bean
     public Jackson2ObjectMapperBuilderCustomizer customizeObjectMapper() {
         return builder -> builder
-                .failOnUnknownProperties(false);
+                .failOnUnknownProperties(false)
+                .modulesToInstall(new AggregateReportModule());
```
wtritch: Since the AggregateQuery models exist in reporting-common, I've added a Jackson module for de/serializing them to reporting-common as well, and registered the module in the aggregate-service and reporting-service where the models are used.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/BasicAggregateReportResult.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Collection;
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * Aggregate report result for {@link org.opentestsystem.rdw.reporting.common.model.BasicAggregateReportQuery}
+ * requests.
+ */
+public class BasicAggregateReportResult extends AggregateReportResult<ImmutableList<BasicAggregateRow>> {
```
wtritch: I'm using an ImmutableList as the response so I don't need to provide a defensive copy on an overridden getPayload() implementation.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AggregateQuery.java
```diff
@@ -0,0 +1,12 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * Implementations of this interface represent Aggregate report queries.
+ */
+public interface AggregateQuery {
```
wtritch: All future aggregate queries will need to implement this shallow interface.  It just requires a single getter that returns a type enum used for easy acceptance/handler testing and deserialization.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AggregateQueryType.java
```diff
@@ -0,0 +1,22 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * This enum contains the available types of aggregate queries.
+ */
+public enum AggregateQueryType {
+    Basic(BasicAggregateReportQuery.class),
+    FilteredSubgroup(FilteredSubgroupReportQuery.class);
```
wtritch: This will be extended with longitudinal, cohort, claim, etc query types.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AggregateRequestMessage.java
```diff
@@ -8,7 +8,8 @@
 public class AggregateRequestMessage {
 
     private long id;
-    private ReportQuery query;
+    @Deprecated private ReportQuery query;
+    private AggregateQuery aggregateQuery;
```
wtritch: This is the object that goes through the RabbitMQ queue.  It will eventually only contain a polymorphic AggregateQuery.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AggregateRow.java
```diff
@@ -0,0 +1,54 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * Base aggregate report row representing an identified collection of measures.
+ */
+public abstract class AggregateRow {
```
wtritch: New base class for the ReportRow replacement models.  As @agorina pointed out, this may be abstracted further since the measures will likely be different for different reports.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
```diff
@@ -47,7 +47,7 @@
     public Jackson2ObjectMapperBuilderCustomizer customizeObjectMapper() {
         return builder -> builder
                 .failOnUnknownProperties(false)
-                .modulesToInstall(new GuavaModule());
+                .modulesToInstall(new AggregateReportModule());
```
wtritch: The GuavaModule is installed via AutoConfiguration.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportResult.java
```diff
@@ -0,0 +1,60 @@
+package org.opentestsystem.rdw.olap.model;
+
+/**
+ * General aggregate report result container.
+ * The generic payload contains the report results.
+ */
+public abstract class AggregateReportResult<T> {
+
+    private T payload;
```
esotrope: isn't this whole class a payload? : J

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/BasicAggregateReportResult.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder extends AggregateReportResult.Builder<Builder, BasicAggregateReportResult, ImmutableList<BasicAggregateRow>> {
+
+        private final List<BasicAggregateRow> rows = newArrayList();
```
esotrope: could use ImmutableList.builder() then .build()

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/NewAggregateReportRequest.java
```diff
@@ -1,58 +1,22 @@
...
-import org.opentestsystem.rdw.reporting.common.model.BasicAggregateReportQuery;
-
-import java.util.List;
-
-import static com.google.common.collect.Lists.newArrayList;
+import org.opentestsystem.rdw.reporting.common.model.AggregateQuery;
 
 /**
  * This class represents a basic aggregate report request.
+ * TODO: How to get subject(s) and schoolyear(s) out of generic AggregateQuery? Should we?
```
wtritch: I'm not sure what to do about this... Right now the back-end is returning a collection of generic Report instances and each has a `subject` and `schoolYear` property.  This is a legacy of starting with PDFs/ExamExport which only allowed a single subject and school year to be specified.  
The problem is that the light-weight AggregateQuery interface only has a `getQueryType` method and I really don't want to pollute it with `getSubjects()` and `getSchoolYears()` since those may not be applicable to all future AggregateQuery implementations and they shouldn't care about how the report-processor is using them.

@esotrope and I discussed whether or not we should even include the Subject and SchoolYear columns in the "My Reports" table since there's no guarantee that they will have meaning for all future Report types...  
However, if we want to keep those columns (and I'll assume we do for now) then I see two options:
1. registered handlers that can pull a collection subjects and school years out of a given AbstractExamReportRequest after testing for and knowing the instance/query classes that it's dealing with.
2. extend AggregateQuery with `getSubjects()` and `getSchoolYear()`
3. leave polymorphic support to the front-end (which would likely need to do something similar to 1. in TypeScript)

Right now my vote is for 1., thoughts?

esotrope: I vote #1.
I think we should eventually have the logic on the frontend, but for now we keep it backend? not sure. that is my only q i think.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportResult.java
wtritch: I'd consider this class an "annotated payload" =P

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/BasicAggregateReportResult.java
wtritch: +1 done

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/NewAggregateReportRequest.java
wtritch: #1 works for me.  I'm comfortable keeping it on the back-end for now just to keep the API a little cleaner.

PR #1146, opened 2018-03-27
DWR-1627 Aggregate Service tests

agorina:
This is not final. I am working on more tests. Just want to have smaller commits.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/QueryPoolExecutor.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/QueryPoolExecutor.java
@@ -1,5 +1,6 @@
-package org.opentestsystem.rdw.olap.repository.impl;
+package org.opentestsystem.rdw.olap.service;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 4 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/QueryPoolExecutor.java
```diff
@@ -1,5 +1,6 @@
-package org.opentestsystem.rdw.olap.repository.impl;
+package org.opentestsystem.rdw.olap.service;
```
agorina: I moved it since it is now used by the service.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/QueryPoolExecutor.java
```diff
@@ -14,9 +15,13 @@
 @EnableAsync
 public class QueryPoolExecutor {
 
-    @Value("${app.aggregate-reports.query-pool-size}")
     private int partitionSize;
 
+    @Autowired
+    public void setPartitionSize(final @Value("${app.aggregate-reports.query-pool-size}") int partitionSize) {
+        this.partitionSize = partitionSize;
```
agorina: Changed this to be able to call it from a test.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/service/impl/DefaultCustomAggregateServiceRST.java
```diff
@@ -65,7 +66,8 @@
 @SuppressWarnings({"SqlNoDataSourceInspection", "SqlResolve"})
 @RunWith(SpringRunner.class)
 @SpringBootTest(classes = {
-        ReportingSystemSettings.class
+        ReportingSystemSettings.class,
+        QueryPoolExecutor.class
 })
```
agorina: This is ignored test, I will delete it soon.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/QueryPoolExecutor.java
wtritch: +1 Although if you'd rather inject an AggregateReportsProperties, I'd be cool with that too =)

PR #1148, opened 2018-03-28
schema update

malaffoon:
This updates the schema dependency which involves adding elas_id, elas_start_at and removing completeness_id and administration_condition_id (only from reporting data mart). All this is doing is making it go, i still have to add the ELAS functionality.


malaffoon:
I'm going to merge so i can push forward, will address any concerns.

agorina:
Looks good.

PR #1150, opened 2018-03-28
corrected creation of cacheKey in caching-data service

dsnyder-fairway:
Issue here was we were expecting options to be passed. Another issue was that `createKey` returns url is options is null (but on line 35 we were passing `{}`, not `null`)

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
+++ b/webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
@@ -32,7 +32,7 @@ export class CachingDataService {
 
   public get(url: string, options?: RequestOptionsArgs): Observable<any> {
 
-    const cacheKey = this.createKey(url, options ? options.params || {} : {});
+    const cacheKey = this.createKey(url, options);
 
     const response = this.responsesByUrl.get(cacheKey);
     if (response) {


jtreuting:
which JIRA is this for?

dsnyder-fairway:
DWR-1649

webapp/src/main/webapp/src/app/shared/data/caching-data.service.ts
```diff
@@ -32,7 +32,7 @@ export class CachingDataService {
 
   public get(url: string, options?: RequestOptionsArgs): Observable<any> {
 
-    const cacheKey = this.createKey(url, options ? options.params || {} : {});
+    const cacheKey = this.createKey(url, options);
```
esotrope: this is not right. remember we want options.params.
should be:
```typescript
options && options.params ? options.params : {}
```
I think what is there will work if we add parens too
```typescript
options ? (options.params || {}) : {}
```

esotrope: can we make some unit tests around this?

dsnyder-fairway: line 60 here starts with 
```
if (options == null) {
  return url;
}
```

esotrope: gotcha - i see - thx 👍

dsnyder-fairway:
I'll work on a UT for this (future PR)

PR #1151, opened 2018-03-28
UT to demonstrate caching data service fix

dsnyder-fairway:


webapp/src/main/webapp/src/app/shared/data/caching-data.service.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      expect(value).toEqual({ url: passedUrl, params: options.params });
+    });
+    service.get(passedUrl, options).subscribe((value) => {
+      expect(dataService.get).toHaveBeenCalledWith(passedUrl, options);
+      expect(dataService.get).toHaveBeenCalledTimes(1);
+      expect(value).toEqual({ url: passedUrl, params: options.params });
+    });
+  });
+
+  it('should handle different options', () => {
```
dsnyder-fairway: this test fails without my changes from #1150

webapp/src/main/webapp/src/app/shared/data/caching-data.service.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      expect(url).toBe(passedUrl);
+      return of({
+        url: passedUrl,
+        params: options.params
+      });
+    });
+
+    service.get(passedUrl, options1).subscribe((value) => {
+      expect(dataService.get).toHaveBeenCalledWith(passedUrl, options1);
+      expect(dataService.get).toHaveBeenCalled();
```
esotrope: i think `toHaveBeenCalled()` is redundant here with `toHaveBeenCalledWith()`

webapp/src/main/webapp/src/app/shared/data/caching-data.service.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  it('should handle options', () => {
+    let passedUrl: string = "/something";
+    let options = {
+      params: {
+        oneParam: "oneValue"
+      }
+    };
+    dataService.get.and.callFake((url, options) => {
+      expect(url).toBe(passedUrl);
```
esotrope: not sure we should assert in the stubbed method

webapp/src/main/webapp/src/app/shared/data/caching-data.service.spec.ts
dsnyder-fairway: 👍 I simplified this to reflect what the intention of each test is. If followed by `toHaveBeenCalledTimes(1)` is to show we hit the cache. When using  `toHaveBeenCalledWith` demonstrates we are populating the cache

webapp/src/main/webapp/src/app/shared/data/caching-data.service.spec.ts
dsnyder-fairway: removed

PR #1154, opened 2018-03-29
added elas to adv-filters

dsnyder-fairway:
not using the filter
screen shot 2018-03-29 at 3 54 36 pm

using the filter
screen shot 2018-03-29 at 3 54 41 pm


report-processor/src/main/resources/application.yml
```diff
@@ -146,3 +146,5 @@ reporting:
   state:
     code: CA
     name: California
+  # english-learners should be lep or elas
+  english-learners: lep
```
dsnyder-fairway: keeping this at `lep` for now. Also, I would like a better name than `english-learners` but couldn't come up with anything better

dsnyder-fairway:
Advanced Filters is at (or has already surpassed) the amount of stuff we can jam into one row. How should we break this up?

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettings.java
```diff
@@ -20,6 +20,23 @@
...
+        this.englishLearners = englishLearners;
+    }
+
+    public boolean isLepEnabled() {
+        return "lep".equals(englishLearners);
+    }
+
+    public boolean isElasEnabled() {
+        return "elas".equals(englishLearners);
+    }
```
dsnyder-fairway: this allows for some flexibility if there was ever a some need for both lep and elas. We make no assumptions on the front end that this is an XOR property

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettings.java
```diff
@@ -20,6 +20,23 @@
...
+
+    public String getEnglishLearners() {
+        return englishLearners;
+    }
+
+    public void setEnglishLearners(final String englishLearners) {
+        this.englishLearners = englishLearners;
+    }
+
+    public boolean isLepEnabled() {
```
esotrope: i know we abbr this but should we? comments should at least expand this text for us

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettings.java
esotrope: wondering if this is more appropriately a collection? and we test enabled with collection.contains

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
```diff
@@ -5,7 +5,7 @@
 <div class="flex-children">
 
   <!-- Test Filters -->
-  <div id="filter-by-test" class="flex-child">
+  <div id="filter-by-test" class="flex-child no-margin">
```
esotrope: this a bootstrap class or ui kit class? if ui-kit i'm surprised we didn't call it `m-0` to go with `mX-0`

webapp/src/main/webapp/src/app/assessments/model/filter-by.model.ts
```diff
@@ -54,6 +59,11 @@ export class FilterBy extends ObservableObject {
         for (let filteredGender of filteredGenders) {
           all.push(property + "." + filteredGender);
         }
+      } else if (property == "elasCodes") {
+        const filteredElasCodes = this.filteredElasCodes;
+        for (let i of filteredElasCodes) {
```
esotrope: i should be reserved for representing `index` also when given the op - change the `let` to `const` when the value isn't updated in the block

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -410,6 +410,13 @@
       "Section504": "504",
       "StudentEnrolledGrade": "Student Enrolled Grade"
     },
+    "elas": {
+      "EO": "English Only",
+      "EL": "English Learner",
+      "IFEP": "Initial Fluent English Proficient",
+      "RFEP": "Reclassified Fluent English Proficient",
+      "TBD": "To Be Determined"
+    },
```
esotrope: this going to be something we disaggregate on in aggregates?

webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.html
dsnyder-fairway: its in the ui kit. Looking around it looks like there are a bunch is a ml-0, mr-0, mt-0, mb-0, but I don't see anything for all margins

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettings.java
dsnyder-fairway: removed abbr and added some javadoc

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettings.java
dsnyder-fairway: good call. Changed

webapp/src/main/webapp/src/assets/i18n/en.json
dsnyder-fairway: I believe so..

PR #1155, opened 2018-03-30
Add the report type to the user_report table for DBA analysis.

wtritch:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/AggregateReportRequest.java
```diff
@@ -15,6 +15,11 @@ public ReportType getReportType() {
         return ReportType.AggregateReportRequest;
     }
 
+    @Override
+    public String getReportTypeDisplayValue() {
+        return getReportType().name() + " " + (query == null ? "null" : query.getQueryType());
+    }
```
wtritch: Since I want aggregate reports to include their query type, I needed a "display value" getter.

esotrope: not sure why it is called display value - do we display this value? we translate it no? Maybe call it a ReportTypeCode?

wtritch: hm... I guess it's a kind of code.  It's the code by default, but can be overridden to include whatever else.  We never translate it, we never display it.  We literally set the value in the DB on insert and ignore it.  That's what I meant by DBA analysis =)

PR #1159, opened 2018-03-30
sort schools by shortest school name first

dsnyder-fairway:
sort by shortest length first. This works well for finding the closest matched text

### DIFF
--- a/common/src/main/resources/common.sql.yml
+++ b/common/src/main/resources/common.sql.yml
@@ -66,6 +66,7 @@ sql:
         WHERE 1=:allow_school
           AND (:name_query IS NULL OR school.name LIKE concat('%', :name_query, '%'))
           AND (1=:statewide OR school.district_group_id IN (:district_group_ids) OR school.district_id IN (:district_ids) OR school.school_group_id IN (:school_group_ids) OR school.id IN (:school_ids))
+        ORDER BY LENGTH(name)
         LIMIT :limit
 
       findOne: >-
--- a/common/src/test/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepositoryIT.java
+++ b/common/src/test/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepositoryIT.java
@@ -101,9 +101,10 @@ public class ReportingJdbcOrganizationRepositoryIT {
     public void itShouldFilterSchoolGroupsByPermissions() {
         final Collection<Organization> allOrgs = repository.findAll(schoolGroups(-10L), allOrgsQuery);
         assertThat(allOrgs.stream().map(Organization::getName))
-                .containsOnly(
-                        "schoolgroup1",
-                        "school1");
+                .containsExactly(
+                        "school1",
+                        "schoolgroup1"
+                        );
     }
 
     @Test
@@ -212,11 +213,11 @@ public class ReportingJdbcOrganizationRepositoryIT {
                 .build();
         final Collection<Organization> matching = repository.findAll(statewide(), namedQuery);
         assertThat(matching.stream().map(Organization::getName))
-                .containsOnly(
-                        "districtgroup1",
+                .containsExactly(
+                        "school1",
                         "district1",
                         "schoolgroup1",
-                        "school1");
+                        "districtgroup1");
     }
 
     @Test
--- a/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
+++ b/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
@@ -23,8 +23,6 @@ import { Observable } from "rxjs/Observable";
 import { SchoolTypeahead } from "../../shared/school/school-typeahead";
 import { map, mergeMap } from "rxjs/operators";
 import { limit } from "../limit";
-import { ordering } from "@kourge/ordering";
-import { byNumber } from "@kourge/ordering/comparator";
 
 @Component({
   selector: 'school-results',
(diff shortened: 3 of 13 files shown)


dsnyder-fairway:
downside here is that searching for part of a school name may lead to still not finding the school you were looking for. On the other hand, this doesn't introduce that issue

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
```diff
@@ -155,10 +157,10 @@ export class SchoolResultsComponent implements OnInit {
...
                 this.organizationService.searchSchoolsWithDistrictsBySchoolName(search)
                   .pipe(map(
-                    (organizations: any[]) =>
-                      organizations.filter(
-                        organization => this.organizations.findIndex(x => organization.equals(x)) === -1
-                      ))
+                    (schools: School[]) =>
+                      schools.filter(
+                        (school: School) => this.organizations.findIndex(x => school.equals(x)) === -1
+                      ).sort(ordering(byNumber).on<School>(s => s.name.length).compare))
```
esotrope: i think ideally we want to do this on the backend no?
could be wrong but i think there is the case where the exact match is not within the natural 20 first results so it doesnt even make its way to the frontend right @wtritch?

webapp/src/main/webapp/src/app/shared/organization/organization.service.ts
```diff
@@ -27,17 +27,15 @@ export class OrganizationService {
...
-          limit: limit
-        }
-      });
-    } else {
-      return this.dataService.get(`${ReportingServiceRoute}/organizations/schools`, {
-        params: {}
-      });
+      options.params = {
+        limit: limit
+      };
```
esotrope: i think you can just always pass `{params: {limit: limit}}` and spring will consider `limit: undefined` to be an empty/missing param

webapp/src/main/webapp/src/app/school-grade/results/school-results.component.ts
malaffoon: That seems correct Landon. This should happen on the backend.

dsnyder-fairway: done (moved to the query)

webapp/src/main/webapp/src/app/shared/organization/organization.service.ts
dsnyder-fairway: you were correct. this has been updated

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
```diff
@@ -68,7 +68,7 @@ export class SBTypeahead implements OnInit {
           .sort(
             join(
               ordering(byString).on<Option>(o => o.group === undefined ? '' : o.group).compare,
-              ordering(byString).on<Option>(o => o.label).compare
+              ordering(byNumber).on<Option>(o => o.label.length).compare
```
wtritch: If the back-end is sorting the results, I wouldn't sort on the front-end as well.

common/src/main/resources/common.sql.yml
```diff
@@ -66,6 +66,7 @@ sql:
         WHERE 1=:allow_school
           AND (:name_query IS NULL OR school.name LIKE concat('%', :name_query, '%'))
           AND (1=:statewide OR school.district_group_id IN (:district_group_ids) OR school.district_id IN (:district_ids) OR school.school_group_id IN (:school_group_ids) OR school.id IN (:school_ids))
+        ORDER BY LENGTH(name)
```
wtritch: Maybe add a validating integration test to verify that aaa comes before aaaa comes before etc?

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
dsnyder-fairway: ah you are right. Forgot this hit the same endpoint :-/

dsnyder-fairway: hmmm. Removing this line, or the entire sort makes the ordering not pretty.

common/src/main/resources/common.sql.yml
dsnyder-fairway: sounds good

webapp/src/main/webapp/src/app/shared/form/sb-typeahead.component.ts
dsnyder-fairway: Oh I see...the Controller returns a set

dsnyder-fairway: done

common/src/test/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcOrganizationRepositoryIT.java
```diff
@@ -101,9 +101,10 @@ public void itShouldFilterDistrictsByPermissions() {
     public void itShouldFilterSchoolGroupsByPermissions() {
         final Collection<Organization> allOrgs = repository.findAll(schoolGroups(-10L), allOrgsQuery);
         assertThat(allOrgs.stream().map(Organization::getName))
-                .containsOnly(
-                        "schoolgroup1",
-                        "school1");
+                .containsExactly(
+                        "school1",
+                        "schoolgroup1"
+                        );
```
dsnyder-fairway: @wtritch some changes here to reflect the ordering. I went with these two tests as these have distinct name lengths. When the name length is the same, 2 may precede 1.

PR #1164, opened 2018-04-02
Fix item-info translations.

wtritch:
I think this is a sign that en.json is getting too large to effectively manage via pull requests alone.  I wonder if we want to create a pre-release task to go check all translation keys for usage?  Or break it up by feature and have a webpack task that consolidates at build time?


dsnyder-fairway:
I think this was just a mix up with my refactoring of the file. Part of my task was to go through and check for usage, but it looks like I missed something. We could introduce a UT to verify usage

malaffoon:
It could also be an artifact of cross-branch merging.

PR #1166, opened 2018-04-02
Filtered subgroup version of the custom analysis report

agorina:
I am still working on more tests, but would like to start the review.

Some background:
Form the API point of view we have two different queries (`FilteredSubgroupReportQuery` and `BasicAggregateReportQuery`) and two output reports (`FilteredSubgroupReportResult` and `BasicAggregateReportResult`). 

When it comes down to building and executing the SQL, it is one input query `OrganizationTypeQuery`, but we need to have two different outputs.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/BasicAggregateReportResult.java
```diff
@@ -1,53 +1,22 @@
...
+ * Aggregate report result for {@link BasicAggregateReportQuery} requests.
  */
-public class BasicAggregateReportResult extends AggregateReportResult<List<BasicAggregateRow>> {
+public class BasicAggregateReportResult extends CustomAggregateReportResult<BasicAggregateRow> {
 
     public static Builder builder() {
         return new Builder();
     }
 
-    public static class Builder extends AggregateReportResult.Builder<Builder, BasicAggregateReportResult, List<BasicAggregateRow>> {
```
agorina: Moved this code to the `CustomAggregateReportResult.class`.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/OrganizationTypeQuery.java
```diff
@@ -33,6 +33,12 @@
     private Set<String> iepCodes;
     private Set<String> economicDisadvantageCodes;
 
+    private String subgroupKey;
+
+    public String getSubgroupKey() {
+        return subgroupKey;
+    }
+
```
agorina: I do not like adding this, but I also did not like alternatives I could think of. 
Would like to know if somebody has a strong opinion.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractCustomAggregateReportRepository.java
```diff
@@ -179,4 +197,30 @@ private Organization findOneByIdAndType(final long id, final OrganizationType ty
...
+                                                                  final OrganizationTypeQuery query) {
+        final Map<ReportRowIdentity, T> reportTemplate = newHashMap();
+        // a template report consists of permutations of all the organizations and assessments
+        for (final Organization organization : queryWithReportTemplate.getOrganizations()) {
+            for (final ActiveAssessment assessment : queryWithReportTemplate.getAssessments()) {
+                for (final Dimension dimension : queryWithReportTemplate.getDimensions()) {
+                    final T row = buildReportRow(organization, assessment, dimension, emptyMeasure, query);
+                    reportTemplate.put(getRowIdentity(row), row);
+                }
+            }
```
agorina: Moved it here since an 'empty' row now depends on what output we are building.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultFilteredSubgroupAggregateService.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    protected Collection<OrganizationTypeQueryDescription> toOrganizationTypeQueries(final FilteredSubgroupReportQuery reportQuery) {
+        final List<OrganizationTypeQueryDescription> queries = newArrayList();
+
+        //each pair of organization/dimension is represented by a separate OrganizationTypeQuery
+        for (final String subgroupsKey : reportQuery.getSubgroups().keySet()) {
+            final OrganizationTypeQuery commonQuery = toQueryBuilderWithCommonAttributes(reportQuery, reportQuery.getSubgroups().get(subgroupsKey))
+                    .subgroupKey(subgroupsKey)
+                    .dimensionType(Overall)
```
agorina: From the repository point of view it is always an Overall dimension.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcBasicAggregateReportRepository.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    protected ReportRowIdentity getRowIdentity(final BasicAggregateRow row) {
+        return new BasicAggregateReportRowIdentity(row);
+    }
+
+    /**
+     * Provides an identity for the {@link BasicAggregateRow}, which is all the attributes but the {@link Measures}
+     */
+    class BasicAggregateReportRowIdentity extends ReportRowIdentity<BasicAggregateRow> {
```
wtritch: nit: static class?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractCustomAggregateReportRepository.java
```diff
@@ -179,4 +197,30 @@ private Organization findOneByIdAndType(final long id, final OrganizationType ty
...
+        }
+        return reportTemplate;
+    }
+
+    /**
+     * Represents an identity of a report row
+     */
+    abstract class ReportRowIdentity<I extends AggregateRow> {
```
wtritch: nit abstract static class?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcFilteredSubgroupAggregateReportRepository.java
```diff
@@ -0,0 +1,92 @@
...
+                .organization(organization)
+                .subgroupKey(query.getSubgroupKey())
+                .measures(measures)
+                .build();
+    }
+
+    /**
+     * Provides an identity for the {@link FilteredSubgroupRow}, which is all the attributes but the {@link Measures}
+     */
+    class FilteredSubgroupAggregateReportRowIdentity extends ReportRowIdentity<FilteredSubgroupRow> {
```
wtritch: same static class nit =)

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/FilteredSubgroupAggregateService.java
```diff
@@ -0,0 +1,15 @@
...
+import org.opentestsystem.rdw.reporting.common.model.FilteredSubgroupReportQuery;
+import org.opentestsystem.rdw.reporting.common.model.FilteredSubgroupRow;
+import org.springframework.security.access.prepost.PreAuthorize;
+
+import java.util.List;
+
+/**
+ * Implementation of this interface is responsible for retrieving the custom aggregate report with a {@link FilteredSubgroupReportQuery}
+ */
+@PreAuthorize("hasAuthority('PERM_CUSTOM_AGGREGATE_READ')")
```
wtritch: Hm... unfortunately, we don't currently respect these prefilters when the request is coming from rabbit (we don't create a security context while processing the message)

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/OrganizationTypeQuery.java
wtritch: I'm ok with it.  If you *really* want it to be generic you could just embed the original abstract query object in it, but that feels kind of messy...

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcBasicAggregateReportRepository.java
agorina: Will fix, in all three places.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/FilteredSubgroupAggregateService.java
agorina: But there is still an optional end point. So I think it should stay?

PR #1168, opened 2018-04-03
fixes display of custom analysis report view

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.ts
```diff
@@ -31,11 +29,6 @@ export class AggregateReportItem {
     }
   };
   organization: Organization;
-
-  // Present for Basic queries
-  dimension?: Dimension;
-
-  // Present for FilteredSubgroup queries
-  subgroup?: SubgroupFiltersListItem;
+  dimension: {id: string, name: string};
```
esotrope: decided to overload this column in the meantime to fix column sorting and grouping. not sure if this is the right long term solution though.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -529,7 +530,9 @@ export class AggregateReportTableComponent implements OnInit {
 
     return ordering(join(
       dimensionTypeAndCodeComparator,
-      enrolledGradeComparator));
+      enrolledGradeComparator,
+      ordering(byString).on(({ dimension }) => dimension.id).compare
```
esotrope: allows sorting to be based on subgroup id as a fallback

PR #1169, opened 2018-04-03
DWR-1627 Added more tests.

agorina:


agorina:
Once again, merging but will address comments later.

PR #1171, opened 2018-04-03
change the SAML attribute used for username; add test

malaffoon:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -31,19 +31,26 @@ public SamlUserDetailsService(@NotNull AuthorizationService authorizationService
...
 
-        final String username = credential.getNameID().getValue();
         final String id = credential.getAttributeAsString("sbacUUID");
         final String firstName = credential.getAttributeAsString("givenName");
         final String lastName = credential.getAttributeAsString("sn");
         final String email = credential.getAttributeAsString("mail");
         final String[] encodedGrants = credential.getAttributeAsStringArray("sbacTenancyChain");
 
+        // for SmarterBalanced SSO the user lookup value (aka username aka login) is configured to be "mail"
+        final String username = email;
```
malaffoon: This is the big change.

webapp/src/test/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+        final User user = (User) service.loadUserBySAML(credential);
+        assertThat(user.getUsername()).isEqualTo("alice@example.com");
+        assertThat(user.getPermissionsById()).isEmpty();
+        assertThat(user.getAuthorities()).isEmpty();
+        assertThat(user.isEnabled()).isFalse();
+    }
+
+    @Test
+    public void itShouldUseMailAndItIsOkayIfNameIdIsWrong() {
```
malaffoon: i visually verified the log message is being emitted

webapp/src/test/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+                createAttribute("givenName", firstName),
+                createAttribute("sn", lastName),
+                createAttribute("mail", username),
+                createAttribute("sbacTenancyChain", sbacTenancyChain)
+        );
+
+        return new SAMLCredential(nameID, assertion, "https://some.sso.provider.org/", attributes, "https://somewhere.org/");
+    }
+
+    private static Attribute createAttribute(final String name, final String value) {
```
malaffoon: mocking XMLObject stuff is annoying; anybody know a less obtuse way that doesn't involve reading XML payloads?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
esotrope: wondering... should this eventually be a tenant configuration?

esotrope: In reality so should saml lol

jtreuting: I don't think it should be because we are tied to the SB SSO in other ways.  We wouldn't be able to switch to another SSO system by simply pointing to it.  There is a custom tenancy chain that is used so I don't think being able to change this would allow any real flexibility.

agorina: My two cents are that we are getting to the point of a highly configurable and highly unmanageable system.  There is a world of possibilities that could happen...but do we really want to make everything configurable?

PR #1174, opened 2018-04-04
makes asmt null safe

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
```diff
@@ -88,7 +88,7 @@ public Subject getSubject() {
      * @return the assessment subject code
      */
     public String getSubjectCode() {
-        return subject.code();
+        return subject != null ? subject.code() : null;
```
esotrope: same q as in `InstructionalResources.java` these aren't nullable in the DB but they are null in some tests it seems. is it a real condition for these to be null? wondering if our builder should enforce notnull at build time.

malaffoon: Yeah, if it is tests only i would extend the builder with a tests builder that forces them to be non-null.

PR #1176, opened 2018-04-04
DO NOT MERGE - initial crack at a subject payload containing subject-…

wtritch:
…scoped messages.


common/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
     compile 'com.fasterxml.jackson.core:jackson-annotations'
     compile 'com.fasterxml.jackson.core:jackson-databind'
     compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-properties'
+    compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'
```
wtritch: I'm using jackson with mixins to map from models to XML

hand-jammed.math.subject.xml
```diff
@@ -0,0 +1,190 @@
+<Subject code="Math" name="Math">
```
wtritch: This is the hand-jammed XML design that I was working off of when creating the models and mixins

Math_subject.xml
```diff
@@ -0,0 +1,3429 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Subject code="Math" name="Math">
```
wtritch: This is the generated, saturated Math definition xml

Math_subject.xml
```diff
@@ -0,0 +1,3429 @@
...
+  <CommonCoreStandardDescription><![CDATA[This column reflects the Common Core State Standard(s) associated with a test item. Test items may be associated with more than one standard.  Items that assess mathematical practices (problem solving, modeling and data analysis, and communicating reasoning) are also aligned to standards that describe the mathematical content with which students engage. For example, an item may ask students to communicate their reasoning when solving a problem about geometry or fractions. For more information, see the <a href=\"https://portal.smarterbalanced.org/library/en/mathematics-content-specifications.pdf\" target=\"_blank\">Mathematics Content Specifications</a>.]]></CommonCoreStandardDescription>
```
wtritch: I believe this is just about the only varying text in the UI on a per-subject basis.  @esotrope ?  (This is displayed as a popup info message on the score column of some tables.)

Math_subject.xml
```diff
@@ -0,0 +1,3429 @@
...
+          <Icon>fa-book</Icon>
+        </SubScore>
+        <SubScore code="3" name="Reasoning">
+          <Icon>fa-question</Icon>
+        </SubScore>
+        <SubScore code="SOCK_2" name="PSMDA (Problem Solving and Modeling &amp; Data Analysis)">
+          <Icon>fa-edit</Icon>
+        </SubScore>
+      </SubScores>
+    </AssessmentType>
```
wtritch: I need some information scoped to an assessment type (ICAs have 4 overall performance levels, and 3 subscore performance levels, with no tertiary scores (up-comming target-based scoring)

Math_subject.xml
```diff
@@ -0,0 +1,3429 @@
...
+      <SubScorePerformanceLevels>
+        <SubScorePerformanceLevel level="1" name="Below Standard"/>
+        <SubScorePerformanceLevel level="2" name="Near Standard"/>
+        <SubScorePerformanceLevel level="3" name="Above Standard"/>
+      </SubScorePerformanceLevels>
+      <SubScores>
+        <SubScore code="1" name="Concepts">
+          <Icon>fa-book</Icon>
+          <TertiaryScores>
+            <TertiaryScore code="N-RN|A" name="Number and Quantities: Extend the properties of exponents to rational exponents."/>
```
wtritch: This is theoretical support for up-comming target-level scoring.  After much discussion, we decided that sub-scores are not tightly bound to claims (Math's SOCK_2 combines claims 2 and 4 I believe, and ELAs SOCK_R combines claims 1-IT and 1-LT.)  Since sub-scores are not tightly bound to claims, it seems unlikely that we can tightly bind tertiary scores to targets while still preserving hierarchical knowledge of which tertiary score belongs to which sub-score and how they relate to claims and targets.  Unfortunately, this leads to some funky duplication...

Math_subject.xml
```diff
@@ -0,0 +1,3429 @@
...
+        <SubScore code="3" name="Reasoning">
+          <Icon>fa-question</Icon>
+        </SubScore>
+        <SubScore code="SOCK_2" name="PSMDA (Problem Solving and Modeling &amp; Data Analysis)">
+          <Icon>fa-edit</Icon>
+        </SubScore>
+      </SubScores>
+    </AssessmentType>
+  </AssessmentTypes>
+  <ReportGrades>
```
wtritch: This element contains all of our PDF report translations.

Math_subject.xml
esotrope: Don't have an answer to your q but - If this is eventually html content i think we can multiline this.

Math_subject.xml
```diff
@@ -0,0 +1,3429 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Subject code="Math" name="Math">
+  <CommonCoreStandardDescription><![CDATA[This column reflects the Common Core State Standard(s) associated with a test item. Test items may be associated with more than one standard.  Items that assess mathematical practices (problem solving, modeling and data analysis, and communicating reasoning) are also aligned to standards that describe the mathematical content with which students engage. For example, an item may ask students to communicate their reasoning when solving a problem about geometry or fractions. For more information, see the <a href=\"https://portal.smarterbalanced.org/library/en/mathematics-content-specifications.pdf\" target=\"_blank\">Mathematics Content Specifications</a>.]]></CommonCoreStandardDescription>
+  <AssessmentTypes>
+    <AssessmentType code="ica">
+      <PerformanceLevels>
+        <PerformanceLevel level="1" name="Did Not Meet Standard"/>
+        <PerformanceLevel level="2" name="Nearly Met Standard"/>
+        <PerformanceLevel level="3" name="Met Standard" rollup="true"/>
```
esotrope: what is rollup? is that where the midpoint is when grouped?

Math_subject.xml
```diff
@@ -0,0 +1,2687 @@
...
+            <ReportPerformanceLevel level="2">{0} demonstrates some ability to solve well-posed mathematics problems by applying his or her knowledge of problem-solving skills and strategies. {0} also demonstrates some ability to analyze real-world problems, and can build and use mathematical models to interpret and solve problems.</ReportPerformanceLevel>
+            <ReportPerformanceLevel level="3">{0} demonstrates a thorough ability to consistently solve a variety of well-posed mathematics problems by applying his or her knowledge of problem-solving skills and strategies. {0} also demonstrates a strong ability to analyze real-world problems, and can build and use mathematical models to interpret and solve problems.</ReportPerformanceLevel>
+          </ReportPerformanceLevel>
+        </ReportClaim>
+      </ReportClaims>
+    </ReportGrade>
+  </ReportGrades>
+  <ItemClaims>
```
wtritch: This element contains all of the claims and their associated targets.  There's a fair bit of duplication among target descriptions, but enough variance that I can't see a viable pattern to reduce it...  Thoughts?

Math_subject.xml
wtritch: I suppose so, but that feels like a super-final just-before-publishing polish task.

Math_subject.xml
esotrope: So is this also where we define if it can be rolled up or not? is that implicit in the presence of this attribute? wondering if it doesn't make sense to pull it up to `<AssessmentType code="ica" performanceLevelGroupingCutpoint="3">`

Math_subject.xml
```diff
@@ -0,0 +1,2687 @@
...
+          <Icon>fa-question</Icon>
+        </SubScore>
+        <SubScore code="SOCK_2" name="PSMDA (Problem Solving and Modeling &amp; Data Analysis)">
+          <Icon>fa-edit</Icon>
+        </SubScore>
+      </SubScores>
+    </AssessmentType>
+    <AssessmentType code="iab">
+      <PerformanceLevels>
+        <PerformanceLevel level="1" name="Below Standard"/>
```
esotrope: if these are always a 1-based sequence should we just imply the level based on the element position? or is there some trouble that could cause later?

wtritch: I really dislike relying on order in XML docs... It's caused me issues in past lives.

esotrope: And just how many lives have you lived Will?

Math_subject.xml
```diff
@@ -0,0 +1,2687 @@
...
+        </ReportClaim>
+        <ReportClaim code="3">
+          <ReportPerformanceLevel>
+            <ReportPerformanceLevel level="1">{0} does not yet demonstrate the ability to put together valid arguments to support his or her own mathematical thinking or to critique the reasoning of others.</ReportPerformanceLevel>
+            <ReportPerformanceLevel level="2">{0} demonstrates some ability to clearly and precisely put together valid arguments to support his or her own mathematical thinking and to critique the reasoning of others.</ReportPerformanceLevel>
+            <ReportPerformanceLevel level="3">{0} demonstrates a thorough ability to clearly and precisely put together valid arguments to support his or her own mathematical thinking and to critique the reasoning of others.</ReportPerformanceLevel>
+          </ReportPerformanceLevel>
+        </ReportClaim>
+        <ReportClaim code="SOCK_2">
+          <ReportPerformanceLevel>
```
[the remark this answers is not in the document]
wtritch: doh, I was originally going to leave this unwrapped (but failed to add the annotation...) but I think you're right.  It's probably better to be consistent with container elements.

Math_subject.xml
wtritch: Hm... actually... will any subject ever rollup subscores?  If we move the rollup to the AssessmentType level it can only ever apply to the overall performance levels.  (I would hate to tack on a `subScorePerformanceLevelGroupingCutpoint="2"` attribute.)  @jtreuting ?

Math_subject.xml
```diff
@@ -0,0 +1,2687 @@
...
+  <CommonCoreStandardDescription><![CDATA[This column reflects the Common Core State Standard(s) associated with a test item. Test items may be associated with more than one standard.  Items that assess mathematical practices (problem solving, modeling and data analysis, and communicating reasoning) are also aligned to standards that describe the mathematical content with which students engage. For example, an item may ask students to communicate their reasoning when solving a problem about geometry or fractions. For more information, see the <a href=\"https://portal.smarterbalanced.org/library/en/mathematics-content-specifications.pdf\" target=\"_blank\">Mathematics Content Specifications</a>.]]></CommonCoreStandardDescription>
+  <AssessmentTypes>
+    <AssessmentType code="ica">
+      <PerformanceLevels>
+        <PerformanceLevel level="1" name="Did Not Meet Standard"/>
+        <PerformanceLevel level="2" name="Nearly Met Standard"/>
+        <PerformanceLevel level="3" name="Met Standard" rollup="true"/>
+        <PerformanceLevel level="4" name="Exceeded Standard"/>
+      </PerformanceLevels>
+      <SubScorePerformanceLevels>
```
wtritch: Assumption: All subscores have the same performance levels.

Math_subject.xml
```diff
@@ -0,0 +1,2687 @@
...
+            <TertiaryScore code="OA|D-1" name="Work with addition and subtraction equations."/>
+            <TertiaryScore code="NBT|D-2" name="Understand place  value."/>
+            <TertiaryScore code="OA|D-3" name="Solve problems involving the four operations, and identify and explain patterns in arithmetic."/>
+            <TertiaryScore code="NBT|D-4" name="Generalize place value understanding for multi-digit whole numbers."/>
+            <TertiaryScore code="NBT|D-5" name="Perform operations with multi-digit whole numbers and with decimals to hundredths."/>
+            <TertiaryScore code="NS|D-6" name="Apply and extend previous understandings of numbers to the system of rational numbers."/>
+            <TertiaryScore code="EE|D-7" name="Solve real-life and mathematical problems using numerical and algebraic expressions and equations."/>
+            <TertiaryScore code="EE|D-8" name="Analyze and solve linear equations and pairs of simultaneous linear equations."/>
+            <TertiaryScore code="OA|D-KG" name="Understand addition as putting together and adding to, and understand subtraction as taking apart and taking from."/>
```
wtritch: Question: Since some of these names can get a little long and contain punctuation should I convert this `@name` attribute into a `<Description>` element?  Should I do the same for the `<SubScore>` element?  Or are we missing a piece of data and the name should be "D-KG" (the target code) and the description should be moved to a `<Description>` element?

Math_subject.xml
wtritch: I'm on my 8th.  Still have 1 to spare!

ELA_subject.xml
```diff
@@ -0,0 +1,2156 @@
+<Subject code="ELA" name="ELA">
```
wtritch: Sample ELA payload

Math_subject.xml
wtritch: I decided to store 3 properties per TeriaryScore:
`code` used to map TRT scores to a particular tertiary score
`name` used in short labels: e.g. `Target X`
`description` used in long labels

Math_subject.xml
malaffoon: Also, as a rule, i don't like optional attributes in XML. How about making it a required attribute on the parent element, `<PerformanceLevels effectLevel='3'>`? I'm working on a better word than `rollup`: effect, accomplish, success?

Math_subject.xml
malaffoon: Shouldn't all the SubScore stuff be under a single element?
```
  <SubScoreXxx>
    <SubScorePerformanceLevels>...</>
    <SubScores>...</>
  </>
```

Math_subject.xml
```diff
@@ -0,0 +1,4678 @@
...
+      </PerformanceLevels>
+      <SubScorePerformanceLevels>
+        <SubScorePerformanceLevel level="1" name="Below Standard"/>
+        <SubScorePerformanceLevel level="2" name="Near Standard"/>
+        <SubScorePerformanceLevel level="3" name="Above Standard"/>
+      </SubScorePerformanceLevels>
+      <SubScores>
+        <SubScore code="1" name="Concepts">
+          <Icon>fa-book</Icon>
+          <TertiaryScores>
```
malaffoon: yech

It's more than just the inconsistency of Score, SubScore, TertiaryScore. I think TertiaryScore is meaningless. And, no, i'm not sure i have a better suggestion.

Math_subject.xml
```diff
@@ -0,0 +1,4678 @@
...
+            <ReportPerformanceLevel level="2">{0} demonstrates some ability to explain and apply mathematical concepts, and the ability to interpret and carry out mathematical procedures with ease and accuracy.</ReportPerformanceLevel>
+            <ReportPerformanceLevel level="3">{0} demonstrates a thorough ability to consistently explain and apply mathematical concepts, and the ability to interpret and carry out mathematical procedures with ease and accuracy.</ReportPerformanceLevel>
+          </ReportPerformanceLevels>
+        </ReportClaim>
+        <ReportClaim code="3">
+          <ReportPerformanceLevels>
+            <ReportPerformanceLevel level="1">{0} does not yet demonstrate the ability to put together valid arguments to support his or her own mathematical thinking or to critique the reasoning of others.</ReportPerformanceLevel>
+            <ReportPerformanceLevel level="2">{0} demonstrates some ability to clearly and precisely put together valid arguments to support his or her own mathematic
```
wtritch: I added standards down here.  For SBAC these represent Common Core Standards, but for Science? Social Studies?

Math_subject.xml
malaffoon: Aren't we using `Target`, `Standard` and `TertiaryScore` to describe basically the same type of things?

Math_subject.xml
wtritch: hm... but then wouldn't it still be optional?  Because IAB PerformanceLevels don't have a rollup?
As for the name... um... groupingCutpoint is the kind of the one I'm leading towards... I tried `goodNuff=true` but didn't get a lot of buy-in.

Math_subject.xml
wtritch: That's how I originally had it, until we got to talking about forbidding sub-scores on IABs, sub-sub-scores on Summatives, and the ability of Science-et-al to have different sub scores for ICA and Summative tests...

Math_subject.xml
wtritch: Are you talking about just re-using an element name, or some kind of fundamental restructuring of the xml?

Math_subject.xml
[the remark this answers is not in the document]
wtritch: Change <ReportPerformanceLevel> to be something like <Target> ?  Sorry, I don't see the "id" you mentioned...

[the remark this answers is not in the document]
wtritch: Correct me if I'm wrong, but what we get in the assessment package that defines an item is something like:
`Targets="1|AAP-R|X, 1|AAPR-R|A-1, etc" Standards="1.G.1, 1.G.2, etc"` ?
And you're suggesting we take that and convert it into a looking key where we know that `Targets` refer to something like 
`[Subject("ELA"), Standard("SBACTarget"), Code("1|AAP-R|X")]`
and `Standards` refer to something like
`[Subject("ELA"), Standard("CCS"), Code("1.G.1")]` ?
And could be backed by a table like:
`id, subject_id, standard_id, code, name, description` ?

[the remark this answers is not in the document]
wtritch: I'm still not sure SubScores TertiaryScores can play in the Standards playground with the arbitrary grouping/splitting of claims that we have, though...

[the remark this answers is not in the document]
wtritch: If we try to flatten our targets into Standards, how do we bind them to their parent claims?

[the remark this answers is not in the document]
wtritch: Ah, good call.  Changed.

Math_subject.xml
malaffoon: If it is that optional (good point) then there should be an optional element, something like `<GroupingCutpoint level='3'/>`?  I feel that we're not quite getting it right still.

Math_subject.xml
malaffoon: Yes  ;)

Math_subject.xml
jtreuting: that's because of the single quotes.  it would feel right with double quotes

Math_subject.xml
jtreuting: @wtritch i took @malaffoon to mean that the `<SubScore>` elements and `<SubScorePerformanceLevels>` are all related to sub scores and therefore should be grouped under a container.  Not that they should be moved from within the ICA assessment type

ELA_subject.xml
```diff
@@ -0,0 +1,4792 @@
...
+        <PerformanceLevel level="3" name="Met Standard" rollup="true"/>
+        <PerformanceLevel level="4" name="Exceeded Standard"/>
+      </PerformanceLevels>
+      <SubScorePerformanceLevels>
+        <SubScorePerformanceLevel level="1" name="Below Standard"/>
+        <SubScorePerformanceLevel level="2" name="Near Standard"/>
+        <SubScorePerformanceLevel level="3" name="Above Standard"/>
+      </SubScorePerformanceLevels>
+      <SubScores>
+        <SubScore code="4-CR" name="Research">
```
agorina: I saw a comment that we do not want to rely on the XML ordering. In this case, should the SubScores have an order defined?

wtritch: I don't think so, because they're uniquely identified by code (rather than by XML order)  We can order them however we like in the handler.

Math_subject.xml
wtritch: oh, duh.  I could see that.

ELA_subject.xml
agorina: I thought this defines Claims (in SB terminology). And I believe SB is very specific about claims ordering. I am not sure I understand what you mean under the handler ordering. How would the handler know the ordering?

wtritch: Hm... that's a good point... My objection to using XML ordering was the suggestion to remove the "level" attribute from the <PerformanceLevel level="1"> element since we could declare that the first element be bound to "1", the next to "2", etc.  I don't like that because it's an implicit binding of element order to identity.  As for determining display order from this package... maybe a displayOrder attribute would be useful?  Or maybe I'm off-base and using element order as an implicit definition of display ordering is viable?

agorina: I have had the same unpleasant experience with relying on the XML ordering and I would vote for an explicit attribute. 

This discussion raises another question: we are not going to support any updates to the  configured subject, right? Once it is loaded, that is it. Or am I wrong about this?

wtritch: I think we will *need* to support updates to allow for adding additional subject-scoped translations...

Math_subject.xml
wtritch: done, added a <SubScoring> container element.

wtritch:
Closing this in favor of a new PR for cleanliness.

PR #1181, opened 2018-04-05
created dimension-prefix for subgroup rows. Expanded ELAS -> English …

dsnyder-fairway:
…Language Acquisition Status on Advanced FIlters as there is enough room to provide clarity


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -423,6 +423,18 @@
       "Section504": "504",
       "StudentEnrolledGrade": "Student Enrolled Grade"
     },
+    "dimension-prefix": {
+      "EconomicDisadvantage": "Economic Disadvantage",
+      "Ethnicity": "Race / Ethnicity",
+      "Gender": "Gender",
+      "IEP": "IEP",
+      "LEP": "English Learners",
+      "ELAS": "ELAS",
```
dsnyder-fairway: Only thing different from `common.dimension`

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -57,7 +57,7 @@
       "assessment-grades-info": "The grade of the assessment that was administered",
       "assessment-grades-empty-error": "Please select an assessment grade.",
       "assessment-type-label": "Assessment Type",
-      "elas-label": "ELAS",
+      "elas-label": "English Language Acquisition Status",
```
dsnyder-fairway: expanding this provides a little more clarity on what `English Language Acquisition Status` under the subgroups is actually doing

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: I'd suggest we shorten the "Ethnicity" prefix. I'd vote for just "Ethnicity".

malaffoon: Perhaps the StudentEnrolledGrade could be shortened as well, to just "Enrolled Grade".

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -423,6 +423,18 @@
...
+      "Ethnicity": "Ethnicity",
+      "Gender": "Gender",
+      "IEP": "IEP",
+      "LEP": "English Learners",
+      "ELAS": "ELAS",
+      "MigrantStatus": "Migrant Status",
+      "Overall": "Overall",
+      "Section504": "504",
+      "StudentEnrolledGrade": "Enrolled Grade"
+    },
```
esotrope: so what is the semantic diff between "dimension" and "dimension-prefix" ? "prefix" doesn't tell me much

dsnyder-fairway: I agree. I'm open to suggestions. "subgroup-row-prefix" or "dimension-row-prefix" would be more appropriate. Thoughts?

esotrope: Hm. Is the intentio just to shorten it? Maybe we just call it `dimension-short-name` like the other stuff

esotrope: I'm wondering if our short displays shouldn't group too like:
Gender: [Male] [Female] Ethnicity: [White] [Asian]

dsnyder-fairway: I don't understand what you mean. Here's an example of the strings in use
screen shot 2018-04-05 at 11 37 15 am

Are you referencing the Report Review section?

PR #1183, opened 2018-04-05
Task/make assessment and subject code lookup dynamic in report processor

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ReportingJdbcAssessmentTypeRepository.java
```diff
@@ -0,0 +1,20 @@
+package org.opentestsystem.rdw.reporting.common.repository.impl;
+
+import org.opentestsystem.rdw.reporting.common.repository.AssessmentTypeRepository;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+public class ReportingJdbcAssessmentTypeRepository extends AbstractJdbcCodedEntityRepository implements AssessmentTypeRepository {
```
esotrope: not sure if this is how we want to do it but i made some common repos that will be shared between the report-processor and reporting-service that lookup entity code-id mappings. The underlying code will work in warehouse as well which is why i put the "Reporting" at the beginning of the repo name here in case we want to make a Warehouse version later

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamReportAssessmentType.java
```diff
@@ -1,32 +0,0 @@
-package org.opentestsystem.rdw.reporting.processor.model;
-
-import org.opentestsystem.rdw.common.model.AssessmentType;
-
-@Deprecated
-public enum ExamReportAssessmentType {
```
esotrope: don't need this anymore which is nice : j

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/AbstractJdbcCodedEntityRepository.java
```diff
@@ -0,0 +1,29 @@
...
+
+public abstract class AbstractJdbcCodedEntityRepository implements CodedEntityRepository {
+
+    private final NamedParameterJdbcTemplate template;
+
+    AbstractJdbcCodedEntityRepository(final NamedParameterJdbcTemplate template) {
+        this.template = template;
+    }
+
+    public Map<String, Integer> findAllIdsByCode() {
```
esotrope: the data structuring here looks like a service concern - i know we talked about this before but basically we have:
controller -> service -> repo -> repo-helper -> other repos 
should the repo helper be calling services? or should the code to ID resolution happen in the service layer?

malaffoon: I'm not quite sure i understand the question. In general a repo should call "up" to the service layer. Of course, you could have a helper that is called "service" that isn't really in the service layer. That said, what am i missing, i don't see this repo calling a service?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcIabReportRepository.java
```diff
@@ -33,13 +34,13 @@
...
 import static java.util.Comparator.comparing;
 import static org.opentestsystem.rdw.reporting.common.jdbc.QueryUtils.getNullable;
-import static org.opentestsystem.rdw.reporting.processor.util.ReportProcessorUtils.getExamFilterParameters;
 
 @Repository
 class JdbcIabReportRepository implements IabReportRepository {
 
     private final NamedParameterJdbcTemplate template;
     private final SecurityParameterProvider securityParameterProvider;
+    private final QueryParameterMapper queryParameterMapper;
```
malaffoon: Try saying that five times quickly  ;)

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/util/QueryParameterMapper.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Creates the query parameters for an {@link ExamQueryParams} instance.
+     *
+     * @param queryParameters The query parameters
+     * @return The SQL query parameters
+     */
+    public Map<String, Object> toParameters(final ExamQueryParams queryParameters) {
+        final Map<String, Object> parameters = new HashMap<>();
+        parameters.put("subject_id", subjectRepository.findAllIdsByCode().get(queryParameters.getSubjectCode()));
+        parameters.put("assessment_type", assessmentTypeRepository.findAllIdsByCode().get(queryParameters.getAssessmentTypeCode()));
```
malaffoon: Oooh. Is this what is bothering you? Yeah, this feels a bit wrong: one repo using a helper that uses another repo.

malaffoon: In aggregate service there are distinct query classes for the API and the repo. The service layer converts the code-based class to an id-based class. Seems like there are three (?) choices:
1. Same as aggregate; have two distinct classes with the service converting them.
2. Change the class to have id for subject and assessment type; UI would have to be aware of that change.
3. Make the class a hybrid with ids and codes for subject and assessment; then, like #1, the service would convert.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/requesthandler/SchoolGradeExamQueryParamsParser.java
```diff
@@ -26,8 +37,10 @@ public ExamQueryParams parseQueryParams(final AbstractExamReportRequest reportRe
                 .schoolYear(request.getSchoolYear())
                 .schoolId(request.getSchoolId())
                 .assessmentTypeCode(request.getAssessmentTypeCode())
+                .assessmentTypeId(assessmentTypeService.getIdsByCode().get(request.getAssessmentTypeCode()))
                 .gradeId(request.getGradeId())
                 .subjectCode(request.getSubjectCode())
+                .subjectId(subjectService.getIdsByCode().get(request.getSubjectCode()))
```
esotrope: changed mapping to be done in the service layer. this could possibly be done less redundantly but it works and we can add caching to the underlying services if we want later.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/util/QueryParameters.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Creates the query parameters for an {@link ExamQueryParams} instance.
+     *
+     * @param queryParameters The query parameters
+     * @return The SQL query parameters
+     */
+    public static Map<String, Object> toParameters(final ExamQueryParams queryParameters) {
+        final Map<String, Object> parameters = new HashMap<>();
+        parameters.put("subject_id", queryParameters.getSubjectId());
+        parameters.put("assessment_type", queryParameters.getAssessmentTypeId());
```
esotrope: should probably rename this param with _id

common/src/main/java/org/opentestsystem/rdw/reporting/common/codedentity/AbstractCodedEntityService.java
```diff
@@ -0,0 +1,20 @@
+package org.opentestsystem.rdw.reporting.common.codedentity;
```
esotrope: created a `codedentity` module under common that the aggregate-service could probably borrow a little from as well. I figure we can consolidate once we make the reporting-service form options provided by the backend.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/requesthandler/SchoolGradeExamQueryParamsParser.java
```diff
@@ -26,8 +37,10 @@ public ExamQueryParams parseQueryParams(final AbstractExamReportRequest reportRe
                 .schoolYear(request.getSchoolYear())
```
malaffoon: perhaps a copy() method on the builder or request?

esotrope:
any objections to merging this?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamQueryParams.java
```diff
@@ -5,7 +5,9 @@
  */
 public class ExamQueryParams {
     private String subjectCode;
+    private Integer subjectId;
     private String assessmentTypeCode;
+    private Integer assessmentTypeId;
```
wtritch: doh, why do we need ids here?

common/src/main/java/org/opentestsystem/rdw/reporting/common/codedentity/CodedEntityService.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Service responsible for providing entity code to ID mappings
+ */
+public interface CodedEntityService {
+
+    /**
+     * @return All entity code to ID mappings
+     */
+    Map<String, Integer> getIdsByCode();
```
wtritch: Should this be `@Cacheable`?

common/src/main/java/org/opentestsystem/rdw/reporting/common/codedentity/AbstractJdbcCodedEntityRepository.java
```diff
@@ -0,0 +1,27 @@
...
+
+public abstract class AbstractJdbcCodedEntityRepository implements CodedEntityRepository {
+
+    private final NamedParameterJdbcTemplate template;
+
+    AbstractJdbcCodedEntityRepository(final NamedParameterJdbcTemplate template) {
+        this.template = template;
+    }
+
+    public List<CodedEntity> findAll() {
```
wtritch: Should this be `@Cacheable` ?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/util/QueryParameters.java
wtritch: Why can't we just use code params?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamQueryParams.java
esotrope: been a while but i think it was because we create this object in the service layer and send it to the repo for lookups and the repo talks in IDs of course. there is a request model layer that is normalized into this, and it accepts only codes. The alternative it looked like was to do the code to id resolution in the repo layer which i was avoiding

agorina: The code is `varchar(10)`, and since this is for the teachers (= high volume usage) it would be better to use ids in the repository. 

This said, I think it highly depends on the complexity of the query and the repository could convert codes to ids. 

If we were to build REST API, I would strongly support using codes, but since we have agreed that our API serves the UI, I think that we should do what is more practical.

esotrope:
closing since this is out of date and we may revisit with a different solution

PR #1184, opened 2018-04-05
corrected createColumnProvider

dsnyder-fairway:
issue was parentColumnIndex was always 0 when the columnCount was MAX_VALUE
If we had the columnCount as 1 we would end up with 4 columns. If we had columnCount as 3 we would have 2 columns.

PR #1186, opened 2018-04-06
 went back to previous implementation for NarrowColumnProvider / WideColumnProvider

dsnyder-fairway:


dsnyder-fairway:
@jtreuting this addresses the single column issue with the following:
custom-agg-details-bug

The reasons for this change can be found here https://github.com/SmarterApp/RDW_Reporting/pull/1184

I wasn't able to figure out the correct way to implement this. I then realized this Provider was only used by two variables. I figured moving back to the previous implementation would be sufficient and readable rather than figuring out the correct implementation of the method

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -7,20 +7,10 @@ import { AssessmentDefinition } from './assessment/assessment-definition';
...
-    return parentColumns;
-  };
-};
-
-const NarrowColumnProvider: ColumnProvider = createColumnProvider(2);
-const WideColumnProvider: ColumnProvider = createColumnProvider();
+const NarrowColumnProvider: ColumnProvider = (organization, assessment, subgroup, filter) =>
+  [ [ organization, assessment ], [ subgroup, filter ] ];
+const WideColumnProvider: ColumnProvider = (organization, assessment, subgroup, filter) =>
+  [ [ organization ], [ assessment ], [ subgroup ], [ filter ] ];
```
esotrope: would at least like us to call these something arbitrary like "section[0]"

```typescript```
const NarrowColumnProvider: ColumnProvider = (...sections) =>
  [ [ sections[0], sections[1] ], [ sections[2], sections[3] ] ];
```

dsnyder-fairway: sounds good

PR #1189, opened 2018-04-06
order by id to maintain consistency between ordering on advanced filt…

dsnyder-fairway:
…ers and custom aggregate reports

Advanced filters
screen shot 2018-04-06 at 12 01 14 pm

Aggregate Reports
screen shot 2018-04-06 at 12 00 49 pm


dsnyder-fairway:
See dwr-1683 for my concerns about alphabetical ordering (changing language would alter the ordering)

dsnyder-fairway:
There is likely more work in aggregate-reports to figure out

PR #1193, opened 2018-04-10
Task/create button group with range constraint

esotrope:
Same as last PR but
1. moved radio functionality to sb-radio-group to simplify already overcomplicated sb-button-group
2. commented up code so hopefully it can be maintained easier - at some point we need to write a test suite for this

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
@@ -3,14 +3,9 @@ import { Forms } from './forms';
 import { AbstractControlValueAccessor } from './abstract-control-value-accessor';
 import * as _ from 'lodash';
 import { Utils } from '../support/support';
+import { Option } from './option';
 
-export type InputType = 'radio' | 'checkbox' | 'range';
-
-export interface Option {
-  readonly value: any;
-  readonly text?: string;
-  readonly analyticsProperties?: any;
-}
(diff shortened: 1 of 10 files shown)


webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -87,7 +87,7 @@ export class Utils {
   static toNgClassObject(value: any): any {
     switch (typeof value) {
       case 'string':
-        return value.split(/s+/g).reduce((object, key) => {
+        return value.split(/\s+/g).reduce((object, key) => {
```
esotrope: my component had class `btn-group- m` instead of `-sm` - doh

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -0,0 +1,435 @@
...
+  get value(): any {
+    return this._value;
+  }
+
+  set value(value: any) {
+    if (this._initialized) {
+      this.setValueAndNotifyChanges(value);
+      this._state = this.computeState(this._options, this._value);
+    } else {
+      this._initialValues = value;
```
esotrope: found that this pattern is needed when components use multiple input properties to compute internal state (options + value for this). Also found that triggering change events based on initial input properties is a bad idea and results in angular throwing the "the state has changed since we last checked" errors

esotrope:
going to push this through so i can put up my PR relying on it

PR #1195, opened 2018-04-11
remove advanced filters from scroll-nav when viewing Customize Subgroups

dsnyder-fairway:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -35,7 +35,8 @@ <h2>{{'aggregate-report-form.heading' | translate}}</h2>
       },
       {
         scrollTo: advancedFiltersSection,
-        text: 'aggregate-report-form.section.subgroup-filters-heading' | translate
+        text: 'aggregate-report-form.section.subgroup-filters-heading' | translate,
+        hidden: isAdvancedFiltersHidden()
```
esotrope: hm. i think we should probably just remove it from the nav entirely since it is joined at the hip with "Subgroup and Filter" section now.

dsnyder-fairway: done

PR #1196, opened 2018-04-11
adds "in-view" directive to lazy load slow rendering components

esotrope:
Kind of a detour but this was a much needed improvement i think

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -500,7 +500,6 @@
        id="previewSection"
        (inView)="onPreviewSectionInView($event)"
        [inViewInvalidator]="previewSectionViewInvalidator">
-       <!--[inViewValid]="holder.previewSectionValid">-->
 
     <div class="col-md-12 col-md-offset-2">
       <div class="well-group wide-content-container">
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -432,7 +432,12 @@
       </div>
     </div>
     <!-- Summary -->
-    <div class="row" #reviewSection id="reviewSection">
+    <div class="row"
+         #reviewSection
+         id="reviewSection"
+         (inView)="onReviewSectionInView($event)"
+         [inViewInvalidator]="reviewSectionViewInvalidator">
+
       <div class="col-md-8 col-md-offset-2">
         <div class="well-group">
           <div class="well">
@@ -490,7 +495,13 @@
 
   </form>
 
-  <div class="row" #previewSection id="previewSection">
+  <div class="row"
+       #previewSection
+       id="previewSection"
+       (inView)="onPreviewSectionInView($event)"
+       [inViewInvalidator]="previewSectionViewInvalidator">
+       <!--[inViewValid]="holder.previewSectionValid">-->
+
     <div class="col-md-12 col-md-offset-2">
       <div class="well-group wide-content-container">
         <div class="well inline-headings">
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -432,7 +432,12 @@ <h3>{{'aggregate-report-form.section.subgroup-filters-heading' | translate}}</h3
       </div>
     </div>
     <!-- Summary -->
-    <div class="row" #reviewSection id="reviewSection">
+    <div class="row"
+         #reviewSection
+         id="reviewSection"
+         (inView)="onReviewSectionInView($event)"
+         [inViewInvalidator]="reviewSectionViewInvalidator">
```
esotrope: listen to `(inView)` for when a decorated element comes into view or is in view at init time.
this will only fire once unless the view is manually invalidated by pushing events to the `[inViewInvalidator]` observable.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -209,14 +213,6 @@ export class AggregateReportFormComponent {
         { messageId: 'aggregate-report-form.field.report-name-file-name-error' }
       ))
     });
-
-    Observable.create((observer) => this.settingsChangedObserver = observer)
-      .pipe(debounceTime(DefaultRenderDebounceMilliseconds))
-      .subscribe(() => this.applySettingsChange());
```
esotrope: removed debouncing logic which can still cause hang while updating the form.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -393,15 +389,41 @@ export class AggregateReportFormComponent {
...
     this.estimatedRowCount = undefined;
-
-    // informs the debounced observable that the view has changed
-    this.settingsChangedObserver.next(undefined);
+    if (this.reviewSectionInvalid) {
+      this.reviewSectionInvalid.next(undefined);
+    }
+    if (this.previewSectionInvalid) {
+      this.previewSectionInvalid.next(undefined);
+    }
```
esotrope: inform each view independently that they need to re-render when the come into view again

webapp/src/main/webapp/src/app/shared/nav/in-view.directive.ts
```diff
@@ -0,0 +1,72 @@
...
+    this._invalidatorSubscription = observable.subscribe(() => {
+      this.valid = false;
+    });
+  }
+
+  get valid(): boolean {
+    return this._valid;
+  }
+
+  set valid(value: boolean) {
```
esotrope: introduced the invalidator observable because unfortunately binding a property directly to this property bypassed angular's change detection.

webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -149,6 +149,25 @@ export class Utils {
...
+  static inView(element: Element, windowReference?: Window): boolean {
+    if (element == null) {
+      return false;
+    }
+    const window = windowReference || window;
+    const bounds = element.getBoundingClientRect();
+    return bounds.bottom > 0
+      && bounds.right > 0
+      && bounds.left < (window.innerWidth || document.documentElement.clientWidth)
+      && bounds.top < (window.innerHeight || document.documentElement.clientHeight);
```
jtreuting: without seeing it in practice it's hard to tell but do we want a little buffer so that it starts loading the view when it gets really close but still not quite there?

esotrope: possibly. i figure we can put this in and refine it as we go. wouldn't be too hard to have this method accept some margin params if we need it

PR #1200, opened 2018-04-11
makes longitudinal reports available for summatives only in the ui

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -95,10 +95,12 @@ export class AggregateReportRequestMapper {
...
 
     // Set report type specific parameters
-    if (settings.reportType === 'GeneralPopulation') {
+    // The assessment definition check is tacked on because the form state can be set to longitudinal cohort
+    // and then the assessment definition can be changed to a type that does not support longitudinal cohort
+    if (settings.reportType === 'GeneralPopulation' || !assessmentDefinition.aggregateReportLongitudinalCohortEnabled) {
       query.assessmentGradeCodes = settings.generalPopulation.assessmentGrades;
       query.schoolYears = settings.generalPopulation.schoolYears;
-    } else if (settings.reportType === 'LongitudinalCohort') {
+    } else if (settings.reportType === 'LongitudinalCohort' && assessmentDefinition.aggregateReportLongitudinalCohortEnabled) {
```
jtreuting: so if they select longitudinal cohort and then switch from summative to ICA, they would essentially get a general population report?  are we blocking them from submitting that query anywhere?

esotrope: yes and the idea is that we don't block them we just ignore their longitudinal report settings and use their general population settings.

esotrope: this is the approach we use for the varying identity columns with IABs.
i could also see us keeping the report type toggle and just disabling the longitudinal option too

PR #1202, opened 2018-04-12
show aggregate writing trait score view for ICAs - not summative beca…

jtreuting:
…use it's item level data

image

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
@@ -126,7 +126,7 @@
                 <li role="menuitem"
                     *ngIf="writingTraitScoresView.display">
                   <a class="dropdown-item"
-                     popover="{{ (!assessmentExam.assessment.isIab ? 'assessment-results.no-writing-trait-scores-for-summative-ica-exams' : 'assessment-results.no-writing-trait-scores') | translate }}"
+                     popover="{{ (writingTraitScoresView.disabled ? 'assessment-results.writing-trait-scores-disabled' : 'assessment-results.no-writing-trait-scores') | translate }}"
                      triggers="{{ !writingTraitScoresView.disabled ? '' : 'mouseenter:mouseleave'}}"
                      placement="right"
                      container="body"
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -126,7 +126,7 @@
                 <li role="menuitem"
                     *ngIf="writingTraitScoresView.display">
                   <a class="dropdown-item"
-                     popover="{{ (!assessmentExam.assessment.isIab ? 'assessment-results.no-writing-trait-scores-for-summative-ica-exams' : 'assessment-results.no-writing-trait-scores') | translate }}"
+                     popover="{{ (writingTraitScoresView.disabled ? 'assessment-results.writing-trait-scores-disabled' : 'assessment-results.no-writing-trait-scores') | translate }}"
```
jtreuting: changed to be tied to if this is disabled or not which before was looking at the IAB status.  so this seemed more logical.  basically it is disabled if it is summative since we don't show item level details for summative.  so changed the translation placeholder to better align

esotrope: nit: would prefer the boolean be modeled in positive form (e.g. enabled, visible, has, is...)

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -159,7 +159,7 @@ export class AssessmentResultsComponent implements OnInit {
   }
 
   get enableWritingTraitScores(): boolean {
-    return this._assessmentExam.assessment.isIab && this.displayItemLevelData;
+    return this.displayItemLevelData;
```
esotrope: so is the idea here to just make item level data display an all or nothing? or will this ever be determined by assessment/subject combination?

jtreuting: previously we were restricting access to only IABs, but in general we don;'t want to show this breakdown because it has item level data and we never show that for summatives independent of the writing trait score concept

esotrope: gotcha, so the "displayItemLevelData" assessment/subject config trumps whether or not writing trait scores are available/enabled or not

jtreuting: correct

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
jtreuting: since that is a shared structure across all the different views and outside this particular change i'll create a tech debt story for it

PR #1203, opened 2018-04-12
corrected estimate row count for custom subgroups

dsnyder-fairway:
Subgroups and advanced filters:

screen shot 2018-04-12 at 10 59 09 am

Customize Subgroups:

screen shot 2018-04-12 at 11 00 43 am

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
@@ -1,6 +1,7 @@
 package org.opentestsystem.rdw.olap.web;
 
 import org.opentestsystem.rdw.olap.service.AggregateQueryHandlerProvider;
+import org.opentestsystem.rdw.reporting.common.model.AggregateQuery;
 import org.opentestsystem.rdw.reporting.common.model.BasicAggregateReportQuery;
 import org.opentestsystem.rdw.reporting.common.web.security.User;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -27,7 +28,7 @@ class CustomAggregateController {
     @PostMapping("/estimatedRowCount")
     public int postEstimateReportRowCount(
             @AuthenticationPrincipal final User user,
-            @RequestBody final BasicAggregateReportQuery query) {
+            @RequestBody final AggregateQuery query) {
         return handlerProvider.findHandler(query).estimateReportRowCount(user, query);
     }
 }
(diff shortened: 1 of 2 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
```diff
@@ -27,7 +28,7 @@
     @PostMapping("/estimatedRowCount")
     public int postEstimateReportRowCount(
             @AuthenticationPrincipal final User user,
-            @RequestBody final BasicAggregateReportQuery query) {
+            @RequestBody final AggregateQuery query) {
```
dsnyder-fairway: @agorina I think this was all that was needed

agorina: Does it 'magically' know how to polymorphically deserialize the request?

dsnyder-fairway: it looks like it. At least from testing it out locally

agorina:
Is 15,912 the right count for your selection? I see only two subgroups + Overall.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
agorina: I am still curious to understand how Spring magic works in this case.

esotrope: What is this magic? i would like to know how jackson is figuring this out without configuration. it was my understanding that it has to look at the report request queryType field to know what to deserialize the query to.

dsnyder-fairway: my guess would be that it sees the Subgroup array on the json

esotrope: hm, i guess if we have unit tests we should be good here. i would just make sure that when we serialize in our unit test we dont include type information. because the queries we send over the wire do not have the java class type info included. can you make sure this is the case?

dsnyder-fairway: I'll debug and/or look into expanding our tests shortly

PR #1205, opened 2018-04-12
corrected IABs having performance levels like "Below Standard Standar…

dsnyder-fairway:
…d" on custom aggregate reports

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
@@ -121,7 +121,7 @@ export class AggregateReportTableExportService {
       if (options.performanceLevelDisplayType === PerformanceLevelDisplayTypes.Grouped) {
         header = this.translateService.instant(`aggregate-report-table.columns.grouped-performance-level-prefix.${level}`);
       } else {
-        header = this.translateService.instant(`common.assessment-type.${options.assessmentDefinition.typeCode}.performance-level.${level}.short-name`);
+        header = this.translateService.instant(`common.assessment-type.${options.assessmentDefinition.typeCode}.performance-level.${level}.name-prefix`);
       }
       return header + ' ' + this.translateService.instant('aggregate-report-table.columns.performance-level-suffix');
     };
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
```diff
@@ -121,7 +121,7 @@ export class AggregateReportTableExportService {
       if (options.performanceLevelDisplayType === PerformanceLevelDisplayTypes.Grouped) {
         header = this.translateService.instant(`aggregate-report-table.columns.grouped-performance-level-prefix.${level}`);
       } else {
-        header = this.translateService.instant(`common.assessment-type.${options.assessmentDefinition.typeCode}.performance-level.${level}.short-name`);
+        header = this.translateService.instant(`common.assessment-type.${options.assessmentDefinition.typeCode}.performance-level.${level}.name-prefix`);
       }
       return header + ' ' + this.translateService.instant('aggregate-report-table.columns.performance-level-suffix');
```
dsnyder-fairway: notice this appends the string of prefix + suffix. `aggregate-report-table.component.html` does something similar when the value is pulled from `getPerformanceLevelColumnHeaderTranslationCode` in the `aggregate-report-table.component.ts`

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -310,14 +310,17 @@
         "performance-level": {
           "1": {
             "name": "Below Standard",
+            "name-prefix": "Below",
             "short-name": "Below Standard"
```
jtreuting: Can you see where we currently use `short-name`?  cause it seems like it works to use short name for ICA level s but not IABs, so i'm wondering if we should change short-name for IABs to "Below" but am wondering what that will impact.

dsnyder-fairway: `short-name` is used in a variety of places, including the School / Group assessments, CSV export, etc. It looks a bit strange to just have "Near" appear in the Student Score Distribution

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -310,14 +310,17 @@
...
           },
           "2": {
             "name": "Near Standard",
+            "name-prefix": "Near",
```
esotrope: you didn't want to go with `even-shorter-name`?

PR #1206, opened 2018-04-13
API for IAB dashboards

dsnyder-fairway:


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupController.java
```diff
@@ -23,4 +24,9 @@
         return service.getGroups(user);
     }
 
+    @GetMapping("/{groupId}")
+    public Group getGroup(@AuthenticationPrincipal final User user, @PathVariable final int groupId) {
```
dsnyder-fairway: will be used with a `forkJoin` to get the total number of students in the group

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/AssessmentItems.java
```diff
@@ -1,4 +1,6 @@
-package org.opentestsystem.rdw.reporting.common.model;
+package org.opentestsystem.rdw.reporting.common.jdbc;
```
dsnyder-fairway: moved some of these to their appropriate package. From org.opentestsystem.rdw.reporting.common.model -> org.opentestsystem.rdw.reporting.common.jdbc

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupController.java
jtreuting: we don't already have the groups on the client side?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/GroupAssessmentController.java
```diff
@@ -38,4 +39,12 @@ public GroupAssessment getLatestGroupAssessment(
         return service.getAllAssessments(user, search);
     }
 
+    @GetMapping("/groups/{groupId}/measuredassessments")
+    public List<MeasuredAssessment> getMeasuredAssessments(
+            @AuthenticationPrincipal final User user,
+            final GroupAssessmentSearch search) {
```
jtreuting: is this search mapping from the querystring?  also, the feature does not need the ability for advanced filters, does the API and the SQL support that right now?

dsnyder-fairway: the SQL only needs groupId and schoolYear, but that is all the GroupAssessmentSearch contains

jtreuting: okay cool.  thanks for clarifying

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupController.java
dsnyder-fairway: we do the call above to get the list of groups on the main page. I suppose we could use the student count from there instead of this endpoint when we want the the number of students in the group.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/group/JdbcGroupAssessmentRepositoryIT.java
```diff
@@ -37,19 +39,30 @@
...
     @Test
     public void findAllShouldFindAll() {
         assertThat(repository.findAll(userBuilder.permissionsById(groupStatewide).build(), search().build()))
                 .usingRecursiveFieldByFieldElementComparator()
                 .containsOnlyElementsOf(TestData.STUDENT10_GROUP10_ASSESSMENTS_WITH_CUT_POINTS);
+
+        final List<MeasuredAssessment> measuredAssessments = repository.findAllMeasuredAssessments(userBuilder.permissionsById(groupStatewide).build(), search().build());
+        assertThat(measuredAssessments)
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsOnlyElementsOf(TestData.STUDENT10_MEASURED_ASSESSMENTS);
```
agorina: I expected more test cases for this call.  Do you feel that it covers all possible use cases?

dsnyder-fairway: Maybe.. This test executes the same query as the `findAll` above it. That `findAll` contains an IAB, ICA, and Summative assessment, but this one contains only the IAB. The other test verifies no assessments are found. Am I missing a test to verify a test with more than one level count? What other tests should there be?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/GroupAssessmentRepository.java
```diff
@@ -1,6 +1,23 @@
...
+    /**
+     * Finds all IAB measured assessments for the given group, assessment grade, school year and permissions
+     *
+     * @param permissionSource the security information to prepare the SQL query parameters for
+     * @param search           the search parameters to limit the results by
+     * @return all IAB measured assessments for the given group, assessment grade, school year and permissions
+     * If the search includes a student group id and {@link PermissionSource} does not have access to the group,
+     * it returns an empty list.
+     */
+    List<MeasuredAssessment> findAllMeasuredAssessments(@NotNull final PermissionSource permissionSource, @NotNull final GroupAssessmentSearch search);
```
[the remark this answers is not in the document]
esotrope: should that limit be imposed in the query for now?
then later our assessment/subject service can inform the UI that "measuredAssessmentTypes" is "iab" only or something

reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -10,12 +10,32 @@ sql:
         sg.name,
         sc.name as school_name,
         sc.id as school_id,
-        s.code as subject_code
+        s.code as subject_code,
+        count(sgm.student_id) as students
```
agorina: Nit: it should be `count(*)`, here and below.

And I wonder, why this change? Do we need to return counts for all user groups somewhere?

reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -10,12 +10,32 @@ sql:
...
       from user_student_group usg
         join student_group sg on sg.id = usg.student_group_id
         join school sc on sc.id = sg.school_id
+        join student_group_membership sgm on sgm.student_group_id = usg.student_group_id
```
esotrope: so if i follow this right - we get the count for the groups on the homepage and just dont display it. we cache the groups with counts and then we look it up again on the individual group page? i'm leaning towards the individual group page getting all the groups and using the counts from there - that page will need to show a select box with all groups anyway right?

reporting-service/src/main/resources/application-no-permission.sql.yml
dsnyder-fairway: yeah this was a little unclear to me the best way to go about this. We could either do this when we get all the groups and just hold onto the counts in cache, or we can do this on every group lookup like `findOneById`. Thoughts? @esotrope @jtreuting conversation will follow here for student count on groups

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/group/JdbcGroupAssessmentRepositoryIT.java
agorina: What about verifying things like:
1. If a student took an asmt twice - it takes the latests one within the school year
2. It returns the right completed_at
3. I just noticed that we do not filter unscored exams. I missed it in the SQL. I am not sure how we want to handle it, but we need to consider that.  Let me know if you need help with adjusting the SQL. 
4. and more ... 

There are several conditional paths in that SQL and the idea for the repository IT is to test all the cases. I am sure my list above is not complete.

reporting-service/src/main/resources/application-no-permission.sql.yml
agorina: My two cents: if we do not display the counts - do not count in MySQL.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/group/JdbcGroupAssessmentRepositoryIT.java
dsnyder-fairway: Thanks @agorina I'll close this PR and work on these

reporting-service/src/main/resources/application-no-permission.sql.yml
dsnyder-fairway: but we will need the group size to display on the IAB pages

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/group/JdbcGroupAssessmentRepositoryIT.java
agorina: No need to close. You could add the tests later. Up to you.

reporting-service/src/main/resources/application-no-permission.sql.yml
agorina: One group or all groups?

dsnyder-fairway: Either X times for a using going through X number of groups (`findOneById`) or once (`findAllForUsername`). But I did both here 😆

jtreuting: it feels to me that we are getting all the groups to populate that dropdown.  so it seems logical to do it at that point and have the counts already.  BUT to Alla's point for performance and load testing reasons since we don't know how often someone will go to the dashboard page and it's not the default maybe we should just call the findOneById and get the count there so it's only calculated when we actually need it.

that is a long rambling way of suggesting the `findOneById` approach and not calculating it for all groups.  let's take the hit when we know we need to.  

P.S @agorina is scared of counting things apparently :)

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/MeasuredAssessment.java
```diff
@@ -0,0 +1,82 @@
...
+
+    }
+
+    public static class Builder {
+        private Assessment assessment;
+        private Measures measures;
+        private Integer studentsTested;
+        private Instant completedAt;
+
+        public Builder() {
```
malaffoon: shouldn't expose the constructor

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/MeasuredAssessment.java
```diff
@@ -0,0 +1,82 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import java.time.Instant;
+
+public class MeasuredAssessment {
```
malaffoon: javadocs, what's a measured assessment?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupRepository.java
```diff
@@ -15,4 +15,12 @@
...
     List<Group> findAllForUsername(String username);
 
+    /**
+     * Find the {@link Group} by group id for the given user name
+     *
+     * @param username the user name
+     * @param groupId the group id
+     * @return a {@link Group}
+     */
+    Group findOneById(String username, int groupId);
```
malaffoon: Bit confused by this: doesn't the groupId fully constrain a group? So this will return null if the user isn't a member of the group? Something feels wrong about that. If we do leave it this way then the method name should probably be findOneByUsernameAndId.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupRepository.java
```diff
@@ -15,4 +15,12 @@
      */
     List<Group> findAllForUsername(String username);
 
+    /**
+     * Find the {@link Group} by group id for the given user name
+     *
+     * @param username the user name
+     * @param groupId the group id
+     * @return a {@link Group}
```
malaffoon: null if not found?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupService.java
```diff
@@ -18,4 +18,12 @@
      */
     Set<Group> getGroups(@NotNull User user);
 
+    /**
+     * Gets the student group by group id that the user has access to
+     *
+     * @param user the {@link User}
+     * @param groupId the group id
+     * @return the group by group id the user has access to
```
malaffoon: throws if not found?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/MeasuredAssessment.java
dsnyder-fairway: oops nice catch

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -99,6 +99,55 @@ sql:
           AND (sg.subject_id is null OR a.subject_id = sg.subject_id)
           AND ${sql.reporting.snippet.studentExamPermissionWithEmbargo}
 
+    measuredassessment:
```
malaffoon: nit: `measuredAssessment`?

dsnyder-fairway: done

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupService.java
dsnyder-fairway: yup

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupRepository.java
dsnyder-fairway: yup

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/GroupRepository.java
dsnyder-fairway: renamed

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/MeasuredAssessment.java
dsnyder-fairway: added

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/MeasuredAssessment.java
dsnyder-fairway: made private

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/exam/group/JdbcGroupAssessmentRepositoryIT.java
agorina: @jtreuting , could you please clarify item 3? 
Should we exclude exams that have null scores from the query?

As for the rest, we could look at it together on Monday if you still have issues with it.

jtreuting: null scores should not be included.

agorina: Hmmmm...this would probably require a change to the index too. Will work on it.

dsnyder-fairway:
I'll re-open this once the issues have been resolved

PR #1210, opened 2018-04-17
Feature/dwr 1674

dsnyder-fairway:
API for group dashboard


webapp/src/main/webapp/src/app/assessments/model/measured-assessment.model.ts
```diff
@@ -0,0 +1,12 @@
+import { Assessment } from './assessment.model';
```
dsnyder-fairway: not used in this PR, but will be used shortly

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
```diff
@@ -9,6 +9,7 @@
     private String schoolName;
     private int schoolId;
     private String subjectCode;
+    private int studentCount;
```
esotrope: do we always populate this? or can it be null? if null let's make this type `Integer` so jackson can omit it from the json payload

webapp/src/main/webapp/src/app/assessments/model/measured-assessment.model.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class MeasuredAssessment {
```
esotrope: would make this an interface with readonly fields.

webapp/src/main/webapp/src/app/assessments/model/measured-assessment.model.ts
```diff
@@ -0,0 +1,12 @@
...
+  assessment: Assessment;
+  level1Count: number;
+  level2Count: number;
+  level3Count: number;
+  avgScaleScore: number;
+  avgStdErr: number;
```
esotrope: nit: abbr

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
dsnyder-fairway: yep this can be null. I'll change this

webapp/src/main/webapp/src/app/assessments/model/measured-assessment.model.ts
esotrope: why not a list?
`studentCountByPerformanceLevel: number[]`

webapp/src/main/webapp/src/app/assessments/model/measured-assessment.model.ts
```diff
@@ -0,0 +1,12 @@
...
+  studentsTested: number;
```
esotrope: seems like a prop that can be computed by looking at the level counts no?

webapp/src/main/webapp/src/app/assessments/model/measured-assessment.model.ts
dsnyder-fairway: removed this class for now

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
jtreuting: why would this be null?  isn't a count 0 if there are no students?

esotrope: its an amorphic model, and we didn't intend on always running count() to populate this value right?

dsnyder-fairway: correct, this is what was done for removing some changes to sql that weren't needed. See https://github.com/SmarterApp/RDW_Reporting/pull/1206#discussion_r181509250

jtreuting: okay, that makes sense then.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/MeasuredAssessment.java
```diff
@@ -0,0 +1,85 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import java.time.Instant;
+
+/**
+ * Represents an {@link Assessment} along with its {@link Measures} and number of students tested
+ */
+public class MeasuredAssessment {
+
+    private MeasuredAssessment() {
```
malaffoon: nit: should move this down (line 18ish)

dsnyder-fairway:
@agorina @malaffoon @esotrope look good? I'll merge if you think this is fine

PR #1212, opened 2018-04-18
adds aggregate assessment service with cut point endpoint

esotrope:
straw man for how we can get the assessment cut point info over each grade year for aggregates.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment.service.ts
@@ -11,17 +11,30 @@ export class AssessmentService {
   constructor(private dataService: DataService) {
   }
 
-  getCutPoints(request: GradeYearSearch): Observable<AssessmentGradeYearCutPoints> {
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment.service.ts
```diff
@@ -0,0 +1,38 @@
...
+          year: serverPoints.gradeYear.schoolYear
+        },
+        cutPoints: serverPoints.cutPoints
+      }))
+    );
+  }
+
+}
+
+export interface GradeYearSearch {
```
esotrope: this is the same model as the current longitudinal report request. The idea here is that this request should share the format so it can be requested using the report query object held in the report object.

Some questions i had when writing this:
1. Should we be requesting this separate to the report payload? or should it be part of the payload?
2. Can there be multiple assessments for a grade/year? If so what do we do?

esotrope: good point. i think i'll break it down by subject code

agorina: About question 1. I think it maybe related to what we are going to output in the CSV export for this report. Are cut points part of the output?

PR #1217, opened 2018-04-18
Feature/dwr 1675 UI for IAB cards

dsnyder-fairway:
a typical group with URL

screen shot 2018-04-18 at 4 03 26 pm

how to access this view

screen shot 2018-04-18 at 4 03 57 pm

screen shot 2018-04-18 at 4 04 10 pm

viewing math 

screen shot 2018-04-18 at 4 04 40 pm

selecting some cards

screen shot 2018-04-18 at 4 04 48 pm

### DIFF
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultExamFilterOptionService.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultExamFilterOptionService.java
@@ -5,6 +5,7 @@ import org.opentestsystem.rdw.reporting.elas.ElasService;
 import org.opentestsystem.rdw.reporting.ethnicity.EthnicityService;
 import org.opentestsystem.rdw.reporting.gender.GenderService;
 import org.opentestsystem.rdw.reporting.schoolyear.SchoolYearService;
+import org.opentestsystem.rdw.reporting.subject.SubjectService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
@@ -17,17 +18,20 @@ public class DefaultExamFilterOptionService implements ExamFilterOptionService {
     private final EthnicityService ethnicityService;
     private final GenderService genderService;
     private final ElasService elasService;
+    private final SubjectService subjectService;
 
     @Autowired
     public DefaultExamFilterOptionService(
             @NotNull final SchoolYearService schoolYearService,
             @NotNull final EthnicityService ethnicityService,
             @NotNull final GenderService genderService,
-            @NotNull final ElasService elasService) {
+            @NotNull final ElasService elasService,
+            @NotNull final SubjectService subjectService) {
         this.schoolYearService = schoolYearService;
         this.ethnicityService = ethnicityService;
         this.genderService = genderService;
         this.elasService = elasService;
+        this.subjectService = subjectService;
     }
 
     @Override
@@ -37,6 +41,7 @@ public class DefaultExamFilterOptionService implements ExamFilterOptionService {
                 .ethnicities(ethnicityService.getEthnicityCodes())
                 .genders(genderService.getGenders())
                 .elasCodes(elasService.getElases())
+                .subjects(subjectService.getSubjects())
                 .build();
     }
 
--- a/webapp/src/main/webapp/src/app/app.routes.ts
+++ b/webapp/src/main/webapp/src/app/app.routes.ts
@@ -1,42 +1,43 @@
-import { HomeComponent } from "./home/home.component";
-import { Routes } from "@angular/router";
-import { GroupResultsComponent } from "./groups/results/group-results.component";
-import { GroupAssessmentsResolve } from "./groups/results/group-assessments.resolve";
-import { SchoolAssessmentResolve } from "./school-grade/results/school-assessments.resolve";
-import { SchoolResultsComponent } from "./school-grade/results/school-results.component";
-import { CurrentSchoolResolve } from "./school-grade/results/current-school.resolve";
-import { StudentResultsComponent } from "./student/results/student-results.component";
-import { StudentExamHistoryResolve } from "./student/results/student-exam-history.resolve";
-import { StudentResponsesResolve } from "./student/responses/student-responses.resolve";
-import { StudentResponsesComponent } from "./student/responses/student-responses.component";
-import { TranslateResolve } from "./home/translate.resolve";
-import { StudentHistoryResponsesExamResolve } from "./student/responses/student-history-responses-exam.resolve";
-import { StudentHistoryResponsesAssessmentResolve } from "./student/responses/student-history-responses-assessment.resolve";
-import { StudentHistoryResponsesStudentResolve } from "./student/responses/student-history-responses-student.resolve";
-import { ReportsResolve } from "./report/reports.resolve";
-import { ReportsComponent } from "./report/reports.component";
-import { ErrorComponent } from "./error/error.component";
-import { AccessDeniedComponent } from "./error/access-denied/access-denied.component";
-import { OrganizationExportComponent } from "./organization-export/organization-export.component";
-import { UserOrganizationsResolve } from "./organization-export/organization/user-organizations.resolve";
-import { InstructionalResourceComponent } from "./admin/instructional-resource/instructional-resource.component";
-import { EmbargoComponent } from "./admin/embargo/embargo.component";
-import { EmbargoResolve } from "./admin/embargo/embargo.resolve";
-import { ImportHistoryComponent } from "./admin/groups/import/history/import-history.component";
-import { FileFormatComponent } from "./admin/groups/import/fileformat/file-format.component";
-import { ImportHistoryResolve } from "./admin/groups/import/history/import-history.resolve";
-import { GroupImportComponent } from "./admin/groups/import/group-import.component";
-import { GroupImportDeactivateGuard } from "./admin/groups/import/group-import.deactivate";
-import { GroupsComponent } from "./admin/groups/groups.component";
-import { SessionExpiredComponent } from "./shared/security/session-expired.component";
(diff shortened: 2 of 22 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultExamFilterOptionService.java
```diff
@@ -17,17 +18,20 @@
...
+    private final SubjectService subjectService;
 
     @Autowired
     public DefaultExamFilterOptionService(
             @NotNull final SchoolYearService schoolYearService,
             @NotNull final EthnicityService ethnicityService,
             @NotNull final GenderService genderService,
-            @NotNull final ElasService elasService) {
+            @NotNull final ElasService elasService,
+            @NotNull final SubjectService subjectService) {
```
dsnyder-fairway: grab list of subjects

dsnyder-fairway:
note -- there will be continue work on this

webapp/src/main/webapp/src/app/group-cards/group-card.service.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Group } from '../groups/group';
+import { AssessmentExamMapper } from '../assessments/assessment-exam.mapper';
+import { MeasuredAssessment } from './measured-assessment';
+
+const ServiceRoute = ReportingServiceRoute;
+
+@Injectable()
+export class GroupCardService {
+  group: Group;
+  schoolYear: number;
```
esotrope: think you forgot to remove these

webapp/src/main/webapp/src/app/group-cards/group-cards.component.ts
```diff
@@ -0,0 +1,121 @@
...
+    this._currentSubject = value;
+    if (this._currentSubject !== 'ALL') {
+      this.measuredAssessments = this.measuredAssessments.filter(
+        measuredAssessment => measuredAssessment.assessment.subject === this._currentSubject);
+    }
+  }
+
+  private updateFilterOptions(): void {
+    this.filterOptions.subjects.push('ALL');
+    this.filterOptions.subjects.sort((a, b) => a.localeCompare(b));
```
esotrope: we should make a dedicated service that does this bit of logic - also we should not alpha sort - we should use the natural order of the backend (Math, ELA)

webapp/src/main/webapp/src/app/groups/group.ts
```diff
@@ -1,7 +1,8 @@
...
-  schoolName: string;
-  schoolId: number;
-  subjectCode: string;
+export interface Group {
+  readonly id: number;
+  readonly name: string;
+  readonly schoolName: string;
+  readonly schoolId: number;
+  readonly subjectCode: string;
+  readonly totalStudentCount?: number;
```
esotrope: i'd pick either `totalStudents` or `studentCount` i think total and count are a bit redundant

webapp/src/main/webapp/src/app/group-cards/measured-assessment.ts
```diff
@@ -0,0 +1,10 @@
+import { Assessment } from '../assessments/model/assessment.model';
+
+export interface MeasuredAssessment {
```
esotrope: 👍

webapp/src/main/webapp/src/app/group-cards/iab-card.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  percentFill(index: number): number {
+    return Math.round((this.measuredAssessment.studentCountByPerformanceLevel[ index ] / this.measuredAssessment.studentsTested) * 100);
+  }
+
+  selectCard(): void {
+    this.selected = !this.selected;
+    this.selectedIab.emit(<IabEvent>{
+      id: this.measuredAssessment.assessment.id,
+      selected: this.selected
```
esotrope: i'd be tempted to call these "assessment cards" and "assessment card events" also i think it might make more sense to have a card model which holds 1. the assessment info and 2. whether or not it is selected or not. then you can just dispatch the card object when clicked

webapp/src/main/webapp/src/app/group-cards/iab-card.component.html
```diff
@@ -0,0 +1,37 @@
...
+      <div class="bar-container">
+        <div class="bar blue-dark" *ngIf="percentFill(0) > 0" [attr.data-width]="percentFill(0)">
+          <span class="percentage">{{ 'common.value-display-type-format.Percent' | translate: { value: percentFill(0)} }}</span>
+          <span class="bar-label sr-only">{{ 'common.assessment-type.iab.performance-level.1.name' }}</span>
+        </div>
+        <div class="bar blue-dark aqua" *ngIf="percentFill(1) > 0" [attr.data-width]="percentFill(1)">
+          <span class="percentage">{{ 'common.value-display-type-format.Percent' | translate: { value: percentFill(1)} }}</span>
+          <span class="bar-label sr-only">{{ 'common.assessment-type.iab.performance-level.2.name' }}</span>
+        </div>
+        <div class="bar aqua" *ngIf="percentFill(2) > 0" [attr.data-width]="percentFill(2)">
```
esotrope: i would pre-compute these kind of things and shove them into a view model so they don't have to be computed redundantly at render time.

jtreuting:
Quick note but will review more tonight.  The CSS for the selected cards is missing the blue circle and icon in the upper right that the code pen had.

webapp/src/main/webapp/src/app/group-cards/group-cards.component.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  addIab(event: IabEvent) {
+    if (event.selected) {
+      this.selectedIabs.push(event.id);
```
esotrope: would change this to `selectedCards` and just hold the card in its entirety no?

webapp/src/main/webapp/src/app/group-cards/group-cards.component.ts
```diff
@@ -0,0 +1,121 @@
...
+      this.disableView = false;
+    } else {
+      this.selectedIabs = this.selectedIabs.filter(id => id !== event.id);
+      if (this.selectedIabs.length === 0) {
+        this.disableView = true;
+      }
+    }
```
esotrope: i would rewrite this whole method as something like:
```typescript

onCardSelectionChange() {
  this.selectedCards = this.cards.filter(card => card.selected);
  this.disableView = this.selectedCards.length === 0;
}
```

webapp/src/main/webapp/src/app/group-cards/group-cards.component.ts
dsnyder-fairway: my understanding was that this would be used to go to the old groups pages with the selected assessments displaying. That's why I am just passing the id for now

webapp/src/main/webapp/src/app/group-cards/group-cards.component.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  get currentGroup(): Group {
+    return this._currentGroup;
+  }
+
+  set currentGroup(value: Group) {
+    this._currentGroup = value;
+    if (this._currentGroup) {
+      this.groupCardService.group = this._currentGroup;
```
esotrope: why is this necessary? can't we just pass the current group into the service method call?

webapp/src/main/webapp/src/app/group-cards/group-cards.component.html
```diff
@@ -0,0 +1,80 @@
...
+                [(ngModel)]="currentSubject"
+                class="form-control"
+                (change)="updateRoute()">
+          <option *ngFor="let subject of filterOptions.subjects"
+                  [ngValue]="subject">{{'common.subject.' + subject + '.short-name' | translate}}
+          </option>
+        </select>
+      </div>
+      <div class="col-md-6">
+        <a href="javascript:void()" [ngClass]="{'disabled': disableView}" class="col-btn btn btn-primary">{{ 'iab-card.view-assessments' | translate }}<i class="ml-xs fa fa-angle-right"></i></a>
```
esotrope: i would flip this var to be `cardViewEnabled`

dsnyder-fairway:
@jtreuting looks like the latest codepen has this in there. I'll have to grab the latest

esotrope:
should the cards have a white bg?

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -216,6 +217,22 @@ export const routes: Routes = [
           studentTestHistoryChildRoute
         ]
       },
+      {
+        path: 'group-cards/:groupId',
```
jtreuting: Should we maybe can this the group-dashboard?

webapp/src/main/webapp/src/app/group-cards/group-cards.component.ts
```diff
@@ -0,0 +1,121 @@
...
+export class GroupCardsComponent implements OnInit {
+
+  measuredAssessments: MeasuredAssessment[] = [];
+  group: Group;
+  groups: Group[];
+  _currentGroup: Group;
+  filterOptions: ExamFilterOptions = new ExamFilterOptions();
+  currentSchoolYear: number;
+  _currentSubject: string;
+  disableView = true;
```
jtreuting: Do we have a standard for the _ vs non names?  I'm having a hard time seeing when one is used versus the other

jtreuting: Send like the underscore should be private?

esotrope: i think we've been only using _ on private members that need getters/setters. might be more straight forward if we always use it for private members though

dsnyder-fairway:
going to close this since I renamed a bunch of files and did a bunch of improvements. I'll issue a separate PR once UT have been completed

PR #1218, opened 2018-04-19
use GroupMessage from RDW_Common

malaffoon:
Mostly i'm delaying real change while i figure out the data flow for this ...


admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupsSource.java
```diff
@@ -21,27 +21,25 @@
 @Service
```
malaffoon: all the changes in here are unrelated fixes to format / warnings

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -179,7 +179,7 @@ private void archiveFile(final MultipartFile file,
     private void validate(final InputStream inputStream,
                           final PermissionScope permissionScope,
                           final Collection<CsvValidationResult> validationResults,
-                          final Map<String, String> fileProperties) throws IOException {
+                          final Map<String, String> fileProperties) {
```
malaffoon: unrelated, fix warning

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupsSourceIT.java
```diff
@@ -17,12 +17,9 @@
...
 
-//@RunWith(SpringRunner.class)
-//@SpringBootTest
-/**
- * DefaultGroupsSourceTest
- */
-public class DefaultGroupsSourceTest {
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class DefaultGroupsSourceIT {
```
malaffoon: I'm not sure why this test was commented out. Runs fine. It is more an IT because it stands up the Spring context.

PR #1219, opened 2018-04-19
fixes test

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -290,7 +290,7 @@ export class AggregateReportTableComponent implements OnInit {
       new Column({ id: 'assessmentGrade', field: 'assessmentGradeCode' }),
       new Column({ id: 'assessmentLabel' }),
       new Column({ id: 'schoolYear' }),
-      new Column({ id: 'dimension', field: 'dimension.id' })
+      new Column({ id: 'dimension', field: 'subgroup.id' })
```
esotrope: we don't use this to directly lookup the property on the row object but i changed it anyway just on principle

PR #1220, opened 2018-04-20
Feature/dwr 1675 UI for Group Dashboard

dsnyder-fairway:
screen shot 2018-04-19 at 7 22 18 pm

getting to group dashboard

screen shot 2018-04-19 at 7 22 39 pm

hover over 

screen shot 2018-04-19 at 7 22 50 pm

selecting some assessments 

screen shot 2018-04-19 at 7 23 06 pm

subject drop down ordering 

screen shot 2018-04-19 at 7 23 16 pm

switched subject, notice View Assessments becomes disabled by doing this. Also note after deselecting all assessments "View Assessments" will become disabled

screen shot 2018-04-19 at 7 23 23 pm


webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -217,6 +218,22 @@ export const routes: Routes = [
           studentTestHistoryChildRoute
         ]
       },
+      {
+        path: 'group-dashboard/:groupId',
```
dsnyder-fairway: @jtreuting moved this over

webapp/src/main/webapp/src/app/group-dashboard/assessment-card.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  constructor(public colorService: ColorService) {
+  }
+
+  ngOnInit() {
+    this.percents = [
+      Math.round(this.measuredAssessment.studentCountByPerformanceLevel[ 0 ].percent),
+      Math.round(this.measuredAssessment.studentCountByPerformanceLevel[ 1 ].percent),
+      Math.round(this.measuredAssessment.studentCountByPerformanceLevel[ 2 ].percent)
+    ];
+    // data widths must sum to 100. This avoids issues where the result is 99 or 101
```
dsnyder-fairway: we round the percentages, which can lead to sums that go over 100 or to 99. I think that is fine if we don't introduce decimal points. But, to make our bar okay we have to squeeze this into 100

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.spec.ts
```diff
@@ -0,0 +1,56 @@
+import { TestModule } from '../../test/test.module';
```
dsnyder-fairway: testing is still mostly "to be completed", but here is a start

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    return <MeasuredAssessment>{
+      assessment: this.assessmentExamMapper.mapAssessmentFromApi(serverAssessment.assessment),
+      averageScaleScore: serverAssessment.measures.avgScaleScore,
+      averageStandardError: serverAssessment.measures.avgStdErr,
+      date: serverAssessment.completedAt,
+      studentsTested: serverAssessment.studentsTested,
+      studentCountByPerformanceLevel: [
+        <DetailsByPerformanceLevel>{
+          studentCount: serverAssessment.measures.level1Count,
+          percent: (serverAssessment.measures.level1Count / serverAssessment.studentsTested) * 100
```
dsnyder-fairway: @esotrope calculate percentage here. Alternative we can move this into `ngOnInit` in the `assessment-card.component.ts`. I am keeping unrounded percentage as it might have other uses somewhere

webapp/src/main/webapp/src/app/shared/test/mock.activated-route.ts
```diff
@@ -1,8 +1,10 @@
 import Spy = jasmine.Spy;
 import createSpy = jasmine.createSpy;
+import { EventEmitter } from '@angular/core';
 
 export class MockActivatedRoute {
   snapshotResult: Spy = createSpy("snapshot");
+  params: EventEmitter<any> = new EventEmitter();
```
dsnyder-fairway: I noticed we had another `mock.activated-route.ts` with this as the only difference. I copied this here and removed the other `mock.activated-route.ts`

webapp/src/main/webapp/src/app/group-dashboard/assessment-card.component.html
```diff
@@ -0,0 +1,40 @@
+<div class="card assessment-aggregate" [ngClass]="{'selected': selected}" (click)="selectCard()">
+
+  <span class="check"><i class="fa fa-check"></i></span>
+  <span
+    class="label green">{{ 'common.assessment-grade-short-label.' + measuredAssessment.assessment.grade | translate}}</span>
```
jtreuting: the `green` class should be generated based on the grade the same way the assessment names are on the group results page.  whatever logic is used there should be used here.

webapp/src/main/webapp/src/app/group-dashboard/assessment-card.component.html
```diff
@@ -0,0 +1,40 @@
...
+  <span class="title">{{ measuredAssessment.assessment.label }}</span>
+  <span class="date">{{ measuredAssessment.date | date }}</span>
+  <div class="students">
+    <span class="caption"
+          [innerHTML]="'assessment-card.student-count' | translate: {studentsTested: this.measuredAssessment.studentsTested, groupCount: group.totalStudents }"></span>
+    <div class="scale-score-graph rounded small">
+      <div class="bar-container">
+        <div class="bar green" [attr.data-width]="studentCountFill"></div>
```
jtreuting: @aaronsleeper were you intending the color of the number of students bar to match the color assigned to the grade, or for it to always be green?

webapp/src/main/webapp/src/app/group-dashboard/assessment-card.component.ts
jtreuting: can you add some tests around this logic since we ran into issues with this in the past when we had the visualization horizontal

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+    return this._currentSubject;
+  }
+
+  @Input()
+  set currentSubject(value: string) {
+    this._currentSubject = value;
+    if (!this._currentSubject) {
+      this.measuredAssessments = this.measuredAssessments.filter(
+        measuredAssessment => measuredAssessment.assessment.subject === this._currentSubject);
```
jtreuting: should this check be on `if (this._currentSubject)`, cause right now it's saying that if there is no subject go find the measuredAssessments where it also doesn't have a subject.  Or I'm missing something.

webapp/src/main/webapp/src/app/group-dashboard/assessment-card.component.html
dsnyder-fairway: I'll correct this

webapp/src/main/webapp/src/app/group-dashboard/assessment-card.component.ts
dsnyder-fairway: working on a test now

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -217,6 +218,22 @@ export const routes: Routes = [
...
+        data: {
+          breadcrumb: { translate: 'groups.name' },
```
esotrope: is it really `groups.name` or `group.name`?

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.ts
dsnyder-fairway: removed the if statement

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.ts
```diff
@@ -0,0 +1,111 @@
...
+    return this.selectedAssessments.length !== 0;
+  }
+
+  updateRoute(changeSource: string): void {
+    this.selectedAssessments = [];
+    this.router.navigate([ 'group-dashboard', this.currentGroup.id, {
+      schoolYear: this.currentSchoolYear,
+      subject: this.currentSubject
+    } ]).then(() => {
+      this.groupService.getGroup(this.currentGroup.id).subscribe((group) => {
```
esotrope: i don't feel like this belongs here - what happens if someone directly links to the group dashboard page? does this loading happen?

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.ts
```diff
@@ -0,0 +1,111 @@
...
+          }
+        });
+      });
+      // TODO analytics
+    });
+  }
+
+  get currentSubject(): string {
+    if (!this._currentSubject) {
+      return 'ALL';
```
esotrope: is it a lift to change this to use `undefined` to represent no All/No filter?

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.ts
dsnyder-fairway: yes, this is used in `ngOnInit` as well

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.ts
dsnyder-fairway: the problem with that is how do we get `undefined` to appear as a value in the drop down list if we are left with two values?

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.service.ts
esotrope: for stuff like this usually we have a second service or "mapper" that takes the service response and produces view models for our views. that way the percent/width calculations can be put into an easily tested mapper service method and stay out of a component

webapp/src/main/webapp/src/app/group-dashboard/group-dashboard.component.ts
```diff
@@ -0,0 +1,111 @@
...
+  private updateFilterOptions(): void {
+    this.filterOptions.subjects.unshift('ALL');
+  }
+
+  onCardSelection(event: AssessmentCardEvent) {
+    if (event.selected) {
+      this.selectedAssessments.push(event.measuredAssessment.assessment.id);
+    } else {
+      this.selectedAssessments = this.selectedAssessments.filter(id => id !== event.measuredAssessment.assessment.id);
+    }
```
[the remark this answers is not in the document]
dsnyder-fairway: I'm not sure I follow "we shouldn't be assuming angular will detect a change via Array.push." This is in response to an event being emitted, so we push to the array. This shouldn't fire any changes elsewhere. When a card is selected we fire the event, here we add that to our selected assessments. We will use this list of selected assessments to bring them to view their assessments

webapp/src/main/webapp/src/app/app.routes.ts
dsnyder-fairway: this is borrowed text (originally `groups.name`). The mockup has `Student Groups` instead of `My Groups`. I suppose I should change this

webapp/src/main/webapp/src/app/group-dashboard/assessment-card.component.ts
dsnyder-fairway: done with test

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -941,6 +941,7 @@
     "template-file": "groups"
   },
   "group-dashboard": {
+    "heading-name": "Student Groups",
     "name": "Assessment Reports"
```
esotrope: Shouldn't the page heading match the link name? also i think we can just call it `heading`

webapp/src/main/webapp/src/app/group-dashboard/measured-assessment.mapper.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  constructor(private assessmentExamMapper: AssessmentExamMapper) {
+  }
+
+  mapMeasuredAssessmentsFromApi(serverAssessments: any[]): MeasuredAssessment[] {
+    return serverAssessments
+      .map(serverAssessment => this.mapMeasuredAssessmentFromApi(serverAssessment));
+  }
+
+  mapMeasuredAssessmentFromApi(serverAssessment: any): MeasuredAssessment {
```
esotrope: just moving the map methods to another class this isn't exactly what i meant - i'll reach out to you on slack

PR #1223, opened 2018-04-20
Design/subject xml payload

wtritch:
After discussion with @malaffoon and @jtreuting I've consolidated the subscores, tertiary scores, claims, and targets into a single "Category" hierarchy:
```
ELA:
SOCK_R
  1-LT
  1-IT
2-W
SOCK_LS
  3-L
  3-S
4-CR

MATH:
1
SOCK_2
  2
  4
3
```

The two files that constitute the deliverable for this PR/task are `Math_subject.xml` and `ELA_subject.xml`  Everything else is just scratch-work and a generation framework.


ELA_subject.xml
```diff
@@ -0,0 +1,3443 @@
+<Subject code="ELA" name="ELA">
+  <CommonCoreStandardDescription><![CDATA[This column reflects the Common Core State Standard(s) associated with a test item. Test items may be associated with more than one standard.  For example, a question that asks students to evaluate research sources may assess both reading and research skills.  For more information, see the <a href=\"https://portal.smarterbalanced.org/library/en/english-language-artsliteracy-content-specifications.pdf\" target=\"_blank\">English Language Arts Content Specifications</a>.]]></CommonCoreStandardDescription>
+  <AssessmentTypes>
+    <AssessmentType code="ica">
+      <PerformanceLevels>
+        <PerformanceLevel level="1" name="Did Not Meet Standard"/>
+        <PerformanceLevel level="2" name="Nearly Met Standard"/>
+        <PerformanceLevel level="3" name="Met Standard">
+          <GroupingCutpoint>true</GroupingCutpoint>
+        </PerformanceLevel>
```
wtritch: I changed the `rollup="true"` optional attribute into an optional "GroupingCutpoint" sub-element to indicate that a particular performance level should be used as the "at-or-above" grouping level.

ELA_subject.xml
```diff
@@ -0,0 +1,3443 @@
...
+        <PerformanceLevel level="4" name="Exceeded Standard"/>
+      </PerformanceLevels>
+      <SubScoring>
+        <SubScorePerformanceLevels>
+          <SubScorePerformanceLevel level="1" name="Below Standard"/>
+          <SubScorePerformanceLevel level="2" name="Near Standard"/>
+          <SubScorePerformanceLevel level="3" name="Above Standard"/>
+        </SubScorePerformanceLevels>
+      </SubScoring>
```
wtritch: This makes me a little nervous.  We now define the available sub-score performance levels a ways away from defining the Categories that can be scored on.  This also assumes that the sub-score performance levels *must always* be the same for all sub-scores for this assessment type (claim-level scores, target-level scores, etc) Thoughts?

ELA_subject.xml
```diff
@@ -0,0 +1,3443 @@
...
+      <ClaimAchievementMessage><![CDATA[{0}''s results are also reported out into the four claims (i.e., areas) that together make up English language arts/literacy: Reading, Writing, Listening, and Research/Inquiry. These four claims are based on California''s academic standards, which describe the knowledge and skills students are expected to learn in grade three. If the student received a score of "No Score Available" for any claim, this means he or she did not complete enough items to receive a score in that area. To view a description of the student''s achievement level claim (i.e., area), select either [Above Standard], [Near Standard], or [Below Standard] from the student-level selections available in the table below.]]></ClaimAchievementMessage>
```
wtritch: I wrapped all of the long message text in CDATA tags just to help me sleep at night.

ELA_subject.xml
```diff
@@ -0,0 +1,3443 @@
...
+              <Description><![CDATA[LANGUAGE USE: Interpret use of language by distinguishing literal from non-literal meanings of words and phrases used in context]]></Description>
+            </Category>
+            <Category code="7-11" name="7-11">
+              <Description><![CDATA[LANGUAGE USE: Determine or analyze the figurative (e.g., euphemism, oxymoron, hyperbole, paradox), or connotative meanings of words and phrases used in context and the impact of those word choices on meaning and tone]]></Description>
+            </Category>
+            <Category code="7-2" name="7-2">
+              <Description><![CDATA[LANGUAGE USE: Interpret use of language by distinguishing literal from non-literal meanings of words and phrases used in context]]></Description>
+            </Category>
+            <Category code="7-3" name="7-3">
+              <Description><![CDATA[LANGUAGE USE: Interpret use of language by distinguishing literal from non-literal meanings of words and phrases us
```
wtritch: I realize we're trying to re-create an actual hierarchy from the shadows of the data we've seen come through, but my gut says that SOCK_LS should be named "Listening & Speaking" and that "3-S" should be labelled "Speaking" ... @jtreuting ?

common/src/main/java/org/opentestsystem/rdw/reporting/common/subject/model/AssessmentType.java
```diff
@@ -0,0 +1,52 @@
+package org.opentestsystem.rdw.reporting.common.subject.model;
+
+import java.util.List;
+
+/**
+ *
```
esotrope: very descriptive : j

wtritch: These won't be checked in =P  They're just little helpers so that I can saturate the payload via code and the DB.  For reading I'd recommend a StAX streaming parser rather than straight Jackson de-serialization into models.

hand-jammed.math.subject.xml
```diff
@@ -0,0 +1,199 @@
...
+        <PerformanceLevel level="2" name="Near Standard"/>
+        <PerformanceLevel level="3" name="Above Standard"/>
+      </PerformanceLevels>
+    </AssessmentType>
+    <AssessmentType code="ica">
+      <PerformanceLevels>
+        <PerformanceLevel level="1" name="Did Not Meet Standard"/>
+        <PerformanceLevel level="2" name="Nearly Met Standard"/>
+        <PerformanceLevel level="3" name="Met Standard" rollup="true"/>
+        <PerformanceLevel level="4" name="Exceeded Standard"/>
```
esotrope: just thinking, does it really make sense that our rollup is pivoted on a performance level? i feel like its an implementation detail that we use an index to cut the list in half. should we instead make this more generic and have like `<PerformanceLevel group="1">`? this isn't necessarily better but i feel we need something to move this away from picking an element to slap "rollup" on. id doesn't really communicate that the selected element and elements after will be grouped into the same group.

hand-jammed.math.subject.xml
```diff
@@ -0,0 +1,199 @@
...
+        <PerformanceLevel level="2" name="Near Standard"/>
+        <PerformanceLevel level="3" name="Above Standard"/>
+      </SubScorePerformanceLevels>
+      <SubScores>
+        <SubScore code="1" name="Concepts &amp; Procedures">
+          <Bindings>
+            <Binding claimCode="1"></Binding>
+          </Bindings>
+          <!-- Optional? -->
+          <Icon>fa-book</Icon>
```
esotrope: i would think this would be optional. maybe there is a tenant that doesn't want icons? i guess they could put `<Icon>no_such-CSS-class</Icon>` there had to

wtritch: It is optional =)

hand-jammed.math.subject.xml
wtritch: Hmm... I can't decide if it's cleaner or more obscure to have `group="1"` on half the elements and `group="2"` on the other half... or if we'll ever want to group n distinct performance levels into more than 2 groups... @jtreuting @malaffoon ?

hand-jammed.math.subject.xml
wtritch: I should probably remove this hand-jammed xml... it was just a scratch-pad for thoughts...

wtritch: I updated the PR description to call out that `Math_subject.xml` and `ELA_subject.xml` are the actual deliverables for this PR.

ELA_subject.xml
malaffoon: Both the optional attribute and optional sub-element have one serious concern: the spec shouldn't allow more than one level to be specified as the grouping level (i think it should allow 0-or-1). Perhaps this should be an optional attribute/element on the parent element? `<PerformanceLevels groupingCutpoint="1">`

hand-jammed.math.subject.xml
malaffoon: Hmm, interesting thought. My gut is telling me that there is a single concern here: is the score sufficient to count as "passing" the standard.

esotrope: yeah that is a good way to think about it i think

Math_subject.xml
```diff
@@ -0,0 +1,4224 @@
...
+      </SubScoring>
+    </AssessmentType>
+  </AssessmentTypes>
+  <ReportGrades>
+    <ReportGrade code="03">
+      <OverallAchievementMessage><![CDATA[{0} took a Smarter Balanced assessment based on the grade three California academic standards for mathematics and received an overall score that ranges from about 2000 to 3000. This overall score falls into one of the four achievement levels shown here. Sample test items may be included that exemplify the kinds of passages and questions that students at different levels of achievement would likely answer correctly.]]></OverallAchievementMessage>
+      <ClaimAchievementMessage><![CDATA[{0}''s results are also reported out into the three claims (i.e., areas) that together make up mathematics: Concepts and Procedures, Problem Solving and Modeling/Data Analysis, and Communicating Reasoning. These three claims (i.e., areas) are based on California''s academic standards, which describe the knowledge and skills students are expected to learn in grade three. If {0} received a score of "No Score Available" for any claim, this means he or she did not complete enough items to receive a score in that area.]]></ClaimAchievementMessage>
```
esotrope: I don't think we'll need the double single apostrophes in this format `''` i think it was just an annoying `.properties` file format thing

ELA_subject.xml
wtritch: hm... I like that idea

ELA_subject.xml
wtritch: I could see that, because if we move it to a subelement of Category we end up with a *lot* of duplication, especially once we support target-level scoring.

ELA_subject.xml
```diff
@@ -0,0 +1,3443 @@
+<Subject code="ELA" name="ELA">
+  <StandardsDescription><![CDATA[This column reflects the Common Core State Standard(s) associated with a test item. Test items may be associated with more than one standard.  For example, a question that asks students to evaluate research sources may assess both reading and research skills.  For more information, see the <a href=\"https://portal.smarterbalanced.org/library/en/english-language-artsliteracy-content-specifications.pdf\" target=\"_blank\">English Language Arts Content Specifications</a>.]]></StandardsDescription>
+  <AssessmentTypes>
+    <AssessmentType code="ica">
```
agorina: Just trying to understand the objective here: does this define an `AssessmentType`, or does it say that this Subject is for the assessments of a predefined type `ica`.

In other words, is the `code` value limited to a set predefined outside of this XML, or could somebody create a new `AssessmentType` here?

hand-jammed.math.subject.xml
wtritch: I know we like to avoid works like "passing" or "failing" but does "passingCutoff" make sense here?  "standardCutoff" just sounds a little obtuse.

ELA_subject.xml
```diff
@@ -0,0 +1,3443 @@
...
+      <ReportPerformanceLevels>
+        <ReportPerformanceLevel level="1"><![CDATA[{0} did not meet the grade three standard for English language arts/literacy needed for likely success in future coursework. Please discuss with {0}''s teacher(s) ways to help {0} improve.]]></ReportPerformanceLevel>
+        <ReportPerformanceLevel level="2"><![CDATA[English language arts/literacy needed for likely success in future coursework. Please discuss with {0}''s teacher(s) ways to help {0} improve.]]></ReportPerformanceLevel>
+        <ReportPerformanceLevel level="3"><![CDATA[{0} has made progress and met the grade three standard for English language arts/literacy. {0} appears ready for future coursework.]]></ReportPerformanceLevel>
+        <ReportPerformanceLevel level="4"><![CDATA[{0} has made progress and exceeded the grade three standard for English language arts/literacy. {0} appears ready for future coursework.]]></ReportPerformanceLevel>
+      </ReportPerformanceLevels>
+      <ReportClaims>
+        <ReportClaim code="4-CR">
```
agorina: This element is named `ReportClaim` and the `code` value matches the code value in the element `Categories/Category`. 

Why not have a name that would show the relationship? Are we intentionally trying to separate them?

ELA_subject.xml
```diff
@@ -0,0 +1,3443 @@
...
+        </ReportClaim>
+      </ReportClaims>
+    </ReportGrade>
+  </ReportGrades>
+  <Categories>
+    <Category code="SOCK_R" name="Reading" displayOrder="1">
+      <Icon>fa-book</Icon>
+      <Categories>
+        <Category code="1-LT" name="Literary Text">
+          <Categories>
```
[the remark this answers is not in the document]
wtritch: Technically "SOCK_R" (the parent) is the scorable category.  I wonder since this is a tree, if we should move the performance levels to either the root Categories (with some duplication) or as an optional element inside the <Categories> container and support nested values.  For instance, I doubt very much that one target or another will have different performance levels, *but* I could see leaf categories having separate performance levels from root categories:
```
<Categories>
  <PerformanceLevels>
    <PerformanceLevel level="1"/>
    ...
  </PerformanceLevels>
  <Category code="SOCK_2">
    ...
    <Categories>
      <Category code="1-LT">
        <Categories>
          <PerformanceLevels>...</PerformanceLevels>
          <Category code="1-1" name="1-1">
            <Description>...</Description>
```
?

Math_subject.xml
wtritch: I'm sure there'll be some touch-up needed once we get these through the full pipeline =/

ELA_subject.xml
wtritch: The `code` value is limited to "ica" "iab" or "sum" is my thought.  Although we are planning to allow *naming* an assessment type within the context of a subject.  (ELPAC IABs will be called "Initial" in the UI...)

ELA_subject.xml
wtritch: Hm... It is and must be a reference to a Category, but I'm wary of having a bunch of different names on our elements... verbosity vs consistency... I'm conflicted.

ELA_subject.xml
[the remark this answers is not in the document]
wtritch: There's still some duplication above, but at least it isn't on the leaves, and would only be required if a lower reporting category actually *has* different performance levels from a parent category.

hand-jammed.math.subject.xml
wtritch: Or maybe "passingLevel" ?

agorina:
I am sure you have thought about it, but just curious: things like ‘math_practice’. 'item_difficulty_cuts'  or ‘depth_of_knowledge’, are they also part of the configuration?

ELA_subject.xml
wtritch: I renamed the element to ReportCategory, we can call out comments in a dtd if/when we make one.

wtritch:
@agorina I added DepthsOfKnowledge and ItemDifficulties to the XML payload.

hand-jammed.math.subject.xml
wtritch: Done.  Moved cutpoint definition to an optional "standardCutoff" attribute on the PerformanceLevels container element.

ELA_subject.xml
```diff
@@ -0,0 +1,3471 @@
+<Subject code="ELA" name="ELA">
+  <StandardsDescription><![CDATA[This column reflects the Common Core State Standard(s) associated with a test item. Test items may be associated with more than one standard.  For example, a question that asks students to evaluate research sources may assess both reading and research skills.  For more information, see the <a href=\"https://portal.smarterbalanced.org/library/en/english-language-artsliteracy-content-specifications.pdf\" target=\"_blank\">English Language Arts Content Specifications</a>.]]></StandardsDescription>
+  <AssessmentTypes>
+    <AssessmentType code="ica">
+      <PerformanceLevels standardCutoff="3">
+        <PerformanceLevel level="1" name="Did Not Meet Standard" color="maroon"/>
+        <PerformanceLevel level="2" name="Nearly Met Standard" color="gray-darkest"/>
+        <PerformanceLevel level="3" name="Met Standard" color="green-dark"/>
+        <PerformanceLevel level="4" name="Exceeded Standard" color="blue-dark"/>
```
wtritch: I've added colors to the PerformanceLevel elements.

ELA_subject.xml
```diff
@@ -0,0 +1,3471 @@
...
+        <ItemDifficulty gradeCode="07" moderateLowEnd="-0.7215" difficultLowEnd="1.06739"/>
+        <ItemDifficulty gradeCode="08" moderateLowEnd="-0.47018" difficultLowEnd="1.34599"/>
+        <ItemDifficulty gradeCode="11" moderateLowEnd="-0.38186" difficultLowEnd="1.5479"/>
+      </ItemDifficulties>
+      <SubScoring>
+        <PerformanceLevels>
+          <PerformanceLevel level="1" name="Below Standard" color="aqua"/>
+          <PerformanceLevel level="2" name="Near Standard" color="textured-blue"/>
+          <PerformanceLevel level="3" name="Above Standard" color="blue-dark"/>
+        </PerformanceLevels>
```
wtritch: After some discussion with @malaffoon and @jtreuting I've decided to leave a single definition for sub-score performance levels in the AssessmentType definition.  The assumption is that all claim, target, etc subscores will all have the same performance levels available, although we will need to confirm with the client.

ELA_subject.xml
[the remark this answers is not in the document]
wtritch: Aborted, see comment above for leaving a single sub-score performance-levels definition per subject per assessment type.

ELA_subject.xml
```diff
@@ -0,0 +1,3471 @@
...
+            </Category>
+            <Category code="6-6" name="6-6">
+              <Description><![CDATA[TEXT STRUCTURES & FEATURES: Relate knowledge of text structures or text features (e.g., layout; visual or auditory elements – lighting, camera effects, music; symbolic or graphic representations) to analyze impact on meaning, style, or presentation]]></Description>
+            </Category>
+            <Category code="6-7" name="6-7">
+              <Description><![CDATA[TEXT STRUCTURES & FEATURES: Relate knowledge of text structures or genrespecific features (visual/graphic/auditory effects) to analyze the impact of those choices on meaning or presentation (e.g., layout; visual or auditory elements – lighting, camera effects, music; symbolic or graphic representations)]]></Description>
+            </Category>
+            <Category code="6-8" name="6-8">
+              <Description><![CDATA[TEXT STRUCTURES/FEATURES: Relate knowledge of text structures or genre features (visual/graphic/auditory effects) to analyze the imp
```
wtritch: I'm not currently calling out *which* categories are scorable... I'm not sure we need/want to... thoughts? @malaffoon ? I know you were expecting us to flag categories as scorable and possibly as "score-type" so that it's easier to group "claim"-level scores and "target"-level scores?  I'm wonder if we want to instead request all sub-scores for an exam and receive a tree like this "Category" tree with scores alongside the descriptions/names?

ELA_subject.xml
```diff
@@ -0,0 +1,3784 @@
...
+      <Reference><![CDATA[https://portal.smarterbalanced.org/library/en/english-language-artsliteracy-content-specifications.pdf#page=54]]></Reference>
+    </DepthOfKnowledge>
+    <DepthOfKnowledge level="4" name="Extended Thinking">
+      <Reference><![CDATA[https://portal.smarterbalanced.org/library/en/english-language-artsliteracy-content-specifications.pdf#page=54]]></Reference>
+    </DepthOfKnowledge>
+  </DepthsOfKnowledge>
+  <ReportGrades>
+    <ReportGrade code="03">
+      <ReportAssessmentTypes>
+        <ReportAssessmentType code="ica">
```
wtritch: Since we support different performance levels per assessment type, I've duplicated all of the SBAC report messages between ICA and SUMMATIVE assessment types... the duplication is unfortunate, but I'm not sure how else to account for potentially different performance levels between assessment types.

RDW_Subject.xsd
```diff
@@ -0,0 +1,275 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns="http://www.smarterbalanced.org/RDW_Subject"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           elementFormDefault="qualified"
+           targetNamespace="http://www.smarterbalanced.org/RDW_Subject">
```
wtritch: I created an XSD schema definition file for our subject payloads.  I'm going to go through and start adding comments to the definition.  I'm thinking this is likely the file we'll put up as a Gist for interaction/feedback from the client.

RDW_Subject.xsd
```diff
@@ -0,0 +1,275 @@
...
+    <xs:restriction base="xs:string">
+      <xs:maxLength value="10"/>
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="String20">
+    <xs:restriction base="xs:string">
+      <xs:maxLength value="20"/>
+    </xs:restriction>
+  </xs:simpleType>
```
wtritch: I don't particularly like the names "String10" and "String20" ... any suggestions for a better naming scheme for length-limited string values?

RDW_Subject.xsd
```diff
@@ -0,0 +1,275 @@
...
+      <xs:enumeration value="gray-darker"/>
+      <xs:enumeration value="gray-dark"/>
+      <xs:enumeration value="gray"/>
+      <xs:enumeration value="gray-light"/>
+      <xs:enumeration value="gray-lighter"/>
+      <xs:enumeration value="gray-lightest"/>
+      <xs:enumeration value="white"/>
+      <xs:enumeration value="textured-blue"/>
+    </xs:restriction>
+  </xs:simpleType>
```
wtritch: These are all the color classes available in the system.

RDW_Subject.xsd
```diff
@@ -0,0 +1,275 @@
...
+
+  <xs:simpleType name="LevelOfPerformance">
+    <xs:restriction base="xs:integer">
+      <xs:minInclusive value="1"/>
+      <xs:maxInclusive value="6"/>
+    </xs:restriction>
+  </xs:simpleType>
```
wtritch: I've limited performance levels to 1 through 6.

RDW_Subject.xsd
```diff
@@ -0,0 +1,275 @@
...
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="03"/>
+      <xs:enumeration value="04"/>
+      <xs:enumeration value="05"/>
+      <xs:enumeration value="06"/>
+      <xs:enumeration value="07"/>
+      <xs:enumeration value="08"/>
+      <xs:enumeration value="11"/>
+    </xs:restriction>
+  </xs:simpleType>
```
wtritch: I've limited the grades to only allow the grades that we currently support.

RDW_Subject.xsd
```diff
@@ -0,0 +1,275 @@
...
+        <xs:element ref="ReportGrades"/>
+        <xs:element ref="Categories"/>
+        <xs:element ref="Standards"/>
+      </xs:sequence>
+      <xs:attribute name="code" use="required" type="String10"/>
+      <xs:attribute name="name" use="required" type="xs:string"/>
+    </xs:complexType>
+  </xs:element>
+
+  <xs:element name="StandardsDescription" type="xs:string"/>
```
wtritch: This is currently the only subject-scoped translation string that we use... but it feels a little strange to just be hanging out as a root subject element... I wonder if we want to create a more extensible <Messages> element that can contain any future subject-scoped message strings... thoughts?

RDW_Subject.xsd
agorina: I think it is good this way. I was able to guess what it is before I found the type definition.

RDW_Subject.xsd
agorina: How are these going to be used? 

Do we plan on pre-loading (hardcoding) all possible codes and use this as a reference to look up db ids, OR are we planning on using these to load the `grade` table?

RDW_Subject.xsd
jtreuting: i like the idea of making it more of a generic descriptions type element that currently only has Standards within it but could have others in the future.

RDW_Subject.xsd
wtritch: Since the grades are a larger concern than subjects, I assume we'll hard-code the grade codes (as we do now) and just use these to reference particular grades in the subject definition.  In this schema they're currently used:
1. to bind report (PDF) messages to a particular grade
2. to bind item difficulty cutpoints to a particular grade

agorina: Sounds good, thank you!

jtreuting: we'll need to discuss with SB but i would assume that we need 9 and 10 and 12 at least so that it covers 3 - 12.  we might also need to add KG, 1 and 2 to cover all elementary school and up.  we know SB doesn't produce content for these but it's probably safe to assume other subjects would be targeted at these levels

wtritch: That... will involve some serious testing.  I wonder we're handling those assessment grades everywhere...  But I'll add them in.

RDW_Subject.xsd
```diff
@@ -0,0 +1,395 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns="http://www.smarterbalanced.org/RDW_Subject"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           elementFormDefault="qualified"
+           targetNamespace="http://www.smarterbalanced.org/RDW_Subject">
+
+  <!--
+    This defines the root Subject element that is defined by the subject payload.
+  -->
```
wtritch: I added a bunch of comments to the XSD.  @malaffoon @jtreuting @agorina If you wouldn't mind scanning them real quick and let me know if I missed/mis-interpreted something I'd appreciate it. =)

RDW_Subject.xsd
```diff
@@ -0,0 +1,395 @@
...
+    <xs:complexType>
+      <xs:sequence>
+        <!-- The performance levels available for assessments of this type -->
+        <xs:element ref="PerformanceLevels"/>
+        <!-- The item difficulty cutpoints of items for assessments of this type -->
+        <xs:element ref="ItemDifficulties" minOccurs="0"/>
+        <!-- Subscore definitions for assessments of this type -->
+        <xs:element ref="SubScoring" minOccurs="0"/>
+      </xs:sequence>
+      <!-- The assessment type being defined. -->
```
jtreuting: might be worth somewhere here saying it must be summative, ica or iab.  i know it is listed in the enumeration but as you read down making it clear there are a fixed set and not open ended

RDW_Subject.xsd
```diff
@@ -0,0 +1,395 @@
...
+    There must be between 1..6 sequential performance levels defined.
+  -->
+  <xs:element name="PerformanceLevels">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element maxOccurs="6" ref="PerformanceLevel"/>
+      </xs:sequence>
+      <!--
+        This optional attribute defines the performance level at which a student
+        has achieved a "successful" level of performance.
```
jtreuting: I would add something like "and is used in the reporting application to group the data visualization from all levels to above/below"

RDW_Subject.xsd
```diff
@@ -0,0 +1,395 @@
...
+      -->
+      <xs:attribute name="standardCutoff" type="LevelOfPerformance"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="PerformanceLevel">
+    <xs:complexType>
+      <!-- The integer value of the level. -->
+      <xs:attribute name="level" use="required" type="LevelOfPerformance"/>
+      <!-- The display string for the performance level -->
+      <xs:attribute name="name" use="required" type="xs:string"/>
```
jtreuting: i'm just realizing that we currently have a short name and long name to accommodate the UI doing Below (in a blue tag) with Standard on a line below it.  This is in the report headers.  Not sure how we want to address/if we want to address

RDW_Subject.xsd
```diff
@@ -0,0 +1,395 @@
...
+      <!-- Assessment Items with a difficulty below the level are defined as "Easy" -->
+      <xs:attribute name="moderateLowEnd" use="required" type="xs:decimal"/>
+      <!-- Assessment Items with a difficulty above this level are defined as "Difficult" -->
+      <xs:attribute name="difficultLowEnd" use="required" type="xs:decimal"/>
+    </xs:complexType>
+  </xs:element>
+
+  <!--
+    This element controls the performance levels at which sub scores for the
+    parent assessment type are reported.
```
jtreuting: I'd add the context that will be easier to understand by SB that by sub scores we mean Claims and Targets for SB.

RDW_Subject.xsd
```diff
@@ -0,0 +1,395 @@
...
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element maxOccurs="unbounded" ref="ReportGrade"/>
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="ReportGrade">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element ref="ReportAssessmentTypes" maxOccurs="2"/>
```
jtreuting: what does the maximum of 2 refer to?

RDW_Subject.xsd
```diff
@@ -0,0 +1,395 @@
...
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="ReportAssessmentType">
+    <xs:complexType>
+      <xs:sequence>
+        <!-- The performance-agnostic overall achievement message -->
+        <xs:element name="OverallAchievementMessage" type="xs:string"/>
+        <!-- The performance-agnostic sub-score achievement message -->
+        <xs:element name="ClaimAchievementMessage" type="xs:string"/>
```
jtreuting: I'm not following what these are.  can you explain more, i'm missing something.

RDW_Subject.xsd
wtritch: done, added a:
```
<Messages>
  <Message key="standards-description">[CDATA[blahblah]]</Message>
</Messages>
```
section for holding unstructured subject-scoped text.
...theoretically, we could blow away the entire "Reporting" section and implement it with unstructured messages with keys like "reporting.ica.03.overall-achievement" ... thoughts? @jtreuting

RDW_Subject.xsd
wtritch: done

RDW_Subject.xsd
wtritch: done

RDW_Subject.xsd
jtreuting: that feels too unstructured to me

RDW_Subject.xsd
wtritch: I've always hated that split, because there's a good chance it'll fall apart in non-English translation (I know, not important at the moment.)  I'd be tempted to ignore it for now since we'll run into other issues as well.  (Like how do we color text as "textured-blue" which is implemented as a background image?)

RDW_Subject.xsd
jtreuting: oops.  i skimmed too fast :)

RDW_Subject.xsd
```diff
@@ -0,0 +1,413 @@
...
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element ref="Messages"/>
+        <xs:element ref="AssessmentTypes"/>
+        <xs:element ref="DepthsOfKnowledge"/>
+        <xs:element ref="ReportGrades"/>
+        <xs:element ref="Categories"/>
+        <xs:element ref="Standards"/>
+      </xs:sequence>
+      <!-- The subject code must be unique for each subject. -->
```
malaffoon: please define `unique`; across the entire system and all vendors, correct?

RDW_Subject.xsd
wtritch: image

RDW_Subject.xsd
```diff
@@ -0,0 +1,413 @@
...
+      <xs:attribute name="code" use="required" type="String10"/>
+      <!-- This is the display name for the subject. -->
+      <xs:attribute name="name" use="required" type="xs:string"/>
+    </xs:complexType>
+  </xs:element>
+
+  <!--
+    This is additional unstructured text displayed to users that is scoped
+    to the subject.
```
malaffoon: Hmm, more docs here: i still have no idea where/how "Messages" will be used?

RDW_Subject.xsd
wtritch: The ClaimAchievementMessage is shown in the picture above.  The OverallAchievementMessage looks to have been removed from the PDF at some point...

RDW_Subject.xsd
wtritch: `The subject code must be unique for this subject` ?

RDW_Subject.xsd
```diff
@@ -0,0 +1,413 @@
...
+          -->
+          <xs:attribute name="key" use="required" type="xs:string"/>
+        </xs:extension>
+      </xs:simpleContent>
+    </xs:complexType>
+  </xs:element>
+
+  <!--
+    AssessmentType elements define the behavior of assessments for this subject
+    within the context of a specific assessment type.
```
malaffoon: I would continue this, something like, `A subject may have up to 3 assessment types defined, one each for summative, ICA, IAB.`

RDW_Subject.xsd
wtritch:
```
  <!--
    Message elements define unstructured subject-scoped text that is displayed to users.
    For example, the "standards-description" message is displayed to the users as an
    informational popover describing the contents of a "Standards" column within an 
    assessment item display table.
  -->
``` ?

File: RDW_Subject.xsd [hunk 32]
```diff
@@ -0,0 +1,413 @@
...
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="DepthOfKnowledge">
+    <xs:complexType>
+      <xs:sequence>
+        <!-- URI to an externally-reachable resource to provide more information -->
+        <xs:element name="Reference" type="xs:anyURI"/>
+      </xs:sequence>
+      <!-- The integer value of the depth. -->
```
malaffoon: Unlike performance level, the order of the DoK values isn't important (i.e. DoK=1 is not necessarily easy/harder than DoK=4). Not sure if it is worth calling out this distinction.

RDW_Subject.xsd
wtritch: done

RDW_Subject.xsd
wtritch: ...so the depth doesn't necessarily represent the depth? ...I hate naming.

RDW_Subject.xsd
```diff
@@ -0,0 +1,413 @@
...
+      <!-- The display value of the depth. -->
+      <xs:attribute name="name" use="required" type="xs:string"/>
+    </xs:complexType>
+  </xs:element>
+
+  <!--
+    ReportGrade elements define the detailed display messages describing a student's
+    performance in generated PDF reports.
+    Each ReportGrade contains messages for each assessment type
+    report. ("ica" and "sum", "iab" reports do not contain subject-scoped messages)
```
malaffoon: The comment in parens is very confusing, perhaps `(ReportGrades are used only for "ica" and "sum" assessment types, the "iab" reports do not contain subject-scoped messages.)`

wtritch: Good call, done

RDW_Subject.xsd
wtritch: Added:
```
NOTE that for SBAC assessments, this represents the performance levels
of Claim and Target-level scores.
```

RDW_Subject.xsd
```diff
@@ -0,0 +1,413 @@
...
+        The code for this category.  NOTE that codes are required to be
+        unique under the same parent but are not required to be unique
+        across the entire subject.
+      -->
+      <xs:attribute name="code" use="required" type="String10"/>
+      <!--
+        Optional attribute declaring display order when displaying
+        categories.  Categories will be displayed in ascending order.
+      -->
+      <xs:attribute name="displayOrder" type="xs:integer"/>
```
malaffoon: nit: i think optional attributes should be listed after required attributes

wtritch: doh, missed one.  That's what I was shooting for.

RDW_Subject.xsd
malaffoon: that helps a lot ... thanks

PR #1225, opened 2018-04-23
Refactor aggregate rows.

agorina:
First some definitions:

**DimensionType** has a double-purpose. From the users point of view it defines default subgroups to return in the report. In terms of SQL what field to use in `GROUP BY.`
For the basic custom aggregate report DimesionType is used in the report query.

**Custom Subgroups** report allows users to build custom subgroups instead of using DimensionTypes. Form the SQL point of view, the `DimensionTYpe = Overall`. **SubGroupKey** is used to identify filters for the subgroup.

Each report (before this PR) had its own type or row: **BasicAggregateRow** and **FilteredSubgroupRow** The difference between them was that the former had a Dimension and the latter had a SubGroupKey.

In addition to that we are about to build the **LongitudinalRow**, and if we were to follow the above pattern we would need another one for the **FilteredSubgroupLongitudinal** report.

The objective of this PR is to reduce the number of different report payloads. It replaces  BasicAggregateRow and FilteredSubgroupRow with **AggregateRow**.

AggregateRow contains Dimension. For the Custom Subgroup reports, the DimenssionType is equal to a new value Custom and Code represents the SubGroupKey.

I admit that I find reusing Dimension a bit confusing, but this reduces the code footprint: 
- Instead of 4 different rows we have 2.
- Instead of 4 different report results we have 2.
- Instead of 4 repositories we have 2.
- More reused code between repositories 
- Reduced testing

I was also considering renaming Dimension/code to someting (not sure exactly what) but since it is stored as JSON in the existing report I do not think it is worth the effort since 'code' is pretty generic.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractReportRepository.java
```diff
@@ -228,21 +218,38 @@ protected Organization findOneByIdAndType(final long id, final OrganizationType
...
     }
 
     /**
-     * Represents an identity of a report row
+     * /**
+     * Provides an identity for instances of {@link AggregateRow}
      */
-    static abstract class ReportRowIdentity<I extends AggregateRow> {
-        final I reportRow;
+    private static class ReportRowIdentity {
```
agorina: It is now the same across all reports.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.mapper.ts
```diff
@@ -52,7 +52,7 @@ export class AggregateReportItemMapper {
                             overall: Subgroup): AggregateReportItem {
 
     const item = this.createRowInternal(assessmentDefinition, row, uuid);
-    const serverSubgroup = subgroups[ row.subgroupKey ];
```
agorina: I need UI folks to validate this and below change, but I have tested and it seems to work.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractReportRepository.java
agorina: This is now the same for all reports.

webapp/src/main/webapp/src/app/report/aggregate-report.ts
```diff
@@ -6,9 +6,6 @@ export interface AggregateReportRow {
   // Present on Basic query
   readonly dimension?: AggregateReportRowDimension;
 
-  // Present on FilteredSubgroup query
-  readonly subgroupKey?: string;
```
esotrope: so we got rid of subgroupKey and use dimension.code instead. Is it possible to further simplify these requests by consolidating the basic & filtered subgroup request objects? i was hoping we could remove the logic branching for `queryType` on the UI as a part of this

PR #1226, opened 2018-04-24
remove some unused methods

malaffoon:
Mostly i'm pontificating as i figure out how to move code around for SIS integration.


admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/StudentGroupBatchRepository.java
```diff
@@ -16,22 +16,6 @@
...
-    StudentGroupBatch findOne(long id);
-
-    /**
-     * * Finds one by a {@link StudentGroupBatch#digest}
-     *
-     * @param digest the digest
-     * @return the found {@link StudentGroupBatch}
-     */
-    StudentGroupBatch findOneByDigest(String digest);
-
```
malaffoon: Methods just aren't used. As i am considering moving this to RDW_Common, i thought minimizing would be a good idea.

PR #1230, opened 2018-04-24
Added models for longitudinal queries.

agorina:
Mostly renaming and breaking classes into reusable parts:

class diagram


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AggregateQuery.java
```diff
@@ -9,4 +11,24 @@
...
+
+    /**
+     * @return assessment type code included into query
+     */
+    String getAssessmentTypeCode();
+
+    /**
+     * @return school years included into query
+     */
+    Set<Integer> getSchoolYears();
```
agorina: Added methods that are expected to be available for all types of custom aggregate reports. The first two are needed for the authorization, and the rest is needed for the UI on the 'My reports' page.

PR #1236, opened 2018-04-26
Task/improve longitudinal chart models

esotrope:
Updated popover view

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
@@ -58,7 +58,7 @@
               <path class="line color-stroke"
                     [attr.d]="performancePath.pathData"></path>
 
-              <circle *ngFor="let point of performancePath.points; let j = index"
+              <circle *ngFor="let point of performancePath.points"
                       [ngClass]="point.styles"
                       r="5"
                       [attr.cx]="point.x"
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
```diff
@@ -24,50 +24,51 @@
...
+          <g *ngFor="let levelPath of chartView.performanceLevelPaths"
+             [ngClass]="levelPath.styles">
 
             <path class="color-fill"
-                  [attr.d]="d3area(area.points)"></path>
+                  [attr.d]="levelPath.pathData"></path>
 
             <path class="scale-score-area-divider"
-                  [attr.d]="d3line(area.points)"></path>
+                  [attr.d]="levelPath.dividerPathData"></path>
```
esotrope: path data is now precomputed

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.ts
```diff
@@ -0,0 +1,71 @@
+import { Organization } from '../../shared/organization/organization';
+import { Subgroup } from '../subgroup/subgroup';
+
+/**
+ * Logical modal for the longitudinal cohort chart
+ */
+export interface LongitudinalCohortChart {
```
esotrope: All the new models to pass to the chart

dsnyder-fairway:
what does the popover view reference?

dsnyder-fairway:
I think I see..2416, 2619 refer to the range of the dark blue (or gray?) background and 2465 is the number on the line graph

esotrope:
@dsnyder-fairway the lower and upper bounds of the performance level the scale score falls in. so in this case the lower and upper bound of "exceeded standard" but you make a good point that it is not entirely clear. it basically can only be deduced by the color behind the scale score

PR #1237, opened 2018-04-27
Added more tests; some changes to the backfilling logic.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractReportRepository.java
```diff
@@ -244,30 +244,43 @@ protected Organization findOneByIdAndType(final long id, final OrganizationType
     }
 
     /**
-     * /**
      * Provides an identity for instances of {@link AggregateRow}
+     * This is used to back-filled missing reports data.
+     * Note that assessment id is ignored. Since this is used only when there is
+     * no data available, it is enough to return an assessment to represent
+     * a missing report row.
```
agorina: This reduces the number of backfilled rows to one (a random one) per asmt grade/subject/year. This however does not reduce a number of the estimated counted rows.


This also `solves` another issue that I ran into that led to a decision to change migrate rules for summative. Let's talk if we still want to change it.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcLongitudinalReportRepositoryRST.java
```diff
@@ -269,6 +279,181 @@ public void itShouldReturnSchools() {
         assertThat(report.getPayload().stream().filter(r -> r.getOrganization().getOrganizationType() == OrganizationType.School).count()).isEqualTo(6);
     }
 
+    @Test
+    public void itShouldFindByQueryWithIEPDimension() {
```
agorina: Added more test for filters/dimensions. Still have more to add.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractReportRepository.java
agorina: I have second thoughts about it. It may not work for IABs...

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractReportRepository.java
```diff
@@ -236,38 +236,55 @@ protected Organization findOneByIdAndType(final long id, final OrganizationType
...
+        final Integer asmtId;
+
+        ReportRowIdentity(final AggregateRow reportRow, final String asmtTypeCode) {
+            final ActiveAssessment assessment = reportRow.getAssessment();
+            this.subjectCode = assessment.getSubjectCode();
+            this.asmtGradeCode = assessment.getGradeCode();
+            this.examSchoolYear = assessment.getExamSchoolYear();
+            this.asmtId = (AssessmentType.IAB.code().equalsIgnoreCase(asmtTypeCode)) ? assessment.getId() : null;
+            this.organization = reportRow.getOrganization();
+            this.dimension = reportRow.getDimension();
```
agorina: Changed it to account for IABs. Not sure how I feel about it. But would like to know what people think. 

And  to explain why it is needed. Without this change, Longitudinal report gets multiple assessments for the same grade/subject/schoolYear.

malaffoon: Minor issue: i think this puts business logic in this POJO; that should probably be in the layer above (repo in this case). Perhaps pass a flag in that indicates whether the asmtId should be set and move the IAB compare code out of here.

jtreuting: So if i understand this right, you are altering the assessment ID for ICA and Summative so if there are multiple it matches as equals when doing that check in the backfill logic.  

I like that it doesn't require an extra pass to determine things in memory, but it feels a little hacky.  Would it work to add this into the `equals` check, or possibly have a different equals method that is just used for backfill equality purposes if we don't want the general equality to match on this.

agorina: @malaffoon, makes sense. Will change it.

PR #1240, opened 2018-04-27
Feature/olap assessment service

esotrope:
Merge dependencies:
1. schema change to add `asmt.max_score` and `asmt.min_score`


aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAssessmentRepositoryRST.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.test.context.jdbc.Sql;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Sets.newHashSet;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Import(JdbcAssessmentRepository.class)
+@Sql(scripts = {"classpath:OlapEntitiesSetup.sql"})
+@CachingTest
+public class JdbcAssessmentRepositoryRST extends RepositoryBackedIT {
```
agorina: I see that this is a CachingTest but do not see it being cached. 
And since this is on Redshift, we should cache it.

esotrope: was going to ask about that. wondering how we want to implement the cache puts, seems like we need to do something manual since we ask for multiple assessments at a time. wouldn't want to break the request to redshift into many requests right?

agorina: Ah, good point. We should probably change it to follow the same pattern as we have for all other repositories in this project: the repository does findAll, and it is cached. The service finds in memory.

agorina:
Just FYI: the schema change is done.

esotrope:
@agorina @malaffoon @wtritch  moved filtering logic to service layer and added cache to repo layer - please re-review

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/CacheConfiguration.java
```diff
@@ -40,47 +42,52 @@
 public class CacheConfiguration {
 
     @Autowired
-    private OrganizationRepository organizationRepository;
+    private ActiveAssessmentRepository activeAssessmentRepository;
```
esotrope: alphabetized

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAssessmentRepositoryRST.java
```diff
@@ -0,0 +1,103 @@
...
+import static com.google.common.collect.Lists.newArrayList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Import(JdbcAssessmentRepository.class)
+@Sql(scripts = {"classpath:OlapEntitiesSetup.sql"})
+@CachingTest
+public class JdbcAssessmentRepositoryRST extends RepositoryBackedIT {
+
+    @Autowired
+    private JdbcAssessmentRepository repository;
```
esotrope: @agorina thanks for the help with running the RST.
I kept the direct dep on the impl class just out of practice let me know if this seems wrong - thanks

agorina: This is not right. It fails to auto-wire and create a proxi. It should be an interface.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcAssessmentRepositoryRST.java
```diff
@@ -0,0 +1,104 @@
...
+import static com.google.common.collect.Lists.newArrayList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Import(JdbcAssessmentRepository.class)
+@Sql(scripts = {"classpath:OlapEntitiesSetup.sql"})
+@CachingTest
+public class JdbcAssessmentRepositoryRST extends RepositoryBackedIT {
+
+    @Autowired
+    private AssessmentRepository repository;
```
esotrope: @agorina fixed

PR #1241, opened 2018-04-30
enable view assessments on group dashboard

dsnyder-fairway:
group dashboard

screen shot 2018-04-30 at 1 18 00 pm

click on "View Assessments"

screen shot 2018-04-30 at 1 18 19 pm

screen shot 2018-04-30 at 1 18 33 pm

click on view dashboard

screen shot 2018-04-30 at 1 18 45 pm

screen shot 2018-04-30 at 1 18 52 pm

from finding a group link on the home page

screen shot 2018-04-30 at 1 19 00 pm


webapp/src/main/webapp/src/app/groups/results/group-assessment.resolve.ts
```diff
@@ -0,0 +1,19 @@
...
+@Injectable()
+export class GroupAssessmentResolve implements Resolve<any> {
+
+  constructor(private service: GroupAssessmentService) {
+  }
+
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<any> {
+    const { groupId, schoolYear, assessmentIds } = route.params;
+    if (!assessmentIds) {
+      return this.service.getMostRecentAssessment(groupId, schoolYear);
```
dsnyder-fairway: when passing the assessmentIds we don't want the most recent assessment

webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -76,6 +76,28 @@ export class GroupDashboardComponent implements OnInit {
...
+          if (!this.currentSubject) {
+            this.measuredAssessments = measuredAssessments;
+          } else {
+            this.measuredAssessments = measuredAssessments.filter(
+              measuredAssessment => measuredAssessment.assessment.subject === this._currentSubject);
+          }
+        });
+      });
+    });
+  }
```
esotrope: how close is this to the other snippet like this on other pages? wondering if we could use a service for this

jtreuting:
when you click View Assessments and go to the group results page, are the Select Assessments are open or is that just to show that the proper 3 are preselected?  I don't think that should be open when you land on the results page.

webapp/src/main/webapp/src/app/groups/results/group-assessment.resolve.ts
```diff
@@ -0,0 +1,19 @@
...
+    }
```
esotrope: aren't we missing the return for when `assessmentIds != null` here?

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
```diff
@@ -2,7 +2,12 @@
 
 <page-heading>
   <ng-template #heading>
-    <h2>{{'groups.name' | translate}}</h2>
+    <div class="col-md-12">
+    <button *ngIf="viewGroupDashboard" class="row btn btn-default btn-sm mb-sm" (click)="viewDashboard()">
```
esotrope: nit: indent

esotrope: also, what is the `col` class without a parent `row` class doing?

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -61,22 +61,23 @@ export class AssessmentsComponent {
...
   @Output()
   export: EventEmitter<any> = new EventEmitter<any>();
 
-  showValuesAsPercent: boolean = true;
+  showValuesAsPercent = true;
   filterDisplayOptions: any = {
     expanded: true
   };
   clientFilterBy: FilterBy;
+  disableMostRecentButton = false;
```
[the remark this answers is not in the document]
jtreuting: i view the View Assessments link as a shortcut to having the assessments pre-selected, all other functionality once on the page should be the same as if you went to the page and selected a few assessments.  So I don't think this functionality should be disabled.

webapp/src/main/webapp/src/app/shared/layout/page-heading.component.html
```diff
@@ -1,12 +1,8 @@
...
-    <div class="vertical-align-bottom">
-      <ng-container *ngTemplateOutlet="heading"></ng-container>
-    </div>
+    <ng-container *ngTemplateOutlet="heading"></ng-container>
   </div>
   <div class="flex-child" *ngIf="content">
-    <div class="vertical-align-bottom">
-      <ng-container *ngTemplateOutlet="content"></ng-container>
-    </div>
+    <ng-container *ngTemplateOutlet="content"></ng-container>
```
esotrope: why the removal of the vertical align bottom?

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -173,14 +174,31 @@ export class AssessmentsComponent {
...
   }
 
+  ngOnInit() {
+    const { assessmentIds } = this.route.snapshot.params;
+    if (assessmentIds) {
+      this.assessmentProvider.getAvailableAssessments().subscribe((value) => {
+        const selectedAssessments = value.filter((assessment) => assessmentIds.split(',').indexOf(assessment.id.toString()) >= 0);
+        selectedAssessments.forEach((assessment) => {
+          assessment.selected = true;
+          this.selectedAssessmentsChanged(assessment);
```
esotrope: do we really want to call this directly? basically this is simulating many assessment clicks?

jtreuting: are you thinking that instead we would have an API that would take multiple assessment IDs so it's a single API call?

webapp/src/main/webapp/src/app/groups/results/group-assessment.resolve.ts
dsnyder-fairway: no. I felt adding a return type that allowed for multiple assessments would add unneeded complexity. Instead I handle loading the assessments elsewhere

webapp/src/main/webapp/src/app/shared/layout/page-heading.component.html
dsnyder-fairway: yeah I know. I wasn't really sure what the styling for vertical align bottom was doing. Checking out the pages before and after appeared to have no differences after resizing the window. This was needed to avoid and issue with placing the button before the header

dsnyder-fairway:
@jtreuting the reason why I have the assessments expanded is because I was having some trouble to get this to behave correctly after the page loads and not having the assessments expanded. I'll work on updating this. Closing for now

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
esotrope: not sure what i'm suggesting. just kind of probing for more info on why this is the way it is. looked a little suspect is all.

webapp/src/main/webapp/src/app/shared/layout/page-heading.component.html
esotrope: button before the header? not sure i follow.
the styling difference will be apparent on pages with h1/h2 (taller) headers

webapp/src/main/webapp/src/app/groups/results/group-assessment.resolve.ts
esotrope: not sure that answered my concern. the method has return type Observable and the method body only returns an observable when assessmentIds == null. shouldn't we have an explicit return statement for when they do not equal null? maybe `return Observable.of(null)`?

dsnyder-fairway: since closing this PR I reverted changes to this file

webapp/src/main/webapp/src/app/groups/results/group-results.component.html
dsnyder-fairway: getting rid of `col` class and surrounding `div`

webapp/src/main/webapp/src/app/shared/layout/page-heading.component.html
dsnyder-fairway: reverted changes to this since closing my PR

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
dsnyder-fairway: I replaced `selectedAssessmentsChanged ` with `loadAssessmentExam` here

PR #1244, opened 2018-05-01
Feature/dwr 1698 has wer

jtreuting:
returning a flag for if an assessment has any WER items.  this allows the results pages to know ahead of time if it is appropriate to have an option enabled for Writing Trait Scores aggregate view.  this is needed because some states won't be using the WER section of an ICA or Summative and so we could know that ahead of time if that item isn't loaded as part of the package.  plus it makes the experience better for ELA IABs that are not the PT one.

NOTE: still need to add integration tests.  will do before merging

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
@@ -121,7 +121,7 @@ public class Assessment {
     }
 
     /**
-     * @return <code>true</code> if the assessment has one or more WER items
+     * @return <code>true</code> if the assessment has one or more WER items. <code>null</code> if there wasn't a check done
      */
     public Boolean isWerItem() {
         return werItem;
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
@@ -21,7 +21,7 @@ public class Assessment {
     private int schoolYear;
     private List<Integer> cutPoints;
     private List<String> claimCodes;
-    private boolean werItem;
+    private Boolean werItem;
 
     private Assessment() {
     }
@@ -123,7 +123,7 @@ public class Assessment {
     /**
      * @return <code>true</code> if the assessment has one or more WER items
      */
-    public boolean isWerItem() {
+    public Boolean isWerItem() {
         return werItem;
     }
 
@@ -160,7 +160,7 @@ public class Assessment {
         private int schoolYear;
         private List<Integer> cutPoints;
         private List<String> claimCodes;
-        private boolean werItem;
+        private Boolean werItem;
 
         public Builder id(final long id) {
             this.id = id;
@@ -207,7 +207,7 @@ public class Assessment {
             return this;
         }
 
-        public Builder werItem(final boolean werItem) {
+        public Builder werItem(final Boolean werItem) {
             this.werItem = werItem;
             return this;
         }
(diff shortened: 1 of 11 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
```diff
@@ -119,6 +120,13 @@ public int getSchoolYear() {
         return claimCodes;
     }
 
+    /**
+     * @return <code>true</code> if the assessment has one or more WER items
+     */
+    public boolean isWerItem() {
```
jtreuting: i tried to make this `hasWerItem` but apparently for the jackson serialization (at least without specific annotations) it needs to start with is for a boolean or else it doesn't serialize it into the JSON response and it is just missing completely

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -258,7 +258,9 @@ sql:
           a.max_score,
           a.cut_point_1,
           a.cut_point_2,
-          a.cut_point_3
+          a.cut_point_3,
+          EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item
```
jtreuting: from my limited testing this seems to perform fine.  the `EXPLAIN` says that it is using the `idx__item__asmt_natural_id` index.  might be helpful to include an index that takes into account item type, but not sure it's needed.

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
```diff
@@ -33,4 +34,8 @@ export class Assessment {
     return this.subject === 'ELA';
   }
 
+  get hasWerItem(): boolean {
+    return this.werItem;
```
jtreuting: yes this is not really needed, but it feels nicer to call `hasWerItem` to me cause it explains better what this is and jackson was mean to me and didn't let me do it, so i turned to angular to make it right

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -126,7 +126,7 @@
                 <li role="menuitem"
                     *ngIf="writingTraitScoresView.display">
                   <a class="dropdown-item"
-                     popover="{{ (writingTraitScoresView.disabled ? 'assessment-results.writing-trait-scores-disabled' : 'assessment-results.no-writing-trait-scores') | translate }}"
+                     popover="{{ (!hasExamsAfterMinimumItemYear ? 'assessment-results.no-writing-trait-scores' : 'assessment-results.writing-trait-scores-disabled') | translate }}"
```
jtreuting: if it's an old year show that message, otherwise the old assessments don't have items and won't have WER items and therefore would always give the message that there aren't any instead of it being older.

reporting-service/src/main/resources/application.sql.yml
malaffoon: Guess the snippet stuff is working if you only had to change one SQL statement!

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Assessment.java
esotrope: maybe `isWerItemHolder`? `isWerItemAssessment`? Not sure what's best but i would avoid putting "`isWerItem`" in the `Assessment` class because assessments aren't items

webapp/src/main/webapp/src/app/assessments/model/assessment.model.ts
```diff
@@ -6,6 +6,7 @@ export class Assessment {
   subject: string;
   claimCodes: string[];
   cutPoints: number[];
+  werItem: boolean;
```
esotrope: since we're in js land now you can change this to `hasWerItem`

jtreuting: that is a very valid point :) we already have the mapper to map between them.

PR #1248, opened 2018-05-01
correcting table sometimes failing to load

dsnyder-fairway:


webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.html
```diff
@@ -142,8 +142,7 @@
                          [standardError]="rowData.exam.standardError"></scale-score>
 
             <div *ngSwitchCase="'claim'">
-              <span *ngIf="rowData.exam.claimScores[column.index].level">{{ 'common.assessment-type.iab.performance-level.' + (rowData.exam.claimScores[column.index] ? rowData.exam.claimScores[column.index].level.toString() : 'missing') + '.name' | translate }}</span>
-              <span *ngIf="!rowData.exam.claimScores[column.index].level">-</span>
+              <span>{{ 'common.assessment-type.iab.performance-level.' + (rowData.exam.claimScores[column.index] ? rowData.exam.claimScores[column.index].level.toString() : 'missing') + '.name' | translate }}</span>
```
dsnyder-fairway: @esotrope this corrects the issue you were seeing

webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.html
```diff
@@ -8,18 +8,18 @@
                              [latestExam]="studentHistoryCard"
                              [exams]="exams"
                              (selectedAssessment)="onCardSelection($event)"></student-assessment-card>
-    <div *ngIf="i % 3 === 0">
+    <ng-container *ngIf="i % 3 === 0 && !hideTableForIndex(i)">
```
jtreuting: seems like this would be clearer if it was `showTableForIndex(i)` (i'm stealing @esotrope comment on this i'm sure)

dsnyder-fairway: agreed. I'll rename

webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.html
jtreuting: maybe this logic should be in a helper method, what do you think

webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.html
esotrope: nope my comment was going to be - why is table hiding conditional on card index? i dont follow the logic

dsnyder-fairway: the idea is that we have a table for every three cards, and we display that table if any of those 3 cards are selected

webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.html
dsnyder-fairway: I'm leaning towards a no, but I'm mostly indifferent. I'd say yeah if we were going to re-use this for sure, but I don't see the need. Then again, if anyone else cares to chime in or you feel it should I'll make the change

webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.html
esotrope: gotcha does the table hold more than one cards data? or this is just about display positioning? this logic locks us into 3 cards a row i guess

dsnyder-fairway: just about display positioning

jtreuting: also @esotrope we locked the CSS grid to 3 columns to make them align.  we worked on a CSS grid solution that would work with flexible grid columns but didn't come up with something better

esotrope: gotcha - thanks for the background - been a bit out of the loop on these dashboard changes

PR #1250, opened 2018-05-01
provide links to the group-dashboard

dsnyder-fairway:
links from home page

screen shot 2018-05-01 at 4 09 30 pm

links from groups page

screen shot 2018-05-01 at 4 09 48 pm

click 

screen shot 2018-05-01 at 4 10 04 pm

screen shot 2018-05-01 at 4 10 11 pm

after clicking view assessments from group-dashboard

screen shot 2018-05-01 at 4 10 18 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/groups.component.html
+++ b/webapp/src/main/webapp/src/app/groups/groups.component.html
@@ -60,6 +60,13 @@
               </a>
             </div>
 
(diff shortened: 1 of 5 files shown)


jtreuting:
It should say IAB Dashboard instead of just Dashboard.  then it is good for now so we can demo it and talk thru the flow with them.

PR #1251, opened 2018-05-01
Task/refactor translation support

wtritch:
This task refactors our translation implementation to remove the concept of frontend/backend namespaces.  This is in support of configurable subjects where display labels may be scoped to frontend (UI), backend (report generation), or both.

Some positive results of this change is that all of our translation messages are now represented in JSON format.  We now also support a single external translation .json file (no need to install both an es.messages AND es.json translation file.)

The base implementation is a composite TranslationProvider: 
embedded xx.json overridden by 
database translations overriden by
optional external (configuration repo)

The composite translation provider is used to back a MessageSource that has a fallback to the English translations in the case that the requested locale's language is not present for a particular message key.

The UI makes a call to the embedded front-end xx.json file, which is then overridden by the call to /api/translations/xx which is backed by the composite TranslationProvider.


common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+        providers.add(new JsonBackedTranslationProvider("classpath:/i18n/", resourceLoader));
+        //Add repository translations
+        providers.add(new RepositoryBackedTranslationProvider(translationRepository));
+
+        if (isNotBlank(settings.getTranslationLocation())) {
+            //Add external translations
+            providers.add(new JsonBackedTranslationProvider(settings.getTranslationLocation(), resourceLoader));
+        }
+
+        return new DelegatingTranslationProvider(providers);
```
wtritch: This configuration declares the override order for translation providers.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProvider.java
```diff
@@ -0,0 +1,33 @@
...
+public class DelegatingTranslationProvider implements TranslationProvider {
+
+    private final List<TranslationProvider> providers;
+
+    public DelegatingTranslationProvider(final List<TranslationProvider> providers) {
+        this.providers = providers;
+    }
+
+    @Override
+    @Cacheable(value = "mergeTranslations", key = "#locale.getLanguage()")
```
wtritch: I'm caching the combined translations for a given language.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSource.java
```diff
@@ -0,0 +1,34 @@
...
+    protected MessageFormat resolveCode(final String code, final Locale locale) {
+        final Map<String, String> messages = provider.getTranslationsByCode(locale);
+        if (messages.containsKey(code)) {
+            return createMessageFormat(messages.get(code), locale);
+        }
+
+        //Fall back to the default locale
+        return DefaultLocale.getLanguage().equalsIgnoreCase(locale.getLanguage())
+                ? null
+                : resolveCode(code, DefaultLocale);
```
wtritch: This is where the MessageSource falls back to the English translation if a value for the specified locale doesn't exist.

common/src/main/resources/i18n/en.json
```diff
@@ -0,0 +1,971 @@
+{
+  "report": {
```
wtritch: This is the old messages.properties resource bundle represented as JSON.

report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -19,7 +19,7 @@
                         </span>
                     </div>
                     <div class="h4"
-                         th:text="#{('report.iab.title.grade.' + ${assessment.gradeCode})(${#strings.toString(exam.schoolYear - 1)}, ${support.getLastTwoDigits(exam.schoolYear)})}"></div>
+                         th:text="#{('report.iab.grade.title.' + ${assessment.gradeCode})(${#strings.toString(exam.schoolYear - 1)}, ${support.getLastTwoDigits(exam.schoolYear)})}"></div>
```
wtritch: I had to invert these keys, we can't have:
```
report.iab.title=blah
report.iab.title.grade=blah
```
which can't be represented as JSON

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProvider.java
esotrope: can we change the cache name to `mergedTranslations` or just `translations`?

wtritch: sure

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonBackedTranslationProvider.java
```diff
@@ -0,0 +1,82 @@
...
+    public Map<String, String> getTranslationsByCode(final Locale locale) {
+        if (locale == null) {
+            return Collections.emptyMap();
+        }
+
+        final ObjectNode objectNode = loadJson(locale);
+        final Properties properties = new Properties();
+        try {
+            final JavaPropsGenerator generator = propertiesMapper.getFactory().createGenerator(properties);
+            generator.writeTree(objectNode);
```
esotrope: what a nice utility

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JdbcTranslationRepository.java
```diff
@@ -1,40 +1,35 @@
...
+    private String findAllForLanguageQuery;
 
     @Autowired
     JdbcTranslationRepository(final NamedParameterJdbcTemplate template) {
         this.template = template;
     }
 
     @Override
-    public List<Translation> findAllForLanguageAndNamespace(final String languageCode, final TranslationNamespace namespace) {
+    public List<Translation> findAllForLanguage(final String languageCode) {
```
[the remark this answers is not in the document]
malaffoon: I know it just sort of happened but why can't the `TranslationRepository` interface be exactly the `TranslationProvider` instead? I'll poke around more but a follow-up question is why have `Translation` and `Map<String,String>`?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonBackedTranslationProvider.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+/**
+ * This translation provider accepts a location accessible via {@link ResourceLoader}
+ * and loads <pre>{translationLocation}xx.json</pre> for the requested locale language.
+ * Translation key and value pairs are parsed from JSON and returned as a flat
+ * key and value map. (e.g. {"my":{"json":{"path":"my_value"}}} becomes "my.json.path=my_value")
+ */
+public class JsonBackedTranslationProvider implements TranslationProvider {
```
malaffoon: nit: i can't come up with a better suggestion so i guess i shouldn't mention it but this is more than json backed, it loads resources from somewhere; perhaps `JsonResourceTranslationProvider`?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonBackedTranslationProvider.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public JsonBackedTranslationProvider(final String translationLocation,
+                                         final ResourceLoader resourceLoader) {
+        this.translationLocation = translationLocation;
+        this.resourceLoader = resourceLoader;
+        this.propertiesMapper = new JavaPropsMapper();
+    }
+
+    @Override
+    public Map<String, String> getTranslationsByCode(final Locale locale) {
```
malaffoon: i'm struggling a little bit with the contract for this method: it is supposed to return an empty map if it can't find the locale? It will never throw an exception or anything?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonBackedTranslationProvider.java
```diff
@@ -0,0 +1,82 @@
...
+        if (locale == null) {
```
malaffoon: what if it is blank?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/RepositoryBackedTranslationProvider.java
```diff
@@ -7,20 +7,19 @@
...
-    public SimpleTranslationService(final TranslationRepository repository, final TranslationNamespace namespace) {
+    public RepositoryBackedTranslationProvider(final TranslationRepository repository) {
         this.repository = repository;
-        this.namespace = namespace;
     }
 
+    @Override
     public Map<String, String> getTranslationsByCode(final Locale locale) {
-        return repository.findAllForLanguageAndNamespace(checkNotNull(locale).getISO3Language(), namespace)
+        return repository.findAllForLanguage(checkNotNull(locale).getISO3Language())
```
malaffoon: Should we be switching this to the 2-letter code? It has bothered me (a teeny bit) that the file is `en.json` but the entries in the database table are `eng`. Do we need to propagate that discrepancy?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSource.java
malaffoon: It doesn't further fall-back to the code itself? How is a null MessageFormat handled upstream?

It seems like that `return null` should have a log message (INFO/WARN?) being emitted so that it can be detected in the log file (since the users don't seem to tell us about weird stuff on the screen).

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/RepositoryBackedTranslationProvider.java
```diff
@@ -7,20 +7,19 @@
...
                 .stream().collect(toImmutableMap(Translation::getCode, Translation::getValue));
```
malaffoon: per comment above, converting from Translation to Map should be unnecessary

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProviderTest.java
```diff
@@ -0,0 +1,46 @@
...
+                .containsEntry("key_B", "value b_b")
+                .containsEntry("key_C", "value b_c");
+
+        final DelegatingTranslationProvider delegatingB_A = new DelegatingTranslationProvider(ImmutableList.of(providerB, providerA));
+        assertThat(delegatingB_A.getTranslationsByCode(Locale.US))
+                .hasSize(3)
+                .containsEntry("key_A", "value a_a")
+                .containsEntry("key_B", "value a_b")
+                .containsEntry("key_C", "value b_c");
+    }
```
malaffoon: nit: use case of child provider return nothing, throw exception, etc.?

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/RepositoryBackedTranslationProviderTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class RepositoryBackedTranslationProviderTest {
+
+    private static final String ns1 = "ns1";
+    private static final String ns2 = "ns2";
+    private static final List<Translation> englishTranslations = newArrayList(
+            new Translation("a", "A"),
+            new Translation("b", "B")
+    );
+    private static final List<Translation> japaneseTranslations = newArrayList(
+            new Translation("c", "C")
```
malaffoon: oh come on, have some fun  `漢字`

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationJsonService.java
```diff
@@ -0,0 +1,21 @@
...
+public interface TranslationJsonService {
+
+    /**
+     * Retrieve the locale-specific messages for the given locale.
+     *
+     * @param locale    A target locale
+     * @param includeUI true to include UI translation key/values
+     * @return  The messages for the given locale
+     */
+    ObjectNode getLocaleMessages(final Locale locale, final boolean includeUI);
```
malaffoon: Hmm. The includeUI thing feels weird. At least it's isolated to this level.

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Locale;
+import java.util.Map;
+import java.util.Properties;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Maps.newHashMap;
+
+/**
+ * This TranslationJsonService provides a json representation of translation key-value pairs
+ * using a TranslationProvider.
```
malaffoon: Perhaps javadocs explaining (briefly) why we aren't just using views for this transformation? Is it because we need more control over the object mapper or something?

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
```diff
@@ -0,0 +1,94 @@
...
+        final List<String> children = newArrayList(objectNode.fieldNames());
+        children.sort(Comparator.naturalOrder());
+        for (final String child : children) {
+            final JsonNode childNode = objectNode.remove(child);
+            orderNode(childNode);
+            objectNode.set(child, childNode);
+        }
+    }
+
+    private Properties getUIMessages(final Locale locale) {
```
malaffoon: Can/should the result of this be cached since they won't change? Premature?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProvider.java
```diff
@@ -0,0 +1,33 @@
...
+public class DelegatingTranslationProvider implements TranslationProvider {
+
+    private final List<TranslationProvider> providers;
+
+    public DelegatingTranslationProvider(final List<TranslationProvider> providers) {
+        this.providers = providers;
+    }
+
+    @Override
+    @Cacheable(value = "translations", key = "#locale.getLanguage()")
```
malaffoon: This means changes to the database or remote config won't be seen without a restart?

webapp/src/test/java/org/opentestsystem/rdw/reporting/i18n/TranslationControllerIT.java
```diff
@@ -48,7 +48,23 @@ public void itShouldUseServiceToGetTranslations() throws Exception {
                 .andExpect(jsonPath("$.parent.childA", equalTo("valueA")))
                 .andExpect(jsonPath("$.parent.childB", equalTo("valueB")));
 
-        verify(translationService).getLocaleMessages(Locale.ENGLISH);
+        verify(translationService).getLocaleMessages(Locale.ENGLISH, false);
+    }
+
+    @Test
+    public void itShouldOptionallyInlcudeUITranslations() throws Exception {
```
malaffoon: nit: typo in test name

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * This translation provider delegates to a collection of translation providers, overriding
+ * values in the order that the providers and given.
+ * (e.g. the last provider in the list wins for a given translation key)
+ */
+public class DelegatingTranslationProvider implements TranslationProvider {
```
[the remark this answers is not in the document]
wtritch: um... `Composite` is probably better.  Changed

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSource.java
wtritch: I wanted to explicitly allow for a null Locale, so that our MessageSource will fall back to using English for any locale (even a missing locale)

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/RepositoryBackedTranslationProvider.java
jtreuting: I vaguely remember the accessibility config XML file using 3 letter codes.  so if we make a change we'll need to consider that import as well.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettings.java
```diff
@@ -154,7 +154,7 @@ public void setTransferAccessEnabled(final boolean transferAccessEnabled) {
     }
 
     /**
-     * This is the location/prefix of translation files (messages_es.properties, es.json, etc)
+     * This is the location/prefix of translation files (es.json, en.json, etc)
```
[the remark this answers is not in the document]
wtritch: s&r text?  Sorry... I'm a little slow today.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProvider.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    @Cacheable(value = "translations", key = "#locale.getLanguage()")
+    public Map<String, String> getTranslationsByCode(final Locale locale) {
+        final Map<String, String> translations = new HashMap<>();
+        for (final TranslationProvider provider : providers) {
+            translations.putAll(provider.getTranslationsByCode(locale));
+        }
+        return ImmutableMap.copyOf(translations);
```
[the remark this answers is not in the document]
wtritch: I started off with that, and was quickly reminding that you can't override values in an ImmutableMap.Builder.
```
builder
  .put("key1", "value1")
  .put("key1", "value2")
```
throws an exception.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonBackedTranslationProvider.java
wtritch: I want it to return an empty map if the locale can't be found (or doesn't exist) so that we can have fallback to English in the MessageSource without have to trap "expected" exceptions.  I could see an argument for logging a Debug message if the requested locale isn't present (the map is empty) in the composite translation provider?

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/RepositoryBackedTranslationProvider.java
wtritch: I'm not sure... @esotrope felt pretty strongly about keeping the three-letter code in the DB.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/RepositoryBackedTranslationProvider.java
wtritch: fixed

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProviderTest.java
wtritch: doh, added

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/RepositoryBackedTranslationProviderTest.java
wtritch: =P

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationJsonService.java
wtritch: Yeah... I originally put it in because I wanted some way to get a single en.json that I could translate into mock spanish for testing... then I documented and left it in because we'd need something like that for production anyway.  Open to suggestions =)

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
wtritch: I feel like that's premature, especially since the value is stored on the local file-system (which is slower than memory but not horribly non-performant)

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProvider.java
wtritch: Correct.  Unless the pod is configured for remote/shared caching (say, via redis.)  But hopefully we'd add a clean mechanic for clearing the cache before recommending that.

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
wtritch: (But this should not be called often, it should only be called by DevOps when they want to get a translation source file to provide an additional language.)

webapp/src/test/java/org/opentestsystem/rdw/reporting/i18n/TranslationControllerIT.java
wtritch: doh, fixed.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonBackedTranslationProvider.java
wtritch: good name, changed

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
wtritch: I'm debating where to put those javadocs... probably on the implementation, because this interface doesn't know or care where the translations are coming from or what they look like.

wtritch:
```
/**
 * This TranslationJsonService provides a json representation of translation key-value pairs
 * using a TranslationProvider.
 * Since our translations are provided as a flat map of "some.prop.path=some_value"
 * we need to transform the flat properties into a hierarchical structure for proper
 * JSON representation using a Jackson JavaPropsMapper .
 */
@Service
class TranslationProviderBackedTranslationJsonService implements TranslationJsonService {
```

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/DelegatingTranslationProvider.java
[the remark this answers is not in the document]
malaffoon: Doh, didn't know that.

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
malaffoon: Oh ... in that case, np. I thought maybe this was being called by the front-end to load up all the translations or something.

PR #1252, opened 2018-05-01
maps query type to expected server values

esotrope:


webapp/src/main/webapp/src/app/report/aggregate-report-request.ts
```diff
@@ -29,17 +29,14 @@ export interface AggregateReportQuery {
...
+  readonly queryType: 'Basic' | 'FilteredSubgroup' | 'BasicLongitudinal' | 'FilteredSubgroupLongitudinal';
 
   // Basic query type params
   readonly studentFilters?: StudentFilters;
 
   // FilteredSubgroup query type params
   readonly subgroups?: {[key: string]: StudentFilters};
 
-  // Needed for mapping back into form state
-  readonly reportType: 'GeneralPopulation' | 'LongitudinalCohort';
```
esotrope: the ui form tracks longitudinal/general pop and basic/filtered subgroup separately while the server side does not

agorina: I think I like the front end model with two pairs of values more than the back end.  And I think it would fit better for the Claim report. Maybe we should do the change in the back end instead?

I know we have talked about it back and forth, but after working on the Claim report yesterday I could see how 'Claim' report could be just a new report type with this approach.

esotrope: works for me : j

agorina: Yeah, I know. I removed my comment before I saw your reply (sorry about it).
There is a reason why it works better for the back end this way. 
Although I still feel that it could be improved on the back end side, not sure exactly how yet.

If this enables Longitudinal in QA, let's go ahead with this change and see where it takes us?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -253,6 +256,42 @@ export class AggregateReportRequestMapper {
...
+      : `${queryType}Longitudinal`;
+  }
+
+  /**
+   * Parses server queryType into client queryType and reportType fields
+   *
+   * @param {AggregateReportQuery} serverQueryType the server query
+   * @returns {{queryType: string; reportType: string}} the form types
+   */
+  private parseTypes(serverQueryType: string): {queryType: string, reportType: string} {
```
wtritch: I don't have a better solution, but this is starting to get difficult to follow... =/

PR #1253, opened 2018-05-02
First draft of the claim query

agorina:


aggregate-service/src/main/resources/claim-report-template.sql.yml
```diff
@@ -0,0 +1,287 @@
...
+                      fe.grade_id,
+                      g.code
+
+      # ------------ filters, must be added on when a dimension and filter do not overlap --------------------------------------------------
+            completeness:
+               clauses:
+                 where: >-
+                     fe.completeness_id in (:completeness_ids)
+
+            administration_condition:
```
jtreuting: i'm seeing different casing and word delimiters: "administration_condition", "StudentEnrolledGrade", "MigrantStatus".  can we make those consistent

agorina: I will change `completeness` and `administration_condition` to be consistent with the rest (and in all three templates).

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
```diff
@@ -24,6 +25,11 @@ public Builder copy() {
         return subjectCodes == null ? newHashSet() : newHashSet(subjectCodes);
     }
 
+    // optional for claim report, can be null
```
esotrope: why not make this a javadoc?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
```diff
@@ -81,7 +89,7 @@ public Builder copy(final AggregateReportQuery query) {
             subjectCodes = query.getSubjectCodes();
             schoolYears = query.getSchoolYears();
             assessmentGradeCodes = query.getAssessmentGradeCodes();
-
+            claimCodes = query.getClaimCodes();
```
esotrope: are these immutable collections being copied? at what point is a copy of the collection made?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
agorina: I do not think I have a good answer. If people agree on reusing the models, I will change this to javadoc.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
```diff
@@ -14,6 +14,7 @@
     private Set<Integer> schoolYears;
     private Set<String> assessmentGradeCodes;
     private Set<String> subjectCodes;
+    private Set<String> claimCodes;
```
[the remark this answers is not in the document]
esotrope: i've got no prob with the nullable field

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportQuery.java
agorina: no, not right now. If it is a keeper, I will make a copy in here.

aggregate-service/src/main/resources/claim-report-template.sql.yml
malaffoon: Hmm. It is a convention to use "kabob" casing in yaml files for spring properties. I realize these aren't *exactly* normal spring properties but i wonder if we should do that. That is, `administration-condition`, `gender-filter`, etc.

agorina: The reason they are in this format is because in the java code we build them dynamically from the enum names.  `completeness` and `administration_condition` are the only two that do not have matching enums.  Of course we could change the java code to remap the names. Not sure if this effort is justifiable.

PR #1254, opened 2018-05-02
removed restriction of only ICA and IAB on assessment items.  needed …

jtreuting:
…for Writing Trait Scores aggregate for Summative

There was a hard-coded restriction for IAB and ICA since it did not allow for the WER items to be returned.  Since we will be restricting sensitive data for summative items during migrate (no responses and no answer keys) I'm thinking we can return these safely.  

The alternative is just to allow this for WER items which seems like an odd restriction, if it's fine for that then it's probably fine for all.

Thoughts?


malaffoon:
I agree that restricting it based on the type of item seems wrong.

PR #1259, opened 2018-05-02
Bump schema version and account for translation schema changes.

wtritch:


common/src/main/resources/common.sql.yml
```diff
@@ -2,7 +2,7 @@ sql:
   translation:
     findAllForLanguage: >-
         select label_code, label
-        from translation
+        from accommodation_translation
```
wtritch: account for the `translation` -> `accommodation_translation` table renaming.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -994,7 +994,7 @@
     }
   },
   "html": {
-    "system-news": ""
+    "system-news": "<h2 class=\"blue-dark h3 mb-md\">Note</h2><div class=\"summary-reports-container mb-md\"><p>Item level data and session IDs are not available for tests administered prior to the 2017-18 school year.</p></div>"
```
wtritch: Since we now heavily discourage using the DB for custom translations, I've copied this from the database initialization script into our embedded messages.

jtreuting: this feels like something specific to CA implementation.  does it belong in this main version of the en.json file?

wtritch: Yes, however we also have:
```
"footer": "© The Regents of the University of California – Smarter Balanced Assessment Consortium",
```
which will also need to be overridden for any non-CA tenant.  I imagine there are probably a couple of others in there as well...  Thoughts?

PR #1261, opened 2018-05-02
hide estimated row count until capable of providing an estimate

dsnyder-fairway:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -409,6 +410,12 @@ export class AggregateReportFormComponent {
     };
   }
 
+  capableOfRowEstimation(): boolean {
+    return (!Utils.isNullOrEmpty(this.settings.schools) || !Utils.isNullOrEmpty(this.settings.districts))
+      && !Utils.isNullOrEmpty(this.settings.subjects)
+      && (!Utils.isNullOrEmpty(this.settings.generalPopulation.schoolYears) || this.settings.longitudinalCohort.toSchoolYear > 0);
```
esotrope: not sure why `form.valid` doesn't cut it. this may be an angular form refresh/revalidation issue.
Also don't think this logic works for statewide only reports.
Also settings.subjects should never be null or empty

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -511,7 +511,7 @@ <h3>{{'aggregate-report-form.section.review-heading' | translate}}</h3>
                 </div>
               </ng-container>
               <ng-template #estimatedRowCountSpinner>
-                <p class="text-muted small">
+                <p *ngIf="capableOfRowEstimation()" class="text-muted small">
```
esotrope: i think this belongs around the  `<ng-container ngTemplateOutlet=estimatedRowCountSpinner`

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
dsnyder-fairway: it may be an issue with initial page load as well. We don't show errors on the page until they have selected a grade and then unselected it, for instance.

dsnyder-fairway: removed settings.subjects and included statewide condition in there.

dsnyder-fairway: I was including settings.subjects check since we are validating this has been set on the backend. We will have to keep this in mind if we ever allow for subjects to be null or empty

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
dsnyder-fairway: I included it in the template. Reason being is this is called from `<ng-container *ngIf="estimatedRowCount != null; else estimatedRowCountSpinner">`

PR #1265, opened 2018-05-04
reduce padding between student count and aggregate graph

dsnyder-fairway:


webapp/src/main/webapp/src/styles.less
```diff
@@ -1242,7 +1242,7 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
     display: grid;
     grid-gap: @spacer / 4;
     grid-template-columns: repeat(3, 1fr);
-    padding-top: @spacer * 2;
+    padding-top: @spacer * 0.88;
```
dsnyder-fairway: evaluates to 14.96px (spacer is 17px)

PR #1267, opened 2018-05-05
improves chart line toggle ui

esotrope:
Some improvements to the chart series toggle UI

The toggles now have the same default sort as the table rows.

image

Also, selection of one highlights only that one. Additional selections will be added to the view instead of incrementally taking things away.

image

I would still like to make an iteration where organizations and subgroups can be toggled independently but for now i think this is an improvement so i'd like to get it out there

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -4,7 +4,7 @@ import { AggregateReportItem } from './aggregate-report-item';
 import { byNumber, byString, Comparator, join, ranking } from '@kourge/ordering/comparator';
 import { ColorService } from '../../shared/color.service';
 import { AssessmentDefinition } from '../assessment/assessment-definition';
-import { District, OrganizationType, School } from '../../shared/organization/organization';
+import { OrganizationType } from '../../shared/organization/organization';
 import { Utils } from '../../shared/support/support';
 import { AggregateReportOptions } from '../aggregate-report-options';
 import { PerformanceLevelDisplayTypes } from '../../shared/display-options/performance-level-display-type';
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -24,30 +25,6 @@ export const IdentityColumnOptions: string[] = [
   'dimension'
 ];
 
-const OverallDimensionType: string = 'Overall';
```
esotrope: moved sorting code to the support method file as it is common now

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
```diff
@@ -6,16 +6,13 @@
...
             class="mb-xs">
-          <button class="btn chart-series color-{{index % 3}} text-left"
-                  [ngClass]="{'chart-series-hidden': !performancePath.visible}"
-                  (click)="performancePath.visible = !performancePath.visible">
-            <div>
-              {{performancePath.organization.name}}
-            </div>
-            <div class="small">
-              {{'common.organization.type.' + performancePath.organization.type | translate}}
-            </div>
```
esotrope: this change does drop the org type name from the display

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.ts
```diff
@@ -219,8 +232,6 @@ export class LongitudinalCohortChartComponent implements OnInit {
 
     this._chartView = <ChartView>{
       performancePaths: this._chart.organizationPerformances
-        // TODO allow different subgroup selection
-        .filter(performance => performance.subgroup.id === 'Overall:')
```
esotrope: removed overall subgroup filter

webapp/src/main/webapp/src/styles.less
```diff
@@ -350,10 +350,11 @@ sb-button-typeahead .input-group-btn > .btn,
 }
 
 .label-max-width {
+  white-space: nowrap;
   text-overflow: ellipsis;
   text-align: left;
   overflow: hidden;
-  max-width: 330px;
+  max-width: 300px;
```
esotrope: didn't mean to change this - going to revert it

agorina:
Looks very nice!

PR #1269, opened 2018-05-07
updates popovers on longitudinal cohort charts

dsnyder-fairway:
screen shot 2018-05-07 at 10 06 12 am

screen shot 2018-05-07 at 10 06 06 am

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
@@ -78,7 +78,9 @@
 </div>
 
 <ng-template #dataPointPopoverTemplate let-point>
-  <span class="small mr-xs">{{point.levelRange.scaleScoreRange.minimum}}</span>
-  <span class="label label-lg {{point.levelRange.level.color}}">{{point.scaleScore}}</span>
-  <span class="small ml-xs">{{point.levelRange.scaleScoreRange.maximum}}</span>
+  <div class="text-center">
+    <span class="label label-lg {{point.levelRange.level.color}}">{{point.scaleScore}}</span>
+    <span>{{ 'longitudinal-cohort-chart.error-band' | translate: {errorBand: point.standardError} }}</span>
+  </div>
+  <div class="mt-xs text-center">{{point.levelRange.level.name}}</div>
 </ng-template>
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
@@ -80,7 +80,7 @@
 <ng-template #dataPointPopoverTemplate let-point>
   <div class="text-center">
     <span class="label label-lg {{point.levelRange.level.color}}">{{point.scaleScore}}</span>
-    <span>{{ 'longitudinal-cohort-chart.error-band' | translate: {errorBand: point.standardError} }}</span>
+    <span>{{ 'common.error-band-value' | translate: {errorBand: point.standardError | number:'1.0-0'} }}</span>
   </div>
   <div class="mt-xs text-center">{{point.levelRange.level.name}}</div>
 </ng-template>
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
```diff
@@ -78,7 +78,9 @@
 </div>
 
 <ng-template #dataPointPopoverTemplate let-point>
-  <span class="small mr-xs">{{point.levelRange.scaleScoreRange.minimum}}</span>
-  <span class="label label-lg {{point.levelRange.level.color}}">{{point.scaleScore}}</span>
-  <span class="small ml-xs">{{point.levelRange.scaleScoreRange.maximum}}</span>
+  <div class="text-center">
+    <span class="label label-lg {{point.levelRange.level.color}}">{{point.scaleScore}}</span>
+    <span>{{ 'longitudinal-cohort-chart.error-band' | translate: {errorBand: point.standardError} }}</span>
```
jtreuting: i'm assuming we have a way to do error band already since we display it in a few places already.  seems like a good place for a common error band translate value

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.mapper.ts
```diff
@@ -87,7 +89,8 @@ function createStubOrganizationPerformances(count: number,
 
       line.push(<YearGradeScaleScore>{
         yearGrade: yearGrade,
-        scaleScore: Math.floor(previous + 50 + 25 * Math.random())
+        scaleScore: Math.floor(previous + 50 + 25 * Math.random()),
+        standardError: Math.floor(Math.max(maximumScaleScore - this.scaleScore, minimumScaleScore + this.scaleScore))
```
jtreuting: i'm guessing this is just the sample data.  @esotrope do we need this in there anymore?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1100,7 +1100,8 @@
     "area-description": {
       "text": "Achievement Level",
       "subtext": "relative to standard"
-    }
+    },
+    "error-band": "± {{errorBand}}"
```
jtreuting: see comment above about making this common possibly

esotrope: shouldn't we just call this var standardError? also is this common text? think it might be reused

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
```diff
@@ -78,7 +78,9 @@
 </div>
 
 <ng-template #dataPointPopoverTemplate let-point>
-  <span class="small mr-xs">{{point.levelRange.scaleScoreRange.minimum}}</span>
-  <span class="label label-lg {{point.levelRange.level.color}}">{{point.scaleScore}}</span>
-  <span class="small ml-xs">{{point.levelRange.scaleScoreRange.maximum}}</span>
+  <div class="text-center">
+    <span class="label label-lg {{point.levelRange.level.color}}">{{point.scaleScore}}</span>
```
esotrope: would be nice if we could center align this with the popover so its not pushed left by the standard error but not sure how we'd do that simply

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
dsnyder-fairway: looks like in other spots we have "± {{ something | translate}}". I'll move this to common and see if I can consolidate those other spots to using this as well

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: We have to be careful about the term. "Standard Error", "Standard Error of the Mean" and "Standard Deviation" are all very specific. SB has used "error band" to generically refer to a ± value included with a number.

dsnyder-fairway: I'll keep this as "error band" then. I will move this to common though

PR #1278, opened 2018-05-08
DO NOT MERGE: demonstration of teacher student group resolution in sql layer

esotrope:
What resolving group data in queries could look like if done in sql.

Another option would be to have the service layer check the group type and call a different repo/sql statement (which would add more redundancy to the sql file but possibly make things simpler?)

### DIFF
--- a/reporting-service/src/main/resources/application.sql.yml
+++ b/reporting-service/src/main/resources/application.sql.yml
@@ -52,14 +52,31 @@ sql:
             JOIN student_group_membership sgm ON e.student_id = sgm.student_id
             JOIN student_group sg ON sg.id = sgm.student_group_id
             JOIN user_student_group usg ON usg.student_group_id = sg.id
+            JOIN teacher_student_group_membership tsgm ON e.student_id = tsgm.student_id
+            JOIN teacher_student_group tsg ON tsg.id = tsgm.teacher_stuent_group_id
             JOIN school s ON e.school_id = s.id
             JOIN student st ON e.student_id = st.id
             JOIN school isch ON st.inferred_school_id = isch.id
           WHERE e.school_year = :school_year
             AND a.id = e.asmt_id
-            AND sg.id = :group_id
-            AND usg.user_login = :user_login
-            AND (sg.subject_id is null OR a.subject_id = sg.subject_id)
+            AND (
+              :group_id IS NULL
+              OR
+              (
+                sg.id = :group_id
+                AND usg.user_login = :user_login
+                AND (sg.subject_id is null OR a.subject_id = sg.subject_id)
+              )
+            )
+            AND (
+              :teacher_group_id IS NULL
+              OR
+              (
+                tsg.id = :teacher_group_id
+                AND tsg.user_login = :user_login
+                AND (tsg.subject_id IS NULL OR tsg.subject_id = :a.subject_id)
+              )
+            )
             AND ${sql.reporting.snippet.studentExamPermissionWithEmbargo}
 
       findLatestExam: >-
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,109 +25,109 @@ services:
       - 5672:5672
       - 15672:15672
 
-#  config-server:
-#    image: smarterbalanced/configuration-service:3.1.0.RELEASE
-#    container_name: config-server
+  config-server:
+    image: smarterbalanced/configuration-service:3.1.0.RELEASE
+    container_name: config-server
+    ports:
+      - 8888:8888
+    links:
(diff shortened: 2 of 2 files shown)


reporting-service/src/main/resources/application.sql.yml
```diff
@@ -52,14 +52,31 @@ sql:
...
+            )
+            AND (
+              :teacher_group_id IS NULL
+              OR
+              (
+                tsg.id = :teacher_group_id
+                AND tsg.user_login = :user_login
+                AND (tsg.subject_id IS NULL OR tsg.subject_id = :a.subject_id)
+              )
+            )
```
esotrope: i think the user_login checks here are redundant with the security snippet but it also might be a good fail-fast

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -52,14 +52,31 @@ sql:
             JOIN student_group_membership sgm ON e.student_id = sgm.student_id
             JOIN student_group sg ON sg.id = sgm.student_group_id
             JOIN user_student_group usg ON usg.student_group_id = sg.id
+            JOIN teacher_student_group_membership tsgm ON e.student_id = tsgm.student_id
+            JOIN teacher_student_group tsg ON tsg.id = tsgm.teacher_stuent_group_id
```
esotrope: table sketches for reference:
```
teacher_student_group
 id
 name
 school_id
 school_year
 subject_id
 user_login
 
teacher_student_group_subject
 teacher_student_group_id
 subject_id
```

jtreuting: In order for this to be able to find teacher or system groups, this would need to be an outer join since a teacher might not have their own teacher groups.

malaffoon: somehow it feels like the JOINs for group and teacher group will need to be left joins?

reporting-service/src/main/resources/application.sql.yml
esotrope: my quick check shows 10 queries. so we can either update 10 queries to support both group types or add 10 more redundant queries : J

jtreuting: For redundant queries we would need to pass in or call a different endpoint knowing we want a teacher created group, we couldn't just treat them as a group is a group is a group.

the other option is to include in the main student group tables with a flag and update those queries to return the flag for if it is a teacher group or not.

I'm under the assumption that this PR is just to demonstrate how it would be for separate tables but that decision on separate tables or one table is yet to be determination and this is meant to help make that decision.  Correct?

esotrope: Correct this is meant to inform our decision.

Note though, that with this implementation the query requires both `:group_id` and `:teacher_group_id` params meaning the java code needs to already know what type of group it is. This would probably end up manifesting as a separate endpoint signature like `/assessments?group=:id` vs `/assessments?teacherGroup=:id`

Also note that if we go with the same table we will have to update the security snippet to check only administrator given groups:
```
sg.group_type = 'administrator_created_group'
```

reporting-service/src/main/resources/application.sql.yml
agorina: It looks like both JOINS (teacher_student_group and student_group) are optional, so they both need to be LEFT joins. With that, the students/exams may not overlap, so it would require a separate JOIN to exam/student.

Almost feels like with these table layout it would need to be a UNION between two separate SQLs. But I am still trying to catch up.

jtreuting: i'm leaning towards the simplicity of having them in the same table with a flag that says one is teacher created.  i know that complicates the idea of being able to fully wipe the reporting schema cause we couldn't just leave some subset of tables alone, we'd have to delete only some from within a table in this case.  but seems more logical and clear in my mind at least.

malaffoon: @wtritch pointed out another problem with a single table: how will groups created in the warehouse (as part of group upload / SIS) avoid id collision with teacher groups created directly in the data mart?

jtreuting: oh yeah that is a good point. i've seen partitions of the ids like odd/even or positive/negative or teacher created groups start at 99999999, but that always feels hacky.

agorina: And if we do it in the data mart, it would have to be a `master` instance. I know we already write to the`master`, but this did not feel right then and does not feel right now.

I guess I am saying that I would rather write into the warehouse.

PR #1279, opened 2018-05-08
find most recent date for the session date to handle edge cases

jtreuting:
There may be timezone issues or data inconsistencies where the dates for the exams aren't all the same.  Right now it is inconsistent what date will show up.. Nohemi saw in QA a different date than I did in QA for the same session.  This changes it to get the most recent date for the session when it determines it.

You can see an April 4th date here but the session says Mar 1.

image


malaffoon:
Is this the cause of https://jira.fairwaytech.com/browse/DWR-1769?

jtreuting:
yep

PR #1284, opened 2018-05-10
adds student search for teacher student groups

esotrope:


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
```diff
@@ -0,0 +1,118 @@
...
+                findAllQuery,
+                new MapSqlParameterSource()
+                        .addValues(securityParameterProvider.getSecurityParameters(permissionSource))
+                        .addValue("group_id", search.getGroupId())
+                        .addValue("school_id", search.getSchoolId()),
+                this::buildStudents
+        );
+    }
+
+    private List<Student> buildStudents(final ResultSet row) throws SQLException {
```
esotrope: the below student construction would be much more straight forward if `student_ethnicity` were denormalized onto `student` into a pipe delimited list i'm thinking. this complicated construction is shared with group/school exam search

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/Student.java
```diff
@@ -0,0 +1,204 @@
+package org.opentestsystem.rdw.reporting.student;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+public class Student {
```
esotrope: this is basically `common.Student` + `StudentContext`. I realize composition or inheritance were options but  thought meh

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -8,6 +8,47 @@ sql:
...
+          OR EXISTS (
+            SELECT 1 FROM student_group_membership sgm
+              JOIN student_group sg ON sg.id = sgm.student_group_id
+            WHERE student_id = st.id
+              AND sg.id = :group_id
+          )
+        )
+        AND (:school_id IS NULL OR e.school_id = :school_id)
+        AND ${sql.reporting.snippet.studentExamPermissionWithEmbargo}
+      ORDER BY st.last_or_surname, st.first_name, st.ssid
```
esotrope: mad props to Alla for whipping this sql into shape. Here's what I had which was producing duplicate students because of the join on student groups:
```sql
SELECT
        st.id,
       ...
      FROM student st
        LEFT JOIN exam e ON (
          e.student_id = st.id
          AND NOT EXISTS (
              SELECT 1 FROM exam e2
              WHERE e2.student_id = st.id
                AND e2.completed_at > e.completed_at
          )
        )
        JOIN asmt a ON a.id = e.asmt_id
        JOIN school s ON s.id = e.school_id
        JOIN student_group_membership sgm ON st.id = sgm.student_id
        JOIN student_group sg ON sg.id = sgm.student_group_id
      WHERE (:school_id IS NULL OR e.school_id = :school_id)
        AND (:group_id IS NULL OR sg.id = :group_id)
      ORDER BY st.last_or_surname, st.first_name, st.ssid
```

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    public void findAllShouldApplyGroupFilterWhenSet() {
+        final List<Student> actual = repository.findAll(
+                userBuilder.permissionsById(individualStatewide).build(),
+                search().groupId(-10L).build()
+        );
+        assertThat(actual)
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyInAnyOrder(student1);
+    }
```
esotrope: security checks delegated to the base exam testing of the security snippet

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Student.java
```diff
@@ -84,7 +83,6 @@ public String getGenderCode() {
         private String lastName;
         private String genderCode;
         private Set<String> ethnicityCodes;
-        private String elasCode;
```
esotrope: unrelated cleanup of an unused prop

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -8,6 +8,47 @@ sql:
...
+      FROM exam e
+        LEFT OUTER JOIN exam e2
+          ON e.student_id = e2.student_id
+          AND (
+            e.completed_at < e2.completed_at
+            OR (
+              e.completed_at = e2.completed_at
+              AND e.id < e2.id
+            )
+          )
```
esotrope: this block makes sure there is no exam more recent than the joined one

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
malaffoon: Could you pivot the ethnicities in the query (instead of denormalizing the data)?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
```diff
@@ -0,0 +1,118 @@
...
+                row -> {
+                    final long studentId = row.getLong("student_id");
+                    final String ethnicityCode = row.getString("ethnicity_code");
+
+                    final Set<String> studentEthnicityCodes = ethnicityCodesByStudentId.get(studentId);
+                    if (studentEthnicityCodes == null) {
+                        ethnicityCodesByStudentId.put(studentId, newHashSet(ethnicityCode));
+                    } else {
+                        studentEthnicityCodes.add(ethnicityCode);
+                    }
```
malaffoon: I thought you liked the compute-if-absent approach? Something like:
```
    enthnicityCodesByStudentId.computeIfAbsent(studentId, id -> newHashSet()).add(ethnicityCode);
```

reporting-service/src/main/resources/application.sql.yml
malaffoon: Why are we restricting to students that have exams? Yes, i understand we need the school to apply the security filter but what if the teacher has a student in a group and there are no exams for that student? We just calling that a corner case and moving past it?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
esotrope: i do - this was copy pasted from the exam repo : j

reporting-service/src/main/resources/application.sql.yml
esotrope: i thought there are is no student unless there is 1+ exams associated with the student? don't we not know about students until they took an exam?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
esotrope: do you mean join ethnicities and process duplicative rows like this:
student1 ethnicity1
student1 ethnicity2
student2 ethnicity1
...
?

agorina: Yes, @malaffoon has a good point: 
```
SELECT
  *,
  GROUP_CONCAT(se.ethnicity_code ORDER BY se.ethnicity_id SEPARATOR '|') AS codes
FROM student s
  LEFT JOIN student_ethnicity se ON se.student_id = s.id
```

reporting-service/src/main/resources/application.sql.yml
agorina: That is an interesting point. Form the group management point of view a presents of exams should be irrelevant. One may want to create a group before exams are given.

malaffoon: Two scenarios that could really happen:
1. Normal group upload will create students before any exam happens. Of course it won't have the student names, just the ssid's.
2. An exam is deleted. The student originally created will still exist in the system. Note that this is very rare.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
esotrope: well this is a nice trick - i think ill change our exam query to do the same

reporting-service/src/main/resources/application.sql.yml
esotrope: @jtreuting what are your thoughts on this?

jtreuting: we have given mockups before when talking thru the admin group mgmt side of things that showed either using an SSID or the student name, so I think it is appropriate to show the SSID if we don't have a student.  I agree with @malaffoon that scenario 1 that he lists is something we should support

agorina: How would we know the student school id in this scenario?

malaffoon: Chatted with Alla offline and this probably won't work. If you look at the permission query it requires a school id (either from the school joined to the exam or the inferred school for a student which comes from exams). So teachers will only be able to see students that have exams. If they want something that works mo' better they need integration with a real roster system.

jtreuting: well that resolves that.

esotrope: cool i'm going to take that as this PR being good to merge then

PR #1286, opened 2018-05-10
adds interfaces for teacher created groups

esotrope:
Wanted to get some early feedback on the teacher created group API interfaces and models.
Please critique : j

### DIFF
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
@@ -4,12 +4,13 @@ import com.google.common.base.MoreObjects;
 
 public class Group {
 
-    private long id;
+    private Long id;
     private String name;
     private String schoolName;
     private int schoolId;
     private String subjectCode;
     private int studentCount;
+    private boolean teacherCreated;
 
     private Group() {
     }
@@ -18,7 +19,7 @@ public class Group {
         return new Builder();
     }
 
-    public long getId() {
+    public Long getId() {
         return id;
     }
 
@@ -42,6 +43,10 @@ public class Group {
         return studentCount;
     }
 
+    private boolean isTeacherCreated() {
+        return teacherCreated;
+    }
+
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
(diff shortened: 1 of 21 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
```diff
@@ -4,12 +4,13 @@
 
 public class Group {
 
-    private long id;
+    private Long id;
```
esotrope: was thinking to reuse this as the form model. not sure what others feel about that though since there are a lot of extra props to ignore when we insert

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/TeacherCreatedGroupAuthorizationService.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public interface TeacherCreatedGroupAuthorizationService {
+
+    /**
+     * Checks if the provided permission source is authorized to use the teacher created student group feature
+     *
+     * @param permissionSource the permission source to check for authority
+     * @return <code>true</code> if the provided permission source has authority to create teacher created student groups
+     */
+    boolean isAuthorized(@NotNull PermissionSource permissionSource);
```
esotrope: idea here is to make the impl enforce PII/PII_GROUP permission is only at school level.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
jtreuting: are you thinking of posting a group without students then adding them to it after that?

esotrope: hm good point. haven't thought of that part of the UI yet. this part of the API probably needs to be UI driven.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/TeacherGroup.java
```diff
@@ -0,0 +1,139 @@
...
+import com.google.common.collect.ImmutableSet;
+import org.opentestsystem.rdw.reporting.common.model.School;
+import org.opentestsystem.rdw.reporting.student.Student;
+
+import java.util.Set;
+
+/**
+ * Represents a teacher created group.
+ */
+public class TeacherGroup {
```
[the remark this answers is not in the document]
agorina: When you create it, you probably need a teacher login name too. It feels to me that it should be broken into two models (as you are suggesting).

PR #1288, opened 2018-05-11
Claim Aggregate report CSV export.

agorina:
This completes the backend work for the Claim Aggregate report with the exception of 25,000 Repository RSTs. Will be working on them next.

### DIFF
--- a/common/src/main/resources/i18n/en.json
+++ b/common/src/main/resources/i18n/en.json
@@ -18,7 +18,8 @@
           "standard-deviation": "ScaleScoreStandardDeviation",
           "student-count": "StudentsTested",
           "subject": "Subject",
-          "year": "SchoolYear"
+          "year": "SchoolYear",
+          "claim": "Claim"
         }
       }
     },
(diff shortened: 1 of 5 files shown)


common/src/main/resources/i18n/en.json
```diff
@@ -18,7 +18,8 @@
           "standard-deviation": "ScaleScoreStandardDeviation",
           "student-count": "StudentsTested",
           "subject": "Subject",
-          "year": "SchoolYear"
+          "year": "SchoolYear",
+          "claim": "Claim"
```
agorina: @jtreuting, Is this the right header for the `claim` column in CSV?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/converter/ClaimAggregateReportCsvHandler.java
```diff
@@ -0,0 +1,70 @@
...
+
+        final AggregateQuery query = ((AggregateReportRequest) resource.getReport().getReportRequest()).getQuery();
+        return query.getReportType() == Claim;
+    }
+
+    @Override
+    protected List<String> getMeasureHeaders(final ClaimAggregateReportQuery query) {
+        final List<String> headers = newArrayList();
+        headers.add(getMessage("report.aggregate.export.header.student-count"));
+        for (int i = 1; i <= 3; i++) {
```
malaffoon: nit: honestly, i would've just done 3 statements instead of the loop. Is the 3 ever going to be data/config-driven?

agorina: Yes, I think `3` will have to be driven by the subject configuration once we have configurable subjects.

PR #1290, opened 2018-05-11
custom aggregate query builder updates

dsnyder-fairway:
Mostly just here to get some early feedback. I am thinking there is likely a more concise approach to handling some of the angular code here.

new report type well
agg-reports

selecting report type affects the allowable assessment types
selecting claim

vice vera
screen shot 2018-05-11 at 9 03 54 am

Report Type: Claim. Subjects: All (or Math + ELA)
screen shot 2018-05-11 at 9 07 08 am

Math
screen shot 2018-05-11 at 9 19 15 am

ELA
screen shot 2018-05-11 at 9 19 22 am


webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -407,12 +410,13 @@ export class SBButtonGroup extends AbstractControlValueAccessor<any[]> implement
    */
   private computeState(options: Option[], values: any[]): State {
     if (this.allOptionEnabled) {
-      const effectivelySelectedAllOption = values.length === options.length;
+      const effectivelySelectedAllOption = values.length >= options.length;
```
dsnyder-fairway: strangely enough when repopulating the values I would end up with more values than options. Now that I think of it this might not be a sufficient check. I should rather see if the optionsFilter is the same length as options, right?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -53,6 +53,8 @@
     "field": {
       "report-type-label": "Report Type",
       "report-type-info": "TODO",
+      "claim-codes-label": "Claim Codes",
+      "claim-codes-info": "TODO",
```
dsnyder-fairway: I'll clean this up. We don't have any info for claim-codes-label

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -108,8 +110,10 @@
       "subgroup-filters-subtext": "Use this feature to include only selected student sub-groups, e.g. only male Hispanic students.",
       "comparative-subgroups-heading": "Subgroups",
       "comparative-subgroups-subtext": "The report will break down results by the subgroups you select.",
+      "report-type-heading": "Report Type",
       "review-heading": "Report Review",
       "preview-heading": "Report Preview",
+      "question-attributes-heading": "Question Attributes",
```
dsnyder-fairway: I'll remove this

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -742,11 +756,36 @@
...
+          "3": {
+            "name": "Reasoning"
+          },
+          "SOCK_2": {
+            "name": "Problem Solving"
+          }
+        }
       }
     },
     "subject-claim-code": {
```
dsnyder-fairway: @wtritch this is incorporating the suggested approach we discussed. TODO is to remove the subject-claim-code

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -39,6 +39,16 @@ export class AggregateReportService {
     return this.dataService.post(`${AggregateServiceRoute}/aggregate/estimatedRowCount`, query);
   }
 
+  /**
+   * Gets the estimated report row count for the provided report request
```
dsnyder-fairway: copy and pasted. I'll correct this

dsnyder-fairway:
Looks like all the claims codes are off by one (2 math when it should be 3. 3 ELA when it should be 4. 6 total when it should be 7). Looking into it

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
```diff
@@ -30,4 +33,11 @@ public int postEstimateReportRowCount(
             @RequestBody final AggregateQuery query) {
         return handlerProvider.findHandler(query).estimateReportRowCount(user, query);
     }
+
+    @PostMapping("/claimCodes")
+    public List<String> postClaimCodes(
+            @AuthenticationPrincipal final User user,
+            @RequestBody final AggregateQuery query) {
+        return handlerProvider.findHandler(query).claimCodes(user, query);
```
agorina: Could you please clarify why you need this?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/ClaimReportQueryHandler.java
```diff
@@ -64,6 +64,11 @@ protected ClaimAggregateReportQuery preProcess(final ClaimAggregateReportQuery q
                 query.copy().claimCodes(findAllClaimCodes(query.getAssessmentTypeCode(), query.getSubjectCodes())).build());
     }
 
+    @Override
+    public ClaimRepository getClaimRepository() {
+        return this.claimRepository;
```
agorina: I am not sure I see why you would need it here?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/AbstractAggregateQueryHandler.java
```diff
@@ -93,6 +95,25 @@ public int estimateReportRowCount(final User user, final Q query) {
                 .sum();
     }
 
+    @Override
+    public List<String> claimCodes(final User user, final Q query) {
```
agorina: You should not need it. `ClaimReportQueryHandler` does this already. Please take a look and let me know if you have any questions.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/AggregateQueryHandler.java
```diff
@@ -38,4 +39,9 @@
      * @return an estimated number of rows
      */
     int estimateReportRowCount(@NotNull User user, @NotNull Q query);
+
+    /**
+     * TODO
+     */
```
agorina: Please see my comments below. I do not think you need this.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
dsnyder-fairway: this is needed for populating the list of claim codes for a given subject before a query is made

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/AbstractAggregateQueryHandler.java
dsnyder-fairway: Good point. I am realizing now that I likely did not have a fully correct query when the estimatedRowCount was called originally. I'll see if I can use the ClaimReportQueryHandler from the call to get the claim codes

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
dsnyder-fairway: Discussed with @agorina. The approach for filtering this will be based on the initial query to get the report options. After that I will filter the claim codes based on the subject that is selected through the front end

jtreuting:
purely screenshot comment right now.  Aaron would prefer us to disable options instead of take them away when we can.  So I think the assessment type should be disabled instead of removed (e.g. when select Claim and IAB goes away) . Same for the Longitudinal option up top.  That should be disabled and have a popover on hover (like we do for the Result Views dropdown option for Results By Item) that says it is only available for Summative assessments.

one important caveat, there is a setting that can turn off Summative in the system completely, so if config setting says there are no summatives, then Longitudinal shouldn't be there at all since they aren't using the system for Summative.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportOptions.java
```diff
@@ -28,6 +29,7 @@
     private List<Integer> schoolYears;
     private List<String> subjects;
     private List<String> summativeAdministrationConditions;
+    private Set<String> claimCodes;
```
esotrope: Maybe this was caught already but i think this should be a list since there is a natural order right?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/ClaimService.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.olap.service;
+
+import org.opentestsystem.rdw.olap.model.Claim;
+
+import java.util.List;
+
+public interface ClaimService {
+
+    List<Claim> findAll();
+}
```
esotrope: nit: interfaces should have docs

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportOptions.java
dsnyder-fairway: I am moving this to a list. A bunch of changes to this PR coming shortly

agorina: Good point, it was not caught. 
I do not think we have a way to order on the back end right now, but I think with the configurable subjects there will be a db column to sort by. 
Maybe a TODO to address it then? @wtritch ?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -123,6 +123,9 @@ export class AggregateReportFormComponent {
    */
   subgroupItems: SubgroupItem[] = [];
 
+  private originalReportTypes: any;
+  private originalAssessmentTypes: any;
+  private originalClaimCodes: any;
```
esotrope: would prefer we just make a copy of the `options` object and call it `filteredOptions`

webapp/src/main/webapp/src/app/aggregate-report/wide-radio-group.component.ts
```diff
@@ -0,0 +1,116 @@
...
+const DefaultButtonStyles = 'btn-primary';
+
+@Component({
+  selector: 'wide-radio-group',
+  templateUrl: './wide-radio-group.component.html',
+  providers: [
+    Forms.valueAccessor(WideRadioGroupComponent)
+  ]
+})
+export class WideRadioGroupComponent extends AbstractControlValueAccessor<any[]> implements OnInit {
```
esotrope: hm cant say i'm thrilled about adding another variant on this component just for "wideness" styles. 
I think (1) i'd challenge the need for the styles and (2) try to solve this in CSS

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
esotrope: can you describe what these changes are trying to solve? i'm a little lost on the need for this

esotrope:
Can we get away with just showing the description for the selected report type instead of all of them at once?

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
dsnyder-fairway: I'll get back to you on this as it may be unneeded with my changes coming up

dsnyder-fairway:
@esotrope I am afraid not

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportOptions.java
wtritch: Yeah, scorable claims will have a specified display order

aaronsleeper:
suggest changing the copy to the following

**General Report** 
Compare aggregate scores across groups

**Claim Focus**
 Examine claim specific performance across groups

**Longitudinal **
Compare performance within a population of cohorts

jtreuting:
@aaronsleeper all showing at once?  @esotrope was suggesting only include text description when you select it.  i'm guessing the concern is about space.

also @dsnyder-fairway seems like that is not full width of 3 equal columns?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/web/CustomAggregateController.java
dsnyder-fairway: removed

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/ClaimReportQueryHandler.java
dsnyder-fairway: removed

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/AbstractAggregateQueryHandler.java
dsnyder-fairway: removed

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/AggregateQueryHandler.java
dsnyder-fairway: removed

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportOptions.java
dsnyder-fairway: now a list

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/ClaimService.java
dsnyder-fairway: added

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -481,6 +509,17 @@ export class AggregateReportFormComponent {
     });
   }
 
+  private filterClaimCodes(): void {
+    this.options.assessmentTypes = this.originalAssessmentTypes.filter(assessmentType => assessmentType.value !== 'iab');
+    this.reportService.getClaimCodes(this.createReportRequest().query)
```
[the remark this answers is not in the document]
dsnyder-fairway: done

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.spec.ts
```diff
@@ -52,7 +52,10 @@ describe('AggregateReportOptionsMapper', () => {
         performanceLevelCount: 0,
         performanceLevelDisplayTypes: [ 'displayTypeA' ],
         performanceLevelGroupingCutPoint: 0,
-        aggregateReportIdentityColumns: [ 'columnA' ]
+        aggregateReportIdentityColumns: [ 'columnA' ],
+        aggregateReportStateResultsEnabled: true,
+        aggregateReportLongitudinalCohortEnabled: true,
+        aggregateReportClaimEnabled: true
```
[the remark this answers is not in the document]
dsnyder-fairway: hmmm, looks like I misread this. I favored using typeCode instead. I can make this array though...

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultReportOptionService.java
```diff
@@ -74,7 +73,7 @@ public ReportOptions getReportOptions() {
         return ReportOptions.builder()
                 .assessmentGrades(codesOf(assessmentGradeService.getAssessmentGrades()))
                 .completenesses(codesOf(completenessService.getCompletenesses()))
-                .claimCodes(ImmutableSet.copyOf(codesOf(claimService.findAll())))
+                .claims(claimService.findAll())
```
esotrope: is this so the UI knows what subject each is associated with? wonder if it doesn't make sense to return `Map<String, List<String>> claimCodesBySubjectCode`

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
dsnyder-fairway: done

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -190,26 +194,43 @@ export class AggregateReportRequestMapper {
...
+
     const defaultLongitudinalCohort = {
       assessmentGrades: [],
       toSchoolYear: options.schoolYears[ 0 ]
     };
 
     let generalPopulation,
-      longitudinalCohort;
+      longitudinalCohort,
+      claimReport;
```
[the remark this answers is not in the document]
dsnyder-fairway: done

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -74,7 +74,7 @@ <h2>{{'aggregate-report-form.heading' | translate}}</h2>
             <h3>{{'aggregate-report-form.section.report-type-heading' | translate}}</h3>
           </div>
           <div class="well pb-0"
-               [hidden]="!currentAssessmentDefinition.aggregateReportLongitudinalCohortEnabled && !currentAssessmentDefinition.aggregateReportClaimEnabled">
+               [hidden]="currentAssessmentDefinition.typeCode === 'iab'">
```
esotrope: yeah we'll want to revert this - the idea is to remove all traces of `xyz == assessmentTypeCode` checks

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
```diff
@@ -54,4 +54,25 @@ export class AggregateReportOptionsService {
...
+          <Claim> {
+            assessmentType: claim.assessmentTypeCode,
+            subject: claim.subjectCode,
+            code: claim.code
+          }
+        );
+      }
+    );
+    return claimArray;
+  }
```
esotrope: most of the time and this time Array.forEach + Array.push = Array.map

webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment-definition.ts
```diff
@@ -3,7 +3,7 @@ export interface AssessmentDefinition {
   /**
    * Reflective reference to the type code of the assessment
    */
-  readonly typeCode: string;
+  readonly typeCode: 'sum' | 'iab' | 'ica';
```
esotrope: makes sense for now i guess. if we ever have configurable assessment types we wouldn't want this though.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
dsnyder-fairway: changed

dsnyder-fairway:
@jtreuting now taking up the full width
screen shot 2018-05-14 at 10 33 05 am

dsnyder-fairway:
@jtreuting 
screen shot 2018-05-14 at 12 30 10 pm
screen shot 2018-05-14 at 12 30 19 pm

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
dsnyder-fairway: updated this a bit

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -410,11 +410,11 @@ export class SBButtonGroup extends AbstractControlValueAccessor<any[]> implement
    */
   private computeState(options: Option[], values: any[]): State {
     if (this.allOptionEnabled) {
-      const effectivelySelectedAllOption = values.length >= options.length;
       const optionsFilter = options.filter(option => values.includes(option.value));
```
esotrope: nit: maybe call this `effectiveValues`?

dsnyder-fairway:
closing for now

PR #1292, opened 2018-05-11
Replaced configurable max-claims with the common constant.

agorina:


agorina:
Merging since we have talked about it. But if needed, will address comments.

PR #1294, opened 2018-05-12
fix preselected from IAB group dashboard not adding new assessments t…

jtreuting:
…o selected list

fixes: https://jira.fairwaytech.com/browse/DWR-1776

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -43,8 +43,6 @@ export class AssessmentsComponent implements OnInit {
     this._hasInitialAssessment = !Utils.isNullOrEmpty(this._assessmentExams);
   }
 
-  private _preselectedAssessments: Assessment[] = [];
-
   /**
    * The provider which implements the AssessmentProvider interface in order
    * to load assessment and exam data.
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -128,13 +126,10 @@ export class AssessmentsComponent implements OnInit {
   }
 
   get selectedAssessments(): Assessment[] {
-    if (!this.showOnlyMostRecent && this._preselectedAssessments.length > 0) {
-      return this._preselectedAssessments;
-    } else if (this.showOnlyMostRecent && this._assessmentExams.length > 0) {
-      this._preselectedAssessments = [];
+    if (this.showOnlyMostRecent && this._assessmentExams) {
```
jtreuting: reverting this block to how it was before

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -199,7 +194,6 @@ export class AssessmentsComponent implements OnInit {
         preselectedAssessments.forEach((assessment) => {
           assessment.selected = true;
           this.loadAssessmentExam(assessment);
-          this._preselectedAssessments = this._preselectedAssessments.concat(assessment);
```
jtreuting: no need to track preselected differently than other selected assessments.  just add them to the select assessments as usual on init and then it works as expected

PR #1297, opened 2018-05-14
single View IAB Dashboard on the landing page defaulting to first group

jtreuting:
sort groups by name at service for consistency on landing page and group dashboard

image

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/group.service.ts
+++ b/webapp/src/main/webapp/src/app/groups/group.service.ts
@@ -4,6 +4,8 @@ import { ReportingServiceRoute } from '../shared/service-route';
 import { map } from 'rxjs/operators';
 import { Group } from '../groups/group';
 import { Observable } from 'rxjs/Observable';
+import { byString } from '@kourge/ordering/comparator';
+import { ordering } from '@kourge/ordering';
 
 @Injectable()
 export class GroupService {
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/groups/group.service.ts
```diff
@@ -13,14 +15,16 @@ export class GroupService {
...
-        };
+      map(serverGroups => serverGroups
+        .sort(ordering(byString).on<any>(x => x.name).compare)
+        .map(serverGroup => {
+          return <Group>{
+            id: serverGroup.id,
+            name: serverGroup.name,
+            schoolName: serverGroup.schoolName,
+            schoolId: serverGroup.schoolId,
+            subjectCode: serverGroup.subjectCode || 'ALL'
```
esotrope: do we still use "ALL" in the code?

jtreuting: i have no idea, these were just indenting changes.  i can check though before merging

jtreuting:
```
    "subject": {
      "ALL": {
        "name": "Math, ELA",
        "short-name": "All"
      },
      "ELA": {
        "name": "ELA",
        "short-name": "ELA"
      },
      "Math": {
        "name": "Math",
        "short-name": "Math"
      }
    },
```
Looks like we do right now, at least on the homepage where we list the groups.  We won't as the configurable subjects gets implemented on the front-end .

PR #1299, opened 2018-05-14
Feature/teacher student group api impl

esotrope:
Hoping to give you guys an early look at these changes and get some feedback.


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,123 @@
...
+    public TeacherGroup saveGroup(final User user, final TeacherGroup group) {
+        if (group.getId() == null) {
+            return createGroup(user, group);
+        }
+        return updateGroup(user, group);
+    }
+
+    @Override
+    public void deleteGroup(final User user, final long groupId) {
+        checkUpdateAuthority(user, groupId);
```
esotrope: Wasn't sure how to split transactions across the tables so long as these security checks exist.
Basically having this is equivalent to having the below to repo methods wrapped in
```
if (groupRepository.findOne(user, groupId) != null) {}
```
Only thing i can see doing is adding user_login security checks into the groupMembershipRepository sql (currently it has none)

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private void checkCreateAuthority(final User user, final TeacherGroup group) {
+        final Set<Long> entitledStudentIds = ImmutableSet.copyOf(studentRepository.findAllIdsByUsername(user));
+        if (!entitledStudentIds.containsAll(group.getStudentIds())) {
+            throw new RuntimeException(String.format(
+                    "Error creating teacher group: User \"%s\" does not have access to one or more of the provided student IDs",
+                    user.getUsername()
+            ));
+        }
```
esotrope: makes sure the user has authority to access the students at group create time

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private void checkUpdateAuthority(final User user, final long groupId) {
```
esotrope: makes sure the user owns the group

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
```diff
@@ -53,4 +56,14 @@
         );
     }
 
+    @Override
+    public List<Long> findAllIdsByUsername(final PermissionSource permissionSource) {
```
esotrope: for making sure requested student IDs are accessible by the user

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -8,7 +8,77 @@ sql:
...
+      WHERE group_id = :group_id
+
   student:
+    findAllIds: >-
+      SELECT e.student_id,
+      FROM exam e
+        JOIN school s ON s.id = e.school_id
+        JOIN school isch ON st.inferred_school_id = isch.id
+        JOIN asmt a ON a.id = e.asmt_id
+      WHERE ${sql.reporting.snippet.studentExamPermissionWithEmbargo}
```
esotrope: don't think this will work actually, probably need to pass in some group ID(s) for it to account for all students accessible through groups

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
wtritch: ...I'd just have a single multi-table transaction for the entire action rather than have funky cleanup sql for partial-failure...

agorina: Let me respectfully disagree. 

1. We have a case when only a group membership is updated, right? In this case wouldn't it make sense to pass security into groupMembershipRepository.
2. Why are we doing a security checkin in Java? Why not SQL?  maybe missing some details here but we seem to be doing all security validation for reporting in SQL. Besides, it will save a SQL call to look up.
3. I do not see a need for the clean up.  From a user point of view, what is the harm if a group fails to delete why the membership is deleted? The user will still have to retry.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,123 @@
...
+            return createGroup(user, group);
+        }
+        return updateGroup(user, group);
+    }
+
+    @Override
+    public void deleteGroup(final User user, final long groupId) {
+        checkUpdateAuthority(user, groupId);
+        groupRepository.delete(groupId);
+        groupMembershipRepository.deleteAllByGroupId(groupId);
```
agorina: The order should be flipped here.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -8,7 +8,77 @@ sql:
...
+        sc.id as school_id,
+        s.code as subject_code,
+        GROUP_CONCAT(tsgm.student_id SEPARATOR '|') AS student_ids
+      FROM teacher_student_group tsg
+        JOIN school sc ON sc.id = tsg.school_id
+        LEFT JOIN subject s ON s.id = tsg.subject_id
+        LEFT JOIN teacher_student_group_membership tsgm ON tsgm.teacher_student_group_id = tsg.id
+      WHERE user_login = :user_login
+        AND tsg.id = :id
+      GROUP BY tsg.id
```
agorina: I am afraid we are abusing `GROUP_CONCAT` feature here. I see why are want to do it, but `GROUP BY` is a relatively expensive call. I think it would be better to load multiple records with duplicate data for a group and filter in java. Curious to know what others think.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -8,7 +8,77 @@ sql:
...
+
+    update: >-
+      UPDATE teacher_student_group tsg SET
+        name = :name,
+        subject_id = :subject_id
+      WHERE tsg.id = :id
+
+    delete: >-
+      DELETE FROM teacher_student_group tsg
+      WHERE tsg.id = :id
```
agorina: About the security comment I made above: you should be able to add `and tsg.user_login = : user_login`.

reporting-service/src/main/resources/application.sql.yml
esotrope: thinking i should have just done a second `SELECT FROM teacher_student_group_membership WHERE teacher_student_group_id = ...`

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private TeacherGroup createGroup(final User user, final TeacherGroup group) {
+        checkCreateAuthority(user, group);
+
+        final TeacherGroup created = groupRepository.create(user.getUsername(), group);
+        for (final long id : group.getStudentIds()) {
+            groupMembershipRepository.create(created.getId(), id);
```
agorina: You should be able to insert multiple records at once. In short, you would need to call `namedParameterJdbcTemplate.batchUpdate`. 

Check out JdbcStudentRepository#executeEthnicityChange in the ingest project for an example.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,123 @@
...
+    private TeacherGroup updateGroup(final User user, final TeacherGroup group) {
+        checkUpdateAuthority(user, group.getId());
+        groupRepository.update(group);
+
+        final Collection<Long> currentIds = group.getStudentIds();
+        final Collection<Long> previousIds = groupMembershipRepository.findAllStudentIdsByGroupId(group.getId());
+
+        for (final long id : currentIds) {
+            if (!previousIds.contains(id)) {
+                groupMembershipRepository.create(group.getId(), id);
```
agorina: Ditto about multiple records in one call.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    private void checkUpdateAuthority(final User user, final long groupId) {
+        final TeacherGroup existing = groupRepository.findOne(user, groupId);
+        if (existing == null) {
+            throw new RuntimeException(String.format(
+                    "Error updating teacher group: User \"%s\" does not have access group with ID: %s",
+                    user.getUsername(), groupId
+            ));
```
esotrope: do you guys like throwing inside transactions or logging then no-oping btw?

reporting-service/src/main/resources/application.yml
```diff
@@ -64,6 +64,13 @@ artifacts:
 jwt:
   secret: ${rdw.reporting.jwt.secret:secret}
 
+jdbc:
+  retry:
+    default:
+      max-attempts: 3
+      delay: 30
+      multiplier: 2
```
esotrope: took these from ingest and changed max-attempts from 4 to 3. let me know what defaults you guys think are sensible here

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
agorina: Oh, one more thing: for this to work, you need to add `&rewriteBatchedStatements=${spring.reporting_datasource.rewrite-batched-statements:true}` to the connection string in the application.yml file. Checkout out admin service yml file for an example.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,142 @@
...
+
+        return Optional.of(group.copy().students(students).build());
+    }
+
+    @Override
+    @Transactional
+    @Retryable(
+            maxAttemptsExpression = "${jdbc.retry.default.max-attempts}",
+            backoff = @Backoff(delayExpression = "${jdbc.retry.default.delay}", multiplierExpression = "${jdbc.retry.default.multiplier}", random = true)
+    )
```
agorina: Should the retry be on the repository?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
agorina: I do not have a strong preference, as long as there is no retry. This will be re-tried, and it shouldn't.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/TeacherGroup.java
```diff
@@ -0,0 +1,150 @@
...
+            group.name = name;
+            group.schoolId = schoolId;
+            group.school = school;
+            group.subjectCodes = safeSetCopy(subjectCodes);
+            group.studentIds = safeSetCopy(studentIds);
+            group.students = safeSetCopy(students);
+            return group;
+        }
+
+        private <T> ImmutableSet<T> safeSetCopy(final Set<T> source) {
```
jtreuting: seems like something that might be useful in common

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupAuthorizationService.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+@Service
+class DefaultTeacherGroupAuthorizationService implements TeacherGroupAuthorizationService {
+
+    @Override
+    public boolean isAuthorized(final PermissionSource permissionSource) {
+        final Permission individualPermission = permissionSource.getPermissionsById().get(ReportingPermission.IndividualPiiRead);
+        final Permission groupPermission = permissionSource.getPermissionsById().get(ReportingPermission.GroupPiiRead);
+        return (individualPermission != null && isSchoolLevel(individualPermission.getScope()))
+                || (groupPermission != null && isSchoolLevel(groupPermission.getScope()));
```
malaffoon: nit: you could put the (duplicate) null check into the helper (if you tweak the signature of the helper)

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupAuthorizationService.java
```diff
@@ -0,0 +1,38 @@
...
+import org.opentestsystem.rdw.reporting.common.security.ReportingPermission;
+import org.opentestsystem.rdw.security.Permission;
+import org.opentestsystem.rdw.security.PermissionScope;
+import org.springframework.stereotype.Service;
+
+/**
+ * Grants access to users with only school-level PII and Group PII access
+ */
+@Service
+class DefaultTeacherGroupAuthorizationService implements TeacherGroupAuthorizationService {
```
malaffoon: Completely unrelated and not anything you have to do but ... we should really address the whole role thing in this system. This service deals with "teacher" even though "teacher" isn't a thing anywhere in the system.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,142 @@
...
+                        .studentIds(group.getStudentIds())
+                        .build()
+                )
+        );
+
+        return Optional.of(group.copy().students(students).build());
+    }
+
+    @Override
+    @Transactional
```
malaffoon: nit: if we take this approach we should consider a custom annotation that is transactional and retryable with the given defaults.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
malaffoon: I haven't caught up with the discussion but i generally agree that transactional/retrying should be hidden in the repository layer when it can be. Specifically i think the repo should have a `save` method for this. And a `delete` method for the next one. There are no cross-repo or complex data model manipulations that would require this to leak into the service layer.  (Think about if we were using JPA, these methods would be hidden there and the service layer wouldn't be worrying about it.)

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
```diff
@@ -0,0 +1,142 @@
...
+        groupMembershipRepository.createAll(group.getId(), createIds);
+        groupMembershipRepository.deleteAll(group.getId(), deleteIds);
+
+        return group;
+    }
+
+    private void checkCreateAuthority(final User user, final TeacherGroup group) {
+        final Set<Long> entitledStudentIds = ImmutableSet.copyOf(studentRepository.findAllIdsByUsername(user));
+        if (!entitledStudentIds.containsAll(group.getStudentIds())) {
+            // log and exit transaction without exception?
```
malaffoon: This supports the approach of separating the concerns: the retry should only capture writing the data to the database, not auth exceptions.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcTeacherGroupMembershipRepository.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+import java.util.stream.StreamSupport;
+
+
+@Repository
+class JdbcTeacherGroupMembershipRepository implements TeacherGroupMembershipRepository {
```
malaffoon: Per comment above i don't think this repo should exist. Group membership should be hidden in the TeacherGroupRepository impl.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcTeacherGroupRepository.java
```diff
@@ -0,0 +1,150 @@
...
+    @Override
+    public void delete(final long groupId) {
+        template.update(
+                deleteQuery,
+                new MapSqlParameterSource()
+                        .addValue("id", groupId)
+        );
+    }
+
+    // TODO use subject repository to map back and forth
```
malaffoon: Or handle this in the queries so you don't have to worry about it here.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/TeacherGroup.java
```diff
@@ -0,0 +1,150 @@
+package org.opentestsystem.rdw.reporting.group;
+
+import com.google.common.collect.ImmutableSet;
+import org.opentestsystem.rdw.reporting.common.model.School;
+import org.opentestsystem.rdw.reporting.student.Student;
+
+import java.util.Set;
+
+/**
+ * Represents a teacher created group.
```
malaffoon: Hmm, i thought you added a `teacherCreated` flag to the existing `Group`. What's this one for? Perhaps add to the javadoc to clarify?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/TeacherGroupController.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @GetMapping("/{id}")
+    public TeacherGroup get(@AuthenticationPrincipal final User user, @PathVariable final long groupId) {
+        return service.getGroup(user, groupId)
+                .orElseThrow(NoSuchElementException::new);
+    }
+
+    @PostMapping
+    public TeacherGroup post(@AuthenticationPrincipal final User user, @RequestBody final TeacherGroup group) {
+        return service.saveGroup(user, group);
```
malaffoon: did we already discuss PUT /id vs. POST with body that has id set?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
```diff
@@ -53,4 +56,14 @@
...
+        return template.queryForList(
+                findAllIdsQuery,
+                new MapSqlParameterSource()
+                        .addValues(securityParameterProvider.getSecurityParameters(permissionSource)),
```
malaffoon: nit: pretty sure you can just pass a map in here so the second param can be just `securityParameterProvider.getSecurityParameters(permissionSource)`

reporting-service/src/main/resources/application.sql.yml
malaffoon: For this i agree with Alla: there aren't that many columns involved so just getting the full join should be fine. That said, i don't know that the GROUP_CONCAT would hurt since this is a single group.

@esotrope you could do that but then you're headed down the home-grown ORM path ... which is never good.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupAuthorizationService.java
esotrope: can do. i'll want to rename the method to `presentAndSchoolLevel` or `notNullAndIsSchoolLevel` or something weird though. I'll play with it

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupAuthorizationService.java
esotrope: Yeah, its an abstraction we understand because of the way the requirements were written. If i were to call this what it is in the system it would be `UserDefinedGroup` and we would have `Group.userDefined` or something to distinguish from `adminDefined`

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/TeacherGroup.java
esotrope: this is meant to embody the payload sent/received by the client when editing a single group. you are right though that this could be joined with the existing `Group` class. It would make the proper field saturation of the class even more situational and mysterious though. but i guess that is what we are already doing across the system...

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
malaffoon: pretty sure java supports annotation composition; i know i've done composite annotations in Spring framework.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/TeacherGroupController.java
esotrope: nope. i think i like having post/put, it would be more restful and it also isn't a big deal to have `f (id == null) post() else put()` logic bleed into the client i think

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultTeacherGroupService.java
esotrope: i think what it might be is that spring's annotation processor supports it and supports it when the composed annotations are spring annotations. could be wrong though

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupAuthorizationService.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.stereotype.Service;
+
+/**
+ * Grants access to users with only school-level PII and Group PII access
+ */
+@Service
+class DefaultUserGroupAuthorizationService implements UserGroupAuthorizationService {
+
+    @Override
+    public boolean isAuthorized(final PermissionSource permissionSource) {
```
esotrope: @malaffoon look better?

reporting-service/src/main/resources/application.sql.yml
esotrope: so we cool with this? or should i make another repo call for this info?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
```diff
@@ -1,80 +1,173 @@
 package org.opentestsystem.rdw.reporting.group;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableSet;
+import org.opentestsystem.rdw.reporting.common.model.School;
+import org.opentestsystem.rdw.reporting.student.Student;
+
+import java.util.Set;
 
 public class Group {
```
esotrope: went ahead and made `Group` more amorphous to support the user created group payload signatures

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/Group.java
```diff
@@ -1,80 +1,173 @@
...
 
+    /**
+     * @return the group's school entity ID
+     */
+    public int getSchoolId() {
+        return schoolId;
+    }
+
+    /**
+     * @deprecated use {@link Group#getSchool()}
```
esotrope: deprecated the following in favor of new fields:
`schoolName` -> `school.name`
`subjectCode` -> `subjectCodes`

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/UserGroupController.java
```diff
@@ -0,0 +1,51 @@
...
+    public Group get(@AuthenticationPrincipal final User user, @PathVariable final long groupId) {
+        return service.getGroup(user, groupId);
+    }
+
+    @PostMapping
+    public Group post(@AuthenticationPrincipal final User user, @RequestBody final Group group) {
+        return service.createGroup(user, group);
+    }
+
+    @PutMapping
```
esotrope: introduced PUT method in the controller

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupService.java
```diff
@@ -0,0 +1,108 @@
...
+        return ImmutableList.copyOf(repository.findAll(user));
+    }
+
+    @Override
+    public Group getGroup(final User user, final long groupId) {
+        final Group group = checkGroupAccess(user, groupId);
+
+        // Resolve group student IDs to student objects
+        // Filter out students the user no longer has access to
+        final Set<Student> students = ImmutableSet.copyOf(
```
esotrope: thinking of moving this to the repo layer

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupService.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /**
+     * Makes sure that the user has rights to add the students to a group
+     *
+     * @param user       the user creating or modifying a group
+     * @param studentIds the IDs of the students the group is being composed of
+     * @throws RuntimeException when the user does not have access rights one or more of the given students
+     */
+    private void checkStudentAccess(final User user, final Collection<Long> studentIds) {
```
esotrope: wondering what you guys think of this security measure. this makes sure the user cant add students to their group that they do not have access to. basically this will prevent them from making unused data in the table as group reads will prevent them from seeing students they do not have access to

esotrope:
If you guys are good with the general code organization here i'll move forward with writing tests for all of this - thanks!

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupService.java
wtritch: So it's basically script-kiddie protection?  The only thing that worries me is `studentRepository.findAllIdsByUsername(user)` which could potentially be massive for a multi-district or state user.

wtritch: I'd be tempted to invert it and send the list of students the user wants to the repository and get back the list of students that the user actually has access to...

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcUserGroupRepository.java
```diff
@@ -0,0 +1,284 @@
...
+    @Override
+    @Transactional
+    @Retryable(
+            maxAttemptsExpression = "${jdbc.retry.default.max-attempts}",
+            backoff = @Backoff(
+                    delayExpression = "${jdbc.retry.default.delay}",
+                    multiplierExpression = "${jdbc.retry.default.multiplier}",
+                    random = true
+            )
+    )
```
wtritch: hm... if you annotate a method with both `@Transactional` AND `@Retryable` is the transaction held through the retries?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupService.java
esotrope: school-level-only access is enforced at the service layer so in practice that wont happen but i'll see if i cant optimize the query anyway

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/ApplicationConfiguration.java
```diff
@@ -41,5 +49,10 @@ public OrganizationRepository repository(final NamedParameterJdbcTemplate templa
         return new ReportingJdbcOrganizationRepository(template, securityParameterProvider);
     }
 
+    @Bean
+    public PlatformTransactionManager transactionManager(final DataSource dataSource) {
+        return new DataSourceTransactionManager(dataSource);
```
agorina: Does it belong here? Or close to the DataSource configuration?

Also, I just realized that since it is in the reporting we cannot reused the existing datasource for the teacher student groups. We need a new one. The existing points to a replica, the new one will have to point to the master.

malaffoon: hmm, good point. Although i'd refer to them as `read-only`/`writeable` instead of `replica`/`master`

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/CollectionDifference.java
```diff
@@ -0,0 +1,22 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import java.util.Set;
+
+/**
+ * Represents the differences between two sets
+ *
+ * @param <T> the type of entries contained in the set
+ */
+public interface CollectionDifference<T> {
```
esotrope: introduced this to make the update student membership process more fluent. figured we could use it other places as well maybe

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/DataSourceConfiguration.java
```diff
@@ -29,9 +30,9 @@ public DataSource dataSource() {
      * @param dataSource the data mart data source
      * @return the jdbc template used to read from the data mart
      */
+    @Primary
```
esotrope: prep for adding writable datasource

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/WritableDataSourceConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.retry.annotation.EnableRetry;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+import javax.sql.DataSource;
+
+@Configuration
+@EnableRetry
+@EnableTransactionManagement
+public class WritableDataSourceConfiguration {
```
esotrope: introduced this datasource config for repos that write.
question: if the repo reads and writes should it be using two datasources for the respective ops?

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.ArgumentMatchers.refEq;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+
+@RunWith(SpringRunner.class)
+@ActiveProfiles("test")
+public class DefaultUserGroupServiceTest {
```
esotrope: added tests

esotrope:
@agorina @malaffoon @wtritch - hey guys, if you have time please give this another gander for good measure, hoping to merge and move on soon

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/WritableDataSourceConfiguration.java
agorina: I concur  Mark's last sentance.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupService.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Makes sure that the user has rights to the given group
+     *
+     * @param user    the user updating the group
+     * @param groupId the group ID
+     * @throws RuntimeException when the user does not have access rights for the given group
+     */
+    private Group checkGroupAccess(final User user, final long groupId) {
```
agorina: I find the name misleading. It really finds it. Maybe 'safeFindOne' or something like this?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcUserGroupRepository.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    @Override
+    @Transactional
+    @Retryable(
+            maxAttemptsExpression = "${jdbc.retry.default.max-attempts}",
+            backoff = @Backoff(
+                    delayExpression = "${jdbc.retry.default.delay}",
+                    multiplierExpression = "${jdbc.retry.default.multiplier}",
+                    random = true
```
agorina: I mentioned it to Landon, but would like to put it here in case somebody has comments or objections. I suggest to remove the re-try and test it in the MultiThreadedApplicationTest to see if it is even needed.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcUserGroupRepository.java
agorina: We need to include JdbcRetryConfiguration, and then it should work. Otherwise the order of the annotation execution is not correct.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/WritableDataSourceConfiguration.java
esotrope: the separate config is because the ds config for readonly is in the common module... not sure if that is a good place for it anymore though.
as for the datasource usage - this is what is currently being done - thanks for the confirmation : j

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupService.java
esotrope: idea here was to copy guava's `Preconditions.notNull` method. it wouldn't be a `safeFindOne` it would be an `unsafeFindOne` so not sure what to call it

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcUserGroupRepository.java
esotrope: 👍 i'll start trying to work that test in

reporting-service/src/main/resources/application.yml
```diff
@@ -21,6 +21,25 @@ spring:
       ?useSSL=${spring.datasource.use-ssl:false}\
       &useLegacyDatetimeCode=${spring.datasource.use-legacy-datetime-code:false}\
       &characterEncoding=${spring.datasource.character-encoding:utf8}\
+      &rewriteBatchedStatements=${spring.datasource.rewrite-batched-statements:true}\
```
agorina: This is the read only data source, right? If the answer is yes, then you do not need this change.

reporting-service/src/main/resources/application.yml
```diff
@@ -21,6 +21,25 @@ spring:
...
+      &useCompression=${spring.datasource.use-compression:true}\
```
agorina: Why do we need this one? 

I see that it exists in reporting (one or two projects) only, and I do not see it in the ingest. I wonder if we need it. I remember researching it in the ingest, but I do not recall the findings. I am assuming that since it is not in the ingest, the conclusion was that we do not need it - but I maybe wrong about it. Wonder if somebody remembers or have a strong opinion.

esotrope: i vaguely feel like it had something to do with aurora exporting csv but not sure at all

reporting-service/src/main/resources/application.yml
esotrope: 👍

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/JdbcRepositoryWithPermissionIT.java
```diff
@@ -35,14 +35,12 @@
     protected final static Map<String, Permission> individualStatewide = ImmutableMap.copyOf(permissions(individualOf(PermissionScope.STATEWIDE)));
     protected final static Map<String, Permission> groupStatewide = ImmutableMap.copyOf(permissions(groupOf(PermissionScope.STATEWIDE)));
 
+    @Deprecated
     protected User.Builder userBuilder;
```
agorina: Why not remove it? It is the tests, so as long as tests works...

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/WritableDataSourceConfiguration.java
malaffoon: Ah, thanks for the explanation of the config, makes sense.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/JdbcRepositoryWithPermissionIT.java
```diff
@@ -53,4 +51,12 @@ public void clearState() {
     protected ReportingSystemSettings getSystemSettings() {
         return systemSettings;
     }
+
+    protected User.Builder userBuilder() {
```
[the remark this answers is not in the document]
agorina: The best explanation I could come up with (besides I just did not think of it) is that it maybe a left over from the Multithreaded tests we used to have. 
As I said above, as long as the tests run - just change it.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/JdbcRepositoryWithPermissionIT.java
esotrope: didn't want to modify a bunch of unrelated tests in the PR. I think i'll do a cleanup round after this PR for things like this

agorina: Sounds good to me. Thanks.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcUserGroupRepository.java
```diff
@@ -0,0 +1,251 @@
...
+            )
+    )
+    public Group create(final String username, final Group group) {
+        final KeyHolder keyHolder = new GeneratedKeyHolder();
+        template.update(
```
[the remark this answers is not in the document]
esotrope: What is this index for btw? I just copied it from one that existed for `student_group` to be consistent.

If we verify before the insert attempt would we do a findByUsernameNameAndSchoolYear query then insert when that yields nothing?

esotrope:
Removed `@Retryable` until cuncurrency tests show they are needed.
Going to merge this for now and create a JIRA to tackle the concurrency testing effort.

PR #1305, opened 2018-05-15
implement 12 options for line color wheel on longitudinal graph

jtreuting:
image

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -1084,16 +1084,52 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
 
   .pallet-a {
     .color-0 {
-      .color-stroke { stroke: @gray-darkest; }
-      .color-fill { fill: @gray-darkest; }
+      .color-stroke { stroke: #6A3F52; }
+      .color-fill { fill: #6A3F52; }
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -1084,16 +1084,52 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
...
-      .color-stroke { stroke: @aqua; }
-      .color-fill { fill: @aqua; }
+      .color-stroke { stroke: #6E4E68; }
+      .color-fill { fill: #6E4E68; }
     }
     .color-2 {
-      .color-stroke { stroke: @green; }
-      .color-fill { fill: @green; }
+      .color-stroke { stroke: #6C5F7E; }
+      .color-fill { fill: #6C5F7E; }
```
jtreuting: didn't think having variables for the colors of the lines was needed since they were only used here and self contained.  i don't see a benefit to reusing elsewhere, but let me know if you disagree

esotrope: would save us from updating the color in two places though no?

esotrope: original idea was to use less to iterate over a colors array and generate these classes but never got to it

PR #1308, opened 2018-05-16
Bugfix/dwr 1815

jtreuting:
add scrollbars to the longitudinal legend when it gets bigger than the height of the graph


webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.html
```diff
@@ -2,7 +2,8 @@
   <div class="flex-child">
 
     <div *ngIf="chartView"
-         class="chart-series-legend pallet-a">
+         class="chart-series-legend pallet-a"
+         [style.height.px]="display.outerHeight">
```
jtreuting: doing this here instead of in CSS size the height of the chart is using this same setting

PR #1309, opened 2018-05-16
Bugfix/dwr 1785

jtreuting:
filter out null or undefined scale scores for the longitudinal paths

fixes: https://jira.fairwaytech.com/browse/DWR-1785


webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.component.ts
```diff
@@ -266,10 +266,14 @@ export class LongitudinalCohortChartComponent implements OnInit {
...
-            x: j,
-            y: scaleScore
-          })),
+          pathData: d3line(
+            performance.yearGradeScaleScores
+              .map(({ scaleScore }, j) => <any>{
+                  x: j,
+                  y: scaleScore
+                })
+              .filter(({ y }) => y != null)
```
jtreuting: doing the filter after the mapping so the index (j) is preserved

PR #1311, opened 2018-05-16
Updated size of the chart and max-size of the legend

jtreuting:
Addressed: https://jira.fairwaytech.com/browse/DWR-1782

Unrelated to that, changed to 10 color wheel lines with new colors

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -932,21 +932,21 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
 @opacity-soft: 0.45;
 @opacity-softened: 0.66;
 
-@colorwheel-0: #6A3F52;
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -961,8 +961,6 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
       &.color-7 { background-color: @colorwheel-7; }
       &.color-8 { background-color: @colorwheel-8; }
       &.color-9 { background-color: @colorwheel-9; }
-      &.color-10 { background-color: @colorwheel-10; }
-      &.color-11 { background-color: @colorwheel-11; }
```
esotrope: @aaronsleeper can you bless us with the syntax to lay these classes out in a less for each loop over a @colors array?

PR #1312, opened 2018-05-17
Display Aggregate Report: Longitudinal as the Report Type on the My R…

jtreuting:
…eports page for Longitudinal reports

image

The differentiator for Longitudinal vs regular custom aggregate appears to be the `request.reportTypeDisplayValue`

image

### DIFF
--- a/webapp/src/main/webapp/src/app/report/reports.component.ts
+++ b/webapp/src/main/webapp/src/app/report/reports.component.ts
@@ -54,11 +54,9 @@ export class ReportsComponent implements OnInit, OnDestroy {
   }
 
   getReportTypeTranslateValue(report: Report) {
-    if (report.request && report.request.reportTypeDisplayValue) {
-      return report.request.reportTypeDisplayValue.replace(' ', '-');
-    }
-
-    return report.reportType;
+    return report.request && report.request.reportTypeDisplayValue
+      ? report.request.reportTypeDisplayValue
+      : report.reportType;
   }
 
   private startPollingStatus(): void {
--- a/webapp/src/main/webapp/src/app/report/reports.component.ts
+++ b/webapp/src/main/webapp/src/app/report/reports.component.ts
@@ -53,6 +53,14 @@ export class ReportsComponent implements OnInit, OnDestroy {
     window.location.reload();
   }
 
+  getReportTypeTranslateValue(report: Report) {
+    if (report.request && report.request.reportTypeDisplayValue) {
+      return report.request.reportTypeDisplayValue.replace(' ', '-');
+    }
+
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -53,6 +53,14 @@ export class ReportsComponent implements OnInit, OnDestroy {
     window.location.reload();
   }
 
+  getReportTypeTranslateValue(report: Report) {
+    if (report.request && report.request.reportTypeDisplayValue) {
+      return report.request.reportTypeDisplayValue.replace(' ', '-');
```
jtreuting: in order to get a `en.json` translate property to work here i'm taking the value of "AggregateReportRequest Longitudinal" and putting a hyphen in there.  If you have any other ideas fire away.  Another option is to split on the spaces and have them be separate translate properties like a report type and a sub type.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1192,6 +1192,9 @@
     "reload": "Reload",
     "report-type": {
       "AggregateReportRequest": "Aggregate Report",
+      "AggregateReportRequest-CustomAggregate": "Aggregate Report",
+      "AggregateReportRequest-Claims": "Aggregate Report: Claims",
```
jtreuting: planning ahead.  @dsnyder-fairway FYI

esotrope: why do we have "Request" at the end of "AggregateReport" here? seems like it should just be "Aggregate" or "AggregateReport"

webapp/src/main/webapp/src/app/report/reports.component.ts
jtreuting: to tell you the truth not sure i tried, i thought it wouldn't like it.  but i'll double check as if it works it makes the code cleaner

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: not sure, but not a priority either :)

webapp/src/main/webapp/src/app/report/reports.component.ts
jtreuting: @esotrope you are right it works with the space in the translate pipeline.  i assumed it wouldn't.  changed it so it's cleaner code

esotrope: nice. had me worried there about the lib we use : J

PR #1314, opened 2018-05-17
summative report repo integration test

jtreuting:
In previous PR, I updated the shared SQL that existed to take in a type_id instead of hard-coding 1 like it did before for ICA.  So the new logic is setting that type_id appropriately.  

Am I being lazy to just add the one test that checks to see if it gets back the summative appropriately?  Opinion: should I be adding an equivalent Summative version for each of the ICA tests that is currently there for each of the 10 ICA tests?


agorina:
My understanding is that you changed from `AND a.type_id=1` to `AND a.type_id=:type_id` in two queries. So as long as you have at least one test for each query that verifies that it is using the passed in parameter and not a hardcoded value, it should be good enough.

PR #1316, opened 2018-05-18
Little clean up

agorina:
Little clean up, mostly of my own mess: I have created `CollectionUtils ` class to safely convert collections into Immutable ones. 

I did a search and replace in the project, but in some places we are using `Iterable`. I left them alone.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateOrganizationQuery.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateOrganizationQuery.java
@@ -8,7 +8,7 @@ import javax.validation.constraints.NotNull;
 import java.util.Objects;
 import java.util.Set;
 
-import static org.opentestsystem.rdw.reporting.common.util.CollectionUtils.asNullSafeImmutable;
+import static org.opentestsystem.rdw.reporting.common.util.CollectionUtils.safeImmutableOf;
 
 /**
  * Represents query parameters for the {@link CustomAggregateReportRepository}.
@@ -76,9 +76,9 @@ public class AggregateOrganizationQuery extends OrganizationTypeQuery {
 
         public AggregateOrganizationQuery build() {
             final AggregateOrganizationQuery query = super.build();
-            query.subjectCodes = asNullSafeImmutable(subjectCodes);
-            query.schoolYears = asNullSafeImmutable(schoolYears);
-            query.assessmentGradeCodes = asNullSafeImmutable(assessmentGradeCodes);
+            query.subjectCodes = safeImmutableOf(subjectCodes);
+            query.schoolYears = safeImmutableOf(schoolYears);
+            query.assessmentGradeCodes = safeImmutableOf(assessmentGradeCodes);
             return query;
         }
 
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateOrganizationQuery.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateOrganizationQuery.java
@@ -8,6 +8,8 @@ import javax.validation.constraints.NotNull;
 import java.util.Objects;
 import java.util.Set;
 
+import static org.opentestsystem.rdw.reporting.common.util.CollectionUtils.asNullSafeImmutable;
+
 /**
  * Represents query parameters for the {@link CustomAggregateReportRepository}.
  * <p>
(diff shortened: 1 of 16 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateOrganizationQuery.java
```diff
@@ -74,9 +76,9 @@ AggregateOrganizationQuery createInstance() {
 
         public AggregateOrganizationQuery build() {
             final AggregateOrganizationQuery query = super.build();
-            query.subjectCodes = asImmutable(subjectCodes);
-            query.schoolYears = asImmutable(schoolYears);
-            query.assessmentGradeCodes = asImmutable(assessmentGradeCodes);
+            query.subjectCodes = asNullSafeImmutable(subjectCodes);
```
agorina: I have create a utility class and replaced multiple implementations with it.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/OrganizationTypeQuery.java
```diff
@@ -273,9 +274,5 @@ public B economicDisadvantageCodes(final Set<String> economicDisadvantageCodes)
             this.economicDisadvantageCodes = economicDisadvantageCodes;
             return (B) this;
         }
-
-        <T> ImmutableSet<T> asImmutable(final Set<T> values) {
-            return values == null ? ImmutableSet.of() : ImmutableSet.copyOf(values);
```
agorina: It was returned `null`, but then in the build object getter was converting it to an empty set. I did not see the need for the middle step, all tests are passing.

common/src/main/java/org/opentestsystem/rdw/reporting/common/util/CollectionUtils.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public final class CollectionUtils {
+
+    public static <T> ImmutableSet<T> asNullSafeImmutable(final Set<T> value) {
```
malaffoon: javadocs please; yeah, what the method does is obvious here but ctrl-J should provide details

malaffoon: nit, and feel free to ignore or outvote me, but i think the name should be `safeAsImmutable` or `safeImmutableOf`. Why? Well, the unsafe version (if we needed such a thing) would be called `asImmutable` or `immutableOf` and when making it null-safe the convention is to prefix with `safe` ... so `safeImmutableOf`, QED.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/OrganizationTypeQuery.java
malaffoon: 👍 i generally prefer constructing the collection during build and just returning the value instead of copying it with every get call.

common/src/main/java/org/opentestsystem/rdw/reporting/common/util/CollectionUtils.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public final class CollectionUtils {
```
malaffoon: We should consider putting this in RDW_Common. We never really attacked this issue when we should have: making utility behavior in RDW_Ingest and RDW_Reporting consistent. There isn't a technical reason why it would be necessary but flipping between the projects can be frustrating with the inconsistencies in which collection library to use, etc.

malaffoon: To be clear, i'm not suggesting making that change now as part of this review, just commenting.

agorina: Agree with both comments. I think we have a technical debt JIRA to address it, some day...

common/src/main/java/org/opentestsystem/rdw/reporting/common/util/CollectionUtils.java
agorina: I like `safeImmutableOf`.

agorina: Ok, ok - adding.

PR #1318, opened 2018-05-18
Bugfix/dwr 1822

jtreuting:
Set to use the code as the message if not found instead of throwing

fixes: https://jira.fairwaytech.com/browse/DWR-1822

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
@@ -43,6 +43,8 @@ public class TranslationConfiguration {
 
     @Bean
     public MessageSource messageSource(final TranslationProvider translationProvider) {
-        return new TranslationProviderMessageSource(translationProvider);
+        final TranslationProviderMessageSource messageSource = new TranslationProviderMessageSource(translationProvider);
+        messageSource.setUseCodeAsDefaultMessage(true);
+        return messageSource;
     }
 }
--- a/common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSourceIT.java
+++ b/common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSourceIT.java
@@ -1,0 +1,28 @@
+package org.opentestsystem.rdw.reporting.common.i18n;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.opentestsystem.rdw.reporting.common.test.ITDataSourceConfiguration;
+import org.opentestsystem.rdw.reporting.common.test.RepositoryIT;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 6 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
```diff
@@ -43,6 +43,8 @@ public TranslationProvider translationProvider(final ResourceLoader resourceLoad
 
     @Bean
     public MessageSource messageSource(final TranslationProvider translationProvider) {
-        return new TranslationProviderMessageSource(translationProvider);
+        final TranslationProviderMessageSource messageSource = new TranslationProviderMessageSource(translationProvider);
+        messageSource.setUseCodeAsDefaultMessage(true);
```
jtreuting: don't throw on missing key, thanks @wtritch for that fix

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSourceIT.java
```diff
@@ -0,0 +1,28 @@
...
+@ContextConfiguration(classes = {ITDataSourceConfiguration.class, JdbcTranslationProvider.class, TranslationConfiguration.class})
+@ActiveProfiles("test")
+public class TranslationProviderMessageSourceIT {
+
+    @Autowired
+    private TranslationProviderMessageSource source;
+
+    @Test
+    public void itShouldUseCodeForAMissingTranslation() {
+        assertThat(source.getMessage("missing_random.key", new Object[]{}, DefaultLocale)).isEqualTo("missing_random.key");
```
jtreuting: added an IT that uses the TranslationConfiguration to make sure it is set to not throw when a key is missing and use that key as the message.  I needed to include the RepositoryIT and the Configuration for those because the TranslationProvider hits the DB to look for the key.

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSourceTest.java
```diff
@@ -53,10 +52,4 @@ public void itShouldProvideFallbackTranslations() {
         assertThat(source.getMessage("key_A", new Object[]{}, Locale.JAPAN))
                 .isEqualTo("value_A");
     }
-
-    @Test(expected = NoSuchMessageException.class)
-    public void itShouldThrowForAMissingTranslation() {
```
jtreuting: removed since this is using a default implementation of TranslationProviderMessageSource and adding that logic here in a unit test didn't make sense to me.

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationProviderMessageSourceIT.java
agorina: +1

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
esotrope: Is'nt it a good idea to throw on missing key? it means there was a developer error to not insert a default english translation to fallback to.

jtreuting: it feels drastic to not be able to generate a report because of that instead of showing the placeholder text as the message.

esotrope: yeah, i was thinking it might help us fail-fast and find issues faster? but we don't have tests setup around that so this works for me.

PR #1319, opened 2018-05-18
Claim UI

dsnyder-fairway:
mostly up for any helpful feedback


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ClaimOrganizationQuery.java
```diff
@@ -19,21 +19,26 @@
 public class ClaimOrganizationQuery extends OrganizationTypeQuery {
     private Set<Integer> schoolYears;
     private Set<String> assessmentGradeCodes;
-    private ImmutableMap<String, Set<String>> claimCodesBySubject;
+    private Map<String, Set<String>> claimCodesBySubject;
+    private Set<String> subjectCodes;
```
dsnyder-fairway: @agorina this was to resolve an issue I had with sending the claim codes originally. Now after groking this a bit more I think maybe I was misusing this endpoint. Say for instance I have a single Math claim code selected and no English claim codes, but both english and math selected. Should I send a `claimCodesBySubject` with values:
`Math->{'some math code'}, ELA->{}`? I was sending this as just `Math->{'some math code'}`...

dsnyder-fairway: Also note here that I had trouble deserializing these to Immutable objects

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -164,6 +168,21 @@ export class AggregateReportFormComponent {
...
+      this.filteredOptions.claimCodes = this.filteredOptions.claimCodes
+        .filter(claim => this.settings.subjects.includes(claim.value.subject))
+        .filter(claim => claim.value.assessmentType === this.settings.assessmentType);
+      if (this.settings.claimReport.claimCodesBySubject.length === 0) {
+        // selected All
+        this.filteredOptions.claimCodes.forEach(claim => {
+          this.settings.claimReport.claimCodesBySubject.push(claim.value);
+        });
+      }
+    }
```
dsnyder-fairway: changes above are done to correctly display Claims that were selected after hitting `View Query` from the results page

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -189,8 +208,8 @@ export class AggregateReportFormComponent {
         this.settings.name,
         fileName({ messageId: 'aggregate-report-form.field.report-name-file-name-error' })
       ],
-      assessmentGrades: [ this.settings.generalPopulation.assessmentGrades ],
-      schoolYears: [ this.settings.generalPopulation.schoolYears ],
+      assessmentGrades: [ this.settings.generalPopulation.assessmentGrades || this.settings.claimReport.assessmentGrades ],
+      schoolYears: [ this.settings.generalPopulation.schoolYears || this.settings.claimReport.schoolYears ],
```
dsnyder-fairway: I will remove this. This was done as a guesstimate to resolve an issue after hitting `View Query`. Now I notice this actually would not resolve it:
`ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'model: '. Current value: 'model: 08'.`

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -375,18 +404,50 @@ export class AggregateReportFormComponent {
...
+
+    let order = this.currentAssessmentDefinition.aggregateReportIdentityColumns.concat();
+    if (this.settings.reportType === 'Claim') {
+      order = order.concat([ 'claim' ]);
+      // this.settings.columnOrder = this.columnItems.map((orderItem) => orderItem.value);
+    }
 
     // Preserve column order between changing assessment types
-    const currentOrder = this.settings.columnOrder.concat();
+    // TODO fix column ordering. Shows as claim first and doesn't respect previous ordering
```
dsnyder-fairway: this TODO is likely outdated. I haven't been having any more issues with column ordering

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -372,7 +386,36 @@ export class AggregateReportRequestMapper {
   }
 
   private fromServerReportType(type: string) {
-    return this.ClientReportTypeBSyerverType[ type ];
+    return this.ClientReportTypeByServerType[ type ];
   }
 
+  private claimMap(claim: any[]): [ [ string, string[] ] ] {
```
dsnyder-fairway: will remove

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -372,7 +386,36 @@ export class AggregateReportRequestMapper {
...
+          arr.find(a => a[ 0 ] === c.subject)[ 1 ].push(c.code);
+        } else {
+          arr.push([ c.subject, [ c.code ] ]);
+        }
+      }
+    );
+    return arr;
+  }
+
+  strMapToObj(strMap: Claim[]) {
```
dsnyder-fairway: will rename and improve

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -135,28 +189,7 @@ export class AggregateReportSummary {
         label: translate('aggregate-report-form.field.subjects-label'),
         values: orAll(options.subjects, settings.subjects, code => translate(`common.subject.${code}.short-name`))
       },
-
-      ...(settings.reportType === 'GeneralPopulation' || !assessmentDefinition.aggregateReportLongitudinalCohortEnabled
```
dsnyder-fairway: where we had stuff like `!assessmentDefinition.aggregateReportLongitudinalCohortEnabled` I removed it because it wasn't clear what doing this added to the functionality, and it only added to my confusion

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -61,36 +68,84 @@ export class AggregateReportTableDataService {
...
+                        Percent: performanceLevelPercents
+                      },
+                      Grouped: {
+                        Number: groupedPerformanceLevelCounts,
+                        Percent: groupedPerformanceLevelCounts
+                      }
+                    },
+                    subgroup: subgroup
+                  };
+                  if (this.getClaimCodeTranslationKey(row) !== this.getClaimCodeTranslation(row)) {
```
dsnyder-fairway: yes, I am checking to see if we have successfully translated something in order to push it through. This was done because we would run into situations with untranslated keys (we would end up inserting a key for something like `common.subject.ELA.claim.1.name` where it only exists for `common.subject.Math.claim.1.name`

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultClaimService.java
```diff
@@ -0,0 +1,25 @@
...
+    private final ClaimRepository repository;
+
+    @Autowired
+    DefaultClaimService(final ClaimRepository repository) {
+        this.repository = repository;
+    }
+
+    @Override
+    public List<Claim> findAll() {
+        return ImmutableList.copyOf(repository.findAll());
```
malaffoon: nit: not sure why we need an extra copy?

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/service/impl/DefaultReportOptionServiceTest.java
```diff
@@ -142,4 +152,10 @@ public void isShouldMapServiceResponsesToFieldsCorrectly() {
                 .collect(toList());
     }
 
+    private List<Claim> claims(final int... ids) {
```
malaffoon: nit: might just be the way github is presenting it but i only see this called once with a single value so i'd just inline it.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ClaimOrganizationQuery.java
agorina: Yes, you should send `Math->{'some math code'}, ELA->{}` and you do not need `subjectCodes`.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/AbstractAggregateQueryHandler.java
```diff
@@ -107,7 +107,8 @@ protected Q preProcess(final Q queryIn) {
         checkArgument(!query.getAssessmentGradeCodes().isEmpty(), "report query is missing asmt grades");
         checkArgument(!query.getSchoolYears().isEmpty(), "report query is missing school years");
         checkArgument(!(!query.isIncludeState() && !query.isIncludeAllDistricts() && query.getDistrictIds().isEmpty() && query.getSchoolIds().isEmpty()), "report query is missing organization");
-        if (query.isIncludeState()) checkArgument(stateAggregateAssessmentTypes.contains(query.getAssessmentTypeCode()), "report query cannot include state results for " + query.getAssessmentTypeCode() + " assessment types");
+        if (query.isIncludeState())
+            checkArgument(stateAggregateAssessmentTypes.contains(query.getAssessmentTypeCode()), "report query cannot include state results for " + query.getAssessmentTypeCode() + " assessment types");
```
agorina: Did something change here?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/ClaimQueryConverter.java
```diff
@@ -24,6 +24,7 @@
     protected Builder getOrganizationTypeQueryBuilder(final ClaimAggregateReportQuery reportQuery, final StudentFilters filters) {
         return ClaimOrganizationQuery.builder()
                 .assessmentGradeCodes(reportQuery.getAssessmentGradeCodes())
+                .subjectCodes(reportQuery.getSubjectCodes())
```
agorina: You should not need it. 

Let me know if you still have questions, but I think after you change the query based on your comment above, it should all work.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ClaimAggregateReportQuery.java
```diff
@@ -14,7 +17,8 @@
 @SuppressWarnings("unchecked")
 public class ClaimAggregateReportQuery extends AbstractAggregateReportQuery {
     private ImmutableSet<Integer> schoolYears;
-    private ImmutableMap<String, Set<String>> claimCodesBySubject;
+    private ImmutableSet<String> subjectCodes;
```
agorina: And I think you know by now, but you do not need this.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -33,18 +34,24 @@ export class AggregateReportTableDataService {
     const organizations = this.createSampleOrganizations(settings, assessmentDefinition);
 
     const gradesAndYears: { grade: string, year: number }[] = [];
-    if (settings.reportType === 'GeneralPopulation' || !assessmentDefinition.aggregateReportLongitudinalCohortEnabled) {
```
agorina: I am not sure if I am reading it right but, longitudinal report has both values: general population and cohort. It looks to me like the code before was allowing to turn on one or another. You change seems to limited it to cohort data only.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -126,6 +140,46 @@ export class AggregateReportSummary {
...
+      assessmentAttributes = [
+        {
+          label: translate('aggregate-report-form.field.assessment-grades-label'),
+          values: inline(orAll(this.options.assessmentGrades, this.settings.claimReport.assessmentGrades,
+            code => translate(`common.assessment-grade.${code}`)))
+        },
+        {
+          label: translate('aggregate-report-form.field.school-year-label'),
+          values: this.settings.claimReport.schoolYears.map(value => this.schoolYearPipe.transform(value))
+        },
```
jtreuting: nit: left over comma

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/AbstractAggregateQueryHandler.java
dsnyder-fairway: this was done by Reformatting code. I can remove

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultClaimService.java
dsnyder-fairway: You mean the `ImmutableList.copyOf`? Just following the patterns from EthnicityService and GenderService used from org.opentestsystem.rdw.olap.service.impl.DefaultReportOptionService

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AggregateRow.java
```diff
@@ -35,7 +35,6 @@ public static Builder builder() {
         private ActiveAssessment assessment;
         private Measures measures;
         private Dimension dimension;
-        private String claimCode;
```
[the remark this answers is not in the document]
dsnyder-fairway: 🤷‍♂️  not sure

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -164,6 +168,21 @@ export class AggregateReportFormComponent {
...
 
     this.options = optionMapper.map(this.aggregateReportOptions);
+    if (!this.settings.assessmentType.includes('sum')) {
+      this.options.reportTypes = this.options.reportTypes.filter(reportType => reportType.value !== 'LongitudinalCohort');
+    }
+    this.filteredOptions = Object.assign({}, this.options);
+    if (this.settings.reportType === 'Claim') {
+      this.filteredOptions.claimCodes = this.filteredOptions.claimCodes
+        .filter(claim => this.settings.subjects.includes(claim.value.subject))
+        .filter(claim => claim.value.assessmentType === this.settings.assessmentType);
```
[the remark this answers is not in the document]
dsnyder-fairway: no. I can change it

dsnyder-fairway:
Hmm. @jtreuting I am not sure the UI really indicates what the results should be. For instance, say we select just a claim code of "Reasoning". When viewing the results of that we will see all Math results only containing "Reasoning" claim codes, but in ELA we see the 4 claim codes selected. I would expect no results for ELA. Should we instead have a list of Claim codes for every subject, each defaulting to "All"?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultClaimService.java
malaffoon: Ok, if it is consistent i guess we can leave it. But i think it is a bad pattern to follow: it is reasonable for a client to assume the collection being returned can be used as is. The only reason to make a new collection is if you want to modify it, or perhaps if you want to change the semantics (e.g. List -> Set).

jtreuting:
@dsnyder-fairway i think that sounds reasonable.  but it might be easier to see it in practice and discuss.  so let's get it in as it is now and then we can play around and see what changes might be needed

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/AbstractAggregateQueryHandler.java
esotrope: i would actually vote for brackets on these one-liner if-do statements

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultClaimService.java
esotrope: We should probably do a cleanup run on this at some point. I think originally I introduced the pattern to make sure the service layer returned immutables, but that isn't really necessary as long as the elements in the collections are immutable.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.ts
```diff
@@ -126,6 +125,28 @@ export interface AggregateReportFormSettings {
 
   };
 
+  /**
+   * Claim report assessment settings
+   */
+  claimReport: {
```
esotrope: this report form has gotten silly hasn't it x j

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
esotrope: you need to introduce new validation form controls for these new form controls. also we will need to programatically switch validators when report type changes. you should see where that is done now in this file

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
esotrope: yep it's _todone_

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
dsnyder-fairway: This is for creating sample data though on the preview table. The ability to change between General Population and Cohort only appeart when viewing the results

esotrope: yeah there is what the user selects on the form and then basically here we are computing "effective form settings" which are not always exactly what the user clicked.
e.g. i may have explicitly set longitudinal settings then changed assessment type to ICA (does not support longitudinals) then submitted the form - this should yield a general population report.
Designing it this way allows for, like Alla said, toggling back and forth form settings.

dsnyder-fairway: Should I instead check to see if the aggregateReportTypes has GeneralPopulation (IAB only)?

esotrope:
@dsnyder-fairway couldn't easily link to the thread so i'm commenting here - i think we need to possibly be more explicit in our if-statements dealing with reportType and assessment def. Maybe we can introduce a util method like this:
```typescript
getEffectiveReportType(selectedReportType: string, assessmentDefinition: AssessmentDefinition): string {
  return assessmentDefinition.aggregateReportTypes.includes(selectedReportType)
    ? selectedReportType
    : DefaultReportType
}
```
then we can switch on the result of this method

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
dsnyder-fairway: updated in 3f29e29 (not sure why change isn't showing up here)

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ClaimAggregateReportQuery.java
dsnyder-fairway: for whatever reason I was able to remove the subjectCodes from the places I previously had it _except for here_. I am left with an error message during deserialization

agorina: I think @esotrope  ran into a similar problem with the longitudinal query. Try annotating 
```
 @Override
    public Set<String> getSubjectCodes()
```
with `@JsonIgnore`

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -310,6 +355,16 @@ export class AggregateReportFormComponent {
   }
 
   onReportTypeChange(): void {
+    if (this.settings.reportType === 'Claim') {
+      this.filteredOptions.assessmentTypes = this.options.assessmentTypes.filter(assessmentType => assessmentType.value !== 'iab');
+      this.filterClaimCodes();
+    } else if (this.settings.reportType === 'LongitudinalCohort') {
+      this.filteredOptions.assessmentTypes = this.options.assessmentTypes.filter(assessmentType => assessmentType.value !== 'iab'
+        && assessmentType.value !== 'ica');
```
esotrope: looks like another place we want effective report type? we should never be checking `== assessment type` btw. that was the point of introducing `AssessmentDefinition`

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -375,18 +430,48 @@ export class AggregateReportFormComponent {
     this.showAdvancedFilters = !this.showAdvancedFilters;
   }
 
-  onAssessmentTypeChange(): void {
+  setColumnOrder() {
+
+    let order = this.currentAssessmentDefinition.aggregateReportIdentityColumns.concat();
```
esotrope: doesn't this effectively reset the order back to defaults? the goal here is to let the user switch between assessment types and to try our best to preserve their order. since this _is_ annoyingly complicated we should probably add tests around this OR just reset the order to defaults when the assessment type changes. @jtreuting @wtritch thoughts?

webapp/src/main/webapp/src/app/aggregate-report/wide-radio-group.component.ts
```diff
@@ -0,0 +1,118 @@
...
+const DefaultButtonStyles = 'btn-primary';
+
+@Component({
+  selector: 'wide-radio-group',
+  templateUrl: './wide-radio-group.component.html',
+  providers: [
+    Forms.valueAccessor(WideRadioGroupComponent)
+  ]
+})
+export class WideRadioGroupComponent extends AbstractControlValueAccessor<any[]> implements OnInit {
```
esotrope: we should address this in a followup. there should be a way to get what we want without a new component here (i hope 🤞 )

webapp/src/main/webapp/src/app/report/aggregate-report-request.ts
```diff
@@ -42,6 +42,9 @@ export interface AggregateReportQuery {
 
   // LongitudinalCohort report type params
   readonly toSchoolYear?: number;
+
+  // Claim report type params
+  claimCodesBySubject?: any;
```
esotrope: nit: readonly

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
dsnyder-fairway: this was a workaround to get the optional (only when claim is selected) orderable column of Claim to appear here. We do lose the preservation of order though.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
dsnyder-fairway: Updated. This is to have the list of assessmentTypes change when selecting a report type.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
jtreuting: let's keep it as resetting to default for now and we can play with it and see how it feels then decide based on that.

PR #1320, opened 2018-05-19
Aligned the test data and code with the latest changes to Redshift schema.

agorina:
Related to SmarterApp/RDW_Schema#275.


aggregate-service/src/main/resources/claim-report-template.sql.yml
```diff
@@ -7,30 +7,31 @@ sqlbuilder:
...
+              ecs.subject_claim_score_id,
               scs.code as claim_code
 
           from: >-
-              fact_exam_claim_score fe
-              JOIN subject_claim_score scs ON scs.id = fe.subject_claim_score_id
-              JOIN asmt a ON a.id = fe.asmt_id
+              exam_claim_score ecs
+              JOIN exam e on e.id = ecs.exam_id
+              JOIN subject_claim_score scs ON scs.id = ecs.subject_claim_score_id
```
agorina: This 'join' is the main change.

aggregate-service/src/main/resources/claim-report-template.sql.yml
```diff
@@ -7,30 +7,31 @@ sqlbuilder:
...
               a.type_id = :asmt_type_id AND
               scs.code in (:subject_claim_score_codes)
 
           groupBy: >-
-              fe.school_year,
-              fe.asmt_id,
-              fe.subject_claim_score_id,
+              e.school_year,
+              e.asmt_id,
+              ecs.subject_claim_score_id,
```
agorina: As mentioned here https://github.com/SmarterApp/RDW_Schema/pull/275#pullrequestreview-121611239, we may need to play with it when we are load-testing.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcClaimReportRepositoryEmbargoRST.java
```diff
@@ -94,7 +94,9 @@ public void isShouldReturnSummativeExamForStateWithEmbargoEnabledForEmbargoAdmin
             statements = {
                     "INSERT INTO state_embargo (aggregate, migrate_id) VALUES (1, -10)",
                     "INSERT INTO asmt_active_year (asmt_id, school_year) VALUES (-11, 1998)",
-                    "UPDATE fact_exam_claim_score SET school_year = 1998 WHERE asmt_id = -11"})
+                    "UPDATE exam_claim_score SET school_year = 1998 WHERE asmt_id = -11",
+                    "UPDATE exam SET school_year = 1998 WHERE asmt_id = -11"
```
agorina: Since we now `JOIN` between these tables, the data should be aligned.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcClaimReportRepositoryRST.java
```diff
@@ -788,12 +788,15 @@ public void itShouldFindByQueryWithMultipleValuesOfDiffParams() {
...
+            statements = {"INSERT INTO  exam_claim_score (id, subject_claim_score_id, exam_id, school_year, asmt_id, category," +
+                    "                                     student_id, completed_at, updated, update_import_id, migrate_id) VALUES " +
+                    "(-500, 4, -34, 2000 , -10, 1,  -9, '2016-08-14 19:05:33.000000', '2016-09-14 19:05:33.000000', -1, -1)",
+                    "INSERT INTO  exam (id, school_year, asmt_id, elas_id, completeness_id," +
+                            "                                    administration_condition_id, performance_level," +
+                            "                                    scale_score, grade_id, student_id, school_id," +
+                            "                                    iep, lep, section504, economic_disadvantage, migrant_status," +
+                            "                                    completed_at, updated, update_import_id, migrate_id) VALUES " +
+                            "(-34, 2000 , -10, -11, -9, -99, 1, 2500, -4, -9, -9, true, true, false, false, true, '2016-08-14 19:05:33.000000', '2016-09-14 19:05:33.000000', -1, -1)",
```
agorina: Ditto.

aggregate-service/src/main/resources/claim-report-template.sql.yml
```diff
@@ -7,30 +7,31 @@ sqlbuilder:
...
               scs.code as claim_code
 
           from: >-
-              fact_exam_claim_score fe
-              JOIN subject_claim_score scs ON scs.id = fe.subject_claim_score_id
-              JOIN asmt a ON a.id = fe.asmt_id
+              exam_claim_score ecs
+              JOIN exam e on e.id = ecs.exam_id
+              JOIN subject_claim_score scs ON scs.id = ecs.subject_claim_score_id
+              JOIN asmt a ON a.id = e.asmt_id
```
malaffoon: i don't think it will matter but you could join this on ecs.asmt_id instead (don't know why but i always like joining tables to the "main" query table in preference to chaining joins -- i know, intellectually, it doesn't matter but it's a habit).

agorina: I see your point in this case since it is more alined with the start schema design. I will change it. Although I highly hope that the query optimizer is smart enough to handle it regardless.

agorina: I am taking it back. I remember the reason I did it this way: `e.school_year, e.asmt_id` is part of the `GROUP BY` and`exam` includes it into its `SORTKEY (school_year, asmt_id, school_id, student_id)`.

PR #1321, opened 2018-05-21
Target report query and returned row models.

agorina:


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetReportingLevels.java
```diff
@@ -0,0 +1,13 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * Represents target report reporting levels
+ */
+public enum TargetReportingLevels {
```
esotrope: I think we can lose the plural s here right?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetReportingLevels.java
```diff
@@ -0,0 +1,13 @@
...
+    Above,
+    Near,
+    Below,
+    InsufficientData,
+    NoData,
+    Excluded
```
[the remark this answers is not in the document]
esotrope: Guessing `Excluded` means there was a level but it is suppressed intentionally?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetReportingLevels.java
agorina: You are right.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetReportingLevels.java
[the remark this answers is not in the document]
agorina: Yes, your guess is correct.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetReportingLevels.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Represents target report reporting levels
+ */
+public enum TargetReportingLevels {
+    Above,
+    Near,
+    Below,
+    InsufficientData,
+    NoData,
```
jtreuting: do we need these for UI purposes or are they really the same from the UI perspective?  I'm guessing InsufficientData is standard error > 0.2 and NoData is .... well ... no data.  I think we will display them the same from the UI but i guess letting the UI decide how to handle that makes sense

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetRow.java
```diff
@@ -0,0 +1,78 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * A row returned as part of the report in response to the {@link TargetReportQuery}
+ * <p>
+ * {@link AggregateRow#measures} represent the overall average scale score and number of
+ * students for the respective organization/assessment/dimension combination
```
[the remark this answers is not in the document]
jtreuting: as discussed offline, Option 1 seems like the best approach for this

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetReportingLevels.java
agorina: This is for both UI and CSV. I am okay with removing it, but I thought since we have this information - why not include?

jtreuting: I believe in the UI we would map insufficient and no data to the same icon or else we would need to expand the legend to have 5 things, when in reality they are both insufficient.  i like the idea of having the ability to handle it either way in the UI.  in practice i think for both the CSV and the UI we would map them to the same thing for display purposes as insufficient.

i'm fine returning it, but then we would have 2 places to keep in sync for what those map to.

i'm fine with either approach i think.  i like the flexibility but don't think we will utilize it right now at least.

agorina: If you think we will not utilize it, then I am going to remove it.

jtreuting: I don't think we will.  @malaffoon ?

agorina: Talked to Mark offline. He agrees with treating it the same as `InsufficientData`. Removing.

PR #1323, opened 2018-05-21
Feature/account for configurable subjects

wtritch:
Account for required schema changes in warehouse due to subject ingest.
See https://github.com/SmarterApp/RDW_Schema/pull/276 for associated schema change PR


wtritch:
Abandoning this as no longer required

PR #1324, opened 2018-05-21
handle loading of all the initial exams as a forkJoin so that the ini…

jtreuting:
…tial selected assessments list always includes all of them since it waits for them to all load

Wanted to get some thoughts on the approach to fixing https://jira.fairwaytech.com/browse/DWR-1776

When the user pre-selects assessments on the IAB Dashboard and passes those IDs to the group results page, we take the IDs and iterate over them to get the results for each assessment.  Because of that, there are cases where it displays the Selected Assessments on the page before all of the exam data is back so it will list 2 out of the 3 selected ones there.

I saw 2 approaches: 1) every time we get a API call back with exam results specifically trigger redrawing the list of selected assessments or 2) wait until we get all of the results back and use a forkJoin to basically tell the user we are Loading results until they are all received.

While option 1 is probably simpler, I went with 2 because it feels like it might be confusing if we start getting results and sorting them on the results and the selected list it seems like it might be confusing.  It feels like if the user specifically selected 4 assessments they will want to see those when the page loads.

But I'm not sold on it and could be convinced otherwise.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.html
@@ -187,8 +187,12 @@
 </ng-container>
 
 <!-- No Results Error -->
-<div class="alert alert-info" *ngIf="assessmentExams.length == 0 || hideAssessments">
+<div class="alert alert-info" *ngIf="(!loadingInitialResults && assessmentExams.length == 0) || hideAssessments">
   {{'common.assessments.no-results-found' | translate}}
+</div>
+
+<div class="alert alert-info" *ngIf="loadingInitialResults && assessmentExams.length == 0">
+  <i class="fa fa-spinner fa-pulse fa-2x"></i> {{'common.messages.loading' | translate}}
 </div>
 
 <div class="fab">
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -16,6 +16,8 @@ import { share } from 'rxjs/operators';
 import { ApplicationSettingsService } from '../app-settings.service';
 import { forkJoin } from 'rxjs/observable/forkJoin';
 import { Utils } from '../shared/support/support';
+import { Exam } from './model/exam.model';
+import { Observable } from 'rxjs/Observable';
 
 /**
  * This component encompasses all the functionality for displaying and filtering
@@ -78,9 +80,10 @@ export class AssessmentsComponent implements OnInit {
 
   filterOptions: ExamFilterOptions = new ExamFilterOptions();
   availableAssessments: Assessment[] = [];
-  assessmentsLoading: any[] = [];
+  assessmentsLoading: Map<number, Assessment> = new Map<number, Assessment>();
   minimumItemDataYear: number;
   exportDisabled = true;
+  loadingInitialResults = true;
 
   get assessmentExams(): AssessmentExam[] {
     return this._assessmentExams;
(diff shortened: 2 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -187,10 +187,14 @@ <h3>{{'assessments.results.title' | translate}}</h3>
 </ng-container>
 
 <!-- No Results Error -->
-<div class="alert alert-info" *ngIf="assessmentExams.length == 0 || hideAssessments">
+<div class="alert alert-info" *ngIf="(!loadingInitialResults && assessmentExams.length == 0) || hideAssessments">
   {{'common.assessments.no-results-found' | translate}}
 </div>
 
+<div class="alert alert-info" *ngIf="loadingInitialResults && assessmentExams.length == 0">
```
jtreuting: new Loading... message instead of the "No results" message showing for a bit and then showing results.  in general this feels like this needs to be there

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -188,17 +191,27 @@ export class AssessmentsComponent implements OnInit {
...
+        const loadingObservables: Observable<Exam[]>[] = [];
+
         const preselectedAssessments = availableAssessment.filter((assessment) => assessmentIds.split(',').indexOf(assessment.id.toString()) >= 0);
         preselectedAssessments.forEach((assessment) => {
           assessment.selected = true;
-          this.loadAssessmentExam(assessment);
+          loadingObservables.push(this.loadAssessmentExam(assessment));
+        });
+
+        forkJoin(loadingObservables).subscribe(args => {
```
jtreuting: this gets all of the calls to load assessment exam and waits for them all to return to process it and set flags

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -268,7 +281,8 @@ export class AssessmentsComponent implements OnInit {
 
   selectedAssessmentsChanged(assessment: Assessment) {
     if (assessment.selected) {
-      this.loadAssessmentExam(assessment);
+      this.loadAssessmentExam(assessment)
+        .subscribe(exams => this.processLoadAssessmentExam(assessment.id, exams));
```
jtreuting: i changed `loadAssessmentExam` to return Observable to allow for the forkJoin above

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -295,34 +309,27 @@ export class AssessmentsComponent implements OnInit {
...
-    }
+    this._assessmentExams.push(assessmentExam);
+    this._assessmentExams.sort(ordering(byGradeThenByName).on<AssessmentExam>(assessmentExam => assessmentExam.assessment).compare);
 
-    this.assessmentsLoading = this.assessmentsLoading.filter(loading => this.selectedAssessments.some(selected => loading.assessment.id == selected.id));
+    this.assessmentsLoading.delete(assessmentId);
+  }
+
+  private removeUnselectedAssessmentExams() {
+    this._assessmentExams = this._assessmentExams.filter(loaded => this.selectedAssessments.some(selected => loaded.assessment.id == selected.id));
```
jtreuting: one thing that was lost here was the ability to stop an inflight API call by unsubscribing if the assessment is removed before the API returns.  i don't think this is a real loss but worth noting the change.

PR #1330, opened 2018-05-22
Task/exam search by user group

esotrope:
Adds support for exam search by "userGroupId" instead of just "groupId"


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/GroupAssessmentSearch.java
```diff
@@ -4,7 +4,8 @@
 
 class GroupAssessmentSearch extends AbstractAssessmentSearch {
 
-    private long groupId;
+    private Long groupId;
+    private Long userGroupId;
```
esotrope: client will pass one or the other.
Looking back, we could probably join the school-grade based search and user group based search into one pipeline instead of having two very similar parallel channels for exam search

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractAssessmentRepository.java
```diff
@@ -36,11 +37,11 @@ protected AbstractAssessmentRepository(
...
-        return ImmutableMap.<String, Object>builder()
-                .putAll(securityParameterProvider.getSecurityParameters(permissionSource))
-                .put("school_year", search.getSchoolYear())
-                .putAll(getParameters(search))
-                .build();
+        final Map<String, Object> parameters = newHashMap();
+        parameters.putAll(securityParameterProvider.getSecurityParameters(permissionSource));
+        parameters.put("school_year", search.getSchoolYear());
+        parameters.putAll(getParameters(search));
+        return parameters;
```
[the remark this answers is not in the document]
malaffoon: Yeah, i ran into that. Should this method return SqlParameterSource instead?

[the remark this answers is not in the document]
esotrope: i wanted it to, but MapSqlParameterSource.addXYZ doesn't allow adding another parameter source *shrug.
I figure if we combine the school grade based exam search with group based we can do away with the AbstractExamXYZ classes anyway

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -343,3 +316,37 @@ sql:
...
+      schoolGradeOrGroupOrUserGroupFilters: >-
+        (
+          (
+            :group_id IS NULL
+            OR EXISTS(
+              SELECT 1 FROM student_group_membership sgm
+                JOIN student_group sg
+                  ON sg.id = sgm.student_group_id
+                JOIN user_student_group usg
+                  ON usg.student_group_id = sgm.student_group_id
```
malaffoon: nit: if this is consistent just ignore me but ... i prefer the `ON` clause to be on the same line as the `JOIN` because, when i'm reading the query i assume the author got the ON correct and all i'm interested in is the table being joined and its alias; the extra ON line disrupts my reading.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -343,3 +316,37 @@ sql:
              OR 1 = :state_embargo_admin
              OR s.district_id IN (:district_embargo_admin_ids)
           )
+
+      # deps: st.id, a.subject_id, :group_id, :user_group_id
+      schoolGradeOrGroupOrUserGroupFilters: >-
```
esotrope: Right now this snippet uses sub-selects to check for group existence. previously we used joins instead of an exists clause. Thoughts on doing this? or should we lean towards joins for performance?

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -343,3 +316,37 @@ sql:
...
+        (
+          (
+            :group_id IS NULL
+            OR EXISTS(
+              SELECT 1 FROM student_group_membership sgm
```
malaffoon: Question: i know it probably won't make a functional difference but why are we checking student membership? Don't we just care if the user is a member of the group?

reporting-service/src/test/resources/integration-test-data.sql
```diff
@@ -165,23 +167,47 @@ insert into student_group_membership (student_group_id, student_id) values
...
-   (-90, 'someone-10@somewhere.com');
-
--- transfer student test data
-insert into student (id, ssid, last_or_surname, first_name, gender_id, gender_code,
-                     birthday, inferred_school_id, update_import_id, updated, migrate_id) values
-  (-100, 'transfer_1', 'from_school3', 'to_school4', -1, 'g1', '1997-01-01 00:00:00.000000', -40, -1, '1997-07-18 20:14:34.000000', -1);
+   (-90, 'someone-10@somewhere.com'),
+   (-100, 'someone-10@somewhere.com');
+
+-- user groups
```
esotrope: added user groups that mirror the other groups setup in this file for ease of testing

reporting-service/src/main/resources/application.sql.yml
esotrope: good point let me tweak it and see if it passes test at least

esotrope: tried this but didn't pass test:
```sql
(
            :group_id IS NULL
            OR EXISTS(
              SELECT 1
              FROM student_group sg
                JOIN user_student_group usg ON usg.student_group_id = sg.id
              WHERE sg.id = :group_id
                AND usg.user_login = :user_login
                AND (sg.subject_id is null OR sg.subject_id = a.subject_id)
            )
          )
```
i think making sure the student belongs to the group is needed for filtering. so this query does two things - makes sure the user owns the requested group and that the student belongs to that group

reporting-service/src/main/resources/application.sql.yml
agorina: To answer your question, we would need to load test the app. Depending on the query complexity, optimizer may convert in into JOIN, or do something else. I have seen conflicting results with the performance in both cases. I am leaning towards keeping it the way you have it and deciding during load testing if this is an issue.

agorina: Oh, I see - you are trying to short circuit it. Good point.

esotrope:
good to merge here?

PR #1333, opened 2018-05-23
adds api support of reports by user group

esotrope:
Same changes as in the last PR but for the remaining group based queries in the report-processor


common/src/main/resources/common.sql.yml
```diff
@@ -227,4 +227,34 @@ sql:
       ei.trait_evidence_elaboration_score as ei_trait_evidence_elaboration_score,
       ei.trait_organization_purpose_score as ei_trait_organization_purpose_score,
       ei.trait_conventions_score as ei_trait_conventions_score,
-      i.performance_task_writing_type
\ No newline at end of file
+      i.performance_task_writing_type
+
+    # dependencies: student st, asmt a, :group_id, :user_group_id
+    groupFilters: >-
```
esotrope: moved to common snippets since report-processor now shares it

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcStudentRepositoryIT.java
```diff
@@ -128,6 +128,28 @@ public void itShouldFindStudentIdsByGroupId() {
         assertThat(g2Students.stream().map(Student::getId)).containsOnly(-1L);
     }
 
+    @Test
```
esotrope: adds spot tests for making sure the queries work with the snippet

common/src/main/resources/common.sql.yml
```diff
@@ -227,4 +227,34 @@ sql:
...
+            SELECT 1
+            FROM teacher_student_group_membership tsgm
+              JOIN teacher_student_group tsg ON tsg.id = tsgm.teacher_student_group_id
+            WHERE tsgm.teacher_student_group_id = :user_group_id
+              AND tsgm.student_id = st.id
+              AND tsg.user_login = :user_login
+              AND (tsg.subject_id IS NULL OR tsg.subject_id = a.subject_id)
+          )
+        )
+      )
```
wtritch: So this clause doesn't provide security, just selection filtering right?

report-processor/src/main/resources/application.sql.yml
```diff
@@ -46,12 +46,8 @@ sql:
       LEFT JOIN exam_item ei ON ei.exam_id=e.id
       LEFT JOIN item i ON i.id=ei.item_id
       ${sql.reportProcessor.snippet.percentileLeftJoin}
-      JOIN student_group_membership sgm ON st.id = sgm.student_id
-      JOIN student_group sg ON sgm.student_group_id = sg.id AND (sg.subject_id is null or a.subject_id = sg.subject_id)
-      JOIN user_student_group usg ON usg.student_group_id = sg.id
```
wtritch: This clause provided both security and selection filtering.  If the `sql.snippet.groupFilters` doesn't provide security, where does the security filtering in this query happen?

esotrope: please elaborate.
the same level of security is still there in the groupFilters snippet when filtering by "admin group" but i see this query does need a check that "there exists a group that the user owns that has the student", missed it because the other queries i modified all had the security snippet at the end

esotrope: double checked - this should be good to merge Will - let me know if you have any more input otherwise i'll probably merge it sometime later tomorrow.

wtritch:
Lgtm

PR #1335, opened 2018-05-23
changed IAB colors to red, yellow, green

jtreuting:
right now i just put this in styles.less instead of the UI Kit.  wanted to get thoughts on this.  they feel RDW specific so not sure if they should go there.

also, I had a hard time with the names of these colors.  there are already some greens and reds so didn't know what to name them.  i didn't want to go with iab1, iab2, iab3 since we wanted to make colors available to the configurable subjects.

thoughts?  @wtritch, @esotrope 

image

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -924,15 +924,15 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
 }
 
 /**
- * New IAB Colors - TODO move to UI Kit?
+ * New IAB Colors
  */
-@iab-green: #2c7f3e;
-@iab-yellow: #eccb1a;
-@iab-red: #750f2c;
+@sb-iab-green: #2c7f3e;
+@sb-iab-yellow: #eccb1a;
+@sb-iab-red: #750f2c;
 
 // Create color classes
 // List the colors
-@colorVars: iab-green, iab-yellow, iab-red;
+@colorVars: sb-iab-green, sb-iab-yellow, sb-iab-red;
 // Call the generator mixin
 // Yes, it's supposed to be called in recursive order
 .generate-color-classes(@colorVars, 1);
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -924,6 +924,28 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
 }
 
 /**
+ * New IAB Colors - TODO move to UI Kit?
+ */
+@iab-green: #2c7f3e;
+@iab-yellow: #eccb1a;
+@iab-red: #750f2c;
+
+// Create color classes
+// List the colors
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -923,6 +923,28 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
...
+/**
+ * New IAB Colors - TODO move to UI Kit?
+ */
+@iab-green: #2c7f3e;
+@iab-yellow: #eccb1a;
+@iab-red: #750f2c;
+
+// Create color classes
+// List the colors
+@colorVars: iab-green, iab-yellow, iab-red;
```
jtreuting: this logic was pulled form the UI Kit to generate a whole bunch of color specific classes that are needed

esotrope: ideally we have some indirection with our colors where we separate a color's role from implementation or pallets from pallet colors
```less
@red: #ff0000
...
@performance-level-pallet-1: @red, @yellow, @green
// generate with method:
.performance-level-pallet-1 {
  .color-1 {
    background-color: @red
  }
  .color-2 {
    background-color: @yellow
  }
  ...
}
```
i think we get into trouble when we call colors "red" "blue" etc. this is an implementation and not a color role/purpose in the app.

jtreuting: while i agree with that in general, we are moving towards the configurable subjects being able to define what colors they use.  so that config file will available certain named colors (like red, green, blue, etc.) that they can choose from and not something like choose pallet-1 or pallet-2.  this also goes for the SB content which would define these colors there as well.

so i do think in this case (which is not the usual case for css naming) we want them to be color named instead of a role or purpose.

@wtritch, you are heading up the config subject part that involves the color, thoughts?

wtritch: I'd argue we should use color names that we can add to the XSD as viable colors and a name like "sbac-iab-green" sounds off...  I don't believe we have a "yellow" yet so that would work for yellow.  How different are these "green" and "red" values from the existing "green" and "red" values?  Could we get away with something like "yellow-green" or "blue-green" or "orange-red" etc?

wtritch: Or maybe change our existing green and red values to match these green and red values?

jtreuting: i'm a bit nervous to do that because they are defined in the UI Kit and IAT might be using them as well, so not sure we should change the existing ones.  You are right we don't have a yellow, so that can just be yellow.

Thoughts on if they are more appropriate in the main UI Kit or the styles.less in RDW?

wtritch: I'd put them in the main UI kit if they're general colors like "yellow"  If they're going to be named something like "sbac-yellow" or "rdw-yellow" or "instructional-resource-yellow" (or whatever the source doc for these colors is) then I'd probably leave them in RDW

PR #1336, opened 2018-05-23
combination visitor as suggested in previous review

malaffoon:
Alla, this is what i was thinking. Let me know if you want me to commit or not.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcClaimReportRepository.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcClaimReportRepository.java
@@ -119,17 +119,13 @@ public class JdbcClaimReportRepository
         final Map<ActiveAssessment, Set<String>> claimsByAssessment = getClaimsByAssessment(queryWithReportTemplate);
 
         final Map<ReportRowIdentity, ClaimRow> reportTemplate = newHashMap();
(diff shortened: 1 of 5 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcClaimReportRepository.java
```diff
@@ -119,17 +119,13 @@ protected ReportRowIdentity createReportRowIdentity(final ClaimRow reportRow, fi
...
-        // a template report consists of permutations of all the organizations and assessments
-        for (final Organization organization : queryWithReportTemplate.getOrganizations()) {
-            for (final ActiveAssessment assessment : queryWithReportTemplate.getAssessments()) {
-                for (final Dimension dimension : queryWithReportTemplate.getDimensions()) {
-                    for (final String claimCode : claimsByAssessment.get(assessment)) {
-                        final ClaimRow row = buildEmptyReportRow(organization, assessment, dimension, query, claimCode);
-                        reportTemplate.put(createReportRowIdentity(row, considerAsmtIdAsPartOfIdentity), row);
-                    }
-                }
+        // a template report consists of combinations of all the organizations and assessments
```
malaffoon: As a mental trick for permutation vs combination: a combination lock should really be called a permutation lock because the order matters and permutations consider order while combinations do not.

PR #1337, opened 2018-05-23
Target report query handler and converter.

agorina:
Sorry for a lot of changes due to renaming. Should have probably done it separately.

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/LongitudinalQueryConverter.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/LongitudinalQueryConverter.java
@@ -34,7 +34,7 @@ class LongitudinalQueryConverter extends AbstractQueryConverter<LongitudinalRepo
     @Override
     protected LongitudinalOrganizationQuery.Builder getOrganizationTypeQueryBuilder(final LongitudinalReportQuery reportQuery,
                                                                                     final StudentFilters filters) {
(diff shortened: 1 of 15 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetReportQuery.java
```diff
@@ -25,6 +25,7 @@ public AggregateReportType getReportType() {
     }
 
     @Override
+    @JsonIgnore
```
agorina: Unrelated, but I think it is needed here. 
Basing it on the previous issues we ran into with other queries.

malaffoon:
Woot, `1337`!

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/LongitudinalQueryConverter.java
```diff
@@ -34,7 +34,7 @@
     @Override
     protected LongitudinalOrganizationQuery.Builder getOrganizationTypeQueryBuilder(final LongitudinalReportQuery reportQuery,
                                                                                     final StudentFilters filters) {
-        checkArgument(reportQuery.getSubjectCodes().size() == 1, "Only one subject is supported for the longitudinal report queries");
+        checkArgument(reportQuery.getSubjectCodes().size() == 1, "invalid query, must include one subject code");
```
malaffoon: nit: perhaps `... must include exactly one subject code`

PR #1338, opened 2018-05-23
display claim reports as having no average scale score, no 4th perfor…

dsnyder-fairway:
…mance level, and cannot group achievement levels

When report type is Claim we treat it much like we do an IAB for display purposes

screen shot 2018-05-23 at 12 45 12 pm


dsnyder-fairway:
There is a minor bug here. On the query builder page whenever we switch  report types to 'Claim' we get a bunch of `AggregateReportTableComponent.html:19 ERROR TypeError: Cannot read property 'id' of undefined
    at Object.eval [as updateDirectives] (AggregateReportTableComponent.html:23)` for a short period of time

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -696,7 +696,7 @@ <h3>{{'aggregate-report-form.section.preview-heading' | translate}}</h3>
                                        analyticsEvent="Change"></sb-radio-button-group>
               </span>
               <span class="pull-right flex-child ml-xs"
-                    *ngIf="currentAssessmentDefinition.performanceLevelDisplayTypes.length > 1">
+                    *ngIf="effectiveReportType !== 'Claim' && currentAssessmentDefinition.performanceLevelDisplayTypes.length > 1">
```
dsnyder-fairway: disable grouping for claim

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -36,26 +36,35 @@ export class AggregateReportTableDataService {
...
       for (const grade of settings.claimReport.assessmentGrades) {
         for (const year of settings.claimReport.schoolYears) {
           gradesAndYears.push({ grade, year });
         }
       }
     }
 
+    if (settings.reportType === 'Claim') {
+      assessmentDefinition = <AssessmentDefinition>{
+        typeCode: 'iab',
```
dsnyder-fairway: this is a bit strange. I am trying to mimic the 'iab' assessment definition since we want to use most of it for display purposes, though iab assessments aren't available for claim reports

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -39,6 +44,20 @@ export class AggregateReportService {
...
   }
 
+  /**
+   * Gets the effective report type
+   *
+   * @param {"GeneralPopulation" | "LongitudinalCohort" | "Claim"} reportType the report type
+   * @param {AssessmentDefinition} definition the assessment definition
+   * @returns {"GeneralPopulation" | "LongitudinalCohort" | "Claim"}
+   */
+  getEffectiveReportType(reportType: 'GeneralPopulation' | 'LongitudinalCohort' | 'Claim', definition: AssessmentDefinition):
```
dsnyder-fairway: moved this from AssessmentDefinitionService to here

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -412,16 +458,18 @@ export class AggregateReportTableComponent implements OnInit {
...
-            id: 'performanceLevel',
-            displayType: displayType,
-            level: level,
-            visible: this.performanceLevelDisplayType === displayType,
-            index: index,
-            field: `performanceLevelByDisplayTypes.${displayType}.${this.valueDisplayType}.${index}`,
-            headerKey: this.getPerformanceLevelColumnHeaderTranslationCode(displayType, level, index),
-            headerColor: this.colorService.getPerformanceLevelColorsByAssessmentTypeCode(assessmentDefinition.typeCode, level)
-          }));
+          if (level !== undefined && !isNaN(level)) {
```
dsnyder-fairway: this `if` might no longer be needed...

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -96,6 +99,27 @@ export class AggregateReportTableComponent implements OnInit {
     return this._performanceLevelDisplayType;
   }
 
+  get cutPoint(): number {
+    if (this._table.reportType === 'Claim') {
+      return this._cutPoint;
+    }
+    return this.table.assessmentDefinition.performanceLevelGroupingCutPoint;
```
jtreuting: it looked like you were defining an assessment definition for use when it is claims above (https://github.com/SmarterApp/RDW_Reporting/pull/1338/files#diff-8bf5cdbf11923e70261bf1468909cc16R60) but this is not using it

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -274,27 +315,32 @@ export class AggregateReportTableComponent implements OnInit {
...
       new Column({ id: 'achievementComparison', sortable: false }),
-      new Column({ id: 'avgScaleScore' }),
+      ...hasClaimCodes ? [] : [ new Column({ id: 'avgScaleScore' }) ],
       ...this.createPerformanceLevelColumns(performanceLevelsByDisplayType, assessmentDefinition)
     ];
 
     this.calculateTreeColumns();
   }
 
+  private configureClaimReport(): AssessmentDefinition {
```
jtreuting: seems like this was defined up here as well?  https://github.com/SmarterApp/RDW_Reporting/pull/1338/files#diff-8bf5cdbf11923e70261bf1468909cc16R60

should this be in the service instead so it's in one place?  also I'm not familiar with the assessment definition like @esotrope is, but seems like we need a way to get a claim Assessment Definition from that service so that it can be used like the other ones.  Name might need changing, but that's a small thing.  And then should this definition also include a flag for if the average scale score should be shown or not.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -263,9 +290,23 @@ export class AggregateReportTableComponent implements OnInit {
     this._orderingByColumnField[ 'assessmentGradeCode' ] = assessmentGradeOrdering;
     this._orderingByColumnField[ 'schoolYear' ] = SchoolYearOrdering;
     this._orderingByColumnField[ 'subgroup.id' ] = subgroupOrdering(item => item.subgroup, options);
-    this._orderingByColumnField[ 'claimCode' ] = ClaimOrdering;
+    if (hasClaimCodes) {
+      this._orderingByColumnField[ 'claimCode' ] = ClaimOrdering;
```
esotrope: doesn't hurt to add even when the table doesn't use it - it will just be ignored

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -263,9 +290,23 @@ export class AggregateReportTableComponent implements OnInit {
...
+      new Column({ id: 'assessmentLabel' }),
+      new Column({ id: 'schoolYear' }),
+      new Column({ id: 'dimension', field: 'subgroup.id' }),
+      ...hasClaimCodes ? [ new Column({ id: 'claim', field: 'claimCode' }) ] : []
+    ];
 
     // create columns
+    if (hasClaimCodes) {
+      assessmentDefinition = this.configureClaimReport();
+    }
```
esotrope: i have to say this seems very hacky. still looking at this and getting into the weeds but it seems like we are bending the current design to work how we want instead of redesigning it based on new variables we didn't anticipate

dsnyder-fairway: I agree. With Claim reports we really need to abandon the assessment definitions. I'll look into making these changes

esotrope: just for reference the assessment definition abstraction was to insulate the aggregate ui from configurable subjects. and was intended to be a product of assessmentType, and subject. We might need another or additional abstraction to represent display config for an assessmentType + subject + reportType

dsnyder-fairway:
re-opened after making the changes to the assessment definition. Now I am trying to understand why I am unable to retrieve state results from a Summative / Claim query

webapp/src/main/webapp/src/app/shared/color.service.ts
```diff
@@ -49,4 +54,15 @@ export class ColorService {
     return PerformanceLevelColorsByAssessmentTypeCode.get(code)[ performanceLevel - 1 ];
   }
 
+  /**
+   * Retrieves the color for the performance level (1-based)
+   *
+   * @param {string} assessment type (ica, iab, sum)
+   * @param {number} performanceLevel (1-based)
+   * @returns {string} the class of the color
+   */
```
dsnyder-fairway: copy-paste doc. I'll update this

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
```diff
@@ -156,9 +157,14 @@ export class AggregateReportOptionsMapper {
    */
   toDefaultSettings(options: AggregateReportOptions): Observable<AggregateReportFormSettings> {
     return this.assessmentDefinitionService.getDefinitionsByAssessmentTypeCode().pipe(
-      map(definitions => {
+      map((definitions: Map<DefinitionKey, AssessmentDefinition>) => {
         const defaultAssessmentType = options.assessmentTypes[ 0 ];
-        const assessmentDefinition = definitions.get(defaultAssessmentType);
+        let assessmentDefinition = null;
+        definitions.forEach((value: AssessmentDefinition, key: DefinitionKey) => {
+          if (key.assessmentType === defaultAssessmentType && key.reportType === 'GeneralPopulation') {
```
dsnyder-fairway: I had to do this because I couldn't get equality to work here for `DefinitionKey`. I am sure there is an easy way to override this somehow...

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -102,6 +102,7 @@
             <performance-level-distribution-chart
               [percentages]="rowData.performanceLevelByDisplayTypes.Separate.Percent"
               [cutPoint]="cutPoint"
+              [performanceLevels]="rowData.claimCode ? 3 : null"
```
esotrope: why is 3 hardcoded here? shouldn't the value here just be assessmentDef.performanceLevelCount?

dsnyder-fairway: we use performanceLevels only when the value is not null

webapp/src/main/webapp/src/app/shared/color.service.ts
esotrope: is how many levels there are be how we lookup the color? i guess this will change when we implement configurable subjects in the ui

webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment-definition.service.ts
```diff
@@ -63,12 +87,22 @@ export class AssessmentDefinitionService {
...
-    ]));
+  public getDefinitionsByAssessmentTypeCode(): Observable<Map<DefinitionKey, AssessmentDefinition>> {
+    return of(
+      new Map([
+        [ <DefinitionKey>{ assessmentType: 'iab', reportType: 'GeneralPopulation' }, Iab ],
+        [ <DefinitionKey>{ assessmentType: 'ica', reportType: 'GeneralPopulation' }, Ica ],
+        [ <DefinitionKey>{ assessmentType: 'ica', reportType: 'Claim' }, ClaimIca ],
+        [ <DefinitionKey>{ assessmentType: 'sum', reportType: 'GeneralPopulation' }, Summative ],
+        [ <DefinitionKey>{ assessmentType: 'sum', reportType: 'LongitudinalCohort' }, Summative ],
+        [ <DefinitionKey>{ assessmentType: 'sum', reportType: 'Claim' }, ClaimSummative ]
```
esotrope: these keys should be top level constant symbols so that === can be used

webapp/src/main/webapp/src/app/shared/color.service.ts
dsnyder-fairway: it is for reportType = Claim since we no longer care if it is ICA or Summative

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.ts
```diff
@@ -88,7 +89,7 @@ export class AggregateReportFormComponent {
   /**
    * Assessment definitions for use in generating sample data
    */
-  assessmentDefinitionsByTypeCode: Map<string, AssessmentDefinition>;
+  assessmentDefinitionsMap: Map<DefinitionKey, AssessmentDefinition>;
```
esotrope: "Map" here is redundant, maybe just omit it

webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment-definition.service.ts
dsnyder-fairway: 🙇 thank you. I was wondering why I couldn't user ===

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -16,121 +16,126 @@
   <ng-template pTemplate="header"
                let-columns>
     <tr>
-      <th *ngFor="let column of columns"
-          [hidden]="!column.visible"
-          [pSortableColumn]="column.field"
-          [pSortableColumnDisabled]="!column.sortable"
-          [ngSwitch]="column.id">
+      <ng-container *ngFor="let column of columns">
+        <ng-container *ngIf="column !== undefined">
```
dsnyder-fairway: only change was here and right above the `<tr`

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -294,7 +294,7 @@ export class AggregateReportTableComponent implements OnInit {
       new Column({ id: 'schoolYear' }),
       new Column({ id: 'dimension', field: 'subgroup.id' }),
       ...hasClaimCodes ? [ new Column({ id: 'claim', field: 'claimCode' }) ] : []
-    ];
+    ].filter(col => col !== undefined);
```
dsnyder-fairway: when hasClaimCodes == false this inserted undefined in the array. To resolve I am just going to filter to non-null values

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1238,7 +1238,7 @@
     "report-type": {
       "AggregateReportRequest": "Aggregate Report",
       "AggregateReportRequest CustomAggregate": "Aggregate Report",
-      "AggregateReportRequest Claims": "Aggregate Report: Claims",
+      "AggregateReportRequest Claim": "Aggregate Report: Claims",
```
dsnyder-fairway: fixes a translation key

dsnyder-fairway:
@esotrope @jtreuting am I good to merge?

dsnyder-fairway:
merging. I'll address any feedback in future PRs

PR #1342, opened 2018-05-24
fixed mapping in admin instructional resource to assessment type inst…

jtreuting:
…ead of from assessment type code

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.ts
+++ b/webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.ts
@@ -72,7 +72,7 @@ export class InstructionalResourceService {
     resource.organizationType = serverResource.organizationType;
     resource.assessmentLabel = serverResource.assessmentLabel;
     resource.assessmentName = serverResource.assessmentName;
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.ts
```diff
@@ -72,7 +72,7 @@ export class InstructionalResourceService {
     resource.organizationType = serverResource.organizationType;
     resource.assessmentLabel = serverResource.assessmentLabel;
     resource.assessmentName = serverResource.assessmentName;
-    resource.assessmentType = serverResource.assessmentTypeCode;
+    resource.assessmentType = serverResource.assessmentType;
```
jtreuting: this mapping caused the assessmentType to be "ica" instead of "ICA" and then when it used this to send back to the server it was not able to bind that to the AssessmentType enum so it failed.

esotrope: would prefer we keep everything in the ui using the code up until we send it back to the server in that case

PR #1344, opened 2018-05-24
updated embedded styles with new IAB colors and updated IAB PDF repor…

jtreuting:
…t template


report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -66,14 +66,19 @@
                                                 th:text="${exam.scaleScore.value}"></span> <span class="error-band">&plusmn;
                                             <span th:text="${support.round(exam.scaleScore.standardError)}"
                                                   th:remove="tag"></span></span></span>
-                                        <div class="bar blue-dark"
+                                        <div class="bar"
+                                             th:classappend="#{'report.iab.level.1.color'}"
```
jtreuting: changed these to pull from the same json file as the dynamic color based on the score so it can all be changed in one place

PR #1345, opened 2018-05-24
Fix aggregate sample org generation to place generated schools under …

wtritch:
…generated districts

The issue was that we were generating 1 district and 2 schools, each with it's own districtId.  This change just binds schools to the "last-generated" district so that we end up with hierarchical generated data.


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -198,7 +198,7 @@ export class AggregateReportTableDataService {
 
       if (settings.schools.length >= schoolId
         || ((settings.districts.length || settings.includeAllDistricts) && settings.includeAllSchoolsOfSelectedDistricts)) {
-        organizations.push(this.createSampleSchool(schoolId++));
+        organizations.push(this.createSampleSchool(schoolId++, districtId - 1));
```
wtritch: small change to bind the school to the last-generated district, or bind all schools to districtId `0` if no districts are being generated.

PR #1347, opened 2018-05-24
Fix aggregate report display table sorting during paging.

wtritch:
This PR fixes the sort order of the aggregate report table when navigating to a new page of results.  The fix was in the primeNG library, so I bumped us to the latest version.  Unfortunately the latest version no longer auto-custom-sorts when the underlying data changes, so I had to put in a trigger to manually sort the table on data load.  Hopefully this won't be required in a future PrimeNG update...

I also re-generated my package-lock and node_modules directory and was unable to get `npm start` to execute due to a missing `@angular-devkit/core` dependency.  Bumping the angular CLI version from 1.5.4 to 1.7.4 resolved the issue.  I flirted with bumping angular CLI to the next major revision `6.0.x` but it caused significant issues with our production build that wasn't easily diagnosable.  We can tackle that if/when we also upgrade Angular to 6.0.x

### DIFF
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -40,12 +40,12 @@
     "mathjs": "^3.15.0",
     "ng2-file-upload": "^1.3.0",
     "ngx-bootstrap": "^2.0.0-beta.9",
-    "primeng": "5.2.2",
+    "primeng": "5.2.7",
     "rxjs": "^5.5.2",
     "zone.js": "^0.8.14"
   },
   "devDependencies": {
-    "@angular/cli": "1.5.4",
+    "@angular/cli": "1.7.4",
     "@angular/compiler-cli": "^5.0.0",
     "@angular/language-service": "^5.0.0",
     "@types/file-saver": "0.0.1",
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -115,6 +115,11 @@ export class AggregateReportTableComponent implements OnInit {
       };
       this.buildAndRender(this._table);
     }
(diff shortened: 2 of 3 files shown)


webapp/src/main/webapp/package.json
```diff
@@ -40,12 +40,12 @@
     "mathjs": "^3.15.0",
     "ng2-file-upload": "^1.3.0",
     "ngx-bootstrap": "^2.0.0-beta.9",
-    "primeng": "5.2.2",
+    "primeng": "5.2.7",
```
wtritch: This is the real fix.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -115,6 +115,11 @@ export class AggregateReportTableComponent implements OnInit {
       };
       this.buildAndRender(this._table);
     }
+    // Latest TurboTable version (1.5.7) does not sort when row data
+    // changes.  Manually trigger a sort after setting row data.
+    setTimeout(() => {
+      this.sort({data: this._table.rows});
+    });
```
wtritch: This is to fix the regression in the latest primeNG version

webapp/src/main/webapp/package.json
```diff
@@ -40,12 +40,12 @@
...
     "rxjs": "^5.5.2",
     "zone.js": "^0.8.14"
   },
   "devDependencies": {
-    "@angular/cli": "1.5.4",
+    "@angular/cli": "1.7.4",
```
wtritch: This is to resolve the build-from-scratch issues I was seeing.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
esotrope: no warning from not including delay ms here?

PR #1353, opened 2018-05-24
create separate summative report template since it needs own labels a…

jtreuting:
…nd footer is changing

allows for more separation of reports as it evolves


common/src/main/resources/i18n/en.json
```diff
@@ -841,6 +841,72 @@
         "name": "Math"
       }
     },
+    "sum": {
```
jtreuting: needed summative specific labels for the PDF report, so copied ICA and edited

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/ExamReportAssessmentType.java
```diff
@@ -1,32 +0,0 @@
-package org.opentestsystem.rdw.reporting.processor.model;
-
-import org.opentestsystem.rdw.common.model.AssessmentType;
-
-@Deprecated
```
jtreuting: was deprecated and not used anywhere so deleted it

report-processor/src/main/resources/templates/ica-body.html
```diff
@@ -122,7 +122,26 @@ <h4 class="text-center" th:text="#{'report.claim.' + ${claimCode} + '.name'}"></
                 </div>
             </div>
 
-            <div th:replace="fragments::bottom"></div>
```
wtritch: interesting, why break this out of being a fragment?

report-processor/src/main/resources/templates/fragments.html
```diff
@@ -112,29 +112,6 @@
     <hr class="mt-sm mb-sm"/>
 </div>
 
-<div th:fragment="bottom">
```
jtreuting: pulled this into the reports as we are implementing new design in ICA and Summative than IAB so won't be able to share, and Summative will not have one of the sections, so just brought into the main body templates

report-processor/src/main/resources/templates/sum-body.html
```diff
@@ -0,0 +1,142 @@
+<!DOCTYPE html>
```
jtreuting: new summative body.  multiple reasons for this:
1) needed to reference summative labels everywhere (could have solved by passing assessment type and constructing placeholder text if i wanted to)
2) we know the summative report will evolve to have different data than the ICA in the future so separating allows that to happen later

report-processor/src/main/resources/templates/ica-body.html
jtreuting: you are too fast to review, i was commenting on it as you were reviewing :)

PR #1354, opened 2018-05-24
Bump the hoek dev dependency version to remove github vulnerability w…

wtritch:
…arnings


webapp/src/main/webapp/package.json
```diff
@@ -55,6 +55,7 @@
     "@types/node": "~6.0.60",
     "codelyzer": "^4.0.1",
     "core-js": "^2.5.1",
+    "hoek": "4.2.1",
```
esotrope: hm i tried this before and still got the warning. cool if it works though

PR #1355, opened 2018-05-24
corrected school years on report preview for longitudinal report

dsnyder-fairway:


dsnyder-fairway:
looks like I need to correct some tests

webapp/src/main/webapp/src/app/admin/instructional-resource/instructional-resource.service.ts
```diff
@@ -66,7 +66,7 @@ export class InstructionalResourceService {
   }
 
   private toServerFormat(resource: InstructionalResource): InstructionalResource {
-    resource.assessmentType = AssessmentType[resource.assessmentType];
+    resource.assessmentType = AssessmentType[resource.assessmentType] ? AssessmentType[resource.assessmentType] : resource.assessmentType;
```
dsnyder-fairway: if we don't have an enum for this, use that the original value

PR #1356, opened 2018-05-25
Bugfix/dwr 1837

jtreuting:
…rticular grade

fixes: https://jira.fairwaytech.com/browse/DWR-1837

image

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/Assessment.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/Assessment.java
@@ -17,6 +17,7 @@ public class Assessment {
     private String typeCode;
     private String subjectCode;
     private String gradeCode;
+    private int gradeSequence;
     private int schoolYear;
     private List<Integer> cutPoints;
 
@@ -60,6 +61,13 @@ public class Assessment {
      */
     public String getGradeCode() {
         return gradeCode;
+    }
+
+    /**
+     * @return the sequence of the grade used for ordering
+     */
+    public int getGradeSequence() {
+        return gradeSequence;
     }
 
     /**
@@ -118,6 +126,7 @@ public class Assessment {
         private String typeCode;
         private String subjectCode;
         private String gradeCode;
+        private int gradeSequence;
         private int schoolYear;
         private List<Integer> cutPoints;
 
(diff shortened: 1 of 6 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/Assessment.java
```diff
@@ -62,6 +63,13 @@ public String getGradeCode() {
         return gradeCode;
     }
 
+    /**
+     * @return the sequence of the grade used for ordering
+     */
+    public int getGradeSequence() {
```
jtreuting: in order to properly sort by grades for this edge case we can't sort by the grade code which is a string

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.mapper.ts
```diff
@@ -202,10 +202,15 @@ export class LongitudinalCohortChartMapper {
...
     const assessmentType = assessments[ 0 ].type;
+    const grades = new Set<string>();
 
     assessments.concat()
       .sort(assessmentYearAscending)
       .forEach(assessment => {
+        // there is an edge case where multiple assessments are returned for a grade, so we only want to take one of them
+        if (grades.has(assessment.grade)) return;
+
+        grades.add(assessment.grade);
```
jtreuting: keep track so that we only take one assessment for each grade

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.mapper.ts
```diff
@@ -113,7 +113,7 @@ function createOrganization(id: number): Organization {
 }
 
 const rowYearAscending = ordering(byNumber).on<AggregateReportRow>(row => row.assessment.examSchoolYear).compare;
-const assessmentYearAscending = ordering(byNumber).on<Assessment>(assessment => assessment.schoolYear).compare;
+const assessmentYearAscending = ordering(byNumber).on<Assessment>(assessment => assessment.gradeSequence).compare;
```
jtreuting: the edge case we have has multiple assessments with the same year so we need to sort by grade

webapp/src/main/webapp/src/app/aggregate-report/results/longitudinal-cohort-chart.mapper.ts
esotrope: is there a case where we want to chose the "most appropriate" of the colliding assessments?

jtreuting: maybe we could take the most recent by academic year?  i could add that to the sorting logic, on line 116

PR #1357, opened 2018-05-25
change from "Scores" to "Points" where appropriate

dsnyder-fairway:
changing just the translations. I thought about changing the keys, but that would lead to confusion as we use writing-trait-scores as the name of some of our components as well


jtreuting:
I'd suggest backing out the Writing Trait Score changes and just leave the student score change and then when we hear back I'll create a new JIRA if needed.

PR #1359, opened 2018-05-25
add appropriate validator for seeing if the years as the result of se…

dsnyder-fairway:
…lected assessment grades exceed the available years for longitudinal aggregate reports

screen shot 2018-05-25 at 1 02 45 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -383,7 +383,7 @@
                     <div *ngIf="showErrors(assessmentGradeRangeControl)">
                       <p *ngIf="assessmentGradeRangeControl.errors.notEmpty" class="help-block small red">
                         {{assessmentGradeRangeControl.errors.notEmpty.messageId | translate}}</p>
-                      <p *ngIf="assessmentGradeRangeControl.errors.withinBounds" class="small red">
+                      <p *ngIf="assessmentGradeRangeControl.errors.withinBounds" class="help-block small red">
                         {{assessmentGradeRangeControl.errors.withinBounds.messageId | translate}}</p>
                     </div>
                   </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -383,6 +383,8 @@
                     <div *ngIf="showErrors(assessmentGradeRangeControl)">
                       <p *ngIf="assessmentGradeRangeControl.errors.notEmpty" class="help-block small red">
                         {{assessmentGradeRangeControl.errors.notEmpty.messageId | translate}}</p>
+                      <p *ngIf="assessmentGradeRangeControl.errors.withinBounds" class="small red">
+                        {{assessmentGradeRangeControl.errors.withinBounds.messageId | translate}}</p>
                     </div>
                   </div>
                 </div>
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
@@ -384,7 +384,7 @@
                       <p *ngIf="assessmentGradeRangeControl.errors.notEmpty" class="help-block small red">
                         {{assessmentGradeRangeControl.errors.notEmpty.messageId | translate}}</p>
                       <p *ngIf="assessmentGradeRangeControl.errors.withinBounds" class="help-block small red">
-                        {{assessmentGradeRangeControl.errors.withinBounds.messageId | translate}}</p>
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -73,9 +73,6 @@ export class AggregateReportTableComponent implements OnInit {
   private _orderingByColumnField: { [ key: string ]: Ordering<AggregateReportItem> } = {};
   private _valueDisplayType: string = ValueDisplayTypes.Percent;
   private _performanceLevelDisplayType: string = PerformanceLevelDisplayTypes.Separate;
-  private _assessmentTypeCode: string;
-  private _cutPoint: number;
-  private _center: boolean;
```
dsnyder-fairway: just cleaning this up

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
```diff
@@ -383,6 +383,8 @@ <h3>{{'aggregate-report-form.section.assessment-heading' | translate}}</h3>
                     <div *ngIf="showErrors(assessmentGradeRangeControl)">
                       <p *ngIf="assessmentGradeRangeControl.errors.notEmpty" class="help-block small red">
                         {{assessmentGradeRangeControl.errors.notEmpty.messageId | translate}}</p>
+                      <p *ngIf="assessmentGradeRangeControl.errors.withinBounds" class="small red">
```
jtreuting: noticing the error above it has a class of `help-block` does that do something that is needed for this one?

webapp/src/main/webapp/src/app/shared/form/validators.ts
```diff
@@ -12,6 +13,20 @@ export function notEmpty(properties: any): ValidatorFn {
   };
 }
 
+export function withinBounds(toSchoolYear: number,
+                             assessmentGrades: string[],
+                             availableSchoolYears: number[],
+                             properties: any): ValidatorFn {
+  return control => {
+    const effectiveSchoolYears = computeEffectiveYears(toSchoolYear, assessmentGrades);
```
jtreuting: does computeEffectiveYears handle the case with HS determining the next year for grade 8 differently than fom grade 8 to grade 7?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -62,6 +62,7 @@
       "assessment-grades-label": "Assessment Grades",
       "assessment-grades-info": "The grade of the assessment that was administered",
       "assessment-grades-empty-error": "Please select an assessment grade.",
+      "assessment-grades-exceed-available-school-years-error": "Range of selected assessment grades exceed number of available academic years.",
```
jtreuting: i really don't know what the text should say here, having a hard time with it.  maybe if we included the actual year in it, so something like "You've selected too many grades.  There are no results prior to 2014-15" within 2014-15 being dynamic based on the avaialble years.

webapp/src/main/webapp/src/app/shared/form/validators.ts
dsnyder-fairway: yes. computeEffectiveYears is used for displaying the Academic years in the report review.  When selecting "HS" and "8" the report review has academic years display "2017-18, 2014-15". When selecting grades "7" and "8" the academic years display "2017-18, 2016-17".

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form.component.html
dsnyder-fairway: I thought this was causing the text to be truncated somehow. Turns out I was wrong. Added it back

webapp/src/main/webapp/src/assets/i18n/en.json

dsnyder-fairway:
I'm going to merge this. If there's any other concerns leave them here or reach out on slack

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -62,7 +62,7 @@
       "assessment-grades-label": "Assessment Grades",
       "assessment-grades-info": "The grade of the assessment that was administered",
       "assessment-grades-empty-error": "Please select an assessment grade.",
-      "assessment-grades-exceed-available-school-years-error": "Range of selected assessment grades exceed number of available academic years.",
+      "assessment-grades-exceed-available-school-years-error": "You've selected too many grades. There are no results prior to {{lowestAvailableSchoolYear}}.",
```
jtreuting: i'm guessing this will display as 2017 and not 2017-18.  there are hepler method to handle that that should be used.

dsnyder-fairway: I use the schoolYear pipe here https://github.com/SmarterApp/RDW_Reporting/pull/1359/commits/5dbfdb250062dde2250f475a20367b3e3b95a111#diff-08bce5334041ddf388f33cd83bcd045b

jtreuting: i missed that pipe on the message.

PR #1361, opened 2018-05-25
Added min student configuration for target report.

agorina:


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportResult.java
```diff
@@ -44,9 +53,18 @@ public V build() {
...
             aggregateReportResult.hasResults = hasResults;
+            aggregateReportResult.insufficientData = insufficientData;
             return result;
         }
 
+        public B merge(final V payload) {
+            createdWhileDataEmbargoed(this.isCreatedWhileDataEmbargoed() || payload.isCreatedWhileDataEmbargoed());
+            hasResults(this.hasResults || payload.isHasResults());
+            insufficientData(this.insufficientData || payload.isInsufficientData());
+            resultCount(this.resultCount + payload.getResultCount());
```
agorina: Moved it from the derived class. It did not seem right that it was there since the data elements are in this class.

I also noticed that it was ignoring `resultCount`. This explains why I am seeing low result counts in production for the queries that take a while to run.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateRowReportResult.java
```diff
@@ -31,9 +31,8 @@ public R build() {
         }
 
         public B merge(final AggregateRowReportResult<T> report) {
+            super.merge(report);
             rows(report.getPayload());
-            createdWhileDataEmbargoed(this.isCreatedWhileDataEmbargoed() || report.isCreatedWhileDataEmbargoed());
-            hasResults(this.isHasResults() || report.isHasResults());
```
agorina: Moved to the base class.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java
```diff
@@ -28,15 +32,47 @@
...
 
     @Autowired
     TargetReportQueryHandler(
             final TargetReportRepository repository,
             final AuthorizationService authorizationService,
             final AggregateReportSettings settings,
             final TargetQueryConverter queryConverter,
-            final TaskExecutor threadPoolTaskExecutor) {
+            final TaskExecutor threadPoolTaskExecutor,
+            @Value("${reporting.target-report.min-number-of-students}") final int minNumberOfStudents) {
```
agorina: Yeah, this is kind of messy...we have `AggregateReportSettings` and this one. But, as I said this morning, I feel that this is a low priority, unless somebody has strong objections.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportStatus.java
```diff
@@ -9,7 +9,8 @@
     COMPLETED(3),
     FAILED(4),
     EXPIRED(5),
-    NO_RESULTS(6);
+    NO_RESULTS(6),
+    INSUFFICIENT_DATA(7);
```
agorina: @wtritch , I have decided to handle it via status instead of metadata. Please let me what you think about it.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandlerTest.java
```diff
@@ -36,14 +43,14 @@
         TargetOrganizationQuery> {
 
     private TargetReportRepository repository;
-    private TargetQueryConverter queryAdapter;
+    private TargetQueryConverter queryConverter;
```
agorina: I just renamed it.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportResult.java
```diff
@@ -44,9 +53,18 @@ public V build() {
...
             aggregateReportResult.resultCount = resultCount;
             aggregateReportResult.hasResults = hasResults;
+            aggregateReportResult.insufficientData = insufficientData;
             return result;
         }
 
+        public B merge(final V payload) {
+            createdWhileDataEmbargoed(this.isCreatedWhileDataEmbargoed() || payload.isCreatedWhileDataEmbargoed());
+            hasResults(this.hasResults || payload.isHasResults());
+            insufficientData(this.insufficientData || payload.isInsufficientData());
```
malaffoon: What's the additive logic here? If i have one payload with sufficient data and add a payload with insufficient data wouldn't the result have sufficient data? I admit i don't really know what a "payload" is here.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java
```diff
@@ -28,15 +32,47 @@
...
+        final TargetReportResult reportResult = super.findByQuery(user, query);
+
+        if (reportResult.getResultCount() == 0) return reportResult;
+
+        for (final TargetRow row : reportResult.getPayload()) {
+            if (row.getDimension().getType() == DimensionType.Overall) {
+
+                if (row.getMeasures().getStudentCount() >= minNumberOfStudents) return reportResult;
+
+                return builder().insufficientData(true).hasResults(false).resultCount(0).build();
```
malaffoon: Just want to verify that you're intentionally not copying the report result because there is nothing in there that would be useful beyond the results?

malaffoon: nit: i know it's personal taste but, with the nested guard-style code this might be slightly more readable using a ternary operator -- it makes it obvious that if we get in here it is returning (and short-circuiting the loop), i.e. 
```
            return row.getMeasures().getStudentCount() >= minNumberOfStudents ? reportResult
                    : builder().insufficientData(true).hasResults(false).resultCount(0).build();
```

aggregate-service/src/main/resources/application.yml
```diff
@@ -85,11 +83,13 @@ jwt:
...
     code: CA
     name: California
-
-  # english-learners should be lep or elas
-  english-learners:
-    - elas
\ No newline at end of file
+  target-report:
+    insufficient-data-cutoff: 0.2
+    min-number-of-students: 0
```
malaffoon: Is this the real default value? If so why did we even bother?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportResult.java
agorina: I was struggling with this too.  I see a few questions here.
1. I think this class should probably be removed/combined with `AggregateRowReportResult`. I  believe the original idea was that we may have different types of the aggregate report results and this is a generic version of it. In reality so far a`payload` is always a collection of AggregateRows.

2. `insufficientData` for the target report means we have at least one row that should not be returned - hence the whole report should not be returned. In this case the additive logic works. But...I see that it may not always be the case. To be honest, this is the easiest solution to the immediate problem.  An alternative I see is to add a report status to the report instead of this flag. Thoughts?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java
agorina: That is correct. Thank you for checking.

aggregate-service/src/main/resources/application.yml
agorina: I believe that idea is that they want to be able to change it. @jtreuting , please help me here.

jtreuting: some states will always want to show results which maybe means the min is 1 and not 0?  so some states will want to show with even a single student while others will be more conservative and block it at 10 or something.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/stream/AggregateReports.java
```diff
@@ -72,7 +73,7 @@ public void handleAggregateRequest(final Message<AggregateRequestMessage> messag
             final AggregateReportResult<?> report = handlerProvider.findHandler(payload.getAggregateQuery()).findByQuery(user, payload.getAggregateQuery());
 
             if (!report.isHasResults()) {
```
wtritch: So for `report.isInsufficientData()` to be true, `report.isHasResults()` *must* be true?  Hidden dependencies like that make me nervous...

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportStatus.java
wtritch: Status sounds reasonable to me.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportResult.java
jtreuting: wait @agorina are we overloading the term Insufficient Data?  I think i'm confused because the UI has a concept of Insufficient Data for a single target row which means that the standard error > 0.2.  but this is for the minimum N right?  maybe we need different terms for those 2 concepts.

@wtritch let's not worry about that right now, we'll discuss more offline.  i think we will be fine, i might have jumped the gun on the warning/heads-up I gave you.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ReportStatus.java
wtritch: ...unless we need to return target-level reports across multiple assessments and some are insufficient and some aren't... but I guess we'll discuss that on Tuesday, because that would require row-level insufficient data results.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AggregateReportResult.java
agorina: @jtreuting , you are right about overloading the term. Let me see if I can make it cleaner.

PR #1363, opened 2018-05-25
more spacing for advanced filters

dsnyder-fairway:
screen shot 2018-05-25 at 3 42 07 pm


dsnyder-fairway:
@jtreuting is this mergeable? I would have made the student section appear with more columns, but I am not sure which elements can be moved to their own column

jtreuting:
sorry didn't realize you were waiting on that.  good point.  i think we can move migrant status, 504 and IEP into the 2nd column and have 4 instead of 3 columns.

dsnyder-fairway:
Here is the result
screen shot 2018-05-30 at 10 51 49 am

Merging

PR #1364, opened 2018-05-26
added validation that assessment grades has more than one selected fo…

dsnyder-fairway:
…r longitudinal reports

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/validators.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/validators.ts
@@ -13,6 +13,27 @@ export function notEmpty(properties: any): ValidatorFn {
   };
 }
 
+/**
+ * Form control validator that makes sure the control value has a length greater than one
+ *
+ * @param properties the properties to propagate when the control value is invalid
+ * @return {ValidatorFn}
+ */
+export function isGreaterThanOne(properties: any): ValidatorFn {
+  return control => {
+    return control.value.length > 1 ? null : { isGreaterThanOne: properties };
+  };
+}
+
+/**
+ * Form control validator that makes sure the computed effective years does not go below the lowest available school year
+ *
+ * @param {number} toSchoolYear a school year
+ * @param {string[]} assessmentGrades a list of selected assessment grades
+ * @param {number} lowestAvailableSchoolYear the lowest school year we allow in the application
+ * @param properties the properties to propagate when the control value is invalid
+ * @returns {ValidatorFn}
+ */
 export function withinBounds(toSchoolYear: number,
                              assessmentGrades: string[],
                              lowestAvailableSchoolYear: number,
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/shared/form/validators.ts
```diff
@@ -13,6 +13,27 @@ export function notEmpty(properties: any): ValidatorFn {
...
+
+/**
+ * Form control validator that makes sure the computed effective years does not go below the lowest available school year
+ *
+ * @param {number} toSchoolYear a school year
+ * @param {string[]} assessmentGrades a list of selected assessment grades
+ * @param {number} lowestAvailableSchoolYear the lowest school year we allow in the application
+ * @param properties the properties to propagate when the control value is invalid
+ * @returns {ValidatorFn}
+ */
```
dsnyder-fairway: forgot some documentation

webapp/src/main/webapp/src/app/shared/form/validators.ts
```diff
@@ -13,6 +13,27 @@ export function notEmpty(properties: any): ValidatorFn {
   };
 }
 
+/**
+ * Form control validator that makes sure the control value has a length greater than one
+ *
+ * @param properties the properties to propagate when the control value is invalid
+ * @return {ValidatorFn}
+ */
+export function isGreaterThanOne(properties: any): ValidatorFn {
```
esotrope: i think we might be able to use `Validators.minlength(1)` though it may be intended for only strings and not just objects with a length prop

dsnyder-fairway: It looks like going that route I would need to still have a notEmpty validator. Also I am not sure how to get that message into the the notificationService when we call `AggregateReportFormComponent.validate`

dsnyder-fairway: also I think you meant `Validators.minlength(2)`. We need two years selected

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -63,7 +63,7 @@
       "assessment-grades-info": "The grade of the assessment that was administered",
       "assessment-grades-empty-error": "Please select an assessment grade.",
       "assessment-grades-exceed-available-school-years-error": "You've selected too many grades. There are no results prior to {{lowestAvailableSchoolYear}}.",
-      "assessment-grades-greater-than-one-error": "Please select more than one assessment grade.",
+      "assessment-grades-less-than-minimum-error": "Please select more than one assessment grade.",
```
dsnyder-fairway: the key should reflect the error we have encountered

webapp/src/main/webapp/src/app/shared/form/validators.ts
esotrope: gotcha - good point. this approach is good then. you could make our validator more flexible though to behave more like minlength() and accept a variable minimum length

PR #1365, opened 2018-05-26
Refactored the way we retrieve students ethnicity.

agorina:
Why?  There are two ways we get students ethnicity: in one repository we used `GROUP_CONCAT` and in another we looked up the ethnicity as a second database call. I need a similar logic for the target report API and I wanted to re-use `GROUP_CONCAT` solution. When I tested it on AWS Dev, it was **very** slow ( I had to kill the query since it was not responding for a while). The main offender is `GROUP BY`. So I changed it to do a second call.

### DIFF
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractExamRepository.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractExamRepository.java
@@ -1,6 +1,5 @@
 package org.opentestsystem.rdw.reporting.exam;
 
-import com.google.common.collect.ImmutableMap;
 import org.opentestsystem.rdw.reporting.common.jdbc.Exams;
 import org.opentestsystem.rdw.reporting.common.jdbc.SecurityParameterProvider;
 import org.opentestsystem.rdw.reporting.common.jdbc.StudentContexts;
@@ -8,20 +7,18 @@ import org.opentestsystem.rdw.reporting.common.model.Exam;
 import org.opentestsystem.rdw.reporting.common.model.Student;
 import org.opentestsystem.rdw.reporting.common.model.StudentContext;
 import org.opentestsystem.rdw.reporting.common.security.PermissionSource;
-import org.springframework.beans.factory.annotation.Value;
+import org.opentestsystem.rdw.reporting.student.StudentEthnicityRepository;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 import javax.validation.constraints.NotNull;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 import static com.google.common.collect.Lists.newArrayList;
-import static com.google.common.collect.Maps.newHashMap;
 import static com.google.common.collect.Sets.newHashSet;
 
 public abstract class AbstractExamRepository<T extends AbstractExamSearch> implements ExamRepository<T> {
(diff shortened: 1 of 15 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/AbstractExamRepository.java
```diff
@@ -94,25 +91,5 @@ protected AbstractExamRepository(
         return exams;
     }
 
-    private Map<Long, Set<String>> findEthnicityCodesForStudentIds(final Collection<Long> studentIds) {
```
agorina: Moved it into a separate repository to be reused.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
```diff
@@ -9,51 +9,71 @@
...
 
     @Override
     public List<Student> findAll(@NotNull final PermissionSource permissionSource,
                                  @NotNull final StudentSearch search) {
-        return template.query(
+
+        final List<Student.Builder> studentBuilders = newArrayList();
+        final Set<Long> studentIds = newHashSet();
+
+        // SQL returns a sorted list, need to preserve the order
```
agorina: @esotrope - thank you for writing a test for this!

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/Student.java
```diff
@@ -128,6 +128,10 @@ public Builder copy() {
         private Boolean section504;
         private Boolean migrantStatus;
 
+        public long getId() {
+            return id;
+        }
```
agorina: I know, but..I need it.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/TestData.java
```diff
@@ -153,7 +153,7 @@
                 .firstName("student1_firstName")
                 .lastName("student1_lastName")
                 .genderCode("g1")
-                .ethnicityCodes(ImmutableSet.of());
+                .ethnicityCodes(ImmutableSet.of("ethnicity-29", "ethnicity-28"));
```
agorina: I added test data.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
```diff
@@ -9,51 +9,71 @@
...
+                            .limitedEnglishProficiency(getNullable(row, row.getBoolean("lep")))
+                            .section504(getNullable(row, row.getBoolean("section504")))
+                            .migrantStatus(getNullable(row, row.getBoolean("migrant_status"))));
+                    return row;
+                });
 
+        final Map<Long, Set<String>> ethnicityCodesByStudentId = ethnicityRepository.findEthnicityCodesByStudentIds(studentIds);
+        return studentBuilders
+                .stream()
+                .map(b -> b.ethnicityCodes(ethnicityCodesByStudentId.getOrDefault(b.getId(), null)).build())
```
esotrope: doesn't Map.get() return null when the key is absent?

agorina: Good point, changing.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentEthnicityRepository.java
```diff
@@ -0,0 +1,50 @@
...
+                row -> {
+                    final long studentId = row.getLong("student_id");
+                    final String ethnicityCode = row.getString("ethnicity_code");
+
+                    final Set<String> studentEthnicityCodes = ethnicityCodesByStudentId.get(studentId);
+                    if (studentEthnicityCodes == null) {
+                        ethnicityCodesByStudentId.put(studentId, newHashSet(ethnicityCode));
+                    } else {
+                        studentEthnicityCodes.add(ethnicityCode);
+                    }
```
malaffoon: nit: i think the computeIfAbsent approach is more readable, i.e.:
```
        final Set<String> studentEthnicityCodes = ethnicityCodesByStudentId.computeIfAbsent(studentId, newHashSet());
        studentEthnicityCodes.add(ethnicityCode);
```

PR #1367, opened 2018-05-28
Target report for teachers API - repository and service

agorina:
This PR depends on the schema changes: https://github.com/SmarterApp/RDW_Schema/pull/277


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/GroupExamWithTargetScoreService.java
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * Gets all {@link ExamWithTargetScore} grouped by {@link Target} for the given {@link GroupExamSearch}
+     *
+     * @param user   the user accessing the group information
+     * @param search the search parameters to limit the results by
+     * @return all {@link ExamWithTargetScore} grouped by {@link Target}; includes all {@link Target}s for
+     * the given {@link GroupExamSearch#assessmentId} regardless of the presents of the target's scores.
+     */
+    Map<Target, Set<ExamWithTargetScore>> getExams(@NotNull User user, @NotNull GroupExamSearch search);
+}
```
agorina: @jtreuting , the controller will just call this. Please let me know if this is good enough for the UI.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -127,6 +127,26 @@ sql:
...
+          JOIN student st on st.id = e.student_id
+          JOIN asmt a on e.asmt_id = a.id
+          JOIN school s on e.school_id = s.id
+          JOIN school isch on st.inferred_school_id = isch.id
+          JOIN exam_target_score ets on ets.exam_id = e.id
+          JOIN target t on t.id = ets.target_id
+         WHERE a.id = :assessment_id
+            AND e.school_year = :school_year
+            AND ${sql.reporting.snippet.groupFilters}
+            AND ${sql.reporting.snippet.studentExamPermissionWithEmbargo}
```
agorina: I am not filtering targets without scores. Since there are two different scores, I think the UI will need to handle it.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/TestData.java
```diff
@@ -424,36 +424,4 @@
                 .claimScaleScores(newArrayList(null, null, null))
                 .school(SCHOOLS.get(2));
     }
-
-    public static Exam.Builder exam20() {
```
agorina: This was not used.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/assessment/JdbcAssessmentItemRepositoryIT.java
```diff
@@ -24,6 +24,69 @@
 @ActiveProfiles("test")
 public class JdbcAssessmentItemRepositoryIT {
 
+    private static final AssessmentItem item1 = AssessmentItem.builder()
```
agorina: Unrelated, OCD.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/DefaultGroupExamWithTargetScoresService.java
```diff
@@ -0,0 +1,45 @@
...
+        final Map<Integer, Target> targetsById = assessmentTargetRepository.
+                findAllByAssessmentId(search.getAssessmentId())
+                .stream()
+                .collect(toMap(Target::getId, t -> t));
+
+        for (final ExamWithTargetScore exam : examWithTargetScoreRepository.findAllByGroupExamSearch(user, search)) {
+            examsByTarget.computeIfAbsent(targetsById.get(exam.getTargetId()), id -> newHashSet()).add(exam);
+        }
+
+        return examsByTarget;
```
agorina: I realized that I forgot to consider `min-number-of-students` here. 

Not sure what is the best way to handle it. 
Should the controller return a different HTTP status? 
Should the UI handle it instead?

malaffoon: Good point about filtering -- that seems to be a definitive argument for doing it on the front-end..

So, do we have to worry about clever users hijacking the payload?

agorina: Is the `min-number-of-students` related to the security?  Or is it just because the data is meaningless?  ( @jtreuting)

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/DefaultGroupExamWithTargetScoresService.java
```diff
@@ -0,0 +1,45 @@
...
+    private final AssessmentTargetRepository assessmentTargetRepository;
+
+    @Autowired
+    DefaultGroupExamWithTargetScoresService(final GroupExamWithTargetScoreRepository examWithTargetScoreRepository,
+                                            final AssessmentTargetRepository assessmentTargetRepository) {
+        this.examWithTargetScoreRepository = examWithTargetScoreRepository;
+        this.assessmentTargetRepository = assessmentTargetRepository;
+    }
+
+    public Map<Target, Set<ExamWithTargetScore>> getExams(@NotNull final User user,
```
agorina: I think I may need to break it into two calls. It looks like it cannot be serialized as JSON, and I see that we have two Controller calls for similar cases.

In this case one call will get all `Target`s and the other all `ExamWithTargetScore`. the front end will have to do what I was doing here. 

Does it sound right?

PR #1369, opened 2018-05-28
new endpoint for user data to aid in diagnostics

dsnyder-fairway:


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/organization/DefaultOrganizationService.java
```diff
@@ -33,6 +34,11 @@ public DefaultOrganizationService(@NotNull final OrganizationRepository reposito
         this.repository = repository;
     }
 
+    @Override
+    public List<School> getSchools(@NotNull final User user) {
+        return getSchools(user, null);
```
esotrope: i would be tempted to pass in just an empty filter here. my guess is that that has a default org limit though?

dsnyder-fairway: yeah we set the limit to Unlimited, then we copy the query if it is available to set the limit again

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/userdetails/DefaultUserDetailsService.java
```diff
@@ -0,0 +1,35 @@
...
+        this.groupService = groupService;
+        this.organizationService = organizationService;
+    }
+
+    @Override
+    public UserDetails getUserDetails(@NotNull final User user) {
+        return UserDetails.builder()
+                .firstName(user.getFirstName())
+                .lastName(user.getLastName())
+                .permissions(user.getPermissionsById())
```
esotrope: my only nit would be that i think we could return this:
```java
.permissions(user.getPermissionsById().keySet())
.permissionsById(user.getPermissionsById())
```
that way we keep the flat permissions in this endpoint as well and add permissions by id.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/userdetails/UserDetailsService.java
```diff
@@ -0,0 +1,11 @@
+package org.opentestsystem.rdw.reporting.userdetails;
+
+import org.opentestsystem.rdw.reporting.common.web.security.User;
+
+import javax.validation.constraints.NotNull;
+
+interface UserDetailsService {
+
+    UserDetails getUserDetails(@NotNull User user);
```
esotrope: + javadocs

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/userdetails/UserDetails.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.opentestsystem.rdw.reporting.common.model.School;
+import org.opentestsystem.rdw.reporting.group.Group;
+import org.opentestsystem.rdw.security.Permission;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+class UserDetails {
```
malaffoon: Hmm, not thrilled with this name. `UserDetails` is a core spring security interface and it is confusing to see it here.

PR #1370, opened 2018-05-29
Feature/dwr 1695 ica writing traits

jtreuting:
New ICA and Summative design with writing trait scores

image

image


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
```diff
@@ -98,6 +100,36 @@ public ScaleScoreView getScaleScoreView(final Exam exam, final Assessment assess
         return new ScaleScoreView(position, ImmutableList.copyOf(levels));
     }
 
+    /**
+     * For ICA and Summative the Writing Claim must be last for the PDF Report so that Writing Trait Scores
+     * are displayed next to the Writing Claim.
+     *
+     * @return Comparator used for sorting claim codes in the ICA and Summative PDF report template
+     */
+    public Comparator<String> getClaimCodeComparator() {
```
jtreuting: used in the thymeleaf template to sort the writing claim last so the writing trait scores can be displayed next to them on their own row.

report-processor/src/main/resources/application.sql.yml
```diff
@@ -65,8 +65,23 @@ sql:
       SELECT
         d.name AS district_name,
         ${sql.snippet.claimColumns},
+        wt.trait_evidence_elaboration_score as ei_trait_evidence_elaboration_score,
+        wt.trait_organization_purpose_score as ei_trait_organization_purpose_score,
+        wt.trait_conventions_score as ei_trait_conventions_score,
+        wt.performance_task_writing_type,
         ${sql.snippet.exam}
       JOIN district d ON s.district_id=d.id
+      LEFT JOIN (
```
jtreuting: for each ICA/Summative for a student i only want to get the WER if it exists.  this is the best way I could think of handling it.  @agorina any thoughts?

report-processor/src/main/resources/templates/batch-reports.html
```diff
@@ -3,7 +3,9 @@
     <head>
         <div th:replace="fragments::meta"></div>
         <title th:text="${request.name}"></title>
-        <div th:replace="fragments::styles"></div>
+        <link rel='stylesheet prefetch' href='https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.css'>
```
jtreuting: the new template used some font-awesome stuff that isn't in the version that is embedded currently in the report-processor styles for the reports.  so I removed the font-awesome from that embedded version and used this as this is what Aaron provided when creating the HTML templates.  @esotrope any issue with this that you can think of?

report-processor/src/main/resources/templates/fragments.html
```diff
@@ -8,74 +8,404 @@
...
....nav-tabs.nav-justified>li{display:table-cell;width:1%}.nav-tabs.nav-justified>li>a{margin-bottom:0;border-bottom:1px solid #ddd;border-radius:5px 5px 0 0}.nav-tabs.nav-justified>.active>a,.nav-tabs.nav-justified>.active>a:focus,.nav-tabs.nav-justified>.active>a:hover{border-bottom-color:#FFF}}.nav-pills>li{float:left}.nav-justified>li,.nav-stacked>li{float:none}.nav-pills>li>a{border-radius:5px}.nav-pills>li+li{margin-left:2px}.nav-pills>li.active>a,.nav-pills>li.active>a:focus,.nav-pills>li.active>a:hover{color:#fff;background-color:#3696E6}.nav-stacked>li+li{margin-top:2px;margin-left:0}.nav-justified{width:100%}.nav-justified>li>a{text-align:center;margin-bottom:5px}.nav-tabs-justified{border-bottom:0}.nav-tabs-justified>li>a{margin-right:0;border-radius:5px}.nav-tabs-justified>.active>a,.nav-tabs-justified>.active>a:focus,.nav-tabs-justified>.active>a:hover{border:1px solid #ddd}@media (min-width:768px){.nav-justified>li{display:table-cell;width:1%}.nav-justified>li>a{margin-bottom:0}.nav-tabs-justified>li>a{border-bottom:1px solid #ddd;border-radius:5px 5px 0 0}.nav-tabs-justified>.active>a,.nav-tabs-justified>.active>a:focus,.nav-tabs-justified>.active>a:hover{border-bottom-
```
jtreuting: separated into 2 styles, libraries and the specific CSS for these reports

jtreuting: this block is the new styles from the HTML provided for the new reports by Aaron

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupportTest.java
```diff
@@ -124,6 +129,14 @@ public void getScaleScoreViewShouldReturnCorrectLevelsForHighStdErrIAB() throws
                 );
     }
 
+    @Test
+    public void getClaimCodeComparatorShouldSortWritingLast() {
```
jtreuting: show that this comparator only moves the Writing claim to the end and the others stay in the same order they started in

webapp/src/main/webapp/src/styles.less
```diff
@@ -945,6 +945,41 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
   .generate-color-classes(@colorName, (@i + 1));
 }
 
+/**
+ * ICA/Summative PDF Report
+ */
+.overview .standard .score.maroon,
```
jtreuting: i added these into the main styles since I believe the report CSS was originally created from these and then pasted to the report fragments (correct me if i'm wrong @esotrope ), so I added them here even though I also added them to the report styles.  maybe there is a different/better way

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
esotrope: this has interesting implications for configurable subjects doesn't it

report-processor/src/main/resources/templates/batch-reports.html
esotrope: Only issues i can think of are those that come along with using a CDN. If we/sbac are cool with those then i see no problem

report-processor/src/main/resources/templates/fragments.html
esotrope: not first-child? interesting

esotrope:
some things i noticed in the pics:
1. `report.faq.p3.header` is not translated
2. text at the bottom doesn't align well with the well columns above it
3. the writing trait well seems to extend too far to the right

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
wtritch: hm... in `warehouse.subject_claim_score` we have a `display_order` column we can use to sort on, we'll just need to make sure it gets migrated =)

jtreuting: we could use that as long as we always want Writing to be the last shown claim in the other contexts as well (whatever those are).  It felt somewhat PDF specific as the need was to place the writing trait scores next to them and that is how it fit.  but if there is a display order that would probably be fine.

report-processor/src/main/resources/templates/sum-body.html
```diff
@@ -4,112 +4,159 @@
...
+    <div class="header container">
+        <div class="block-group blue right">
+            <span th:text="#{'report.subject.' + ${report.assessment.subject} + '.name'}"></span>
+            <span>
+                <span th:text="${#strings.toString(report.exam.schoolYear - 1)}" th:remove="tag"></span> -
+                <span th:text="${support.getLastTwoDigits(report.exam.schoolYear)}" th:remove="tag"></span>
+            </span>
+
+        </div>
+        <img src="https://s3-us-west-2.amazonaws.com/s.cdpn.io/1501143/sbac%20logo%20color.svg" alt="Smarter Balanced Assessment Consortium" class="logo">
```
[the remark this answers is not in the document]
jtreuting: good point, i'm guessing it was where Aaron put it maybe just for the mockups.  we'll need to put that somewhere else i think.  i'll do that.

PR #1372, opened 2018-05-29
Feature/target report query ui

wtritch:
This PR adds a new target report query route and component.
I've modified the sample aggregate report data generator to be a little more declarative.
I've modified the aggregate report table display to also define the entire column set based upon the report type rather than conditionally switch individual columns on and off.

Note that the feature is not yet complete, the "Generate Report" button doesn't currently execute any method, but the PR kept getting larger and this is a good place to get some feedback.  Also, merging at this point shouldn't affect existing aggregate report behavior.

target_report_2


webapp/src/main/webapp/package.json
```diff
@@ -38,6 +38,7 @@
     "json-unflat": "^1.0.1",
     "lodash": "^4.17.4",
     "mathjs": "^3.15.0",
+    "ng-inline-svg": "6.2.1",
```
wtritch: This is the library I'm using to load SVG files as icons so that I can still style/color them using CSS.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.ts
```diff
@@ -96,7 +96,7 @@ export interface AggregateReportFormSettings {
   /**
    * Defines the report type (standard or longitudinal)
    */
-  reportType: 'GeneralPopulation' | 'LongitudinalCohort' | 'Claim';
+  reportType: 'GeneralPopulation' | 'LongitudinalCohort' | 'Claim' | 'Target';
```
wtritch: I added 'Target' as a ReportType with all of the fiddly-bits that that involves.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
```diff
@@ -197,6 +197,11 @@ export class AggregateReportOptionsMapper {
         migrantStatuses: options.studentFilters.migrantStatuses,
         section504s: options.studentFilters.section504s
       },
+      targetReport: {
+        assessmentGrade: options.assessmentGrades[ 0 ],
+        schoolYear: options.schoolYears[ 0 ],
+        subjectCode: options.subjects[ 0 ]
+      },
```
wtritch: Since the assessment grade, school year, and subject selectors are all radio buttons in the target form, I'm just defaulting them to their "first" option which simplifies a lot of the query form logic. (We always have enough information to generate sample data.)

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -189,10 +216,6 @@ export class AggregateReportSummary {
         label: translate('aggregate-report-form.field.assessment-type-label'),
         values: [ translate(`common.assessment-type.${settings.assessmentType}.short-name`) ]
       },
-      {
-        label: translate('aggregate-report-form.field.subjects-label'),
-        values: orAll(options.subjects, settings.subjects, code => translate(`common.subject.${code}.short-name`))
-      },
```
wtritch: I had to break this out of the shared "common" rows because a target report query contains a single subject at `settings.targetReport.subjectCode` rather than a collection of subjects at `settings.subjects`

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -24,154 +24,230 @@ const MaximumOrganizations = 3;
...
+      }];
+    }
+  };
+
   constructor(private translate: TranslateService,
               private subgroupMapper: SubgroupMapper,
               private reportService: AggregateReportService) {
   }
 
   createSampleData(assessmentDefinition: AssessmentDefinition,
```
wtritch: I modified this method from a single generation method with a bunch of *if* statements to a collection of providers that supply values for AggregateReportItem properties.
Example: three providers, one that gives two values for `organization` and one that gives three values for `subgroup`, and one that provides a single value for `performanceLevelByDisplayTypes` will result in 2*3*1=6 rows, the full combinatorial set of provided values.
This let me define the generation of sample target rows without a giant copy-paste if-else clause.

webapp/src/main/webapp/src/app/aggregate-report/assessment/assessment-definition.service.ts
```diff
@@ -71,6 +71,17 @@ const ClaimSummative: AssessmentDefinition = {
...
+const TargetSummative: AssessmentDefinition = {
+  typeCode: 'sum',
+  interim: false,
+  performanceLevels: [1, 2, 3],
+  performanceLevelCount: 3,
+  performanceLevelDisplayTypes: [ PerformanceLevelDisplayTypes.Separate ],
+  aggregateReportIdentityColumns: ['claim', 'target', 'dimension'],
+  aggregateReportStateResultsEnabled: false,
+  aggregateReportTypes: [ 'Target' ]
+};
```
wtritch: Right now all Target reports are Summative reports, so I only needed a single AssessmentDefinition in here.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.ts
```diff
@@ -31,5 +31,8 @@ export class AggregateReportItem {
   };
   organization: Organization;
   subgroup: Subgroup;
+  studentRelativeResidualScoresLevel?: 'Above' | 'Near' | 'Below' | 'InsufficientData' | 'Excluded';
+  standardMetRelativeResidualLevel?: 'Above' | 'Near' | 'Below' | 'InsufficientData' | 'Excluded';
+  targetNaturalId?: string;
```
wtritch: new optional properties driven by the java `TargetRow` class

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -12,7 +12,7 @@
          [sortMode]="rowSortingEnabled && 'single'"
          [customSort]="true"
          [autoLayout]="true"
-         styleClass="table table-striped table-hover overflow table-row-size-fixed">
+         styleClass="table table-striped table-hover overflow">
```
wtritch: Now that we allow word-wrapping on our very long target descriptions, I had to play with the CSS a bit to keep our tables looking nice in both existing aggregate reports and the target report.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -84,7 +84,22 @@
...
+
+          <div *ngSwitchCase="'target'"
+               class="text-right wrapping-max-width">
+            <span *ngIf="!preview; else targetPreview"
+                  [innerHTML]="'aggregate-report-table.target-description' | translate: getTargetDisplay(rowData)">
+            </span>
+            <ng-template #targetPreview>
+              <span [innerHTML]="'aggregate-report-table.sample.target' | translate:{name: rowData.targetNaturalId}"></span>
+            </ng-template>
           </div>
```
wtritch: Since we want `<strong>` tags around the target name to separate it from the description, I had to use the innerHTML hack here =/

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -128,6 +143,57 @@
...
+                    class="sb-iab-red">
+                <sb-icon class="fa-lg" icon="sbac-below-average"></sb-icon>
+              </span>
+              <span *ngSwitchCase="'InsufficientData'"
+                    class="gray-light">
+                <sb-icon class="fa-lg" icon="sbac-insufficient-data"></sb-icon>
+              </span>
+            </span>
+            {{'aggregate-report-table.target.overall.' + rowData.studentRelativeResidualScoresLevel | translate}}
+          </div>
```
wtritch: I could move this
```
'Above' => ('sb-iab-green', 'sbac-above-average')
```
logic out of the template and into the component... but I'm not sure it'd be cleaner.  Thoughts?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -128,6 +143,57 @@
...
             </span>
           </div>
+
+          <div *ngSwitchCase="'studentRelativeResidualScoresLevel'">
+
+            <!-- pick the icon -->
+            <span [ngSwitch]="rowData.studentRelativeResidualScoresLevel">
+              <span *ngSwitchCase="'Above'"
+                    class="sb-iab-green">
+                <sb-icon class="fa-lg" icon="sbac-above-average"></sb-icon>
```
wtritch: I created an `<sb-icon>` component that can load style-able .svg icons from `/assets/svgs/{{icon}}.svg`  They'll respect CSS-controlled color, size, svg effects like blurring, etc.

agorina:
I see that this assumes multiple subjects/grades/school years. Do we have a decision on this?  (as you know, the back-end currently assumes/supports only one of each).

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -37,6 +37,9 @@ const AssessmentLabelOrdering: Ordering<AggregateReportItem> = ordering(byString
 const ClaimOrdering: Ordering<AggregateReportItem> = ordering(byString)
   .on(item => item.claimCode);
 
+const TargetOrdering: Ordering<AggregateReportItem> = ordering(byString)
+  .on(item => item.targetNaturalId);
```
wtritch: This will have to change... We could have a target with natural id, names (AAA|X, X) and (BBB|A, A) but we'll want to sort them based upon their translated names...

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -287,6 +288,14 @@ export const routes: Routes = [
...
             component: AggregateReportFormComponent
           },
+          {
+            path: 'targets',
+            pathMatch: 'full',
+            resolve: {
+              settings: AggregateReportFormSettingsResolve
+            },
+            component: TargetReportFormComponent
+          },
```
wtritch: This intercepts and routes the user to the target form before attempting to match 'targets' as a numeric id in the child route below: `:id`

wtritch:
@agorina I'm not sure what you mean?  I'm currently assuming a single subject/grade/school year?

agorina:
@wtritch , the way UI looks I assumed that multiple selections are allowed. Sorry about it.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -24,154 +24,230 @@ const MaximumOrganizations = 3;
...
 
-  getClaimCodeTranslation(row: AggregateReportItem): string {
-    return this.translate.instant(this.getClaimCodeTranslationKey(row));
+    const valueProvider: ValueProvider = valueProviders.shift();
+    for (let values of valueProvider.getValues(context)) {
+      row = Object.assign({}, row, values);
+      context.row = row;
+      items = items.concat(this.createRows(row, valueProviders, context));
+    }
+    valueProviders.unshift(valueProvider);
```
esotrope: why do we need to shift and unshift? seems complicated

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
esotrope: yeah in cases like these i would be tempted to make view models that just hold the icon code. the question is do we want the complexity in the view template or the js code. i generally tend toward the js code but i'm fine either way

webapp/src/main/webapp/src/app/aggregate-report/target/target-report-form.component.ts
```diff
@@ -0,0 +1,264 @@
...
+import { fileName } from "../../shared/form/validators";
+import { Forms } from "../../shared/form/forms";
+import { AssessmentDefinition } from "../assessment/assessment-definition";
+import { Utils } from "../../shared/support/support";
+
+@Component({
+  selector: 'target-report-form',
+  templateUrl: './target-report-form.component.html'
+})
+export class TargetReportFormComponent implements OnInit {
```
esotrope: having all this duplicated form code is a bit frightening. we will really have to make sure changes to one happen to the other.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
wtritch: Because we go up and down the tree to cover all the combinations:
```
org a -> subgroup a -> grade a
         -> subgroup a -> grade b
         -> subgroup b -> grade a
         -> subgroup b -> grade b
org b -> subgroup a -> grade a
etc
```

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -84,7 +84,22 @@
 
           <div *ngSwitchCase="'claim'"
                class="text-right">
-            {{ getClaimCodeTranslation(rowData) }}
+            <span *ngIf="!preview || table.reportType !== 'Target'; else targetClaimPreview">
+              {{ getClaimCodeTranslation(rowData) }}
+            </span>
+            <ng-template #targetClaimPreview>
+              {{'aggregate-report-table.sample.claim' | translate:{id: rowData.claimCode} }}
```
esotrope: odd that we have to do this property mapping dance. if anything i would have expected `{code: rowData.claimCode}`

webapp/src/main/webapp/src/app/aggregate-report/target/target-report-form.component.ts
wtritch: Yeah... The alternative was even more nested "ifs" with large copy-pasted sections sprinkled throughout the existing controller... =/

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
wtritch: good point, changed from "id" to "code"

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -24,154 +24,230 @@ const MaximumOrganizations = 3;
...
+              delete bothDisplayTypes.performanceLevelByDisplayTypes.Grouped;
+              return [ bothDisplayTypes ];
+            }
+          }, {
+            getValues: (context) => [ {subjectCode: context.settings.subjects[0]} ]
+          }, {
+            getValues: (context) => context.settings.claimReport.assessmentGrades
+              .map(grade => {
+                return { assessmentGradeCode: grade }
+              })
```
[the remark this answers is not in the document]
wtritch: ah, I like that a lot better.  Changed.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -24,154 +24,230 @@ const MaximumOrganizations = 3;
...
+            }
+          });
+        break;
+
+      case 'Claim':
+        valueProviders.push(
+          this.defaultOrganizationProvider,
+          this.defaultSubgroupProvider, {
+            getValues: (context) => {
+              //Strip out grouped performance level results
```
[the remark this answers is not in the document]
wtritch: hmmm no it doesn't  wonder if I need to update my lint settings...

jtreuting:
@wtritch is there a reason the icons for Insufficient data aren't the same in the 2 columns?

wtritch:
Not really except that one seems to use badging and the other used labelled icons... It'll be inconsistent either way.  Do you want to use an icon for insufficient data in the right column so it doesn't match the other values in the column OR do you want to use an Insufficient Data badge in the right column so it doesn't match the other Insufficient Data display?

jtreuting:
@wtritch I would vote for using the same icon for insufficient data in both columns since they mean the same thing in both cases.  @aaronsleeper have a strong opinion on this?

aaronsleeper:
@wtritch @jtreuting, yes, it was my intention to use the same icon in both columns

PR #1373, opened 2018-05-29
fix issue where table would be re-drawn on student history table clic…

dsnyder-fairway:
…k events

issue was that the columns would be redrawn on any table click event. It is only necessary to redraw the columns when we are able to toggle to Claims


dsnyder-fairway:
Gonna merge this since this has been a big issue for a while. I'll address anything that comes up in future PRs

PR #1375, opened 2018-05-30
Modify the target report form component to submit target report queries.

wtritch:
Update styles to re-introduce bottom padding on cell values.

I also modified the column headers to allow word-wrapping on titles that are significantly wider than their content.

target_report_4

aggregate_report

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
@@ -419,13 +419,15 @@ export class AggregateReportRequestMapper {
   private ServerReportTypeByClientType = {
     GeneralPopulation: 'CustomAggregate',
     LongitudinalCohort: 'Longitudinal',
-    Claim: 'Claim'
+    Claim: 'Claim',
+    Target: 'Target'
   };
 
   private ClientReportTypeByServerType = {
     CustomAggregate: 'GeneralPopulation',
     Longitudinal: 'LongitudinalCohort',
-    Claim: 'Claim'
+    Claim: 'Claim',
+    Target: 'Target'
   };
 
   private toServerReportType(type: string) {
--- a/webapp/src/main/webapp/src/app/aggregate-report/target/target-report-form.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/target/target-report-form.component.ts
@@ -1,4 +1,4 @@
-import { Component, Inject, OnInit, ViewChild } from "@angular/core";
+import { Component, Inject, ViewChild } from "@angular/core";
 import { AggregateReportFormSettings } from "../aggregate-report-form-settings";
 import { AggregateReportFormOptions } from "../aggregate-report-form-options";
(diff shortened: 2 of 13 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -419,13 +419,15 @@ export class AggregateReportRequestMapper {
...
+    Claim: 'Claim',
+    Target: 'Target'
   };
 
   private ClientReportTypeByServerType = {
     CustomAggregate: 'GeneralPopulation',
     Longitudinal: 'LongitudinalCohort',
-    Claim: 'Claim'
+    Claim: 'Claim',
+    Target: 'Target'
```
wtritch: During the submission I realized there were yet more sets of "all report types" to modify.

webapp/src/main/webapp/src/app/aggregate-report/target/target-report-form.component.ts
```diff
@@ -53,7 +53,14 @@ export class TargetReportFormComponent implements OnInit {
...
   @ViewChild('organizationTypeahead')
-  organizationTypeahead: OrganizationTypeahead;
+  set organizationTypeahead(value: OrganizationTypeahead) {
+    this._organizationTypeahead = value;
+    if (!Utils.isNullOrUndefined(this.organization)) {
+      setTimeout(() => {
+        this._organizationTypeahead.value = this.organization.name;
+      });
+    }
+  }
```
wtritch: I discovered that the organizationTypeahead ViewChild isn't initialized at OnInit time due to a parent `*ngIf`  This initializes the value once the child component is initialized and a reference is injected.

webapp/src/main/webapp/src/styles.less
```diff
@@ -563,13 +563,18 @@ body .select-menu-group {
...
   &.pt-0 {
     padding-top: 0;
   }
+  &.pb-0 {
+    padding-bottom: 0;
+  }
+
+  .scale-score {
+    margin: -.15em;
+  }
```
wtritch: once I re-introduced bottom padding and stripped said bottom padding from the bar chart column, the scale-score component was driving the cell height since it uses an over-sized font.  This lead to awkward gaps between bar-chart vertical bars so I had to shrink the margins on the scale-score to keep it vertically centered but not driving cell height.  I don't like experimentally-determined css values but the only other option is to force the font-size of the "2400 +- 50" to be 1em, which makes the "+- 50" text small enough to be difficult to read...

webapp/src/main/webapp/src/app/aggregate-report/target/target-report-form.component.ts
esotrope: that's fun. my guess is there is no typeahead when we know they only have one org entitlement?

wtritch: nope, but since we initialize the value when the component loads, it should be handled and pre-populated in the case of one-org-entitlement.

PR #1381, opened 2018-05-31
Collapse results by default for school/grade results

jtreuting:
Request was to have the school/grade results collapsed by default as an administrator is more likely to want the aggregate view than the individual student results.  So we updated the language to say "Expand Results" since there is not the context of having just collapsed the results.  

Going to merge since I'd like it to be there tomorrow for the demo, but will address any feedback and update if needed.

### DIFF
--- a/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
+++ b/webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
@@ -30,6 +30,7 @@
                [hideAssessments]="gradesAreUnavailable"
                [assessmentProvider]="assessmentProvider"
                [assessmentExporter]="assessmentExporter"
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
```diff
@@ -30,6 +30,7 @@ <h2>{{currentSchool.name}}<span *ngIf="currentGrade"> - {{'common.assessment-gra
                [hideAssessments]="gradesAreUnavailable"
                [assessmentProvider]="assessmentProvider"
                [assessmentExporter]="assessmentExporter"
+               [isDefaultCollapsed]="true"
```
esotrope: we cant pick one or the other for both views? seems odd and inconsistent

PR #1386, opened 2018-05-31
sort claim codes on query screen to maintain consistency between quer…

dsnyder-fairway:
…y and results pages

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
@@ -5,7 +5,14 @@ import { CachingDataService } from '../shared/data/caching-data.service';
 import { OrganizationMapper } from '../shared/organization/organization.mapper';
 import { map } from 'rxjs/operators';
 import { AggregateServiceRoute } from '../shared/service-route';
-import { AssessmentTypeOrdering, BooleanOrdering, CompletenessOrdering } from '../shared/ordering/orderings';
+import {
+  AssessmentTypeOrdering,
+  BooleanOrdering,
+  ClaimCodeOrdering,
+  CompletenessOrdering,
+  SubjectOrdering
+} from '../shared/ordering/orderings';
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
```diff
@@ -27,7 +34,9 @@ export class AggregateReportOptionsService {
       map(serverOptions => <AggregateReportOptions>{
         assessmentGrades: serverOptions.assessmentGrades.concat(),
         assessmentTypes: serverOptions.assessmentTypes.concat().sort(assessmentTypeComparator),
-        claims: this.mapClaims(serverOptions.claims.concat()),
+        claims: this.mapClaims(serverOptions.claims.concat()).sort(join(
```
esotrope: could save some object creation by extracting this joined comparator to a constant outside the map fn

webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
```diff
@@ -1,9 +1,11 @@
 import { ranking } from '@kourge/ordering/comparator';
 import { ordering } from '@kourge/ordering';
 
-export const AssessmentTypeOrdering = ordering(ranking(['sum', 'ica', 'iab']));
+export const AssessmentTypeOrdering = ordering(ranking([ 'sum', 'ica', 'iab' ]));
 
-export const SubjectOrdering = ordering(ranking(['Math', 'ELA']));
+export const SubjectOrdering = ordering(ranking([ 'Math', 'ELA' ]));
+
+export const ClaimCodeOrdering = ordering(ranking([ '1', 'SOCK_2', '3', 'SOCK_R', 'SOCK_LS', '2-W', '4-CR' ]));
```
esotrope: do we not get the natural claim code order from the backend? its ideal if we can rely on the backend supplied order instead of hardcoding a copy here

dsnyder-fairway: 👍 I'll look around

PR #1390, opened 2018-06-01
calculate target aggregate values from the API response models

jtreuting:


webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.spec.ts
```diff
@@ -20,71 +20,50 @@ describe('Exam Calculator', () => {
...
-    let exams = [ 2580, 2551, 2850, 2985, 2650, 2651 ].map(x => {
-      let exam = new Exam();
-      exam.score = x;
-      return exam;
-    });
+    let scores = [ 2580, 2551, 2850, 2985, 2650, 2651 ];
 
     let fixture = new ExamStatisticsCalculator();
-    let actual = fixture.calculateAverage(exams);
+    let actual = fixture.calculateAverage(scores);
```
jtreuting: most of these changes are because i made the calculate average and standard error of the mean methods more general by accepting the scores instead of exams, so I could reuse for aggregate target reports

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -27,15 +29,28 @@ export class ExamStatisticsCalculator {
...
-    if (scores.length == 0) {
+    if (scored.length == 0) {
       return 0;
     }
 
-    return math.std(scores) / math.sqrt(scores.length);
+    return math.std(scored) / math.sqrt(scored.length);
+  }
+
+  calculateStandardError(scores: number[]): number {
```
jtreuting: @malaffoon I'm not using this right now because I looked at the SQL on the redshift side of things and it looks like it is using the Standard Error of the Mean.

```
round(avg(e.scale_score)) AS score,
round(stddev_samp(e.scale_score)/sqrt(count(*))) as std_err,
```

But when I looked at the scoring doc provided it looked like it should be this method instead.

image

thoughts?

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -113,6 +128,50 @@ export class ExamStatisticsCalculator {
     return pointFields;
   }
 
+  aggregateTargetScores(targetScoreExams: TargetScoreExam[]): AggregateTargetScoreRow[] {
+    let grouped = targetScoreExams.reduce((l, r) => {
```
jtreuting: this groups by target ID to get all the residual scores together

webapp/src/main/webapp/src/app/assessments/model/target-score-exam.model.ts
```diff
@@ -0,0 +1,7 @@
+import { Exam } from './exam.model';
+
+export class TargetScoreExam extends Exam {
```
jtreuting: this is what comes back from the new API.  FYI @dsnyder-fairway in case you started working this too on the service side.

webapp/src/main/webapp/src/app/assessments/model/aggregate-target-score-row.model.ts
```diff
@@ -0,0 +1,5 @@
+export class AggregateTargetScoreRow {
```
jtreuting: might need more info as the actual display is built out but just starting with the basics for calculation purposes

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -113,6 +128,50 @@ export class ExamStatisticsCalculator {
...
+          this.calculateStandardErrorOfTheMean(entry.studentScores)
+        )
+      };
+    });
+
+    return rows;
+  }
+
+  // TODO: enum?
+  mapTargetScoreDeltaToReportingLevel(delta: number, standardError: number): number {
```
jtreuting: again starting basic but it seems like this would naturally be an enum.  also will need to make the 0.2 value pulled from the server since it's configurable.

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
malaffoon: If you pull that apart you'll find it is the SEM: stddev_sample = sqrt(sum(delta^2)/(n-1)) so that formula is just dividing that by sqrt(n).

jtreuting: okay, let me check on that.  thanks for clarifying.

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
esotrope: really? `l` and `r`? what is this written in? c? 😩

jtreuting: e1, e2? exam1, exam2? prev, next?, left, right?  joe, bob?

esotrope: just giving you a hard time : j 
i was going to say go with `(groupedExams, exam)` but i think i like `(joe, bob)` better : j

jtreuting: i like groupedExams, exam

PR #1392, opened 2018-06-01
Task/improve user group ui

esotrope:
Adds tabs

image

Fixes subjects display

image

Adds form error help text and required name field

image

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/group-tabs.component.html
+++ b/webapp/src/main/webapp/src/app/groups/group-tabs.component.html
@@ -1,0 +1,21 @@
+<tabset *ngIf="initialized">
+  <tab>
+    <ng-template tabHeading>
+      {{'home.groups-well.assigned-groups-heading' | translate}}
+      <span class="badge ml-xs">{{groups.length}}</span>
+    </ng-template>
+    <div class="well">
+      <groups [groups]="groups"></groups>
+    </div>
+  </tab>
+  <tab *ngIf="userGroups"
+       [active]="userGroups.length">
+    <ng-template tabHeading>
+      {{'home.groups-well.created-groups-heading' | translate}}
+      <span class="badge ml-xs">{{userGroups.length}}</span>
+    </ng-template>
+    <div class="well">
+      <user-groups [groups]="userGroups"></user-groups>
+    </div>
+  </tab>
+</tabset>
--- a/webapp/src/main/webapp/src/app/groups/group-tabs.component.ts
+++ b/webapp/src/main/webapp/src/app/groups/group-tabs.component.ts
@@ -1,0 +1,40 @@
+import { Component, OnInit } from '@angular/core';
+import { Group } from './group';
+import { UserGroup } from '../user-group/user-group';
+import { UserService } from '../user/user.service';
+import { UserGroupService } from '../user-group/user-group.service';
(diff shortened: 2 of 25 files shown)


webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -32,65 +29,8 @@
            (ngModelChange)="onSearchChange($event)">
   </div>
 
-  <!-- My Groups Search Results List -->
   <div class="table-list-container">
-    <p-table [columns]="columns"
```
esotrope: broke this out into the `group-table` component

webapp/src/main/webapp/src/app/groups/group-tabs.component.ts
```diff
@@ -0,0 +1,40 @@
...
+              private groupService: GroupService,
+              private userGroupService: UserGroupService) {
+  }
+
+  ngOnInit(): void {
+    forkJoin(
+      this.groupService.getGroups(),
+      this.userGroupService.getGroups().pipe(
+        // interpret error as permission denied
+        catchError(() => of(null))
```
esotrope: permission denied handler includes all errors for now

webapp/src/main/webapp/src/app/groups/group-tabs.component.html
```diff
@@ -0,0 +1,21 @@
+<tabset *ngIf="initialized">
```
esotrope: now a wrapper `group-tabs` component fetches both group sets so it can know whether or not to display a tab and also how many groups there are of each type

webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -7,19 +7,16 @@
     <label>{{'groups.search-label' | translate}}</label>
 
     <ul class="list-unstyled list-inline pull-right">
-      <li>
-        <a *ngIf="defaultGroup"
-           routerLink="/group-dashboard/{{defaultGroup.id}}"
+      <li *ngIf="defaultGroup">
```
esotrope: moved these if's up to the <li> element in a bunch of places on the homepage to prevent margin issues when links are absent

webapp/src/main/webapp/src/app/shared/form/forms.ts
```diff
@@ -54,6 +54,20 @@ export class Forms {
     return control.invalid && (control.dirty || control.touched);
   }
 
+  public static submit(form: FormGroup, onValid: () => void, onInvalid?: () => void): void {
```
esotrope: @wtritch had to repeat this code for the user group form so i created this helper that can be used on the aggregate form as well.

webapp/src/main/webapp/src/app/user-group/user-group.component.ts
```diff
@@ -278,8 +283,6 @@ export class UserGroupComponent implements OnInit, OnDestroy {
 
   private updateSaveButtonDisabled(): void {
     this._saveButtonDisabled = !this.initialized
-      || this.groupForm == null
-      || !this.groupForm.formGroup.valid
```
esotrope: save button is now clickable even when the form is invalid so that we can display helpful text for why it is invalid under the form inputs

PR #1393, opened 2018-06-01
Add All/Remove All buttons for user groups

dsnyder-fairway:
initial view
screen shot 2018-06-01 at 2 26 58 pm

demonstrating works with filters
screen shot 2018-06-01 at 2 27 10 pm
screen shot 2018-06-01 at 2 27 16 pm

### DIFF
--- a/webapp/src/main/webapp/src/app/user-group/user-group-form.component.html
+++ b/webapp/src/main/webapp/src/app/user-group/user-group-form.component.html
@@ -24,7 +24,19 @@
 
   </div>
   <div class="form-group">
-    <label>{{'user-group-form.field.students.label' | translate}}</label>
+    <div class="flex-children">
+      <div class="flex-child grow pb-sm">
+        <label>{{'user-group-form.field.students.label' | translate}}</label>
+      </div>
+      <div *ngIf="group.students.length" class="flex-child">
+        <button type="submit"
+                class="btn btn-danger btn-sm"
+                (click)="removeAllStudents()">
+          <i class="fa fa-minus"></i>
+          {{'user-group-form.field.students.remove-all-button' | translate}}
+        </button>
+      </div>
+    </div>
     <input type="hidden"
            name="students"
            formControlName="students"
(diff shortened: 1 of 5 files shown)


esotrope:
I like the positioning of the buttons.
I think it might look better if we make them links without underlines though. Also we should be consistent with using either `x` or `-` to represent remove.

webapp/src/main/webapp/src/app/user-group/user-group-form.component.html
```diff
@@ -24,7 +24,19 @@
...
   <div class="form-group">
-    <label>{{'user-group-form.field.students.label' | translate}}</label>
+    <div class="flex-children">
+      <div class="flex-child grow pb-sm">
+        <label>{{'user-group-form.field.students.label' | translate}}</label>
+      </div>
+      <div *ngIf="group.students.length" class="flex-child">
+        <button type="submit"
+                class="btn btn-danger btn-sm"
+                (click)="removeAllStudents()">
```
esotrope: I would give some indirection here like `(click)="removeAllStudentsButtonClick()` also why is the button `type=submit`?

dsnyder-fairway: Sounds good. Not sure (copy and pasted). Should it be `type="button"`?

webapp/src/main/webapp/src/app/user-group/user-group.component.ts
```diff
@@ -195,16 +195,24 @@ export class UserGroupComponent implements OnInit, OnDestroy {
...
   onFormStudentClick(student: Student): void {
+    this.addStudents(student);
+  }
+
+  addAllStudents() {
+    this.addStudents(...this.filteredStudents);
+  }
+
+  private addStudents(...student: Student[]) {
+    const studentsToAdd = student.filter(aStudent => !this.group.students.includes(aStudent));
```
esotrope: not sure this is needed. we filter the student search results already based on those in the group already

dsnyder-fairway: ah gotcha. I was passing along `this.students` while I was doing this initially until I realized that would be incorrect. I'll remove this

webapp/src/main/webapp/src/app/user-group/user-group-form.component.html
esotrope: yeah you can either omit it or make it type=button

PR #1395, opened 2018-06-02
Task/refactor aggregate query form

wtritch:
This PR refactors the existing aggregate-report-query-form into a separate sub-component for each report type: GeneralPopulation, LongitudinalCohort, and Claim.  It also refactors the Target report-type query into a sub-component and includes it within the same query form container.

aggregate_report_query


webapp/src/main/webapp/src/app/aggregate-report/query-forms/aggregate-query-form-container.component.html
```diff
@@ -0,0 +1,80 @@
...
+                              (navItemChange)="setNavItems('LongitudinalCohort', $event)"
+                              [submitAction]="submitActionsByReportType['LongitudinalCohort']"></longitudinal-cohort-form>
+
+    <claim-report-form [hidden]="reportType !== 'Claim'"
+                              (navItemChange)="setNavItems('Claim', $event)"
+                              [submitAction]="submitActionsByReportType['Claim']"></claim-report-form>
+
+    <target-report-form [hidden]="reportType !== 'Target'"
+                        (navItemChange)="setNavItems('Target', $event)"
+                        [submitAction]="submitActionsByReportType['Target']"></target-report-form>
```
wtritch: The container is responsible for displaying the left-hand scroll-nav, the embargo warning, testing for access, and picking which report-type query form to display.

It asks the sub-components for the scroll-nav display items and injects a submission trigger called by the "Generate Report" button in the scroll-nav.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/aggregate-query-form-container.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  setNavItems(reportType: string, navItems: ScrollNavItem[]): void {
+    navItems = this._globalNavItems.concat(navItems);
+    this.navItemsByReportType[reportType] = navItems;
+    if (this.reportType === reportType) {
+      this.navItems = navItems;
+    }
+  }
+
+  onReportTypeChange(): void {
+    this.navItems = this.navItemsByReportType[this.reportType] || [];
```
esotrope: i think pre-population might save us some complexity here

jtreuting:
general UI comment before digging thru the code.  The suggestion during the demo which I agree with is to have the the report types in 2 columns and 2 rows like this:

Yearly Report                  Claim Report
Longitudinal Report       Target Report

With the descriptions under the buttons like they are now.

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -26,7 +28,28 @@ export class ScrollNavComponent {
...
+            && !Utils.isNullOrUndefined(item.id)) {
+            item = Object.assign(item, {
+              scrollTo: document.getElementById(item.id)
+            });
+          }
+
+          //Saturate text from translation key
+          if (Utils.isNullOrUndefined(item.text)
+            && !Utils.isNullOrUndefined(item.translationKey)) {
+            item = Object.assign(item, {
```
esotrope: note object.assign does mutate the first argument. if you want to create a new object here you need to do this:
```typescript
item = Object.assign({}, item, {...})
```

wtritch: Yeah, I'm not interested in modifying the original.  I suppose I could just add `{}` as a first param to make that a little more obvious?

webapp/src/main/webapp/src/app/aggregate-report/query-forms/aggregate-query-form-container.component.ts
wtritch: done

wtritch:
@jtreuting I'm going to merge this now to unblock anybody else who wants to work in aggregate-report-query land. (I believe @dsnyder-fairway has some claim query UI work he wanted to do.) 
I'll update the report selection UI in a followup PR.

wtritch:
@esotrope agreed... The TS files feel cleaner, but the html templates are still pretty heavily re-used between components.  Unfortunately, I couldn't find a clean way to break them up and share them among the 4 query components that didn't result in an explosion of tertiary components with only minor re-usability.

PR #1400, opened 2018-06-04
Modify the wide radio group component to accept a row count.

wtritch:
Update aggregate report query UI to display report type options on two rows.

image

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -1479,11 +1479,12 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
   font-size: 100%;
 }
 wide-radio-group {
+
   .btn-toolbar-grid {
-    .btn-toolbar;
     display: grid;
     grid-template-rows: repeat(2, auto);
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -1479,11 +1479,12 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
   font-size: 100%;
 }
 wide-radio-group {
+
   .btn-toolbar-grid {
-    .btn-toolbar;
```
wtritch: This looked to be vestigial?

webapp/src/main/webapp/src/styles.less
```diff
@@ -1479,11 +1479,12 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
...
     display: grid;
     grid-template-rows: repeat(2, auto);
     grid-auto-flow: column;
+    grid-gap: 0 @spacer;
```
wtritch: No vertical gap, standard horizontal gap

PR #1401, opened 2018-06-04
Teacher Student Group API: Concurrency Tests

agorina:


common/src/test/java/org/opentestsystem/rdw/reporting/common/multithreaded/MultiThreadedApplicationHelper.java
```diff
@@ -0,0 +1,16 @@
+package org.opentestsystem.rdw.reporting.common.multithreaded;
+
+import org.springframework.boot.builder.SpringApplicationBuilder;
+
+public class MultiThreadedApplicationHelper {
```
agorina: Since we have these tests in two projects, I moved the shared code into common location. This should probably be moved into RDW_Common and shared with Ingest...(when/if we have time for this clean up).

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/multithreaded/MultiThreadedApplicationTest.java
```diff
@@ -1,75 +1,22 @@
...
-                logger.info("Running test for [" + test + "] with thread count =" + TestThreadCount);
-                final boolean passed = ctx.getBean(test, MultiThreadedTest.class).run(pool, TestThreadCount);
-                logger.info(test + " " + (passed ? "passed" : "failed"));
-
-                failed.add(passed ? 0 : 1);
-            }
-        }
-
-        System.exit(failed.stream().mapToInt(Integer::intValue).sum());
+        MultiThreadedApplicationHelper.run(args, MultiThreadedApplicationTest.class);
```
agorina: Simplified version due to the common code.

reporting-service/build.gradle
```diff
@@ -28,3 +28,13 @@ dependencies {
...
 }
+
+//This will allow for running of the multithreaded test application
+task runMTTest(type: JavaExec) {
+    group = 'Verification'
+    description = 'Runs MultiThreadedApplicationTest'
+    outputs.upToDateWhen { false }
+    doFirst { println "Running multithreaded tests..." }
+    classpath = sourceSets.test.runtimeClasspath
+    main = 'org.opentestsystem.rdw.reporting.multithreaded.MultiThreadedApplicationTest'
```
agorina: @malaffoon , could you please add this to CI?

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/JdbcUserGroupRepository.java
```diff
@@ -25,7 +25,7 @@
 
 
 @Repository
-class JdbcUserGroupRepository implements UserGroupRepository {
+public class JdbcUserGroupRepository implements UserGroupRepository {
```
agorina: I wanted Multithreaded test to load only what it needs. Scanning the folder pulls way more than needed. I changed it to `public` to include it as `import` for the test.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/multithreaded/MultiThreadedUserGroupRepository.java
```diff
@@ -0,0 +1,72 @@
...
+                repository.create("testUser", Group.builder()
+                        .userCreated(true)
+                        .name(UUID.randomUUID().toString())
+                        .subjectCodes(ImmutableSet.of("Math"))
+                        .studentIds(ImmutableSet.of(-30L))
+                        .build()
+                ).copy()
+                        .studentIds(ImmutableSet.of(-1L, -30L))
+                        .build()
+        );
```
agorina: This is the goal of this PR - to prove that we do not deadlock when create/update groups.

reporting-service/build.gradle
malaffoon: It is already running MT tests.

common/src/test/java/org/opentestsystem/rdw/reporting/common/multithreaded/MultiThreadedCommandLineRunner.java
```diff
@@ -0,0 +1,79 @@
...
+@Component
+@Profile("mtTest")
+public class MultiThreadedCommandLineRunner implements CommandLineRunner {
+
+    private static final Logger logger = LoggerFactory.getLogger(MultiThreadedCommandLineRunner.class);
+    @Autowired
+    private ApplicationContext appContext;
+
+    // expected number of {@link MultiThreadedTest}s.
+    @Value("${mtTest.expectedTests:1}")
```
esotrope: not a big deal but i think concurrency is actually what we are testing for right? we should probably call things concurrency tests. Another thought i had is we could suffix the tests with "CT" like we do for "IT"

reporting-service/build.gradle
agorina: I was not sure if anything needs to be change to include this one. Thank you.

PR #1403, opened 2018-06-04
Update aggregate target result display to account for client feedback.

wtritch:
The column header was updated to re-use shared header language.
Added a min-width to the "Above" "Below" and "Near" labels so they display at the same width with default display text.
Updated the overall-relative values to use up, down, and equals icons.  Made all icons blue.
Updated "Insufficient Data" display value to use label-styling rather than an icon.
Centered data column values in target results table to align with other aggregate results tables.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
@@ -154,51 +154,55 @@
             </span>
           </div>
 
-          <div *ngSwitchCase="'studentRelativeResidualScoresLevel'">
+          <div *ngSwitchCase="'studentRelativeResidualScoresLevel'"
+               class="text-center">
 
             <!-- pick the icon -->
-            <span [ngSwitch]="rowData.studentRelativeResidualScoresLevel"
-              style="font-size: 1.2em">
-              <span *ngSwitchCase="'Above'"
-                    class="sb-iab-green">
-                <sb-icon class="fa-lg" style="stroke-width: 1px; stroke: #000;" icon="sbac-above-average"></sb-icon>
+            <ng-container [ngSwitch]="rowData.studentRelativeResidualScoresLevel">
+              <span *ngSwitchCase="'Above'">
+                <span class="blue"><i class="fa fa-lg fa-arrow-up"></i></span>
+                {{'aggregate-report-table.target.overall.' + rowData.studentRelativeResidualScoresLevel | translate}}
               </span>
-              <span *ngSwitchCase="'Near'"
-                    class="sb-iab-yellow">
-                <sb-icon class="fa-lg" style="stroke-width: 1px; stroke: #000;" icon="sbac-average"></sb-icon>
+              <span *ngSwitchCase="'Near'">
+                <span class="blue">
+                  <sb-icon class="fa-lg" icon="fa-equals"></sb-icon>
+                </span>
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -154,51 +166,55 @@
...
+              <span *ngSwitchCase="'Above'">
+                <span class="blue"><i class="fa fa-lg fa-arrow-up"></i></span>
+                {{'aggregate-report-table.target.overall.' + rowData.studentRelativeResidualScoresLevel | translate}}
               </span>
-              <span *ngSwitchCase="'Near'"
-                    class="sb-iab-yellow">
-                <sb-icon class="fa-lg" style="stroke-width: 1px; stroke: #000;" icon="sbac-average"></sb-icon>
+              <span *ngSwitchCase="'Near'">
+                <span class="blue">
+                  <sb-icon class="fa-lg" icon="fa-equals"></sb-icon>
```
wtritch: We have an older version of FontAwesome, which doesn't include the equals icon.  Rather than update our dependency version, (which may change any and every icon in the system,) I just manually included the equals icon as SVG.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -154,51 +166,55 @@
...
 
-          <div *ngSwitchCase="'standardMetRelativeResidualLevel'">
+          <div *ngSwitchCase="'standardMetRelativeResidualLevel'"
+               class="text-center">
 
             <!-- pick the label color -->
             <span [ngSwitch]="rowData.standardMetRelativeResidualLevel">
               <span *ngSwitchCase="'Above'"
-                    class="label sb-iab-green">
+                    class="label label-lg performance-min-width sb-iab-green">
```
wtritch: I created a performance-min-width class that sets a min-width of 75px.  If a tenant chooses to change the label text via translation-override the label will still scale to whatever length text they choose.

webapp/src/main/webapp/src/assets/svgs/fa-equals.svg
```diff
@@ -0,0 +1,6 @@
+<svg aria-hidden="true" data-prefix="fas" data-icon="equals" role="img" xmlns="http://www.w3.org/2000/svg"
+     viewBox="0 0 448 512" class="svg-inline--fa fa-equals fa-w-14 fa-5x">
+  <path fill="currentColor"
+        d="M416 304H32c-17.67 0-32 14.33-32 32v32c0 17.67 14.33 32 32 32h384c17.67 0 32-14.33 32-32v-32c0-17.67-14.33-32-32-32zm0-192H32c-17.67 0-32 14.33-32 32v32c0 17.67 14.33 32 32 32h384c17.67 0 32-14.33 32-32v-32c0-17.67-14.33-32-32-32z"
+        class=""></path>
+</svg>
```
wtritch: Grabbed from the font-awesome website.

webapp/src/main/webapp/src/styles.less
```diff
@@ -1552,3 +1552,7 @@ user-group {
   }
 }
 
+.performance-min-width {
```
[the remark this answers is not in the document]
wtritch: I'd prefer `performance-label` to `mid-width-md`  Something like this is extremely content-based rather than layout based like our existing `label-max-width`

PR #1406, opened 2018-06-04
Task/dwr 1853 target appsettings

jtreuting:
Bring server side config settings for target report to the client

### DIFF
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -132,4 +132,4 @@ reporting:
 
   target-report:
     insufficient-data-cutoff: 0.2
-    min-number-of-students: 1
+    min-number-of-students: 0
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -129,3 +129,7 @@ reporting:
   # english-learners should be lep or elas
   english-learners:
     - elas
+
+  target-report:
+    insufficient-data-cutoff: 0.2
+    min-number-of-students: 1
--- a/webapp/src/main/webapp/src/app/app-settings.service.ts
+++ b/webapp/src/main/webapp/src/app/app-settings.service.ts
@@ -28,6 +28,10 @@ export class ApplicationSettingsService {
           code: serverSettings.state.code,
           name: serverSettings.state.name
         },
+        targetReport: {
+          insufficientDataCutoff: serverSettings.targetReport.insufficientDataCutoff,
+          minNumberOfStudents: serverSettings.targetReport.minNumberOfStudents
+        },
         transferAccess: serverSettings.transferAccessEnabled,
         uiLanguages: serverSettings.uiLanguages,
         userGuideUrl: serverSettings.userGuideUrl
--- a/webapp/src/main/webapp/src/app/app-settings.service.ts
+++ b/webapp/src/main/webapp/src/app/app-settings.service.ts
@@ -30,7 +30,7 @@ export class ApplicationSettingsService {
         },
         targetReport: {
           insufficientDataCutoff: serverSettings.targetReport.insufficientDataCutoff,
-          minNumberOfStudents: serverSettings.targetReport.minNumberOfStudents
+          minimumStudentCount: serverSettings.targetReport.minNumberOfStudents
         },
         transferAccess: serverSettings.transferAccessEnabled,
         uiLanguages: serverSettings.uiLanguages,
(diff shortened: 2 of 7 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -129,3 +129,7 @@ reporting:
   # english-learners should be lep or elas
   english-learners:
     - elas
+
+  target-report:
```
jtreuting: this is a duplicate of what @agorina did in the aggregate-service.  Should it be moved to somewhere common?  thoughts?

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.html
```diff
@@ -139,23 +139,23 @@
             <!-- pick the label color -->
             <span [ngSwitch]="getTargetReportingLevelString(rowData.standardMetRelativeLevel)">
               <span *ngSwitchCase="'Above'"
-                    class="label sb-iab-green">
+                    class="label label-lg sb-iab-green">
```
jtreuting: not related, but bigger labels on the report output before i forgot

webapp/src/main/webapp/src/app/app-settings.service.ts
```diff
@@ -28,6 +28,10 @@ export class ApplicationSettingsService {
           code: serverSettings.state.code,
           name: serverSettings.state.name
         },
+        targetReport: {
+          insufficientDataCutoff: serverSettings.targetReport.insufficientDataCutoff,
+          minNumberOfStudents: serverSettings.targetReport.minNumberOfStudents
```
esotrope: would prefer `minimumStudentCount`

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
```diff
@@ -11,10 +11,20 @@ import { Subgroup } from '../../aggregate-report/subgroup/subgroup';
...
+
   constructor(private examStatisticsCalculator: ExamStatisticsCalculator,
               private subgroupMapper: SubgroupMapper) {
   }
 
+  /**
+   * Set the standard error cutoff used when determining if the data is sufficient or not
+   * @param {number} cutoff
+   */
+  setInsufficientDataCutoff(cutoff: number): void {
```
esotrope: nit: you could use a setter:
```typescript
set insufficientDataCutoff(cutoff: number): void {
  //...
}
```

jtreuting: i always forget about those in typescript, thanks

webapp/src/main/webapp/src/app/app-settings.service.ts
jtreuting: that's fine with me, was being consistent with the existing setting, but i'll change it on the client side

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -128,12 +125,15 @@ export class TargetReportComponent implements OnInit {
...
-
-    ).subscribe(([ allTargets, targetScoreExams, subgroupOptions ]) => {
+      this.applicationSettingsService.getSettings()
+    ).subscribe(([ allTargets, targetScoreExams, subgroupOptions, applicationSettings ]) => {
       this.targetScoreExams = targetScoreExams;
       this.subgroupOptions = subgroupOptions;
       this.allTargets = allTargets;
 
+      this.minimumStudents = applicationSettings.targetReport.minNumberOfStudents;
+      this.targetStatisticsCalculator.setInsufficientDataCutoff(applicationSettings.targetReport.insufficientDataCutoff);
```
esotrope: i think ideally we pass this in as a method param, or the service makes the call to get this value at method invocation time and returns an observable

jtreuting: i originally was passing it in, but it is used about 4 method calls within the main one and that didn't feel right.  it's not really a concern of the first call to generate the aggregate rows.  it also feels weird to have what is essentially a set of helper methods call a service and get the observable.

webapp/src/main/resources/application.yml
malaffoon: As discussed offline, copy and paste is okay for this. Please add to configuration samples in RDW.

jtreuting: yep, done: https://github.com/SmarterApp/RDW/pull/32

PR #1408, opened 2018-06-04
Task/dwr 1853 default claim sort

jtreuting:
Handle claim ordering on a per subject basis

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/model/aggregate-target-score-row.model.ts
+++ b/webapp/src/main/webapp/src/app/assessments/model/aggregate-target-score-row.model.ts
@@ -4,6 +4,7 @@ export class AggregateTargetScoreRow {
   targetId: number;
   target: string;
   claim: string;
+  claimOrder: number;
   subgroup: Subgroup;
   studentsTested: number;
   standardMetRelativeLevel: TargetReportingLevel;
--- a/webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
@@ -16,6 +16,15 @@ export class TargetStatisticsCalculator {
   }
 
   /**
+   * Used to determine the ordering of the claims on the Target report
+   * @type {Map<string, number>}
+   */
+  claimOrder: Map<string, number> = new Map([
+    ['1', 1], ['2', 2], ['3', 3], ['4', 4],
+    ['1-LT', 1], ['1-IT', 2], ['2-W', 3], ['3-L', 4], ['3-S', 5], ['4-CR', 6]
+  ]);
+
+  /**
    * Calculate the overall scores for each target from the raw API data
    * @param {Target[]} allTargets
    * @param {TargetScoreExam[]} targetScoreExams
@@ -78,6 +87,7 @@ export class TargetStatisticsCalculator {
     return <AggregateTargetScoreRow>{
       targetId: groupedScore.targetId,
       claim: groupedScore.claim,
+      claimOrder: this.claimOrder.get(groupedScore.claim),
       target: groupedScore.target,
       subgroup: groupedScore.subgroup,
       studentsTested: numStudents,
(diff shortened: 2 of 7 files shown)


webapp/src/main/webapp/src/app/assessments/model/aggregate-target-score-row.model.ts
```diff
@@ -4,6 +4,7 @@ export class AggregateTargetScoreRow {
   targetId: number;
   target: string;
   claim: string;
+  claimOrder: number;
```
jtreuting: field used by the data table sorting, so this is just an int (the index in the ordering array)

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
```diff
@@ -31,9 +32,9 @@ export class TargetStatisticsCalculator {
    * @param {TargetScoreExam[]} targetScoreExams
    * @returns {AggregateTargetScoreRow[]}
    */
-  aggregateOverallScores(allTargets: Target[], targetScoreExams: TargetScoreExam[]): AggregateTargetScoreRow[] {
+  aggregateOverallScores(subjectCode: string, allTargets: Target[], targetScoreExams: TargetScoreExam[]): AggregateTargetScoreRow[] {
```
jtreuting: need to pass in the assessment subject now in order to get the appropriate claim orderings for the claimOrder property.  as i'm typing this, wonder if this is something that I post process in the target report component since it's really a report concern?

webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
```diff
@@ -8,3 +8,13 @@ export const SubjectOrdering = ordering(ranking(['Math', 'ELA']));
 export const BooleanOrdering = ordering(ranking([ 'yes', 'no', 'undefined' ]));
 
 export const CompletenessOrdering = ordering(ranking([ 'Complete', 'Partial' ]));
+
+export const SubjectClaimOrder: Map<string, string[]> = new Map([
```
jtreuting: holds the order of the claims which can be used independently when needed, but also used in the ordering below

webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
```diff
@@ -8,3 +8,13 @@ export const SubjectOrdering = ordering(ranking(['Math', 'ELA']));
...
+  ['Math', ['1', '2', '3', '4']],
+  ['ELA', ['1-LT', '1-IT', '2-W', '3-L', '3-S', '4-CR']]
+]);
+
+export const SubjectClaimOrderings: Map<string, Ordering<string>> = new Map([
```
jtreuting: use the claim code array as the ordering.

webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
esotrope: is there a reason we need this and cant just use the `SubjectClaimOrderings`?

jtreuting: i'm using it to get a `claimOrder` on the row that can be used more easily by the data table as the sort field.  not sure it's absolutely needed, probably could find another way, but it felt like it gave more flexibility as well and it is not duplicating data since one uses the other.

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
wtritch: I like it here, since it at least gives us the option of off-loading it to the back-end in the future if we don't want to front-load all targets to the front-end.

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -191,7 +190,7 @@ export class TargetReportComponent implements OnInit {
 
     this.aggregateTargetScoreRows.sort(
       join(
-        ordering(byString).on<AggregateTargetScoreRow>(row => row.claim).compare,
+        SubjectClaimOrderings.get(this.assessment.subject).on<AggregateTargetScoreRow>(row => row.claim).compare,
```
wtritch: Won't this break if we don't have claim orderings defined for "ELPAC" ?

jtreuting: yes it will.  i wasn't really trying to address new subjects as part of it, but it would be easy enough to check if it exists and if not do a basic sort on the claim field alphabetically right now.  sound good?

wtritch: Sgtm

PR #1414, opened 2018-06-05
only use the selected subjects when generating the claims by subject …

jtreuting:
…on the claim report

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
@@ -106,10 +106,7 @@ export class AggregateReportRequestMapper {
     } else if (this.reportService.getEffectiveReportType(settings.reportType, assessmentDefinition) === 'Claim') {
       query.assessmentGradeCodes = settings.claimReport.assessmentGrades;
       query.schoolYears = settings.claimReport.schoolYears;
-      query.claimCodesBySubject = this.claimsBySubjectMapping(options.subjects.map(
-        subject => subject.value),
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -106,10 +106,7 @@ export class AggregateReportRequestMapper {
     } else if (this.reportService.getEffectiveReportType(settings.reportType, assessmentDefinition) === 'Claim') {
       query.assessmentGradeCodes = settings.claimReport.assessmentGrades;
       query.schoolYears = settings.claimReport.schoolYears;
-      query.claimCodesBySubject = this.claimsBySubjectMapping(options.subjects.map(
-        subject => subject.value),
-        settings.claimReport.claimCodesBySubject
-      );
+      query.claimCodesBySubject = this.claimsBySubjectMapping(settings.subjects, settings.claimReport.claimCodesBySubject);
```
jtreuting: was using all subjects instead of only the selected ones

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -444,7 +441,9 @@ export class AggregateReportRequestMapper {
       obj[ subject ] = [];
     }
     for (const claim of claims) {
-      obj[ claim.subject ].push(claim.code);
+      if (obj[ claim.subject ] != null) {
```
jtreuting: need to protect against the subject of all claims since we are only prepping for the selected ones

PR #1420, opened 2018-06-05
Feature/target aggregate report results

wtritch:
This PR updates the aggregate report table, service, mappers, etc to properly display aggregate target reports.

Still TODO: Display overall performance metrics for the assessment as a whole.

target_report


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -102,4 +104,38 @@ export class AggregateReportService {
       });
   }
 
+  /**
+   *
+   * @param {number} id
+   * @returns {Observable<BasicReport>}
+   * @deprecated With configurable subjects, we should get target display text via translation
+   */
+  getTargetReport(id: number): Observable<BasicReport> {
```
wtritch: This method just fetches the report data, follows it with a call to get the targets for the specified assessment, then merges the two to get target name and description into the response payload.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.mapper.ts
```diff
@@ -39,24 +36,30 @@ export class AggregateReportItemMapper {
...
 
-    let totalTested = 0;
-
     for (let level = 1; level <= assessmentDefinition.performanceLevelCount; level++) {
       const count = measures[ `level${level}Count` ] || 0;
-      totalTested += count;
       itemPerformanceLevelCounts.push(count);
     }
-    item.studentsTested = totalTested;
+    item.studentsTested = row.measures.studentCount;
```
wtritch: I switched this mapper to use the returned student count rather than sum up all of the performance level counts to calculate the total number of students.  (With target reports, we don't have any performance level counts to sum over.)

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -168,60 +168,62 @@
...
-            <!-- pick the icon -->
-            <ng-container [ngSwitch]="rowData.studentRelativeResidualScoresLevel">
-              <span *ngSwitchCase="'Above'">
-                <span class="blue"><i class="fa fa-lg fa-arrow-up"></i></span>
-                {{'aggregate-report-table.target.overall.' + rowData.studentRelativeResidualScoresLevel | translate}}
-              </span>
-              <span *ngSwitchCase="'Near'">
-                <span class="blue">
-                  <sb-icon class="fa-lg" icon="fa-equals"></sb-icon>
+            <span *ngIf="!rowData.studentsTested">-</span>
```
wtritch: I added a fallback to display the standard no-data `-` if this row has no students.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -168,60 +168,62 @@
...
-                    class="label label-lg performance-label sb-iab-red">
-                {{'aggregate-report-table.target.standard.Below' | translate}}
-              </span>
-              <span *ngSwitchCase="'InsufficientData'"
-                    class="label label-lg gray-light">
-                {{'aggregate-report-table.target.standard.InsufficientData' | translate}}
-              </span>
-              <span *ngSwitchCase="'Excluded'">
-                {{'aggregate-report-table.target.standard.Excluded' | translate}}
+            <span *ngIf="!rowData.studentsTested">-</span>
```
dsnyder-fairway: could use `{{'common.missing' | translate}}`

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -35,11 +36,15 @@ const SchoolYearOrdering: Ordering<AggregateReportItem> = ordering(byNumber)
...
 
-const ClaimOrdering: Ordering<AggregateReportItem> = ordering(byString)
+const ScorableClaimOrdering: Ordering<AggregateReportItem> = ordering(byString)
   .on(item => item.claimCode);
 
-const TargetOrdering: Ordering<AggregateReportItem> = ordering(byString)
-  .on(item => item.targetNaturalId);
+const OrganizationalClaimOrderingProvider: (subjectCode: string) => Ordering<AggregateReportItem> = (subjectCode) =>
+  (SubjectClaimOrderings.get(subjectCode) || ordering(byString))
+    .on(item => item.claimCode);
```
wtritch: Note that we now have two "claim" orderings: one for scorable claims (e.g. SOCK_R) and one for organizational claims (e.g. 1-LT)
The problem is that item.claimCode represents a scorable claim for Claim reports, and it represents an organizational claim for Target reports.  I toyed with the idea of breaking item.claimCode into item.scorableClaimCode and item.organizationalClaimCode but decided it was easier to add some "if reportType === 'Target'" checks instead... thoughts?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -57,20 +62,20 @@ const TargetOrdering: Ordering<AggregateReportItem> = ordering(byString)
     '[class.disabled-table]': 'preview'
   }
 })
-export class AggregateReportTableComponent implements OnInit {
```
dsnyder-fairway: nit: could remove `ngOnInit`

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -284,7 +288,7 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
             options: this.options,
             assessmentDefinition: this.assessmentDefinition,
             rows: [ item ],
-            reportType: hasClaimCode ? 'Claim' : 'GeneralPopulation'
+            reportType: this.report.request.query.reportType
```
wtritch: This confused me a bit... Anyone remember why we didn't just pass the report type on through and instead condensed it down into just two of the previous three options?

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -198,15 +199,6 @@ export class TargetReportComponent implements OnInit, ExportResults {
...
 
   sortRows() {
-    const byTarget = (a: string, b: string) => {
-      let numA = Number(a);
-      let numB = Number(b);
-
-      if (!Number.isNaN(numA) && !Number.isNaN(numB)) return numA - numB;
-
-      return a.localeCompare(b);
-    };
```
wtritch: I consolidated this into a shared "byNumericString" comparator in Utils.

webapp/src/main/webapp/src/app/shared/target/target.service.ts
```diff
@@ -14,6 +15,19 @@ export class TargetService  {
...
+        )
+      );
+    }
```
wtritch: I won't be committing this, but I wanted to call out the kind of shenanigans that are required when running target reports locally.  QA is providing me aggregate results.  I then take the assessment id from those results and look up target name and description using my localhost Reporting service.
Since QA and my local DB have different assessments with different IDs, this doesn't work out so well, and I have to hand-jam in some mocked data that I manually pulled from QA.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
wtritch: good call, switched that everywhere in the table component

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
wtritch: shoot, could have sworn I did that.  Nice catch, done

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -102,4 +104,38 @@ export class AggregateReportService {
...
+        return report.rows.length === 0
+          ? Observable.of({report: report, targets: []})
+          : this.targetService.getTargetsForAssessment(report.rows[0].assessment.id)
+            .pipe(
+              map((targets) => <any>{report: report, targets: targets})
+            );
+      })
+      .pipe(
+        map(({report, targets}) => {
+          const targetByNaturalId = {};
```
jtreuting: any reason to not use a Map here?  feels cleaner since you know the intention instead of the object syntax.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.mapper.ts
```diff
@@ -39,24 +36,30 @@ export class AggregateReportItemMapper {
     item.schoolYear = row.assessment.examSchoolYear;
     item.organization = this.organizationMapper.map(row.organization);
     item.claimCode = row.claimCode;
+    item.targetNaturalId = row.targetNaturalId;
+    item.studentRelativeResidualScoresLevel = row.studentRelativeResidualScoresLevel;
+    item.standardMetRelativeResidualLevel = row.standardMetRelativeResidualLevel;
+    item.targetDescription = row.targetDescription;
+
+    const dashIndex = row.targetCode ? row.targetCode.indexOf('-') : -1;
```
jtreuting: not sure what is considered best practice with typescript, but i've leaned towards doing a `row.targetCode == null` which handles null and undefined.  this is maybe more a question than a suggestion to change.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.ts
```diff
@@ -33,6 +33,9 @@ export class AggregateReportItem {
   subgroup: Subgroup;
   studentRelativeResidualScoresLevel?: 'Above' | 'Near' | 'Below' | 'InsufficientData' | 'Excluded';
```
jtreuting: I have an enum on the Teacher side of things, maybe we should use the same approach?  i also added NoResults to more simply handle displaying the - when there are no results in my table switch statement so they are all checks on that same property

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -168,60 +168,62 @@
...
-              </span>
-              <span *ngSwitchCase="'InsufficientData'"
-                    class="label label-lg gray-light">
-                {{'aggregate-report-table.target.overall.' + rowData.studentRelativeResidualScoresLevel | translate}}
-              </span>
-              <span *ngSwitchDefault>
-                {{'aggregate-report-table.target.overall.' + rowData.studentRelativeResidualScoresLevel | translate}}
-              </span>
+                <span *ngSwitchCase="'Near'">
+                  <span class="blue">
```
jtreuting: cool, i'll grab these icons

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -221,7 +213,7 @@ export class TargetReportComponent implements OnInit, ExportResults {
     this.aggregateTargetScoreRows.sort(
       join(
         claimOrdering.on<AggregateTargetScoreRow>(row => row.claim).compare,
-        ordering(byTarget).on<AggregateTargetScoreRow>(row => row.target).compare,
+        ordering(Utils.byNumericString).on<AggregateTargetScoreRow>(row => row.target).compare,
```
jtreuting: nice, thanks

webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -250,4 +250,21 @@ export class Utils {
...
 
+  /**
+   * Comparator for ordering two strings which may represent numbers such that
+   * "A" < "B" and "2" < "10"
+   *
+   * @param {string} a  A string value
+   * @param {string} b  A string value
+   * @returns {number}  0 if equal, less than 0 if a < b, greater than 0 if a > b
+   */
+  static byNumericString = (a: string, b: string) => {
```
jtreuting: does this belong in the shared orderings.ts instead?

webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -250,4 +250,21 @@ export class Utils {
...
+    let numA = Number(a);
+    let numB = Number(b);
```
jtreuting: these could be `const`, landon caught this in some changes he just did and put up a PR for

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.mapper.ts
esotrope: yeah i think we've been inconsistent with it. using == is a must when checking a numeric value though since zero evaluates to false

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
esotrope: report type check is fine for now but i would drop this comment into the code for later

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -269,15 +270,24 @@ export class AggregateReportTableComponent implements OnInit {
...
   getClaimCodeTranslation(row: AggregateReportItem): string {
     return this.translate.instant(this.getClaimCodeTranslationKey(row));
   }
 
-  getTargetDisplay(row: AggregateReportItem): string {
-    return "Fetch target display from API here";
+  getTargetDisplay(row: AggregateReportItem): {name: string, description: string} {
+    const targetName: string = row.targetCode
+      ? row.targetCode
+      : this.translate.instant("common.unknown") + ' (' + row.targetNaturalId + ')';
```
esotrope: nit: `"` -> `'`

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -243,9 +243,14 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   private loadReport(): void {
     this.spinnerModal.loading = true;
 
-    const observable = this.isLongitudinal
-      ? this.reportService.getLongitudinalReport(this.report.id)
-      : this.reportService.getAggregateReport(this.report.id);
+    let observable;
+    if (this.effectiveReportType === "LongitudinalCohort") {
```
esotrope: should we put this routing behind a service call?
```typescript
this.reportService.getReportByTypeAndId(type, id)
```

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -102,4 +104,38 @@ export class AggregateReportService {
...
+   *
+   * @param {number} id
+   * @returns {Observable<BasicReport>}
+   * @deprecated With configurable subjects, we should get target display text via translation
+   */
+  getTargetReport(id: number): Observable<BasicReport> {
+    return this.getAggregateReport(id)
+      .flatMap((report) => {
+        return report.rows.length === 0
+          ? Observable.of({report: report, targets: []})
```
esotrope: nit: `Observable.of` -> `of` (rxjs recommendation)

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -102,4 +104,38 @@ export class AggregateReportService {
...
+      .pipe(
+        map(({report, targets}) => {
+          const targetByNaturalId = {};
+          targets.forEach(target => {
+            targetByNaturalId[target.naturalId] = target;
+          });
+          for (let row of report.rows) {
+            const target = targetByNaturalId[row.targetNaturalId];
+            if (target) {
+              (<any>row).targetCode = target.code;
```
esotrope: if it is not readonly maybe we should remove the readonly flag?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
esotrope: @dsnyder-fairway any insight? i've got nothing

webapp/src/main/webapp/src/app/shared/support/support.ts
esotrope: so what is being modeled with numeric strings instead of numbers?

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
wtritch: good call, done

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
dsnyder-fairway: I think I introduced using the Claim report type and everything before that was using GeneralPopulation? I don't see how it could hurt doing this the correct way though ;)

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
wtritch: Nope, switched to:
```
          const targetByNaturalId: Map<number, Target> = targets.reduce((targetMap, target) => {
            targetMap.set(target.naturalId, target);
            return targetMap;
          }, new Map<number, any>());
```

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.mapper.ts
wtritch: Switched to:
```
    const dashIndex = Utils.isNullOrUndefined(row.targetCode) ? -1 : row.targetCode.indexOf('-');
    item.targetCode = dashIndex === -1
      ? row.targetCode
      : row.targetCode.substring(0, dashIndex);
```

webapp/src/main/webapp/src/app/shared/support/support.ts
wtritch: Sometimes target names look like `A, B, C, etc` and other times they look like `1, 2, ...11, etc` And we want "C" to come after "A" and "11" to come after "2"

wtritch: @jtreuting good call, moved

webapp/src/main/webapp/src/app/shared/support/support.ts
wtritch: changed

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
wtritch: doh, changed

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
wtritch: That feels like bleeding component logic into the service layer...  although it doesn't really belong in either location... The only reason it's here rather than in a resolver is because we need to be able to warn the user when they're about to download and display 100MB of report data...  I'm tempted to leave it.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
wtritch: doh, changed

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
wtritch: Good idea.  Ideally it would be, but these properties are caught in a strange halfway state until we can finalize configurable subjects.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-item.ts
wtritch: Switched to using the enum, but I'd like to leave the no-results switch on studentsTested == 0 to be consistent with the other "no-results" checks in this table.

PR #1422, opened 2018-06-05
sort when re-ordering columns

dsnyder-fairway:
downsides are:

1. briefly see the rows get re-ordered when changing the order of columns.

2. if someone sorts a column, after reordering the column they lose that sort

upsides:

1. we no longer have duplicate or missing rows


dsnyder-fairway:
Is this alright to merge? Addresses two major bugs

PR #1429, opened 2018-06-05
changed My Groups -> Groups

dsnyder-fairway:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1091,7 +1091,7 @@
     },
     "groups-well": {
       "assigned-groups-heading": "Assigned Groups",
-      "created-groups-heading": "My Groups"
+      "created-groups-heading": "Groups"
```
esotrope: i think we'll want to revert this - i think the confusion was on the results page not the home page. also i already patched the result page to say "Groups" instead of "My Groups"

dsnyder-fairway: Ah my bad. We can revert

PR #1432, opened 2018-06-06
Change reportType to be an enum to make it clear when we're dealing w…

wtritch:
…ith server vs client report types

I had a lot of difficulty distinguishing between the two, so I "enumified" the values to make it clear when we're dealing with each.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
@@ -12,7 +12,7 @@ import { ordering } from '@kourge/ordering';
 import { AggregateReportQuery } from '../../report/aggregate-report-request';
 import { DisplayOptionService } from '../../shared/display-options/display-option.service';
 import { TranslateService } from '@ngx-translate/core';
-import { AggregateReportRequestMapper, AggregateReportServerType } from '../aggregate-report-request.mapper';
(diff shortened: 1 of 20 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -87,7 +89,7 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   }
 
   get effectiveReportType() {
-    return this.reportService.getEffectiveReportType(<'GeneralPopulation' | 'LongitudinalCohort' | 'Claim' | 'Target'>this.report.request.query.reportType, this.assessmentDefinition);
+    return this.reportService.getEffectiveReportType(this.mapToReportType(this.query.reportType), this.assessmentDefinition);
```
wtritch: This was one issue, the query reportType is a server reportType

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -288,7 +295,7 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
             options: this.options,
             assessmentDefinition: this.assessmentDefinition,
             rows: [ item ],
-            reportType: this.report.request.query.reportType
+            reportType: this.effectiveReportType
```
wtritch: This was the other issue, also due to the query reportType being a server report type

wtritch:
@esotrope good idea, changed

PR #1433, opened 2018-06-06
Bugfix/iab dashboard group count and more

esotrope:


reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -39,11 +39,9 @@ sql:
       SELECT
         tsg.id,
         tsg.name,
-        s.code as subject_code,
-        count(tsgm.student_id) as student_count
```
esotrope: turns out we don't need this in the ui when doing a findAll on user groups.
yw @agorina

webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -41,17 +42,16 @@ export class GroupDashboardComponent implements OnInit {
   ngOnInit() {
     const { groupId, userGroupId, schoolYear } = this.route.snapshot.params;
     forkJoin(
+      groupId != null
```
esotrope: adds back the call to get a singular group so we have the student count in hand. ty for the catch @jtreuting

webapp/src/main/webapp/src/app/groups/group-table.component.html
```diff
@@ -30,7 +30,7 @@
 
         <div *ngSwitchCase="'group'"
              class="label-max-width">
-          <a [routerLink]="['group-exams', {groupId: group.id}]"
+          <a [routerLink]="['/', 'group-exams', {groupId: group.id}]"
```
esotrope: first attempt at resolving 1901. we'll see how it goes

reporting-service/src/main/resources/application-no-permission.sql.yml
agorina: 👍

PR #1435, opened 2018-06-06
Fix WER item display

wtritch:


webapp/src/main/webapp/src/app/groups/results/group-assessment.provider.ts
```diff
@@ -16,9 +16,9 @@ export class GroupAssessmentProvider implements AssessmentProvider {
               private stateProvider: StateProvider) {
   }
 
-  getAssessmentItems(assessmentId: number, itemTypes?: string[]): Observable<AssessmentItem[]> {
+  getAssessmentItems(assessmentId: number, types?: string[]): Observable<AssessmentItem[]> {
     const { group, schoolYear } = this.stateProvider;
-    return this.service.getAssessmentItems(this.addGroup({schoolYear, assessmentId, itemTypes}, group));
+    return this.service.getAssessmentItems(this.addGroup({schoolYear, assessmentId, types}, group));
```
wtritch: Small typo, the query parameter in the argument resolver is named "types" rather than "itemTypes"

PR #1436, opened 2018-06-06
Fix front-end aggregate csv export for Claim and Target reports

wtritch:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
```diff
@@ -98,6 +130,31 @@ export class AggregateReportTableExportService {
...
+      return builder
+        .withColumn(
+          this.translateService.instant('aggregate-report-table.columns.target'),
+          (item: AggregateReportItem) => {
+            return item.targetCode
+              ? item.targetCode
+              : this.translateService.instant('common.unknown') + ' (' + item.targetNaturalId + ')';
+          }
+        )
+    }
```
wtritch: We were missing target and claim columns here, causing a javascript exception when the button was pushed =)

wtritch:
Merging now, will resolve comments as they arrive

PR #1440, opened 2018-06-06
support LEP for groups showing target reports

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -121,6 +113,24 @@ export class TargetReportComponent implements OnInit, ExportResults {
               private filterOptionService: ExamFilterOptionsService,
               private assessmentService: GroupAssessmentService,
               private applicationSettingsService: ApplicationSettingsService) {
+
+    applicationSettingsService.getSettings().subscribe(settings => {
+      this.allSubgroups = [
```
esotrope: just a code style nit, but i would create an array and build it then assign it to the field.
like:
```
const subgroups = [];
//...
this.allSubgroups = subgroups;
```
easiest way to do this is to tuck it into a method like `this.createAllSubgroups(settings)`

PR #1443, opened 2018-06-06
Preserve value display preferences when switching between general and…

wtritch:
… cohort longitudinal displays

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
@@ -280,6 +280,15 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
     const rowMapper: (query, assessmentDefinition, row, index) => AggregateReportItem =
       (query, assessmentDefinition, row, index) => this.itemMapper.createRow(query, assessmentDefinition, row, index, measuresGetter);
 
+    // Preserve existing display type choices if they exist
+    const displayBySubject: Map<string, any> = (this.reportViews || []).reduce((map, view) => {
+      map.set(view.subjectCode, {
+        valueDisplayType: view.valueDisplayType,
+        performanceLevelDisplayType: view.performanceLevelDisplayType
+      });
+      return map;
+    }, new Map());
+
     this.reportViews = rows.reduce((views, row, index) => {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -280,6 +280,15 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
...
       (query, assessmentDefinition, row, index) => this.itemMapper.createRow(query, assessmentDefinition, row, index, measuresGetter);
 
+    // Preserve existing display type choices if they exist
+    const displayBySubject: Map<string, any> = (this.reportViews || []).reduce((map, view) => {
+      map.set(view.subjectCode, {
+        valueDisplayType: view.valueDisplayType,
+        performanceLevelDisplayType: view.performanceLevelDisplayType
+      });
+      return map;
+    }, new Map());
```
wtritch: We regenerate the "views" that back the aggregate tables when switching between general and cohort populations in Longitudinal reports.  This just preserves the currently-selected value display types when re-generating the views.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -289,6 +298,11 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
         : this.report.request.query.columnOrder;
 
       if (!view) {
+        const displayTypes = displayBySubject.get(subjectCode) || {
+          valueDisplayType: this.query.valueDisplayType,
+          performanceLevelDisplayType: this.query.achievementLevelDisplayType,
+        };
```
wtritch: Use either the UI-preserved or query-default values for the display types for this view.

esotrope: just curious, can this all be avoided by using `[hidden]` instead of `*ngIf` in the template? this screams to me that we accidentally used `*ngIf`

wtritch: I'm afraid not, because we're re-generating rather than modifying views.  And since we're binding to the "view" instance, I don't believe the display table would properly handle modifications of deeper properties.

PR #1445, opened 2018-06-06
Prevent modifying options school year list during display.

wtritch:
The default school year was 2018 BEFORE display, and 2015 AFTER display, causing the hard-to-reproduce bug.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
@@ -68,7 +68,7 @@ export class AggregateReportOptionsMapper {
           value => translate(`common.administration-condition.${value}`),
           value => `Manner of Administration: ${value}`
         )),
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
```diff
@@ -68,7 +68,7 @@ export class AggregateReportOptionsMapper {
           value => translate(`common.administration-condition.${value}`),
           value => `Manner of Administration: ${value}`
         )),
-      schoolYears: options.schoolYears.sort()
+      schoolYears: options.schoolYears.concat().sort()
```
wtritch: This was remarkably difficult to track down...

esotrope: yeah i had another bug that was hard to track before and wound up being this very issue. i blame js for not returning a sorted copy ; j

PR #1446, opened 2018-06-07
Bugfix/fix claim report

wtritch:
The claim report had some issues around initializing selection, initializing with a single subject, and changing selected assessment type.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ClaimAggregateReportQuery.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ClaimAggregateReportQuery.java
@@ -1,6 +1,6 @@
 package org.opentestsystem.rdw.reporting.common.model;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 
@@ -13,6 +13,7 @@ import static org.opentestsystem.rdw.reporting.common.util.CollectionUtils.safeI
  * This class represents a query for the aggregate report with {@link ClaimRow}s
  */
 @SuppressWarnings("unchecked")
+@JsonIgnoreProperties(value="subjectCodes", allowGetters = true)
 public class ClaimAggregateReportQuery extends AbstractAggregateReportQuery {
     private ImmutableSet<Integer> schoolYears;
     private ImmutableMap<String, Set<String>> claimCodesBySubject;
@@ -37,7 +38,6 @@ public class ClaimAggregateReportQuery extends AbstractAggregateReportQuery {
     }
 
     @Override
-    @JsonIgnore
     public Set<String> getSubjectCodes() {
         return getClaimCodesBySubject().keySet();
     }
--- a/webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
@@ -17,7 +17,6 @@ import { MultiOrganizationQueryFormComponent } from './multi-organization-query-
 import { fileName, notEmpty } from '../../shared/form/validators';
 import { Utils } from '../../shared/support/support';
 import { AggregateReportFormOptions } from '../aggregate-report-form-options';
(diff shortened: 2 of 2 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ClaimAggregateReportQuery.java
```diff
@@ -13,6 +13,7 @@
  * This class represents a query for the aggregate report with {@link ClaimRow}s
  */
 @SuppressWarnings("unchecked")
+@JsonIgnoreProperties(value="subjectCodes", allowGetters = true)
```
wtritch: We need to return the "subjectCodes" property to the front-end in order to initialize the `settings.subjects` property.  But we need to tell Jackson to not attempt to set values in a map keyset.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
```diff
@@ -63,14 +62,6 @@ export class ClaimReportFormComponent extends MultiOrganizationQueryFormComponen
 
     this.assessmentDefinition = this.assessmentDefinitionService.get(this.settings.assessmentType, this.settings.reportType);
 
-    this.filterClaimCodes();
-    if (this.settings.claimReport.claimCodesBySubject.length === 0) {
-      // selected All
-      this.filteredOptions.claimCodes.forEach(claim => {
-        this.settings.claimReport.claimCodesBySubject.push(claim.value);
-      });
-    }
```
wtritch: This was unused.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
```diff
@@ -140,11 +130,11 @@ export class ClaimReportFormComponent extends MultiOrganizationQueryFormComponen
 
     this.updateColumnOrder();
     this.markOrganizationsControlTouched();
-    this.onSettingsChange();
+    this.initializeClaimsForAssessmentType();
+    this.onSubjectsChange();
```
wtritch: When we change assessment types we need to update our claim options and selection to bind to the claims for the new assessment type.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
```diff
@@ -176,24 +166,38 @@ export class ClaimReportFormComponent extends MultiOrganizationQueryFormComponen
...
 
-  private initializeClaimsBySubject(): void {
+  private initializeClaimsForAssessmentType(): void {
     this.filteredOptions.subjects.forEach(subject => {
       const subjectCode = subject.value;
-      this.claimsBySubject[ subjectCode ] = this.filteredOptions.claimCodes.filter(claim => claim.value.subject === subjectCode && claim.value.assessmentType === this.settings.assessmentType);
+      this.claimsBySubject[ subjectCode ] = this.filteredOptions.claimCodes
+          .filter(claim => claim.value.subject === subjectCode
+            && claim.value.assessmentType === this.settings.assessmentType);
     });
```
wtritch: This just initializes the claim options for each subject as before based upon the current assessment type.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
```diff
@@ -176,24 +166,38 @@ export class ClaimReportFormComponent extends MultiOrganizationQueryFormComponen
...
+    for (let subject of this.settings.subjects) {
+      if (selections.has(subject)) {
+        // Initialize selection based on settings values
+        this.selectionBySubject[ subject ] = selections.get(subject);
+      } else {
+        // Set selection to 'All'
+        this.selectionBySubject[ subject ] = this.claimsBySubject[ subject ]
+          .map(option => option.value);
+      }
+    }
```
wtritch: This is the big change.  At the end of this method we want our selection for each subject to be either:
1. The `this.settings.claimReport.claimCodesBySubject` Claim values that apply to the subject
2. All available Claim values for the subject

Before we were interpreting both "all" and "none" selections to mean "all" but that caused some nasty logic issues when we create a combined Claim list to set as `this.settings.claimReport.claimCodesBySubject`  Now only "all" means "all" and "none" is never a selection state.

PR #1447, opened 2018-06-07
removed subject restriction on the permission check used by the searc…

jtreuting:
…h students SQL

Still have not tested locally but check it out

### DIFF
--- a/reporting-service/src/main/resources/application.sql.yml
+++ b/reporting-service/src/main/resources/application.sql.yml
@@ -34,7 +34,6 @@ sql:
         JOIN student st on st.id = e.student_id
         JOIN school s ON s.id = e.school_id
         JOIN school isch ON st.inferred_school_id = isch.id
-        JOIN asmt a ON a.id = e.asmt_id
       WHERE e2.student_id IS NULL
         AND (
           :group_id IS NULL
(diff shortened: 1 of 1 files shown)


reporting-service/src/main/resources/application.sql.yml
```diff
@@ -335,3 +335,43 @@ sql:
            OR 1 = :state_embargo_admin
            OR s.district_id IN (:district_embargo_admin_ids)
         )
+
+      studentSearchPermissions: >-
```
jtreuting: combined the `sql.snipper.studentExamPermissionWIthoutEmbargo` and the check above, but removed the restriction on subject that was part of the studentExamPermissionWIthoutEmbargo check

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -335,3 +335,43 @@ sql:
...
+            )
+          )
+          OR
+          # Group permissions
+          (
+            EXISTS(
+               SELECT 1 FROM student_group_membership gm
+                     JOIN student_group sg ON sg.id = gm.student_group_id
+                     JOIN user_student_group usg ON usg.student_group_id = gm.student_group_id
+                        WHERE gm.student_id = e.student_id AND usg.user_login = :user_login
```
jtreuting: here is where there was also a check on the subject

agorina: 👍

agorina:
LGTM. I am assuming it passed the ITs?

PR #1449, opened 2018-06-07
fixes ordering of claims on claim report table

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -294,12 +289,14 @@ export class AggregateReportTableComponent {
     this._orderingByColumnField[ 'schoolYear' ] = SchoolYearOrdering;
     this._orderingByColumnField[ 'claimCode' ] = reportType === AggregateReportType.Target
       ? OrganizationalClaimOrderingProvider(rows[0].subjectCode)
-      : ScorableClaimOrdering;
+      : ordering(ranking(options.claims.map(claim => claim.code))).on<AggregateReportItem>(item => item.claimCode);
```
esotrope: now sorts on the natural order of the claims provided by the option service.
I didn't bother to break the ranking up into subjects because we never show the subjects together anyway.

PR #1450, opened 2018-06-07
implemented sorting on target reports results view for groups

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -199,6 +215,42 @@ export class TargetReportComponent implements OnInit, ExportResults {
...
+        : event.data.sort(claimOrdering.on<AggregateTargetScoreRow>(row => row.claim).reverse().compare);
+    } else if (event.field === 'target') {
+      ascending ? event.data.sort(ordering(byString).on<AggregateTargetScoreRow>(row => row.target).compare)
+        : event.data.sort(ordering(byString).on<AggregateTargetScoreRow>(row => row.target).reverse().compare);
+    } else if (event.field === 'subgroup') {
+      // to keep Overall at the top we shift -> sort -> unshift
+      const overall = event.data.shift();
+      ascending ? event.data.sort(SubgroupOrdering.on<AggregateTargetScoreRow>(row => row.subgroup).compare)
+        : event.data.sort(SubgroupOrdering.on<AggregateTargetScoreRow>(row => row.subgroup).reverse().compare);
+      event.data.unshift(overall);
```
esotrope: why do we need to shift/unshift?

dsnyder-fairway: keeps Overall at the top with all the other subgroups sorted below

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -199,6 +215,42 @@ export class TargetReportComponent implements OnInit, ExportResults {
...
   }
 
+  /**
+   * Sort the data
+   *
+   * @param event {{order: number, field: string}} An optional sort event
+   */
+  public sort(event?: SortEvent): void {
+    const ascending = event.order > 0;
+    if (event.field === 'claimOrder') {
```
esotrope: nit: would add this at the top to remove all the redundant dereferencing
```
const { field, data } = event;
```

esotrope: also a good candidate for a switch statement/ map based lookup like we use the aggregate table.

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
wtritch: Oh wow... this is really hard to wrap my head around... Is there a better way to break all this up?

dsnyder-fairway: changes are coming

esotrope: sounds ominous 8 l

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -199,6 +215,43 @@ export class TargetReportComponent implements OnInit, ExportResults {
...
+   * @param event {{order: number, field: string}} An optional sort event
+   */
+  public sort(event?: SortEvent): void {
+    const { field, data } = event;
+    const ascending = event.order > 0;
+    if (field === 'claimOrder') {
+      // when there isn't a subject specific claim ordering, then default ot a simple alpha sort
+      const claimOrdering: Ordering<string> = (SubjectClaimOrderings.get(this.assessment.subject) || ordering(byString));
+      ascending ? data.sort(claimOrdering.on<AggregateTargetScoreRow>(row => row.claim).compare)
+        : data.sort(claimOrdering.on<AggregateTargetScoreRow>(row => row.claim).reverse().compare);
```
wtritch: You could simplify all this with something like:
```
let columnOrdering: Ordering<AggregateTargetScoreRow>;
if (field === 'claimOrder') {
  columnOrdering = (SubjectClaimOrderings.get(this.assessment.subject) || ordering(byString))
    .on(row => row.claim);
} else if (blah) {
  ...
}
if (event.order < 0) {
  columnOrdering = columnOrdering.reverse();
}
data.sort(columnOrdering);
```

dsnyder-fairway: yeah I more or less did this in my latest changes

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -199,6 +215,48 @@ export class TargetReportComponent implements OnInit, ExportResults {
...
+   * @param event {{order: number, field: string}} An optional sort event
+   */
+  public sort(event?: SortEvent): void {
+    const { field, data } = event;
+    const ascending = event.order > 0;
+    const order = this.getOrdering(field);
+
+    if (field === 'subgroup') {
+      // to keep Overall at the top we shift -> sort -> unshift
+      const overall = data.shift();
```
wtritch: You can do this using a combined Ordering that sorts overall via ranking then falls through to sorting alphabetically.  You can't assume that "overall" will always be the first value when coming into this method (e.g. if they sorted by a different field just prior, overall may be at any position.)

dsnyder-fairway:
@esotrope I think you are right. I was confused about how subgroups are ordered. I'll remove the overall shift / unshift

PR #1452, opened 2018-06-07
fixes report type sort and breaks up report view into two components

esotrope:


webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -4,96 +4,14 @@ <h2>{{'reports.heading' | translate}}</h2>
...
-      <tr>
-        <td [attr.colspan]="columns.length">
-          {{'common.no-results' | translate}}
-        </td>
-      </tr>
-    </ng-template>
-  </p-table>
+  <ng-container *ngIf="reports.length; else noReports">
+    <p>{{ 'reports.instructions' | translate }}</p>
+    <report-table [reports]="reports"></report-table>
```
wtritch: I wonder if `my-reports-table` might be more tightly scoped?  `reports` is just a very generic component/feature name =/  (Sorry, I'm pretty sure I'm the one that named it once-upon-a-time...)

esotrope: yeah.. not concerned with that atm. but if we do a rename i would opt for "UserReport" to follow the convention of user-created / user-specific things being prefixed with "user"

PR #1453, opened 2018-06-07
Correctly sort by target performance level in the target aggregate re…

wtritch:
…sults table.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -19,6 +19,7 @@ import { BaseColumn } from '../../shared/datatable/base-column.model';
 import { byNumericString, SubjectClaimOrderings } from "../../shared/ordering/orderings";
 import { IdentityColumnOptions } from '../assessment/assessment-definition.service';
 import { AggregateReportType } from "../aggregate-report-form-settings";
+import { byTargetReportingLevel } from "../../assessments/model/aggregate-target-score-row.model";
 
 export const SupportedRowCount = 10000;
 export const DefaultRowsPerPageOptions = [ 100, 500, 1000 ];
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/model/aggregate-target-score-row.model.ts
```diff
@@ -13,12 +13,12 @@ export class AggregateTargetScoreRow {
...
-  Below,
-  InsufficientData,
-  Excluded,
-  NoResults
+  Above = 'Above',
+  Near = 'Near',
+  Below = 'Below',
+  InsufficientData = 'InsufficientData',
+  Excluded = 'Excluded',
+  NoResults = 'NoResults'
```
wtritch: In aggregate-target land these are returned from the server as string values.  I just changed the enum to respect that rather than perform a mapping from string to enum/int value.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -295,6 +296,10 @@ export class AggregateReportTableComponent {
       : ordering(ranking(options.claims.map(claim => claim.code))).on<AggregateReportItem>(item => item.claimCode);
     this._orderingByColumnField[ 'subgroup.id' ] = subgroupOrdering(item => item.subgroup, options);
     this._orderingByColumnField[ 'targetNaturalId' ] = TargetOrdering;
+    this._orderingByColumnField[ 'studentRelativeResidualScoresLevel' ] = ordering(byTargetReportingLevel)
+      .on(item => item.studentRelativeResidualScoresLevel);
+    this._orderingByColumnField[ 'standardMetRelativeResidualLevel' ] = ordering(byTargetReportingLevel)
+      .on(item => item.standardMetRelativeResidualLevel);
```
wtritch: By default we sort columns without explicit ordering `byNumber` (for student count, achievement level, etc)
We need to sort these fields by explicit target reporting level.

PR #1455, opened 2018-06-07
Initialize default aggregate settings to copies of option information.

wtritch:
@jtreuting  this may correct some hard-to-reproduce bugs

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
@@ -160,7 +160,7 @@ export class AggregateReportOptionsMapper {
     const assessmentDefinition = this.assessmentDefinitionService.get(defaultAssessmentType, defaultReportType);
     return of(<AggregateReportFormSettings>{
       assessmentType: defaultAssessmentType,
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
```diff
@@ -188,21 +188,21 @@ export class AggregateReportOptionsMapper {
...
-        migrantStatuses: options.studentFilters.migrantStatuses,
-        section504s: options.studentFilters.section504s
+        economicDisadvantages: options.studentFilters.economicDisadvantages.concat(),
+        ethnicities: options.studentFilters.ethnicities.concat(),
+        genders: options.studentFilters.genders.concat(),
+        individualEducationPlans: options.studentFilters.individualEducationPlans.concat(),
+        limitedEnglishProficiencies: options.studentFilters.limitedEnglishProficiencies.concat(),
+        englishLanguageAcquisitionStatuses: options.studentFilters.englishLanguageAcquisitionStatuses.concat(),
+        migrantStatuses: options.studentFilters.migrantStatuses.concat(),
+        section504s: options.studentFilters.section504s.concat()
```
wtritch: When creating our default settings object, we want to initialize it with a copy of the available options rather than pass a reference.

PR #1457, opened 2018-06-07
display message when users do not have permissions to view page conte…

dsnyder-fairway:
…nt after login

someone lacking the permissions will see this
screen shot 2018-06-07 at 3 18 11 pm

@jtreuting we probably want to change this message now that Custom Aggregate Report is an option

### DIFF
--- a/webapp/src/main/webapp/src/app/home/home.component.html
+++ b/webapp/src/main/webapp/src/app/home/home.component.html
@@ -73,6 +73,6 @@
 
 <ng-template #noPermissions>
   <div class="well">
-    <label>{{'home.no-permissions' | translate}}</label>
+    <label [innerHTML]="'home.no-permissions' | translate"></label>
   </div>
 </ng-template>
--- a/webapp/src/main/webapp/src/app/home/home.component.html
+++ b/webapp/src/main/webapp/src/app/home/home.component.html
@@ -29,19 +29,21 @@
   <!-- Left Column Content Area -->
   <div class="col-md-9">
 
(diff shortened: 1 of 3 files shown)


jtreuting:
I'm guessing we'll want it to say something about the Custom Aggregate Reports right?  I see that as a button in the top as well.  I'm assuming they do have access to that?

dsnyder-fairway:
@jtreuting 
screen shot 2018-06-07 at 3 32 51 pm

wtritch:
Hm... I'm not sure I'd call out specific buttons or we'll just hit this again with the next button / button renaming.  Maybe something more generic like:
"Depending on your exact roles, you may have access to other system features above."
?

jtreuting:
That's a good point @wtritch .  I like that.

malaffoon:
Same comment about `above`, what if we move things around? Do we have to give them more instructions? Can't we just say, `You do not have permission to view Individual Assessment Results.`

dsnyder-fairway:
@malaffoon I think we just wanted to highlight that they aren't getting some poorly implemented access denied page. We want to show them that they may have other options available to them. If they can't find them, then there is likely something wrong with their login

dsnyder-fairway:
@jtreuting @wtritch @malaffoon good to merge?

jtreuting:
I like the idea of saying that they likely have other features that they can use.  I do think that just saying they don't have access might confuse them and they might not notice the other options.  So I like calling it out.  I get @malaffoon 's point that if we move the buttons they won't be above anymore but can't think of anything better myself.  So i'd say let's include it as is.

PR #1460, opened 2018-06-07
Fix aggregate report actions in My Reports view

wtritch:


webapp/src/main/webapp/src/app/report/report-action.service.ts
```diff
@@ -118,7 +118,7 @@ class DefaultActionProvider implements ActionProvider {
 class AggregateReportActionProvider extends DefaultActionProvider {
 
   public supports(report: Report): boolean {
-    return report.reportType === AggregateReportType
+    return report.reportType.startsWith(AggregateReportType)
```
wtritch: report.reportType now looks like `AggregateReportRequest Target`

PR #1461, opened 2018-06-08
When sorting by aggregate target performance level, no results should…

wtritch:
… always be sorted to the bottom

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -412,14 +412,23 @@ export class AggregateReportTableComponent {
    */
   private getComparator(field: string, order: number): Comparator<AggregateReportItem> {
     const ascending: boolean = order >= 0;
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -412,14 +412,23 @@ export class AggregateReportTableComponent {
...
+  }
+
+  private getNoResultsLastComparator(comparator: Comparator<any>, order: number): Comparator<AggregateReportItem> {
+    const ascending: boolean = order >= 0;
+    const noResultsComparator: Ordering<AggregateReportItem> = ordering(byNumber)
+      .on(item => item.studentsTested === 0 ? 1 : 0);
+    return join(
+      (ascending ? noResultsComparator.compare : noResultsComparator.compare),
+      comparator
+    );
```
wtritch: I broke out the piece that sorts "no results" last regardless of sort order.  As a side benefit, we now sort "no results" rows to the bottom when sorting on *any* column, not just performance columns.

esotrope: neet

PR #1462, opened 2018-06-08
fixes sort order of report table columns

esotrope:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/report/report-table.component.ts
@@ -1,5 +1,6 @@
 import { Component, Input } from '@angular/core';
 import { Report } from './report.model';
+import { TranslateService } from '@ngx-translate/core';
 
 @Component({
   selector: 'report-table',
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/report/report-table.component.ts
```diff
@@ -8,27 +9,67 @@ import { Report } from './report.model';
...
 
-  constructor({ id, field }) {
+  constructor({ id, field, sortField = '' }) {
     this.id = id;
     this.field = field;
+    this.sortField = sortField ? sortField : field;
   }
 }
+
+interface ReportView {
```
esotrope: created this wrapper which holds pre-translated fields to sort on alphabetically

webapp/src/main/webapp/src/app/report/report-table.component.ts
```diff
@@ -8,27 +9,67 @@ import { Report } from './report.model';
...
+      report: report,
+      translatedAssessmentType: translate(report.assessmentTypeCode
+        ? `common.assessment-type.${report.assessmentTypeCode}.short-name`
+        : 'reports.all'
+      ),
+      // Not ideal, we should not have empty/null subject codes for reports,
+      // it should be the subjects the report was created for at the time of creation
+      translatedSubjectCodes: report.subjectCodes.length
+        ? report.subjectCodes.map(code => translate(`common.subject.${code}.name`)).join(', ')
+        : translate('common.subject.ALL.name'),
```
esotrope: right now some reports have `[]` and `['Math', 'ELA']`. And we write out `[]` as `Math, ELA` This will not be correct when/if new subjects enter the system and ideally should reflect the exact subjects the report was made for at the time of creation.

wtritch: I think we should change `common.subject.ALL.name` to just be `All` or `All Subjects`

PR #1463, opened 2018-06-08
avoid flickering popover

dsnyder-fairway:
strangely enough, I can reproduce this error on the ngx-bootstrap site when viewing the mouseover popover when my browser is less than 1350px wide. https://valor-software.com/ngx-bootstrap/#/popover#triggers-custom

### DIFF
--- a/webapp/src/main/webapp/src/app/student/results/student-results.component.html
+++ b/webapp/src/main/webapp/src/app/student/results/student-results.component.html
@@ -16,7 +16,7 @@
     <ng-template #content>
       <ul class="list-unstyled list-inline">
         <li>
-          <button class="btn btn-default btn-sm"
+          <button class="btn btn-default btn-sm ml-md"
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -16,7 +16,7 @@ <h2>{{'common.person-name' | translate:examHistory.student}}</h2> <div class="st
     <ng-template #content>
       <ul class="list-unstyled list-inline">
         <li>
-          <button class="btn btn-default btn-sm"
+          <button class="btn btn-default btn-sm ml-md"
```
esotrope: does this have a different look than before? can you add a screenshot?

dsnyder-fairway: no difference in look

screen shot 2018-06-07 at 5 44 31 pm

PR #1465, opened 2018-06-08
fix embargo admin display Math count not displaying

jtreuting:
This is the simplest fix at the moment.  Since configurable subjects will be changing this I didn't think a bigger change made sense. 

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.html
+++ b/webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.html
@@ -36,7 +36,7 @@
 
         <div *ngSwitchCase="'subject'"
              class="text-right">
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.html
```diff
@@ -36,7 +36,7 @@
 
         <div *ngSwitchCase="'subject'"
              class="text-right">
-          {{embargo.examCountsBySubject[column.code] | number}}
+          {{embargo.examCountsBySubject[column.serverSubjectCode] | number}}
```
esotrope: i take it the server sends back a `Map<Subject, Integer>` instead of `Map<String, Integer>`?

PR #1466, opened 2018-06-08
added missing "All" selection were appropriate on aggregate report su…

dsnyder-fairway:
…mmary


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -25,6 +25,8 @@ const createColumnProvider = (columnCount: number = Number.MAX_VALUE): ColumnPro
   };
 };
 
+const equalSize = Utils.hasEqualLength;
+const inline = values => [ values.join(', ') ];
```
dsnyder-fairway: looks like we could extract these consts

PR #1470, opened 2018-06-08
Bugfix/fix aggregate target binding

wtritch:
We may have targets in two separate claims with the same natural id. (e.g. "3-8")  This PR references targets by claim code as well as target natural id when resolving aggregate report target display text.


wtritch:
The teacher side being correct is how nohemi found the issue with the aggregate display =)

jtreuting:
+1 JT ;)

PR #1471, opened 2018-06-08
Change IAB Dashboard and Student History page icons: blue icon on gra…

jtreuting:
…y background

image


webapp/src/main/webapp/src/styles.less
```diff
@@ -1526,7 +1526,7 @@ wide-radio-group {
   vertical-align: middle;
 
   > svg {
-    width: 100%;
+    width: 96%;
```
esotrope: why not scale proportionally?

webapp/src/main/webapp/src/styles.less
```diff
@@ -1572,12 +1572,19 @@ user-group {
   min-width: 75px;
 }
 
+.icon-wrapper {
+  padding: 5px;
+  background-color: @gray-lightest;
+  border-radius: 6px;
```
esotrope: there should be some vars to rely on here for padding and radius

webapp/src/main/webapp/src/styles.less
jtreuting: the height should be as well, but i forgot to go back and do that.  this was the only way I could get the icon to not overlap the background on the right.

webapp/src/main/webapp/src/styles.less
jtreuting: do all browsers (like IE 11 and MS Edge) respect styling custom elements?

esotrope: hm good q. i know we have quite a few styles done that way already. if not an alternative is adding `host: {class: 'component-name'}` to the `@Component()` to give it a class to bind to. done that before too

PR #1474, opened 2018-06-08
corrected some incorrect mappings

dsnyder-fairway:
this doesn't resolve DWR-1937, but I noticed a couple mistakes that needed correction

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.html
@@ -134,10 +134,10 @@
               </span>
               <span *ngSwitchCase="'Excluded'"
                     class="label label-lg gray-light">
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/TargetRow.java
```diff
@@ -17,11 +17,11 @@ public String getTargetNaturalId() {
     }
 
     public TargetReportingLevel getStandardMetRelativeResidualLevel() {
-        return studentRelativeResidualScoresLevel;
+        return standardMetRelativeResidualLevel;
     }
 
     public TargetReportingLevel getStudentRelativeResidualScoresLevel() {
-        return standardMetRelativeResidualLevel;
+        return studentRelativeResidualScoresLevel;
```
esotrope: good find - somebody made an oopsiee x J

PR #1475, opened 2018-06-09
DWR-1928 Simplify query that returns students for the teacher group and add ITs

agorina:


common/src/main/resources/common.sql.yml
```diff
@@ -124,7 +124,7 @@ sql:
         (
            # Individual permissions
           (
-            (1 = :individual_statewide OR s.district_group_id IN (:individual_district_group_ids) OR s.district_id IN (:individual_district_ids) OR s.school_group_id IN (:individual_school_group_ids) OR e.school_id IN (:individual_school_ids))
+            (1 = :individual_statewide OR s.district_group_id IN (:individual_district_group_ids) OR s.district_id IN (:individual_district_ids) OR s.school_group_id IN (:individual_school_group_ids) OR s.id IN (:individual_school_ids))
             OR
```
agorina: Unrelated, notice that it is inconsistent.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -21,33 +23,42 @@ sql:
...
+                 JOIN user_student_group usg ON usg.student_group_id = gm.student_group_id
+               WHERE gm.student_id = st.id AND usg.user_login = :user_login
+           )
+           AND
+           (
+             (1 = :group_statewide OR sc.district_group_id IN (:group_district_group_ids) OR sc.district_id IN (:group_district_ids) OR sc.school_group_id IN (:group_school_group_ids) OR sc.id IN (:group_school_ids))
+           )
+         )
+        )
+      ORDER BY st.last_or_surname, st.first_name, st.ssid;
```
agorina: This query takes 2s 562ms in UAT and about 6 seconds in the DEV. And this is the warmed up time.

I have tried breaking it into two queries - first find the latest exam and then use the ids to pass into this one. A query to find the latest exam took 1s 204ms , and the second one took 649ms. 

Once I added  `
ALTER TABLE exam  ADD INDEX test (student_id, school_id, completed_at);
`, these query improved to 1s 48ms.

Based on @jtreuting 's latest discovery of 24 min queries, if we decided to derive the latest exam outside of the reporting, then the index is not really needed. 

Let's talk on Monday.

reporting-service/src/test/resources/integration-test-data.sql
```diff
@@ -38,14 +38,14 @@ insert into school_year (year) values
   (1998);
 
 insert into student (id, ssid, last_or_surname, first_name, gender_id, gender_code, birthday, inferred_school_id, update_import_id, updated, migrate_id) values
-  (-1, 'student1_ssid', 'student1_lastName', 'student1_firstName', -1, 'g1', '1997-01-01 00:00:00.000000', -10, -1, '1997-07-18 20:14:34.000000', -1),
-  (-2, 'student2_ssid', 'student2_lastName', 'student2_firstName', -1, 'g1', '1997-01-01 00:00:00.000000', -20, -1, '1997-07-18 20:14:34.000000', -1),
-  (-3, 'student3_ssid', 'student3_lastName', 'student3_firstName', -1, 'g1', '1997-01-01 00:00:00.000000', -10, -1, '1997-07-18 20:14:34.000000', -1),
+  (-1, 'student1_ssid', 'student1_lastName', 'student1_firstName', -1, 'g1', '1997-01-01 00:00:00.000000', -30, -1, '1997-07-18 20:14:34.000000', -1),
+  (-2, 'student2_ssid', 'student2_lastName', 'student2_firstName', -1, 'g1', '1997-01-01 00:00:00.000000', -30, -1, '1997-07-18 20:14:34.000000', -1),
+  (-3, 'student3_ssid', 'student3_lastName', 'student3_firstName', -1, 'g1', '1997-01-01 00:00:00.000000', null, -1, '1997-07-18 20:14:34.000000', -1),
```
agorina: The inferred school was not set up correctly. Now I think I may have broken some other test. Will check it.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
```diff
@@ -95,7 +157,16 @@ public void findAllShouldApplySchoolFilterWhenSet() {
     }
 
     @Test
-    public void findAllShouldApplyGroupFilterWhenSet() {
+    public void itShouldFindByStudentGroupIgnoringGroupSubject() {
+        //verify that the test data is set up as expected: the group's subject does not match student's latest exam subject
+        assertThat(template.queryForObject("SELECT subject_id FROM student_group WHERE id = -10", Integer.class)).isEqualTo(2);
```
agorina: This tests the issue we found in UAT.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
```diff
@@ -73,18 +89,64 @@
...
         );
         assertThat(actual)
                 .usingRecursiveFieldByFieldElementComparator()
                 .containsExactlyElementsOf(AllStudentsWithExams);
     }
 
     @Test
-    public void findAllShouldApplySchoolFilterWhenSet() {
+    public void isShouldFindExamsForUserWithPIIPermission() {
+
```
agorina: Since we are not using a common permission snippet, I added these tests.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
```diff
@@ -94,8 +156,17 @@ public void findAllShouldApplySchoolFilterWhenSet() {
                 .containsExactlyInAnyOrder(student4);
     }
 
+    @Sql(scripts = {"classpath:integration-test-data.sql"}, statements = {"update student_group set subject_id = 2 where id = -10"})
```
agorina: Reset test data to set up a use case we ran into in UAT.

reporting-service/src/main/resources/application.sql.yml
jtreuting: Yeah sounds like even more evidence to derive the latest exam.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
```diff
@@ -73,18 +89,64 @@
...
+
+        Assertions.assertThat(repository.findAll(userBuilder.permissionsById(permissions(individualOf(districts(-20L)))).build(), emptySearch).size()).isEqualTo(2);
+        Assertions.assertThat(repository.findAll(userBuilder.permissionsById(permissions(individualOf(districts(-99L)))).build(), emptySearch)).isEmpty();
+
+        Assertions.assertThat(repository.findAll(userBuilder.permissionsById(permissions(individualOf(districtGroups(-40L)))).build(), emptySearch).size()).isEqualTo(2);
+        Assertions.assertThat(repository.findAll(userBuilder.permissionsById(permissions(individualOf(districtGroups(-99L)))).build(), emptySearch)).isEmpty();
+    }
+
+    @Test
+    public void isShouldFindExamsForUserWithGroupPermission() {
```
jtreuting: this should be `itShouldFind...` instead of `isShouldFind...`

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
jtreuting: this should be `itShouldFind...` instead of `isShouldFind...`

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
```diff
@@ -106,14 +177,24 @@ public void findAllShouldApplyGroupFilterWhenSet() {
...
+
+        assertThat(repository.findAll(
+                userBuilder.permissionsById(permissions(groupOf(schools(-10L)))).build(), newArrayList(-1L, -2L, -3L, -4L)))
                 .usingRecursiveFieldByFieldElementComparator()
-                .containsExactlyInAnyOrder(student1, student2);
+                .containsExactlyInAnyOrder(student4);
     }
 
+    @Test
+    public void isShouldFindNothingForUserWithoutRequiredPermission() {
```
jtreuting: same

reporting-service/src/main/resources/application.sql.yml
agorina: Interesting, I played with that 24 min query, and the latest exam is not the offender. 
Replacing EXISTS with the JOIN changes the response time to 600ms.

This does however introduce a challenge with having one query for both `student_group` and `teacher_student_group`.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepositoryIT.java
agorina: thanks, will fix...everywhere.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupService.java
```diff
@@ -36,12 +35,7 @@ public Group getGroup(final User user, final long groupId) {
 
         // Resolve group student IDs to student objects
         // Filter out students the user no longer has access to
-        final Set<Student> students = ImmutableSet.copyOf(
-                studentRepository.findAll(user, StudentSearch.builder()
-                        .studentIds(group.getStudentIds())
-                        .build()
-                )
-        );
+        final Set<Student> students = ImmutableSet.copyOf(studentRepository.findAll(user, group.getStudentIds()));
```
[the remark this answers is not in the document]
malaffoon: 👍

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -21,33 +23,42 @@ sql:
...
-      WHERE e2.student_id IS NULL
-        AND (
-          :group_id IS NULL
-          OR EXISTS (
-            SELECT 1 FROM student_group_membership sgm
-              JOIN student_group sg ON sg.id = sgm.student_group_id
-            WHERE sgm.student_id = st.id
-              AND sg.id = :group_id
+      FROM student st
+        JOIN school sc ON st.inferred_school_id = sc.id
```
malaffoon: If a student has no exams, the inferred school will not be set and the student will be excluded. Is that okay? 

Use case: school admin loads rosters and teachers are tweaking groups early in the school year before any exams are submitted.

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupServiceTest.java
```diff
@@ -53,7 +53,7 @@
     public void before() {
         when(user.getUsername()).thenReturn("username");
         when(repository.findOne(any(User.class), eq(1L))).thenReturn(group1.build());
-        when(studentRepository.findAll(any(User.class), any(StudentSearch.class))).thenReturn(ImmutableList.of(
+        when(studentRepository.findAll(any(User.class), any(Collection.class))).thenReturn(ImmutableList.of(
```
malaffoon: i thought the new matchers didn't need this specificity?

reporting-service/src/test/resources/integration-test-data.sql
malaffoon: Unrelated but why is the birthday including the time, can't it just be a date?

reporting-service/src/main/resources/application.sql.yml
agorina: I believe we have talked about this use case and decided that it is okay since we want to enforce the permissions (https://github.com/SmarterApp/RDW_Reporting/pull/1284). 

I am open to reconsider it?

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/group/DefaultUserGroupServiceTest.java
agorina: Maybe, I can check it out.

reporting-service/src/test/resources/integration-test-data.sql
agorina: Since it is `date` in the schema, we should be able to use just dates. I will clean it up.

reporting-service/src/main/resources/application.sql.yml
agorina: I have tested this query on `awsdev`. 

The group version of it has the same issue with `EXISTS`:  `EXISTS` takes 30 seconds, while the `JOIN` returns right away.

The school version benefits from the index mentioned above, but it is still slow for a school with a large number of exams:
- a school with 111,850 exams :  6s 205ms with index, 15s 964ms without
- a school with 87,149, exams 1708: 3s 641ms vs 9s 483ms
- smaller numbers where close what I saw in UAT.

Not sure if this justifies deriving the latest exam outside of this query or if we should try to use the index for now and consider the other approach as 'future enhancement'. 

I have also tested `JOIN` version on the smaller db (UAT) and it performs fine. So I am going to make this change everywhere. Probably in a separate PR.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/student/JdbcStudentRepository.java
```diff
@@ -49,10 +61,10 @@
                 findAllQuery,
                 new MapSqlParameterSource()
                         .addValues(securityParameterProvider.getSecurityParameters(permissionSource))
-                        .addValue("group_id", search.getGroupId())
-                        .addValue("school_id", search.getSchoolId())
-                        .addValue("search_student_ids", search.getStudentIds() != null)
-                        .addValue("student_ids", nullOrEmptyToDefault(search.getStudentIds(), null)),
+                        .addValue("group_id", search == null ? null : search.getGroupId())
+                        .addValue("school_id", search == null ? null : search.getSchoolId())
+                        .addValue("search_student_ids", !isEmpty(studentIdsToSearch))
```
esotrope: if we are going to split the methods i think we should make the impl for findAll(by search) and findAll(by student IDs) separate. here we are rechecking for something we already know (search == null)

agorina: Agree, was lazy. Will change.

agorina:
I am short of an approved review, but I would like to merge to move on with the SQL changes. Will address any new comments if needed, so please post them.

reporting-service/src/main/resources/application.sql.yml
malaffoon: i thought that was when we were sharing the security snippet and didn't want to mess with things. Seems like, if the teacher could see a student because of assigned group membership, they should be able to see the student whether they have exams or not.

Of course, it is June 10 so, if it takes more than an hour or two to figure out and write tests for, i wouldn't do it.

agorina: I do not think it will take too long to implement from the API point of view, but there are two business concerns here:
 1. it would remove *ANY* security check from the API that creates the teacher group. 
2. it will return students without a name and all other attributes.

malaffoon: Perhaps we can talk about it Monday. It seem like when filtering on PII (school-based) permissions you should join the inferred school but when filtering on PII_GROUP permissions you should not.

agorina: Ok, let's talk.

PR #1477, opened 2018-06-09
accessibility skip link

jtreuting:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1085,18 +1085,19 @@
     "view-dashboard": "View IAB Dashboard"
   },
   "home": {
+    "groups-well": {
+      "assigned-groups-heading": "Assigned Groups",
+      "created-groups-heading": "My Groups"
+    },
+    "no-permissions": "You do not have permissions to view Assessment Results.<p><p>Depending on your exact roles, you may have access to other system features above.",
```
jtreuting: alphabetized these since i was in the section

PR #1479, opened 2018-06-10
Accessibility updates adding for attribute and fieldset

jtreuting:
From the WCAG analysis report:

> Provide a descriptive text label for each field. Code the HTML <label> tag so the "for" attribute matches the "id" attribute of the field.
>
> For cases such as the radio buttons on the “Student: Print ICA Report” page, we recommend that the radio buttons and labels should also be enclosed within a `fieldset` tag.

This PR adds the `for` attribute to labels where they were missing.  It also adds a fieldset around the label and radio groups and checkbox groups.  

I checked that the fieldset CSS is set to not do anything (no border, padding, margin, etc.) Checked in Chrome and FF and it looks fine there.


webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -13,11 +13,14 @@ <h3>{{'aggregate-report-form.section.organization.heading' | translate}}</h3>
 
           <!-- Typeahead -->
           <div *ngIf="!hasDefaultSchoolOrganization">
-            <label>{{'aggregate-report-form.field.organization-label' | translate}}</label>
-            <span [hidden]="!organizationTypeahead.loading"><i class="fa fa-spinner fa-pulse"></i></span>
-            <organization-typeahead #organizationTypeahead
-                                    [options]="organizationTypeaheadOptions"
-                                    (selected)="onOrganizationTypeaheadSelect($event)"></organization-typeahead>
+            <fieldset>
```
jtreuting: all of these blocks are going to be adding fieldset around the label and inputs and adding the `for` attribute.  Mostly indenting because of the fieldset wrapping

webapp/src/main/webapp/src/app/aggregate-report/subgroup/subgroup-filters.component.html
```diff
@@ -1,140 +1,168 @@
...
+  <fieldset>
+    <div class="row">
+      <div class="col-md-6">
+        <label for="gender-subgroup-filter"
+               info-button
+               title="{{'aggregate-report-form.field.gender-label' | translate}}"
+               content="{{'aggregate-report-form.field.gender-info' | translate}}"></label>
+      </div>
+      <div class="col-md-6">
+        <sb-checkbox-group id="gender-subgroup-filter"
```
jtreuting: i named these with `subgroup-filter` since they are on the same page as the advanced filters and didn't want conflicts

jtreuting:
I did not test it specifically.  I believe they did during the WCAG compliance audit that generated these items to address.  But I'll run through that and see if any adjustments are needed.

PR #1480, opened 2018-06-10
Add aria-live="polite" to error messages so screen readers are prompt…

jtreuting:
…ed to notify the user

Approach based on this: https://cantina.co/form-errors-screen-readers-can-access/

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-download.component.html
+++ b/webapp/src/main/webapp/src/app/report/report-download.component.html
@@ -32,7 +32,6 @@
                      maxlength="255"
                      [(ngModel)]="options.name"
                      [aria-invalid]="resourceControl.invalid && (resourceControl.dirty || resourceControl.touched) && resourceControl.errors.required"
-                     aria-required="true"
                      required>
             </div>
             <div *ngIf="resourceControl.invalid && (resourceControl.dirty || resourceControl.touched)">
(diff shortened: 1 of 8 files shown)


dsnyder-fairway:
Is this missing from `user-group-form.component.html`and `target-report-form.component.html` as well? This was just seeing what files had `help-block` class in them

jtreuting:
@dsnyder-fairway I'll check.  I was looking for the `.errors` that is used to find the errors, but maybe there are other ways.

webapp/src/main/webapp/src/app/report/report-download.component.html
```diff
@@ -35,6 +35,7 @@ <h4 class="modal-title pull-left">{{title}}</h4>
             </div>
             <div *ngIf="resourceControl.invalid && (resourceControl.dirty || resourceControl.touched)">
               <p *ngIf="resourceControl.errors.required"
+                 aria-live="polite"
                  class="help-block small red">{{'common.form-field-message.required' | translate}}</p>
             </div>
```
wtritch: hm... this is just going to prompt a screen-reader to say "Required" without any indication of which control is required...

jtreuting: good point.  i initially started it with the error messages that had more context.  i'll do a review of these to see where we might need to add the field label to the required error message.

PR #1482, opened 2018-06-10
attempt at fixing grid for ms edge

esotrope:
Applied changes recommended here:
https://stackoverflow.com/questions/45786788/css-grid-layout-not-working-in-edge-and-ie-11-even-with-ms-prefix

It seems only the latest version of edge supports repeat() and ie11 still doesn't?

Anyway, this is an untested shot at a fix so if someone with ie11 could pull the branch and test it in edge that would be awesome 👌

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -1310,6 +1310,7 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
 .card-grid {
   display: grid;
   grid-gap: @spacer;
+  -ms-grid-columns: (@spacer * 21) (@spacer * 21) (@spacer * 21);
   grid-template-columns: repeat(@cards-per-row, @spacer * 21);
 }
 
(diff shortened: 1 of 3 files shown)


jtreuting:
In order for the cells to display in the proper column/row, each cell must have the following styles.  To get the cell in the second row and third column the div has to have `-ms-grid-column: 3; -ms-grid-row: 2;`

I think it might make more sense to move to another approach like bootstrap columns or flexbox.  Maybe for some simpler ones like the Report Type we can just manually add these styles.  But for things like the IAB Dashboard where it was to wrap, it seems like a bot so great solution to have the CSS define that it's 3 columns AND angular know that it's 3 columns and track it to define the column and row on each div.

Here is an example of me playing with the styles in Edge and seeing them start moving into cells.

image

jtreuting:
sounds reasonable

webapp/src/main/webapp/src/styles.less
```diff
@@ -1310,6 +1310,7 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
 .card-grid {
   display: grid;
   grid-gap: @spacer;
+  -ms-grid-columns: (@spacer * 21) (@spacer * 21) (@spacer * 21);
   grid-template-columns: repeat(@cards-per-row, @spacer * 21);
```
wtritch: ...what's "@spacer * 21" represent/mean?

esotrope: *shrug* don't know but you reminded me i need to close this PR : j

esotrope:
closing in favor of PR to convert to bs grid

PR #1483, opened 2018-06-10
An attempt to fix slow queries with the common group filter snippet.

agorina:
For those who did not see Jeff's findings: there are at least two SQLs with the common group filter /snippet that take about 25 min on `awsdev`. The source of the problem seems to be an `EXISTS` clause. 

There are total of 10 SQLs that use this common snippet.

My objective is to avoid two different SQLs based on the type of the student groups. This review is an attempt to try a different version of the common snippets. I have tested it with the 2 found offenders and it works, but I have not tested the other 8 queries. I think it will be faster to use `jmeter` for this. 

And yes, we do use `EXISTS` in our common permission snippet, but for some reason it seems to work there just fine.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/UserGroupId.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/UserGroupId.java
@@ -1,0 +1,37 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * Represents either a teacher group or an admin created user group id
+ */
+public class UserGroupId {
+    private long id;
+    private int type;
+
+    /**
+     * Default constructor for serialization
+     */
+    private UserGroupId() {
+    }
+
+    /**
+     * Constructor
+     *
+     * @param groupId an admin created groups id
+     * @param userGroupId a user created group id
+     */
+    public UserGroupId(final Long groupId, final Long userGroupId) {
+        checkArgument((groupId != null || userGroupId != null) && (groupId != userGroupId), "one and only one group id must be not null");
+        this.id = groupId == null ? userGroupId : groupId;
+        this.type = groupId == null ? 2 : 1;
+    }
+
+    public int getType() {
+        return type;
+    }
+
+    public long getId() {
+        return id;
+    }
+}
--- a/common/src/main/resources/common.sql.yml
+++ b/common/src/main/resources/common.sql.yml
@@ -241,32 +241,23 @@ sql:
       ei.trait_conventions_score as ei_trait_conventions_score,
       i.performance_task_writing_type
 
-    # dependencies: student st, asmt a, :group_id, :user_group_id
+    # dependencies: sg (sql.snippet.selectFromBothGroupTypes), student st, asmt a
     groupFilters: >-
-      (
-        (
-          :group_id IS NULL
-          OR EXISTS(
-            SELECT 1
-            FROM student_group_membership sgm
-              JOIN student_group sg ON sg.id = sgm.student_group_id
-              JOIN user_student_group usg ON usg.student_group_id = sgm.student_group_id
-            WHERE sgm.student_group_id = :group_id
-              AND sgm.student_id = st.id
-              AND usg.user_login = :user_login
-              AND (sg.subject_id is null OR sg.subject_id = a.subject_id)
-          )
-        )
-        AND (
-          :user_group_id IS NULL
-          OR EXISTS(
-            SELECT 1
-            FROM teacher_student_group_membership tsgm
-              JOIN teacher_student_group tsg ON tsg.id = tsgm.teacher_student_group_id
-            WHERE tsgm.teacher_student_group_id = :user_group_id
-              AND tsgm.student_id = st.id
-              AND tsg.user_login = :user_login
-              AND (tsg.subject_id IS NULL OR tsg.subject_id = a.subject_id)
-          )
-        )
+      sg.id = :group_id
+      AND sg.user_login = :user_login
+      AND sg.group_type = :group_type
+      AND sg.student_id = st.id
+      AND (sg.subject_id IS NULL OR sg.subject_id = a.subject_id)
+
+    selectFromBothGroupTypes: >-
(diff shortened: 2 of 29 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/UserGroupId.java
```diff
@@ -0,0 +1,37 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * Represents either a teacher group or an admin created user group id
+ */
+public class UserGroupId {
```
agorina: My understanding is that we never search by both groups, just by one at a time. And we had two ids in each object that needed it. Instead I introduced this one.

common/src/main/resources/common.sql.yml
```diff
@@ -241,32 +241,23 @@ sql:
       ei.trait_conventions_score as ei_trait_conventions_score,
       i.performance_task_writing_type
 
-    # dependencies: student st, asmt a, :group_id, :user_group_id
+    # dependencies: sg (sql.snippet.selectFromBothGroupTypes), student st, asmt a
```
agorina: This is the main change.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -185,26 +189,28 @@ sql:
...
-              FROM exam ex2
-                JOIN asmt a ON a.id = ex2.asmt_id
-                JOIN student st ON st.id = ex2.student_id
-            WHERE ex2.type_id = 2
-                  AND ex2.school_year = :school_year
-                  AND ex2.scale_score IS NOT NULL
-                  AND ex2.scale_score_std_err IS NOT NULL
-                  AND ex2.performance_level IS NOT NULL
+            ( SELECT e.id, e.student_id, e.asmt_id, e.school_year, e.completed_at
+              FROM exam e
```
agorina: Ditto, from `ex2` to `e`.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/GroupExamReportRequest.java
```diff
@@ -1,29 +1,21 @@
...
+    private UserGroupId groupId;
 
-    public GroupGrant getGroupGrant() {
-        return groupGrant;
-    }
-
-    public Long getGroupId() {
+    public UserGroupId getGroupId() {
         // Bridge for existing reports.
         if (groupGrant != null) {
```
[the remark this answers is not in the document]
esotrope: its for adapting old report requests - what you have here looks right 👍

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/UserGroupId.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Constructor
+     *
+     * @param groupId an admin created groups id
+     * @param userGroupId a user created group id
+     */
+    public UserGroupId(final Long groupId, final Long userGroupId) {
+        checkArgument((groupId != null || userGroupId != null) && (groupId != userGroupId), "one and only one group id must be not null");
+        this.id = groupId == null ? userGroupId : groupId;
+        this.type = groupId == null ? 2 : 1;
```
esotrope: i know this is probably all just for poc but if we go this route i suggest we extract these and use a builder instead of a two-param constructor.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/UserGroupId.java
esotrope: neat idea 👍

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/GroupAssessmentSearchArgumentResolver.java
```diff
@@ -26,18 +28,11 @@ public Object resolveArgument(
...
-        }
-
-        return search.build();
+        return search.groupId(new UserGroupId(tryToParseLong(request.getParameter("groupId")), tryToParseLong(request.getParameter("userGroupId")))).build();
     }
 
+    private static Long tryToParseLong(final String value) {
+        if (value == null) return null;
+        return Longs.tryParse(value);
+    }
```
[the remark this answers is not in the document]
esotrope: `opentestsystems.rdw.common.support.Longs2`?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/UserGroupId.java
jtreuting: Do we really need a builder for a 2 parameter object?  I'm up for whatever the consensus is but it seems like if you have 1 or 2 (and maybe 3) parameters using the constructor seems appropriate to me.

agorina: Wasn't the consensus: below 3 parameters - no builder?

Looking at it now, I think it maybe cleaner to define an Enum for the GroupType and change the signature to `public UserGroupId(final Long groupId, final GroupType groupType)`.

But I really would like to try it with jmeter to see if it solves the problem before I make this change.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/exam/group/GroupAssessmentSearchArgumentResolver.java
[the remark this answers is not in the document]
malaffoon: I'm not sure i agree that it should be put in a utility class and i definitely don't think it deserves its own new module. Unless we plan on sweeping through one or both projects and collecting all these.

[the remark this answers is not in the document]
agorina: This is a good discussion, but considering the timeline and where we are I do not think I want to spend time on changing this. Unless somebody has a very strong objection.

PR #1484, opened 2018-06-11
Defect/msedge grid issues

esotrope:
Converts css grids to bootstrap grids on
1. The aggregate form report type control
2. group/student history dashboard
3. group dashboard card chart

### DIFF
--- a/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.html
+++ b/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.html
@@ -26,52 +26,59 @@
 <ng-template #loaded>
   <ng-container *ngIf="group; else groupNotFound">
 
-    <!-- Group Select -->
-    <div class="col-md-3">
-      <label for="select-group">{{'group-results.select-group' | translate}}</label>
-      <!--TODO: type ahead? -->
-      <select id="select-group"
-              class="form-control"
-              [compareWith]="groupEquals"
-              [(ngModel)]="group"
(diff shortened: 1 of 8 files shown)


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.html
```diff
@@ -81,34 +88,32 @@ <h2>{{'group-dashboard.heading' | translate}}</h2>
...
+  </ng-container>
+  <ng-template #hasMeasuredAssessments>
 
-      <div class="card-grid">
-        <div *ngFor="let measuredAssessment of filteredMeasuredAssessments">
-          <group-assessment-card (selectedAssessment)="onCardSelection($event)"
-                                 [group]="group"
-                                 [measuredAssessment]="measuredAssessment"></group-assessment-card>
-        </div>
+    <div class="row" *ngFor="let row of rows">
```
esotrope: initially i did not want to resort to this and just use float-left fixed-width divs but this lead to awkward stacking of cards when they were not equal height. in order to get away with that approach the cards would need to be equal height.

webapp/src/main/webapp/src/styles.less
```diff
@@ -1572,19 +1561,18 @@ user-group {
   min-width: 75px;
 }
 
-.icon-wrapper {
+@assessment-icon-size: 80px;
+
+.assessment-icon {
```
esotrope: fixed unrelated issue with icon borders on resize:

screen shot 2018-06-10 at 11 13 28 pm

PR #1485, opened 2018-06-11
Don't call get latest assessment when assessments are pre-selected wh…

jtreuting:
…en going to the assessment results page

Remove unused call to available measured assessments when leaving the IAB Dashboard page

Addresses: https://jira.fairwaytech.com/browse/DWR-1942

### DIFF
--- a/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
+++ b/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
@@ -125,10 +125,6 @@ export class GroupDashboardComponent implements OnInit {
     this.router.navigate([ 'group-exams', this.stateAsNavigationParameters ]).then(() => {
       // reset selected assessments to avoid issues with going back to previous page
       this.selectedAssessments = [];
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -125,10 +125,6 @@ export class GroupDashboardComponent implements OnInit {
     this.router.navigate([ 'group-exams', this.stateAsNavigationParameters ]).then(() => {
       // reset selected assessments to avoid issues with going back to previous page
       this.selectedAssessments = [];
-      this.groupDashboardService.getAvailableMeasuredAssessments(this.createSearch(this.group))
```
jtreuting: i didn't see a use for this call.  it is called when assessments are selected on the IAB Dashboard and click View Assessments.  Anyone know if i'm missing something?

esotrope: since we cache the state of the dashboard when returning to the page maybe it's to cover the case that the back button is hit? that doesn't make sense though * shrug *

PR #1494, opened 2018-06-12
Remove results test for exam export report.  Report will now return a…

wtritch:
… report CSV containing only a header row if there are no results.

This will result in a header-only CSV download if there are no results.

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/ExamRepository.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/ExamRepository.java
@@ -19,12 +19,4 @@ public interface ExamRepository {
      */
     String export(@NotNull PermissionSource permissionSource, @NotNull ExportExamReportRequest query, @NotNull String uriLocation);
 
-    /**
-     * Determines if the {@link #export(PermissionSource, ExportExamReportRequest, String)} method would contain results.
-     *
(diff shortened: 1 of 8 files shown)


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/ExamRepository.java
```diff
@@ -19,12 +19,4 @@
...
     String export(@NotNull PermissionSource permissionSource, @NotNull ExportExamReportRequest query, @NotNull String uriLocation);
 
-    /**
-     * Determines if the {@link #export(PermissionSource, ExportExamReportRequest, String)} method would contain results.
-     *
-     * @param permissionSource permissions to apply during exam report
-     * @param query            exam report query params
-     * @return true if and only if any exams are found
-     */
-    boolean hasExams(@NotNull PermissionSource permissionSource, @NotNull ExportExamReportRequest query);
```
wtritch: I debated leaving the legacy functionality in there... but if we want to re-introduce it we'd likely do something totally different, like maybe try to return the row count from the `SELECT INTO OUTFILE` statement?  Regardless, I don't like leaving vestigial code and SQL hanging around in a codebase.

malaffoon: 👍

PR #1495, opened 2018-06-12
fixes scroll to top when page filters are changed

esotrope:


webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -86,13 +97,18 @@ export class AppComponent {
...
+
       // if the user is going back, don't do the window scroll
       // let the browser take the user back to the position where they last were
-      if (event.url == this._lastPoppedUrl) {
+      if (event.url === this._lastPoppedUrl) {
         this._lastPoppedUrl = undefined;
-      } else {
+
+      // do not scroll to the top if the route params are the only change to the route
+      } else if (currentUrlWithoutParams !== this._previousUrlWithoutParams) {
```
esotrope: was hoping the `NavigationEnd` or `ActivatedRoute` would provide ways to do this but couldn't find it.

esotrope:
Yes, noticed this when fixing the back button state restoration issue. What this change does is prevent the page from jumping to the top when you change the route-changing form controls like group, school and year. I think I'll close this for now though. I can see us wanting to jump to the top when you click back and forward but not when clicking on the form controls.

PR #1496, opened 2018-06-12
allows old reports without reportType to be correctly handled without…

esotrope:
… errors

Old reports don't have reportType defined

image

### DIFF
--- a/webapp/src/main/webapp/src/app/report/report-action.service.ts
+++ b/webapp/src/main/webapp/src/app/report/report-action.service.ts
@@ -1,9 +1,9 @@
-import { Injectable } from "@angular/core";
-import { Report } from "./report.model";
-import { ReportService } from "./report.service";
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/report/report-action.service.ts
```diff
@@ -118,13 +118,14 @@ class DefaultActionProvider implements ActionProvider {
 class AggregateReportActionProvider extends DefaultActionProvider {
 
   public supports(report: Report): boolean {
-    return report.reportType.startsWith(AggregateReportType)
+    return report.reportType != null
```
esotrope: the change.

when this reportType check fails the default handler will  handle the report correctly, treating it as a non-aggregate report.

jtreuting:
@esotrope merged since you are out so we can QA it

PR #1499, opened 2018-06-12
Fix a bug with the student group sql.

agorina:


common/src/main/java/org/opentestsystem/rdw/reporting/common/sqlbuilder/DefaultUserGroupSqlProvider.java
```diff
@@ -25,12 +25,12 @@ public DefaultUserGroupSqlProvider(@Value("${sql.snippet.adminUserGroups.join}")
...
-        teacherProperties.put("groupFilterPlaceholder", adminUserGroupsFilterSql);
+        teacherProperties.put("groupJoinPlaceholder", teacherGroupsJoinSql);
+        teacherProperties.put("groupFilterPlaceholder", teacherGroupsFilterSql);
 
         adminProperties = new Properties();
-        adminProperties.put("groupJoinPlaceholder", teacherGroupsJoinSql);
-        adminProperties.put("groupFilterPlaceholder", teacherGroupsFilterSql);
+        adminProperties.put("groupJoinPlaceholder", adminUserGroupsJoinSql);
+        adminProperties.put("groupFilterPlaceholder", adminUserGroupsFilterSql);
     }
```
agorina: My guess is the reason it is passing all the ITs is because we use the same ids in the test scripts for both groups. I would like to merge this but then address the tests because they should have caught this.

PR #1501, opened 2018-06-12
make cards tab selectable

dsnyder-fairway:


webapp/src/main/webapp/src/styles.less
```diff
@@ -1318,6 +1318,7 @@ distractor-analysis .ui-widget.ui-table table .ui-table-tbody td.green {
   will-change: transform, border-color;
   transition-duration: @duration-exit;
   transition-timing-function: @easing-decel;
+  width: 100%;
```
dsnyder-fairway: this was necessary as cards with small titles (summatives) shrunk in size.

jtreuting:
can you test across a few different browsers?  for some reason having the card be a button feels like it might cause issues in older browsers, but i might be off on that.  i can test in Edge once we have it in QA

dsnyder-fairway:
@jtreuting 

- Chrome ✅ 
- Firefox ✅ 
- Opera ✅ 
- Safari ❌ (Safari tab navigation seems to be lacking in general from RDW_Reporting. Was only able to select input fields & grade dropdown on main page)

dsnyder-fairway:
@jtreuting yes. The only one that is difficult to see is Firefox as it looks like by default there is no blue selector box around the item when tabbing. You are able to see the indicator in the top right though

PR #1502, opened 2018-06-12
Handle tabbing the sb-button-group

jtreuting:
Right now the sb-button-group's are not navigable via keyboard tabbing.  This changes makes it so each item can be tabbed to and selected by hitting Enter

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
@@ -124,22 +124,25 @@ const ControllerByInputType: {[inpuType: string]: InputController} = {
   selector: 'sb-button-group',
   template: `
     <ng-template #button let-option let-isAllOption="isAllOption">
-      <label class="btn"
-             [ngClass]="computeStylesInternal(buttonStyles, { 
-                 active: isAllOption ? stateInternal.selectedAllOption : stateInternal.selectedOptions.has(option), 
-                 disabled: disabled 
-             })">
+      <a class="btn"
+         href="javascript:void(0)"
+         (click)="isAllOption ? onAllOptionClickInternal() : onOptionClickInternal(option)"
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -124,22 +124,25 @@ const ControllerByInputType: {[inpuType: string]: InputController} = {
...
-                 active: isAllOption ? stateInternal.selectedAllOption : stateInternal.selectedOptions.has(option), 
-                 disabled: disabled 
-             })">
+      <a class="btn"
+         href="javascript:void(0)"
+         (click)="isAllOption ? onAllOptionClickInternal() : onOptionClickInternal(option)"
+         [ngClass]="computeStylesInternal(buttonStyles, { 
+             active: isAllOption ? stateInternal.selectedAllOption : stateInternal.selectedOptions.has(option), 
+             disabled: disabled 
+         })">
```
jtreuting: changed from label to link and moved the click event up to hear from the checkbox

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -124,22 +124,25 @@ const ControllerByInputType: {[inpuType: string]: InputController} = {
...
         <input type="checkbox"
                [attr.checked]="isAllOption ? stateInternal.selectedAllOption : stateInternal.selectedOptions.has(option)"
                [name]="name"
+               notab="notab"
+               tabindex="-1"
```
jtreuting: stack overflow suggested the notab but it didn't work in Chrome (maybe in some browsers) but the tabindex="-1" worked.  basically without this, when you tab to the first item it is highlighted (the link) then you tab again and nothing is highlighted (the checkbox gets the focus) then on the 3rd tab you get the 2nd option selected

jtreuting:
will address any feedback that comes later

PR #1507, opened 2018-06-12
Handle NULL session IDs in the UI

jtreuting:
To support an optional session TRT field, the UI needs to handle having no session value.  Made sure the session toggle buttons are generated correctly and that the session column is populated.

image


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.spec.ts
```diff
@@ -136,6 +136,27 @@ describe('AssessmentResultsComponent', () => {
     expect(component.exams.length).toBe(0);
   });
 
+  it('should handle all null sessions', () => {
```
jtreuting: it was already handling it but added a couple tests around it

PR #1508, opened 2018-06-13
updated TODO message on target reports when we don't have enough results

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.html
```diff
@@ -45,9 +45,8 @@
 
 <!-- Student Results Table -->
 <p-table #dataTable
-         *ngIf="showResults"
          [columns]="columns"
-         [value]="aggregateTargetScoreRows"
+         [value]="showResults ? aggregateTargetScoreRows : []"
```
dsnyder-fairway: if we can't show the results we default to an empty list to trigger our emptymessage

PR #1509, opened 2018-06-13
Do check on whether a user can create a teacher group on not having a…

jtreuting:
…ny assigned groups for teachers.  it was using not having any teacher created groups

### DIFF
--- a/webapp/src/main/webapp/src/app/user-group/user-groups.component.ts
+++ b/webapp/src/main/webapp/src/app/user-group/user-groups.component.ts
@@ -5,6 +5,7 @@ import { FilterOptionsService } from '../shared/filter/filter-options.service';
 import { PermissionService } from '../shared/security/permission.service';
 import { forkJoin } from 'rxjs/observable/forkJoin';
 import { Router } from '@angular/router';
+import { Group } from '../groups/group';
 
 @Component({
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/user-group/user-groups.component.ts
```diff
@@ -37,11 +47,11 @@ export class UserGroupsComponent implements OnInit {
       this.permissionService.getPermissions()
     ).subscribe(([ options, permissions ]) => {
       this.filteredGroups = this.groups.concat();
-      if (this.groups.length) {
+      if (this.groups.length !== 0) {
         this.defaultGroup = this.groups[ 0 ];
       }
       this.subjects = options.subjects;
-      this.createButtonDisabled = this.groups.length === 0
+      this.createButtonDisabled = this.assignedGroups.length === 0
```
jtreuting: this is the fix.  it was checking if there were no teacher created groups so it was blocking the user from creating a new group if they hadn't already created one

esotrope: thx Jeff! wonder if we shouldn't eventually just make createButtonDisabled an `@Input()` and let the group-tabs component determine it

PR #1512, opened 2018-06-13
Make the iframe not tabbable, otherwise the focus gets trapped in the…

jtreuting:
… iframe and can't tab out of it

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
@@ -1,14 +1,4 @@
 <div>
-  <div *ngIf="irisIsLoading"
-       class="loader">
-    <div class="inner">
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-viewer/item-viewer.component.html
```diff
@@ -22,12 +12,24 @@
...
+
+  <div *ngIf="irisIsLoading"
+       class="loader">
+    <div class="inner">
+      <i class="fa fa-spinner fa-pulse fa-3x fa-fw"></i>
+      <span class="text">
+        {{ 'common.messages.loading' | translate }}
+      </span>
+    </div>
+  </div>
```
esotrope: i think this whole block can now be replaced with `<spinner *ngIf="irisIsLoading" text="{{'common.messages.loading' | translate}}"></spinner>`

jtreuting: oh cool, will do. thanks.

PR #1516, opened 2018-06-13
make sure advanced filters work with null values

dsnyder-fairway:


dsnyder-fairway:
Hmm, need to check something

dsnyder-fairway:
I ran through UTs and ITs without any issues

jtreuting:
@dsnyder-fairway but it seems like it would be useful to add NULL values to the integration test data so we know that there are examples of them being NULL used during the tests.

dsnyder-fairway:
@jtreuting working on it now

webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.html
```diff
@@ -82,7 +82,7 @@
         </div>
 
         <div *ngSwitchCase="'status'">
-          <span *ngIf="exam.administrativeCondition !== 'Valid'" class="label border-only maroon">{{ 'common.administration-condition.' + exam.administrativeCondition | translate }}</span>
+          <span *ngIf="exam.administrativeCondition && exam.administrativeCondition !== 'Valid'" class="label border-only maroon">{{ 'common.administration-condition.' + exam.administrativeCondition | translate }}</span>
```
wtritch: Might be better to use the `optional` pipe here so we get a `-` if administrative condition isn't provided?

wtritch:
You'll also likely need to change `filter-options.service.ts` Check out the diff between `StrictBoolean` and `Boolean`

wtritch:
also check out `student.service.ts` to see how we map an undefined migrantStatus to the 'undefined' string for handling in the UI.  You might just want to scan for "migrantStatus" and make sure our nullable handling for it is mirrored in the other now-nullable fields.

webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.html
jtreuting: i think the problem in this case is that the - would be inside a bordered box since we are making these like tags.  So I don't think it would fit cleanly here.

dsnyder-fairway:
@wtritch searching `student.service.ts` and other uses of migrantStatus show no other necessary changes

PR #1519, opened 2018-06-14
Small performance improvements

wtritch:
Small query and caching performance improvements.

### DIFF
--- a/reporting-service/src/main/resources/application.sql.yml
+++ b/reporting-service/src/main/resources/application.sql.yml
@@ -98,7 +98,8 @@ sql:
     assessment:
       findAll: >-
           SELECT distinct
-            ${sql.reporting.snippet.assessmentColumns}
+            ${sql.reporting.snippet.assessmentColumns},
+            EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item
           FROM asmt a
             JOIN exam e ON e.asmt_id = a.id
             JOIN school s ON e.school_id = s.id
@@ -113,6 +114,7 @@ sql:
       findLatestExam: >-
           SELECT distinct
            ${sql.reporting.snippet.assessmentColumns},
+           EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item,
            max(e.completed_at) as max_completed
           FROM exam e
             JOIN asmt a ON e.asmt_id=a.id
@@ -123,7 +125,7 @@ sql:
           WHERE e.school_year = :school_year
             AND $[groupFilterPlaceholder]
             AND ${sql.reporting.snippet.studentExamPermissionWithEmbargo}
-          GROUP BY a.id
+          GROUP BY ${sql.reporting.snippet.assessmentColumns}, has_wer_item
           ORDER BY max_completed desc
           LIMIT 1
 
@@ -235,7 +237,8 @@ sql:
     assessment:
       findAll: >-
         SELECT distinct
-          ${sql.reporting.snippet.assessmentColumns}
+          ${sql.reporting.snippet.assessmentColumns},
+          EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item
         FROM asmt a
         WHERE EXISTS (
             SELECT e.asmt_id
@@ -253,6 +256,7 @@ sql:
       findLatestExam: >-
         SELECT distinct
           ${sql.reporting.snippet.assessmentColumns},
+          EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item,
           max(e.completed_at) as max_completed
         FROM exam e
           JOIN asmt a ON e.asmt_id=a.id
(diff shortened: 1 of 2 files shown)


reporting-service/src/main/resources/application.sql.yml
```diff
@@ -122,7 +123,8 @@ sql:
           WHERE e.school_year = :school_year
             AND $[groupFilterPlaceholder]
             AND ${sql.reporting.snippet.studentExamPermissionWithEmbargo}
-          ORDER BY e.completed_at desc
+          GROUP BY a.id
+          ORDER BY max_completed desc
           LIMIT 1
```
wtritch: This drops query execution time from over a second to ~175ms in awsdev.  I'm just grouping the assessments by the max completed_at so that when we sort the table, it only has to filesort 23 rows rather than 18k.

After changes:
explain_after_changes

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
```diff
@@ -47,6 +48,7 @@ public TranslationProviderBackedTranslationJsonService(final TranslationProvider
     }
 
     @Override
+    @Cacheable(value = "translation_properties", key = "#locale.getLanguage() + #includeUI")
```
wtritch: This does some serialization/de-serializing to turn our key/value translation map into json.  I'm not sure we want to do that on every request... thoughts?

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -98,7 +98,8 @@ sql:
     assessment:
       findAll: >-
           SELECT distinct
-            ${sql.reporting.snippet.assessmentColumns}
+            ${sql.reporting.snippet.assessmentColumns},
+            EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item
```
wtritch: I decided to pull the `has_wer_item` column out of the `${sql.reporting.snippet.assessmentColumns}` snippet so that I can use the same snippet in a `GROUP BY` clause below.  MySQL 5.7+ requires all non-aggregated column selections to exist in the GROUP BY clause, and while Aurora is running 5.6 I'm not sure every developer is...

wtritch: UPDATE: it looks like we're doing this elsewhere in my sql (grouping by a single column, but selecting multiple)  so I converted this back to a simple `GROUP BY a.id`

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
esotrope: post-serialization caching sounds good to me : j

reporting-service/src/main/resources/application.sql.yml
malaffoon: Two thoughts:
1. I know it probably won't matter but does grouping by `e.asmt_id` behave exactly the same as `a.id`?
2. Can you put a comment near the sql explaining that the group by / max thing is a performance improvement. Because when i read the way it is now i immediately think, "why bother doing that grouping and max thing?"

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
malaffoon: Doesn't this short-circuit any caching behavior in the translation providers themselves? Even the database one? So, if the customer wants to tweak a message (e.g. the landing page one), they will have to bounce the pods? That's fine but we better document it that way.

reporting-service/src/main/resources/application.sql.yml
wtritch: 1. I ran a test and grouping by e.asmt_id doesn't improve performance at all, but gives the same results.  But since we're actually selecting `a.id` I'd rather group by it unless I'm missing something?
2. Will do

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonService.java
wtritch: It does... so we're double-caching a bit, once at the repo level (which is cached in the reporting-common module) and once here after some post-processing.  And yeah, if the user changes translations they should invalidate their translation caches.  We really should look into adding a cache invalidation endpoint...

reporting-service/src/main/resources/application.sql.yml
agorina: I have to admit that I am surprised to see `GROUP BY` and `DISTINCT` (below) as the performance improvement. It feels like we are missing an index for the query to run fast without it. 

 But I am assuming you have played with it, and if it works, then it is the most important part.

agorina: Hmmm....can we chat about it?

wtritch: @agorina yeah, that was my first thought too, but if you notice from the image above it *is* actually using an index that includes `completed_at` I guess it just has to join on so many tables it's unable to use it for sorting the final results?
I tried creating a separate completed_at-only index, adding it to the school index, creating a whole new multi-field index, and for reasons beyond my ken MySQL refuses to use an exam index for final ordering =/ So the best I could do was restrict the number of rows it had to order.

PR #1520, opened 2018-06-14
added claim and target info icons

jtreuting:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -43,6 +43,18 @@
           <div>{{ 'aggregate-report-table.columns.performance-level-suffix' | translate }}</div>
         </div>
 
+        <span *ngSwitchCase="'claim'"
+              info-button
+              title="{{'target-report.columns.claim' | translate}}"
+              content="{{('common.info.claim') | translate}}">
```
esotrope: nit: could lose the parens

PR #1522, opened 2018-06-14
student test history pluralize result text

jtreuting:


webapp/src/main/webapp/src/app/dashboard/student-dashboard/student-assessment-card.component.html
```diff
@@ -37,6 +37,6 @@
     {{'common.unscored' | translate}}
   </div>
   <div class="result-count center-block mt-sm">
-    {{'student-assessment-card.results' | translate:{resultCount: resultCount} }}
+    {{(resultCount == 1 ? 'student-assessment-card.result' : 'student-assessment-card.results') | translate:{resultCount: resultCount} }}
```
malaffoon: The translation pipeline is accepting the resultCount, how is that supposed to be used? I ask because this now hard-codes american english pluralization rules.

jtreuting: does it hard-code american english?  it allows for there to be a singular and a plural version that can be used.  they can be used however is appropriate right?  we don't have a pluralize type pipe that takes into account language (at least that is my understanding).  the resultCount passed to the translation pipe is just as a replacement in the specific translation that we use based on if it is singular or plural.

PR #1525, opened 2018-06-14
fix when distractor analysis has no results and we select another ses…

dsnyder-fairway:
…sion and attempt to view other results view


webapp/src/main/webapp/src/app/assessments/results/view/distractor-analysis/distractor-analysis.component.ts
```diff
@@ -56,23 +56,23 @@ export class DistractorAnalysisComponent implements OnInit, ExportResults {
     return this._exams;
   }
 
-  loading: boolean = true;
+  loading = true;
   columns: Column[];
   filteredMultipleChoiceItems: AssessmentItem[] = [];
 
-  private _multipleChoiceItems: AssessmentItem[];
+  private _multipleChoiceItems: AssessmentItem[] = [];
```
dsnyder-fairway: the fix

PR #1527, opened 2018-06-14
deselect cards on advanced filter change

dsnyder-fairway:
Unselect cards on change of advance filters, or change of subject / year / assessment type

### DIFF
--- a/webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.ts
+++ b/webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.ts
@@ -9,6 +9,7 @@ import { InstructionalResource } from '../../../assessments/model/instructional-
 import { map } from 'rxjs/operators';
 import { TranslateService } from '@ngx-translate/core';
 import * as _ from 'lodash';
+import { StudentResultsFilterService } from '../student-results-filter.service';
 
 @Component({
   selector: 'student-history-table',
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.ts
```diff
@@ -43,11 +44,16 @@ export class StudentHistoryTableComponent implements OnInit {
...
   constructor(private actionBuilder: MenuActionBuilder,
               private instructionalResourcesService: InstructionalResourcesService,
-              private translateService: TranslateService) {
+              private translateService: TranslateService,
+              private studentResultsFilterService: StudentResultsFilterService) {
     this.createColumns();
   }
 
   ngOnInit(): void {
+    this.studentResultsFilterService.filterChange.subscribe(() => {
```
esotrope: if i gather this correctly we have a child component listening to changes on its parent indirectly through a service. imo the parent should always be defining how two things interact, the child should not be telling the parent how to behave.

can we get away with just doing this?
https://github.com/SmarterApp/RDW_Reporting/pull/1528/files#diff-9c03f89ba081dcbd45913e423bf57461R159

dsnyder-fairway: yeah. We have the StudentResultsFilterComponent.filterChange emitting whenever a filter change happens, which is picked up by StudentResultsComponent, and we use the service from there to indicate that the filter has been changed

PR #1534, opened 2018-06-15
corrected styling for Subjects and Claims selections

dsnyder-fairway:
Also, now all subject selectors now have a vertical display (like Claims did).

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
+++ b/webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
@@ -145,15 +145,16 @@
                 <label for="subjects">{{'aggregate-report-form.field.subjects-label' | translate}}</label>
               </div>
               <div class="col-md-3">
-                <sb-checkbox-group id="subjects"
-                                   name="subjects"
-                                   analyticsEvent="FilterClick"
-                                   analyticsCategory="AggregateQueryBuilder"
-                                   [allOptionAnalyticsProperties]="{label: 'Subject: All'}"
-                                   [options]="filteredOptions.subjects"
-                                   [(ngModel)]="settings.subjects"
-                                   [ngModelOptions]="{standalone: true}"
-                                   (change)="onSettingsChange()"></sb-checkbox-group>
+                <sb-button-group id="subjects"
+                                 name="subjects"
+                                 analyticsEvent="FilterClick"
+                                 analyticsCategory="AggregateQueryBuilder"
+                                 [vertical]="true"
+                                 [allOptionAnalyticsProperties]="{label: 'Subject: All'}"
+                                 [options]="filteredOptions.subjects"
+                                 [(ngModel)]="settings.subjects"
+                                 [ngModelOptions]="{standalone: true}"
+                                 (change)="onSubjectsChange()"></sb-button-group>
               </div>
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -145,16 +145,15 @@ <h3>{{'aggregate-report-form.section.assessment-heading' | translate}}</h3>
...
-                                 name="subjects"
-                                 [vertical]="true"
-                                 analyticsEvent="FilterClick"
-                                 analyticsCategory="AggregateQueryBuilder"
-                                 [allOptionAnalyticsProperties]="{label: 'Subject: All'}"
-                                 [options]="filteredOptions.subjects"
-                                 [(ngModel)]="settings.subjects"
-                                 [ngModelOptions]="{standalone: true}"
-                                 (change)="onSubjectsChange()"></sb-button-group>
+                <sb-checkbox-group id="subjects"
```
esotrope: is this to fix the styling? we don't know why button group styles don't work and checkbox group styles do?

dsnyder-fairway: hmm, I might have just needed to put type="checkbox" on this. I forgot sb-checkbox-group was deprecated

webapp/src/main/webapp/src/styles.less
```diff
@@ -429,6 +429,20 @@ body .select-menu-group {
   }
 }
 
+.nested-btn-group {
+  > a:first-child {
+    display: table-cell;
+  }
+  &.vertical {
+    > a:first-child {
```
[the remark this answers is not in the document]
dsnyder-fairway: Yes, and I had to add border-top/border-bottom radiuses. Not sure where those are applied, but they are missing from anchors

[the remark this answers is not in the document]
esotrope: ok cool. can you also add a comment that this should be moved to ui kit so we can sort it out later

webapp/src/main/webapp/src/app/aggregate-report/query-forms/target-report-form.component.html
```diff
@@ -68,14 +68,16 @@ <h3>{{'aggregate-report-form.section.assessment-heading' | translate}}</h3>
                 <label for="subject">{{'aggregate-report-form.field.subject-label' | translate}}</label>
               </div>
               <div class="col-md-6">
-                <sb-radio-group id="subject"
```
dsnyder-fairway: I'll revert this change

webapp/src/main/webapp/src/app/shared/sb-toggle.component.ts
```diff
@@ -7,7 +7,7 @@ const DefaultButtonGroupStyles = 'btn-group-sm';
 const DefaultButtonStyles = 'btn-primary';
 
 /**
- * @deprecated use <sb-button-group type="radio">
+ * @deprecated use <sb-radio-group>
```
dsnyder-fairway: `<sb-button-group type="radio">` no longer works

esotrope: : / that's not good

PR #1537, opened 2018-06-15
order ELA claim codes and fix the width of claim button groups

dsnyder-fairway:


webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
```diff
@@ -9,6 +9,8 @@ export const BooleanOrdering = ordering(ranking([ 'yes', 'no', 'undefined' ]));
 
 export const CompletenessOrdering = ordering(ranking([ 'Complete', 'Partial' ]));
 
+export const ClaimReportElaClaimOrder = ordering(ranking(['SOCK_R', '2-W', 'SOCK_LS', '4-CR']));
```
jtreuting: We have started to refer to these types of claims as "Scorable Claims" so maybe call this ScorableClaimOrder.  Also seems like we could put the Math ones in here as well since the exact index in the array doesn't matter.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
```diff
@@ -170,8 +172,12 @@ export class ClaimReportFormComponent extends MultiOrganizationQueryFormComponen
     this.filteredOptions.subjects.forEach(subject => {
       const subjectCode = subject.value;
       this.claimsBySubject[ subjectCode ] = this.filteredOptions.claimCodes
-          .filter(claim => claim.value.subject === subjectCode
-            && claim.value.assessmentType === this.settings.assessmentType);
+        .filter(claim => claim.value.subject === subjectCode
+          && claim.value.assessmentType === this.settings.assessmentType);
+      if (subjectCode === 'ELA') {
```
jtreuting: if we follow the pattern for SubjectClaimOrdering then we won't need to hard-code this type of check i think.  It has a map and it can handle when there is a new subject that it doesn't know about whcih is important for new subjects in the future.  check out how/where that is used to see if that pattern will work here.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -333,8 +333,13 @@ export class AggregateReportTableComponent implements OnInit {
     this._orderingByColumnField[ 'assessmentGradeCode' ] = assessmentGradeOrdering;
     this._orderingByColumnField[ 'schoolYear' ] = SchoolYearOrdering;
     this._orderingByColumnField[ 'claimCode' ] = reportType === AggregateReportType.Target
-      ? OrganizationalClaimOrderingProvider(rows[0].subjectCode, this.preview)
-      : ordering(ranking(options.claims.map(claim => claim.code))).on<AggregateReportItem>(item => item.claimCode);
+      ? OrganizationalClaimOrderingProvider(rows[ 0 ].subjectCode, this.preview)
+      : ordering(
+        join(
+          ranking(options.claims.filter(claim => claim.subject !== 'ELA').map(claim => claim.code)),
```
esotrope: is this because there is overlap in the codes between subjects?

dsnyder-fairway:
@jtreuting I was trying to get this into `ordering.ts` but haven't had any luck yet

PR #1540, opened 2018-06-15
Fix change notification on sb input components

wtritch:
There was a large issue with `(change)` handlers not being called on sb-button-group instances.

I also gave the stylings a once-over to make sure we were highlighting on focus/focus-within

radio highlight
image

button-group highlight:
image

horizontal button-group highlight:
image

checkbox-group highlight:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
@@ -124,25 +124,22 @@ const ControllerByInputType: {[inpuType: string]: InputController} = {
   selector: 'sb-button-group',
   template: `
     <ng-template #button let-option let-isAllOption="isAllOption">
-      <a class="btn"
-         href="javascript:void(0)"
-         (click)="isAllOption ? onAllOptionClickInternal() : onOptionClickInternal(option)"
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -124,25 +124,22 @@ const ControllerByInputType: {[inpuType: string]: InputController} = {
   selector: 'sb-button-group',
   template: `
     <ng-template #button let-option let-isAllOption="isAllOption">
-      <a class="btn"
-         href="javascript:void(0)"
-         (click)="isAllOption ? onAllOptionClickInternal() : onOptionClickInternal(option)"
-         [ngClass]="computeStylesInternal(buttonStyles, { 
-             active: isAllOption ? stateInternal.selectedAllOption : stateInternal.selectedOptions.has(option), 
-             disabled: disabled 
-         })">
```
wtritch: Binding click events to an anchor tag here prevented (change) handlers from being called.  I just reverted to a label and made sure tab-navigation and focus-styling were functional, although it looks like a lot of the other focus-styling work fixed 90% of the issues.

PR #1541, opened 2018-06-16
adds some TODOs on things that will need to change to support configurable…

esotrope:
… subjects.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -30,6 +30,7 @@ const SchoolYearOrdering: Ordering<AggregateReportItem> = ordering(byNumber)
 const AssessmentLabelOrdering: Ordering<AggregateReportItem> = ordering(byString)
   .on(item => item.assessmentLabel);
 
+// TODO:ConfigurableSubjects this needs to be provided by the backend
 const OrganizationalClaimOrderingProvider: (subjectCode: string, preview: boolean) => Ordering<AggregateReportItem> = (subjectCode, preview) => {
   const currentOrdering: Ordering<string> = !preview && SubjectClaimOrderings.has(subjectCode)
     ? SubjectClaimOrderings.get(subjectCode)
--- a/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
+++ b/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
@@ -258,10 +258,12 @@ export class CsvBuilder {
     );
   }
 
+  // TODO:ConfigurableSubjects needs to support configurable subjects
   withMathClaimScores(getExam: (item: any) => Exam) {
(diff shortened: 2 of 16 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -30,6 +30,7 @@ const SchoolYearOrdering: Ordering<AggregateReportItem> = ordering(byNumber)
 const AssessmentLabelOrdering: Ordering<AggregateReportItem> = ordering(byString)
   .on(item => item.assessmentLabel);
 
+// TODO:ConfigurableSubjects this needs to be provided by the backend
 const OrganizationalClaimOrderingProvider: (subjectCode: string, preview: boolean) => Ordering<AggregateReportItem> = (subjectCode, preview) => {
```
jtreuting: i'm not sure if the configurable subjects XML file supports a specific order for claims.  if it does, great.  if not then the idea of this provider is to have the SB ordering hard-coded and then fall back to basic string ordering for everything else.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -258,10 +258,12 @@ export class CsvBuilder {
...
   }
 
+  // TODO:ConfigurableSubjects needs to support configurable subjects
   withMathClaimScores(getExam: (item: any) => Exam) {
     return this.withClaimScores([ '1', 'SOCK_2', '3' ], getExam);
   }
 
+  // TODO:ConfigurableSubjects needs to support configurable subjects
   withELAClaimScores(getExam: (item: any) => Exam) {
     return this.withClaimScores([ 'SOCK_R', 'SOCK_LS', '2-W', '4-CR' ], getExam);
```
jtreuting: i know it isn't part of this PR but, these are ordered incorrectly.  we need to have these match the ordering used in the rest of the site which would be SOCK_R, 2-W, SOCK_LS, 4-CR

webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
```diff
@@ -3,17 +3,20 @@ import { Ordering, ordering } from '@kourge/ordering';
...
 export const AssessmentTypeOrdering = ordering(ranking(['sum', 'ica', 'iab']));
 
+// TODO:ConfigurableSubjects these should be provided by the backend
 export const SubjectOrdering = ordering(ranking(['Math', 'ELA']));
 
 export const BooleanOrdering = ordering(ranking([ 'yes', 'no', 'undefined' ]));
 
 export const CompletenessOrdering = ordering(ranking([ 'Complete', 'Partial' ]));
 
+// TODO:ConfigurableSubjects this needs to be provided by the backend
```
jtreuting: as mentioned, if we have claim ordering defined in the XML file then i agree, if not then this is the hard-coded SB specific rules and it falls back to a basic string ordering which is appropriate.

esotrope:
merging these todo comments

PR #1542, opened 2018-06-17
need to filter by assessment type otherwise it gets duplicate claims …

jtreuting:
…for ICA and Summative

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
@@ -152,8 +152,7 @@ export class AggregateReportTableDataService {
                 .find(subject => true);
 
               return claims
-                // do not filter on assessment type
-                .filter(claim => claim.subject === subjectCode)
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -152,8 +152,7 @@ export class AggregateReportTableDataService {
                 .find(subject => true);
 
               return claims
-                // do not filter on assessment type
```
jtreuting: @dsnyder-fairway, @esotrope any idea why this was specifically called out to not filter by assessment type?  when not filtering on assessment type, it grabs 3 claims for ICA and 3 claims for Summative and therefore the duplicate rows

wtritch: Hm... that's strange.  I'm pretty sure we *should* be filtering on assessment type?

dsnyder-fairway: This was done to address DWR-1896. I would just confirm that changing this doesn't re-introduce the issue

PR #1543, opened 2018-06-18
add configuration options for the landing page and access denied urls

jtreuting:
As part of having multiple tenants, they need the ability to have their own controlled reporting landing page where they can then direct folks to a tenant specific installation.  These options will allow them to keep the existing page or move to an externally Smarter Balanced controlled page/site.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
@@ -20,9 +20,7 @@ public class ViewController implements ErrorController {
     @RequestMapping(value = {"/"})
     public String index(@AuthenticationPrincipal final User user) {
         if (user == null) {
-            return this.landingPageUrl == null || this.landingPageUrl.isEmpty()
-                    ? "forward:/landing.html"
-                    : "redirect:" + this.landingPageUrl;
+            return this.landingPageUrl;
         }
 
         return "forward:/index.html";
@@ -36,9 +34,7 @@ public class ViewController implements ErrorController {
 
     @RequestMapping(value="/error/accessDenied", method = { RequestMethod.GET, RequestMethod.POST })
     public String accessDenied() {
-        return this.accessDeniedUrl == null || this.accessDeniedUrl.isEmpty()
-                ? "forward:/assets/public/access-denied.html"
-                : "redirect:" + this.accessDeniedUrl;
+        return this.accessDeniedUrl;
     }
 
     @RequestMapping(value="/error")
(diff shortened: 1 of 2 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -18,8 +18,10 @@ permissionservice:
 
 reporting:
   analytics-tracking-id: ${rdw.reporting.analytics.trackingId:}
+  access-denied-url:
   interpretive-guide-url: https://portal.smarterbalanced.org/library/en/reporting-system-interpretive-guide.pdf
   iris-vendor-id: ${rdw.reporting.iris.vendorId}
+  landing-page-url:
```
jtreuting: i considered having this have a default of "forward:/landing.html" instead of having that in the code as the default but it felt safer there so that if nothing is provided we still have something that works.

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -10,10 +11,18 @@
...
+    @Value("${reporting.landing-page-url}")
+    private String landingPageUrl;
+
     @RequestMapping(value = {"/"})
     public String index(@AuthenticationPrincipal final User user) {
         if (user == null) {
-            return "forward:/landing.html";
+            return this.landingPageUrl == null || this.landingPageUrl.isEmpty()
+                    ? "forward:/landing.html"
+                    : "redirect:" + this.landingPageUrl;
```
malaffoon: If the value is required it will never be null. Our convention has been to put a default values in the embedded application.yml (or encoded into Value annotation expression) and not worry about null/empty.

webapp/src/main/resources/application.yml
malaffoon: Practically, that is never a problem. Put the default in here. They have to provide a string to override it (putting the key in the yml without a value will *not* override the embedded value with an empty string).

webapp/src/main/resources/application.yml
```diff
@@ -18,8 +18,10 @@ permissionservice:
 
 reporting:
   analytics-tracking-id: ${rdw.reporting.analytics.trackingId:}
+  access-denied-url: forward:/assets/public/access-denied.html
   interpretive-guide-url: https://portal.smarterbalanced.org/library/en/reporting-system-interpretive-guide.pdf
   iris-vendor-id: ${rdw.reporting.iris.vendorId}
+  landing-page-url: forward:/landing.html
```
jtreuting: @malaffoon I'm guessing the annotated config file is the place to put a comment saying something about using "redirect:url" when wanting to redirect to an external site/page?

malaffoon: Yes please.

PR #1547, opened 2018-06-18
 send latest assessment exam to assessment component when necessary

dsnyder-fairway:


webapp/src/main/webapp/src/app/groups/results/group-results.component.ts
```diff
@@ -116,7 +117,15 @@ export class GroupResultsComponent implements OnInit, StateProvider {
     // update latest assessment when resolved route data changes
     // this and the resolve could be replaced later by a manual invocation when the route params change
     this.route.data.subscribe(({ assessment }) => {
-      this.updateAssessment(assessment);
+      this.latestAssessmentExam = assessment;
+      if (!this.latestAssessmentExam) {
```
esotrope: i think this PR could be simplified to just removing the `assessmentIds == null` guard in the `group-assessment.resolve`. ideally we don't want to resolve the assessment latest assessment unless assessmentIDs are not specified though

jtreuting: i think there are 2 solutions right.

1) Pull the latest assessment from within the group results if needed and not provided already (not provided because the resolver ignored it because assessment IDs were selected)

2) Always get the latest in the resolver even if it isn't needed for the group results initially (and might not be unless they click Most Recent).

the pros of option 1 are that it is more efficient to load the page and might save a call that isn't ever used.

pros of option 2 is that it;s the simplest

sound right?

esotrope: sounds right. also we want the route assessmentIds to trickle down to the assessment-results component so it can load the right assessments when the browser back/forward buttons are hit

dsnyder-fairway: @esotrope hmmm I thought that wasn't working for me at first, but after debugging again it looks like that solution works. I'll open a new PR with those changes

PR #1548, opened 2018-06-18
Task/admin well on homepage

jtreuting:
Moved the admin related buttons from the top area to their own well on the homepage.  It was decided that if you had access to custom aggregate reports or to manage the application those were important and more important than the individual results you might also be able to access.

image


webapp/src/main/webapp/src/app/home/home.component.html
```diff
@@ -2,49 +2,28 @@
...
-  </ng-template>
 </page-heading>
 
 <div class="row">
 
   <!-- Left Column Content Area -->
   <div class="col-md-9">
+    <admin-tools></admin-tools>
 
-    <div *hasAnyPermission="['GROUP_PII_READ', 'INDIVIDUAL_PII_READ', 'GROUP_PII_READ']; else noPermissions">
```
jtreuting: no need for the "No permissions, check the buttons above" message since those buttons are now in the main content area when a user has them

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1088,6 +1088,32 @@
     "view-dashboard": "View IAB Dashboard"
   },
   "home": {
+    "administrator": {
```
esotrope: nit: to follow our convention this should be "admin-tools"

webapp/src/main/webapp/src/app/home/admin-tools.component.html
```diff
@@ -0,0 +1,59 @@
...
+      <div class="row">
+        <div *hasPermission="'GROUP_WRITE'" class="col-xs-6">
+          <p class="small">{{ 'home.administrator.student-groups.description' | translate}}</p>
+          <a routerLink="/admin-groups"
+             class="btn btn-primary btn-xs">
+            <i class="fa fa-edit"></i>
+            {{ 'home.administrator.student-groups.button' | translate}}
+          </a>
+        </div>
+        <div *hasPermission="'INSTRUCTIONAL_RESOURCE_WRITE'" class="col-xs-6">
```
esotrope: what does it look like if you have instructional resource, group and embargo write?

jtreuting: one of them wraps down a line

webapp/src/main/webapp/src/assets/i18n/en.json
jtreuting: thx, done

PR #1549, opened 2018-06-18
correct "most recent" selected assessment button after viewing assess…

dsnyder-fairway:
…ments from group iab dashboard


webapp/src/main/webapp/src/app/groups/results/group-assessment.resolve.ts
```diff
@@ -12,11 +11,6 @@ export class GroupAssessmentResolve implements Resolve<AssessmentExam> {
   }
 
   resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<AssessmentExam> {
-    // if there are pre-selected assessments, don't call to get the most recent assessment
-    const { schoolYear, assessmentIds } = route.params;
-    if (schoolYear == null || assessmentIds != null) {
```
esotrope: school year guard we actually should keep as this will save us from double loading the latest assessment

dsnyder-fairway: done

PR #1554, opened 2018-06-19
claim ordering across the board

jtreuting:
made a general scorable claim ordering provider that accepts a function to get the claim.  this is needed because some places have just the claim code as a string and others have a class with a claim property.

### DIFF
--- a/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
+++ b/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
@@ -276,17 +276,17 @@ export class CsvBuilder {
       ['SOCK_R', 0], ['SOCK_LS', 1], ['2-W', 2], ['4-CR', 3]
     ]);
 
-    claims.forEach((claim, idx) => {
+    claims.forEach((claim) => {
       this.withColumn(
         this.translateService.instant(`common.subject-claim-code.${claim}`),
         (item) => {
           const exam: Exam = getExam(item);
-          const idx = claimCodeToDatabaseIndex.get(claim);
-          if (!exam || !exam.claimScores[ idx ].level) {
+          const index = claimCodeToDatabaseIndex.get(claim);
+          if (!exam || !exam.claimScores[ index ].level) {
             return '';
           }
-          return this.translateService.instant(exam.claimScores[ idx ].level
-            ? `common.assessment-type.iab.performance-level.${exam.claimScores[ idx ].level}.name`
+          return this.translateService.instant(exam.claimScores[ index ].level
+            ? `common.assessment-type.iab.performance-level.${exam.claimScores[ index ].level}.name`
             : 'common.missing'
           );
         }
--- a/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
+++ b/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
@@ -14,6 +14,7 @@ import { TranslateDatePipe } from "../shared/i18n/translate-date.pipe";
 import { TranslateNumberPipe } from "../shared/i18n/translate-number.pipe";
 import { ApplicationSettingsService } from "../app-settings.service";
 import { AggregateTargetScoreRow, TargetReportingLevel } from '../assessments/model/aggregate-target-score-row.model';
+import { ScorableClaimOrder } from '../shared/ordering/orderings';
 
 @Injectable()
 export class CsvBuilder {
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -260,20 +261,27 @@ export class CsvBuilder {
...
-    return this.withClaimScores([ 'SOCK_R', 'SOCK_LS', '2-W', '4-CR' ], getExam);
+    return this.withClaimScores(ScorableClaimOrder.get('ELA'), getExam);
   }
 
   withClaimScores(claims: string[], getExam: (item: any) => Exam) {
+    // TODO:ConfigurableSubjects
+    const claimCodeToDatabaseIndex: Map<string, number> = new Map([
+      ['1', 0], ['SOCK_2', 1], ['3', 2],
+      ['SOCK_R', 0], ['SOCK_LS', 1], ['2-W', 2], ['4-CR', 3]
+    ]);
```
jtreuting: the csv claims were in a specific order to match the database column ordering.  therefore i needed to map the claim code to the DB column order so that it pulls the right value for the claim.  will need to make changes here based on configurable subjects.

webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.html
```diff
@@ -26,7 +26,7 @@
 <ng-template #tableTemplate>
 
     <div *ngIf="assessmentType !== 'iab'"
-         class="btn-group btn-group-md"
+         class="btn-group btn-group-xs pull-right mb-xs"
```
jtreuting: started by ordering claims on the student test history page and made the overall/claim toggle smaller buttons and aligned right to match how it is on the assessment results pages

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -260,20 +261,27 @@ export class CsvBuilder {
...
+
     claims.forEach((claim, idx) => {
```
dsnyder-fairway: might want to remove this `idx`

webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
```diff
@@ -20,6 +20,13 @@ export const ScorableClaimOrderings: Map<string, Ordering<string>> = new Map([
   [ 'ELA', ordering(ranking(ScorableClaimOrder.get('ELA'))) ]
 ]);
 
+export const ScorableClaimOrderingProvider: (subjectCode: string, f: (data: any) => string) => Ordering<any> = (subjectCode, f) => {
+  const currentOrdering: Ordering<string> = ScorableClaimOrderings.has(subjectCode)
+    ? ScorableClaimOrderings.get(subjectCode)
+    : ordering(byString);
+  return currentOrdering.on(f);
```
esotrope: i would just return the ordering here and let the consumer optionally add the `.on()` 
also, since this is a factory method i would call it `createScorableClaimOrdering`

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
wtritch: hmm... so this is basically just the `exam_claim_score_mapping` table duplicated in the UI?

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -260,20 +261,27 @@ export class CsvBuilder {
...
       this.withColumn(
         this.translateService.instant(`common.subject-claim-code.${claim}`),
         (item) => {
           const exam: Exam = getExam(item);
+          const idx = claimCodeToDatabaseIndex.get(claim);
```
esotrope: what's an idx?

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
dsnyder-fairway: nit: could have this defined outside of the class

webapp/src/main/webapp/src/app/student/results/tables/student-history-table.component.ts
```diff
@@ -175,14 +176,16 @@ export class StudentHistoryTableComponent implements OnInit {
...
-        id: 'claim',
-        field: 'exam.claimScores.' + index + '.level',
-        claim: claim,
-        index: index
+    return this.exams[ 0 ].assessment.claimCodes
+      .sort(ScorableClaimOrderingProvider(this.exams[ 0 ].assessment.subject, code => code).compare)
+      .map((claim: string, index: number) => {
+        return new Column({
+          id: 'claim',
+          field: 'exam.claimScores.' + index + '.level',
```
esotrope: nit: could be: `exam.claimScores.${index}.level`

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
jtreuting: @wtritch sounds like it probably is :) in this case we didn't have that in the UI and I figured this would have to change for the configurable subjects so I'm not sure it's worth pulling that in right now before the release version.

@dsnyder-fairway i think it could but i see it being refactored soon and it is only used here right now so felt like keeping it close made more sense right now.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
jtreuting: yeah i think i should.  it seems to work fine but yeah that is confusing, thanks.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
jtreuting: it's the same thing it was before i made any changes :)

webapp/src/main/webapp/src/app/shared/ordering/orderings.ts
jtreuting: i could do that instead of passing in the function, i was keeping with an existing pattern of having the pattern return the `on` part, but i'm fine changing that.  i think it will feel cleaner to the consumer of it.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
esotrope: is it not more straight forward to just sort a copy of these claims then iterate over them?

jtreuting: they are sorted first how we want them displayed.  but then the index of the claim that we iterate over does not match with the index in the claimCodes array returned from the server.  so i'm mapping from the code to the server index so we get the right value for the claim.  everywhere else we are sorting claims the claim code is returned with the value so we know which it is and don't have to have this map.  i'm assuming with the configurable subject changes we will return that as well to make this unnecessary at that point.

PR #1556, opened 2018-06-19
fixes caching issue with available assessments

esotrope:


webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -351,27 +356,23 @@ export class AssessmentsComponent implements OnInit {
...
 
-  private removeUnselectedAssessmentExams() {
+  private removeUnselectedAssessmentExams(): void {
     this._assessmentExams = this._assessmentExams.filter(loaded => this.selectedAssessments.some(selected => loaded.assessment.id == selected.id));
   }
 
-  private getAvailableAssessments() {
+  private getAvailableAssessments(): Observable<Assessment[]> {
     if (this._expandAssessments) {
-      const observable = this._allAvailableAssessments != null && this._allAvailableAssessments.length != 0
```
esotrope: we were not loading new assessments when the group/school-grade/school year changed because we were caching based on `_allAvailableAssessments` which was only set in ngOnInit(). seems like before we probably had a manual flush of this property somewhere but lost it.

Now this caches based on the route params.

webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -251,3 +251,18 @@ export class Utils {
...
+
+/**
+ * Creates a string based on the provided parameters
+ * This method assumes the parameters is a basic javascript object and contains only primitive types
+ * This is useful when caching results based on request or route parameters
+ *
+ * @param parameters the object to create the key from
+ * @returns {string} the key
+ */
+export function serializeURLParameters(parameters: any): string {
```
esotrope: common function extracted from caching data service

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -126,10 +127,18 @@ export class AssessmentsComponent implements OnInit {
 
   set expandAssessments(value: boolean) {
     this._expandAssessments = value;
-    this.getAvailableAssessments();
+    this.getAvailableAssessments().subscribe(result => {
+      // TODO fix this so that we don't need an Array.map callback with side-effects
+      this.availableAssessments = result.map(available => {
```
esotrope: before this was in the `this.getAvailableAssessments()` method which returns an observable but it didnt make much sense to subscribe to the observable then return it. also the ngOnInit() overwrote the results of this subscription when the page was loaded with assessmentIds in the route...

PR #1558, opened 2018-06-19
use student's ssid when no name is provided

dsnyder-fairway:


common/src/main/resources/i18n/en.json
```diff
@@ -778,6 +778,7 @@
       "school": {
         "label": "School:"
       },
+      "ssid": "{0}",
```
jtreuting: do we need this?  could we just include the SSID directly when needed

webapp/src/main/webapp/src/app/shared/layout/sb-breadcrumbs.component.ts
```diff
@@ -134,20 +134,28 @@ export class SbBreadcrumbs implements OnInit {
...
+      if (typeof options.translate === 'string') {
+        text = options.translateResolve
+          ? this.translateService.instant(options.translate, Utils.getPropertyValue(<string>options.translateResolve, routeData))
+          : this.translateService.instant(options.translate);
+      } else {
+        // secondary translation available. Attempt translations until we translate the text or until we run out of translation keys
+        let i = 0;
+        do {
+          text = this.translateService.instant(options.translate[ i ], Utils.getPropertyValue(options.translateResolve, routeData));
+        } while (options.translate[ i + 1 ] && text === this.translateService.instant(options.translate[ i++ ]));
```
esotrope: not sure i follow why we need to process an array of strings can you elaborate?

webapp/src/main/webapp/src/app/student/results/student-results.component.html
```diff
@@ -83,7 +83,7 @@ <h3>{{'student-results.title' | translate}}</h3>
   </div>
 
   <span student-report-download #reportDownloader
-        [title]="'student-results.create-report' | translate:examHistory.student"
+        [title]="'student-results.create-report' | studentNameTranslation:examHistory.student"
```
esotrope: is there translation involved or just formatting?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -695,12 +695,17 @@
     "menus": {
       "resources": "Instructional Resources",
       "responses": "{{firstName}}'s Responses",
+      "responses-ssid": "{{ssid}}'s Responses",
```
jtreuting: I'm not sure we need the "-ssid" versions of these.  Couldn't we just use the SSID to replace the existing {{firstName}}?  Or maybe even better changing {{firstName}} to something more generic (not sure what, maybe {{name}} or {{student}}?) and then it could be replaced with either "student.firstName" or "student.ssid"

webapp/src/main/webapp/src/app/shared/layout/sb-breadcrumbs.component.ts
dsnyder-fairway: I wanted some sort of fall-back. If we pass an array, we will use the keys of the array until we have a successful translation, or we run out of keys. We use this to get the user First name and last name, or SSID

webapp/src/main/webapp/src/app/student/results/student-results.component.html
dsnyder-fairway: it gets the students first name or ssid

common/src/main/resources/i18n/en.json
dsnyder-fairway: good point. I'll update this

webapp/src/main/webapp/src/assets/i18n/en.json
dsnyder-fairway: we could

jtreuting: that feels cleaner to me.  @esotrope thoughts?

webapp/src/main/webapp/src/app/student/results/student-results.component.html
esotrope: i would prefer we pass in a custom handler for this in the route def. something like:
```
{
  breadcrumbs: {
    transform: (student) => studentPipe.transform(student)
  }
}
```

webapp/src/main/webapp/src/app/shared/layout/sb-breadcrumbs.component.ts
esotrope: i would prefer we pass in a custom handler for this in the route def. something like:
```
{
  breadcrumbs: {
    transform: (student) => studentPipe.transform(student)
  }
}
```

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: ideally it would be nice if the template could have a fallback baked in with `?:` but since we don't i'm tempted to delegate this to the `studentPipe` so that the fallback logic is in one place. maybe:
```
studentPipe.transform(student, {omitLastName: true})
```

webapp/src/main/webapp/src/app/shared/layout/sb-breadcrumbs.component.ts
dsnyder-fairway: thanks for the suggestion. I think this is a better approach

dsnyder-fairway:
@esotrope @jtreuting how are we feeling about this?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1399,8 +1399,9 @@
     "title": "Student's Responses"
   },
   "student-results": {
-    "create-report": "Create {{firstName}}'s Printable Report",
-    "crumb": "Student: {{lastName}}, {{firstName}}",
+    "create-report": "Create {{value}}'s Printable Report",
+    "create-report-ssid": "Create {{ssid}}'s Printable Report",
```
jtreuting: do we still need this one?

dsnyder-fairway: nope. Removed

PR #1562, opened 2018-06-20
Add a subject definition endpoint to the reporting service

wtritch:
This PR adds a "/subjectdefinitions" endpoint to the reporting-service to allow the front-end to retrieve subject definitions within the context of an assessment type.

Sample request: 
http://localhost:8080/api/reporting-service/subjectdefinitions

Sample payload:
```
[
{
asmtTypeCode: "ica",
subjectCode: "ELA",
performanceLevelCount: 4,
performanceLevelStandardCutoff: 3,
claimScorePerformanceLevelCount: 3
},
{
asmtTypeCode: "iab",
subjectCode: "ELA",
performanceLevelCount: 3
},
{
asmtTypeCode: "sum",
subjectCode: "ELA",
performanceLevelCount: 4,
performanceLevelStandardCutoff: 3,
claimScorePerformanceLevelCount: 3
},
{
asmtTypeCode: "ica",
subjectCode: "Math",
performanceLevelCount: 4,
performanceLevelStandardCutoff: 3,
claimScorePerformanceLevelCount: 3
},
{
asmtTypeCode: "iab",
subjectCode: "Math",
performanceLevelCount: 3
},
{
asmtTypeCode: "sum",
subjectCode: "Math",
performanceLevelCount: 4,
performanceLevelStandardCutoff: 3,
claimScorePerformanceLevelCount: 3
}
]
```

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
@@ -1,0 +1,79 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * This model represents a subject's properties within the scope of an Assessment Type.
+ */
+public class SubjectDefinition {
+    private String asmtTypeCode;
+    private String subjectCode;
+    private int performanceLevelCount;
+    private Integer performanceLevelStandardCutoff;
+    private Integer claimScorePerformanceLevelCount;
+
+    public String getAsmtTypeCode() {
+        return asmtTypeCode;
+    }
+
+    public String getSubjectCode() {
+        return subjectCode;
+    }
+
+    public int getPerformanceLevelCount() {
+        return performanceLevelCount;
+    }
+
+    public Integer getPerformanceLevelStandardCutoff() {
+        return performanceLevelStandardCutoff;
+    }
+
+    public Integer getClaimScorePerformanceLevelCount() {
+        return claimScorePerformanceLevelCount;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private String asmtTypeCode;
+        private String subjectCode;
+        private int performanceLevelCount;
+        private Integer performanceLevelStandardCutoff;
+        private Integer claimScorePerformanceLevelCount;
+
+        public SubjectDefinition build() {
+            final SubjectDefinition definition = new SubjectDefinition();
+            definition.asmtTypeCode = asmtTypeCode;
+            definition.subjectCode = subjectCode;
+            definition.performanceLevelCount = performanceLevelCount;
+            definition.performanceLevelStandardCutoff = performanceLevelStandardCutoff;
+            definition.claimScorePerformanceLevelCount = claimScorePerformanceLevelCount;
+            return definition;
+        }
+
+        public Builder asmtTypeCode(final String asmtTypeCode) {
+            this.asmtTypeCode = asmtTypeCode;
+            return this;
+        }
+
+        public Builder subjectCode(final String subjectCode) {
+            this.subjectCode = subjectCode;
+            return this;
+        }
+
+        public Builder performanceLevelCount(final int performanceLevelCount) {
+            this.performanceLevelCount = performanceLevelCount;
+            return this;
+        }
+
+        public Builder performanceLevelStandardCutoff(final Integer performanceLevelStandardCutoff) {
+            this.performanceLevelStandardCutoff = performanceLevelStandardCutoff;
+            return this;
+        }
+
+        public Builder claimScorePerformanceLevelCount(final Integer claimScorePerformanceLevelCount) {
+            this.claimScorePerformanceLevelCount = claimScorePerformanceLevelCount;
+            return this;
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/SubjectDefinitionRepository.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.opentestsystem.rdw.reporting.common.model.SubjectDefinition;
+
+import java.util.Collection;
+
+/**
+ * Implementations of this interface are responsible for querying for
+ * {@link SubjectDefinition} instances.
+ */
+public interface SubjectDefinitionRepository {
```
wtritch: I put the repository and model in reporting-common because this will likely be needed by the report-processor for PDF generation as well.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/SubjectDefinitionRepository.java
```diff
@@ -0,0 +1,18 @@
...
+
+    /**
+     * Find all SubjectDefinitions within the system.
+     * @return All SubjectDefinitions
+     */
+    Collection<SubjectDefinition> findAll();
```
wtritch: I just created a single "findAll" call right now... since this will be called once per user per login I wasn't sure we wanted caching... but maybe we do. Thoughts?

jtreuting: i would think we would since this is something that changes very infrequently.  especially since under load the login process has a lot going on

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public String getAsmtTypeCode() {
+        return asmtTypeCode;
+    }
+
+    public String getSubjectCode() {
+        return subjectCode;
+    }
+
+    public int getPerformanceLevelCount() {
```
jtreuting: if we just have the count, then the consumer of this would need to loop from 1 to count to get the labels right?  my gut reaction is that it would be nice to have this as an array so the consumer can just iterate over it instead of knowing that we are enforcing the levels to be 1 - N.

wtritch: I'm not exactly sure what you're asking, maybe we can discuss in the parking lot?

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/SubjectDefinitionRepository.java
malaffoon: Alla and i were talking about this yesterday and feel that (re)loading subjects requiring a bounce of applications is just fine for now. The only issue here is that there might be a process involved:
1. Load new subjects.
2. Bounce ingest/migrate services.
3. Wait for migrate.
4. Bounce reporting services.
Steps 2-4 could be combined but then an eager user / smoke test might load reporting caches before the data is migrated.

wtritch: good point, done

common/src/main/resources/common.sql.yml
```diff
@@ -116,6 +116,18 @@ sql:
...
+    findAll: >-
+      SELECT
+        at.code as asmt_type_code,
+        s.code as subject_code,
+        sat.performance_level_count,
+        sat.performance_level_standard_cutoff,
+        sat.claim_score_performance_level_count
+      FROM subject_asmt_type sat
+        JOIN asmt_type at ON at.id = sat.asmt_type_id
+        JOIN subject s ON s.id = sat.subject_id
```
malaffoon: Woohoo a SQL query i can understand in one reading.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/subject/DefaultSubjectService.java
```diff
@@ -19,6 +19,7 @@ public DefaultSubjectService(@NotNull final SubjectRepository repository) {
     }
 
     @Override
+    @Cacheable("subjectdefinitions")
     public List<String> getSubjects() {
         return ImmutableList.copyOf(repository.findAll());
```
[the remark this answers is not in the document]
wtritch: Ah shoot.  Thanks for pointing this out... I cached the wrong service method >_<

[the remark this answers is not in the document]
wtritch: As for your original question... my guess is that we don't want consumers modifying the list?  I'm not sure *why* they would modify the list of subjects...

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/SubjectDefinitionRepository.java
agorina: You may want to update this documentation to reflect this:
https://github.com/SmarterApp/RDW/blob/develop/docs/Runbook.md#system-configuration

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
```diff
@@ -0,0 +1,79 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * This model represents a subject's properties within the scope of an Assessment Type.
+ */
+public class SubjectDefinition {
+    private String asmtTypeCode;
```
esotrope: is it really that laborious to write assessment?

PR #1563, opened 2018-06-20
Load subject display text into translation system

wtritch:
I debated creating separate repositories / translation providers for accommodation translations and subject translations, but thought a UNION with a single DB call felt cleaner.  Thoughts?

### DIFF
--- a/common/src/main/resources/common.sql.yml
+++ b/common/src/main/resources/common.sql.yml
@@ -1,9 +1,17 @@
 sql:
   translation:
     findAllForLanguage: >-
-        select label_code, label
-        from accommodation_translation
-        where language_code=:language_code
+        SELECT
+          label_code,
+          label
+        FROM accommodation_translation
+        WHERE language_code=:language_code
+        UNION
+        SELECT
+          label_code,
+          label
+        FROM subject_translation
+        WHERE 'eng' = :language_code
 
   organization:
     reporting:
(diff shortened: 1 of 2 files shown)


common/src/main/resources/common.sql.yml
```diff
@@ -1,9 +1,17 @@
...
+          label_code,
+          label
+        FROM accommodation_translation
+        WHERE language_code=:language_code
+        UNION
+        SELECT
+          label_code,
+          label
+        FROM subject_translation
+        WHERE 'eng' = :language_code
```
jtreuting: noticing this is different than the accommodatin_tanslation.  so the first part will get translations for other langauges where the subject will only get anything if english is passed in?

common/src/main/resources/common.sql.yml
```diff
@@ -1,9 +1,17 @@
...
     findAllForLanguage: >-
-        select label_code, label
-        from accommodation_translation
-        where language_code=:language_code
+        SELECT
+          label_code,
+          label
+        FROM accommodation_translation
+        WHERE language_code=:language_code
+        UNION
```
agorina: `UNION ALL` ?

common/src/main/resources/common.sql.yml
wtritch: Right, because they're defined in the XML and I don't want to force DevOps to shove translations into the subject definition XML.  So different language/overrides of subject display text will be done via external .yml, in the same file used for all other UI/PDF translation overrides.

common/src/main/resources/common.sql.yml
wtritch: good call, changed

PR #1565, opened 2018-06-20
remove sessions filter for summative assessments

dsnyder-fairway:


dsnyder-fairway:
@jtreuting yup just double checked. It shows all the results for the sessions

PR #1567, opened 2018-06-20
when changing from group to user created group it was including both …

jtreuting:
…userGroupId, groupId in the route

only userGroupId or groupId should be included in the route


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -151,8 +151,10 @@ export class GroupDashboardComponent implements OnInit {
     if (group) {
       if (group.userCreated) {
         parameters.userGroupId = group.id;
+        delete parameters.groupId;
       } else {
         parameters.groupId = group.id;
+        delete parameters.userGroupId;
```
esotrope: thanks. definitely introduced this with the back button changes.

PR #1568, opened 2018-06-20
Move no results to the bottom ONLY when sorting on a "value" column

wtritch:
Before we were pushing "no-results / student-count 0" rows to the bottom when sorting on *any* column.  This PR only pushes those rows to the bottom when sorting on "value" columns where if the row has no results a `-` is displayed.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -372,7 +372,7 @@ export class AggregateReportTableComponent implements OnInit {
         dataColumns.push(
           new Column({ id: 'studentsTested' }),
           new Column({ id: 'achievementComparison', sortable: false }),
-          new Column({ id: 'avgScaleScore' }),
+          new Column({ id: 'avgScaleScore', valueColumn: true }),
           ...this.createPerformanceLevelColumns(performanceLevelsByDisplayType, assessmentDefinition)
         );
         break;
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -609,12 +625,14 @@ class Column implements BaseColumn {
...
                 headerKey = '',
-                headerColor = ''
+                headerColor = '',
+                valueColumn = false
               }) {
     this.id = id;
     this.field = field ? field : id;
     this.sortable = sortable;
     this.visible = visible;
+    this.valueColumn = valueColumn;
```
wtritch: I'm tagging some columns as "value" columns.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -459,7 +459,18 @@ export class AggregateReportTableComponent implements OnInit {
     if (!ascending) {
       rowOrdering = rowOrdering.reverse();
     }
-    return this.getNoResultsLastComparator(rowOrdering.compare, order);
+
+    return this.getColumn(field).valueColumn
+      ? this.getNoResultsLastComparator(rowOrdering.compare, order)
+      : rowOrdering.compare;
```
wtritch: Only push no-results to the bottom if we're sorting on a "value" column.

wtritch:
Yup, unfortunately no specs.

PR #1570, opened 2018-06-21
Remove extra empty page when generating PDF reports.

wtritch:
TIL we can't just page-break willy-nilly on every footer, because that leaves us with an extra empty page on all our PDF reports.

This PR only adds a page-break to a report if there is another report after it.

### DIFF
--- a/report-processor/src/main/resources/templates/batch-reports.html
+++ b/report-processor/src/main/resources/templates/batch-reports.html
@@ -8,25 +8,41 @@
         <div th:replace="fragments::styles-general"></div>
     </head>
     <body class="pdf">
-        <div th:each="studentCompositeReport : ${studentCompositeReports}">
+        <div th:each="studentCompositeReport, iStat : ${studentCompositeReports}"
+             th:class="${iStat.last}? '' : 'break-page'">
             <div th:if="${studentCompositeReport.iabMath != null}"
-                 th:with="report=${studentCompositeReport.iabMath}">
+                 th:with="report=${studentCompositeReport.iabMath}"
+                 th:class="${studentCompositeReport.iabEla != null ||
+                             studentCompositeReport.icaMath != null ||
+                             studentCompositeReport.icaEla != null ||
+                             studentCompositeReport.summativeMath != null ||
+                             studentCompositeReport.summativeEla != null}? 'break-page'">
                 <div th:replace="iab-body::body"></div>
             </div>
             <div th:if="${studentCompositeReport.iabEla != null}"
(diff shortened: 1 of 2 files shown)


report-processor/src/main/resources/templates/batch-reports.html
```diff
@@ -8,25 +8,41 @@
...
+        <div th:each="studentCompositeReport, iStat : ${studentCompositeReports}"
+             th:class="${iStat.last}? '' : 'break-page'">
             <div th:if="${studentCompositeReport.iabMath != null}"
-                 th:with="report=${studentCompositeReport.iabMath}">
+                 th:with="report=${studentCompositeReport.iabMath}"
+                 th:class="${studentCompositeReport.iabEla != null ||
+                             studentCompositeReport.icaMath != null ||
+                             studentCompositeReport.icaEla != null ||
+                             studentCompositeReport.summativeMath != null ||
+                             studentCompositeReport.summativeEla != null}? 'break-page'">
```
wtritch: This feels... sub-optimal.  Ideally we'd iterate over the reports in the composite and switch their display based upon an advertised type.  But that's a larger change than I'm willing to make at this late stage.

report-processor/src/main/resources/templates/batch-reports.html
```diff
@@ -8,25 +8,41 @@
         <div th:replace="fragments::styles-general"></div>
     </head>
     <body class="pdf">
-        <div th:each="studentCompositeReport : ${studentCompositeReports}">
+        <div th:each="studentCompositeReport, iStat : ${studentCompositeReports}"
+             th:class="${iStat.last}? '' : 'break-page'">
```
wtritch: This breaks between students on a multi-student report.

report-processor/src/main/resources/templates/batch-reports.html
```diff
@@ -8,25 +8,41 @@
...
             <div th:if="${studentCompositeReport.iabMath != null}"
-                 th:with="report=${studentCompositeReport.iabMath}">
+                 th:with="report=${studentCompositeReport.iabMath}"
+                 th:class="${studentCompositeReport.iabEla != null ||
```
esotrope: interesting syntax - no need for `:`?

wtritch: Nope, it weirded me out when I googled it as well

esotrope: wouldn't be a complete java templating language without some janky syntax i guess

PR #1571, opened 2018-06-21
fixes group report api integration

esotrope:


webapp/src/main/webapp/src/app/report/report.service.ts
```diff
@@ -79,7 +79,12 @@ export class ReportService {
    * @returns {Observable<Report>} the handle used the get status on the download
    */
   public createGroupExamReport(group: Group, options: ReportOptions): Observable<Report> {
-    const search = Object.assign(group.userCreated ? {userGroupId: group.id} : {groupId: group.id}, options);
+    const search = Object.assign({
+      groupId: {
+        id: group.id,
+        type: group.userCreated ? 'Teacher' : 'Admin'
+      }
```
esotrope: API expects this format now that we have @agorina's fancy changes ; J

esotrope:
passes tests so i'm merging

webapp/src/main/webapp/src/app/report/report.service.ts
agorina: :)

PR #1572, opened 2018-06-21
avoid empty school year when viewing groups with no results

dsnyder-fairway:
debugging showed schoolYear was initially set in updateRouteWithDefaultFilters and then subscribeToRouteChanges was called which resulted in the schoolYear becoming unset

### DIFF
--- a/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
+++ b/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
@@ -91,7 +91,7 @@ export class GroupDashboardComponent implements OnInit {
       const { reload, group, schoolYear, subject, measuredAssessments } = resolvedParameters;
       if (reload) {
         this.group = group;
-        this.schoolYear = Number.parseInt(schoolYear) || undefined;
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -91,7 +91,7 @@ export class GroupDashboardComponent implements OnInit {
       const { reload, group, schoolYear, subject, measuredAssessments } = resolvedParameters;
       if (reload) {
         this.group = group;
-        this.schoolYear = Number.parseInt(schoolYear) || undefined;
+        this.schoolYear = Number.parseInt(schoolYear) || this.schoolYear;
```
esotrope: how does this make sense, the schoolYear should be in the route by now no?

dsnyder-fairway: I am getting resolvedParameters = groupId when clicking on this button
screen shot 2018-06-21 at 9 55 55 am
and my first group does not contain any test results for the most recent year

dsnyder-fairway:
Having trouble reproducing this one now. Continuing to investigate

PR #1579, opened 2018-06-21
 Display districts section for school-typeahead only when district is populated

dsnyder-fairway:


webapp/src/main/webapp/src/app/shared/school/school-typeahead.ts
```diff
@@ -24,7 +24,7 @@ import { TypeaheadMatch } from "ngx-bootstrap";
 
     <ng-template #schoolTemplate let-school="item" let-index="index" let-query="query">
       <p class="mb-0">{{school.name}}</p>
-      <p class="h6"><span class="label label-default">{{'school-typeahead.district-label' | translate}}</span>
+      <p *ngIf="school.districtName" class="h6"><span class="label label-default">{{'school-typeahead.district-label' | translate}}</span>
```
esotrope: shouldn't it always be populated though?

jtreuting: Yeah feels like the problem is that it isn't populated.  Can you see if the API returns it and it's a mapper issue, or if the API isn't returning it and it's a SQL issue.

dsnyder-fairway: yeah from looking at this from the API it looks like using nacostaIP@example.com retrieves the districts on login, but the districts list is empty. Looking into why this is

dsnyder-fairway:
closing until I investigate this more

PR #1581, opened 2018-06-21
 allow switching groups with exclusive subjects when subject is selected

dsnyder-fairway:


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -94,7 +94,12 @@ export class GroupDashboardComponent implements OnInit {
         this.schoolYear = Number.parseInt(schoolYear) || undefined;
         this.updateMeasuredAssessments(measuredAssessments);
       }
-      this.subject = subject;
+      if (group == null || group.subjectCode == null || group.subjectCode === subject) {
+        this.subject = subject;
+      } else {
+        delete this.subject;
+        this.updateRoute(true);
+      }
```
esotrope: what was happening before this change?

dsnyder-fairway: You needed to have the following:
1. Two user created groups with two different subjects (no shared subject) `UserGroup: ELA` and `UserGroup: Math`
2. Select the subject from `All` to whatever subject was in there
3. Switch to the other group
4. Notice that the selector would default to All, but the url would still have the old subject, and there would be no resultant cards displayed

dsnyder-fairway:
@jtreuting @esotrope is this okay to merge?

webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
esotrope: only remaining q is what subject options do we display when the selected group is one subject only? All and the subject? or do we lock it to the one subject?

dsnyder-fairway: We display all and the subject (switch between subjects two groups with two different subjects defaults to "All", but now correctly displays). When switching between a group with one subject and two subjects, the subject will stay selected

PR #1583, opened 2018-06-22
Fix claim score accessing

wtritch:
Looks like this was a quicker fix than I thought.  @jtreuting the aggregations at the top match what I'm seeing in the columns now.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
@@ -92,10 +92,17 @@ export class ResultsByStudentComponent implements OnInit {
     }
 
     return this.assessment.claimCodes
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/results/view/results-by-student/results-by-student.component.ts
```diff
@@ -92,10 +92,17 @@ export class ResultsByStudentComponent implements OnInit {
...
+        new Column({
+          id: 'claim',
+          field: `claimScores.${index}.level`,
+          index: index,
+          claim: code
+        })
+      )
+      .sort(createScorableClaimOrdering(this.assessment.subject)
+        .on((column: Column) => column.claim)
+        .compare);
```
wtritch: Just a quick fix to leave the data claim ordering alone while generating our access `index` values, then I'm visually sorting the columns.

PR #1591, opened 2018-06-22
makes dok work with configurable subjects

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/AssessmentItems.java
```diff
@@ -23,7 +24,11 @@ public static AssessmentItem map(final ResultSet row) throws SQLException {
                 .claimCode(row.getString("claim_code"))
                 .targetCode(row.getString("target_code"))
                 .targetId(row.getInt("target_id"))
-                .depthOfKnowledgeCode(row.getString("dok_level_subject_id"))
+                .depthOfKnowledge(DepthOfKnowledge.builder()
```
esotrope: now we need just the level to lookup the dok name and referenceUrl which used to be stored in en.json

webapp/src/main/webapp/src/app/assessments/assessment-exam.mapper.ts
```diff
@@ -147,32 +147,35 @@ export class AssessmentExamMapper {
...
+  private mapAssessmentItemFromApi(serverItem: any): AssessmentItem {
+    const item: AssessmentItem = new AssessmentItem();
+
+    item.id = serverItem.id;
+    item.bankItemKey = serverItem.bankItemKey;
+    item.position = serverItem.position;
+    item.claim = serverItem.claimCode;
+    item.target = this.formatTarget(serverItem.targetCode);
+    item.targetId = serverItem.targetId;
+    item.depthOfKnowledge = {
```
esotrope: the only real change here

webapp/src/main/webapp/src/app/assessments/items/item-info/item-info.component.html
```diff
@@ -1,39 +1,37 @@
...
+    <strong>{{ 'common.results.assessment-item-target' | translate:item}}: </strong>
     {{ targetDescription }}
   </p>
-  <p>
-    <b>{{ 'item-info.dok-label' | translate}}</b>
-    <!-- TODO:ConfigurableSubjects needs to be subject.${subject}.dok.${item.dokId}.name -->
-    {{ 'item-info.dok-description.' + item.depthOfKnowledge | translate }}
-    <!-- TODO:ConfigurableSubjects needs to be pulled from DB -->
-    <a href="{{ 'item-info.dok-reference.' + item.depthOfKnowledge | translate }}" target="_blank">{{ 'item-info.view-ref' | translate}}</a>
+  <p *ngIf="item.depthOfKnowledge">
```
esotrope: added guard (1) so the specs pass and (2) in case it becomes optional later

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/AssessmentItems.java
esotrope: nope, isn't that the subject definers job?

esotrope: just checked - they match up on my local at least

wtritch: huh... `'item-info.dok-reference.*'` doesn't even exist in en.json...  where the heck was it coming from?

wtritch: ah, looks like it was removed a while ago.  But the old references match the ones in the DB so I think we're good =)

esotrope: yeah i removed them in a recent PR to break things on purpose : J

PR #1592, opened 2018-06-22
dont' include longitudinal and target report types if summative is no…

jtreuting:
…t an assessment type

with summative not include in the configuration options, it looks like this.

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
@@ -36,7 +36,9 @@ export class AggregateReportOptionsService {
         dimensionTypes: serverOptions.dimensionTypes.concat(),
         interimAdministrationConditions: serverOptions.interimAdministrationConditions.concat(),
         queryTypes: [ 'Basic', 'FilteredSubgroup' ],
-        reportTypes: [ AggregateReportType.GeneralPopulation, AggregateReportType.LongitudinalCohort, AggregateReportType.Claim, AggregateReportType.Target ],
+        reportTypes: serverOptions.assessmentTypes.some(x => x.value == 'sum')
+          ? [ AggregateReportType.GeneralPopulation, AggregateReportType.LongitudinalCohort, AggregateReportType.Claim, AggregateReportType.Target ]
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
```diff
@@ -36,7 +36,9 @@ export class AggregateReportOptionsService {
         dimensionTypes: serverOptions.dimensionTypes.concat(),
         interimAdministrationConditions: serverOptions.interimAdministrationConditions.concat(),
         queryTypes: [ 'Basic', 'FilteredSubgroup' ],
-        reportTypes: [ AggregateReportType.GeneralPopulation, AggregateReportType.LongitudinalCohort, AggregateReportType.Claim, AggregateReportType.Target ],
+        reportTypes: serverOptions.assessmentTypes.some(x => x.value == 'sum')
```
jtreuting: have the logic be in the service not in the component.  then the component uses what is there as it;s rule

esotrope: are assessmentTypes complex objects? i thought they were just the string codes

webapp/src/main/webapp/src/app/aggregate-report/query-forms/aggregate-query-form-container.component.ts
```diff
@@ -57,4 +57,25 @@ export class AggregateQueryFormContainerComponent {
...
+  get hasClaimReport(): boolean {
+    return this.hasReportType(AggregateReportType.Claim)
+  }
+
+  get hasTargetReport(): boolean {
+    return this.hasReportType(AggregateReportType.Target)
+  }
+
+  private hasReportType(type: AggregateReportType): boolean {
+    return this.filteredOptions.reportTypes.some(x => x.value == type);
```
esotrope: semantically, `find(..) != null`  is a better fit for this i think.

esotrope: actually instead of introducing all the methods i would be tempted to have a `supportedAssessmentTypes: string[]` array and in the template call `supportedAssessmentTypes.includes('Claim')` etc.

PR #1593, opened 2018-06-22
Adjusted for the schema changes.

agorina:


admin-service/src/test/resources/org/opentestsystem/rdw/admin/search/group/groups.sql
```diff
@@ -4,9 +4,9 @@ insert into district (id, name, natural_id) values
 insert into school (id, district_id, name, natural_id, import_id) values
   (-20, (select id from district where natural_id = 'district_1'), 'Test School 1', 'school_1', -1);
 
-insert into subject (id, code) values
-  (-1, 'subject_1'),
-  (-2, 'subject_2');
+insert into subject (id, code, update_import_id, migrate_id) values
```
agorina: These two columns have been changed to `NOT NULL`.

PR #1597, opened 2018-06-23
Task/config subjects org export

jtreuting:


webapp/src/main/webapp/src/app/organization-export/organization/organization.service.ts
```diff
@@ -5,11 +5,8 @@ import { OrganizationMapper } from "./organization.mapper";
 import { CachingDataService } from "../../shared/data/caching-data.service";
 import { map } from 'rxjs/operators';
 import { forkJoin } from 'rxjs/observable/forkJoin';
-import { ReportingServiceRoute } from '../../shared/service-route';
 import { OrganizationService as ExtendOrganizationService } from "../../shared/organization/organization.service";
 
-const ServiceRoute = ReportingServiceRoute;
```
jtreuting: unrelated but saw it wasn't used.  the base class access the ReportingServiceRoute directly from there

PR #1598, opened 2018-06-23
integrates group dashboard with configurable subjects

esotrope:


webapp/src/main/webapp/src/app/shared/color.service.ts
```diff
@@ -52,7 +52,8 @@ export class ColorService {
    */
   // TODO:ConfigurableSubjects drive through subject service
   getPerformanceLevelColorsByAssessmentTypeCode(code: string, performanceLevel: number): string {
-    return PerformanceLevelColorsByAssessmentTypeCode.get(code)[ performanceLevel - 1 ];
+    return 'gray-darkest';
+    // return PerformanceLevelColorsByAssessmentTypeCode.get(code)[ performanceLevel - 1 ];
```
esotrope: intentionally breaking the color service here just to make things we need to change stand out

webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-assessment-card.component.ts
```diff
@@ -5,16 +5,26 @@ import { ColorService } from '../../shared/color.service';
 import { GradeCode } from '../../shared/enum/grade-code.enum';
 import { ExamStatisticsCalculator } from '../../assessments/results/exam-statistics-calculator';
 
+export interface GroupCard {
```
esotrope: made wrapper model since cards are readonly on the input data and the data does not change over the lifetime of a card instance

PR #1600, opened 2018-06-24
join on subject to get code instead of using subject_id and hard-code…

jtreuting:
…d enum that mapped to code

image

image

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Assessments.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Assessments.java
@@ -17,7 +17,9 @@ public final class Assessments {
             "claim1_score_code",
             "claim2_score_code",
             "claim3_score_code",
-            "claim4_score_code"
+            "claim4_score_code",
+            "claim5_score_code",
+            "claim6_score_code"
     };
 
     private Assessments() {
@@ -36,7 +38,7 @@ public final class Assessments {
                 .label(row.getString(prefix + "label"))
                 .gradeCode(row.getString(prefix + "grade_code"))
                 .type(type)
-                .subject(Subject.valueOf(row.getInt(prefix + "subject_id")))
+                .subjectCode(row.getString(prefix + "subject_code"))
                 .schoolYear(row.getInt(prefix + "school_year"));
 
         // IABs do not have claims
@@ -61,7 +63,7 @@ public final class Assessments {
                         row.getInt(prefix + "min_score"),
                         row.getInt(prefix + "cut_point_1"),
                         row.getInt(prefix + "cut_point_2"),
-                        row.getInt(prefix + "cut_point_3"),
+                        row.getInt(prefix + "cut_point_3"), // TODO:ConfigurableSubjects add cut_point_4, cut_point_5?
                         row.getInt(prefix + "max_score")
                 ));
 
--- a/reporting-service/src/main/resources/application.sql.yml
+++ b/reporting-service/src/main/resources/application.sql.yml
@@ -102,6 +102,7 @@ sql:
           FROM asmt a
             JOIN exam e ON e.asmt_id = a.id
             JOIN school s ON e.school_id = s.id
+            JOIN subject su ON a.subject_id = su.id
             JOIN student st ON e.student_id = st.id
             JOIN school isch ON st.inferred_school_id = isch.id
             $[groupJoinPlaceholder]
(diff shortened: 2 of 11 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Assessments.java
```diff
@@ -36,7 +38,7 @@ private Assessments() {
                 .label(row.getString(prefix + "label"))
                 .gradeCode(row.getString(prefix + "grade_code"))
                 .type(type)
-                .subject(Subject.valueOf(row.getInt(prefix + "subject_id")))
+                .subjectCode(row.getString(prefix + "subject_code"))
```
jtreuting: no longer use the Subject enum to get the code and get the code from the SQL

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -124,6 +126,10 @@ export class AverageScaleScoreComponent {
   }
 
   get performanceLevels(): ExamStatisticsLevel[] {
+    if (this.statistics == null) {
```
jtreuting: needed some guards since the subjectDefinition is not set until the ngOnInit

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/TestData.java
```diff
@@ -172,7 +172,7 @@
                 .label("ica1")
                 .gradeCode("g1")
                 .schoolYear(1997)
-                .subject(Subject.MATH)
+                .subjectCode(Subject.MATH.code())
```
agorina: Should this be replaced with the string values so that we could remove `Subject` enum?

jtreuting: Yep agreed

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Assessments.java
```diff
@@ -61,7 +63,7 @@ private Assessments() {
                         row.getInt(prefix + "min_score"),
                         row.getInt(prefix + "cut_point_1"),
                         row.getInt(prefix + "cut_point_2"),
-                        row.getInt(prefix + "cut_point_3"),
+                        row.getInt(prefix + "cut_point_3"), // TODO:ConfigurableSubjects add cut_point_4, cut_point_5?
```
wtritch: yeah we should probably add all the cutpoints, and wrap these with getNullable

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -319,18 +324,20 @@ sql:
...
         a.claim4_score_code,
+        a.claim5_score_code,
+        a.claim6_score_code,
         a.min_score,
         a.max_score,
         a.cut_point_1,
         a.cut_point_2,
         a.cut_point_3,
-        EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item
+        EXISTS(SELECT 1 FROM item i WHERE i.asmt_id = a.id AND i.type = 'WER') as has_wer_item,
```
wtritch: Should this EXISTS also limit to ELA?

malaffoon: Why limit it to ELA? Is there an assumption that WER's only occur in ELA assessments?

wtritch:
👍

reporting-service/src/main/resources/application.sql.yml
wtritch: I believe so, in that we only support Writing Trait Scores for SBAC ELA assessments, not for configurable subjects, I believe.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Assessments.java
agorina: The nullable comment is applicable to the min/max as well, right?

reporting-service/src/main/resources/application.sql.yml
jtreuting: this didn't change.  i'm not sure if the front end uses the combination of "has_wer_item" and assessment type to determine if Writing Trait Scores should be shown or not.  i'd have to look.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Assessments.java
jtreuting: okay, i'll get this updated

PR #1601, opened 2018-06-25
display "0%" instead of empty "%" when score count is 0

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/model/assessment-item.model.ts
```diff
@@ -48,7 +48,7 @@ export class AssessmentItem {
     const scoreCount = this.scores.reduce((count, score) =>
       score.points >= 0 ? count + 1 : count,
       0);
-    return this.scores.length > 0
+    return this.scores.length > 0 && scoreCount > 0
```
jtreuting: i think this can just be scoreCount > 0 since that takes into account the scores.length when determining it

dsnyder-fairway: good call. Updated

PR #1602, opened 2018-06-25
add cutpoint 4 and 5 and make nullable

jtreuting:


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/Assessments.java
```diff
@@ -59,11 +62,13 @@ private Assessments() {
 
     public static <A extends Assessment.Builder> A mapWithCutPoints(final ResultSet row, final A builder, final String prefix) throws SQLException {
         map(row, builder, prefix)
-                .cutPoints(ImmutableList.of(
+                .cutPoints(Arrays.asList(
```
jtreuting: since these are nullable can't use the Immutable list anymore.  i am personally not saddened by this.

PR #1604, opened 2018-06-25
config subjects for embargo

jtreuting:
Slight change of data returned from the API, looking for thoughts on it?

image

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
@@ -74,18 +74,20 @@ public class JdbcEmbargoRepository implements EmbargoRepository {
                                     .individualEnabled(getNullable(rs, rs.getBoolean("individual")))
                                     .aggregateEnabled(getNullable(rs, rs.getBoolean("aggregate")))
                                     .schoolYear(rs.getInt("school_year"));
-                            for (final Subject subject : Subject.values()) {
-                                innerBuilder.examCount(subject, 0);
-                            }
+//                            for (final Subject subject : Subject.values()) {
+//                                innerBuilder.examCount(subject, 0);
+//                            }
                             return innerBuilder;
                         } catch (SQLException e) {
                             throw new DataRetrievalFailureException(e.getMessage());
                         }
                     });
-                    final Integer subjectId = getNullable(rs, rs.getInt("subject_id"));
-                    if (subjectId != null) {
-                        builder.examCount(Subject.valueOf(subjectId), rs.getLong("count"));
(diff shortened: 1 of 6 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -74,18 +74,18 @@ public JdbcEmbargoRepository(@Qualifier("warehouseJdbcTemplate") final NamedPara
                                     .individualEnabled(getNullable(rs, rs.getBoolean("individual")))
                                     .aggregateEnabled(getNullable(rs, rs.getBoolean("aggregate")))
                                     .schoolYear(rs.getInt("school_year"));
-                            for (final Subject subject : Subject.values()) {
```
jtreuting: this pre-filled all the subjects with 0.  since the UI can handle it not being there and just display a 0 if it;s missing i didn't think it was necessary to get all the subjects up front.

admin-service/src/test/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepositoryIT.java
```diff
@@ -103,13 +102,12 @@ public void findAllShouldIncludeExamCounts() {
...
-        assertThat(state.getExamCounts().get(Subject.ELA)).isEqualTo(1);
+        assertThat(state.getExamCounts().get("Math")).isEqualTo(2);
+        assertThat(state.getExamCounts().get("ELA")).isEqualTo(1);
 
         final Embargo district = ((List<Embargo>) embargos).get(1);
-        assertThat(district.getExamCounts()).hasSize(2);
-        assertThat(district.getExamCounts().get(Subject.MATH)).isEqualTo(2);
-        assertThat(district.getExamCounts().get(Subject.ELA)).isEqualTo(0);
+        assertThat(district.getExamCounts()).hasSize(1);
+        assertThat(district.getExamCounts().get("Math")).isEqualTo(2);
```
jtreuting: updated the test since the empty ELA is not returned with a 0 now

webapp/src/main/webapp/src/app/admin/embargo/embargo-table.component.ts
```diff
@@ -57,6 +64,11 @@ export class EmbargoTable implements OnInit {
       this.overridingEmbargo ? this.overridingEmbargo.aggregateEnabled : undefined);
   }
 
+  getExamCount(embargo: Embargo, subjectCode) {
+    const examCount = embargo.examCountsBySubject[subjectCode];
+    return examCount || 0;
```
jtreuting: handle the 0 here

PR #1605, opened 2018-06-26
get claim order for assessment results

dsnyder-fairway:
hotfix


wtritch:
hm... are we planning on deploying a hotfix independently of configurable subjects?
And should we fix this on `develop` as well?

dsnyder-fairway:
@wtritch yeah I think we are planning on a hotfix independent of configurable subjects. I have some changes ready if we feel I should make this to develop as well.

PR #1608, opened 2018-06-26
config subject updates for instructional resources

jtreuting:
pretty straight forward.  needed to return the assessment subject code so that the performance levels could be displayed

### DIFF
--- a/admin-service/src/main/resources/application.sql.yml
+++ b/admin-service/src/main/resources/application.sql.yml
@@ -177,6 +177,7 @@ sql:
       SELECT DISTINCT
         GROUP_CONCAT(DISTINCT a.label ORDER BY a.id DESC SEPARATOR ', ') AS asmt_label,
         a.type_id AS asmt_type_id,
+        su.code as asmt_subject_code,
         ir.asmt_name,
         ir.performance_level,
         ir.org_level,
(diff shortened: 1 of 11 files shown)


webapp/src/main/webapp/src/app/admin/instructional-resource/create-instructional-resource.modal.ts
```diff
@@ -71,6 +76,11 @@ export class CreateInstructionalResourceModal implements OnDestroy {
     ).subscribe(() => {
       this.cancel();
     });
+
+    this.subjectService.getSubjectDefinitions()
+      .subscribe(subjectDefinitions => {
+        this.subjectDefinitions = subjectDefinitions;
```
jtreuting: getting them all because we can choose any assessment and then we will have them already

admin-service/src/main/resources/application.sql.yml
```diff
@@ -227,12 +230,13 @@ sql:
             )
           )
         )
-      GROUP BY a.type_id, ir.asmt_name, ir.performance_level, ir.org_level, ir.resource, ir.org_id, district_group.name, district.name, school_group.name
+      GROUP BY a.type_id, su.code, ir.asmt_name, ir.performance_level, ir.org_level, ir.resource, ir.org_id, district_group.name, district.name, school_group.name
 
     findOne: >-
       SELECT
         GROUP_CONCAT(DISTINCT a.label ORDER BY a.id DESC SEPARATOR ', ') AS asmt_label,
```
agorina: Unrelated to your change, but this explains why it takes a long time to load the instructional resources.

PR #1609, opened 2018-06-27
 support state level instructional resources for user created groups

dsnyder-fairway:


reporting-service/src/main/resources/application-no-permission.sql.yml
```diff
@@ -200,6 +200,34 @@ sql:
...
+          ON ir.org_level = 'DistrictGroup'
+          AND district_group.id = ir.org_id
+        LEFT JOIN district
+          ON ir.org_level = 'District'
+          AND district.id = ir.org_id
+        LEFT JOIN school_group
+          ON ir.org_level = 'SchoolGroup'
+          AND school_group.id = ir.org_id
+      WHERE asmt.id = :asmt_id
+        AND
```
jtreuting: I don't think we need a second SQL query and could use the main one and pass in school_id NULL.  Should be able to edit the line `AND school.id = :school_id` to be `AND (:school_id IS NULL OR school.id = :school_id)`

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/instructionalresource/JdbcInstructionalResourceRepository.java
```diff
@@ -41,4 +44,20 @@ public JdbcInstructionalResourceRepository(final NamedParameterJdbcTemplate temp
                                 .build()
         );
     }
+
+    @Override
+    public Collection<InstructionalResource> findByAssessment(final long assessmentId) {
```
jtreuting: Also think we could have a single method with schoolId being NULL

reporting-service/src/main/resources/application-no-permission.sql.yml
dsnyder-fairway: that was my first attempt, but that resulted in 8 rows being returned for a single instructional resource

dsnyder-fairway: I had to `UNION` these two selects. See https://stackoverflow.com/a/1255523

dsnyder-fairway:
Attn @agorina

PR #1612, opened 2018-06-27
Task/fix claim score overview ordering

wtritch:


webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -160,3 +172,12 @@ export class AverageScaleScoreComponent {
...
 }
+
+/**
+ * This class provides an orderable reference to claim score statistics.
+ */
+class ClaimReference {
+  code: string;
+  dataIndex: number;
+  stats: ClaimStatistics;
+}
```
wtritch: Just created a sortable wrapper around our claim stats that we can order independently from our input data models.

PR #1614, opened 2018-06-27
update average scale score component to do claim ordering once assess…

jtreuting:
…ment and statistics are set

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import { Component, EventEmitter, Input, Output } from '@angular/core';
 import { AssessmentExam } from '../model/assessment-exam.model';
 import { ExamStatistics, ExamStatisticsLevel } from '../model/exam-statistics.model';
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -23,13 +23,17 @@ enum ScoreViewState {
...
 })
-export class AverageScaleScoreComponent implements OnInit {
+export class AverageScaleScoreComponent {
 
   @Input()
   showValuesAsPercent: boolean = true;
 
   @Input()
-  public assessmentExam: AssessmentExam;
+  set assessmentExam(value: AssessmentExam) {
```
jtreuting: change to set method and renamed below to use private variable

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -23,13 +23,17 @@ enum ScoreViewState {
...
+    this._assessmentExam = value;
```
wtritch: Does this also need a getter?  Does the HTML reference it at all?

jtreuting: it did reference it, but when i looked at it, it was using it to get the assessment and we have a getter for that purpose, so decided to use that

PR #1615, opened 2018-06-27
 handle invalid schoolYears (defaults to most recent)

dsnyder-fairway:
I was not able to find out how we get schoolYear=undefined in the url, but this update handles `schoolYear=`, `schoolYear=abc`, `schoolYear=2`, and `schoolYear` not specified


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -77,11 +77,13 @@ export class GroupDashboardComponent implements OnInit {
           || (previousParameters.groupId != null && previousParameters.groupId != groupId)
           || (previousParameters.userGroupId != null && previousParameters.userGroupId != userGroupId);
 
+        const exit = isNaN(Number(schoolYear)) || schoolYear === '';
```
jtreuting: can we make this variable name more descriptive?  for some reason it's hard for me to connect the dots with the generic name exit.

dsnyder-fairway: Yeah I had trouble with the naming too. Really the intent of this, if true, is to do nothing and have the handling be done in `updateRouteWithDefaultFilters`. How about `doNothing` or `defaultFiltersNeeded`?

PR #1618, opened 2018-06-28
correcting CSV export of null values

dsnyder-fairway:


report-processor/src/main/resources/application.sql.yml
```diff
@@ -215,21 +215,21 @@ sql:
...
-        (CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'Asian') THEN 'Yes' ELSE 'No' END ),
-        (CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'BlackOrAfricanAmerican') THEN 'Yes' ELSE 'No' END ),
-        (CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'White') THEN 'Yes' ELSE 'No' END ),
-        (CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'NativeHawaiianOrOtherPacificIslander') THEN 'Yes' ELSE 'No' END ),
-        (CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'TwoOrMoreRaces') THEN 'Yes' ELSE 'No' END ),
-        (CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'Filipino') THEN 'Yes' ELSE 'No' END ),
+        (CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'HispanicOrLatinoEthnicity') THEN 'Yes'  WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id) THEN 'No' ELSE '' END ),
```
agorina: Could you please clarify what we need to return here?
It seems to me that we would need something like this:
```
(CASE WHEN EXISTS (SELECT 1 from student_ethnicity WHERE student_id = st.id AND ethnicity_code = 'HispanicOrLatinoEthnicity') THEN 'Yes' ELSE '' END )
```
But I maybe missing some requirements, that is why I am asking.

dsnyder-fairway: Yeah, what I am trying to achieve is getting the student ethnicity, if any is available. Say a student has ethnicity code `HispanicOrLatinoEthnicity`, then they will have `HispanicOrLatinoEthnicity` of `Yes` and `No` for everything else. If they have no ethnicities for the student, then all of the columns would be empty (`''`) since the student's ethnicity is not available.

agorina: A few comments:
1. keep in mind that a student could have more than one ethnicity
2. why not return "Yes" for the columns with the student's ethnicity and "" for everything else?

dsnyder-fairway: Response on:
`1.` yup
`2.` I think the absence of a student being on the ethnicity table may have a different meaning than a student having a few (or one) ethnicities on the table

dsnyder-fairway: 👍 I'll change this shortly

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -345,8 +361,11 @@ export class CsvBuilder {
       this.withColumn(
         ethnicity,
         (item) => {
-          const polarEnum = getExam(item).student.ethnicityCodes.some(code => code == ethnicity) ? 1 : 2;
-          return this.getPolarTranslation(polarEnum);
+          if (getExam(item).student.ethnicityCodes.some(code => code == ethnicity)) {
+            return this.getPolarTranslation(1);
+          } else {
+            return '';
+          }
```
dsnyder-fairway: note these changes were also needed to have consistent ethnicity reporting

PR #1619, opened 2018-06-28
Modified aggregate report queries to support optional data elements.

agorina:


aggregate-service/src/main/resources/claim-report-template.sql.yml
```diff
@@ -126,10 +126,10 @@ sqlbuilder:
                 select: >-
                      'Gender' AS dimension,
                       s.gender_id  AS dimension_id,
-                      gn.code AS dimension_code
+                      CASE WHEN s.gender_id IS NULL THEN 'undefined' ELSE gn.code END AS dimension_code
                 join: >-
                      JOIN student s ON e.student_id = s.id
-                     JOIN gender gn on gn.id = s.gender_id
+                     LEFT JOIN gender gn on gn.id = s.gender_id
```
agorina: I have decided to take this approach instead of introducing a new gender code.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcCustomAggregateReportRepositoryRST.java
```diff
@@ -506,11 +506,30 @@ public void itShouldFindByQueryWithMigrantStatusDimensionAnELASFilter() {
...
+                .organizationType(School)
+                .organizationIds(of(-9L))
+                .dimensionType(Gender)
+                .build(), School).getPayload();
+
+        assertThat(rows.size()).isEqualTo(genderCount + 1);
+        assertThat(rows.stream().filter(r -> r.getDimension().getCode().equals("undefined")).count()).isEqualTo(1);
+        assertThat(rows.stream().filter(r -> r.getDimension().getCode().equals("gender-18")).count()).isEqualTo(1);
+        assertThat(rows.stream().filter(r -> r.getDimension().getCode().equals("gender-19")).count()).isEqualTo(1);
     }
```
agorina: I know, I know - I am being lazy with the RSTs here. If somebody strongly objects I will add more, but it feels to me that it is not such a high priority for now.

aggregate-service/src/main/resources/claim-report-template.sql.yml
agorina: Does something need to be changed in the UI to support this new code?

aggregate-service/src/test/resources/TargetEntitiesSetup.sql
```diff
@@ -30,18 +30,18 @@ INSERT INTO asmt (id, grade_id, subject_id, type_id, school_year, name, label, c
...
-  (-21,  'OA|D',  '2'),
-  (-22,  'OA|A',  '2'),
-  (-31,  'NF|C',  '3'),
-  (-32,  'MD|D',  '3'),
-  (-33,  'MD|E',  '3'),
-  (-34,  'OA|E',  '3'),
-  (-41,  'OA|E',  '4'),
-  (-42,  'MD|D',  '4'),
-  (-43,  'OA|A',  '4');
+INSERT INTO target(id, natural_id, claim_code, subject_id) VALUES
```
agorina: This tells me that RSTs are disabled in the CI/reporting. (@wtritch ) 

I now refuse to write any more of them! Never ever again!

wtritch: >_< I'll go re-enable them

aggregate-service/src/main/resources/claim-report-template.sql.yml
wtritch: hm... so before we had NULL, Male, Female, NonBinary and now we have 'undefined', Male, Female, NonBinary?  ...why introduce 'undefined' ?

wtritch: n/m, looks like NULL wasn't fully supported in the UI with code like:
```
withGender(getStudent: (item: any) => Student) {
    return this.withColumn(
      this.translateService.instant('csv-builder.gender'),
      (item) => this.translateService.instant(`common.gender.${getStudent(item).genderCode}`)
    )
  }
```

However, with migrant status it looks like we mapped NULL to 'undefined' in the UI... sometimes:
```
migrantStatus: this.toBooleanCode(serverStudent.migrantStatus)
```
and other times we allowed NULL to percolate through to the views which interpreted it as FALSE:
```
  withMigrantStatus(getExam: (item: any) => Exam) {
    return this.withColumn(
      this.translateService.instant('csv-builder.migrant-status'),
      (item) => {
        const polarEnum = getExam(item).migrantStatus ? 1 : 2;
        return this.getPolarTranslation(polarEnum);
      }
    );
  }
```

I'd be tempted to leave the API dealing with NULLs and let the UI map NULL to whatever it wants...  although it looks like the UI needs a good once-over for consistency.  Thoughts?

agorina: We did not have NULL before.  Instead records with the  NULL gender were silently excluded.

agorina: Let's chat offline to get on the same page :)

agorina: Discussed offline. I will create a JIRA to address this new gender code in the UI.

PR #1621, opened 2018-06-28
fix csv export of target column

dsnyder-fairway:


webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -370,10 +370,16 @@ export class CsvBuilder {
...
+  withTarget(getAssessment: (item: any) => Assessment, getAssessmentItem: (item: any) => AssessmentItem) {
     return this.withColumn(
       this.translateService.instant('csv-builder.target'),
-      (item) => this.translateService.instant('common.results.assessment-item-target', getAssessmentItem(item))
+      (item) => {
+        const assessment = getAssessment(item);
+        const assessmentItem = getAssessmentItem(item);
+        const target = this.translateService
+          .instant(`subject.${assessment.subject}.claim.${assessmentItem.claim}.target.${assessmentItem.targetNaturalId}.name`);
+        return this.translateService.instant('common.results.assessment-item-target', { target: target });
```
wtritch: For CSV export I believe we just return the name without the title. (Example "A" rather than "Target A")

dsnyder-fairway: done 👍

PR #1626, opened 2018-06-28
avoid issue of embargoed popover text appearing on Export CSV while n…

dsnyder-fairway:
…ot embargoed

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.html
@@ -163,12 +163,12 @@
         </button>
       </li>
       <li>
-        <button class="btn btn-default btn-xs"
-                [ngClass]="{disabled: exportDisabled}"
+        <button *ngIf="!exportDisabled" class="btn btn-default btn-xs"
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/assessments.component.html
```diff
@@ -163,12 +163,12 @@ <h3>{{'assessments.results.title' | translate}}</h3>
         </button>
       </li>
       <li>
-        <button class="btn btn-default btn-xs"
-                [ngClass]="{disabled: exportDisabled}"
+        <button *ngIf="!exportDisabled" class="btn btn-default btn-xs"
+                (click)="callExport()"><i class="fa fa-bold fa-table"></i> {{'common.export-csv' | translate}}</button>
+        <button *ngIf="exportDisabled" class="btn btn-default btn-xs disabled"
                 popover="{{'common.embargo-export-disabled' | translate}}"
-                [triggers]="exportDisabled ? 'mouseenter:mouseleave' : ''"
```
dsnyder-fairway: for certain users (I imagine users with embargoing privileges) this exportDisabled trigger would always occur on mouseenter:mouseleave and display the popover text. However, they were still able to export and the button did not appear disabled.

PR #1630, opened 2018-06-29
Bugfix/config subjects aggregate

jtreuting:
fixed query builder new subject selection and preview

image

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
@@ -19,6 +19,7 @@ import { Subgroup } from './subgroup/subgroup';
 import { Claim } from './aggregate-report-options.service';
 import { AggregateReportService } from './aggregate-report.service';
 import { Utils } from "../shared/support/support";
+import { SubjectDefinition } from '../subject/subject';
 
 const MaximumOrganizations = 3;
 
@@ -53,9 +54,9 @@ export class AggregateReportTableDataService {
       const studentsTested = context.row.studentsTested;
       const performanceLevelCounts = [];
       const performanceLevelPercents = [];
-      const performanceLevelCount = Math.floor(studentsTested / context.assessmentDefinition.performanceLevelCount);
-      const performanceLevelPercent = Math.floor(100 / context.assessmentDefinition.performanceLevelCount);
-      for (let i = 0; i < context.assessmentDefinition.performanceLevelCount; i++) {
+      const performanceLevelCount = Math.floor(studentsTested / context.subjectDefinition.performanceLevelCount);
+      const performanceLevelPercent = Math.floor(100 / context.subjectDefinition.performanceLevelCount);
+      for (let i = 0; i < context.subjectDefinition.performanceLevelCount; i++) {
         performanceLevelCounts.push(performanceLevelCount);
         performanceLevelPercents.push(performanceLevelPercent);
       }
@@ -83,6 +84,7 @@ export class AggregateReportTableDataService {
   }
 
   createSampleData(assessmentDefinition: AssessmentDefinition,
+                    subjectDefinition: SubjectDefinition,
                     settings: AggregateReportFormSettings,
                     options: AggregateReportOptions): AggregateReportItem[] {
 
@@ -197,6 +199,7 @@ export class AggregateReportTableDataService {
     const rowTemplate: AggregateReportItem = this.createRowTemplate();
     return this.createRows(rowTemplate, valueProviders, {
       assessmentDefinition: assessmentDefinition,
+      subjectDefinition: subjectDefinition,
       settings: settings,
       options: options,
       itemId: 1
@@ -299,6 +302,7 @@ interface ValueProvider {
 
 class RowContext {
   assessmentDefinition: AssessmentDefinition;
+  subjectDefinition: SubjectDefinition;
   settings: AggregateReportFormSettings;
   options: AggregateReportOptions;
   itemId: number;
--- a/webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
@@ -21,6 +21,9 @@ import { Observer } from "rxjs/Observer";
 import { Observable } from "rxjs/Observable";
 import { AggregateReportColumnOrderItemProvider } from "../aggregate-report-column-order-item.provider";
 import { OrderableItem } from "../../shared/order-selector/order-selector.component";
+import { SubjectDefinition } from '../../subject/subject';
+import { SubjectService } from '../../subject/subject.service';
+import { forkJoin } from 'rxjs/observable/forkJoin';
 
 /**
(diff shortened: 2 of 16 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
```diff
@@ -83,6 +84,7 @@ export class AggregateReportTableDataService {
   }
 
   createSampleData(assessmentDefinition: AssessmentDefinition,
+                    subjectDefinition: SubjectDefinition,
```
jtreuting: still working on getting rid of assessmentDefinition completely, will followup with more

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
```diff
@@ -172,15 +173,15 @@ export class AggregateReportTableExportService {
       if (options.performanceLevelDisplayType === PerformanceLevelDisplayTypes.Grouped) {
         header = this.translateService.instant(`aggregate-report-table.columns.grouped-performance-level-prefix.${level}`);
       } else {
-        // TODO:ConfigurableSubjects hardcoded Math here for now
-        header = this.translateService.instant(`subject.Math.asmt-type.${options.assessmentDefinition.typeCode}.level.${level}.short-name`);
+        header = this.translateService.instant(`subject.${options.subjectDefinition.subject}.asmt-type.${options.subjectDefinition.assessmentType}.level.${level}.short-name`);
       }
-      return header + ' ' + this.translateService.instant(`subject.Math.asmt-type.${options.assessmentDefinition.typeCode}.level.${level}.suffix`);
+      const suffix = this.translateService.instant(`subject.${options.subjectDefinition.subject}.asmt-type.${options.subjectDefinition.assessmentType}.level.${level}.suffix`);
+      return header + (suffix ? ' ' + suffix : '');
```
jtreuting: make suffix optional, so no extra space if blank

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -38,9 +38,9 @@
              class="text-center">
           <div class="label"
                [ngClass]="column.headerColor">
-            {{ column.headerKey | translate }}
+            {{ column.headerText }}
```
jtreuting: changed to do the tranlsation on the component to be consistent since I need to see if there is a suffix translation

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -106,6 +109,16 @@ export class AggregateReportTableComponent implements OnInit {
     this._initialized = true;
   }
 
+  @Input()
+  set subjectDefinition(value: SubjectDefinition) {
+    console.log('subjectDefinition', value);
```
jtreuting: oops, will remove

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -125,15 +138,18 @@ export class AggregateReportTableComponent implements OnInit {
   }
 
   get cutPoint(): number {
-    return this.table.assessmentDefinition.performanceLevelGroupingCutPoint;
+    return this.subjectDefinition.performanceLevelStandardCutoff;
+    //return this.table.assessmentDefinition.performanceLevelGroupingCutPoint;
```
jtreuting: same here, will remove

webapp/src/main/webapp/src/app/shared/assessment/performanc-level/performance-level-distribution-chart.component.ts
```diff
@@ -23,7 +24,7 @@ export class PerformanceLevelDistributionChart implements OnInit {
   private _visible: boolean = true;
   private _performanceLevelBarsByDisplayType: Map<string, Map<boolean, PerformanceLevelBars>> = new Map();
 
-  constructor(private colorService: ColorService) {
+  constructor(private translateService: TranslateService) {
```
jtreuting: get color from translations instead of color service

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
wtritch: I wonder if we'll be able to since it contains things like default column orders...  might need to munge the two together somehow or separate their concerns in aggregate reports.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
```diff
@@ -179,6 +196,7 @@ export abstract class BaseAggregateQueryFormComponent implements OnInit, OnDestr
       options: this.aggregateReportOptions,
       rows: this.tableDataService.createSampleData(
         this.getAssessmentDefinition(),
+        this.subjectDefinition,
```
wtritch: Careful with stuff like this, because it may not be initialized yet... 99.9% of the time it will be but...

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
```diff
@@ -172,15 +173,15 @@ export class AggregateReportTableExportService {
       if (options.performanceLevelDisplayType === PerformanceLevelDisplayTypes.Grouped) {
         header = this.translateService.instant(`aggregate-report-table.columns.grouped-performance-level-prefix.${level}`);
       } else {
-        // TODO:ConfigurableSubjects hardcoded Math here for now
-        header = this.translateService.instant(`subject.Math.asmt-type.${options.assessmentDefinition.typeCode}.level.${level}.short-name`);
+        header = this.translateService.instant(`subject.${options.subjectDefinition.subject}.asmt-type.${options.subjectDefinition.assessmentType}.level.${level}.short-name`);
```
wtritch: Could this throw an NPE if options.subjectDefinition isn't resolved yet?

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-table-data.service.ts
jtreuting: yeah my hope was to be able to but if i run into anything where i can't then they might live side by side

webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
jtreuting: any suggestions on a pattern to follow to avoid it?  we can talk offline if you have thoughts

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
jtreuting: hmmm, yes it could.  not sure what we should/could show if subjectDefinition is null.  things would not be good and we couldn't provide headers at all or know how many we should be showing

PR #1633, opened 2018-06-29
correctly removing writing trait scores dropdown from non-ELA assessm…

dsnyder-fairway:
…ents

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/items/item-tab.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/items/item-tab.component.html
@@ -10,7 +10,7 @@
       </item-scores>
     </div>
   </tab>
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/items/item-tab.component.html
```diff
@@ -10,7 +10,7 @@
       </item-scores>
     </div>
   </tab>
-  <tab *ngIf="item.performanceTaskWritingType && responsesAssessmentItem && subject === 'ELA'" (select)="selectTab('Writing Trait Scores')"
+  <tab *ngIf="item.performanceTaskWritingType && responsesAssessmentItem" (select)="selectTab('Writing Trait Scores')"
```
dsnyder-fairway: previous change was incorrect

PR #1634, opened 2018-06-30
change to vertical display if we have more than two subjects

dsnyder-fairway:


webapp/src/main/webapp/src/app/user-group/user-group-form.component.html
```diff
@@ -24,6 +24,7 @@
     <label for="subjects">{{'user-group-form.field.subjects.label' | translate}}</label>
     <sb-button-group id="subjects"
                      name="subjects"
+                     [vertical]="options.subjects.length > 2"
```
dsnyder-fairway: we might want this as well on the aggregate report forms. Right now it is always vertical

PR #1636, opened 2018-07-02
support "Not Stated" options for boolean options and provide appropri…

dsnyder-fairway:
…ate translations for undefined subgroups

I removed Strict boolean related classes as we no longer are using this


jtreuting:
@dsnyder-fairway Does this add the 3rd option to all the forms that previously only had 2?  Like advanced filters and customize subgroups?  I'm not sure if we want that in all places which is why i'm asking.

dsnyder-fairway:
Yes, this allows users to select Not Stated for LEP, IEP, and Economic Disadvantage. We also return the correct translated string on the results

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/DefaultReportOptionService.java
```diff
@@ -69,18 +65,17 @@
...
     public ReportOptions getReportOptions() {
         final List<String> booleans = codesOf(booleanRepository.findAll());
-        final List<String> strictBooleans = codesOf(strictBooleanRepository.findAll());
         return ReportOptions.builder()
                 .assessmentGrades(codesOf(assessmentGradeService.getAssessmentGrades()))
                 .completenesses(codesOf(completenessService.getCompletenesses()))
                 .claims(claimService.findAll())
                 .dimensionTypes(DimensionType.orderedDisplayValues())
-                .economicDisadvantages(strictBooleans)
+                .economicDisadvantages(booleans)
```
agorina: I do not think this is what we want. @jtreuting , please correct me if needed, but I think we wanted to keep this as is.

jtreuting: We don't want to be able to filter by not stated, just handle it on the display as a row in the results.

dsnyder-fairway: fixing this pr now

webapp/src/main/webapp/src/app/aggregate-report/support.ts
```diff
@@ -85,15 +85,19 @@ export function organizationOrdering<T>(organizationGetter: (item: T) => Organiz
 
 export function subgroupOrdering<T>(subgroupGetter: (item: T) => Subgroup, options: AggregateReportOptions): Ordering<T> {
 
+  function getNullableOption(option: string[]) {
+    return option.concat('undefined');
+  }
```
dsnyder-fairway: changes here are to correct the ordering when viewing the results

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -315,7 +315,7 @@ export class AggregateReportSummary {
           values: inline(orAll(
             optionFilters.individualEducationPlans,
             settingFilters.individualEducationPlans,
-            code => translate(`common.strict-boolean.${code}`)
+            code => translate(`common.boolean.${code}`)
```
dsnyder-fairway: I don't think this has any impact either way, but I'm going to see what reverting this change does

dsnyder-fairway:
I'll address any comments as they come

PR #1639, opened 2018-07-03
Fix to a broken build.

agorina:
Rolled back some changes.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcTargetReportRepository.java
```diff
@@ -165,6 +165,7 @@ private TargetRow buildEmptyReportRow(final Organization organization,
                 .organization(organization)
                 .dimension(buildDimension(dimension, query))
                 .measures(Measures.getEmptyMeasure())
+                .claimCode(target.getClaimCode())
```
agorina: We need claim code for the target report.

PR #1642, opened 2018-07-03
Use configured access denied URL on the client side to redirect if ne…

jtreuting:
…eded

the config setting for access denied url is by default `forward:/assets/public/access-denied.html` or it can be changed to `redirect:http://external.url`

so this uses any redirect: setting and redirects the user away from the access denied component if appropriate.  it will not magically make other `forward:/internal/page/html` work.  That is beyond the scope.


webapp/src/main/webapp/src/app/error/access-denied/access-denied.component.html
```diff
@@ -4,8 +4,14 @@ <h2>{{'access-denied.title' | translate}}</h2>
   </ng-template>
 </page-heading>
 
+
+
 <div class="well">
-  <p>{{'access-denied.message' | translate}}</p>
-  <a class="btn btn-primary" href="javascript:void(0)" onclick="window.document.logout.submit()">{{'access-denied.logout' | translate}}</a>
-  <form name="logout" method="post" action="/saml/logout" class="hidden"></form>
+  <spinner *ngIf="loading"></spinner>
```
jtreuting: added a spinner since we subscribe to the setting service

webapp/src/main/webapp/src/app/error/access-denied/access-denied.component.ts
```diff
@@ -1,14 +1,25 @@
...
+  loading: boolean = true;
 
-  constructor() { }
+  constructor(private service: ApplicationSettingsService) { }
 
   ngOnInit() {
+    this.loading = true;
+    this.service.getSettings().subscribe(settings => {
+      if (settings.accessDeniedUrl.startsWith('redirect:')) {
+        window.location.href = settings.accessDeniedUrl.substring(9);
```
jtreuting: i couldn't find a better/different way to do this with my quick googling, but it felt like there should be some Angular window service to inject like there was in AngularJS.

wtritch: I'd be tempted to do your sub-stringing in the mapper and then switch here based upon whether or not the value is undefined.
If you really want to use the "Angular" way, you could inject a Document reference and modify that:
https://stackoverflow.com/a/42452665
but I'm ok with window.location.href as well.

jtreuting: good call on the mapper, i'll change that

PR #1645, opened 2018-07-03
Update PDF generation to account for subject assessment type labels.

wtritch:
image

image


jtreuting:
@wtritch I thought we wanted the short text in the dark blue to be SUM, ICA and IAB still?

wtritch:
@jtreuting I thought we were debating that as a way to avoid having multiple labels in the XML?  But now that we support both a long and short name, I thought it was more appropriate to follow the same pattern as the SBAC assessments with `[{{ShortName}} | {{LongName}} Report]` ?

wtritch:
Merging now, will resolve additional comments as they come in

PR #1650, opened 2018-07-04
Remove graph border radius from PDF graphs

wtritch:
With the corner border radius, wkhtmltopdf seems to do some funky antialiasing on the bars at the left and right edges of the graph, which makes them appear about a pixel or two shorter than the bars in the middle.

image

image

After removing the corner border radius the bars are drawn with a uniform height:

image

image


wtritch:
Merging now, will resolve any comments if/when they come.

PR #1656, opened 2018-07-05
return empty string when level counts are null on csv export of aggre…

dsnyder-fairway:
…gate reports


dsnyder-fairway:
merging

PR #1657, opened 2018-07-05
Fix generic labels for aggregate report/ IAB export.

agorina:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/converter/AbstractAggregateReportCsvHandler.java
```diff
@@ -343,16 +343,9 @@ private void printCsv(final Iterator<R> results, final Q query, final OutputStre
...
-            for (int i = 1; i <= 3; i++) {
-                //this should not happen for IABs
-                if (!prefix.isEmpty()) throw new IllegalStateException("invalid invocation for iab");
-                performanceLevelHeaders.add(getMessage("report.iab.level." + prefix + i + ".label"));
-            }
-        } else {
-            for (int i = 1; i <= AssessmentBoundaries.MaxPerformanceLevels; i++) {
-                performanceLevelHeaders.add(getMessage("report.generic.performance-level." + prefix + i + ".label"));
-            }
+        final int maxPerformanceLevels = IAB.code().equals(query.getAssessmentTypeCode()) ? 3 : AssessmentBoundaries.MaxPerformanceLevels;
```
agorina: I did not see a reason to have different labels between IAB and other asmt types for this export.

Any objections?

jtreuting: these headers are `Level1`, `Level2`, etc right?  yeah i don't see any reason either

agorina:
The headers are as you said: 'Level1' and so on.

PR #1659, opened 2018-07-05
handle switching between groups with no common subject on IAB Dashboard

dsnyder-fairway:


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -69,15 +69,17 @@ export class GroupDashboardComponent implements OnInit {
...
 
         const reload = previousParameters == null
           || previousParameters.schoolYear != schoolYear
+          || (previousParameters.subject != null && previousParameters.subject != subject)
           || (previousParameters.groupId != null && previousParameters.groupId != groupId)
           || (previousParameters.userGroupId != null && previousParameters.userGroupId != userGroupId);
 
-        const defaultsParametersRequired = isNaN(Number(schoolYear)) || schoolYear === '';
+        const defaultsParametersRequired = (isNaN(Number(schoolYear)) || schoolYear === '')
+          || (subject != null && (!this.subjects || this.subjects.indexOf(subject) < 0));
```
jtreuting: `this.subjects` is a string array right?  if so the indexOf check could be replaced with `!this.subjects.includes(subject)` which reads cleaner to me.

dsnyder-fairway: sounds good. Update coming

PR #1661, opened 2018-07-05
allow for only a single subject (or "All" option) to be selected on u…

dsnyder-fairway:
…ser groups

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
@@ -8,7 +8,7 @@ import { Option } from './option';
 /**
  * All input types that can be used to change the component click behavior
  */
-export type InputType = 'checkbox' | 'range';
+export type InputType = 'checkbox' | 'radio' | 'range';
 
 /**
  * Default component styles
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -8,7 +8,7 @@ import { Option } from './option';
 /**
  * All input types that can be used to change the component click behavior
  */
-export type InputType = 'checkbox' | 'range';
+export type InputType = 'checkbox' | 'radio' | 'range';
```
dsnyder-fairway: while we do have an `sb-radio-group.ts` there is not an option to select "All" from there

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -112,8 +125,9 @@ class Range implements InputController {
   }
 }
 
-const ControllerByInputType: {[inpuType: string]: InputController} = {
+const ControllerByInputType: { [ inpuType: string ]: InputController } = {
```
jtreuting: nit: there was an existing typo on the param `inpuType` (that's what you get for fixing the format ;) )

dsnyder-fairway:
@jtreuting unfortunately it looks like there is no test coverage for `sb-button-group.ts` yet

PR #1665, opened 2018-07-06
corrected withinBounds validation for longitudinal reports

dsnyder-fairway:
single validation message
year at 2014
screen shot 2018-07-06 at 6 43 04 pm
move year up one
screen shot 2018-07-06 at 6 43 11 pm
add a grade
screen shot 2018-07-06 at 6 43 15 pm

corrected notification too. There was some a `{{lowestAvailableSchoolYear}}` appearing in the notification string before
screen shot 2018-07-06 at 6 43 26 pm


webapp/src/main/webapp/src/app/aggregate-report/query-forms/longitudinal-cohort-form.component.ts
```diff
@@ -160,4 +164,26 @@ export class LongitudinalCohortFormComponent extends MultiOrganizationQueryFormC
...
+   * Form control validator that makes sure the computed effective years does not go below the selected school year
+   *
+   * @param {number} lowestAvailableSchoolYear the lowest school year we allow in the application
+   * @param properties the properties to propagate when the control value is invalid
+   * @returns {ValidatorFn}
+   */
+  private withinBounds(lowestAvailableSchoolYear: number,
+                       properties: any): ValidatorFn {
+    return control => {
+      if ((control.root.get('toSchoolYear') || control.root.get('assessmentGradeRange')) === null) {
```
dsnyder-fairway: wasn't sure how to have these two values being read with how we build the formGroup. So I now grab the root and check its children

webapp/src/main/webapp/src/app/aggregate-report/query-forms/longitudinal-cohort-form.component.ts
```diff
@@ -97,12 +98,15 @@ export class LongitudinalCohortFormComponent extends MultiOrganizationQueryFormC
       assessmentGradeRange: [
         this.settings.longitudinalCohort.assessmentGrades, [
           isGreaterThan(1, { messageId: 'aggregate-report-form.field.assessment-grades-less-than-minimum-error' }),
-          withinBounds(this.settings.longitudinalCohort.toSchoolYear,
```
dsnyder-fairway: Looks like this `this.settings.longitudinalCohort.toSchoolYear` didn't reflect the current selection

webapp/src/main/webapp/src/app/aggregate-report/query-forms/longitudinal-cohort-form.component.ts
dsnyder-fairway: Also the primary reason for moving this out of `validators.ts` into this file as it is referencing children of the root which is constructed here

wtritch: hm... it feels strange to bind a two-value validator to a single input, but I'm not going to advocate refactoring now.  What exactly is the problem being fixed?  And if control A is invalid, does it suddenly go valid by fiddling with control B?  Or do you need to touch control A again to get it to revalidate?

dsnyder-fairway: this fixes an issue where you could generate a longitudinal report from 2011 to 2014 by selecting the most recent academic year of 2014-2015 and 4 grades. After looking at this again it does require a touch to get it to revalidate

dsnyder-fairway: @wtritch recent changes get rid of the need to touch in order to revalidate

dsnyder-fairway:
@jtreuting @wtritch should I merge this? Original issue was that we had 

1. Never accounted for a change in Most Recent Academic Year when validating the form. This resulted in the ability to query for test results going back to 2011 if we selected Most Recent Academic Year of 2014
2. had an notification message "You've selected too many grades. There are no results prior to {{lowestAvailableSchoolYear}}."

jtreuting:
@dsnyder-fairway I'd like to see it in practice.  Let's screenshare today and run through it.

PR #1669, opened 2018-07-06
Order group target table by standard subgroup order.

wtritch:
ordered_targets

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
@@ -14,7 +14,7 @@ import { Subscription } from 'rxjs/Subscription';
 import { forkJoin } from 'rxjs/observable/forkJoin';
 import { Target } from '../../../model/target.model';
 import { Ordering, ordering } from '@kourge/ordering';
-import { byNumber, Comparator, join } from '@kourge/ordering/comparator';
+import { byNumber, Comparator, join, ranking } from '@kourge/ordering/comparator';
 import { TargetService } from '../../../../shared/target/target.service';
 import { BaseColumn } from '../../../../shared/datatable/base-column.model';
 import { Table } from 'primeng/table';
@@ -35,17 +35,6 @@ import { SortEvent } from 'primeng/api';
 import { AggregateReportItem } from '../../../../aggregate-report/results/aggregate-report-item';
 import { Utils } from '../../../../shared/support/support';
 import { SubjectDefinition } from '../../../../subject/subject';
-
-const SubgroupOrdering = ordering((a: Subgroup, b: Subgroup) => {
-  // Overall should be first
-  if (a.name.startsWith('Overall') && !b.name.startsWith('Overall')) {
-    return -1;
-  }
-  if (!a.name.startsWith('Overall') && b.name.startsWith('Overall')) {
-    return 1;
-  }
-  return a.name.localeCompare(b.name);
-});
 
 @Component({
   selector: 'target-report',
@@ -149,10 +138,6 @@ export class TargetReportComponent implements OnInit, ExportResults {
               private dataTableService: DataTableService,
               private filterOptionService: ExamFilterOptionsService,
               private applicationSettingsService: ApplicationSettingsService) {
-
-    applicationSettingsService.getSettings().subscribe(settings => {
-      this.allSubgroups = this.createAllSubgroups(settings);
-    });
   }
 
   ngOnInit(): void {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -36,17 +36,6 @@ import { AggregateReportItem } from '../../../../aggregate-report/results/aggreg
...
-const SubgroupOrdering = ordering((a: Subgroup, b: Subgroup) => {
-  // Overall should be first
-  if (a.name.startsWith('Overall') && !b.name.startsWith('Overall')) {
-    return -1;
-  }
-  if (!a.name.startsWith('Overall') && b.name.startsWith('Overall')) {
-    return 1;
-  }
-  return a.name.localeCompare(b.name);
-});
```
wtritch: This was the old subgroup ordering

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -369,6 +356,41 @@ export class TargetReportComponent implements OnInit, ExportResults {
...
+      const dimensionA = toDimension(a);
+      const dimensionB = toDimension(b);
+      const orderedValues: any[] = dimensionValuesByType.get(dimensionA.type);
+      if (!orderedValues) return 0;
+
+      return orderedValues.indexOf(dimensionA.value) - orderedValues.indexOf(dimensionB.value);
+    };
+
+    return ordering(join(typeComparator, valueComparator));
+  }
```
wtritch: This was a little more in-depth than I was expecting, since we need value ordering as well as type ordering (Example: Gender: * should come before Ethnicity: * but Gender: Female should come before Gender: NonBinary)

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -284,7 +270,8 @@ export class TargetReportComponent implements OnInit, ExportResults {
           this.translate.instant(`subject.${this.assessment.subject}.claim.${row.claim}.target.${row.targetNaturalId}.name`)
         );
       case 'subgroup':
-        return SubgroupOrdering.on<AggregateTargetScoreRow>(row => row.subgroup);
+        const subgroupOrdering: Ordering<Subgroup> = this.getSubgroupOrdering();
+        return subgroupOrdering.on<AggregateTargetScoreRow>(row => row.subgroup);
```
jtreuting: nit: couldn't these 2 lines be combined

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -369,6 +356,41 @@ export class TargetReportComponent implements OnInit, ExportResults {
...
   }
 
+  private getSubgroupOrdering(): Ordering<Subgroup> {
+    const toDimension = (subgroup: Subgroup) => <any>{
+      type: subgroup.dimensionGroups[0].type,
+      value: subgroup.dimensionGroups[0].values[0] ? subgroup.dimensionGroups[0].values[0].code : null
+    };
+    const dimensionTypes: string[] = this.allSubgroups
+      .map(subgroup => subgroup.code);
+    dimensionTypes.unshift("Overall");
```
jtreuting: could also use the fancy typescript syntax `['Overall', ...this.allSubgroups.map(blah blah)]`

wtritch: changed

webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
wtritch: indeed, done

wtritch:
@jtreuting yeah, it looks like the API that gives options doesn't include that the same way that we're doing migrant status.  There's some hackiness in the target-statistics-calculator that's doing something strange:
```
    subgroupCodes.forEach(subgroupCode => {
      let subgroupValues = [];
      switch (subgroupCode) {
        case 'Gender':
          subgroupValues = subgroupOptions.genders;
          break;
        case 'Ethnicity':
          subgroupValues = subgroupOptions.ethnicities;
          break;
        case 'ELAS':
          subgroupValues = [...subgroupOptions.elasCodes, null]; // adding NULL since we have Not Stated data but not a filter option for it
          break;
        case 'LEP':
        case 'Section504':
        case 'IEP':
          subgroupValues = [true, false];
          break;
        case 'MigrantStatus':
          subgroupValues = [true, false, undefined];
          break;
      }
```
But I'm a little scared to mess with it at this point in time.

PR #1674, opened 2018-07-09
renamed LEPStatus/Limited English Proficiency to EnglishLearner/Engli…

dsnyder-fairway:
…sh Learner

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/DimensionType.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/DimensionType.java
@@ -16,7 +16,7 @@ public enum DimensionType {
     Overall(-1, "Overall"),
     Gender(1, "Sex"),
     Ethnicity(2, "Ethnicity"),
-    LEP(3, "LEPStatus"),
+    LEP(3, "EnglishLearner"),
     ELAS(3, "EnglishLanguageAcquisitionStatus"),
     Section504(4, "Section504Status"),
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/DimensionType.java
```diff
@@ -16,7 +16,7 @@
     Overall(-1, "Overall"),
     Gender(1, "Sex"),
     Ethnicity(2, "Ethnicity"),
-    LEP(3, "LEPStatus"),
+    LEP(3, "EnglishLearner"),
```
dsnyder-fairway: @jtreuting any reason why we cannot change this text? Comments above state 
```
* Note: The codes are associated with SBAC data terms defined by Organization Exam Export
* functionality.
```

jtreuting: @malaffoon do you remember if this was to align with CEDS?

malaffoon: Yes, CEDS refers to that as LEP, pretty definitively. Bit surprised they'd want to change that, but i guess if CDE wants it to be something else, they can do whatever they want.

dsnyder-fairway:
Closing for now, I'll open another PR since which will be the reverse of this (changing English Learner to LEP / Limited English Proficiency)

PR #1677, opened 2018-07-11
Move claims to the next "row" when we have more than 4 claims

dsnyder-fairway:


webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -85,18 +85,20 @@
               title="{{'average-scale-score.student-claim-distribution' | translate}}"
               content="{{'common.info.claim' | translate}}"></strong>
       <div class="ui-table-wrapper">
-        <table class="table table-striped table-hover student-claim-distribution">
+        <table *ngFor="let index of claimReferenceRows"
+               class="table table-striped table-hover student-claim-distribution"
+               [ngClass]="{'limit-width': index > 0, 'column-count-1': claimReferences[index].length === 1, 'column-count-2': claimReferences[index].length === 2, 'column-count-3': claimReferences[index].length === 3}">
```
dsnyder-fairway: there is probably a cleaner way to get the `column-count` class. Probably something like `column-count-{{number}}`. I'll look into that

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -193,18 +196,28 @@ export class AverageScaleScoreComponent {
     return results;
   }
 
+  getClaimScoreClass(index): string {
+    return 'table table-striped table-hover student-claim-distribution ' +
```
jtreuting: let's move the static classes out to the HTML and just have the column-count- part of the `[ngClass]` that calls this method.  i think you can have something like `class="table table-striped ...." [ngClass]=getClaimScoreClass(..)"`

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -193,18 +196,28 @@ export class AverageScaleScoreComponent {
...
       .subscribe(ordering => {
-        this.claimReferences = this.assessment.claimCodes.map((code, idx) => <ClaimReference>{
+        const claimReferenceSingleArray = this.assessment.claimCodes.map((code, idx) => <ClaimReference>{
           code: code,
           dataIndex: idx,
-          stats: this.statistics.claims[idx]
+          stats: this.statistics.claims[ idx ]
         }).sort(ordering.on((reference: ClaimReference) => reference.code).compare);
+
+        this.claimReferenceRows = claimReferenceSingleArray.filter((value, index) => index % 4 === 0).map((value, index) => index);
```
jtreuting: can you add a comment describing why this is being done.  it took me a while to understand it

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
```diff
@@ -193,18 +196,28 @@ export class AverageScaleScoreComponent {
     return results;
   }
 
+  getClaimScoreClass(index): string {
+    return 'table table-striped table-hover student-claim-distribution ' +
+      (index > 0 ? `limit-width column-count-${this.claimReferences[ index ].length}` : '');
```
jtreuting: this means it's only limiting the table width if there is more than 4 claims?

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
dsnyder-fairway: creates an array [0,..n] where the n is the length of claimReferenceSingleArray mod 4 = 0. So if we have 6 claimReferences, it is [0, 1] and if we had 16 it would be [0, 1, 2, 3]

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
dsnyder-fairway: yup

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
dsnyder-fairway: done

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.ts
jtreuting: just to be clear, i meant add a comment in the code :) maybe you did but clarifying

dsnyder-fairway: oops haha, alright I'll update this

PR #1680, opened 2018-08-01
Hotfix/1.2.2

malaffoon:
I didn't find a spec.ts for this file so no changes to tests. And i didn't run it locally because ... well, i'm not set up for it. So, my "plan" was to commit, build, deploy to QA/UAT and see what happens.

Obviously, i'm open to suggestions.


webapp/src/main/webapp/src/app/organization-export/organization/organization.service.ts
```diff
@@ -17,12 +17,15 @@ export class OrganizationService extends ExtendOrganizationService {
...
+   * Because the component caches the schools and doesn't go back to the server
+   * during type-ahead search or when adding districts/groups, this needs to
+   * get *every* school for the user. So use a large limit.
    *
    * @returns {Observable<UserOrganizations>}
    */
   getUserOrganizations(): Observable<UserOrganizations> {
     return forkJoin(
-      this.getSchools(),
+      this.getSchools(1000),
```
jtreuting: will we have the same issue with districts and need a bigger number there?

malaffoon:
Oops, this should be against `master` not `develop`

PR #1687, opened 2018-12-18
Upgrading bootstrap to fix vulnerability reported by github

burzum619:
Upgrading bootstrap from `3.3.7` to `4.1.2`, as recommended by github. Because we were importing some stylesheets that have since been removed from the bootstrap package, I had to pull those assets directly into our project. We likely will want to note this in release notes when we are ready to do an official release.


aaronsleeper:
@malaffoon @burzum619 you'd have to check with the other SBAC products that consume the kit. it does make sense to upgrade the kit if the upgrade is needed rather than overwriting the kit at the project level

burzum619:
Are we actually overwriting the kit in this case? The only real code change I'm making is to our `styles.less`, and thats only changing the import path from the `node_modules` folder (which contains the bootstrap library pulled in via `npm`) to the bootstrap directory I added to `/assets`. I don't that we're overwriting anything that wasn't being overwritten already.

aaronsleeper:
honestly i'd have to take a look to understand better. maybe you could walk
me through it on a call?

PR #1688, opened 2018-12-21
RP-31, add targetReport to backend data model and repo

malaffoon:
This adds the targetReport flag to the backend data model. Which isn't really that useful but doesn't hurt anything and nibbles away at the work for this JIRA.


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
```diff
@@ -13,6 +13,7 @@
     private int performanceLevelCount;
     private Integer performanceLevelStandardCutoff;
     private Integer claimScorePerformanceLevelCount;
+    private boolean targetReport;
```
malaffoon: Although this is `Boolean` on the warehouse side, once we're in reporting we know it will be set so we can use the primitive.

PR #1690, opened 2018-12-22
Landing page changes based on Brad's spec

burzum619:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -950,8 +950,7 @@
         "description": "Create and manage student groups for teachers."
       },
       "titles": {
-        "analyze": "Analyze Data",
-        "manage": "Manage"
+        "administrator-tools": "Administrator Tools"
```
malaffoon: Hmm, i think we need to keep track of changes like this for the update documentation. If the user happened to change those titles by customizing their en.json resource, they would want to know about it. Perhaps you could add to the CHANGELOG.md something like:
```
* Change administrator tools in landing page
    * home.admin-tools.titles.{analyze,manage} replaced with home.admin-tools.titles.administrator-tools.
```

PR #1693, opened 2018-12-27
RP-35, add language table to reporting / olap

malaffoon:


common/src/main/resources/common.sql.yml
```diff
@@ -9,17 +9,11 @@ sql:
...
+        SELECT label_code, label FROM accommodation_translation WHERE language_code = :language_code
         UNION ALL
-        SELECT
-          label_code,
-          label
-        FROM subject_translation
-        WHERE 'eng' = :language_code
+        SELECT label_code, label FROM subject_translation WHERE 'eng' = :language_code
+        UNION ALL
+        SELECT CONCAT('common.languages.', code) AS label_code, name AS label FROM language WHERE 'eng' = :language_code
```
malaffoon: @burzum619 , note the translation key, i think this is what you had in your code; it'll be something like `common.languages.spa` -> `Spanish`

malaffoon:
I'm going to merge this, i'll address any concerns as they come up.

PR #1694, opened 2018-12-28
RP-44, add language as an aggregate query dimension

malaffoon:
Most of this was adding `language` to the various model objects. And tweaking tests. And fixing unrelated warnings.

But there were also changes needed because this is a dimension that supports both filtering and subgrouping while not being backfilled.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportTemplate.java
```diff
@@ -29,8 +29,6 @@
     private Map<String, Set<String>> claimCodesBySubject;
     private Map<Integer, Set<Target>> targetsByAssessmentId;
 
-    private boolean includeStudentEnrolledGradeDimension = false;
```
malaffoon: A little explanation: because StudentEnrolledGrade is not backfilled and can't be a filter, it was handled separately from the other dimensions. This bothered me a bit and, with Language sharing half the characteristics, i decided to pull all the dimensions in. So no need for this flag. If StudentEnrolledGrade is specified as a dimension then it is captured as a dimension with no value. And an attribute on DimensionType allows the non-backfilling behavior.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcClaimReportRepository.java
```diff
@@ -159,7 +159,7 @@ public int estimateReportRowCountByOrganizationType(@NotNull final ClaimOrganiza
 
         return (reportTemplate.getOrganizations().size()
                 * claimsByAssessment.values().stream().flatMap(Collection::stream).collect(Collectors.toList()).size()
-                * (reportTemplate.getDimensions().size() + (reportTemplate.isIncludeStudentEnrolledGradeDimension() ? 1 : 0)));
+                * reportTemplate.getDimensions().size());
```
malaffoon: See note above: all the dimensions are captured in the list so no special handling for StudentEnrolledGrade.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/impl/QueryProviderRepositoryHelper.java
```diff
@@ -124,24 +126,32 @@
      * @return all possible values for the given {@link DimensionType} filtered based on the given {@link OrganizationTypeQuery}
      */
     public List<Dimension> findDimensionsByTypeAndQueryFilters(final DimensionType dimensionType, final OrganizationTypeQuery query) {
```
malaffoon: This is the method that changed the most: It knows about DimensionType flags and acts accordingly.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/impl/query/AbstractSqlQueryWithReportTemplateProvider.java
```diff
@@ -240,45 +245,49 @@ private void addFilterParams(final Builder queryBuilder,
...
 
             case ELAS:
                 queryBuilder.parameter("all_elas", dimensionFilters.isEmpty())
-                        .parameter("elas_ids", paddedFiltersToMaxSize);
+                        .parameter("elas_ids", padIntToSize(dimensionFilters, repositoryHelper.maxSize(dimensionType)));
+                break;
+
+            case Language:
+                queryBuilder.parameter("all_languages", dimensionFilters.isEmpty())
+                        .parameter("language_ids", padIntToSize(dimensionFilters, 8));
```
malaffoon: Yeah, magic number. To explain: Redshift optimizes queries based on the size of the query params. For example, if a query has `WHERE id IN (1, 2, 3)`, changing the number of values messes up the optimizer. So, to improve performance we put in non-matchable values to pad the list, e.g. `WHERE id IN (1, -1, -2)`. For most dimensions, the length of that list can simply be the max number of codes. For language that is silly, there are 500. So i arbitrarily chose 8 as a reasonable max number of language codes. If a user chooses more, then all of them will be passed in, the optimizer just won't do as good a job. But up to 8 values, the optimizer will be happy.

PR #1696, opened 2018-12-31
RP-43, add languages to exam filter options

malaffoon:
Because the filtering happens in the UI, this change involves adding the language code to the student/exam payloads, and including the languages in the exam filter options. 

I also renamed a couple methods to make them consistent, and fixed some warnings. Sorry, that adds some noise to this review.

### DIFF
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
@@ -13,6 +13,7 @@ public class ExamFilterOptions {
     private List<String> ethnicities;
     private List<String> genders;
     private List<String> elasCodes;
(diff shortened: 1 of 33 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/ExamFilterOptions.java
```diff
@@ -13,6 +13,7 @@
     private List<String> ethnicities;
     private List<String> genders;
     private List<String> elasCodes;
+    private List<String> languages;
     private List<String> subjects;
```
malaffoon: These fields are inconsistent, but i didn't want to change a payload that is going to the front-end. Personally, i think the field names should indicate these are codes so `ethnicityCodes`, `genderCodes`, etc. But i'm not going to lose any sleep over it.

PR #1698, opened 2019-01-03
Feature/rp 31 target report flag

burzum619:
- Back-end/front-end changes to object models support "complex" subject data model.
- Updating frontend mapping logic
- Honoring target report flag in custom aggregate - target report section and in student group "Results view"
- Unit test fixes


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
```diff
@@ -11,17 +11,26 @@
...
 
-    JdbcSubjectRepository(
-            final NamedParameterJdbcTemplate template,
-            @Value("${sql.subject.findAll}") final String findAllQuery) {
-        super(template, findAllQuery);
+    @Value("${sql.subject.findAll}")
+    private String findAllQuery;
+
+    JdbcSubjectRepository(final NamedParameterJdbcTemplate template) {
+      this.template = template;
```
malaffoon: what are your tab settings?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/impl/QueryProviderRepositoryHelper.java
```diff
@@ -365,11 +365,26 @@ private int toId(final String code, final CodedEntityRepository repository, fina
...
+    }
+
     private Set<Integer> toIds(final Collection<String> codes, final CodedEntityRepository repository) {
         final TreeMap<String, Integer> codeToId = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
         repository.findAll().forEach(entity -> codeToId.put(entity.getCode(), entity.getId()));
         return codes.stream()
                 .map(codeToId::get)
                 .collect(Collectors.toSet());
     }
+
```
burzum619: Not loving having this duplicate code with only the method signature being different but couldn't figure out a  clean way around it.

aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -113,10 +113,15 @@ sql:
...
-      ORDER BY id
+      SELECT
+        s.id,
+        s.code,
+        sat.target_report,
+        at.code as asmt_type_code
+      FROM subject s
+        JOIN subject_asmt_type sat ON s.id = sat.subject_id
+        JOIN asmt_type at ON at.id = sat.asmt_type_id
+      ORDER BY s.id
```
malaffoon: I'll wait to see the rest of the PR but isn't this now returning 3x as many rows as before?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
burzum619: image

This look OK? I though I imported the RDW IDE settings file

aggregate-service/src/main/resources/application.sql.yml
burzum619: Yes, this will return every subject/assessment type permutation. We then filter on the client side by the specific assessment type, and set the appropriate subject code based on that.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Subject.java
```diff
@@ -0,0 +1,51 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * Represents a subject/assessment type combination
+ */
+public class Subject extends CodedEntity {
+    /**
```
malaffoon: To be consistent (and more useful to other devs) i think the javadocs should be on the getters/setters, not the internal fields.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Subject.java
```diff
@@ -0,0 +1,51 @@
...
+     * Indicates whether the subject/assessment type combination has target reporting enabled
+     */
+    private boolean targetReport;
+
+    /**
+     * The assessment type "key"
+     */
+    private String assessmentType;
```
malaffoon: Hmm, i would've expected this to be either AssessmentType or int (type-id). I'll see how it's used ...

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -98,13 +98,16 @@ sql:
     assessment:
       findAll: >-
           SELECT distinct
-            ${sql.reporting.snippet.assessmentColumns}
+            ${sql.reporting.snippet.assessmentColumns},
+            sat.target_report as target_report_enabled
```
malaffoon: Why the alias? Just curious - I find using the native column name is less prone to error in general.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Subject.java
burzum619: Ah, didn't realize there was an existing enum in `common`. I can switch over to using that instead.

reporting-service/src/main/resources/application.sql.yml
burzum619: I generally agree although I though the column name in this particular case is a bit vague. Totally fine removing the alias if you prefer.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
malaffoon: Those look correct, i wonder why it only indented 2?

reporting-service/src/main/resources/application.sql.yml
malaffoon: Perhaps we should change the column name. I'd prefer to not have an alias here.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
burzum619: Good question. Running the auto-formatting in my IDE has corrected the issue though, at least for this particularly file 🤷‍♂️

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -268,6 +274,7 @@ sql:
           JOIN subject su ON a.subject_id = su.id
           JOIN student st ON e.student_id = st.id
           JOIN school isch ON st.inferred_school_id = isch.id
+          JOIN subject_asmt_type sat on su.id = sat.subject_id AND at.id = sat.asmt_type_id
```
malaffoon: Shouldn't the last condition be `... AND a.type_id = sat.asmt_type_id`?

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -196,7 +196,9 @@ export class AssessmentResultsComponent implements OnInit {
   }
 
   get displayTargetReport(): boolean {
-    return this.allowTargetReport && this._assessmentExam.assessment.isSummative;
+    return this.allowTargetReport
+      && this._assessmentExam.assessment.isSummative
+      && this._assessmentExam.assessment.targetReportEnabled;
```
malaffoon: Hmm. This is fine for now but the check against `isSummative` was basically a proxy for `targetReportEnabled` before we had that flag. In theory, you should be able to check just `targetReportEnabled`.

reporting-service/src/main/resources/application.sql.yml
```diff
@@ -115,14 +118,17 @@ sql:
...
+            max(e.completed_at) as max_completed,
+            sat.target_report as target_report_enabled
           FROM exam e
             JOIN asmt a ON e.asmt_id=a.id
             JOIN school s ON e.school_id=s.id
             JOIN subject su ON a.subject_id = su.id
             JOIN student st ON e.student_id = st.id
             JOIN school isch ON st.inferred_school_id = isch.id
+            JOIN asmt_type at ON at.id = a.type_id
+            JOIN subject_asmt_type sat on su.id = sat.subject_id AND at.id = sat.asmt_type_id
```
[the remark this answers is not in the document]
burzum619: Whoops! Nice catch - thanks for pointing that out. The join on `asmt_type` is unnecessary.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/impl/QueryProviderRepositoryHelper.java
malaffoon: Hmm, like Grade you could just inline the helper code in the toSubjectId and toSubjectIds methods since it isn't being used in more than one place.

reporting-service/src/main/resources/application.sql.yml
burzum619: Same as the other query you pointed out - that additional join is not necessary. Fixing.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
burzum619: I considered removing it but was a bit uncertain of what the side effect might be. If I have your blessing, I'll go ahead and remove it from the conditional.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
```diff
@@ -11,17 +11,26 @@
...
-class JdbcSubjectRepository extends AbstractJdbcCodedEntityRepository implements SubjectRepository {
+class JdbcSubjectRepository implements SubjectRepository {
+    private final NamedParameterJdbcTemplate template;
 
-    JdbcSubjectRepository(
-            final NamedParameterJdbcTemplate template,
-            @Value("${sql.subject.findAll}") final String findAllQuery) {
-        super(template, findAllQuery);
+    @Value("${sql.subject.findAll}")
+    private String findAllQuery;
```
malaffoon: I'm not sure i care that much but why the change in wiring this?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
```diff
@@ -11,17 +11,26 @@
...
     @Override
-    public List<CodedEntity> findAll() {
-        return super.findAll();
+    public List<Subject> findAll() {
+        return template.query(findAllQuery, (row, index) -> Subject.builder()
+                .id(row.getInt("id"))
+                .code(row.getString("code"))
+                .targetReport(row.getBoolean("target_report"))
+                .assessmentType(row.getString("asmt_type_code"))
+                .build());
```
malaffoon: I know these tests are a PITA but i'd like to see one that checks this change. If you want, kick it to me with a reminder and i can add it.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
burzum619: This class no longer  implements `AbstractJdbcCodedEntityRepository`, so that required having to manually wire this up a bit more. I want to return a `List<Subject>` - not the more generic `List<CodedEntity>` in the `findAll()` method here. I used `JdbcClaimRepository` as an example for this repository class.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Subject.java
```diff
@@ -0,0 +1,50 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+/**
+ * Represents a subject/assessment type combination
+ */
+public class Subject extends CodedEntity {
```
malaffoon: If it represents a subject/assessment maybe we should call it SubjectAssessment?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/JdbcSubjectRepository.java
malaffoon: I understand why you refactored the class, just not sure why you changed the wiring for the SQL. We use both techniques so i'm fine with it, just curious.

PR #1701, opened 2019-01-03
Debug/skip-minification flag for npm build

burzum619:


webapp/build.gradle
```diff
@@ -65,7 +65,9 @@ task buildFe(type: NpmTask) {
     workingDir = file("src/main/webapp")
     outputs.dir('build/resources/static')
     dependsOn npm_install
-    args = ['run', 'build']
+    args = project.hasProperty('debug')
+            ? ['run', 'build-debug']
+            : ['run', 'build']
     // move build from dist to static resources
```
burzum619: I wanted the default action to remain the same if no flag is provided (minification turned ON). Let me know if you think the default should be to disable minification (in which case I would suggest changing this flag to `prod` or something similar, and renaming the npm script names appropriately (`build` and `build-prod`).

malaffoon: I like the way you've done it.

PR #1703, opened 2019-01-06
RP-47, use school-year associated with accommodation translations

malaffoon:


common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/JdbcTranslationProviderIT.java
```diff
@@ -2,35 +2,33 @@
...
 import java.util.Locale;
+import org.opentestsystem.rdw.reporting.common.test.ITDataSourceConfiguration;
+import org.opentestsystem.rdw.reporting.common.test.RepositoryIT;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 @RunWith(SpringRunner.class)
 @RepositoryIT
 @ContextConfiguration(classes = {ITDataSourceConfiguration.class, JdbcTranslationProvider.class})
-@Sql(scripts = {"/translations.sql"})
```
malaffoon: It seems inconsistent to use a separate script for one of the three tests, so i deleted it and just inlined it in the test.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ReportViewSupport.java
```diff
@@ -43,6 +43,7 @@ public ReportViewSupport(final MessageSource messageSource) {
      */
     public List<String> getSortedAccommodationNames(final Exam exam, final Locale locale) {
         return exam.getAccommodationCodes().stream()
+                .map(value -> value + "." + exam.getSchoolYear())
```
malaffoon: The change: instead of looking up `NEA_Abacus` it will now look for `NEA_Abacus.2018`

PR #1707, opened 2019-01-09
Set min-width on the well holding the admin buttons so they don't squ…

stevemock:
…ish into each other.

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -1706,6 +1706,7 @@ tr.highlight {
 }
 
 .admin-tools-well {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/styles.less
```diff
@@ -1706,6 +1706,7 @@ tr.highlight {
 }
 
 .admin-tools-well {
+  min-width: 590px;
```
malaffoon: Is this how the other wells on that page are protected? Just curious.

stevemock: No, the other wells on that page flow in a way that they don't really need protection.

PR #1711, opened 2019-01-17
Feature/rp 43

stevemock:
Language filter widget added to the advanced filters reports.

### DIFF
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -49,6 +49,7 @@
     "primeng": "^7.0.4",
     "rxjs": "^6.3.3",
     "tslib": "^1.9.0",
+    "webpack-dev-server": "^3.1.11",
     "zone.js": "^0.8.14"
   },
   "devDependencies": {
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -290,9 +290,7 @@ export class AssessmentsComponent implements OnChanges {
     } else if (property.indexOf('ethnicities') > -1) {
       this.removeEthnicity(property.substring(property.indexOf('.') + 1));
     } else if (property.indexOf('genders') > -1) {
-      console.log("this.clientFilterBy['genders'] starts as ", this.clientFilterBy['genders']);
       this.removeGender(property.substring(property.indexOf('.') + 1));
-      console.log("this.clientFilterBy['genders'] ends as ", this.clientFilterBy['genders']);
     } else if (property.indexOf('elasCodes') > -1) {
       this.removeElasCode(property.substring(property.indexOf('.') + 1));
     } else if (property.indexOf('languageCodes') > -1) {
--- a/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
+++ b/webapp/src/main/webapp/src/app/assessments/assessments.component.ts
@@ -276,7 +276,7 @@ export class AssessmentsComponent implements OnChanges {
     this.clientFilterBy.elasCodes = Object.assign({}, this.clientFilterBy.elasCodes);
   }
 
-  removeLanguageFilter(languageCode) {
+  removeLanguageCode(languageCode) {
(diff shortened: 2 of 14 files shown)


webapp/src/main/webapp/package.json
```diff
@@ -49,6 +49,7 @@
     "primeng": "^7.0.4",
     "rxjs": "^6.3.3",
     "tslib": "^1.9.0",
+    "webpack-dev-server": "^3.1.11",
```
malaffoon: thanks

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -272,6 +276,12 @@ export class AssessmentsComponent implements OnChanges {
     this.clientFilterBy.elasCodes = Object.assign({}, this.clientFilterBy.elasCodes);
   }
 
+  removeLanguageFilter(languageCode) {
```
malaffoon: nit: to be consistent shouldn't this be `removeLanguage` or `removeLanguageCode`?

But then again, i'm not entirely grokking the diff between language code and the other filters.

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
```diff
@@ -180,6 +192,10 @@ export class ExamFilterService {
     return exam.student && exam.student.ethnicityCodes.some(ethnicity => filterValue.some(code => code == ethnicity));
   }
 
+  private filterByLanguageCode(exam: Exam, filterValue: any): boolean {
+    return exam.student && !filterValue.length || filterValue.some(languageCode => languageCode == exam.languageCode);
```
malaffoon: nit: perhaps add parens to make it easier to read (yeah, precedence takes care of it but parens help clarify)

webapp/src/main/webapp/src/app/assessments/model/filter-by.model.ts
```diff
@@ -34,6 +35,10 @@ export class FilterBy extends ObservableObject {
     return this.filterArray(this._elasCodes);
   }
 
+  get filteredLanguages(): any[] {
+    return this.filterArray(Object.assign({}, ...this._languageCodes));
```
malaffoon: Feel free to ignore this (cause it's just me not knowing) but ... what exactly is this doing? Why is it different than the other things?

webapp/src/main/webapp/src/app/assessments/model/filter-by.model.ts
```diff
@@ -64,6 +69,11 @@ export class FilterBy extends ObservableObject {
         for (const filteredElasCode of filteredElasCodes) {
           all.push(property + "." + filteredElasCode);
         }
+      } else if (property == "languageCodes") {
+        const filteredLanguages = this.filteredLanguages;
+        for (const filteredLanguage of filteredLanguages) {
```
malaffoon: nit: this is now repeated 4 times, i wonder if a tiny helper method with an intuitive name is worthwhile?

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
stevemock: Sure thing.

webapp/src/main/webapp/src/app/assessments/assessments.component.ts
stevemock: Agreed.

webapp/src/main/webapp/src/app/assessments/model/filter-by.model.ts
stevemock: The _languageCodes are passed in in a slightly different format than say the _elasCodes or the _genders (because of the way that the language widget and it's underlying components work) so this properly formats that object to work with the filters in the same way as the other filter types.

malaffoon: That was my guess ... thanks for 'splaining.

PR #1714, opened 2019-01-22
Bug/rp 55 hamburger menu fixed

stevemock:


webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -5,9 +5,10 @@
   <div class="container">
     <div class="well">
       <div class="well-body">
+
         <nav class="navbar navbar-default">
           <div class="navbar-header">
-            <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navigation-bar-collapse" aria-expanded="false">
+            <button (click)="toggleNavCollapse()" type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navigation-bar-collapse" aria-controls="#navigation-bar-collapse" aria-expanded="false">
```
malaffoon: So what will clicking on it do?

stevemock: It opens and closes the menu. I'll pop you some images.

PR #1715, opened 2019-01-23
Feature/scrollable aggregate summary tables

bgsavage:
Candidate solution for RP-26.

I've implemented a CSS solution to add scrollbars to the tbody for these aggregate report tables.  I would have preferred to use the scrollable property on the primeng datatable, but it no longer is able to align the row cells with the header cells.  This seems to be the best solution for now, but needs to be tested across different browsers / operating systems.  The limitation with this solution is that it means the table cells now need to be a fixed width, which can be less ideal if there are cells that would normally be wider because of the data it contains.

Let me know your thoughts.

image


webapp/src/main/webapp/src/styles.less
```diff
@@ -1768,3 +1768,38 @@ fieldset.marginLeft50 {
...
+    overflow: auto;
+
+    tr {
+      td:first-child {
+        width: 175px;
+        white-space: nowrap;
+        text-overflow: ellipsis;
+        text-align: left;
+        overflow: hidden;
+      }
```
bgsavage: I made the first cell (Organization) wider because it seems to be the most important.  All other cells will be the same width.

bgsavage:
For subgroup, the existing implementation has a text-overflow, but depending on the other data in the table, you might see something like `English Language Ac...` in the table.  Now with the fixed width, no matter what other data is in the table, you will always see something like `English L...`.  I did notice those labels show a tooltip with the full text on hover.

bgsavage:
Fixed as in all column widths are equal, with the exception of the first column which I've set a specific width for.

malaffoon:
Note to dev: the results are paged. If we add scrolling we need to think about removing the paging or something. I'm not sure how the DOM works with a scrolling table and a gazillion entries but the user experience of a scroll bar and paging is not good.

Looking at the screenshot, i'm not thrilled with the page-size selector either. Perhaps we could clean that up as part of this.

PR #1719, opened 2019-01-24
Feature/rp 69 Language added to subgroups filtering in Target Reports

stevemock:


webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
```diff
@@ -192,6 +196,9 @@ export class TargetStatisticsCalculator {
       case 'MigrantStatus': return exam.migrantStatus;
       case 'StudentEnrolledGrade': return exam.enrolledGrade;
 
+      // if languageCode is undefined (for now) that means it was eng / English
+      case 'Language': return exam.languageCode != undefined ? exam.languageCode : 'eng';
```
malaffoon: hmm, can we discuss this; i just want to make sure i understand it

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
```diff
@@ -79,8 +78,11 @@ export class TargetStatisticsCalculator {
         });
       });
     });
-
-    return groupedScores.map(entry => this.mapToAggregateTargetScoreRow(entry));
+    return groupedScores.filter( groupedScore => {  // filter out when the language results were 0 or we'll get a list of every language every time
+      if(! ((groupedScore.standardMetScores.length == 0) && (groupedScore.subgroup.dimensionGroups[0].type === "Language"))) {
+        return groupedScore;
+      }
+    }).map(entry => this.mapToAggregateTargetScoreRow(entry));
```
malaffoon: Discussed offline: for Language, instead of getting all the possible languages from subgroupOptions, we need to extract just the language codes that appear in the targetScoreExams. That would happen further up and then this filter goes away.

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
```diff
@@ -57,20 +57,22 @@ export class TargetStatisticsCalculator {
    * @returns {AggregateTargetScoreRow[]}
    */
   aggregateSubgroupScores(subjectCode: string, allTargets: Target[], targetScoreExams: TargetScoreExam[], subgroupCodes: string[], subgroupOptions: ExamFilterOptions): AggregateTargetScoreRow[] {
+    // Special case for Languages, pull out just the used languageCodes for generatingSubgroupTargets
+    const activeLanguageCodes =  [...Array.from(new Set(targetScoreExams.map(e => e.languageCode))), 'und' ].filter( Boolean );
```
malaffoon: sorry about the noob questions but ... what does the `[].filter( Boolean )` do?

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
```diff
@@ -192,6 +196,9 @@ export class TargetStatisticsCalculator {
       case 'MigrantStatus': return exam.migrantStatus;
       case 'StudentEnrolledGrade': return exam.enrolledGrade;
 
+      // if languageCode is undefined (for now) that means it was eng / English
```
malaffoon: comment is no longer correct

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
stevemock: Np. That's actually been simplified away. What it does however, is remove non truthy elements from an array.

webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
stevemock: Removed.

PR #1720, opened 2019-01-24
RP-71, add LanguageAltCode to export; RP-74, fix IRiS calls

malaffoon:


report-processor/src/main/resources/application.sql.yml
```diff
@@ -245,6 +247,7 @@ sql:
         JOIN school s ON e.school_id = s.id
         JOIN district d ON d.id = s.district_id
         JOIN school isch ON st.inferred_school_id = isch.id
+        LEFT JOIN language lan ON e.language_code = lan.code
```
malaffoon: @agorina if you have a minute to think about this ...

This is the district/school export SQL. The exam table (in reporting) has language_code, a 3-digit varchar. But we also need the alternate code which is in the language lookup table. I played with this query with a few thousand results and the join didn't seem to hurt but what do you think, is joining on a short string okay? The alternative would be to keep language_id in the reporting exam table - i can do that but it would be a bit of frantic coding at the last minute so i'd rather not unless you think this will be a performance issue. Thanks.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -450,21 +450,27 @@ public FilterChainProxy samlFilter() throws Exception {
...
         chains.add(new DefaultSecurityFilterChain(new AntPathRequestMatcher("/saml/discovery/**"), samlIDPDiscovery()));
         final FilterChainProxy filterChainProxy = new FilterChainProxy(chains);
-        filterChainProxy.setFirewall(allowSemicolonHttpFirewall());
+        filterChainProxy.setFirewall(reportingHttpFirewall());
         return filterChainProxy;
     }
 
     /**
-     * The app URLs use semi-colons to separate params so weaken the firewall a bit.
+     * Spring security uses StrictHttpFirewall by default. But it has some restrictions
```
malaffoon: The IRiS problem was due to the strict firewall not allowing un-normalized URLs. Since this is the second problem we've encountered with the firewall, i decided to revert it to the default firewall (which is what was there before we upgraded Spring Security).

PR #1722, opened 2019-01-24
Added LanguageCode to the exportAssessmentExams CSV exporter.

stevemock:
LanguageAltCode still needs to be addressed as described in RP-75


webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -360,7 +360,17 @@ export class CsvBuilder {
...
-    )
+    );
+  }
+
+  withLanguageCode(getExam: (item: any) => Exam) {
+    return this.withColumn(
+      this.translateService.instant('csv-builder.language'),
+      (item) => {
+        const languageCode = getExam(item).languageCode;
+        return languageCode ? this.translateService.instant(`common.languages.${getExam(item).languageCode}`) : '';
```
malaffoon: So this will output the translated language value, i.e. "Chinese; Cantonese"? Not just the code, i.e. "chi"?

stevemock: Oh, good point! Do we want the code or the language? Or both?

stevemock: I've gone with both for the moment...easy enough to remove one or the other.

malaffoon: I think that is a good approach for now.

I was thinking about this last night: for the front-end export, the resulting CSV is supposed to "be like what the teacher sees." In that case, the most appropriate thing to see is the translated language code. And the alt-code wouldn't even enter into it. I guess we can review it with SB/CDE during/after the UAT.

PR #1726, opened 2019-01-25
RP-74, fix calls to IRiS

malaffoon:
This was frustrating because we had a fix for this in-place that just stopped working. I never did figure out why Spring decided the filter bean didn't need to be instantiated. After trying a bunch of things, i built up a new version of the filter that does basically the exact same thing. And it worked.


webapp/src/main/java/org/opentestsystem/rdw/reporting/iris/IrisFormRequestPreFilter.java
```diff
@@ -67,38 +65,52 @@ public IrisAwareFormBodyWrapperFilter() {
...
+            wrapper = new IrisFormBodyRequestWrapper(request);
             context.setRequest(wrapper);
         }
         context.getZuulRequestHeaders().put("content-type", wrapper.getContentType());
+
         return null;
     }
 
-    static class FormBodyRequestWrapper extends Servlet30RequestWrapper {
+    static class IrisFormBodyRequestWrapper extends HttpServletRequestWrapper {
```
malaffoon: Instead of making a copy of Servlet30RequestWrapper and extending it, i just incorporated the changes into this Wrapper; it was just the getRequest() call.

webapp/src/main/resources/application.yml
```diff
@@ -117,7 +117,8 @@ spring:
 zuul:
   set-content-length: true
   FormBodyWrapperFilter:
-    pre: disable
+    pre:
+      disable: true
```
malaffoon: This is what i thought the problem was, they changed the property. But it wasn't (just) this.

PR #1727, opened 2019-01-26
RP-79, modify embargo query to return 0-count records if there are no exams

malaffoon:


admin-service/src/main/resources/application.sql.yml
```diff
@@ -42,14 +42,15 @@ sql:
         st.*,
```
malaffoon: The change forces rows to exist for the org+subject before left joining the counts (before, only the org rows were being forced to exist). Then, for rows where the count doesn't exist, coalesce to 0.

admin-service/src/test/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepositoryIT.java
```diff
@@ -2,9 +2,6 @@
```
malaffoon: Added a new test but there are some warning fixes too.

PR #1729, opened 2019-01-28
Added Language select filter to teacher group creation page.

stevemock:
Fixed bug with All button not deselecting properly in language selection widget.


webapp/src/main/webapp/src/app/shared/filter/student-filters.component.ts
```diff
@@ -3,6 +3,8 @@ import { StudentFilterOptions } from './student-filter-options';
 import { StudentFilter } from './student-filter';
```
malaffoon: I'm confused: is this component not being used in both places? Just wondering why this change is happening now instead of earlier when this was added in the individual report screen.

stevemock: The sb-typeahead-group is being used in several places, yes. This particular instance of it just got missed in the original round of places it needed to go.

malaffoon: discussed offline, apparently adv-filters.component is similar to this component and i was confused

PR #1730, opened 2019-01-28
Fixed bug with Claim section not rendering.

stevemock:


webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -175,18 +175,18 @@ <h3>{{'aggregate-report-form.section.assessment-heading' | translate}}</h3>
 
         <div class="well gray-lighter" *ngIf="!claimsCollapsed">
           <ng-container *ngFor="let subject of settings.subjects">
-            <div class="row mt-md" *ngIf="claimsBySubject[subject].length != 0">
+            <div class="row mt-md" *ngIf="claimsBySubject[subject.code].length != 0">
```
malaffoon: Was this just an oversight from the changes Ernie made?

stevemock: Looks that way to me. It's not something I recall touching.

PR #1735, opened 2019-01-31
Feature/rp 98 Display English + a few randomly selected languages in preview aggregate report summary table

stevemock:


webapp/src/main/webapp/src/app/aggregate-report/subgroup/subgroup.mapper.ts
```diff
@@ -98,6 +98,19 @@ export class SubgroupMapper {
...
+   * adds the first language (should be 'eng' for English) to the front of the returned array
+   */
+  reduceLanguageDimensionCodes(codes: string[], configurationType: string, numLangs: number): string[] {
+    if(configurationType === "Language" && codes.length > numLangs) {
+      let shuffled = codes.slice(1,codes.length -1).sort(() => 0.5 - Math.random());
+      shuffled.unshift(codes[0]); //add first element to the beginning (should be 'eng')
+      return shuffled.slice(0,numLangs);
+    }
+    return codes;
+  }
```
malaffoon: It does seem like there could be a test or two for this method.

stevemock: Right you are.

PR #1737, opened 2019-02-01
Added proper sorting to the teacher CSV export for the subjects via t…

stevemock:
…he subject codes.


webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -71,6 +73,12 @@ export class CsvExportService {
         .withScoreAndErrorBand(getExam);
 
       subjectClaims.forEach(entry => {
+        this.orderingService
+          .getScorableClaimOrdering(entry.subject, entry.type).subscribe(
+          ordering => {
+            entry.claims.sort(ordering.on((claim) => claim).compare);
+          });
```
malaffoon: I assume under the covers somewhere, this is ordering the claims by display-order?

PR #1740, opened 2019-02-01
New JSON entry for the table display specifically and referencing it …

stevemock:
…to shorten English Language Acquisition Status to ELAS.

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -435,19 +435,6 @@
       "Section504": "504",
       "StudentEnrolledGrade": "Student Enrolled Grade"
     },
-    "dimension-table": {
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -435,6 +435,19 @@
...
+      "IEP": "IEP",
+      "LEP": "Limited English Proficiency",
+      "ELAS": "ELAS",
+      "Language": "Primary Language",
+      "MigrantStatus": "Migrant Status",
+      "Overall": "Overall",
+      "Section504": "504",
+      "StudentEnrolledGrade": "Student Enrolled Grade"
+    },
     "dimension-short-name": {
```
malaffoon: I *think* we could just use the "dimension-short-name" values instead of adding a new set.

PR #1742, opened 2019-02-02
try to get prod build working

malaffoon:


webapp/src/main/webapp/src/app/aggregate-report/subgroup/subgroup-filters.component.html
```diff
@@ -17,7 +17,7 @@
                            analyticsEvent="FilterClick"
                            analyticsCategory="AggregateQueryBuilder"
                            [allOptionAnalyticsProperties]="{label: 'Gender: All'}"
-                           (change)="onSettingChangeInternal()"></sb-checkbox-group>
+                           (change)="onSettingChangeInternal($event)"></sb-checkbox-group>
```
malaffoon: I have *no* idea if this is correct. The method requires an event. I hope `$event` is the right thing to do.

stevemock: Yep, if you need the event, then that's how you pass it in.

PR #1747, opened 2019-02-05
Reduced the warning from 10k rows to 4k rows for aggregate reports.

stevemock:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -25,7 +25,7 @@ import { map } from "rxjs/operators";
 import { Observable ,  forkJoin } from "rxjs";
 import { SubjectDefinition } from '../../subject/subject';
 
-export const SupportedRowCount = 10000;
+export const SupportedRowCount = 4000;
 export const DefaultRowsPerPageOptions = [ 50, 100, 500, 1000 ];
```
malaffoon: Since we're not paging any more, can we get rid of this?

malaffoon: It would be great if SupportedRowCount and RowsPerPage were configured instead of hard-coded. How about you ignore the previous comment and i put in a JIRA that asks for both those changes?

stevemock: Yes. Removed all references to rows per page

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -1,6 +1,5 @@
 <p-table #dataTable
          [columns]="columns"
-         [rows]="rowsPerPageOptions[0]"
```
malaffoon: so this isn't necessary at all? how is it deciding how many rows to put in the table?

stevemock: nope. it's determined by the height of the table.

PR #1749, opened 2019-02-07
tweak ISR to have an asterisk beside score range; make summative look like ICA

malaffoon:


common/src/main/resources/i18n/en.json
```diff
@@ -190,7 +190,7 @@
           "header": "Overall Achievement"
```
malaffoon: This was an ad-hoc ask by SB, came up during a requirements meeting

PR #1750, opened 2019-02-11
Tweak ITs, RSTs and a bit of code

malaffoon:
ITs work, RSTs work. There are little problems in the app, but the query appears to work there too. I think you can merge this and move forward.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/sqlbuilder/impl/query/AbstractSqlQueryWithReportTemplateProvider.java
```diff
@@ -188,12 +189,14 @@ Builder addFiltersAndDimension(final Builder queryBuilder, final Q query) {
         final Set<Integer> section504Ids = repositoryHelper.toBooleanIds(query.getSection504Codes());
```
malaffoon: I think the changes in here were the critical ones that were missing.

aggregate-service/src/test/java/org/opentestsystem/rdw/olap/repository/impl/JdbcClaimReportRepositoryRST.java
```diff
@@ -790,13 +812,13 @@ public void itShouldFindByQueryWithMultipleValuesOfDiffParams() {
 
         final List<ClaimRow> rowsWithNonZeroValues = reportRows.stream().filter(r -> (r.getMeasures().getStudentCount() > 0)).collect(Collectors.toList());
 
-        assertThat(rowsWithNonZeroValues.stream().map(AggregateRow::getOrganization).distinct().collect(toSet()))
```
malaffoon: unrelated, no need to `distinct` before collecting in a Set.

aggregate-service/src/main/resources/claim-report-template.sql.yml
```diff
@@ -187,11 +187,15 @@ sqlbuilder:
               clauses:
                 select: >-
                   'MilitaryConnectedCode' AS dimension,
-                  e.military_connected_code AS dimension_id
+                  e.military_connected_id AS dimension_id,
+                  mc.code AS dimension_code
+                join: >-
```
malaffoon: in the OLAP store, the code is not stored in the exam table, so we have to join

aggregate-service/src/main/resources/claim-report-template.sql.yml
```diff
@@ -187,11 +187,15 @@ sqlbuilder:
...
+                  JOIN military_connected mc ON mc.id = e.military_connected_id
                 where: >-
-                  ( true = :all_military_connected_codes OR e.military_connected_code in (:military_connected_code_ids) )
+                  ( true = :all_military_connected OR e.military_connected_id in (:military_connected_ids) )
```
malaffoon: We can quibble on parameter names, i was trying to be consistent with the other dimensions.

PR #1753, opened 2019-02-12
Feature/rp 87

stevemock:
Add filter for military connected to individual reports and filter student results.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
@@ -149,9 +149,7 @@ export class AggregateReportRequestMapper {
 
     // Returns the first argument that is not null or empty
     const or = (a: any, b: any) => Utils.isNullOrEmpty(a) ? b : a;
-
-    console.log("options", options);
-
+    
     // Safely sorts the provided values ranked by the provided options
     const sort = (values: any[], options: any[]) => (values || []).sort(ordering(ranking(options)).compare);
 
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
@@ -150,6 +150,8 @@ export class AggregateReportRequestMapper {
     // Returns the first argument that is not null or empty
     const or = (a: any, b: any) => Utils.isNullOrEmpty(a) ? b : a;
 
+    console.log("options", options);
+
     // Safely sorts the provided values ranked by the provided options
     const sort = (values: any[], options: any[]) => (values || []).sort(ordering(ranking(options)).compare);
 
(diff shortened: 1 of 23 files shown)


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -149,7 +149,7 @@ export class AggregateReportRequestMapper {
 
     // Returns the first argument that is not null or empty
     const or = (a: any, b: any) => Utils.isNullOrEmpty(a) ? b : a;
-
+
```
malaffoon: wasting a perfectly good tab?

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
```diff
@@ -148,6 +153,12 @@ export class ExamFilterService {
       filters = filters.filter(x => x.indexOf('languageCodes') == -1);
       filters.push('languageCodes');
     }
+    if (filters.some(x => x.indexOf('militaryConnectedCodes') > -1 )) {
+      // remove individual 'militaryConnectedCode.code' and add just one militaryConnectedCode
+      // as militaryConnectedCode need to be evaluated all at once.
```
malaffoon: nit (really a nit, don't change if you disagree): it seems like the comments on line 157-158 should be stated once in the function documentation, to avoid repeating it for all the non-binary options.

webapp/src/main/webapp/src/app/assessments/model/filter-by.model.spec.ts
```diff
@@ -74,6 +74,30 @@ describe('FilterBy model', () =>{
     expect(actual).toBeTruthy();
   });
 
+  it('should detect changes to languageCodes', () =>{
```
malaffoon: woot, backfilling tests!

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -383,6 +383,16 @@ export class CsvBuilder {
     );
   }
 
+  withMilitaryConnectedCode(getExam: (item: any) => Exam) {
+    return this.withColumn(
+      this.translateService.instant('csv-builder.military-connected-code'),
+      (item) => {
+        const militaryConnectedCode = getExam(item).militaryConnectedCode;
+        return militaryConnectedCode ? this.translateService.instant(`common.military-connected-code.${getExam(item).militaryConnectedCode}`) : '';
```
malaffoon: Does this need to deal with null? That is, `getExam(item).militaryConnectedCode` can be null can't it?

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
stevemock: makes sense to me.

webapp/src/main/webapp/src/app/assessments/model/filter-by.model.spec.ts
stevemock: +1

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
stevemock: I know, I know!

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
stevemock: No, it just outputs the empty string from the ternary operator if it is null. I've tested the CSV export with the mcc and since the majority of them are blank/null see that empty string in that column.

PR #1759, opened 2019-02-18
Task/npm git hooks

esotrope:
Adds enforcement of code formatting on UI code commit where we previously had none


esotrope:
@malaffoon I understand the argument against the pre-commit hooks however we had a lot of success with this one on the clover project. I found it provided us the following:
1. speedier ramp-up as there is no IDE formatting settings to share
2. enforced consistency alleviates the devs from having to think/worry about their code formatting before checkin
3. ends the conversation around how we should format and format nits in code reviews

PR #1760, opened 2019-02-18
Task/aggregate report form side nav position sticky css

esotrope:
This should improve performance of the sticky nav on the aggregate report page as it relies on the browser built-in css rendering instead of js


webapp/src/main/webapp/src/app/shared/nav/sticky.directive.ts
```diff
@@ -1,91 +0,0 @@
...
-const PositionFixed = 'fixed';
-
-/**
- * Makes a decorated element stick to the top when the user scrolls past its vertical position
- * Also, the elements original page x-position is maintained when scrolling horizontally
- */
-@Directive({
-  selector: '[sticky]'
-})
-export class StickyDirective {
```
esotrope: this is worth it just because we can get rid of this maintenance obligation

webapp/src/main/webapp/src/styles.less
```diff
@@ -793,6 +793,17 @@ aggregate-report-table {
 }
 
 aggregate-report-form, .aggregate-report-form {
+  display: grid;
```
esotrope: removed use of bootstrap row, col classes in favor of css grid

webapp/src/main/webapp/src/app/aggregate-report/query-forms/aggregate-query-form-container.component.html
```diff
@@ -20,28 +20,29 @@ <h2>{{'aggregate-report-form.heading' | translate}}</h2>
...
+          <div class="ml-sm mt-xs">
+            <ng-container *ngTemplateOutlet="submitButtonTemplate"></ng-container>
+          </div>
+        </scroll-nav>
       </div>
     </div>
 
-    <!-- Report Type -->
-    <div class="row" id="reportTypeSection">
-      <div class="col-md-8 col-md-offset-2">
```
esotrope: rm `class="row"` `class="col-*"`

esotrope:
@malaffoon absotutely. doing some more due diligence before i merge as well

esotrope:
sticky

PR #1765, opened 2019-02-25
adds change detection performance enhancements

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -35,7 +35,8 @@ const WideColumnProvider: ColumnProvider = createColumnProvider();
   templateUrl: 'aggregate-report-summary.component.html',
   host: {
     'class': 'aggregate-report-summary'
-  }
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush
```
esotrope: this actually made a noticeable difference in the render time for both the summary and table views. 
We are able to apply this change detection strategy to views that are "readonly" or "input-only" other views that are readonly and "read-once" at init time can be optimized by calling `changeDetector.detach()`

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -442,7 +442,7 @@ <h3>{{'aggregate-report-form.section.subgroup-filters-heading' | translate}}</h3
     <div class="well-group">
       <div class="well">
         <h3>{{'aggregate-report-form.section.review-heading' | translate}}</h3>
-        <ng-container *ngIf="formGroup.valid">
+        <ng-container *ngIf="formGroup.valid && capableOfRowEstimation()">
```
esotrope: this was in the wrong place in a few of the templates. an old copy-paste error probably

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -593,3 +593,10 @@ <h3>{{'aggregate-report-form.section.preview-heading' | translate}}</h3>
           (click)="submitAction.emit($event)">{{'aggregate-report-form.submit' | translate}}
   </button>
 </ng-template>
+
+<ng-template #warning let-text>
```
esotrope: another old copy paste error i think - this template was being referenced in these views but was not present

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.html
```diff
@@ -170,8 +170,7 @@
 
           <div *ngSwitchCase="'studentsTested'"
                class="text-right">
-            <!--{{rowData.studentsTested}}-->
-            {{rowData.itemId}}
+            {{rowData.studentsTested}}
```
esotrope: going to merge this PR since this was a bug introduced by my other PR

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -595,12 +596,13 @@ export class AggregateReportTableComponent implements OnInit {
...
       if (_.isEqual(previousColumns.slice().sort(), newColumns.slice().sort())) {
-        this.columns = sortColumns(
+        const sortedColumns = sortColumns(
           columns,
           newColumns
         );
+        this.columns = sortedColumns;
         this._identityColumnComparators = identityColumnComparators(
-          columns,
+          sortedColumns,
```
esotrope: fixes a column sorting bug

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
malaffoon: I know these are html templates so i probably won't like the answer but ... is there any sort of test we can put in place to catch stuff like this. A missing template seems like a testable thing.

esotrope: you're right it is. i'll see if i cant add something to cover it next go around

PR #1766, opened 2019-02-25
fixes bug with sorting of military connected subgroups on group target reports

esotrope:


webapp/src/main/webapp/src/app/assessments/results/view/target-report/target-report.component.ts
```diff
@@ -374,7 +374,7 @@ export class TargetReportComponent implements OnInit, ExportResults {
     dimensionValuesByType.set("Ethnicity", this.subgroupOptions.ethnicities);
     dimensionValuesByType.set("ELAS", this.subgroupOptions.elasCodes);
     dimensionValuesByType.set("Language", this.subgroupOptions.languages);
-    dimensionValuesByType.set("MilitaryConnected", this.subgroupOptions.militaryConnectedCodes);
+    dimensionValuesByType.set("MilitaryConnectedCode", this.subgroupOptions.militaryConnectedCodes);
```
esotrope: the time to just setup my local to test this - just for one line x l

malaffoon: could we add unit tests for this?

esotrope: We can. it will need a minute of refactoring to make that go nicely though - do we have budget for tech debt in general?

PR #1768, opened 2019-02-26
Defect/rp 123 query subjects not reflected

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.spec.ts
```diff
@@ -234,51 +240,62 @@ describe('AggregateReportRequestMapper', () => {
...
       subjects: [
         {
           code: 'Math',
           assessmentType: 'iab'
-        }, {
+        },
+        {
+          code: 'Math',
+          assessmentType: 'ica'
+        },
```
esotrope: only thing i actually changed

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -21,6 +21,12 @@ const idsOf = values => values.map(value => value.id);
 const hasOption = (options: any[], value) => options.find(option => option === value) != null;
 const notNullOrEmpty = (value) => !Utils.isNullOrEmpty(value);
 
+// Returns the first argument that is not null or empty
+const or = (a: any, b: any) => Utils.isNullOrEmpty(a) ? b : a;
+
+// Safely sorts the provided values ranked by the provided options
+const sort = (values: any[], options: any[]) => (values || []).sort(ordering(ranking(options)).compare);
```
esotrope: moved up top since this will technically be more performant than defining the method in another method

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-request.mapper.ts
```diff
@@ -260,9 +260,20 @@ export class AggregateReportRequestMapper {
           query.subjectCode,
           defaultTargetReport.subjectCode
         )
-      }
+      };
     }
 
+    const subjects = sort(
```
esotrope: the fix - before we were not taking the exact option instances out of the options array which caused the strict equality check in our select component to fail

malaffoon: Hmm, so is the correct fix to take the exact option instances ... OR ... to change the equality check?

esotrope: i went with taking the object instances as this is the existing pattern. I feel like if we are going to venture into refactor land there are bigger things to throw on the chopping block first

esotrope:
going to go ahead and merge this so we can QA it

PR #1769, opened 2019-02-28
removes unused print options

esotrope:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/model/PrintOptions.java
```diff
@@ -1,26 +0,0 @@
...
-
-    private boolean grayscale;
-
-    /**
-     * No-arg constructor for serialization
-     */
-    private PrintOptions() {}
-
-    public PrintOptions(final boolean grayscale) {
-        this.grayscale = grayscale;
```
esotrope: whole point of this was to pass grayscale as a request param when generating reports - i think this was dropped in favor of just using the grayscale print option from your browser/computer print menu though as this option is no longer presented in the UI

esotrope: image

PR #1771, opened 2019-03-05
report-inheritance

malaffoon:
@esotrope, i'd like to hand this off to you. I made the changes to fix the `private`/`protected` issue. Then i started pulling schoolYears into the base class. But i got stuck with the UI code because it is not obvious to me which model objects are for the UI and which are for calling the back-end API. We can discuss at standup but i was hoping to pass this off to you to finish ...

### DIFF
--- a/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java
+++ b/aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java
@@ -93,11 +93,11 @@ class TargetReportQueryHandler extends AbstractAggregateQueryHandler<TargetRow,
         checkArgument(!query.isFilteredSubgroupQuery(), "invalid query, filtered subgroups are not supported");
         checkArgument(query.getSubjectCode() != null, "invalid query, missing a subject");
         checkArgument(query.getAssessmentGradeCodes().size() == 1, "invalid query, must have one assessment grade");
+        checkArgument(query.getSchoolYears().size() == 1, "invalid query, must have exactly one school year");
         checkArgument(!query.isIncludeAllDistrictsOfSchools() && !query.isIncludeAllDistricts()
                 && !query.isIncludeState() && !query.isIncludeAllSchoolsOfDistricts()
                 && ((query.getSchoolIds().size() == 1 && query.getDistrictIds().isEmpty()) ||
                 (query.getDistrictIds().size() == 1 && query.getSchoolIds().isEmpty())), "invalid query, supports one school or one district only");
-        checkArgument(!query.isFilteredSubgroupQuery(), "does not support filtered subgroups");
 
         final String asmtTypeCode = query.getAssessmentTypeCode() != null ? query.getAssessmentTypeCode() : AssessmentType.SUMMATIVE.code();
         checkArgument(AssessmentType.SUMMATIVE.code().equals(asmtTypeCode), "supports summative assessments only");
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
@@ -15,33 +15,33 @@ import static org.opentestsystem.rdw.reporting.common.util.CollectionUtils.safeI
 /**
  * This class provides base functionality for all aggregate report queries.
  */
-@SuppressWarnings("unchecked")
 public abstract class AbstractAggregateReportQuery extends AbstractReportQuery implements AggregateQuery {
 
-    protected Set<String> subjectCodes;
-    protected String assessmentTypeCode;
-    protected ImmutableSet<String> assessmentGradeCodes;
-    protected boolean includeState;
-    protected boolean includeAllDistricts;
-    protected boolean includeAllSchoolsOfDistricts;
-    protected boolean includeAllDistrictsOfSchools;
-    protected ImmutableSet<Long> districtIds;
-    protected ImmutableSet<Long> schoolIds;
-    protected ImmutableSet<String> completenessCodes;
-    protected ImmutableSet<String> administrativeConditionCodes;
+    private Set<String> subjectCodes;
+    private String assessmentTypeCode;
+    private ImmutableSet<String> assessmentGradeCodes;
+    private ImmutableSet<Integer> schoolYears;
+    private boolean includeState;
+    private boolean includeAllDistricts;
+    private boolean includeAllSchoolsOfDistricts;
+    private boolean includeAllDistrictsOfSchools;
+    private ImmutableSet<Long> districtIds;
+    private ImmutableSet<Long> schoolIds;
+    private ImmutableSet<String> completenessCodes;
+    private ImmutableSet<String> administrativeConditionCodes;
 
     // display preferences used by the UI to present the report
-    protected boolean showEmpty = true;
-    protected String valueDisplayType;
-    protected String achievementLevelDisplayType;
-    protected ImmutableList<String> columnOrder;
+    private boolean showEmpty = true;
+    private String valueDisplayType;
+    private String achievementLevelDisplayType;
+    private ImmutableList<String> columnOrder;
 
     // use of these is mutually exclusive with subgroups
-    protected ImmutableSet<DimensionType> dimensionTypes;
-    protected StudentFilters studentFilters;
+    private ImmutableSet<DimensionType> dimensionTypes;
+    private StudentFilters studentFilters;
 
     // defines custom subgroups
-    protected Map<String, StudentFilters> subgroups;
+    private Map<String, StudentFilters> subgroups;
 
     @Override
     public boolean isFilteredSubgroupQuery() {
@@ -58,6 +58,10 @@ public abstract class AbstractAggregateReportQuery extends AbstractReportQuery i
 
     public Set<String> getAssessmentGradeCodes() {
         return safeImmutableOf(assessmentGradeCodes);
+    }
+
+    public Set<Integer> getSchoolYears() {
+        return safeImmutableOf(schoolYears);
     }
 
     public boolean isIncludeState() {
@@ -120,13 +124,14 @@ public abstract class AbstractAggregateReportQuery extends AbstractReportQuery i
         return safeImmutableOf(subgroups);
     }
 
-    public abstract <T extends AbstractAggregateReportQuery, B extends Builder<T, B>> Builder<T, B> copy();
+    public abstract <B extends Builder> B copy();
 
+    @SuppressWarnings("unchecked")
     public static abstract class Builder<A extends AbstractAggregateReportQuery, B extends Builder<A, B>> extends AbstractReportQuery.Builder<A, B> {
-
         private Set<String> subjectCodes;
         private String assessmentTypeCode;
         private Set<String> assessmentGradeCodes;
+        private Set<Integer> schoolYears;
         private boolean includeState;
         private boolean includeAllDistricts;
         private boolean includeAllSchoolsOfDistricts;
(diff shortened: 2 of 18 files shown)


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java
```diff
@@ -93,11 +93,11 @@ protected TargetReportQuery preProcess(final TargetReportQuery query) {
         checkArgument(!query.isFilteredSubgroupQuery(), "invalid query, filtered subgroups are not supported");
         checkArgument(query.getSubjectCode() != null, "invalid query, missing a subject");
         checkArgument(query.getAssessmentGradeCodes().size() == 1, "invalid query, must have one assessment grade");
+        checkArgument(query.getSchoolYears().size() == 1, "invalid query, must have exactly one school year");
         checkArgument(!query.isIncludeAllDistrictsOfSchools() && !query.isIncludeAllDistricts()
                 && !query.isIncludeState() && !query.isIncludeAllSchoolsOfDistricts()
                 && ((query.getSchoolIds().size() == 1 && query.getDistrictIds().isEmpty()) ||
                 (query.getDistrictIds().size() == 1 && query.getSchoolIds().isEmpty())), "invalid query, supports one school or one district only");
-        checkArgument(!query.isFilteredSubgroupQuery(), "does not support filtered subgroups");
```
malaffoon: duplicate check

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -143,52 +148,53 @@ public StudentFilters getStudentFilters() {
...
-            value.administrativeConditionCodes = safeImmutableOf(administrativeConditionCodes);
-            value.showEmpty = showEmpty;
-            value.valueDisplayType = valueDisplayType;
-            value.achievementLevelDisplayType = achievementLevelDisplayType;
-            value.columnOrder = safeImmutableOf(columnOrder);
-            value.dimensionTypes = dimensionTypes == null ? null : ImmutableSet.copyOf(dimensionTypes);
-            value.studentFilters = studentFilters;
-            value.subgroups = subgroups == null ? null : ImmutableMap.copyOf(subgroups);
-            return value;
+        protected AbstractAggregateReportQuery build(final AbstractAggregateReportQuery reportQuery) {
```
malaffoon: The fix is to fully type the parameter.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/jackson/ReportQueryDeserializer.java
```diff
@@ -43,7 +42,7 @@
     }
 
     @Override
-    public AbstractReportQuery deserialize(final JsonParser parser, final DeserializationContext context) throws IOException, JsonProcessingException {
```
malaffoon: fix warning

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/jackson/ReportQueryDeserializer.java
```diff
@@ -73,26 +72,22 @@ private ObjectNode update(final ObjectNode node) {
 
         // this is an aggregate query
         if (node.has("query")) {
-            final ObjectNode query = (ObjectNode) node.remove("query");
-
-            final String aggregateType = query.remove("reportType").asText();
-
-            // Target reports no longer have these unused properties
```
malaffoon: I went back and forth on this. Now that Target extends the base class it seems like we should let the properties through. Yeah, the Target report generation ignores them but throwing them away seems pointless.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/jackson/ReportQueryDeserializer.java
```diff
@@ -73,26 +72,22 @@ private ObjectNode update(final ObjectNode node) {
...
-            if (type.equals(ReportQueryType.Target.name())) {
```
malaffoon: Note that this code wasn't actually getting executed. It was comparing "AggregateReportQuery" to "Target" and never entering the block. Fixed.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/jackson/ReportQueryDeserializer.java
```diff
@@ -73,26 +72,22 @@ private ObjectNode update(final ObjectNode node) {
...
-                query.remove("subjectCodes");
-                query.remove("subgroups");
-                query.remove("valueDisplayType");
+            final ObjectNode queryNode = (ObjectNode) node.remove("query");
+
+            final String aggregateType = queryNode.remove("reportType").asText();
+
+            if (aggregateType.equals(ReportQueryType.Target.name())) {
+                // target-specific schoolYear was replaced with single-value schoolYears from base class
+                if (queryNode.has("schoolYear")) {
```
malaffoon: So the only remaining adjustment for Target is to convert "schoolYear" value to "schoolYears" array of size 1.

common/src/test/java/org/opentestsystem/rdw/reporting/common/model/jackson/ReportQueryDeserializerTest.java
```diff
@@ -431,7 +431,8 @@ public void itShouldDeserializeLegacyTargetReportQuery() {
                                 .build()
                         )
                         .name("Keith's Report")
-                        // old requests send this
+                        // these values are not important for a target report but aren't
+                        // actually excluded so we have to set them for the assertion
```
malaffoon: see comment above

webapp/src/main/webapp/src/app/report/report.ts
```diff
@@ -199,20 +196,19 @@ export interface AggregateReportQuery extends ReportQuery {
   studentFilters?: AggregateReportStudentFilters;
   schoolIds?: number[];
   districtIds?: number[];
+  schoolYears?: number[];
```
malaffoon: Is this the right model to adjust?

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java
esotrope: I think the test looks for this specific failure message so we'll need to change that expectation to the other message maybe

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
esotrope: does this work when scaled though? i think this only works because the next model extending this is a final model and not an intermediary that needs to override the previous builder's `A build(A thing)` method no?

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractBatchPrintableReportQuery.java
```diff
@@ -22,15 +22,15 @@ public B order(final PrintableReportOrder order) {
...
+        protected B copy(final AbstractBatchPrintableReportQuery reportQuery) {
+            return super.copy(reportQuery)
+                    .order(reportQuery.order);
         }
 
-        protected A build(final A value) {
-            super.build(value);
-            value.order = order;
-            return value;
+        protected AbstractBatchPrintableReportQuery build(final AbstractBatchPrintableReportQuery reportQuery) {
```
esotrope: interesting, so java does allow you to override the super's method here despite the explicit type not being the same... shrug

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/DistrictSchoolExportReportQuery.java
```diff
@@ -66,21 +67,22 @@ public Builder districtGroupIds(final Set<Long> districtGroupIds) {
             return this;
         }
 
-        public Builder copy(final DistrictSchoolExportReportQuery value) {
-            return super.copy(value)
-                    .schoolIds(value.schoolIds)
-                    .schoolGroupIds(value.schoolGroupIds)
-                    .districtIds(value.districtIds)
-                    .districtGroupIds(value.districtGroupIds);
+        public Builder copy(final DistrictSchoolExportReportQuery reportQuery) {
```
esotrope: one thing i like about using the var name `value` is that it is context agnostic and thus more portable or copy-paste friendly. I'm still not sure why the java community code style standard is to always parrot the class name
```
is there a good reason i'm oblivious to?

File: common/src/main/java/org/opentestsystem/rdw/reporting/common/model/jackson/ReportQueryDeserializer.java (hunk 5 above)
Comment (esotrope) [reply common/src/main/java/org/opentestsystem/rdw/reporting/common/model/jackson/ReportQueryDeserializer.java]: good catches Mark - thanks!

File: aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java (hunk 1 above)
Comment (malaffoon) [reply aggregate-service/src/main/java/org/opentestsystem/rdw/olap/service/impl/TargetReportQueryHandler.java]: Oh, good point. I did run the tests during dev but might not have after i noticed this.

File: webapp/src/main/webapp/src/app/report/report.ts [hunk 11]
```diff
@@ -199,20 +196,19 @@ export interface AggregateReportQuery extends ReportQuery {
...
 /**
  * A basic aggregate report query
  */
-export interface CustomAggregateReportQuery extends AggregateReportQuery {
-  schoolYears?: number[];
-}
+export interface CustomAggregateReportQuery extends AggregateReportQuery {}
 
 /**
  * A longitudinal aggregate report query
```
esotrope: another goofy thing to work out here is that longitudinals now have schoolYears and toSchoolYear.

webapp/src/main/webapp/src/app/report/report.ts
esotrope: yep

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
malaffoon: It can be made to work, although you do end up with order-sensitive chaining. 

I actually prefer the template pattern where the base class calls something like `newInstance()`; i feel it works out better in the long run. But i didn't want to change it that much.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractBatchPrintableReportQuery.java
malaffoon: Technically we're not overriding, we're hiding. The external call hits the explicit `build` method in the concrete child class, which then explicitly calls its super.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/DistrictSchoolExportReportQuery.java
malaffoon: I'm fine with reverting it back to `value`. I like it when the variable name in the body of the method makes it clear what i have. So `reportQuery.districtIds` works better for me than `value.districtIds`. Of course, if the method is short (as all methods should be) then that is a weak argument.

webapp/src/main/webapp/src/app/report/report.ts
malaffoon: Yeah, i started to change that and then decided it wouldn't be entirely obvious what was going on. At that point i almost reverted all the schoolYear changes. Sigh.

esotrope: Yeah i think i might not want to take on changing school years at this time. I need a break from overhaul-size refactoring for a bit x l

PR #1772, opened 2019-03-05
Feature/rp 143 api user query

esotrope:


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### 1.4.0
 
+* Add saved queries
+* Add toggle for empty aggregate report rows
```
esotrope: adding this entry for the previous work on this feature

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
```diff
@@ -0,0 +1,186 @@
...
+                        .addValue("user_login", query.getUser())
+                        .addValue("label", query.getQuery().getName())
+                        .addValue("query", safeSerialize(query.getQuery()))
+                        .addValue("query_type", query.getQuery().getType().name()),
+                keyHolder
+        );
+        if (insertCount != 1) {
+            throw new IllegalStateException("Unable to create user query");
+        }
+        return findByUserAndId(query.getUser(), keyHolder.getKey().longValue());
```
esotrope: idea here is that the db generates the created/updated datetime column values so this gets them fresh - do we want instead for the repo layer to pass these to avoid the extra db roundtrip?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
```diff
@@ -0,0 +1,186 @@
...
+        final int updateCount = template.update(
+                update,
+                new MapSqlParameterSource()
+                        .addValue("id", query.getId())
+                        .addValue("user_login", query.getUser())
+                        .addValue("label", query.getQuery().getName())
+                        .addValue("query", safeSerialize(query.getQuery()))
+                        .addValue("query_type", query.getQuery().getType().name()));
+
+        if (updateCount != 1) {
```
esotrope: i pushed security into the queries using `where id = :id and user_login = :user_login` then check here for a hit or not - not sure if this is the pattern we had or not though - not many user-saved things in the system.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
```diff
@@ -0,0 +1,186 @@
...
+
+
+    private ObjectMapper createObjectMapper(final ObjectMapper mapper) {
+        final ObjectMapper copy = mapper.copy();
+        copy.configure(FAIL_ON_UNKNOWN_PROPERTIES, false);
+        copy.registerModule(new ReportModule());
+        return copy;
+    }
+
+    private String safeSerialize(final Object query) {
```
esotrope: these could probably both be extracted at least to a common place for user report/query repos

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/UserQueryService.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.security.access.prepost.PreAuthorize;
+
+import javax.validation.constraints.NotNull;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Service responsible for managing user saved queries
+ */
+@PreAuthorize("hasAnyAuthority('PERM_INDIVIDUAL_PII_READ', 'PERM_GROUP_PII_READ', 'PERM_CUSTOM_AGGREGATE_READ')")
```
esotrope: not sure if this is exactly necessary as i'm not sure if you can enter the system without at least one of these right?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Autowired
+    public JdbcUserQueryRepository(final NamedParameterJdbcTemplate template,
+                                   final ObjectMapper objectMapper) {
+        this.template = template;
+        this.objectMapper = createObjectMapper(objectMapper);
+    }
+
+    @Override
+    @Transactional(readOnly = true)
```
esotrope: i noticed in the `JdbcReportRepository` we don't use `@Transactional(readonly = true)` on the reads and the delete method is missing a `@Transactional` is this by design?

report-processor/src/main/resources/application-no-permission.sql.yml
```diff
@@ -66,12 +66,52 @@ sql:
...
+      (user_login, label, query, query_type)
+      values
+      (:user_login, :label, :query, :query_type)
+
+    update: >-
+      update user_query
+      set
+        label = :label,
+        query = :query,
+        query_type = :query_type
```
esotrope: created/updated columns aren't here because this is their config:
```
  created timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
  updated timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
```

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
malaffoon: You don't need transactional if it is a single SQL query, especially for read-only.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
malaffoon: Nope, this is good.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
```diff
@@ -0,0 +1,186 @@
...
+            throw new NoSuchElementException(String.format(
+                    "Unable to update report with id \"%s\" for user \"%s\"",
+                    query.getId(), query.getUser()
+            ));
+        }
+        return findByUserAndId(query.getUser(), query.getId());
+    }
+
+    @Override
+    @Transactional
```
malaffoon: don't need transactional here

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcUserQueryRepository.java
malaffoon: strongly agree; i think these methods should be helpers in a utility someplace so the database layer doesn't have to "know" about how serialization works.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/UserQueryRepository.java
```diff
@@ -0,0 +1,52 @@
...
+     * Finds all queries associated with the user
+     *
+     * @param username The user
+     * @return All queries associated with the user
+     */
+    List<UserQuery> findAllByUser(final String username);
+
+    /**
+     * Finds a query by username and ID
+     *
```
malaffoon: Might note why somebody needs anything other than the id.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/UserQueryRepository.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public interface UserQueryRepository {
+
+    /**
+     * Finds all queries associated with the user
+     *
+     * @param username The user
+     * @return All queries associated with the user
+     */
+    List<UserQuery> findAllByUser(final String username);
```
malaffoon: nit: IJ (and lint) complain about `final` in interface definitions

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultUserQueryService.java
```diff
@@ -0,0 +1,59 @@
...
+        this.repository = repository;
+    }
+
+    @Override
+    public List<UserQuery> get(final User user) {
+        return repository.findAllByUser(user.getUsername());
+    }
+
+    @Override
+    public Optional<UserQuery> get(final User user, long id) {
```
malaffoon: nit: `final long id`

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/UserQueryService.java
malaffoon: There isn't something like `isAuthenticated()`?

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/web/UserQueryControllerIT.java
```diff
@@ -0,0 +1,253 @@
...
+@RunWith(SpringRunner.class)
+@WebMvcTest(UserQueryController.class)
+@Import({
+        DefaultJsonContentConfiguration.class,
+        WebExceptionHandlerConfiguration.class,
+        UserQueryControllerIT.ObjectMapperConfiguration.class
+})
+public class UserQueryControllerIT {
+
+    static class ObjectMapperConfiguration {
```
malaffoon: it is a red flag for me if a test has to configure an object mapper; there isn't any way this can use the object mapper that is part of the default application context?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/UserQueryService.java
esotrope: There is. i'm wondering now if this isn't what we want actually as there could technically be a day where someone enters the system on some other permission that does not get reports.

report-processor/src/test/java/org/opentestsystem/rdw/reporting/processor/web/UserQueryControllerIT.java
esotrope: yeah i think because this is an IT we actually want the app's object mapper here. i'll see what i can do. this was copied form the existing report related IT

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/configuration/JacksonConfiguration.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.reporting.processor.configuration;
+
+import com.fasterxml.jackson.datatype.guava.GuavaModule;
+import org.opentestsystem.rdw.reporting.common.model.jackson.ReportModule;
+import org.springframework.boot.autoconfigure.jackson.Jackson2ObjectMapperBuilderCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JacksonConfiguration {
```
esotrope: moved it to it's own config class so it could be singled out for import by test classes

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/ReportQueryMapper.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.io.IOException;
+
+import static org.apache.commons.lang.StringUtils.isBlank;
+
+/**
+ * Responsible for serializing and deserializing report queries safely
+ */
+@Component
+public class ReportQueryMapper {
```
esotrope: thought of making a more generic "SafeObjectMapper" that returned optionals for `writeAsString` and `readValue` but this ended up not being an exact fit as we would have to run an if statement for presence and fire off a log messages which would require another wrapper class so i just went with this

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/UserQueryRepository.java
esotrope: Added some comments here. let me know if this is what you were looking for or not

PR #1775, opened 2019-03-05
fixes multi threaded test config and adds user query multi threaded test

esotrope:


report-processor/src/test/resources/application-mtTest.yml
```diff
@@ -1,3 +1,6 @@
 spring:
   datasource:
-    url-schema: reporting_test
\ No newline at end of file
+    url-schema: reporting_test
+
+mtTest:
+  expectedTests: 2
```
esotrope: @malaffoon do you recall why we need this? seems like a strange self-imposed constraint

malaffoon: Just a check that tests aren't getting disabled or skipped by accident.

esotrope: ah got it - a test to make sure tests are run ha

PR #1779, opened 2019-03-08
adds ability to delete reports from the UI

esotrope:
Added "Delete Report" option for all reports.
Also added "View Report Query" option for District/School Exports since it is supported now

image


webapp/src/main/webapp/src/app/report/reports.component.html
```diff
@@ -1,57 +1,66 @@
...
-        ></user-query-table>
-      </ng-container>
-      <ng-template #noUserQueries>
-        <p
-          [innerHtml]="
-            'reports.tab.user-queries.no-user-queries-instructions' | translate
-          "
-        ></p>
-      </ng-template>
+<ng-container [ngSwitch]="loadingStatus">
```
esotrope: added loading state and wrapped view with switch statement for loading/loaded/loading failed states

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -20,56 +29,70 @@ import Timer = NodeJS.Timer;
...
-    this.updateReports(reports.data);
-
-    /*
-     Start report status polling
-     Since reports currently cannot be generated on this page we do not have to dynamically update the IDs to pull
-     */
-    if (this.resolution.isOk()) {
-      this.startPollingStatus();
-    }
+    forkJoin(
```
esotrope: now both reports and queries use stores & reactive programming to support read/write elegantly in the views

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -81,10 +104,39 @@ export class ReportsComponent implements OnInit, OnDestroy {
...
+      this.userReportStore.setState(
+        this.userReportStore.state.filter(({ id }) => id !== userReport.id)
+      );
+    });
+  }
+
+  onSaveQuery(userReport: UserReport): void {
+    this.userQueryService.createQuery(userReport.query).subscribe(userQuery => {
+      this.userQueryStore.setState([userQuery, ...this.userQueryStore.state]);
+      this.openQueryTab();
```
esotrope: saving a query now opens the query tab when complete

webapp/src/main/webapp/src/app/report/reports.resolve.ts
```diff
@@ -1,26 +0,0 @@
...
-export class ReportsResolve implements Resolve<Resolution<UserReport[]>> {
-
-  constructor(private service: UserReportService) {
-  }
-
-  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Resolution<UserReport[]>> {
-    return this.service.getReports()
-      .pipe(
-        map(Resolution.ok),
-        catchError(Resolution.errorObservable)
```
esotrope: removed another unnecessary resolve in favor of loading in the view - this allows us to navigate to the view faster and see a loading spinner during the loading process instead of getting more of a hanging response when navigating to the view

malaffoon:
Trying to understand one thing: how/why can you both View Report Query and Save Report Query? If it is already saved, then Save shouldn't be an option (but you can view it). If it isn't yet saved then you can Save it but not View it? Will there be a save button on the View screen? Sorry, just thinking out loud.

webapp/src/main/webapp/src/app/report/reports.component.html
malaffoon: Is this just a general improvement? Is there something that makes loading slower now?

esotrope:
Yeah there will be a save checkbox or button on the form screen.
Currently the user reports don't know that their query has been saved or not so they are not context aware like that - they simply allow you to create a corresponding user query for the query used to build that report

webapp/src/main/webapp/src/app/report/reports.component.ts
malaffoon: That's good, right?  ;)

esotrope: better? yes! good? who knows ; P

webapp/src/main/webapp/src/app/report/reports.component.html
esotrope: general improvement. before we relied on loading the data before the view even started to render and looked at a resolution object to see if it succeeded or not - now the view does an initial render with a spinner and loads the data thus pushing an extra "loading" state to display down to the view.

malaffoon:
So if i keep selecting Save Report Query over and over will it create a bunch of copies?

esotrope:
> So if i keep selecting Save Report Query over and over will it create a bunch of copies?

yes sir : j

PR #1781, opened 2019-03-12
Feature/rp 150 add aggregate save query button and checkbox

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/query-forms/aggregate-query-form-container.component.html
```diff
@@ -1,97 +1,102 @@
-<aggregate-embargo-alert *hasPermission="'EMBARGO_READ'"></aggregate-embargo-alert>
+<aggregate-embargo-alert
```
esotrope: no real changes in this file - just a var rename

webapp/src/main/webapp/src/app/aggregate-report/query-forms/aggregate-query-form-container.component.ts
```diff
@@ -1,48 +1,53 @@
...
-  reportType: string;
-  reportTypes: string[];
+  queryType: string;
+  queryTypes: string[];
   filteredOptions: AggregateReportFormOptions;
   navItems: ScrollNavItem[];
-  formByReportType: {[reportType: string]: BaseAggregateQueryFormComponent} = {};
-
-  // TODO this necessary? doesn't appear to really do anything useful
-  submitActionsByReportType: {[reportType: string]: EventEmitter<Event>} = {};
```
esotrope: removed this extra indirection in favor of just calling the respective form onSubmit handlers directly

webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
```diff
@@ -13,24 +13,28 @@ import { AggregateReportOptions } from '../aggregate-report-options';
...
+import { isEqualReportQuery } from '../../report/reports';
+import { Utils } from '../../shared/support/support';
 
 /**
  * Base query component implementation for all aggregate report types.
  */
-export abstract class BaseAggregateQueryFormComponent implements OnInit, OnDestroy {
-
-  @Input()
-  submitAction: EventEmitter<Event>;
```
esotrope: this was duplicative and not needed now that we call onSubmit directly from the form container

webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
```diff
@@ -130,82 +143,168 @@ export abstract class BaseAggregateQueryFormComponent implements OnInit, OnDestr
   abstract getSupportedAssessmentTypes(): string[];
 
   ngOnInit(): void {
-    this.subjectService.getSubjectDefinitions().subscribe(subjectDefinitions => {
+    const { query, options } = this.route.snapshot.data;
```
esotrope: one complication that fell out of this change was that i needed to make the initialization of each form async - as you'll see below there is a new abstract method `initialize` that the extending forms can use to hook in their specific initializations

webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
```diff
@@ -130,82 +143,168 @@ export abstract class BaseAggregateQueryFormComponent implements OnInit, OnDestr
...
+    if (this.saveQuerySubscription != null) {
+      this.saveQuerySubscription.unsubscribe();
+    }
+    if (this.submissionSubscription != null) {
+      this.submissionSubscription.unsubscribe();
+    }
+  }
+
+  get saveQueryButtonDisabled(): boolean {
+    return (
```
esotrope: luckily i could add most of the changes to the base form component from which each report type form extends

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -1,602 +1,922 @@
...
-          <div class="form-group">
-            <input type="hidden"
-                   [(ngModel)]="organizations"
-                   formControlName="organizations">
-            <div *ngIf="showErrors('organizations')">
-              <p *ngIf="getControl('organizations').errors.invalid"
-                 aria-live="polite"
-                 class="help-block small red">
-                {{getControl('organizations').errors.invalid.messageId | translate}}</p>
+<ng-container *ngIf="formGroup">
```
esotrope: had to wrap each form template with this since this they are no longer initialized in the constructor but after some async event

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -1,602 +1,922 @@
...
+            "
+          ></aggregate-report-table>
+        </div>
       </div>
     </div>
   </div>
-</div>
+</ng-container>
+
+<ng-template #saveQueryTemplate>
```
esotrope: the real annoying part about how we have these forms split up is that they don't share the same html template so this had to be copy-pasted into each of them even though it is identical between them...

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
```diff
@@ -48,52 +48,86 @@ export class ClaimReportFormComponent extends MultiOrganizationQueryFormComponen
...
+      requestMapper,
+      route,
+      router,
+      subgroupMapper,
+      tableDataService
+    );
+  }
 
-    this.assessmentDefinition = this.assessmentDefinitionService.get(this.settings.assessmentType, this.settings.reportType);
+  initialize(): void {
```
esotrope: only real change in these aggregate form components was moving the content of the constructor into these initialize() methods

webapp/src/main/webapp/src/app/aggregate-report/report-query.resolve.ts
```diff
@@ -0,0 +1,34 @@
...
+import { UserQueryService } from '../report/user-query.service';
+import { getQueryFromRouteQueryParameters } from '../report/reports';
+import { UserReportService } from '../report/user-report.service';
+
+/**
+ * This resolver is responsible for fetching an aggregate report based upon
+ * an optional report id query parameter.
+ */
+@Injectable()
+export class ReportQueryResolve implements Resolve<ReportQuery> {
```
esotrope: before, there was a settings resolver which took the query and turned it into settings and then we lost reference to the query once we got to the forms - now the components are responsible for that transform and the resolve just passes the query so we can store it as the "initial state" and use it for comparison with any changes made to the form

webapp/src/main/webapp/src/app/aggregate-report/support.ts
```diff
@@ -1,8 +1,14 @@
-import { District, Organization, OrganizationType, School } from '../shared/organization/organization';
```
esotrope: formatting only, added something to this file then removed it..

webapp/src/main/webapp/src/app/organization-export/organization-export.component.html
```diff
@@ -133,36 +133,32 @@ <h2>{{ 'organization-export.title' | translate }}</h2>
...
-              <div class="form-group checkbox-form-group">
+              <div
+                *ngIf="!showSaveQueryButton"
+                class="form-group inline-checkbox-form-group"
+              >
                 <input
                   id="save-query-checkbox"
                   type="checkbox"
-                  class="btn btn-info aqua"
+                  [ngClass]="{ 'text-muted': saveQueryCheckboxDisabled }"
```
esotrope: revisited org export form to give the checkbox a disabled style when it is disabled

webapp/src/main/webapp/src/app/organization-export/organization-export.component.ts
```diff
@@ -74,7 +66,7 @@ export class OrganizationExportComponent implements OnInit {
   transferAccess: boolean;
   showSaveQueryButton: boolean;
   userOrganizations: UserOrganizations;
-  initialOrganizationExport: OrganizationExport;
+  initialQuery: DistrictSchoolExportReportQuery;
```
esotrope: now the forms compare queries to queries instead of some intermediary model to another intermediary model. this allows for the comparison logic to be the same across report forms

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -104,6 +106,17 @@ export class ReportsComponent implements OnInit, OnDestroy {
     window.location.reload();
   }
 
+  onCopyUserQuery(userQuery: UserQuery): void {
```
esotrope: unrelated - added a copy/duplicate query method so we could see if we liked it or not

webapp/src/main/webapp/src/app/report/reports.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+          }
+        }),
+        expected: false
+      }
+    ]
+  }
+];
+
+// TODO add set comparison assertions for subgroups and studentFilters
+describe('isEqualReportQuery', () => {
```
esotrope: tests for report query equality checking

webapp/src/main/webapp/src/app/report/reports.ts
```diff
@@ -108,3 +111,115 @@ export function getQueryFromRouteQueryParameters<T extends ReportQuery>(
...
+  };
+}
+
+/**
+ * Checks recursively for report query equality
+ *
+ * @param a The first query to compare
+ * @param b The second query to compare
+ */
+export function isEqualReportQuery(a: ReportQuery, b: ReportQuery): boolean {
```
esotrope: the report query equality checking method.
the trick this uses is it preps or "normalizes" queries by creating a copy of them and sorting all their arrays that are supposed to be treated as sets - this is so that the lodash _.isEqual method will correctly perceive them as equal - this is the only strategy i could come up with that had the smallest code footprint, everything else required custom recursive equality checking methods which would require a bunch more tests.

esotrope: another note here is that one _really_ annoying thing about javascript `Set` is that they cannot be serialized to and from json otherwise we would be using these as they are the correct in-memory representation of the object

webapp/src/main/webapp/src/app/report/reports.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    });
+
+    Object.entries(query).forEach(([key, value]) => {
+      const isPrimitiveValue = typeof value !== 'object';
+
+      const differentValue = isPrimitiveValue
+        ? getDifferentValueOfSamePrimitiveType(value)
+        : null;
+
+      it(`${
```
esotrope: wasn't sure of a more thorough way to test this, but basically the strategy here is to generate a test for every field in the query object, it will mutate the field on the right hand side, then compare and expect failure. This gets more complicated as the datastructure nests however so there are some custom tests you can declare in the test cases array above. Ideally, this would recursively iterate and mutate fields but i did not want to go to far down that path as then the test file would need tests itself.

webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
```diff
@@ -130,82 +143,168 @@ export abstract class BaseAggregateQueryFormComponent implements OnInit, OnDestr
...
-          .filter(subject => subject.value.targetReport && subject.value.assessmentType === this.settings.assessmentType)
-          .map(subject => subject.value.code)
-      : this.filteredOptions.subjects
-          .filter(subject => subject.value.assessmentType === this.settings.assessmentType)
-          .map(subject => subject.value.code);
+    const { settings } = this;
+    const validSubjectDefinitions = this.subjectDefinitions.filter(
+      x => x.assessmentType == settings.assessmentType
+    );
+    const enabledSubjects =
```
malaffoon: seems like you could avoid a teeny bit of duplication by moving the conditional inside the filter?

webapp/src/main/webapp/src/app/aggregate-report/query-forms/general-population-form.component.ts
```diff
@@ -98,30 +132,40 @@ export class GeneralPopulationFormComponent extends MultiOrganizationQueryFormCo
...
+        translationKey: 'aggregate-report-form.section.organization.heading'
+      },
+      {
+        id: 'generalPopulationAssessmentSection',
+        translationKey: 'aggregate-report-form.section.assessment-heading'
+      },
+      {
+        id: 'generalPopulationSubgroupSection',
+        translationKey:
+          'aggregate-report-form.section.comparative-subgroups-heading'
```
malaffoon: Seems inconsistent, i guess it must've cross over the line-length limit. This is why hard-and-fast formatting rules drive me nuts.

webapp/src/main/webapp/src/app/report/reports.ts
malaffoon: I gather there isn't an equivalent to assertj's usingRecursiveFieldByFieldElementComparator family of helper methods?

webapp/src/main/webapp/src/app/aggregate-report/query-forms/general-population-form.component.ts
esotrope: yeah that is a side effect that would normally drive me nuts - i think i may have learned not to care as much and it is healthier for me x D

webapp/src/main/webapp/src/app/report/reports.ts
esotrope: you're looking at it - lodash `isEqual` is field by field recursive and supports `Set`s! it's just that we store things as Arrays so they can be easily serialized to/from json from the server. its kind of a battle of choosing whether we want the set conversion logic in the equality checker which we seldom need or have it in the api mapping layer which we use all the time. this way is at least "code sparse"

webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
esotrope: you could, i think i'll refrain from changing the logic at this point though because i don't want to test more changes for this PR ; p

PR #1783, opened 2019-03-12
fixes org export school and district options. renames createUuid to c…

esotrope:
…reateCompositeId

### DIFF
--- a/webapp/src/main/webapp/src/app/organization-export/organization-export.component.ts
+++ b/webapp/src/main/webapp/src/app/organization-export/organization-export.component.ts
@@ -2,7 +2,7 @@ import { Component, OnInit } from '@angular/core';
 import { OrganizationType } from './organization/organization-type.enum';
 import { TranslateService } from '@ngx-translate/core';
 import { Tree } from './organization/tree';
-import { Organization } from './organization/organization';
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/organization-export/organization-export.component.ts
```diff
@@ -193,19 +193,13 @@ export class OrganizationExportComponent implements OnInit {
     });
   }
 
-  set selectedSchools(value: Organization[]) {
-    const {
-      formGroup: {
-        value: { schools }
-      },
```
esotrope: before the method was referencing stale data - it was doing a read, then a write and then not re-reading the data you could say

webapp/src/main/webapp/src/app/organization-export/organization-exports.ts
```diff
@@ -9,7 +9,7 @@ import { Option } from '../shared/form/sb-typeahead.component';
  * @param {number} id the organization entity ID
  * @returns {string} the UUID
  */
-function createUuid(type: OrganizationType, id: number): string {
+function createCompositeId(type: OrganizationType, id: number): string {
```
esotrope: i remembered something

PR #1785, opened 2019-03-13
remove propdeps

malaffoon:
Similar to change in common: Dave had trouble with the repo for propdeps not being available. Since we don't really use the optional and provided features, it seems prudent to stop trying to get it. Note that we do pull in the spring boot spring-boot-configuration-processor using optional, so i switched that to compileOnly which is morally equivalent. I also pulled that common dependency into the root build file since every module was repeating it. And we did use `provided` in common-web but i can't quite figure out why since i *think* the reason it was provided is because the service using common-web is pulling it in, and that means `compile` is the same as `provided` in that situation.  Maybe @wtritch will correct me if i'm wrong.

### DIFF
--- a/admin-service/build.gradle
+++ b/admin-service/build.gradle
@@ -10,7 +10,6 @@ dependencies {
     compile 'org.opentestsystem.rdw.common:rdw-common-security'
     compile 'org.opentestsystem.rdw.common:rdw-common-status'
     compile 'org.opentestsystem.rdw.common:rdw-common-utils'
-    optional 'org.springframework.boot:spring-boot-configuration-processor'
     runtime 'org.springframework.boot:spring-boot-devtools'
     compile 'org.springframework.boot:spring-boot-starter-aop'
     compile 'org.springframework.boot:spring-boot-starter-jdbc'
--- a/common-web/build.gradle
+++ b/common-web/build.gradle
@@ -7,11 +7,11 @@ bootRun.enabled = false
 
 dependencies {
     compile project(':rdw-reporting-common')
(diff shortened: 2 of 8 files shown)


common-web/build.gradle
```diff
@@ -7,11 +7,11 @@ bootRun.enabled = false
 
 dependencies {
     compile project(':rdw-reporting-common')
+    compile 'io.jsonwebtoken:jjwt'
+    compile 'org.springframework:spring-messaging'
     compile 'org.springframework.boot:spring-boot-starter-actuator'
     compile 'org.springframework.boot:spring-boot-starter-web'
     compile 'org.springframework.boot:spring-boot-starter-security'
-    compile 'io.jsonwebtoken:jjwt'
-    provided 'org.springframework:spring-messaging'
```
malaffoon: @wtritch this.

admin-service/build.gradle
```diff
@@ -10,7 +10,6 @@ dependencies {
     compile 'org.opentestsystem.rdw.common:rdw-common-security'
     compile 'org.opentestsystem.rdw.common:rdw-common-status'
     compile 'org.opentestsystem.rdw.common:rdw-common-utils'
-    optional 'org.springframework.boot:spring-boot-configuration-processor'
```
esotrope: what was this for? runtime code reloading?

malaffoon: No, this is for IJ properly handling @ConfigurationProperties. I didn't remove this, just moved it to the root build.gradle file since every module uses it.

PR #1792, opened 2019-03-19
adds new exam search filter service with dynamic filters based on con…

esotrope:
…fig and permissions

example config:
```yaml
reporting:
  studentFilters:
    ethnicity: admin
    gender: disabled
```


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SearchFilter.java
```diff
@@ -0,0 +1,10 @@
+package org.opentestsystem.rdw.reporting.common.model;
+
+import java.util.List;
+
+public interface SearchFilter<ID, V> {
```
esotrope: idea here is we could eventually use this for any filter like school years etc.

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultUserExamSearchFilterService.java
```diff
@@ -0,0 +1,114 @@
...
+            @NotNull final SchoolYearService schoolYearService,
+            @NotNull final EthnicityService ethnicityService,
+            @NotNull final GenderService genderService,
+            @NotNull final ElasService elasService,
+            @NotNull final LanguageService languageService,
+            @NotNull final MilitaryConnectedService militaryConnectedService,
+            @NotNull final SubjectService subjectService) {
+
+
+        final Supplier<List<String>> getStrictBooleans = () -> ImmutableList.of("yes", "no");
```
esotrope: might be a better place for this but since it is only used here for now i just kept it local

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/filter/DefaultUserExamSearchFilterServiceTest.java
```diff
@@ -0,0 +1,263 @@
...
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyElementsOf(studentFilters);
+
+        assertThat(service.get(schoolAdmin()).getStudentFilters())
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyElementsOf(studentFilters);
+
+        assertThat(service.get(embargoAdmin()).getStudentFilters())
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyElementsOf(studentFilters);
```
esotrope: we do have the opportunity to remove all filters for embargo read users if we want - right now they are just treated as "not admin" like teachers are

reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultUserExamSearchFilterService.java
```diff
@@ -0,0 +1,117 @@
...
+        return ExamSearchFilters.builder()
+                .schoolYears(schoolYearService.getSchoolYears())
+                .studentFilters(createStudentFilters(user))
+                .subjects(subjectService.getSubjectCodes())
+                .build();
+    }
+
+    private List<SearchFilter<StudentFilterType, String>> createStudentFilters(final User user) {
+
+        final ImmutableList.Builder<SearchFilter<StudentFilterType, String>> filters = ImmutableList.builder();
```
malaffoon: Perhaps define `StudentSearchFilter extends SearchFilter<StudentFilterType, String>` so you don't have to have the awkward templating everywhere.

PR #1794, opened 2019-03-20
add considerations for english learners settings

esotrope:
now the filters respect the `englishLearners` settings

### DIFF
--- a/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultUserExamSearchFilterService.java
+++ b/reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultUserExamSearchFilterService.java
@@ -20,14 +20,10 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import javax.validation.constraints.NotNull;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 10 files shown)


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/filter/DefaultUserExamSearchFilterService.java
```diff
@@ -91,9 +93,21 @@ public ExamSearchFilters get(final User user) {
...
 
+        // Filter out english learner search filters that are not enabled
+        final List<StudentFilterType> types = Arrays.stream(StudentFilterType.values())
+                .filter(type -> (
+                        type != EnglishLanguageAcquisitionStatus
+                        || settings.getEnglishLearners().contains(ELAS)
+                ) && (
+                        type != LimitedEnglishProficiencies
+                        || settings.getEnglishLearners().contains(LEP)
+                ))
```
malaffoon: nit: would two separate filter calls be more readable?

esotrope: actually i was thinking last night - did we want to just get rid of `reporting: englishLearning:`? since the same thing can be represented in the new config approach as
```
reporting:
  studentFilters:
    lep: disabled
```

esotrope:
That sounds like good measure - i can add that

PR #1796, opened 2019-03-20
remove unused methods

malaffoon:
When researching the use of cut-points i noticed these methods aren't used. Landon, can you sanity check me? Do you think this is just leftover cruft or future-speculation?

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/scale-score.service.ts
+++ b/webapp/src/main/webapp/src/app/assessments/results/scale-score.service.ts
@@ -1,6 +1,5 @@
-import { Injectable } from "@angular/core";
-import { Assessment } from "../model/assessment.model";
-import {ExamStatisticsLevel} from "../model/exam-statistics.model";
+import { Injectable } from '@angular/core';
+import { ExamStatisticsLevel } from '../model/exam-statistics.model';
 
 @Injectable()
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/assessments/results/scale-score.service.ts
```diff
@@ -54,66 +63,10 @@ export class ScaleScoreService {
...
-
-
-  /**
-   * Determines what level a scale score is in based on the cut points
-   *
-   * @param score the scale score
-   * @param standardError the standard error
-   * @returns {number} the level number (starting with 1)
-   */
-  public calculateLevelNumber(assessment: Assessment, score: number, standardError: number): number {
```
malaffoon: This is the method i removed, didn't see it being used anywhere.

esotrope: you made sure to check both in `*.html` and `*.ts`? sometimes there are method references in the html for things

malaffoon: Yup, grep'd the whole project. I guess we used to display the achievement bar and might've needed it then. Okay, i'm going to assume it's okay to remove.

PR #1798, opened 2019-03-21
add test for loading reporting settings

malaffoon:
Mostly demonstrating the way to test loading from yaml files.


common/src/test/resources/application-test-reporting-settings.yml
```diff
@@ -0,0 +1,16 @@
...
+    EconomicDisadvantage: disabled
+    EnglishLanguageAcquisitionStatus: admin
+    Ethnicity: enabled
+    Gender: enabled
+    IndividualEducationPlan: admin
+    LimitedEnglishProficiencies: disabled
+    MigrantStatus: admin
+    MilitaryStudentIdentifier: admin
+    PrimaryLanguage: admin
+    Section504: admin
```
malaffoon: Landon, is there anything else you want me to test as long as i have this going?

common/src/test/resources/application-test-reporting-settings.yml
```diff
@@ -0,0 +1,16 @@
+reporting:
+  state:
+    code: CA
+  iris-vendor-id: bob
+  studentFilters:
```
malaffoon: One thought (and i apologize in advance): these settings are going to be used to suppress data in exports as well as suppress the filters. So, is `studentFilters` a bit too specific? Should it be `studentFields` or some such?

common/src/test/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettingsTest.java
```diff
@@ -0,0 +1,52 @@
...
+        assertThat(studentFilters.get(StudentFilterType.EnglishLanguageAcquisitionStatus)).isEqualTo(Admin);
+        assertThat(studentFilters.get(StudentFilterType.Ethnicity)).isEqualTo(Enabled);
+        assertThat(studentFilters.get(StudentFilterType.Gender)).isEqualTo(Enabled);
+        assertThat(studentFilters.get(StudentFilterType.IndividualEducationPlan)).isEqualTo(Admin);
+        assertThat(studentFilters.get(StudentFilterType.LimitedEnglishProficiencies)).isEqualTo(Disabled);
+        assertThat(studentFilters.get(StudentFilterType.MigrantStatus)).isEqualTo(Admin);
+        assertThat(studentFilters.get(StudentFilterType.MilitaryStudentIdentifier)).isEqualTo(Admin);
+        assertThat(studentFilters.get(StudentFilterType.PrimaryLanguage)).isEqualTo(Admin);
+        assertThat(studentFilters.get(StudentFilterType.Section504)).isEqualTo(Admin);
+        assertThat(studentFilters).hasSize(10);
```
esotrope: why not use the recursive field by field compare as it covers size etc.

common/src/test/resources/application-test-reporting-settings.yml
esotrope: Interesting point. i think that makes more sense to call it `studentFields` if that is the case

common/src/test/resources/application-test-reporting-settings.yml
esotrope: you should also test omission of an entry maybe. other than that i think we're good

common/src/test/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemSettingsTest.java
malaffoon: Mostly cause there isn't an easy constructor for a map with more than 4 entries.

common/src/test/resources/application-test-reporting-settings.yml
malaffoon: Okay, will rename and update PR.

common/src/test/resources/application-test-reporting-settings.yml
malaffoon: Will do.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SearchFieldPermissionLevel.java
```diff
@@ -0,0 +1,20 @@
+package org.opentestsystem.rdw.reporting.common.model;
```
malaffoon: just a rename, i didn't add/delete the files

PR #1799, opened 2019-03-21
fixes report query creation

esotrope:


webapp/src/main/webapp/src/app/app-settings.service.ts
```diff
@@ -1,43 +1,53 @@
 import { ApplicationSettings } from './app-settings';
-import { Observable ,  of } from 'rxjs';
+import { Observable, of } from 'rxjs';
 import { Injectable } from '@angular/core';
 import { CachingDataService } from './shared/data/caching-data.service';
 import { catchError, map } from 'rxjs/operators';
+import { isNullOrEmpty } from './shared/support/support';
 
 const EmptySettings = of(<any>{});
+const DefaultReportLanguages = ['en'];
```
esotrope: added explicit default for language - the backend was not filling this in when passed null

webapp/src/main/webapp/src/app/app-settings.service.ts
```diff
@@ -1,43 +1,53 @@
...
+            )
+              ? serverSettings.accessDeniedUrl.substring(9)
+              : undefined,
+            irisVendorId: serverSettings.irisVendorId,
+            lepEnabled: serverSettings.limitedEnglishProficienciesEnabled,
+            minItemDataYear: serverSettings.minItemDataYear,
+            percentileDisplayEnabled: serverSettings.percentileDisplayEnabled,
+            reportLanguages: isNullOrEmpty(serverSettings.reportLanguages)
+              ? DefaultReportLanguages
+              : serverSettings.reportLanguages,
```
esotrope: the empty option guard

webapp/src/main/webapp/src/app/report/provider/language-field.provider.ts
```diff
@@ -24,7 +24,7 @@ export const useFactory = (
         )
       })),
       defaultValue: reportLanguages[0],
-      excluded: reportLanguages == null || reportLanguages.length < 2
+      hidden: reportLanguages == null || reportLanguages.length < 2
```
esotrope: backend requires language we just dont show it in the UI when there is only one to select so i introduced a "hidden" abstraction for this case

webapp/src/main/webapp/src/app/report/service/report-form.service.ts
```diff
@@ -87,7 +90,11 @@ export class ReportFormService {
         this.translate,
         group
       ),
-      schoolYear
+      schoolYear,
+      groupId: {
+        id: group.id,
+        type: group.userCreated ? 'Teacher' : 'Admin'
+      }
```
esotrope: the missing pieces in query creation

PR #1801, opened 2019-03-21
fixes loading aggregate report query by report id

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -194,7 +213,10 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
   }
 
   onUpdateRequestButtonClick(): void {
-    this.router.navigate([ '..' ], { relativeTo: this.route, queryParams: { src: this.report.id } });
+    this.router.navigate(['..'], {
+      relativeTo: this.route,
+      queryParams: { userReportId: this.report.id }
```
esotrope: the only change and the fix to the defect was changing `src` to the new param name `userReportId`

webapp/src/main/webapp/src/app/report/provider/order-field.provider.ts
```diff
@@ -1,8 +1,7 @@
-import { FactoryProvider } from '@angular/core/src/di/provider';
```
esotrope: changed imports from `core/src/...` to `core` to avoid issues if they move the internal file location later

PR #1803, opened 2019-03-22
fixes build

esotrope:


webapp/src/main/webapp/src/app/app-settings.ts
```diff
@@ -2,37 +2,47 @@
...
+  readonly userGuideUrl: string;
+  readonly state: StateSettings;
   readonly targetReport: TargetReportSettings;
+  readonly transferAccess: boolean;
+}
+
+export interface StateSettings {
+  readonly code: string;
+  readonly name: string;
 }
```
esotrope: before we were doing a hard cast to `<AppSettings>` when i took this away TS noticed that the interface didn't actually have the state field on it

malaffoon: Good thing we have all those tests!  ;)

PR #1805, opened 2019-03-23
minimal changes for reporting to work with configurable subjects schema change

malaffoon:
Because most of the configurable subject changes to reporting are additive, there were just a couple tweaks to keep things running.

@esotrope, this (obviously) doesn't make all the changes we'll need to get the configurable subject changes into the UI and backing API calls.

### DIFF
--- a/common/src/main/resources/common.sql.yml
+++ b/common/src/main/resources/common.sql.yml
@@ -1,10 +1,12 @@
 sql:
   claim:
+    # TODO - deal with ALT scores and remove "WHERE scs.score_type_id=3"
     findAll: >-
       SELECT scs.id, scs.code, s.code as subject_code, at.code as asmt_type_code
       FROM subject_score scs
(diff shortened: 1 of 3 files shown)


common/src/main/resources/common.sql.yml
```diff
@@ -2,7 +2,7 @@ sql:
   claim:
     findAll: >-
       SELECT scs.id, scs.code, s.code as subject_code, at.code as asmt_type_code
-      FROM subject_claim_score scs
+      FROM subject_score scs
         JOIN subject s ON scs.subject_id = s.id
         JOIN asmt_type at ON at.id = scs.asmt_type_id
       ORDER BY scs.display_order
```
malaffoon: Hmm. I wonder if i should put a temporary filter on these results, `WHERE score_type_id=3`, so it only gets the CLAIM scores until we've finished coding things.

PR #1807, opened 2019-03-25
enhance SubjectDefinition to have new configurable subject info

malaffoon:
Baby step #1.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
@@ -13,8 +13,11 @@ public class SubjectDefinition {
     private int performanceLevelCount;
     private Integer performanceLevelStandardCutoff;
     private Integer claimScorePerformanceLevelCount;
+    private Integer altScorePerformanceLevelCount;
     private boolean targetReport;
+    private boolean printedReport;
     private List<String> scorableClaims;
+    private List<String> altScores;
 
     public String getAsmtTypeCode() {
         return asmtTypeCode;
@@ -36,12 +39,24 @@ public class SubjectDefinition {
         return claimScorePerformanceLevelCount;
     }
 
+    public Integer getAltScorePerformanceLevelCount() {
+        return altScorePerformanceLevelCount;
+    }
+
     public boolean isTargetReport() {
         return targetReport;
     }
 
+    public boolean isPrintedReport() {
+        return printedReport;
+    }
+
     public List<String> getScorableClaims() {
         return scorableClaims == null ? ImmutableList.of() : scorableClaims;
+    }
+
+    public List<String> getAltScores() {
+        return altScores == null ? ImmutableList.of() : altScores;
     }
 
     public Builder copy() {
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
```diff
@@ -13,8 +13,11 @@
     private int performanceLevelCount;
     private Integer performanceLevelStandardCutoff;
     private Integer claimScorePerformanceLevelCount;
+    private Integer altScorePerformanceLevelCount;
     private boolean targetReport;
+    private boolean printedReport;
     private List<String> scorableClaims;
+    private List<String> altScores;
```
malaffoon: It may be that we need more information than this for alt-scores ... we'll see.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/JdbcSubjectDefinitionRepository.java
```diff
@@ -20,25 +26,43 @@
     private final NamedParameterJdbcTemplate template;
```
malaffoon: I'm opinionated on this: previously this repo was populating only the "flat" fields, it wasn't adding the list of claim codes. Instead, the service above this was merging in the claim codes. That really doesn't make sense to me: the repository is responsible for hydrating the object from the database. Period. So i changed it to do that, and to populate the new alt-score list.

common/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultSubjectDefinitionService.java
```diff
@@ -24,39 +16,15 @@
...
-                                           final ClaimRepository claimRepository) {
+    public DefaultSubjectDefinitionService(final SubjectDefinitionRepository definitionRepository) {
         this.definitionRepository = definitionRepository;
-        this.claimRepository = claimRepository;
     }
 
     @Override
     @Cacheable("subjectdefinitions")
     public Collection<SubjectDefinition> findAll() {
-        final Collection<SubjectDefinition> definitions = definitionRepository.findAll();
```
malaffoon: see note above: i don't think the service should be doing multiple repository calls for this.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
```diff
@@ -36,14 +39,26 @@ public Integer getClaimScorePerformanceLevelCount() {
...
+    public Integer getAltScorePerformanceLevelCount() {
+        return altScorePerformanceLevelCount;
+    }
+
     public boolean isTargetReport() {
         return targetReport;
     }
 
+    public boolean isPrintedReport() {
+        return printedReport;
```
esotrope: this is like a "supported in printed reports" flag right?

common/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultSubjectDefinitionService.java
esotrope: 👍

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/SubjectDefinition.java
malaffoon: Yup. I'm beginning to worry about the naming because it specifically refers to the ISR.

PR #1811, opened 2019-03-26
report names now support any character

esotrope:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/converter/DefaultReportHandler.java
```diff
@@ -43,7 +43,7 @@ public void writeResource(final ReportContentResource resource, final HttpOutput
      * @param headers   The response http headers (the value is modified)
      */
     protected void setHeaders(final ReportContentResource resource, final HttpHeaders headers) {
-        final String filename = resource.getReport().getLabel().replaceAll("\\s", "_") + "." + getExtension(resource.getMediaType());
+        final String filename = FileNames.format(resource.getReport().getLabel(), resource.getMediaType());
```
esotrope: idea here is that now the UI allows any character to be input as a report name but we sanitize it when they click download so the filename is permitted on the file system

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.html
```diff
@@ -722,18 +722,6 @@ <h3>
                   'aggregate-report-form.default-report-name' | translate
                 }}"
               />
-              <div *ngIf="showErrors('reportName')">
```
esotrope: dropping validation messages

webapp/src/main/webapp/src/app/aggregate-report/query-forms/claim-report-form.component.ts
```diff
@@ -111,12 +111,7 @@ export class ClaimReportFormComponent extends MultiOrganizationQueryFormComponen
               };
         }
       ],
-      reportName: [
-        this.settings.name,
-        fileName({
```
esotrope: dropping validation constraint

webapp/src/main/webapp/src/app/report/component/printable-report-form-modal/printable-report-form-modal.component.ts
```diff
@@ -157,8 +157,8 @@ export class PrintableReportFormModalComponent implements OnDestroy {
   }
 
   onCancelButtonClick(): void {
-    this.close();
     this.cancelled.emit();
+    this.close();
```
esotrope: uncrossing streams again x l
I'm going to play around with how i can add a test for this now.

webapp/src/main/webapp/src/app/shared/form/validators.ts
```diff
@@ -46,21 +46,3 @@ export function isGreaterThan(
...
   };
 }
-
-/**
- * Form control validator that makes sure the control value is a valid filename
- *
- * @param properties the properties to propagate when the control value is invalid
- * @return {ValidatorFn}
- */
-export function fileName(
```
esotrope: no longer used

webapp/src/main/webapp/src/app/report/reports.component.ts
```diff
@@ -127,6 +130,11 @@ export class ReportsComponent implements OnInit, OnDestroy {
         )
       );
     });
+    modal.userReportCreated.subscribe(created => {
+      console.log({ created });
+      this.userReportStore.setState([created, ...this.userReportStore.state]);
+      this.openReportTab();
```
esotrope: now when you run a report from a saved query it opens the reports tab back up for you to view its status

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/util/FileNames.java
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @param text The text to sanitize
+     */
+    private static String sanitize(@NotNull final String text) {
+        return text.replaceAll("[<>:;,%?*|\"]", "");
+    }
+
+    /**
+     * Formats the given text and media type into a file name.
+     * This method removes all invalid file system name characters and replaces spaces with underscores
```
malaffoon: Replacing tabs and stuff with underscores is probably necessary but don't most OS's allow spaces in file names? I'm not suggesting a change because spaces in a filename can be annoying to deal with, just curious if you thought about it.

esotrope: yeah i actually think spaces are fine since they are technically allowed. Only reason i left it was to maintain functional parity

webapp/src/main/webapp/src/app/report/reports.component.ts
malaffoon: 👍

PR #1813, opened 2019-03-26
Feature/as outlined buttons

esotrope:
image

image

image


webapp/src/main/webapp/src/btn-outlined.less
```diff
@@ -0,0 +1,53 @@
+// Button variants
+//
+// Easily pump out default styles, as well as :hover, :focus, :active,
+// and disabled options for all buttons
+
+.btn-outlined(@color; @background; @border) {
```
esotrope: thanks to @aaronsleeper for his loyal contributions to this project ; J

malaffoon: 👍

PR #1814, opened 2019-03-27
add alt score codes to exams

malaffoon:
@esotrope please think about the ScaleScore question. We can address that independently.


common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/ScaleScores.java
```diff
@@ -1,31 +1,109 @@
 package org.opentestsystem.rdw.reporting.common.jdbc;
```
malaffoon: Moved all the helper methods for ScaleScores in here (deleted ClaimScores helper class)

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/ScaleScores.java
```diff
@@ -1,31 +1,109 @@
...
+        for (final String prefix : prefixes) {
+            final String code = getNullable(row, row.getString(prefix + "score_code"));
+            if (code == null || code.isEmpty()) break;
+
+            final Integer level = getNullable(row, row.getInt(prefix + levelSuffix));
+            final Integer score = getNullable(row, row.getInt(prefix + "scale_score"));
+            if (level == null && score == null) {
+                scaleScores.add(null);
+            } else {
+                // TODO - ScaleScore should allow null level/value - but will that ripple?
```
malaffoon: @esotrope, this.

For ELPAC, there will be no claim scores (the test vendor doesn't have them), only claim levels. Right now, that means we get an exam payload with stuff like this in it (note claim score objects):
```
      "scaleScore": {"level": 4, "value": 1787, "standardError": 30.0},
      "claimScaleScores": [
        {"level": 3, "value": 0, "standardError": 0.0},
        {"level": 3, "value": 0, "standardError": 0.0},
        {"level": 3, "value": 0, "standardError": 0.0},
        {"level": 3, "value": 0, "standardError": 0.0}
      ],
      "altScaleScores": [
        {"level": 4, "value": 1792, "standardError": 41.0},
        {"level": 4, "value": 1782, "standardError": 39.0}
      ],
```
Personally, i think it would be better if ScaleScore had nullable fields and didn't default to 0. How much ripple effect do you think that will have in the UI if i change it?

reporting-service/src/test/java/org/opentestsystem/rdw/reporting/TestData.java
```diff
@@ -443,6 +452,7 @@
                 .completenessCode("Partial")
                 .scaleScore(null)
                 .claimScaleScores(newArrayList(null, null, null))
+                .altScaleScores(newArrayList((ScaleScore) null))
```
malaffoon: Just like claim scores, if an assessment has alt scores defines then an unscored exam will have an array of null score objects.

common/src/main/java/org/opentestsystem/rdw/reporting/common/jdbc/ScaleScores.java
esotrope: we'll just have to try and see what breaks i think

PR #1820, opened 2019-03-31
ScaleScore should allow values to be null 

malaffoon:
Not all scores will have values and/or levels set. 

I ran this locally and everything seemed to work fine with ELPAC claims, which have no score/std-err.


reporting-service/src/test/java/org/opentestsystem/rdw/reporting/TestData.java
```diff
@@ -429,15 +429,15 @@
                 .sessionId("session5")
                 .dateTime(Instants.from("1997-01-05"))
                 .scaleScore(
-                        ScaleScore.builder().level(2).value(2400).standardError(24).build()
+                        ScaleScore.builder().level(2).value(2400).standardError(24.0).build()
                 )
                 .claimScaleScores(newArrayList(
-                        ScaleScore.builder().level(1).value(1200).standardError(120).build(),
-                        ScaleScore.builder().level(2).value(2200).standardError(220).build(),
-                        ScaleScore.builder().level(3).value(3200).standardError(320).build()
```
malaffoon: I tweaked one of the test exams to have null values for claims.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -973,7 +973,7 @@
       },
       "instructional-resources": {
         "button": "Instructional Resources",
-        "description": "Upload instructional resources in the system."
+        "description": "Upload links to instructional resources in the system."
```
malaffoon: unrelated change from SB feedback

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ScaleScore.java
```diff
@@ -5,9 +5,9 @@
  */
 public class ScaleScore {
 
-    private int level;
-    private int value;
-    private double standardError;
+    private Integer level;
+    private Integer value;
+    private Double standardError;
```
esotrope: is it possible for some of the values to be null and not all? i think before we opted for an absent/null entry in an array to represent all of these values being null.

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: 👍

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/ScaleScore.java
malaffoon: Yes, that functionality is still there (handled in the service code): if all the values are missing a null value is stuffed in the ScaleScore list.

PR #1821, opened 2019-04-01
adds toggle button to alt scores

esotrope:
alt-scores

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/model/score-statistics.ts
+++ b/webapp/src/main/webapp/src/app/assessments/model/score-statistics.ts
@@ -1,6 +1,11 @@
 import { SubjectDefinition } from '../../subject/subject';
 
 /**
+ * Declares all the different score types
+ */
+export type ScoreType = 'Overall' | 'Claim' | 'Alternate';
+
+/**
  * High level set of statistics computed from a set of exams for a specific assessment
  */
 export interface ScoreStatistics {
--- a/webapp/src/main/webapp/src/app/assessments/model/score-statistics.ts
+++ b/webapp/src/main/webapp/src/app/assessments/model/score-statistics.ts
@@ -3,7 +3,7 @@ import { SubjectDefinition } from '../../subject/subject';
 /**
  * Declares all the different score types
  */
-export type ScoreType = 'Overall' | 'Claim' | 'Alternate';
+export type ScoreType = 'Overall' | 'Alternate' | 'Claim';
 
 /**
  * High level set of statistics computed from a set of exams for a specific assessment
--- a/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
@@ -313,17 +313,6 @@
       {{ 'common.buttons.display-overall' | translate }}
     </button>
     <button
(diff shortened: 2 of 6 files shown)


webapp/src/main/webapp/src/app/assessments/model/score-statistics.ts
```diff
@@ -1,5 +1,10 @@
 import { SubjectDefinition } from '../../subject/subject';
 
+/**
+ * Declares all the different score types
+ */
+export type ScoreType = 'Overall' | 'Claim' | 'Alternate';
```
malaffoon: Can we put `Alternate` between the other two?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -403,6 +403,7 @@
     },
     "buttons": {
       "all": "All",
+      "display-alternate": "Alt Scores",
```
malaffoon: This label (and `"Claim Scores"`) should come from, for example, `subject.ELPAC.asmt-type.sum.alt-score.name` and `subject.ELPAC.asmt-type.sum-claim-score.name` (substituting the subject code and asmt-type code of course).

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -1,320 +1,339 @@
...
+    <button
+      *ngIf="subjectDefinition.alternateScores"
+      (click)="onScoreTypeChange('Alternate')"
+      class="btn btn-default z-index-over-data-table"
+      angulartics2On="click"
+      angularticsAction="ViewClaimScores"
+      angularticsCategory="AssessmentResults"
+      [ngClass]="{ active: scoreType === 'Alternate' }"
+    >
+      {{ 'common.buttons.display-alternate' | translate }}
```
malaffoon: see note below about the key for this

webapp/src/main/webapp/src/app/assessments/results/average-scale-score.component.html
```diff
@@ -1,320 +1,339 @@
...
+      (click)="onScoreTypeChange('Claim')"
       class="btn btn-default z-index-over-data-table"
       angulartics2On="click"
       angularticsAction="ViewClaimScores"
       angularticsCategory="AssessmentResults"
-      [ngClass]="{ active: isClaimScoreSelected }"
+      [ngClass]="{ active: scoreType === 'Claim' }"
     >
-      <i class="fa fa-bold fa-table"></i>
       {{ 'common.buttons.display-claim' | translate }}
```
malaffoon: see note below about the key for this

webapp/src/main/webapp/src/app/assessments/model/score-statistics.ts
malaffoon: Okay, i guess the order here doesn't matter that much. But the buttons should probably have alt-scores in the middle.

esotrope: yeah i can do that

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: interesting ok. sounds like a feature enhancement ; j

PR #1824, opened 2019-04-03
deal with null std-err

malaffoon:
The std-err associated with a score may be null. This tweaks the UI and CSV export to deal with that.

### DIFF
--- a/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
+++ b/webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
@@ -13,7 +13,10 @@ import { WritingTraitAggregate } from '../assessments/model/writing-trait-aggreg
 import { TranslateDatePipe } from '../shared/i18n/translate-date.pipe';
 import { TranslateNumberPipe } from '../shared/i18n/translate-number.pipe';
 import { ApplicationSettingsService } from '../app-settings.service';
-import { AggregateTargetScoreRow, TargetReportingLevel } from '../assessments/model/aggregate-target-score-row.model';
+import {
+  AggregateTargetScoreRow,
+  TargetReportingLevel
+} from '../assessments/model/aggregate-target-score-row.model';
 import { SubjectDefinition } from '../subject/subject';
 
 @Injectable()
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -252,37 +283,53 @@ export class CsvBuilder {
   withErrorBandMin(getExam: (item: any) => Exam) {
     return this.withColumn(
       this.translateService.instant('csv-builder.error-band-min'),
-      (item) => {
+      item => {
         const exam: Exam = getExam(item);
-        return !exam.score ? '' : exam.score - exam.standardError;
+        return exam.score && exam.standardError
+          ? exam.score - exam.standardError
+          : '';
```
malaffoon: The change.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -252,37 +283,53 @@ export class CsvBuilder {
...
   withErrorBandMax(getExam: (item: any) => Exam) {
     return this.withColumn(
       this.translateService.instant('csv-builder.error-band-max'),
-      (item) => {
+      item => {
         const exam: Exam = getExam(item);
-        return exam.score ? exam.score + exam.standardError : '';
+        return exam.score && exam.standardError
+          ? exam.score + exam.standardError
+          : '';
```
malaffoon: And here.

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -13,7 +13,10 @@ import { WritingTraitAggregate } from '../assessments/model/writing-trait-aggreg
 import { TranslateDatePipe } from '../shared/i18n/translate-date.pipe';
```
malaffoon: most of this is the auto-formatting, look for the comments for the only changes

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
esotrope: my only q is what should the behavior be when they are zero because in js
zero is treated as false.
My guess is this is more accurately expressed as
`exam.score != null && exam.standardError != null`

malaffoon: Hmm. An SEM of zero is non-sensical but, i guess for the CSV export, we should probably honor it. For the display in the GUI i don't think we want to do +- 0.

esotrope: 👍

PR #1827, opened 2019-04-03
Task/run prettier

esotrope:
ran prettier on all web app files with
```
prettier --write src/{app,test}/**/**/**/**/**/**/*.{ts,js,json,less,html,md}"
```
yes `/**/*.xyz` syntax is not sufficient *eyeroll*


malaffoon:
Thanks Landon.

PR #1828, opened 2019-04-04
increase query timeout for district export

malaffoon:


docker-compose.yml
```diff
@@ -54,8 +54,16 @@ services:
...
+# if running other services from IntelliJ, comment out the links entry (above) and
+# change the zuul route to use the mac docker bridge; for example:
+#        - zuul.routes.report-processor.url=http://host.docker.internal:8085
     environment:
       - spring.profiles.active=local-docker
+      - spring.datasource.url-server=host.docker.internal:3306
+      - zuul.routes.admin-service.url=http://admin-service:8080
+      - zuul.routes.aggregate-service.url=http://aggregate-service:8080
+      - zuul.routes.report-processor.url=http://report-processor-service:8080
+      - zuul.routes.reporting-service.url=http://reporting-service:8080
```
malaffoon: This is unrelated. When i was trying to debug this, i wanted to run report-processor in IJ while running everything else in docker. The way things were set up i couldn't do it. So i cleaned up the configuration for local-docker and moved the environment-specific settings in here. Not sure how people were doing this, perhaps running all the services from within IJ instead of mixing with docker?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/repository/JdbcExamRepository.java
```diff
@@ -37,30 +35,27 @@
...
+                        .addValues(securityParameterProvider.getSecurityParameters(permissionSource, query.isDisableTransferAccess()))
+                        .addValue("query_district_group_ids", nullOrEmptyToDefault(query.getDistrictGroupIds(), UNMATCHABLE_IDS))
+                        .addValue("query_district_ids", nullOrEmptyToDefault(query.getDistrictIds(), UNMATCHABLE_IDS))
+                        .addValue("query_school_group_ids", nullOrEmptyToDefault(query.getSchoolGroupIds(), UNMATCHABLE_IDS))
+                        .addValue("query_school_ids", nullOrEmptyToDefault(query.getSchoolIds(), UNMATCHABLE_IDS))
+                        .addValue("school_year", query.getSchoolYear())
+                        .addValue("disable_transfer_access", query.isDisableTransferAccess())
+                        .addValue("export_location", uriLocation),
+                ps -> {
+                    ps.setQueryTimeout(0);
```
malaffoon: This is the change. Since we're using an interceptor to set the timeout, we need to have our hands on the statement *after* it is created. This works.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/ExportExams.java
```diff
@@ -56,15 +56,11 @@ public void exportExams(final Message<ReportRequestMessage> message) {
...
-                .reportResourceUri(exportUri)
-                .build();
-
-        userReportService.update(updated);
+        try {
+            final URI exportUri = exportService.exportExams(user, (DistrictSchoolExportReportQuery) report.getQuery(), reportId);
+            userReportService.update(report.copy().status(ReportStatus.COMPLETED).reportResourceUri(exportUri).build());
+        } catch (final Exception e ) {
+            userReportService.update(report.copy().status(ReportStatus.FAILED).build());
+        }
```
malaffoon: Tangentially related: the timeout failures were not making their way back to the report status, so those reports were just left in PENDING state.

report-processor/src/main/resources/application.yml
```diff
@@ -97,7 +97,7 @@ spring:
       &characterEncoding=${spring.datasource.character-encoding:utf8}\
       &rewriteBatchedStatements=${spring.datasource.rewrite-batched-statements:true}\
       &connectTimeout=${spring.datasource.connect-timeout:10000}\
-      &socketTimeout=${spring.datasource.socket-timeout:130000}\
+      &socketTimeout=${spring.datasource.socket-timeout:600000}\
```
malaffoon: When the query timeout is disabled (setting it to 0 in the java code above) it falls back to the driver socket timeout. I've arbitrarily chosen 10 minutes as the maximum time. Note that the 2 minute query timeout for all the other queries is still in effect.

reporting-service/src/test/resources/application-test.yml
```diff
@@ -1,3 +1,8 @@
+logging:
+  level:
+    # emit statistics for the underlying context cache
+    org.springframework.test.context.cache: debug
+
```
malaffoon: Unrelated. @ftdholbrook discovered that many of our integration tests have distinct contexts. This logging shows the cache hits and misses. Something like 1800 hits and 48 misses during a full build of reporting-service. As Dave noted we should probably try to reduce the number of distinct contexts at some point. This is just sort of a reminder for that.

docker-compose.yml
esotrope: hm. i've never had an issue interchanging IJ-run services with docker run services - what was the specific scenario you had

malaffoon: Discussed offline: the issue is that i was running the webapp in docker and the backing service in IntelliJ. Up until now, it was always the opposite. Or running both in IntelliJ.

PR #1831, opened 2019-04-05
updates csv exports to include alt scores

esotrope:
teacher export

image

image

student history export

image

### DIFF
--- a/webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
+++ b/webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
@@ -13,18 +13,64 @@ import { ExportTargetReportRequest } from '../assessments/model/export-target-re
 import { Assessment } from '../assessments/model/assessment';
 import { ordering } from '@kourge/ordering';
 import { ranking } from '@kourge/ordering/comparator';
-import { Observable } from 'rxjs';
+import { forkJoin } from 'rxjs';
 import { SubjectService } from '../subject/subject.service';
-import { OrderingService } from '../shared/ordering/ordering.service';
-import { map } from 'rxjs/operators';
+import { isNullOrEmpty } from '../shared/support/support';
+
+/**
+ * Represents a specific type of score for an assessment (e.g. claim, alternate)
+ * This is used when collecting information about what score codes exist and what columns to add to the CSV
+ */
+interface AssessmentScoreType {
+  subjectCode: string;
+  assessmentTypeCode: string;
+  codes: string[];
+}
+
+/**
+ * Gets the score codes for the given assessment holding values
+ *
+ * @param values The assessment holders
+ * @param subjectCodes All system subject codes used for sorting
+ * @param getAssessment Method to get the assessment from the value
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -259,49 +391,4 @@ export class CsvExportService {
...
-
-  /**
-   * Scans the provided data for what scorable claims are present and returns a collection of
-   * subject - scorable claim collection pairs to iterate over when generating export columns
-   *
-   * @param {any[]} data the data to scan
-   * @param {(datum) => Assessment} getAssessment used to get the assessment from a data entry
-   * @returns {Observable<SubjectScorableClaims[]>} subject - scorable claim collection pairs present in the given data set
-   */
-  private getSubjectScorableClaims(
```
esotrope: refactored this to more generic "getScoreCodes" method to be used for alt and claim scores

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -673,6 +673,7 @@
       "assessment-exam-columns": {
         "alternateScaleScore": "{{name}} Scale Score",
         "alternateScaleScoreLevel": "{{name}} Level",
+        "alternateScaleScoreStandardError": "{{name}} Standard Error",
```
esotrope: sort of doesn't belong here as it is not an actual column in this table but it also sort of does as the export is supposed to represent the table so meh

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -13,18 +13,64 @@ import { ExportTargetReportRequest } from '../assessments/model/export-target-re
...
+): AssessmentScoreType[] {
+  return (values || [])
+    .reduce((scoreCodes: AssessmentScoreType[], value: T) => {
+      const assessment = getAssessment(value);
+      if (assessment != null && !isNullOrEmpty(getScoreCodes(assessment))) {
+        const { subject: subjectCode, type: assessmentTypeCode } = assessment;
+        const score = scoreCodes.find(
+          score => score.subjectCode === subjectCode
+        );
+        if (score == null) {
```
malaffoon: might help in three months to have a comment here why the score can be null?

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: i see your point but it seems fine to me, and i don't know where else it would go. I guess, in theory we could have a whole separate set of keys, `"assessment-exam-export-columns"` but, given that the export is supposed to mirror what is on the page, that seems ... silly.

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
esotrope: ah. this is just a poor-mans createIfAbsent

PR #1832, opened 2019-04-08
change show-more/show-less to show-results/hide-results

malaffoon:
show-results
hide-results


malaffoon:
@esotrope, i like the change in verbiage but i yield to you as to whether the component should be renamed or not. I thought changing the name was a good idea to discourage reuse in other places. It may make more sense to leave the name alone and parameterize the component, passing in the show/hide text. But i didn't want to do that cause i'd mess it up.

webapp/src/main/webapp/package-lock.json
```diff
@@ -96,10 +96,10 @@
           "integrity": "sha512-RZXPviBTtfmtka9n9sy1N5M5b82CbxWIR6HIis4s3WQTXDJamc/0gpCWNGz6EWdWp4DOfjzJfhz/AS9zVPjjWg==",
```
malaffoon: sigh, didn't mean to include this

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -73,10 +73,10 @@
     </ng-container>
   </div>
 
-  <show-more-divider
+  <show-results-divider
     [collapsed]="collapsed"
     (collapsedChanged)="collapsed = $event"
-  ></show-more-divider>
+  ></show-results-divider>
```
esotrope: actually thinking again this could be slightly improved by just exposing the text as a property that can be changed.
e.g.
```
@Input() label: string; 

<show-more-divider label="{{'my-label-key' | translate}}">

```

PR #1833, opened 2019-04-08
Task/refactor color service

esotrope:
started added alt scores to the student view and got side tracked standardizing the way we get colors for assessment types and grade codes


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-assessment-card.component.ts
```diff
@@ -18,68 +23,50 @@ export interface AssessmentCardEvent {
 
 @Component({
   selector: 'group-assessment-card',
-  templateUrl: './group-assessment-card.component.html'
+  templateUrl: './group-assessment-card.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
```
esotrope: optimized group and student card views in the process

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -164,13 +162,6 @@ export class StudentResultsComponent implements OnInit {
     });
   }
 
-  getAssessmentTypeColor(assessmentType: string): string {
-    const index = ['ica', 'iab', 'sum'].indexOf(assessmentType);
-    const totalAssessmentTypes = 3;
-    const colorIndex = index >= 0 ? index + 1 : totalAssessmentTypes;
-    return this.colorService.getColor(colorIndex);
-  }
```
esotrope: extracted to `colors.ts`

webapp/src/main/webapp/src/app/shared/colors.ts
```diff
@@ -0,0 +1,26 @@
...
+  'green-dark',
+  'blue-dark'
+];
+
+export function color(index: number): string {
+  return colors[index % colors.length];
+}
+
+export function gradeColor(gradeCode: string): string {
+  const index = gradeCodes.indexOf(gradeCode);
```
malaffoon: Does this make it so the `grade.xx.color` values from common en.json are no longer needed?

webapp/src/main/webapp/src/app/shared/colors.ts
```diff
@@ -0,0 +1,26 @@
...
+  return color(index >= 0 ? index : 0);
+}
+
+export function assessmentTypeColor(assessmentType: AssessmentType): string {
+  const index = assessmentTypes.indexOf(assessmentType);
```
malaffoon: I like the indirection approach.

webapp/src/main/webapp/src/app/shared/colors.ts
esotrope: hm.. wasn't aware of their presence. My guess is those are being used for server side PDFs and this was used on the UI. I'm thinking now we should only be using en.json...

esotrope: The upside to doing it in the code is we can apply fallback colors if the grade code or whatever is absent for some reason in the translation file this however should never be a real case though so i'm leaning toward single source of truth for this

PR #1835, opened 2019-04-09
makes the claim heading dynamic

esotrope:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/exam/component/claim-score-summary/claim-score-summary.component.html
+++ b/webapp/src/main/webapp/src/app/exam/component/claim-score-summary/claim-score-summary.component.html
@@ -2,7 +2,18 @@
   <div class="col-md-12">
     <strong
       info-button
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/exam/component/claim-score-summary/claim-score-summary.component.html
```diff
@@ -2,7 +2,18 @@
...
+            : {
+                scoreName:
+                  'subject.' +
+                    _subjectDefinition.subject +
+                    '.asmt-type.' +
+                    _subjectDefinition.assessmentType +
+                    '.claim-score.name' | translate
+              }
+      }}"
       content="{{ 'common.info.claim' | translate }}"
```
esotrope: Does this need to be subject-specific?

image

esotrope: yeah i think ideally it is subject driven OR we just remove it ; J

PR #1836, opened 2019-04-09
Feature/rp 135 respect printable report flag

esotrope:
removes "ELPAC" option from menu

image

removes "print {name} summative report" option from menu

image


webapp/src/main/webapp/src/app/report/provider/subject-field.provider.ts
```diff
@@ -2,19 +2,30 @@ import { TranslateService } from '@ngx-translate/core';
...
+    map(([values, subjectDefinitions]) => ({
       name: 'subject',
       type: 'select',
       label: translateService.instant('common.subject-select-label'),
-      options: [undefined, ...values].map(value => ({
+      options: [
+        undefined,
+        ...values.filter(
+          value =>
+            subjectDefinitions.find(({ subject }) => subject === value)
```
malaffoon: I guess we can refactor when a second subject field is needed but it seems like the part finding the matching subject should be a common bit of code, and extracting the desired field should be a functor or something.

PR #1837, opened 2019-04-09
Task/alt score demo feedback

esotrope:
image

image

image


webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -34,6 +34,15 @@ export function deepEqualsIgnoringNullAndFalse(a: Object, b: Object): boolean {
   );
 }
 
+/**
+ * Strips out HTML tags from the given string
+ *
```
malaffoon: Perhaps a comment as to why this exists so that, in three months, a developer doesn't just come along and "clean up" the source text?

PR #1845, opened 2019-04-12
fixes org grouping issue

esotrope:


webapp/src/main/webapp/src/app/organization-export/organization-export-queries.ts
```diff
@@ -139,5 +140,5 @@ function groupIfAllDescendantsSelected(
  * @returns {boolean} True if the node is a valid grouping node
  */
 function validGroupingNode(node: Tree<Organization>) {
-  return node.value.name != null;
+  return !isNullOrBlank(node.value.name);
```
esotrope: good thing a test caught this

esotrope: in a better solution we should somehow separate the data tree view of the orgs from the display tree view

PR #1848, opened 2019-04-15
fixes standard description key mapping

esotrope:


webapp/src/main/webapp/src/app/assessments/results/view/distractor-analysis/distractor-analysis.component.html
```diff
@@ -38,8 +38,8 @@
             'common.results.assessment-item-columns.standard' | translate
           }}"
           content="{{
-            'common.results.assessment-item-columns.standard-info.' +
-              assessment.subject | translate
+            'subject.' + assessment.subject + '.standard-description'
```
esotrope: the replaced code did not even have corresponding text in the en.json anymore - this now maps directly to the subject definitions "status-description"

PR #1849, opened 2019-04-15
makes student exam table refresh everytime params change

esotrope:


webapp/src/main/webapp/src/app/student/component/student-exam-table/student-exam-table.component.ts
```diff
@@ -93,28 +90,19 @@ function createClaimColumns(scoreCodes: string[]): Column[] {
   selector: 'student-exam-table',
   templateUrl: './student-exam-table.component.html'
 })
-export class StudentExamTableComponent implements OnInit {
-  @Input()
-  student: Student;
-
+export class StudentExamTableComponent {
```
esotrope: instead of making one of these per card we are making one per row to support the layout. this meant the component only initialized once per display row instead of per-exam. now it refreshes every time the subject def changes.

malaffoon:
Should this be merged into `develop`?

PR #1851, opened 2019-04-16
makes display of grouping control conditional on standard cutoff config

esotrope:
This will theoretically disable the score level grouping in the aggregate result page but i have not yet been able to test this locally


malaffoon:
Figured it out Landon ... for local-docker, the school year is 2018. But embargo is on. So the 2018 results were being suppressed. I'll post a PR that bumps the default school year to 2019.

esotrope:
verified
image

PR #1858, opened 2019-04-25
adds change detection. confirmation when navigating away

esotrope:
change-detection

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-editor/pipeline-editor.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-editor/pipeline-editor.component.ts
@@ -8,6 +8,7 @@ import {
 import {
   CompilationError,
   Pipeline,
+  PipelineScript,
   PipelineTest,
   TestResult
 } from '../../model/pipeline';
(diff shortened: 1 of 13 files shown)


webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-editor/pipeline-editor.component.ts
```diff
@@ -89,6 +90,9 @@ export class PipelineEditorComponent {
   @Input()
   selectedItem: Item;
 
+  @Input()
+  selectedItemLoading: boolean = true;
```
esotrope: added support for lazy loading test content - still not sure if it will be necessary and would simplify things if we didn't bother

webapp/src/main/webapp/src/app/admin/ingest-pipeline/service/pipeline.service.stubs.ts
```diff
@@ -68,8 +68,165 @@ export const stubPipelineTests: PipelineTest[] = [1, 2, 3, 4, 5].map(
...
-  `,
-    input: '',
-    output: ''
+  `
   })
 );
+
+export const stubPipelineTest: PipelineTest = {
+  input: `<TDSReport>
+  <Test academicYear="2018" assessmentType="Interim" assessmentVersion="11111" bankKey="777" contract="MOCK" grade="11"
```
esotrope: added example trt from ingest

image

PR #1859, opened 2019-04-26
Sandbox Configuration

burzum619:
- Stubbing out backend calls - `sandbox.service.ts` will be completely reworked once the APIs are ready
- Temporarily keeping sandbox configuration mockData in memory - added TODOs to remove these mocks once backend is in place and integration is complete
- No integration for "Configuration Properties" section yet - this will be done in a separate PR
- For now, all sandboxes are editable in a single screen. May refactor this and create a completely separate screen for edit depending on client feedback next week

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
@@ -1,0 +1,95 @@
+import { Component } from '@angular/core';
+import { Router } from '@angular/router';
+import { DataTemplate } from './sandbox-configuration';
+import { SandboxService } from './sandbox.service';
+import {
+  FormBuilder,
+  FormGroup,
+  FormArray,
+  Validators,
+  FormControl
+} from '@angular/forms';
+import { RdwTranslateLoader } from '../../shared/i18n/rdw-translate-loader';
+import { SandboxConfigurationProperty } from './sandbox-configuration-property';
+
+@Component({
+  selector: 'new-sandbox',
+  templateUrl: './new-sandbox.component.html'
+})
+export class NewSandboxConfigurationComponent {
+  dataTemplates: DataTemplate[];
+  sandboxForm: FormGroup;
+  // Contains the full list of localization overrides, with default values
+  localizationOverrides: SandboxConfigurationProperty[] = [];
+  // Contains the full list of configuration properties, with default values
+  configurationProperties: SandboxConfigurationProperty[] = [];
+  // An array containing localization overrides that have been modified
+  modifiedLocalizationOverrides: SandboxConfigurationProperty[] = [];
+  // An array containing configuration properties that have been modified
+  modifiedConfigurationProperties: SandboxConfigurationProperty[] = [];
+
+  constructor(
+    private service: SandboxService,
+    private formBuilder: FormBuilder,
+    private translationLoader: RdwTranslateLoader,
+    private router: Router
+  ) {}
+
+  ngOnInit(): void {
+    this.sandboxForm = this.formBuilder.group({
+      label: [null, Validators.required],
+      description: [null],
+      template: [null, Validators.required],
+      configurationProperties: this.formBuilder.array([]),
+      localizationOverrides: this.formBuilder.array([])
+    });
+
+    this.service
+      .getAvailableDataTemplates()
+      .subscribe(templates => (this.dataTemplates = templates));
+
+    this.translationLoader
+      .getFlattenedTranslations('en')
+      .subscribe(translations => {
+        let locationOverrideFormArray = <FormArray>(
+          this.sandboxForm.controls['localizationOverrides']
+        );
+        for (let key in translations) {
+          // check also if property is not inherited from prototype
+          if (translations.hasOwnProperty(key)) {
+            let value = translations[key];
+            this.localizationOverrides.push(
+              new SandboxConfigurationProperty(key, value)
+            );
+            locationOverrideFormArray.controls.push(new FormControl(value));
+          }
+        }
+      });
+  }
+
+  onSubmit(): void {
+    let newSandbox = {
+      ...this.sandboxForm.value,
+      localizationOverrides: this.modifiedLocalizationOverrides
+    };
+    this.service.create(newSandbox);
+    this.router.navigate(['sandbox']);
+  }
+
+  updateOverride(override: SandboxConfigurationProperty, index: number): void {
+    const overrides = <FormArray>(
+      this.sandboxForm.controls['localizationOverrides']
+    );
+    const newVal = overrides.controls[index].value;
+
+    if (this.modifiedLocalizationOverrides.indexOf(override) > -1) {
+      let existingOverride = this.modifiedLocalizationOverrides[
+        this.modifiedLocalizationOverrides.indexOf(override)
+      ];
+      existingOverride.value = newVal;
+    } else {
+      override.value = newVal;
+      this.modifiedLocalizationOverrides.push(override);
+    }
+  }
+}
--- a/webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
@@ -23,10 +23,6 @@ export class NewSandboxConfigurationComponent {
   localizationOverrides: SandboxConfigurationProperty[] = [];
   // Contains the full list of configuration properties, with default values
   configurationProperties: SandboxConfigurationProperty[] = [];
-  // An array containing localization overrides that have been modified
-  modifiedLocalizationOverrides: SandboxConfigurationProperty[] = [];
-  // An array containing configuration properties that have been modified
-  modifiedConfigurationProperties: SandboxConfigurationProperty[] = [];
 
   constructor(
     private service: SandboxService,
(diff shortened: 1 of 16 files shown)


webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
```diff
@@ -0,0 +1,98 @@
...
+  FormControl
+} from '@angular/forms';
+import { RdwTranslateLoader } from '../../shared/i18n/rdw-translate-loader';
+import { SandboxConfigurationProperty } from './sandbox-configuration-property';
+
+@Component({
+  selector: 'new-sandbox',
+  templateUrl: './new-sandbox.component.html'
+})
+export class NewSandboxConfigurationComponent {
```
esotrope: what's a new sandbox vs an existing one? how do we display things differently?

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-configuration.ts
```diff
@@ -0,0 +1,48 @@
...
+   * The map containing key/value pairings of configuration properties
+   */
+  configurationProperties?: SandboxConfigurationProperty[];
+}
+
+export interface DataTemplate {
+  /**
+   * The unique key for a given data template
+   */
+  key: string;
```
esotrope: i think we would want to call this `code` to be consistent with other models

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.html
```diff
@@ -0,0 +1,386 @@
+<div class="sandbox-container pt-md pb-md" *ngIf="!editMode">
```
esotrope: is sandbox-details the "existing sandbox" view?

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.ts
```diff
@@ -0,0 +1,159 @@
...
+  templateUrl: './sandbox-details.component.html'
+})
+export class SandboxConfigurationDetailsComponent implements OnInit {
+  @Input()
+  sandbox: SandboxConfiguration;
+  @ViewChild('resetModal')
+  resetModal;
+  @ViewChild('archiveModal')
+  archiveModal;
+  @ViewChild('deleteModal')
```
esotrope: i'd recommend eventually introducing separate modal components and using the `BsModalService` to launch them

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.ts
```diff
@@ -0,0 +1,159 @@
...
+  deleteModal;
+
+  expanded = false;
+  editMode = false;
+  configurationProperties: SandboxConfigurationProperty[] = [];
+  localizationOverrides: SandboxConfigurationProperty[] = [];
+  modifiedConfigurationProperties: SandboxConfigurationProperty[] = [];
+  modifiedLocalizationOverrides: SandboxConfigurationProperty[] = [];
```
esotrope: doesn't
`configurationProperties` and ` localizationOverrides` exist in `SandboxConfiguration` model?
why not just have a `sandboxConfiguration` field and maintain the overrides inside the current `formGroup.value` ?

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.ts
```diff
@@ -0,0 +1,159 @@
...
+          if (translations.hasOwnProperty(key)) {
+            const value = translations[key];
+            const localizationOverrides =
+              this.sandbox.localizationOverrides || [];
+            const override = localizationOverrides.find(
+              override => override.key === key
+            );
+            if (override) {
+              this.localizationOverrides.push(
+                new SandboxConfigurationProperty(
```
esotrope: i would separate out these methods.
to me creating the array of `SandboxConfigurationProperty`s and creating `FormArray` are two separate tasks.
also not sure why the service isn't delivering the sandbox configuration model in it's entirety

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.ts
```diff
@@ -0,0 +1,159 @@
...
+      key: this.sandbox.key,
+      template: this.sandbox.template,
+      ...this.sandboxForm.value,
+      localizationOverrides: this.modifiedLocalizationOverrides
+    };
+    this.service.update(newSandbox);
+    this.editMode = false;
+  }
+
+  updateOverride(override: SandboxConfigurationProperty, index: number): void {
```
esotrope: is there any way we can make `new-sandbox` and `sandbox-details` the same component?

webapp/src/main/webapp/src/app/admin/sandbox/sandbox.service.ts
```diff
@@ -0,0 +1,83 @@
+import { Observable } from 'rxjs';
+import { Injectable } from '@angular/core';
+import { DataService } from '../../shared/data/data.service';
+import { DataTemplate, SandboxConfiguration } from './sandbox-configuration';
+
+/**
+ * Service responsible for managing organization embargo settings
+ */
+@Injectable()
```
esotrope: `providedIn: 'root'`

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -148,6 +150,37 @@ export const adminRoute = {
           resolve: { embargoesByOrganizationType: EmbargoResolve }
         }
       ]
+    },
+    {
+      path: 'sandbox',
```
esotrope: to be 'restfull' this should be `sandboxes`

webapp/src/main/webapp/src/app/shared/i18n/rdw-translate-loader.ts
```diff
@@ -58,6 +76,26 @@ export class RdwTranslateLoader implements TranslateLoader {
       .pipe(catchError(() => EmptyObservable));
   }
 
+  private flattenJsonObject(ob): any {
```
esotrope: i really hope lodash has this already - can you check?

webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
burzum619: Maybe the naming can be better. This is the "create" screen for creating a new sandbox.

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.html
burzum619: `sandbox-details` is the child of the `sandbox` component, which is the view containing all sandbox. If we decide to break the "edit" into its own screen as opposed to part of the main view where all sandboxes are listed, this will likely be renamed `sandbox-edit` and will not be a child component of `sandbox` - it will be its own child page.

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.ts
burzum619: I can add a story to handle this refactor work

webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
burzum619: `AddSandbox...` more appropriate naming convention?

webapp/src/main/webapp/src/app/shared/i18n/rdw-translate-loader.ts
burzum619: I peeked - I couldn't find anything relevant on lodash docs or on any SOs.  There's another small npm library someone recommended in the comment here, but figured we might as well write our own method at that point
https://stackoverflow.com/questions/44076106/flatten-nested-objects-with-lodash

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.ts
burzum619: At the moment that are pretty distincts (although I agree that they share some basic characteristics). I think I will refactor in this way if we decided that editing a single sandbox at a time (and moving it to its own screen) is going to be the way to go.

webapp/src/main/webapp/src/app/admin/sandbox/sandbox-details.component.ts
burzum619: The service delivers the "overrides" - it doesn't include the full list of all configuration parameters, hence the need here to check if an override exists before setting the value of the form control.

webapp/src/main/webapp/src/app/admin/sandbox/new-sandbox.component.ts
esotrope: main question here is how do we display the view differently when it is new vs existing? can it be the same view?

burzum619: Yeah, at the moment the views are significantly different (with the exception of the accordions with the datatables). I don't think it makes sense to combine them, although I will reconsider if we end up refactoring the edit page into its own individual page

PR #1865, opened 2019-04-29
Change high school assessment grades to be the grade instead of "HS";…

malaffoon:
… this is an accepted short-term fix for the conflict between SB assessments and those from other vendors.


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -372,7 +368,11 @@
       "06": "G6",
       "07": "G7",
       "08": "G8",
-      "09": "HS",
+      "09": "G9",
+      "10": "G10",
+      "11": "G11",
+      "12": "G12",
+      "13": "G13",
```
esotrope: so no more references to "High School"?
We may also need to change this in the `src/java/resources/en.json` for report processor as well

PR #1866, opened 2019-04-29
disables pipeline test buttons when there are unsaved changes. adds h…

esotrope:
…elpful tooltips for this behavior

image

image

image

image


webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
```diff
@@ -360,7 +362,38 @@ export class PipelineComponent implements ComponentCanDeactivate {
   }
 
   private setPipelineTests(tests: PipelineTest[]): void {
-    this.testButtonDisabled = this.testing || tests.length === 0;
     this.pipeline.tests = tests;
+    this.updateButtonStates();
+  }
+
+  private updateButtonStates(): void {
+    // TODO need to move back to reactive approach with observables to make this simpler
```
esotrope: moved away from having everything be an observable/behavior subject for simplicity but now i'm thinking i should move back because of all this manual "updateState()" method calling on dependent state changes - yuck

webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
```diff
@@ -360,7 +362,38 @@ export class PipelineComponent implements ComponentCanDeactivate {
...
+
+    const hasUnsavedChanges =
+      this.items.some(({ type, changed }) => type === 'Script' && changed) ||
+      (this.selectedItem.type === 'Script'
+        ? this.items.some(({ type, changed }) => type === 'Test' && changed)
+        : this.selectedItem.changed);
```
esotrope: the nuance here is that if you are editing a test and that test and the script are saved you should be able to run that one test without having to save another test. if you are editing the script the test button should be disabled if any test is not saved - i guess i should add this as a comment instead of explaining here huh

webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-editor/pipeline-editor.component.html
```diff
@@ -9,73 +9,91 @@ <h3>Assessments Pipeline</h3>
...
-        (click)="scriptPublish.emit(selectedItem.value)"
-      >
-        <ng-container *ngIf="publishing; else notPublished">
-          <i class="fa fa-spinner fa-pulse"></i> Publishing
-        </ng-container>
-        <ng-template #notPublished>
-          Publish
-        </ng-template>
-      </button>
+      <div [tooltip]="saveButtonDisabledTooltip">
```
esotrope: nasty annoying thing about the tooltip & popover utils is that they don't/can't? launch when applied to a disabled button

webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
malaffoon: 🥇

esotrope:
> Since i'm pretty sure i'll have opportunity to see the user workflow and provide more feedback ... rubber stamp for you!

yay.
I have to say those big stamp images you post are satisfying to see - maybe we should make a github plugin for it

PR #1867, opened 2019-04-29
makes user group form student filters permission based

esotrope:


webapp/src/main/webapp/src/app/user-group/user-group.component.ts
```diff
@@ -95,67 +96,67 @@ export class UserGroupComponent implements OnInit, OnDestroy {
...
-                value: code,
-                text: this.translate.instant('subject.' + code + '.name'),
-                analyticsProperties: {
-                  label: `Subject: ${code}`
-                }
+    ).subscribe(([options, studentFormOptions, filterOptions]) => {
+      // TODO get options etc from this
+      this.studentFields = this.examSearchFilterService
+        .getExamSearchFilters()
+        .pipe(map(({ studentFilters }) => studentFilters.map(({ id }) => id)));
```
esotrope: this assignment is the only real change. the only other thing i did in this file was remove the app settings retrieval as it is no longer used

PR #1872, opened 2019-05-02
Task/demo feedback for pipeline scripts

esotrope:
Now when compilation fails it displays a message at the top to that effect.
This message will be displayed after on-demand compilation when editing code and also if compilation fails during test & publish processes.

image

Added validation to test forms.
Now running a test or all tests will be prevented when the tests are invalid or "not runnable"

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/code-editor/code-editor.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/code-editor/code-editor.component.ts
@@ -13,6 +13,7 @@ import { ControlValueAccessor } from '@angular/forms';
 import { map } from 'rxjs/operators';
 
 declare var ace: any;
+declare var UndoManager: any;
 
 export type MessageType = 'error' | 'information' | 'warning';
 
@@ -120,6 +121,8 @@ export class CodeEditorComponent implements ControlValueAccessor {
     this.withEditor(editor => {
       editor.setValue(value || '');
       editor.clearSelection();
+      // start change history after initialization
+      editor.getSession().setUndoManager(new ace.UndoManager());
     });
   }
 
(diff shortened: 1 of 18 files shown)


webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/code-editor/code-editor.component.ts
```diff
@@ -120,6 +121,8 @@ export class CodeEditorComponent implements ControlValueAccessor {
     this.withEditor(editor => {
       editor.setValue(value || '');
       editor.clearSelection();
+      // start change history after initialization
+      editor.getSession().setUndoManager(new ace.UndoManager());
```
esotrope: there was a fun bug that if you just open an editor and hit `ctrl + z` the entire script goes blank : J

webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-editor/pipeline-editor.component.html
```diff
@@ -123,27 +134,27 @@ <h3>Assessments Pipeline</h3>
     (deleteTestButtonClick)="testDelete.emit($event)"
   ></pipeline-explorer>
 
-  <div>
+  <div *ngFor="let item of items" [hidden]="item !== selectedItem">
```
esotrope: this gives each item (test or script) it's own right hand side view that is just shown or hidden instead of keeping one view and just changing the model associated with it (this is less error prone and requires less dom manipulation)

webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
```diff
@@ -262,38 +272,46 @@ export class PipelineComponent implements ComponentCanDeactivate {
             });
         } else {
           this.testState = null;
+          this.compilationState = 'Failed';
         }
       });
   }
 
   onTestCreate(): void {
```
esotrope: @malaffoon made this a client side op - the record will be inserted on the first save

PR #1875, opened 2019-05-03
Feature/cloud bus

ftdholbrook:
RP-301 config service - modify for RDW use case and verify client usage of @RefreshScope and Cloud Bus

- Force to do some refactor on DatabaseStatusIndicator implementations after aop wrapping for @RefreshScope JdbcTemplate no longer available in scope for auto-injection, have to fall back to actual concrete NamedParameterJdbcTemplate.  Made minimal possible change by using NamedParameterJdbcTemplate.getJdbcOperations()

-Reduced scope of global interceptors to remove conflict with cloud bus


docker-compose.yml
```diff
@@ -84,6 +90,9 @@ services:
       - spring.profiles.active=local-docker
       - CONFIG_SERVICE_ENABLED=true
       - CONFIG_SERVICE_URL=http://config-server:8888
+      - SPRING_RABBITMQ_ADDRESSES=rabbitmq:5672
+      - SPRING_RABBITMQ_USERNAME=guest
+      - SPRING_RABBITMQ_PASSWORD=guest
```
ftdholbrook: @malaffoon Any option on this for configuration?  Should we add a config section to each of the .yml files and allow some indirection with fallback, or use this style (similar to config service).

admin-service/src/main/java/org/opentestsystem/rdw/admin/status/DatabaseStatusIndicator.java
```diff
@@ -24,11 +25,11 @@
...
     private String sqlReadTest;
 
-    private final JdbcTemplate jdbcTemplate;
+    private final JdbcOperations jdbcOperations;
 
     @Autowired
-    DatabaseStatusIndicator(final JdbcTemplate jdbcTemplate) {
-        this.jdbcTemplate = jdbcTemplate;
+    DatabaseStatusIndicator(final NamedParameterJdbcTemplate namedParameterJdbcTemplate) {
+        this.jdbcOperations = namedParameterJdbcTemplate.getJdbcOperations();
```
malaffoon: Curious about this. Is it because the template creation isn't going through the same steps?

admin-service/src/main/resources/application.yml
```diff
@@ -41,6 +41,9 @@ spring:
     maxActive: 10
     minIdle: ${spring.reporting_datasource.initialSize}
     maxIdle: ${spring.reporting_datasource.maxActive}
+    removeAbandoned: false
+    removeAbandonedTimeout: 60
+    logAbandoned: false
```
malaffoon: Shouldn't all these be defaulting to `removeAbandoned: true`?

docker-compose.yml
malaffoon: I'll ask the obvious question: i know it needs this to subscribe to the config event bus but is there any way the apps can ask the config service itself for this info? You know, directly ask for the configuration and *then* subscribe?

admin-service/src/main/resources/application.yml
ftdholbrook: for now I'm just setting them to the defaults https://tomcat.apache.org/tomcat-8.5-doc/jdbc-pool.html

I'm going to make another ticket to actually do the tuning, so those are effectively placeholders.

report-processor/src/main/resources/application.yml
```diff
@@ -116,6 +116,9 @@ spring:
     minIdle: ${spring.datasource.initialSize}
     maxIdle: ${spring.datasource.maxActive}
     maxWait: 10000
+    removeAbandoned: false
+    removeAbandonedTimeout: 60
```
malaffoon: This one needs to be larger. The default is 2m for all report-processor queries and we disable that default and fall back to the 10m socket timeout for the district export query.

admin-service/src/main/java/org/opentestsystem/rdw/admin/status/DatabaseStatusIndicator.java
ftdholbrook: It seems to cause the injection error right after I put the AOP annotation on the datasoruce.  I don't know the exact reason for it, but we don't actually have any plain JdbcTemplates configured.  I assume some spring magic was allowing for the downgrade, but the AOP wrapping broke it.

docker-compose.yml
ftdholbrook: yes it can, that's kind of what I'm asking, we need the rabbit config for the cloud bus to work, should we put it in the yml files, or leave it fully external like the config service.

webapp/src/main/resources/application.yml
```diff
@@ -6,6 +6,10 @@ management:
   health:
     redis:
       enabled: false
+#  endpoints:
+#    web:
+#      exposure:
+#        include: *
```
malaffoon: hint?

ftdholbrook: leftover garbage form early experimentation I'll remove.

docker-compose.yml
malaffoon: Oh. Well, my generally philosophy is to boot-strap as little as possible. So just enough to talk to the config service and no more. But i guess that means adding rabbit to the yml files for these services even though the services don't care about rabbit except for configuration? (am i getting the gist of the question yet?)

ftdholbrook: -it is an hard requirement for cloud bus
-if we don't want cloud bus (say for local always static config) we don't need it, and I don't think it will error (which is it's own problem, took a long time to figure out what was going on)
-it only matters if you want to update the config
-the rabbit dependency is already there, so seems like not a big deal to me

So to put it another way, if you remove that from docker compose, the app will work, but configuration will not update on a service unless you directly call the rest /refresh endpoint on that particular service.

My think was that I prefer obvious dependencies called out in the yml files, so I understand what config I need.  I think this hidden spring boot stuff is configuring, especially after moving away from it for a few years.

report-processor/src/main/resources/application.yml
ftdholbrook: for now I have  removeAbandoned: false (disabling the feature, also the default pool behavior).  When we tune it we absolutely have to make it line up with the other config options.

admin-service/src/main/resources/application.yml
ftdholbrook: https://jira.fairwaytech.com/browse/RP-311

PR #1877, opened 2019-05-07
Feature/273 pipeline script crud

esotrope:


admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineRepository.java
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Repository responsible for accessing available pipelines
+ */
+public interface PipelineRepository {
+
+    /**
+     * @return gets all pipelines
+     */
+    List<CodedEntity> findAll();
```
esotrope: wanted to make a `Pipeline` model that just extended CodedEntity but found that i would have to rework how that class does its builder so i gave up.. i mean postponed it

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/PipelineScript.java
```diff
@@ -0,0 +1,93 @@
+package org.opentestsystem.rdw.admin.model;
+
+import java.time.Instant;
+
+public class PipelineScript {
```
malaffoon: nit: javadocs

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineRepository.java
malaffoon: I'm confused, is this an old comment that no longer applies?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineRepository.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.admin.repository;
+
+import org.opentestsystem.rdw.reporting.common.model.CodedEntity;
+
+import java.util.List;
+
+/**
+ * Repository responsible for accessing available pipelines
+ */
+public interface PipelineRepository {
```
malaffoon: Perhaps add to the javadocs, something like `Pipeline represents the type of a pipeline, e.g. Exam or Group`. Otherwise, my mind goes immediately to thinking Pipeline is the whole pipeline.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineTestRepository.java
```diff
@@ -0,0 +1,46 @@
...
+     * @return The updated test
+     */
+    PipelineTest update(PipelineTest test);
+
+    /**
+     * Deletes a pipeline test
+     *
+     * @param test The test to delete
+     */
+    void delete(PipelineTest test);
```
malaffoon: Isn't delete normally by just id?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcPipelineScriptRepository.java
```diff
@@ -0,0 +1,111 @@
...
+        );
+    }
+
+    @Override
+    public PipelineScript findById(final int id) {
+        try {
+            return template.queryForObject(
+                    findById,
+                    new MapSqlParameterSource()
+                            .addValue("id", id),
```
malaffoon: nit: should probably be consistent (on line 42 you use the constructor version)

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineTestRepository.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Repository responsible for managing pipeline scripts
+ */
+public interface PipelineTestRepository {
+
+    /**
+     * @param pipelineId The pipeline ID
+     * @return all tests for the given pipeline ID
+     */
+    List<PipelineTest> findAllByPipeline(int pipelineId);
```
malaffoon: Since this is a list (with implied order), you might note what the expected order is in the javadocs.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineScriptRepository.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Repository responsible for managing pipeline scripts
+ */
+public interface PipelineScriptRepository {
+
+    /**
+     * @param pipelineId The pipeline ID
+     * @return all scripts for the given pipeline ID
+     */
+    List<PipelineScript> findAllByPipeline(int pipelineId);
```
malaffoon: Although physically modeled to allow multiple scripts, logically there is only one script per pipeline. So i'm fine with the repository method being this way, just wondering where that degeneration will happen. In the service? In the UI?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineRepository.java
esotrope: 👍

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineRepository.java
esotrope: hm. i'll have to check out how those did it

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineScriptRepository.java
esotrope: i was thinking the UI - that way we remain flexible

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineTestRepository.java
esotrope: Good point

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/PipelineTestRepository.java
esotrope: yeah, right now it uses the "updatedBy" field for the error message is basically the reason for accepting the whole object but i can change that

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcPipelineScriptRepository.java
esotrope: 👍

PR #1879, opened 2019-05-07
Sandbox/Tenant Configuration Bugfixes

burzum619:
- Fixing issue where line breaks were not being displayed for the description despite the textarea allowing linebreaks when editing
- Creating a custom validator to trim control values before checking for null/empty string (surprised the standard angular validator does not already do this!)
- Adding a nice message for when there are no sandboxes/tenants to display

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/validator/custom-validators.ts
+++ b/webapp/src/main/webapp/src/app/shared/validator/custom-validators.ts
@@ -1,0 +1,20 @@
+import { FormControl, ValidationErrors } from '@angular/forms';
+import { isString } from 'lodash';
+
+export class CustomValidators {
+  /**
+   * A static validator that trims input before checking if it is null or an empty string
+   * @param control
+   * @returns an error if the control value is null, or is composed of only whitespace
+   */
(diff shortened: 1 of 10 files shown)


webapp/src/main/webapp/src/app/shared/validator/custom-validators.ts
```diff
@@ -0,0 +1,20 @@
+import { FormControl, ValidationErrors } from '@angular/forms';
+import { isString } from 'lodash';
+
+export class CustomValidators {
+  /**
+   * A static validator that trims input before checking if it is null or an empty string
+   * @param control
+   * @returns an error if the control value is null, or is composed of only whitespace
+   */
+  static required(control: FormControl): ValidationErrors | null {
```
esotrope: i'd maybe name this something different lie "notBlank" and point out the shortcomings of `Validators.required` in the comment otherwise it seems like the developer might not know about `Validators.required` and coded it from scratch.
Also, maybe replace `: ValidationErrors | null` with `: { notBlank: boolean } | null`

PR #1880, opened 2019-05-08
pipeline service modeling

esotrope:


admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineScriptService.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Service responsible for managing pipeline scripts
+ */
+public interface PipelineScriptService {
+
+    /**
+     * @param pipelineId The pipeline entity ID
+     * @return All scripts for the given pipeline
+     */
+    List<PipelineScript> get(int pipelineId);
```
malaffoon: I'm beginning to think we should introduce a pipeline enum, like we do with some of the other reference types. That would make these methods more type-safe. This sort of thing ...
```
public enum Pipeline {
    Exam(1),
    Group(2),
    Assessment(3);

    private final int id;

    Pipeline(final int id) {
        this.id = id;
    }
    ...
```

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineScriptService.java
```diff
@@ -0,0 +1,52 @@
...
+     * @throws java.util.NoSuchElementException when there is no script for the given parameters
+     */
+    PipelineScript get(int pipelineId, int id);
+
+    /**
+     * Creates a pipeline script
+     *
+     * @param script The script to create
+     * @return The created script
+     * @throws java.util.NoSuchElementException when there is no script for the given parameters
```
malaffoon: c&p error

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineTestService.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Deletes a pipeline test
+     *
+     * @param id The test ID
+     * @throws java.util.NoSuchElementException when there is no test for the given parameters
+     */
+    void delete(int id);
+
+    /**
+     * Runs all test of a given pipeline on the given pipeline
```
malaffoon: nit: grammar, `Run all tests ...` ?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineTestService.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @param pipelineId The pipeline entity ID
+     * @return All tests for the given pipeline
+     */
+    List<PipelineTest> get(int pipelineId);
+
+    /**
+     * @param pipelineId The pipeline entity ID
+     * @param id The test entity ID
+     * @return All tests for the given pipeline
```
malaffoon: `The identified test for the pipeline`?

This does make me ask: why do we need the pipelineId if we have the test id?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineTestService.java
```diff
@@ -0,0 +1,71 @@
...
+     * @throws java.util.NoSuchElementException when there is no test for the given parameters
+     */
+    PipelineTest get(int pipelineId, int id);
+
+    /**
+     * Creates a pipeline test
+     *
+     * @param test The test to create
+     * @return The created test
+     * @throws java.util.NoSuchElementException when there is no test for the given parameters
```
malaffoon: c&p error?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineScriptService.java
esotrope: Hm. feels redundant in that if the pipeline id is a miss the repo will throw/ return null though

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineScriptService.java
esotrope: cp error - create wont throw i dont think.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineScriptService.java
esotrope: Also, if we change the IDs that's two places we've got to change no?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineTestService.java
esotrope: 👍

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineTestService.java
esotrope: because the input/output text _can_ be large the UI gets it lazily when the test is "opened"

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineTestService.java
esotrope: どうも すみません でした

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/PipelineTestValidation.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static org.opentestsystem.rdw.reporting.common.util.CollectionUtils.safeImmutableOf;
+
+/**
+ * Represents a pipeline test validation run
+ */
+public class PipelineTestValidation {
+
+    private PipelineTest test;
+    private List<String> errors;
```
esotrope: @gregcharles let me know if there will be more than a list of string error messages from the validation process and i'll correct this

PR #1881, opened 2019-05-08
Feature/reporting settings tenant

ftdholbrook:
make reporting settings tenant aware


admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
```diff
@@ -36,9 +40,28 @@
...
+    }
+
+    @Bean
+    @ConfigurationProperties("reporting")
+    @RefreshScope
+    public ReportingSystemSettings reportingSystemSettings() {
+        return new ReportingSystemSettings();
+    }
+
+    @Bean("reportingSystemPropertiesResolver")
```
malaffoon: nit: don't need the name

admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
```diff
@@ -36,9 +40,28 @@
...
+    @RefreshScope
+    ReportingSystemPropertiesResolver reportingSystemPropertiesResolver(ReportingTenantIdResolver reportingTenantIdResolver,
+                                                                        ReportingSystemSettings reportingSystemSettings) {
+        return new ReportingSystemPropertiesResolver(reportingTenantIdResolver, reportingSystemSettings);
+    }
+
     @Bean(name = "reportingOrganizationRepository")
```
malaffoon: ditto  (unless i'm missing something?)

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemPropertiesResolver.java
```diff
@@ -0,0 +1,165 @@
...
+
+    private Optional<ReportingSystemProperties> getResolvedReportingSystemProperties() {
+        //TODO: need an id -> human readable key indirection, just using id for now
+        final Optional<String> tenantId = tenantIdResolver.getTenantId();
+        if (!tenantId.isPresent()) {
+            logger.warn("No tenantId Found. Will fall back to default dataSource configuration.");
+            return Optional.empty();
+        }
+        final Optional<ReportingSystemProperties> dataSourceElementsTenant = tenantId
+                .flatMap(k -> Optional.ofNullable(reportingSystemSettings.getTenants().get(k)));
```
malaffoon: It's probably just me but i don't find this easy to read. I looked it up and i get it, but isn't `tenantId` just an optional string? Couldn't this just be `Optional.ofNullable(reportingSystemSettings.getTenants().get(tenantId.get()))`? Not really suggesting a change, just hoping you can help me understand why `flatMap` is even a method on Optional.

ftdholbrook: This came out of me not being able to diagnose why a value wasn't resolving to the value I thought it should (typo in the config file).  There were two possibilities, one there was no tenant defined, or there was one defined, but it didn't have an override set.  what you have is pretty much what I started with.  This could be overkill, and I can revert it back.  I'm missing functional Either style disjuctions that would cleanly let you explain which path you took in a flow.  Java 8 dosn't have them yet... looks like they may be in java 11.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/ApplicationConfiguration.java
```diff
@@ -58,6 +63,25 @@
...
+    }
+
+    @Bean
+    @ConfigurationProperties("reporting")
+    @RefreshScope
+    public ReportingSystemSettings reportingSystemSettings() {
+        return new ReportingSystemSettings();
+    }
+
+    @Bean("reportingSystemPropertiesResolver")
```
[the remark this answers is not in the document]
ftdholbrook: You are seeing the detritus of me fighting with spring auto-wire resolution.

admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
```diff
@@ -36,9 +40,28 @@
 })
 @EnableCaching
 @EnableGlobalMethodSecurity(prePostEnabled = true)
-@EnableConfigurationProperties(ReportingSystemSettings.class)
+@EnableConfigurationProperties
 public class ApplicationConfiguration {
 
+    @Bean("tenantIdResolver")
+    public ReportingTenantIdResolver tenantIdResolver() {
```
esotrope: yeah i'm curious about the bean naming too. do we need it if the return signature is `TenantIdResolver` here?

admin-service/src/test/java/org/opentestsystem/rdw/admin/configuration/TestApplicationConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Bean
+    @ConfigurationProperties("reporting")
+    public ReportingSystemSettings reportingSystemSettings() {
+        return new ReportingSystemSettings();
+    }
+
+    @Bean("reportingSystemPropertiesResolver")
+    ReportingSystemPropertiesResolver reportingSystemPropertiesResolver(ReportingTenantIdResolver reportingTenantIdResolver,
```
esotrope: could the other beans be package private too? i think by default we make them all public but i guess they never had to be? not sure about anything now : j

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceServiceTest.java
```diff
@@ -62,7 +62,7 @@
 
     @Before
     public void setup() throws Exception {
-        final ReportingSystemSettings systemSettings = new ReportingSystemSettings();
+        final ReportingSystemPropertiesImpl systemSettings = new ReportingSystemPropertiesImpl();
```
esotrope: if there is nothing to say but `Impl` we usually use the `DefaultXYZ` naming convention. If there is a strategy at play we go for `[Strategy]XYZ` naming

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AbstractReportingSystemProperties.java
```diff
@@ -0,0 +1,22 @@
+package org.opentestsystem.rdw.reporting.common.configuration;
+
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+public abstract class AbstractReportingSystemProperties implements ReportingSystemProperties {
```
esotrope: is this to house shared service props?

admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
ftdholbrook: Reporting tenant id resolver is a specific implementation of TenantIdResolver, early on in the switch spring was getting confused about the sub-classing.  I switched to more explicit naming, the BeanName and methodName match are belt-and-suspenders because I was getting sick of not finding problems until the app started.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemPropertiesResolver.java
```diff
@@ -0,0 +1,165 @@
...
+        return getResolvedReportingSystemProperties()
+                .map(ReportingSystemProperties::getStudentFields)
+                .orElse(reportingSystemSettings.getStudentFields());
+    }
+
+    @Override
+    public List<String> getEffectiveReportLanguages() {
+        return getResolvedReportingSystemProperties()
+                .map(ReportingSystemProperties::getEffectiveReportLanguages)
+                .orElse(reportingSystemSettings.getEffectiveReportLanguages());
```
esotrope: we resorted to this because we couldn't have spring flatten the ymls before setting up this bean right?

common/src/test/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemPropertiesTest.java
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Test loading of reporting reportingSystemProperties from a yaml file
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = {ReportingSystemPropertiesTest.TestConfiguration.class})
+@ActiveProfiles(profiles = {"test-reporting-settings", "test-reporting-settings_nv"})
+@EnableConfigurationProperties
+public class ReportingSystemPropertiesTest {
+
+    public static class TestConfiguration {
```
esotrope: could this be its own `@Configuration` file - is this different than the real one?

admin-service/src/test/java/org/opentestsystem/rdw/admin/configuration/TestApplicationConfiguration.java
ftdholbrook: in a test package, probably, but not that important

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultInstructionalResourceServiceTest.java
ftdholbrook: There are like 4 versions of that interface... and I had to pick one that was pretty much just the interface?  Impl seems like the norm in java for that.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AbstractReportingSystemProperties.java
ftdholbrook: There is one computed prop that is not backed by data, it is used by all the implementations.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemPropertiesResolver.java
ftdholbrook: Yes, it was either this pattern, or we would have to reference parent properties in the config files like `some-prop: ${parent.some-prop}` for every property.  We chose to put the complexity in the code, not the config.

PR #1884, opened 2019-05-08
Task/273 make ui match api

esotrope:
basically prep for the webapp services to match the backend ones


webapp/src/main/webapp/src/app/shared/layout/sb-breadcrumbs.component.ts
```diff
@@ -21,14 +21,33 @@ export const BreadCrumbsTitleDelimiter = ' < ';
  * data: { breadcrumb: { translate: 'code' } }
  * data: { breadcrumb: { translate: 'path.to.property', translateResolve: 'path.to.property.holder.in.route.data' } }
  * data: { breadcrumb: { resolve: 'path.to.property.in.route.data' }
+
+ * @deprecated use {@link BreadcrumbFactory}
  */
 export interface BreadcrumbOptions {
```
esotrope: found it impossible to make the breadcrumb text i wanted without changing this to support dynamic translation code generation based on resolved data

webapp/src/main/webapp/src/app/shared/layout/sb-breadcrumbs.component.ts
```diff
@@ -158,37 +175,48 @@ export class SbBreadcrumbs implements OnInit {
...
-        text = this.translateService.instant(options.translate);
-      }
-
-      return {
-        text: text,
-        routerLinkParameters: routerLinkParameters
-      };
-    }
-    if (options.resolve) {
+    if (typeof options === 'function') {
```
esotrope: added this new block and simplified the below block - thinking i'll rewrite the way we configure breadcrumbs to not need the below block anyway

PR #1889, opened 2019-05-10
Fix for  RP-323 Error: Cannot read property 'forEach' of undefined

ftdholbrook:
root cause: AOP proxy was being serialized fixed with Jackson @JsonView


ftdholbrook:
possibly a cleaner fix than this, but I couldn't find one.

PR #1890, opened 2019-05-10
Feature/272 implement pipeline apis

esotrope:


admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineScriptService.java
```diff
@@ -1,51 +1,37 @@
...
 /**
  * Service responsible for managing pipeline scripts
  */
 public interface PipelineScriptService {
 
     /**
-     * @param pipelineId The pipeline entity ID
-     * @return All scripts for the given pipeline
-     */
-    List<PipelineScript> get(int pipelineId);
```
esotrope: shuffled some stuff around - mostly moving things from pipeline test and script services into the generic pipeline service

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineService.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public List<Pipeline> getPipelines() {
+        return repository.findAll();
+    }
+
+    @Override
+    public List<PipelineScript> getPublishedScripts(final int pipelineId) {
+        // TODO
```
esotrope: the todos here are where we would plug into the stuff you are developing @gregcharles

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/PipelineScript.java
```diff
@@ -42,6 +42,10 @@ public static Builder builder() {
         return new Builder();
     }
 
+    public Builder copy() {
+        return new Builder().copy(this);
```
malaffoon: nit: this can be `return builder().copy(this);`

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/SchemaException.java
```diff
@@ -0,0 +1,7 @@
+package org.opentestsystem.rdw.admin.model;
+
+/**
+ * Placeholder for schema exceptions
+ */
+public class SchemaException {
```
malaffoon: nit: a bit queasy about calling something XxxException if it isn't actually an exception.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineScriptService.java
```diff
@@ -0,0 +1,48 @@
...
+                        "Pipeline script with ID \"%s\" does not exist",
+                        id
+                )));
+    }
+
+    @Override
+    public PipelineScript create(final User user, final PipelineScript script) {
+        return repository.create(
+                script.copy()
+                        .updatedBy(user.getUsername())
```
malaffoon: If the only reason to pass user in is to get the username, i think the signature should change to take just the username. 

Which reminds me ... where is the permission being enforced?

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/PipelineController.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @GetMapping("/{id}/publishedScripts")
+    public List<PipelineScript> getPublishedScripts(@PathVariable final int id) {
+        return service.getPublishedScripts(id);
+    }
+
+    @GetMapping("/{id}/scripts")
+    public List<PipelineScript> getScripts(@PathVariable final int id) {
+        return service.getScripts(id);
+    }
```
malaffoon: All of the scripts are published except one, right? The naming of `publishedScripts` feels odd. Perhaps it could be a query param like `GET ../pipelines/1/scripts?published=true` or something. Willing to talk it through.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/PipelineScriptController.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/pipelineScripts")
```
malaffoon: Hmm. Aren't scripts part of a pipeline? Not sure it makes sense that they have their own independent endpoint. I was thinking the endpoint prefix would be something like `pipelines/{id}/scripts`.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineScriptService.java
esotrope: we still need to add the permission to the permission service then i'll add it in a `@PreAuthorize`

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/PipelineController.java
esotrope: hm i suppose. the published scripts will be sourced from s3 while the regular scripts the db. also later if we had multiple scripts per pipeline we wouldn't really be getting published scripts we'd be getting published pipelines so maybe this should really be `/pipelines/{id}?published=true`

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/PipelineScriptController.java
esotrope: the idea here is that the script is its own entity and can be addressed directly by ID without the preceding pipeline id. I don't think we've had a crud interface for a tree shaped entity yet in the reporting - is something like that done over in ingest?

malaffoon: No existing pattern in the project to follow. Can a script exist without belonging to a pipeline or can a script belong to more than one pipeline? If the answer to both of those is no, then i think having the pipeline id in the path makes more sense.

esotrope: 👍

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineScriptService.java
```diff
@@ -1,51 +1,40 @@
...
 import org.opentestsystem.rdw.admin.model.PipelineScript;
-import org.opentestsystem.rdw.admin.model.CompilationError;
-
-import java.util.List;
+import org.springframework.security.access.prepost.PreAuthorize;
 
 /**
  * Service responsible for managing pipeline scripts
  */
+@PreAuthorize("hasPermission('PERM_PIPELINE_USER_SCRIPT_READ')")
```
esotrope: put read permission check at the top level (that makes everything require it) and write at specific method levels.
That makes sense right?

PR #1891, opened 2019-05-10
updates overall score view based on feedback

esotrope:
image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-table.component.html
+++ b/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-table.component.html
@@ -11,7 +11,6 @@
       [paginator]="true"
       [alwaysShowPaginator]="false"
       [rows]="20"
-      (onPage)="paginate($event)"
     >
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-table.component.html
```diff
@@ -11,7 +11,6 @@
       [paginator]="true"
       [alwaysShowPaginator]="false"
       [rows]="20"
-      (onPage)="paginate($event)"
```
esotrope: broke aot build

webapp/src/main/webapp/src/app/exam/component/score-table/score-table.component.less
```diff
@@ -81,3 +81,7 @@ table {
   align-items: center;
   justify-content: right;
 }
+
+.distribution-heading {
+  padding-top: 3px;
```
esotrope: so toggling between views doesn't make the table jump around

esotrope:
@malaffoon this good?

PR #1892, opened 2019-05-11
fixes rendering of pipeline history page

esotrope:


webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline-publishing-history/pipeline-publishing-history.component.ts
```diff
@@ -22,15 +23,15 @@ export class PipelinePublishingHistoryComponent {
   ) {}
 
   ngOnInit(): void {
-    const pipelineId = this.route.params.pipe(
-      map(({ id }) => Number(id)),
-      share()
```
esotrope: sharing was not caring

PR #1894, opened 2019-05-14
Feature/datasource rename

ftdholbrook:
there is a matching set of changes on rdw_config in gitlab
----
@ftdholbrook i renamed the top-level, let me know if it looks okay. I'll need to tweak gitlab too.


README.md
```diff
@@ -125,7 +125,7 @@ The integration tests dealing with Redshift have been separated out because they
 and they take a while to run. To run these tests you must set credentials -- please see the comment in 
 aggregate-service/build.gradle. By default it uses the CI database instances:
 ```bash
-(export SPRING_OLAP_DATASOURCE_PASSWORD=password; \
+(export SPRING_OLAP_RO_DATASOURCE_PASSWORD=password; \
```
malaffoon: We'll need to adjust the CI job just before merging this.

admin-service/src/main/resources/application.yml
```diff
@@ -15,21 +15,21 @@ server:
     mime-types: application/json,application/xml,text/html,text/xml,text/plain,text/css,text/javascript,application/javascript,text/csv
 
 spring:
```
malaffoon: Should this still be in `spring:` scope? I was thinking that if we changed this then the UI wouldn't have to artificially group the datasource properties.

Hmm, what if we put it under `datasource:`? Then we wouldn't even have to repeat datasource in all the names:
```
datasource:
  reporting_ro:
    ...
  warehouse_rw:
    ...
```

Feel free to ignore me, just pontificating.

ftdholbrook: I thought about 2 improvements 
1) move datasources up to the top level as you describe
2) rename the datasource beans to match the configuration
with the test/fix cycle I think it is probably 2-3 hours to get that working correctly, and decided it was low priority.  I'll open up a ticket for cleanup / improvement.
This much at least gets us some level of consistency.

ftdholbrook:
> Were you going to disable spring's automagical datasource stuff? Or do we actually need to do anything if `spring.datasource` doesn't exist?

yes.  It broke some things and I didn't want to spend any more time investigating.  I'll add a datasource consolidation / cleanup Jira to circle back around on this.

admin-service/src/test/resources/application-adminTest.yml
```diff
@@ -1,7 +1,7 @@
 spring:
-  reporting_datasource:
+  reporting_ro_datasource:
     url-schema: reporting_test
-  warehouse_datasource:
+  warehouse_rw_datasource:
```
malaffoon: oops, missed one ... how did the IT pass i wonder?

ftdholbrook: weird

PR #1895, opened 2019-05-14
tech debt: rename and document ArtifactRepository

malaffoon:


reporting-service/src/main/java/org/opentestsystem/rdw/reporting/artifact/DefaultArtifactRepository.java
```diff
@@ -9,13 +8,16 @@
 import java.io.InputStream;
 import java.text.MessageFormat;
 
+/**
+ * A resource-loader backed implementation of {@link ArtifactRepository}.
+ */
 @Repository
-public class AmazonS3ArtifactRepository implements ArtifactRepository {
+public class DefaultArtifactRepository implements ArtifactRepository {
```
malaffoon: I could see nothing really AmazonS3-specific about this class and it was just confusing the heck out of me.

PR #1898, opened 2019-05-15
Feature/create multi tenant service

vgonzalez9:
Updated to work with new API layout
Updated APIs create and update to use just the TenantOverrides for requests.
Created new support files for reporting and datasource , needed @JsonProperty to handle json using yml names that are not Java friendly. Includes property not in existing class.
Added TenantConfigurationstub class to return prebaked TenantConfigurations and TenantOverrides
Cleaned up files - updated todos for actual property retrieval from services

NOTE : integration testing is still required

TODO : 
need to retrieve yml information, atm should be on a per request basis to pick up any manual changes. 
Create & update yml files 
Need to determine what delete means and what archive means. 
remove Stub information 

UPDATE:
Modifications for PR and updated for new API json contract with UI 
Updated getAll to return the default properties with a list of tenants 
Use existing DataSource and ReportSettings classes, 
Removed multitenant package from Configuration package, moved tenant classes to Model 
Added inMemoryMap repository for tenantService to use 
Added jackson propertyNameSerializer to handle the yml setting name Cleaned up Classes, comments and import 
TODO : update with real properties - still using mock data for initial UI integration


admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/DataSource.java
```diff
@@ -0,0 +1,96 @@
+package org.opentestsystem.rdw.admin.configuration.multitenant;
+
+/**
+ * POJO for the TenantConfiguration
```
malaffoon: javadocs don't match class name?

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/DataSource.java
```diff
@@ -0,0 +1,96 @@
...
+    private String username;
+    private String password;
+    private Integer initialSize;
+    private Integer maxActive;
+
+    public String getUrl() {
+        return url;
+    }
+
+    public void setUrl(String url) {
```
malaffoon: If we have a builder do we need setters? Is this going through serialization?

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/RemoveStaleReports.java
```diff
@@ -0,0 +1,39 @@
+package org.opentestsystem.rdw.admin.configuration.multitenant;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class RemoveStaleReports {
```
malaffoon: Feel free to get rid of this, we're getting rid of the task and the settings for it.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/DataSource.java
```diff
@@ -0,0 +1,96 @@
+package org.opentestsystem.rdw.admin.configuration.multitenant;
+
```
ftdholbrook: why not use org.opentestsystem.rdw.multitenant.datasource.DataSourceElements?

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/RemoveStaleReports.java
```diff
@@ -0,0 +1,39 @@
...
+
```
ftdholbrook: I think we aren't doing this anymore

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/Reporting.java
```diff
@@ -0,0 +1,167 @@
...
+import org.opentestsystem.rdw.reporting.common.model.StudentFieldType;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Reporting {
+
+    @JsonProperty("analytics-tracking-id")
```
malaffoon: Two thoughts on this: 
1. Do we really need to change all the JSON names? 
2. I thought there was an annotation for JSON naming convention, i.e. some way of saying "snake-case everything"

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/Task.java
```diff
@@ -0,0 +1,21 @@
+package org.opentestsystem.rdw.admin.configuration.multitenant;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * A Tenant Configuration property
+ */
+public class Task {
```
malaffoon: ditto (feel free to get rid of this)

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfiguration.java
```diff
@@ -0,0 +1,228 @@
...
+    public void setLocalization(LocalizationProperties localization) {
+        this.localization = localization;
+    }
+
+
+    public static TenantConfiguration.Builder builder() {
+        return new TenantConfiguration.Builder();
+    }
+
+    public static class Builder {
```
malaffoon: It seems odd to have a builder for a `@Configuration` class ... is this for testing?

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/LocalizationProperties.java
```diff
@@ -0,0 +1,21 @@
+package org.opentestsystem.rdw.admin.configuration.multitenant;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class LocalizationProperties {
```
esotrope: why the wrapper?

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/Reporting.java
```diff
@@ -0,0 +1,167 @@
...
+import org.opentestsystem.rdw.reporting.common.model.SearchFieldPermissionLevel;
+import org.opentestsystem.rdw.reporting.common.model.StudentFieldType;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Reporting {
+
```
ftdholbrook: Is there a reason to duplicate reporting again?  if there is a good reason should we at least implement org.opentestsystem.rdw.reporting.common.configuration.ReportingSystemProperties

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfiguration.java
```diff
@@ -0,0 +1,228 @@
...
+                    },
+                    "state":{
+                       "code":"CA",
+                       "name":"California"
+                    }
+                 }
+              }
+           }
+        }
+*/
```
esotrope: did you mean to check in this comment?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantService.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Implementation of this service is responsible for management actions of a Tenant
+ * TODO Add Auth checks
+ */
+public interface TenantService {
+    /**
+     * Retrieve all known tenants and their configurations
+     *
+     */
```
malaffoon: nit: javadocs are incomplete. Personally i usually *just* document the return, something like:
```
/**
 * @return all known tenants and their configurations
 */
```

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfigurationStubs.java
```diff
@@ -0,0 +1,372 @@
...
+            "         \"common.ethnicity.NativeHawaiianOrOtherPacificIslander\":\"Native Hawaiian or Other Pacific Islander\", " +
+            "         \"common.ethnicity.DemographicRaceTwoOrMoreRaces\":\"Two or More Races\", " +
+            "         \"gender.Male\":\"Male\", " +
+            "         \"gender.Female\":\"Female\", " +
+            "         \"gender.Nonbinary\":\"Neutral\" " +
+            "      } " +
+            "   } " +
+            "}";
+
+    static public String tenantOverrideJson = "{ " +
```
esotrope: i would put these in a file then load the file. that way we don't have all this concatination and escaping going on

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfiguration.java
```diff
@@ -0,0 +1,228 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.context.annotation.Configuration;
+
+import java.util.Map;
+
+/**
+ * This contains tenant configuration - the properties and localization for a tenant
+ */
+@Configuration
+public class TenantConfiguration {
```
ftdholbrook: This looks like a duplicate  of org.opentestsystem.rdw.multitenant.TenantProperties?  I don't understand the purpose.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
```diff
@@ -0,0 +1,43 @@
...
+@RestController
+public class TenantController {
+
+    private TenantService tenantService;
+
+    public TenantController(TenantService tenantService) {
+        this.tenantService = tenantService;
+    }
+
+    @GetMapping(path = "/tenants/getall", produces = MediaType.APPLICATION_JSON_VALUE)
```
esotrope: i think json is the default

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
```diff
@@ -0,0 +1,43 @@
...
+import org.opentestsystem.rdw.admin.configuration.multitenant.TenantOverride;
+import org.opentestsystem.rdw.admin.service.TenantService;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
+
+@RestController
+public class TenantController {
```
esotrope: if you add `@RequestMapping("/tenants")` at the top you can omit it in each of the handlers below

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantOverride.java
```diff
@@ -0,0 +1,39 @@
+package org.opentestsystem.rdw.admin.configuration.multitenant;
+
+import org.opentestsystem.rdw.multitenant.TenantProperties;
+
+/**
+ * A TenantConfiguration property used to maintain Tenants
+ */
+public class TenantOverride {
```
malaffoon: Why is this a TenantOverride? It feels like "override" is a use-case not an object name. I suck at naming things but maybe i can help if you explain what this is exactly?

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
esotrope: also the endpoint should be `/tenants` to be restful

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
```diff
@@ -0,0 +1,43 @@
...
+    public TenantController(TenantService tenantService) {
+        this.tenantService = tenantService;
+    }
+
+    @GetMapping(path = "/tenants/getall", produces = MediaType.APPLICATION_JSON_VALUE)
+    List<TenantConfiguration> getAllTenants() {
+        return tenantService.getAll();
+    }
+
+    @PostMapping(path = "/tenants/create", produces = MediaType.APPLICATION_JSON_VALUE)
```
esotrope: same comment. the rest interface should look like this:
```
GET /tenants (gets all tenants)
POST /tenants (creates a tenant)
PUT /tenants (updates a tenant)
DELETE /tenants/{id} (deletes a tenant for the given id)
```

admin-service/src/test/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfigurationTest.java
```diff
@@ -0,0 +1,229 @@
...
+//        serviceProperties.forEach((k,v) -> {
+//            logger.info("services={} and value = {}", k, v);
+//            System.out.println("Key : " + k + " Value : " + v);
+//
+//        });
+
+    }
+
+    private TenantOverride getTestTenantOverride(String testTenantOverride) {
+        Gson gson = new GsonBuilder().serializeNulls().create();
```
esotrope: still using this? we probably don't need to use gson anywhere when we have jackson

admin-service/src/test/java/org/opentestsystem/rdw/admin/web/TenantControllerTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+import static org.mockito.ArgumentMatchers.any;
+
+import static org.mockito.Mockito.*;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.*;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@WebMvcTest(controllers = {TenantController.class}, secure = false)
+public class TenantControllerTest {
```
esotrope: 👍 does the secure=false thing make it to where we need not specify a user in the request?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantService.java
```diff
@@ -0,0 +1,131 @@
...
+@Service
+public class DefaultTenantService implements TenantService {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultTenantService.class);
+
+    private TenantProperties tenantProperties;
+    // used to store the default information. But atm will be the initial Mock Tenant with
+    // no tenantOverride information.
+    private TenantConfiguration defaultTenantConfiguration;
+    // Todo replace this initial in memory storage
+    private HashMap<String, TenantConfiguration> tenantConfigurationMap = new HashMap<>();
```
malaffoon: What i like to do for this is create a TenantRepository which the service uses, then provide an InMemoryTenantRepository until i get around to creating the real one. That way, the service won't have to change at all, only the repo.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantService.java
```diff
@@ -0,0 +1,131 @@
...
+
+    // todo use TenantProperties tenantProperties & reportingSettings
+    @Autowired
+    DefaultTenantService(TenantProperties tenantProperties) {
+        tenantProperties = new TenantProperties(); // temp
```
malaffoon: Likewise, i like to code the service as if i have everything and just use a temporary bean to inject the temp stuff.

I admit that these are conventions, just pushing the mock/temp code around a bit. Just helps me think through where to put logic.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
```diff
@@ -0,0 +1,43 @@
+package org.opentestsystem.rdw.admin.web;
+
+import org.opentestsystem.rdw.admin.configuration.multitenant.TenantConfiguration;
+import org.opentestsystem.rdw.admin.configuration.multitenant.TenantOverride;
+import org.opentestsystem.rdw.admin.service.TenantService;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.*;
```
malaffoon: Can you adjust your IDE settings to avoid import splats? Thanks.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
malaffoon: This should just be `path = "/tenants"`

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
```diff
@@ -0,0 +1,43 @@
...
+public class TenantController {
+
+    private TenantService tenantService;
+
+    public TenantController(TenantService tenantService) {
+        this.tenantService = tenantService;
+    }
+
+    @GetMapping(path = "/tenants/getall", produces = MediaType.APPLICATION_JSON_VALUE)
+    List<TenantConfiguration> getAllTenants() {
```
malaffoon: Is order significant? If it isn't perhaps this is just a Collection?

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
malaffoon: `path = "/tenants"`

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
```diff
@@ -0,0 +1,43 @@
...
+        return tenantService.getAll();
+    }
+
+    @PostMapping(path = "/tenants/create", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(value = HttpStatus.CREATED)
+    TenantConfiguration createTenant(@RequestBody final TenantOverride tenantOverride) {
+        return tenantService.createTenant(tenantOverride);
+    }
+
+    @PutMapping(path = "/tenants/save", produces = MediaType.APPLICATION_JSON_VALUE)
```
malaffoon: `path = "/tenants"`

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
```diff
@@ -0,0 +1,43 @@
...
+    @ResponseStatus(value = HttpStatus.OK)
+    TenantConfiguration saveTenant(@RequestBody final TenantOverride tenantOverride) {
+        return tenantService.updateTenant(tenantOverride);
+    }
+
+    @DeleteMapping("tenants/delete")
```
malaffoon: `"tenants/{tenantId}"`

admin-service/src/test/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfigurationTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+        ConfigurationProperties configurationProperties = tenantOverride.getConfigurationProperties();
+        assertThat(configurationProperties).isNotNull();
+        Task task = mapper.convertValue(configurationProperties.getTask(), Task.class);
+        assertThat(task).isNotNull();
+        logger.info("Cfg Prop's Task = ", task.getRemoveStaleReports());
+
+    }
+
+    //@Test
```
malaffoon: Perhaps leave the `@Test` annotation and add an `@Ignore("ignoring because ...")`

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
ftdholbrook: > `"tenants/{tenantId}"`

actually `tenants/{tenantKey}` would be preferable

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantServiceTest.java
```diff
@@ -0,0 +1,158 @@
...
+        service.addTenantConfiguration(tenantConfiguration);
+
+        TenantConfiguration tenantConfigurationNV = TenantConfigurationStubs.createTenantWithMockDefaultsNV();
+        tenantConfiguration.setTenantOverride(TenantConfigurationStubs.getTenantOverrideNV());
+        service.addTenantConfiguration(tenantConfigurationNV);
+
+        List<TenantConfiguration> tenantList = service.getAll();
+        verifyTenantList(tenantList, 2);
+
+        // visual dump of tenants
```
malaffoon: To be replaced with assertions right?  :)

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantServiceTest.java
```diff
@@ -0,0 +1,158 @@
...
+        verifyTenantList(tenantList, 1, tenantConfiguration.getTenantOverride());
+
+        // now ok to delete
+        service.deleteTenant(tenantConfiguration.getTenant().getId());
+
+        // verify tenant was removed from list
+        tenantList.clear();
+        tenantList = service.getAll();
+        assertThat(tenantList).isNotNull();
+        assertThat(tenantList).isEmpty();
```
malaffoon: fyi, you can chain assertions so these 4 lines could be replaced with 
```
    assertThat(service.getAll()).isNotNull().isEmpty();
```

malaffoon: Actually you shouldn't need to check `isNotNull()` - services should never return null for collections.

admin-service/src/test/java/org/opentestsystem/rdw/admin/web/TenantControllerTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.test.web.servlet.MockMvc;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+
+import static org.mockito.ArgumentMatchers.any;
+
+import static org.mockito.Mockito.*;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.*;
```
malaffoon: ditto (about import splats)

admin-service/src/test/java/org/opentestsystem/rdw/admin/web/TenantControllerTest.java
```diff
@@ -0,0 +1,114 @@
...
+                TenantConfigurationStubs.createTenantWithMockDefaultsNV());
+        when(service.getAll()).thenReturn(tenantConfigurationList);
+
+        mvc.perform(
+                get("/tenants/getall")
+                .contentType(MediaType.APPLICATION_JSON)
+                .accept(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk());
+
+        verify(service, times(1)).getAll();
```
malaffoon: nit: `times(1)` is the default, don't need to specify it.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
malaffoon: sure, i was just matching the method declaration

admin-service/src/test/java/org/opentestsystem/rdw/admin/web/TenantControllerTest.java
esotrope: cool i didn't know that one

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/DataSource.java
vgonzalez9: After review other code I saw that appeared to be the preference so I has added.
And yes since it is part of the TentantOverride object that is automagically getting serialized when the API is called.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/Reporting.java
vgonzalez9: without them it was failing, but I'll look into a generic way to handle them.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfiguration.java
vgonzalez9: Yeah it was then, I changed what I was doing. Will remove.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/LocalizationProperties.java
vgonzalez9: Its not need, I can use the map directly here.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfiguration.java
vgonzalez9: Though it could be helpful for future reference. But it does take up a ton of space.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantService.java
vgonzalez9: I'll go update those accordingly.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantController.java
vgonzalez9: would that affect the /tenants default route that displays te default page?

esotrope: nope : j - magical, i know

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantOverride.java
vgonzalez9: The TenantConfiguration obj contains the default properties and the TenantOverride contains tenant specific properties. Changes only occur in the TenantOverride, none of the default properties are ever modified.  So only the TenantOverride is sent to the APIs, for create & update, and each return a full TenantConfiguration so the UI will have a tenant's default and modified properties. 
A better description would have help.

admin-service/src/test/java/org/opentestsystem/rdw/admin/web/TenantControllerTest.java
vgonzalez9: yes, without it an Auth error is thrown

admin-service/src/test/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfigurationTest.java
vgonzalez9: No we can remove.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/ConfigurationProperties.java
```diff
@@ -0,0 +1,46 @@
...
+    private Task task;
+
+
+    public ConfigurationProperties () {}
+
+    public Reporting getReporting() {
+        return reporting;
+    }
+
+    public void setReporting(Reporting reporting) {
```
burzum619: Sort of a nit, but in general I'd add the `final` keyword to all method arguments with immutable inputs. There's a way to configure intellij to do this automatically for generated getters/setters

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/Reporting.java
burzum619: https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/PropertyNamingStrategy.html#KEBAB_CASE

using this name strategy (`KEBAB_CASE`) should remove the need for all the `@JsonProperty` annotations

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/DataSource.java
vgonzalez9: will fix

admin-service/src/test/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfigurationTest.java
```diff
@@ -0,0 +1,229 @@
...
+        ObjectMapper mapper = new ObjectMapper();
+
+        TenantConfiguration tenantConfiguration = getTestTenantConfig(tenantConfigurationJson);
+        assertThat(tenantConfiguration).isNotNull();
+        logger.info("tenantConfiguration = {}", tenantConfiguration);
+        assertThat(tenantConfiguration.getTenant()).isNotNull();
+        logger.info("tenantConfiguration.getTenant() = {}", tenantConfiguration.getTenant());
+        logger.info("tenantConfiguration.getTenant().getId() = {}", tenantConfiguration.getTenant().getId());
+        assertThat(tenantConfiguration.getConfigurationProperties()).isNotNull();
+        logger.info("tenantConfiguration.getTenant().getId() = {}", tenantConfiguration.getTenant().getId());
```
burzum619: Personally I prefer not to have log statements in test code (except in rare cases). I haven't taken a good look of RDW backend tests though, so if this is the norm ignore this comment.

admin-service/src/test/java/org/opentestsystem/rdw/admin/web/TenantControllerTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Test
+    public void itShouldCreateTenant() throws Exception {
```
burzum619: good test coverage

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/DataSource.java
malaffoon: In general POJO's are either built internally OR serialized so they seldom have both. But this may be the case where we need to do both.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantOverride.java
malaffoon: Maybe we can whiteboard the container classes. This POJO still feels like "tenant configuration", the override part is just how it is used. And I haven't tracked back on Dave's comment that this feels duplicative with TenantProperties somehow.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/Reporting.java
malaffoon: I was thinking about this last night ... it seems wrong to have to create a POJO for all this. Aren't we going to get the properties by reading the configuration into generic collections? We'll need the high level wrapper, but it seems like the properties themselves shouldn't be hard-coded anywhere?

admin-service/src/test/java/org/opentestsystem/rdw/admin/configuration/multitenant/TenantConfigurationTest.java
malaffoon: +42.  I was assuming this was here an interim fiddling and would be converted to assertions.

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/multitenant/ConfigurationProperties.java
```diff
@@ -0,0 +1,46 @@
+package org.opentestsystem.rdw.admin.configuration.multitenant;
```
ftdholbrook: I don't think these beans belong here in the `org.opentestsystem.rdw.admin.configuration` package.  that package seems to be for spring configuration objects.  I think `org.opentestsystem.rdw.admin.model` is more appropriate.

ftdholbrook:
I think we need to figure out some naming conventions for the structure:

maybe something like:
Now:

TenantOverride
	TenantProperties.Tenant tenant
	ConfigurationProperties configurationProperties
	LocalizationProperties localization
	
Proposed Naming:

TenantConfiguration
	TenantProperties.Tenant tenant
	ApplicationConfiguration configuration
	LocalizationProperties localization

ftdholbrook:
Is the admin service a directly user accessible public API?  If not this will need to move to webapp.

malaffoon:
@ftdholbrook , all the reporting services are proxied (zuul) through the webapp.

PR #1902, opened 2019-05-16
Remove the stale report cleanup task

malaffoon:
There were some helper methods that existed only for the task, so i removed those.
Cleaned up some warnings.
Fixed an unreported bug.

----
There are some documentation changes that go along with this ... i'll submit a separate PR.


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/DefaultExamReportService.java
```diff
@@ -87,7 +89,12 @@ public UserReport createReport(
 
     @Override
     public void delete(final User user, final long id) {
+        final Optional<UserReport> found = service.findOneById(id);
+        if (!found.isPresent()) return;
+
+        // even though this is a find/delete, it doesn't really need to be transactional
         service.delete(user.getUsername(), id);
+        contentService.deleteContent(found.get());
```
malaffoon: This was actually a bug: we were wiping out the database record for the report but not deleting the content out on S3.

esotrope:
so let me get this straight - you removed the stale stale report remover?
Also, what will this do to pagination on the my reports page? are we going to crash peoples browsers?

PR #1904, opened 2019-05-17
RP-304 - Integration adjustments to fix top level names

vgonzalez9:
Update the datasource and archive names per Json contract with UI


vgonzalez9:
Added @JsonProperty annotation for UI integration

PR #1905, opened 2019-05-18
Feature/create multi tenant service

vgonzalez9:
Add mock tenants 
added clearTenant to repo for testing


vgonzalez9:
Add 2 mock tenants for default integration

vgonzalez9:
Add 2 mock tenants for default integration

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantServiceTest.java
```diff
@@ -39,6 +39,8 @@ public void setUp() {
         //creating a new repo so the cache is empty at the start of the tests
         inMemoryTenantRepository = new InMemoryTenantRepository();
         service = new DefaultTenantService(tenantProperties, inMemoryTenantRepository);
+        // todo remove once mock is gone
+        inMemoryTenantRepository.clearTenants();  // atm service is automatically populating Mock data
```
esotrope: in the service test we should be mocking the repository layer i think so i'm not sure if this is necessary. this would imply that the service test is dependent on the state of the repo - but as said that should be mocked.
There should be some examples using `@Mock` and `MockitoTestRunner`

PR #1907, opened 2019-05-20
gist to discuss multitenant roles

malaffoon:


common/src/main/java/org/opentestsystem/rdw/reporting/common/multitenant/ReportingTenantIdResolver.java
```diff
@@ -9,8 +9,6 @@
  */
 public class ReportingTenantIdResolver implements TenantIdResolver {
 
-    public static final String NO_TENANT_ID = "NO_TENANT_ID";
```
malaffoon: Unrelated to the main gist of this PR: do we need to duplicate the NO_TENANT_ID constant?

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/ReportingPermission.java
```diff
@@ -9,5 +12,20 @@
...
+    public static final String EmbargoAdminRole = "EMBARGO_ADMIN";
+    public static final String GroupAdminRole = "GROUP_ADMIN";
+    public static final String InstructionalResourceAdminRole = "Instructional Resource Admin";
+    public static final String PipelineAdminRole = "PIPELINE_ADMIN";
+    public static final String TenantAdminRole = "TENANT_ADMIN";
+
+    public static final List<String> ReportingRoles = Arrays.asList(
+            GeneralRole, PiiRole, PiiGroupRole, CustomAggregateReporterRole, EmbargoAdminRole,
+            GroupAdminRole, InstructionalResourceAdminRole, PipelineAdminRole, TenantAdminRole
+    );
```
malaffoon: The current list of reporting roles.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -68,12 +65,11 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
                     .build();
         }
 
-        //TODO: Need list of applicable roles
```
malaffoon: Thinking about this: what if a customer decides to create a whole new set of roles, mapping them to the reporting permissions? For example, maybe they actually create roles that make sense, like "Teacher", "School Admin", etc. Since everything else in the reporting app is permission-based, hardcoding the roles seems like a bad idea. I guess we could have a configurable list of roles to look for?

Anyway, i wanted to put this PR up to force the discussion.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -68,12 +65,11 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
...
         final String stateId = grants.stream()
                 .filter(g -> g.getRole() != null && g.getStateId() != null)
                 .findFirst()
```
malaffoon: If we go this route, i'd add a `.filter(g -> ReportingRoles.contains(g.getRole())`

common/src/main/java/org/opentestsystem/rdw/reporting/common/multitenant/ReportingTenantIdResolver.java
ftdholbrook: We could push it up to TenantIdResover in common.  

I don't think it is used much anymore, most application code is using the TenantKey construct now, and there are some situations where we want to have a "NO_TENANT_KEY" (like cache keys in use prior to login).  It's worth looking into removing completely.

malaffoon: It is already in common TenantIdResolver, that's what confused me.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
ftdholbrook: It makes sense to me to have a instance wide configuration that lists the roles.

common/src/main/java/org/opentestsystem/rdw/reporting/common/multitenant/ReportingTenantIdResolver.java
ftdholbrook: Then we can defiantly pull it :-)

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
malaffoon: Ok. I'm going to be lazy and put a TODO on this. So really all i've done is moved the TODO from one place to another  ;)

webapp/src/test/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsServiceTest.java
```diff
@@ -85,7 +85,7 @@ public void itShouldUseMailAndItIsOkayIfNameIdIsWrong() {
 
     @Test
     public void itShouldAssignTenant() {
-        final String chain = "|CA|ASMTDATALOAD|STATE|SBAC||||CA||||||||||";
+        final String chain = "|CA|PII|STATE|SBAC||||CA||||||||||";
```
malaffoon: proof that the change did something

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
esotrope: think i'm lacking context on this - why do we need to enumerate the roles in the app?

malaffoon: @esotrope: We need to determine a user's tenant from their tenancy chain. However, a tenancy chain may contain grants from multiple tenants so we agreed with Smarter Balanced to use the first tenant we encounter. However, there are users that have Digital Library grants from multiple tenants but RDW grants from a single tenant. So we need to extract just the grants that are intended for RDW. In theory, we could punt on this and then blame SB when the tenancy chain has multiple tenants but i wanted to make a reasonable effort to do the Right Thing.

esotrope: ah thanks for the nuance - could we not also add a new permission "REPORTING_READ" to all the RDW roles?

PR #1912, opened 2019-05-21
Defect/revert tenant name property

vgonzalez9:
Removing the Jackson PropertyName from the ApplicationConfiguration. 
Other services are depending on Camelcase, not kebab. Remove for now, will need to address TenantService and TenantUI interface.


admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Service
+public class DefaultTenantPropertiesServiceTest {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultTenantPropertiesServiceTest.class);
+
+    //ReportingServiceTenantProperties reportingServiceTenantProperties;
+    String uri = "http://localhost:4200/api/settings";
+    
+    String reportSettings ="{" +
```
esotrope: i think you can put this in a file under `/resources` and do `return IOUtils.readFully(getClass().getClassLoader().getResourceAsStream("filename"), "UTF-8")
just did something similar recently

vgonzalez9:
Ugh - only one was suppose to be pushed

vgonzalez9:
Closing this pull request so only the Needed change is merged.

PR #1914, opened 2019-05-22
Feature/git client

ftdholbrook:
Not 100% ready, but very close

Still TODO:
- externalize the confiuration properties 
- set up credentials on our dev gitlab rdw_config, encrypt the password and put it here
- a few more tests
- some documentation

Interested in early feedback @malaffoon, @vgonzalez9, @burzum619


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
```diff
@@ -0,0 +1,136 @@
...
+        File directory = new File(gitClientProperties.getLocalRepositoryPath());
+        try {
+            FileUtils.forceMkdir(directory);
+        } catch (IOException e) {
+            throw new GitClientException("Unable to create working directory.", e);
+        }
+        return directory;
+    }
+
+    public void saveTenantConfiguration(String tenantKey, ApplicationTenantConfiguration applicationTenantConfiguration) throws IOException, GitAPIException {
```
ftdholbrook: need to wrap these exceptions in the custom GitClientException

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
```diff
@@ -0,0 +1,136 @@
...
+        getLocalRepositoryToCleanState();
```
ftdholbrook: This saveTenantConfiguartion is the only public method, it really should only be called one-at-a-time... I could implement a locking mechanism, but maybe having this driving off of a queue is the best way to do that.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
```diff
@@ -0,0 +1,136 @@
...
+import org.yaml.snakeyaml.nodes.Node;
+import org.yaml.snakeyaml.nodes.NodeTuple;
+import org.yaml.snakeyaml.nodes.SequenceNode;
+import org.yaml.snakeyaml.nodes.Tag;
+import org.yaml.snakeyaml.representer.Representer;
+
+import java.io.File;
+import java.io.IOException;
+
+public class GitClient {
```
malaffoon: nit: can/should we have an interface to facilitate unit testing, etc.?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClientException.java
```diff
@@ -0,0 +1,20 @@
+package org.opentestsystem.rdw.admin.multitenant.git;
+
+public class GitClientException extends RuntimeException {
```
malaffoon: I'll be curious to see how this is used: it doesn't add anything to the base exception so it is effectively a "marker" exception. Will it be handled differently than other runtime exceptions?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
malaffoon: I never looked closely but it's interesting that ApplicationTenantConfiguration doesn't include the tenant key.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
```diff
@@ -0,0 +1,136 @@
...
+        FileUtils.forceMkdir(tenantDirectory);
+
+        File tenantFile = new File(tenantDirectory.getPath() + File.separator + "application.yml");
+        logger.debug("tenantFile {}", tenantFile.getPath());
+        FileUtils.write(tenantFile, output, "UTF-8");
+
+        logger.debug("read file", FileUtils.readFileToString(tenantFile));
+
+        Git git = Git.open(new File(gitClientProperties.getLocalRepositoryPath()));
+        git.add().addFilepattern(".").call();
```
malaffoon: should this be more specific since we know the filename?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
```diff
@@ -0,0 +1,136 @@
...
+        options.setIndent(2);
+        options.setCanonical(false);
+
+
+        SkipEmptyRepresenter representer = new SkipEmptyRepresenter();
+        representer.addClassTag(ApplicationTenantConfiguration.class,Tag.MAP);
+
+        Yaml yaml = new Yaml(representer, options);
+
+        String output = yaml.dump(applicationTenantConfiguration);
```
malaffoon: nit: it feels like lines 66-77 should be in a helper or something (everything feels blended in this method)

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClientException.java
ftdholbrook: It wraps the non-runtime exceptions in run-time executions.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
ftdholbrook: it is indirectly in there, not worth digging out

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClient.java
ftdholbrook: yes

admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
```diff
@@ -90,5 +92,16 @@ public OrganizationRepository warehouseOrganizationRepository(@Qualifier("wareho
         return new WarehouseJdbcOrganizationRepository(template, securityParameterProvider);
     }
 
+    @Bean
+    @ConfigurationProperties("tenant-configuration-git-client")
```
malaffoon: nit: that's a really long property name; are we going to have other stuff so we can break it up perhaps?

meh, i'm probably just wrong but i have to type those things in a lot

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClientException.java
malaffoon: Yeah, i got that part, just wondering why we can't just use `RuntimeException`?

admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
ftdholbrook: I can change it to whatever, I'm terrible at naming things

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitClientException.java
ftdholbrook: Style points?  Pretty much the way I've always done it.  When I'm searching logstash for a specific error it can be handy.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationGitClient.java
```diff
@@ -0,0 +1,136 @@
...
+import org.yaml.snakeyaml.nodes.Node;
+import org.yaml.snakeyaml.nodes.NodeTuple;
+import org.yaml.snakeyaml.nodes.SequenceNode;
+import org.yaml.snakeyaml.nodes.Tag;
+import org.yaml.snakeyaml.representer.Representer;
+
+import java.io.File;
+import java.io.IOException;
+
+public class TenantConfigurationGitClient {
```
malaffoon: i'd be tempted to make this something like:
```
public class GitTenantConfigurationRepository implements TenantConfigurationRepo {
```

malaffoon: or is that coming later and will use this?

ftdholbrook: "Repository" is a super overloaded term in spring land... From a DDD perspective it is, but I'm kind of loath to put it in the same category as the RDBMS interactions.

malaffoon: `Service`?

ftdholbrook: there is already org.opentestsystem.rdw.admin.repository.TenantConfigurationRepository that is @vgonzalez9 put in.  I think this thing is closer to a datasource than to a repository, it will only save the data all the reading is coming from elsewhere.

malaffoon: Oh, i thought this was going to be part of the real TenantConfigurationRepository ... the exact one that Vicky created.

ftdholbrook: This will be called by that for add and update (both will call save)

ftdholbrook:
I did a rename pass
TenantConfigurationPersistenceService (interface)
TenantConfigurationPersistenceServiceGit (impl)

config root:
 tenant-configuration-persistence:

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGitIT.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+// we don't currently have a git integration test server setup
+// to run this test set tenant-configuration-persistence.enable=true
+// also need to set remote repository uri, username, and password
+//  java -Dtenant-configuration-persistence.enable=true \
+//       -Dtenant-configuration-persistence.remote-repository-uri=https://github.com/example/rdw_config.git
+//       -Dtenant-configuration-persistence.git-username=exampleuser
+//       -Dtenant-configuration-persistence.git-password=examplepwd
+@IfProfileValue(name = "tenant-configuration-persistence.enable", value = "true")
```
malaffoon: remind me to set up CI to run this as part of the build

PR #1915, opened 2019-05-22
fix interim administration conditions

malaffoon:


malaffoon:
I'm going to merge this ... 'cause it is an obvious bug.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ReportOptions.java
```diff
@@ -128,8 +128,8 @@ public Builder subjects(final List<SubjectAssessment> subjects) {
             return this;
         }
 
-        public Builder interimAdministrationConditions(final List<String> summativeAdministrationConditions) {
-            this.summativeAdministrationConditions = summativeAdministrationConditions;
+        public Builder interimAdministrationConditions(final List<String> interimAdministrationConditions) {
+            this.interimAdministrationConditions = interimAdministrationConditions;
```
esotrope: uh oh. that looks disastrous

PR #1916, opened 2019-05-22
fix js date sort

malaffoon:


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -476,7 +476,10 @@ export class AssessmentResultsComponent implements OnInit {
     this.showPercentileHistory = !this.showPercentileHistory;
     if (Utils.isNullOrUndefined(this.percentileGroups)) {
       const results = this.assessmentExam;
-      const dates = results.exams.map(exam => new Date(exam.date)).sort();
```
malaffoon: even i know you can't trust the default js sort  👎

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -476,7 +476,10 @@ export class AssessmentResultsComponent implements OnInit {
...
+      // sort dates ascending
+      const dates = results.exams
+        .map(exam => exam.date)
+        .sort((a, b) => (a > b ? 1 : a < b ? -1 : 0));
```
malaffoon: no idea if there is a more elegant solution available

esotrope: you could be slightly more fancy and do this
```
const dates = results.exams
        .map(exam => exam.date)
        .sort((a, b) => a - b);
```

PR #1920, opened 2019-05-23
Feature/rp 367 add support for default and tenant properties

vgonzalez9:
Lots of TODOs for archive and aggregate properties
NOTE TenantUpdate will throw NoSuchElementExceptions 404 if the key is not found 
           TenantCreate will throw a IllegalArgumentException 400 if the key exists already
FIX: for add/delete/add - it was using the tenantProperties to verify if a tenant exists, but that list wasn't being updated when tenants are being added and delete 
UPDATE: 
Refactored the service name, DefaultActuatorClientRdwService & ActuatorClientRdwDefaultPropertiesService
and the various PR comments, also added some additional testing, found that the repo was not returning the the newly added or updated tenantConfiguration, it is now. 
Keeping a


admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
```diff
@@ -90,5 +92,17 @@ public OrganizationRepository warehouseOrganizationRepository(@Qualifier("wareho
         return new WarehouseJdbcOrganizationRepository(template, securityParameterProvider);
     }
 
+    @Bean
+    @ConfigurationProperties("tenant-configuration-lookup")
```
malaffoon: This is a second set of things that seem to belong under "tenant configuration", can we put them under the same top-level property?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfiguration.java
```diff
@@ -122,8 +122,8 @@ public void setLocalization(Map<String, String> localization) {
             "sum",
             "iab"
          ],
-         "statewide-user-assessment-types":"sum",
-         "state-aggregate-assessment-types":"sum"
+         "statewide-user-assessment-types":["sum"],
+         "state-aggregate-assessment-types":["sum"]
```
malaffoon: do we have to make this change?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantServiceDefaultPropertiesService.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.opentestsystem.rdw.admin.model.ApplicationTenantConfiguration;
+import org.opentestsystem.rdw.archive.ArchivePropertiesTenant;
+import org.opentestsystem.rdw.multitenant.datasource.DataSourceElementsTenant;
+import org.opentestsystem.rdw.olap.configuration.AggregateReportingPropertiesTenant;
+import org.opentestsystem.rdw.reporting.common.configuration.ReportingSystemPropertiesImpl;
+
+import java.util.Optional;
+
+public interface TenantServiceDefaultPropertiesService {
```
malaffoon: i know naming is problematic but i have to ask why the second `Service` in the name? Can/should this be `TenantDefaultPropertiesService`?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantServiceDefaultPropertiesService.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Retrieve the Default settings ApplicationTenantConfiguration
+     * @return ApplicationTenantConfiguration
```
malaffoon: nit (style): if the javadocs don't add value, just document the return, e.g.
```
    /**
     * @return the default settings ApplicationTenantConfiguration
     *
     */
```

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceService.java
```diff
@@ -0,0 +1,256 @@
...
+import org.springframework.stereotype.Service;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+
+@Service
+public class DefaultTenantPropertiesServiceService implements TenantServiceDefaultPropertiesService {
```
malaffoon: Okay, now i'm really confused with the naming ... i get that we are getting service properties and this is a service that does that but ... still confused. Maybe we can discuss naming.

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupProperties.java
```diff
@@ -0,0 +1,18 @@
+package org.opentestsystem.rdw.admin.client;
+
+import java.util.Map;
+
+/**
+ * This contains the tenant services and their corresponding endpoints
+ */
+public class TenantConfigurationLookupProperties {
+    private Map<String, RdwService> services;
```
malaffoon: Are we likely to need more info about the apps, could this just be Map<String, String> mapping name to url?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupProperties.java
```diff
@@ -0,0 +1,18 @@
+package org.opentestsystem.rdw.admin.client;
+
+import java.util.Map;
+
+/**
+ * This contains the tenant services and their corresponding endpoints
```
malaffoon: I think my confusion about names starts here. The services aren't "tenant services", they are "RDW services", perhaps we could use that (this will make more sense when you read my later comments).

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantServiceDefaultPropertiesService.java
malaffoon: Okay, having  read a bit perhaps this could be `RdwServiceDefaultPropertiesService`

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceService.java
malaffoon: Having read more, perhaps `ActuatorClientRdwServiceDefaultPropertiesService`?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceService.java
```diff
@@ -0,0 +1,256 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(DefaultTenantPropertiesServiceService.class);
+
+    private TenantConfigurationLookupProperties tenantConfigurationLookupProperties;
+
+    private TenantConfigurationLookupClient tenantConfigurationLookupClient;
+
+    private ObjectMapper mapper = new ObjectMapper();
```
malaffoon: I know we got in trouble before reusing the global mapper but it seems like could reuse it here? Just add that deserialization feature in the application.yml?

admin-service/src/main/resources/application.yml
```diff
@@ -126,3 +126,15 @@ reporting:
...
+tenant-configuration-lookup:
+  services:
+    admin-service:
+      url: http://admin-service:8008
+    aggregate-service:
+      url: http://aggregate-service:8008
+    report-processor:
+      url: http://report-processor-service:8008
+    reporting-service:
+      url: http://reporting-service:8008
```
malaffoon: per question above, could this just be:
```
  services:
    admin-service: http://admin-service:8008
    aggregate-service: http://aggregate-service:8008
    ...
```
Not a big deal, just curious.

admin-service/src/test/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClientTest.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.CoreMatchers.is;
+
+//this test only runs against a local docker instance
+// set this property to run this test (or temporarily comment out this annotation)
+// java -Dtenant-client.docker-it-enable=true
+@IfProfileValue(name = "tenant-client.docker-it-enable", value = "true")
```
malaffoon: please put a TODO here ... we have to get rid of this restriction

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceService.java
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Returns the ApplicationTenantConfiguration default reading DataSourceElementsTenant
+     * @param responseEntities from the service calls
+     * @return DataSourceElementsTenant
+     */
+    public Optional<DataSourceElementsTenant> getDataSourceElementsTenant(List<ResponseEntity<String>> responseEntities) {
+
+        Optional<DataSourceElementsProperties> dataSourceElementsTenantProperties = tenantConfigurationLookupClient.getDataSources(responseEntities);
+        if(dataSourceElementsTenantProperties.isPresent()) {
+            logger.info("testing response to get for ReportingSystemProperties = {}", dataSourceElementsTenantProperties.get());
```
ftdholbrook: minor: probably should drop this to debug

admin-service/src/test/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClientTest.java
```diff
@@ -0,0 +1,173 @@
...
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@ActiveProfiles({"tenant_client_docker"})
+@EnableConfigurationProperties
+public class TenantConfigurationLookupClientTest {
```
ftdholbrook: It looks to me like this is an End to End test that requires local docker to be up and running.  I think this will fail in the CI environment.

admin-service/src/test/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClientTest.java
ftdholbrook: Created JIRA https://jira.fairwaytech.com/browse/RP-394 to track this and other ETE tests

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantServiceDefaultPropertiesService.java
esotrope: so can we not write them when what the method does is obvious ? : J

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceService.java
esotrope: We need a service to get the service : J

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceService.java
esotrope: don't forget about `ObjectMapper.copy()` you can autowire the global one in, copy and modify from there

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfiguration.java
vgonzalez9: I updated to match the AggregateReportingProperties which are all Sets.  Is there only ever one type assigned at a time?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantServiceDefaultPropertiesService.java
vgonzalez9: Renaming sounds find to me.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantPropertiesServiceService.java
vgonzalez9: How about ActuatorClientRdwDefaultPropertiesService? since it gathering from multiple rdw services?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfiguration.java
vgonzalez9: Left as a Set for now. But if we do, the UI will need to handle a list of type if the user enters multiple types.

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupProperties.java
vgonzalez9: The rdwService allow for a more obvious mapping of the url, and I wasn't sure if that could be used to possibly reference a yml file or path updates.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantServiceDefaultPropertiesService.java
vgonzalez9: Updated to DefaultActuatorClientRdwService implements ActuatorClientRdwDefaultPropertiesService

admin-service/src/test/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClientTest.java
vgonzalez9: Updated with @Ignores, would like to keep for future  local testing

ftdholbrook: I added this ticket to track this test and some others, feel free to update https://jira.fairwaytech.com/browse/RP-394

vgonzalez9:
These changes were included in the RP366
closing this request.

PR #1921, opened 2019-05-23
Feature/rp 390 tenant integration

burzum619:
Includes some refactoring. I pulled the service call for the default localization values out of the child component into the main `tenant.component.ts` page component to avoid redundant calls when more than one tenant is configured.

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
@@ -65,6 +65,7 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
   expandOrCollapse(node: TreeNode): void {
     node.expanded = !node.expanded;
     // Change detection is not triggered unless the TreeNode array is replaced
+    // TODO: See if we can avoid this hack
     this.configurationPropertiesTreeNodes = cloneDeep(
       this.configurationPropertiesTreeNodes
     );
--- a/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
@@ -9,8 +9,20 @@ import { forOwn, cloneDeep, get } from 'lodash';
   templateUrl: './property-override-tree-table.component.html'
 })
 export class PropertyOverrideTreeTableComponent implements OnInit {
+  _configurationProperties: any;
+
+  get configurationProperties(): any {
+    return this._configurationProperties;
+  }
+
   @Input()
-  configurationProperties: any;
+  set configurationProperties(properties: any) {
+    this._configurationProperties = properties;
+    if (this.form) {
+      this.createConfigurationPropertyTree();
+    }
+  }
+
   @Input()
   propertiesArrayName: string;
   @Input()
@@ -18,13 +30,11 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
 
   showModifiedPropertiesOnly = false;
   configurationPropertiesTreeNodes: TreeNode[] = [];
-  filteredConfigurationPropertiesTreeNodes: TreeNode[] = [];
 
   constructor() {}
 
   ngOnInit(): void {
     this.createConfigurationPropertyTree();
-    this.filteredConfigurationPropertiesTreeNodes = this.configurationPropertiesTreeNodes;
   }
 
   updateOverride(override: ConfigurationProperty): void {
@@ -36,12 +46,12 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
     const key = override.formControlName.split(/\.(.+)/)[1];
 
     configurationProperties = <ConfigurationProperty[]>(
-      this.configurationProperties[group]
+      this._configurationProperties[group]
     );
 
     if (!configurationProperties) {
       // If we couldn't find the group within the top-level property groups, lets peek at datasources...
-      const datasources = this.configurationProperties['datasources'];
+      const datasources = this._configurationProperties['datasources'];
       configurationProperties = <ConfigurationProperty[]>datasources[group];
     }
 
@@ -80,7 +90,9 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
   }
 
   private createConfigurationPropertyTree(): void {
-    forOwn(this.configurationProperties, (configGroup, groupKey) => {
+    const groupNodes: TreeNode[] = [];
+
+    forOwn(this._configurationProperties, (configGroup, groupKey) => {
       const childrenNodes: TreeNode[] = [];
       // For each configuration group, create a root-level node
 
@@ -113,8 +125,10 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
         expanded: false,
         leaf: false
       };
-      this.configurationPropertiesTreeNodes.push(groupNode);
+      groupNodes.push(groupNode);
     });
+
+    this.configurationPropertiesTreeNodes = groupNodes;
   }
 
   private mapLeafNodes(
--- a/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
@@ -64,8 +64,7 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
 
   expandOrCollapse(node: TreeNode): void {
     node.expanded = !node.expanded;
-    // Change detection is not triggered unless the TreeNode array is replaced
-    // TODO: See if we can avoid this hack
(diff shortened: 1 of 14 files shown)


webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
```diff
@@ -55,6 +65,7 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
   expandOrCollapse(node: TreeNode): void {
     node.expanded = !node.expanded;
     // Change detection is not triggered unless the TreeNode array is replaced
+    // TODO: See if we can avoid this hack
```
burzum619: Open to suggestions on this one. I tried using `ChangeDetectorRef` to force the change detect changes, and that didn't seem to do the trick.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
```diff
@@ -9,22 +9,32 @@ import { forOwn, cloneDeep, get } from 'lodash';
...
+    this._configurationProperties = properties;
+    if (this.form) {
+      this.createConfigurationPropertyTree();
+    }
+  }
+
   @Input()
   propertiesArrayName: string;
   @Input()
   form: FormGroup;
```
esotrope: this where the race condition is?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-details.component.ts
```diff
@@ -28,59 +40,30 @@ export class TenantConfigurationDetailsComponent implements OnInit {
...
-            } else {
-              this.localizationOverrides.push(
-                new ConfigurationProperty(key, value)
-              );
-              locationOverrideFormArray.controls.push(new FormControl(value));
-            }
-          }
-        }
-      });
+  ngOnChanges(changes: SimpleChanges): void {
```
esotrope: is this only fired on `@Input()` changes? i've actually never used this

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-details.component.ts
```diff
@@ -104,8 +87,14 @@ export class TenantConfigurationDetailsComponent implements OnInit {
...
     };
-    console.log(updatedTenant);
-    this.service.update(updatedTenant);
+
+    this.service
+      .update(updatedTenant)
+      .subscribe(
+        () => this.tenantUpdated.emit(),
+        error =>
+          this.notificationService.error({ id: 'tenant-config.errors.update' })
```
esotrope: 👍

webapp/src/main/webapp/src/app/admin/sandbox-tenants/mapper/tenant.mapper.ts
```diff
@@ -178,3 +172,19 @@ export function mapConfigurationProperties(
...
+  );
+  return configProperties;
+}
+
+function mapFromLocalizationOverrides(overrides: ConfigurationProperty[]): any {
+  let localizationOverrides = {};
+  overrides.forEach(override => {
+    localizationOverrides[override.key] = override.value;
+  });
+  return localizationOverrides;
```
esotrope:
```typescript
return overrides.reduce((localizationOverrides, { key, value}) => {
  localizationOverrides[key] = value;
}, {});
```

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/tenant.component.ts
```diff
@@ -32,8 +46,8 @@ export class TenantConfigurationComponent {
     let modal: DeleteTenantConfigurationModalComponent = modalReference.content;
     modal.tenant = tenant;
     this._modalSubscriptions.push(
-      modal.deleted.subscribe(tenant => {
-        console.log(tenant);
+      modal.deleted.subscribe(() => {
+        this.getTenants();
```
esotrope: i'd go with a store for this - no use in hitting the endpoint again.
you can see the pattern in `reports.component`

webapp/src/main/webapp/src/app/admin/sandbox-tenants/service/tenant.service.ts
```diff
@@ -2,11 +2,12 @@ import { Observable } from 'rxjs';
...
-  mapFromSandbox,
-  mapFromTenant,
-  mapTenant
-} from '../mapper/tenant.mapper';
+import { mapFromTenant, mapTenant } from '../mapper/tenant.mapper';
+import { catchError, map } from 'rxjs/operators';
+import { AdminServiceRoute } from '../../../shared/service-route';
+import { ResponseUtils } from '../../../shared/response-utils';
+
+const ServiceRoute = `${AdminServiceRoute}/tenants`;
```
esotrope: I'd call this `ResourceRoute`

webapp/src/main/webapp/src/app/admin/sandbox-tenants/service/tenant.service.ts
```diff
@@ -17,244 +18,58 @@ import {
...
+      .get(`${ServiceRoute}/`)
+      .pipe(map(apiTenants => apiTenants['applicationTenantConfiguration']));
   }
 
   /**
    * Gets all sandbox configurations for the system
    */
   getAll(): Observable<TenantConfiguration[]> {
-    return new Observable(observer => observer.next(this.mockData));
+    return this.dataService.get(`${ServiceRoute}/`).pipe(
```
esotrope: nit: no need for the tailing slash

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
burzum619: Sort of. Very similar situation here as well, but this one work with just the check on the setter.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/service/tenant.service.ts
```diff
@@ -17,244 +18,58 @@ import {
...
-  update(tenant: TenantConfiguration): void {
-    let mappedTenant = mapFromTenant(tenant);
-    // TODO: Call the update API
-    let index = this.mockData.findIndex(s => s.code === tenant.code);
-    this.mockData[index] = tenant;
+  /**
+   * Updates an existing tenant
+   * @param tenant - The tenant to update
+   */
+  update(tenant: TenantConfiguration): Observable<void> {
```
esotrope: update usually returns the updated object

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-details.component.ts
burzum619: Yep. You can peek at the `SimpleChanges`  to see which fields triggered the change and what old/new values are. In this case I don't care what order they are defined in though, I just need to make sure they are all defined before I process.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/service/tenant.service.ts
```diff
@@ -17,244 +18,58 @@ import {
...
   }
 
-  create(tenant: TenantConfiguration): void {
-    // TODO: Call the create API
-    this.mockData.push(tenant);
+  /**
+   * Creates a new tenant
+   * @param tenant - The tenant to create
+   */
+  create(tenant: TenantConfiguration): Observable<void> {
```
esotrope: create usually returns the created object - usefull for when the ID is dynamically assigned once inserted in a DB

webapp/src/main/webapp/src/app/admin/sandbox-tenants/mapper/tenant.mapper.ts
burzum619: thanks, I always forget about reduce.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/tenant.component.ts
burzum619: thanks, I wasn't sure if we were using rxjs in this app

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
esotrope: what if the setter is invoked before the form is input though?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/service/tenant.service.ts
burzum619: yep - in this case the user is having to define the unique ids/keys on creation time, so I thought it would be unnecessary/unused. But I believe the backend API is in fact returning the object anyway, so I'll go ahead and change that here.

PR #1922, opened 2019-05-24
Adding tenant ID and reset button informational popover

burzum619:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -912,6 +913,7 @@
       "search": "Search by key or value"
     },
     "empty": "No tenant configurations found. Create a new tenant by clicking the \"Create New Tenant\" button",
+    "reset-info": "Click this button to reset the value back to its original default value",
```
burzum619: Happy to take suggestions on better informational texts

ftdholbrook: You don't want my advice on that stuff. :-)

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.html
```diff
@@ -103,6 +103,10 @@
                 rowNode.node.data.value !== rowNode.node.data.originalValue
               "
               (click)="resetClicked(rowNode.node.data)"
+              popover="{{ 'tenant-config.reset-info' | translate }}"
+              placement="left"
+              container="body"
+              triggers="mouseenter:mouseleave"
```
esotrope: will there be a modal in conjunction with this?
wondering if we don't just force them to click through a modal to confirm

burzum619: I don't think it's necessary since the reset can be cancelled and changes need to be saved be explicitly clicking the "Save" button for the tenant.

esotrope: ah i see. so you would hit reset and then save

webapp/src/main/webapp/src/app/admin/sandbox-tenants/mapper/tenant.mapper.ts
```diff
@@ -12,6 +12,7 @@ export function mapTenant(
 ): TenantConfiguration {
   return <TenantConfiguration>{
     code: tenant.tenant['key'],
+    id: tenant.tenant['id'],
     label: tenant.tenant['name'],
     description: tenant.tenant['description'],
```
esotrope: looks like we are missing typing or digging at internals here? why the `['prop']` syntax?
also `...tenant.tenant` would give you the first 4 props

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: only suggestion is original and default is redundant i'd maybe say `system default value` or something

PR #1926, opened 2019-05-24

vgonzalez9:
Created the initial Sandbox service with CRUD and a reset endpoint
it uses an inMemory repo to store the sandboxConfigurations
and simple checks on create & update
At startup to populates the repo with 2 mock sandboxes.

NOTE: this is post before fully complete to allow for UI integration.
Still need to pull in real data lookup and saving sandbox data.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/DataSet.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/DataSet.java
@@ -1,0 +1,28 @@
+package org.opentestsystem.rdw.admin.model;
+
+/**
+ * DataSet is a preconfigured set of metadata the Sandbox uses.
+ */
+public class DataSet {
+    private String dataSet; // label name of the DataSet
+    private String key; // unique key
+
+    DataSet() {}
+
+    public String getDataSet() {
+        return dataSet;
+    }
+
+    public void setDataSet(final String dataSet) {
+        this.dataSet = dataSet;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public void setKey(final String key) {
+        this.key = key;
+    }
+
+}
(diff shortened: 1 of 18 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/DataSet.java
```diff
@@ -0,0 +1,28 @@
+package org.opentestsystem.rdw.admin.model;
+
+/**
+ * DataSet is a preconfigured set of metadata the Sandbox uses.
+ */
+public class DataSet {
+    private String dataSet; // label name of the DataSet
```
burzum619: Can we rename this to "label"? The object itself is the `DataSet`

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/InMemorySandboxConfigurationRepository.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.stereotype.Repository;
+
+import java.util.HashMap;
+import java.util.Set;
+
+@Repository
+public class InMemorySandboxConfigurationRepository implements SandboxConfigurationRepository {
+
+    private HashMap<String, SandboxConfiguration> sandboxConfigurationMap = new HashMap<>();
```
burzum619: when declaring the property here,  can we use the interface (`Map`)?

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/SandboxController.java
```diff
@@ -0,0 +1,60 @@
...
+@RequestMapping("/sandboxes")
+public class SandboxController {
+
+    private SandboxService service;
+
+    public SandboxController(SandboxService service) {
+        this.service = service;
+    }
+
+    @GetMapping
```
burzum619: Does this also need @ResponseBody ?

PR #1928, opened 2019-05-25
make the Tenant instances stored in Tenant Properties a Map

ftdholbrook:
- Map<TenantKey,Tenant> instead of a List<Tenant>
- simplifies modification of tenant data in git, ties it directly to other tenant configuration
- depends on changes in common


admin-service/src/test/resources/application-tenant_az.yml
```diff
@@ -1,4 +1,12 @@
 #simulating the spring cloud config search path feature with "tenant_az" profile for testing purposes
+tenantProperties:
+  tenants:
+    NVX:
+      id: AZ
+      key: AZ #To verify id and key need not match
```
malaffoon: but the id and the key DO match?

PR #1929, opened 2019-05-25

vgonzalez9:
Fixed the IT build failure that could not locate the class AggregateReportingPropertiesTenant.
Moved the AggregateReporting configuration classes to the rdw-reporting-common project.
Remove the service to service dependency from build.gradle
A new Jira has been created to revist this change 
https://github.com/SmarterApp/RDW_Reporting/pull/1929#issue-282257644

Moved the files to .../rdw/reporting/common/configuration/ 
Updated affected files.
Successfully ran ./gradlew clean build IT


malaffoon:
I'd vote for either not having the `olap` qualifier in the package or flipping it so it is under `configuration`, and putting this with the other common configuration (so `.../rdw/reporting/common/configuration/` instead of `.../rdw/olap/configuration/`). My thought is that all this configuration stuff belongs together. Perhaps `configuration` is overused but that's what the package name is right now. And, yeah, i get that these aren't quite as common as ReportingSystemProperties but i think grouping them is fine.

PR #1930, opened 2019-05-26
Feature/git persist extended

ftdholbrook:
- add support to persist TenantProperties/Tenant
- change API to accept full TenantConfiguration
- still TODO localization


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/ApplicationTenantConfigurationPersistence.java
```diff
@@ -59,4 +64,12 @@ public void setDatasources(Map<String, Map<String, Map<String, DataSourceElement
     public void setArchive(Map<String, Map<String, ArchiveProperties>> archive) {
         this.archive = archive;
     }
+
+    public Map<String, Map<String, Tenant>> getTenantProperties() {
```
malaffoon: perhaps javadocs so i can understand what the return is, exactly? Is it map of tenantKey to the map of tenantKey to tenant? Why would each tenant have a different set of tenant maps??

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGit.java
```diff
@@ -30,6 +34,7 @@
 
     private static final Logger logger = LoggerFactory.getLogger(TenantConfigurationPersistenceServiceGit.class);
     private static final String configFileName = "application.yml";
+    private static final String tenantsFileName = "application.yml";
```
malaffoon: I have a completely indefensible nit: can these be `ConfigFileName` and `TenantsFileName`? I really dislike the ALL_CAPS const thing but, for whatever reason, i like the leading uppercase.

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGitIT.java
```diff
@@ -104,10 +106,17 @@ public void testSaveFile() throws Exception {
         applicationTenantConfiguration.setDataSources(ImmutableMap.of("reporting_ro", dataSourceElements));
         applicationTenantConfiguration.setArchiveProperties(archivePropertiesTenant);
 
-        ApplicationTenantConfigurationPersistence applicationTenantConfigurationPersistence = new ApplicationTenantConfigurationPersistence("CAX", applicationTenantConfiguration);
+        TenantProperties.Tenant tenant = new TenantProperties.Tenant();
+        tenant.setId("CAX");
+        tenant.setKey("CAX");
+        tenant.setName("California Test");
```
malaffoon: I ran into this too, perhaps we can add a constructor that takes the three values id, key, name?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGit.java
ftdholbrook: really?  I think that is the worst of all worlds, but I'll do it.

burzum619: What's wrong with the `ALL_CAPS_CONSTS`?   🤔

ftdholbrook: These are private string literals, there's nothing special about them, they aren't part of a public api.  Coming from an immutable by default language (where you signal the exceptional mutability) it strikes me as really weird treating stuff like this as special. :-)

malaffoon: Like i said, it's indefensible. I'm fine with leaving them.

PR #1931, opened 2019-05-27

vgonzalez9:
Moving the AggregateReporting configuration files to rdw_reporting_common
first commit with successful IT build. There will be another one with all of the SandboxService changes.

This was done with the latest code, and because automatic and patched merges failed.
The second commit is the Sandbox API changes.


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/ApplicationSandboxConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+import org.opentestsystem.rdw.multitenant.TenantProperties;
+import org.opentestsystem.rdw.reporting.common.configuration.AggregateReportingPropertiesTenant;
+import org.opentestsystem.rdw.reporting.common.configuration.ReportingSystemPropertiesImpl;
+
+import java.util.Set;
+
+public class ApplicationSandboxConfiguration {
+    private TenantProperties.Tenant tenant; // Tenant used to create sandbox, and where sandbox exists
+    private Set<DataSet> dataSets; // collection of dataSets preconfigured for this tenant
+    private ReportingSystemPropertiesImpl reporting; // default Reporting system properties
```
malaffoon: please put these comments in the javadocs for the getters

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/ApplicationSandboxConfiguration.java
```diff
@@ -0,0 +1,47 @@
+package org.opentestsystem.rdw.admin.model;
+
+import org.opentestsystem.rdw.multitenant.TenantProperties;
+import org.opentestsystem.rdw.reporting.common.configuration.AggregateReportingPropertiesTenant;
+import org.opentestsystem.rdw.reporting.common.configuration.ReportingSystemPropertiesImpl;
+
+import java.util.Set;
+
+public class ApplicationSandboxConfiguration {
```
malaffoon: overview javadoc would be helpful

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/Sandbox.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public void setId(final String id) {
+        this.id = id;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public void setKey(final String key) {
```
malaffoon: I'm assuming `key` here is the sandbox equivalent of tenant key?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/DataSet.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public String getLabel() {
+        return label;
+    }
+
+    public void setLabel(final String label) {
+        this.label = label;
+    }
+
+    public String getKey() {
```
malaffoon: What is `key` here? I'm guessing it is an identifier of some kind so i'd much prefer `id`.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/Sandbox.java
```diff
@@ -0,0 +1,59 @@
+package org.opentestsystem.rdw.admin.model;
+
+public class Sandbox {
+    private String id;
+    private String key;
+    private String name;
+    private String description;
```
malaffoon: Is there any way we can share the common Tenant class here? Perhaps we can move `Tenant` out of `TenantProperties` and then use it as a base class here? That way it becomes much more obvious that Sandbox isA Tenant.

ftdholbrook: I like that idea.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/SandboxConfigurationRepository.java
```diff
@@ -0,0 +1,40 @@
+package org.opentestsystem.rdw.admin.repository;
+
+import org.opentestsystem.rdw.admin.model.SandboxConfiguration;
+
+import java.util.Set;
+
+public interface SandboxConfigurationRepository {
+    /**
+     * @return the sandbox default configuration and the list of existing sandboxes
```
malaffoon: Bit confusing, is the default configuration just one of the entries in the set?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/SandboxConfigurationRepository.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    Set<SandboxConfiguration> getAll();
+
+    /**
+     * @param sandboxConfiguration to create
+     * @return the newly added sandboxConfiguration
+     */
+    SandboxConfiguration add(SandboxConfiguration sandboxConfiguration);
```
malaffoon: `create` would be more consistent

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/SandboxConfigurationRepository.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @param key of the Sandbox to reset its DataSet
+     */
+    void reset(String key);
+
+    /**
+     * @param key of the sandbox to check
+     * @return true if the sandbox key exists
+     */
+    Boolean sandboxExists(String key);
```
malaffoon: This should be `boolean exists(String key)`: it is a method on the SandboxRepository so `sandbox` is redundant, and the return will always be a value so using the primitive `boolean` is appropriate.

malaffoon:
@vgonzalez9, i merged develop (with my changes to common) into this branch ... didn't want you to have to deal with conflicts.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/Sandbox.java
vgonzalez9: Will fix.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/SandboxConfigurationRepository.java
vgonzalez9: You are correct, it really just the defaults of a tenant.
will clarify.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/Sandbox.java
malaffoon: I've separated out Tenant so you don't have to worry about that part.

PR #1935, opened 2019-05-28
Feature/sandbox service apis update

vgonzalez9:
Fix some of UI Integration fixes
Added missing @JsonProperty that was accidentally removed.
Added customer deserializers for Set and List in ReportingSystemSettings & AggregateReportingProperties to handle a single item to convert to a list
Fix the PUT for the tenantService - it's checking tenantProperties when that isn't ready to be checked
Fix the PUT for SandboxService - it was using tenantKey for the save and sandbox.key
Also for a sandbox create, the id & key  are automatically generated, using the tenantId and a string version of the currentTime in ms


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/ApplicationSandboxConfiguration.java
```diff
@@ -11,12 +12,12 @@
     /**
      * Default Reporting system properties
      */
+    @JsonProperty("reporting")
```
malaffoon: I thought it used the variable name by default; just curious what it was calling the property?

vgonzalez9: its using the class name

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
```diff
@@ -88,4 +93,13 @@ private void validateExistingSandbox(final SandboxConfiguration sandboxConfigura
...
         }
     }
+
+    /**
+     * TODO Improve
+     * @return a tenant based ID - using current time
+     */
+    private String generateId(String tenantId) {
+        String ts = String.valueOf(System.currentTimeMillis());
+        return tenantId + ts;
```
malaffoon: perhaps a 4-6 character hash of the timestamp?

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AggregateReportingPropertiesTenant.java
```diff
@@ -1,13 +1,19 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
 import java.util.Set;
 
 public class AggregateReportingPropertiesTenant implements AggregateReportingProperties {
 
+    @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
     private Set<String> assessmentTypes;
 
+    @JsonDeserialize(using = CustomSetStringDeserializer.class)
```
burzum619: Shouldn't all of these properties have the same annotations?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/ApplicationSandboxConfiguration.java
burzum619: I think by default its using property/variable names.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AggregateReportingPropertiesTenant.java
vgonzalez9: I created 2 different serializers one for List and one for Set. Or are you suggesting the whole class should be using the same one?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/ApplicationSandboxConfiguration.java
vgonzalez9: And you both are correct that annotation is not needed and the name is use by default.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AggregateReportingPropertiesTenant.java
vgonzalez9: Thanks simplified how this is being done, and now properly handling the effectiveLanguages List.

PR #1936, opened 2019-05-28
stubbed out localization persistence with mock data in anticipation o…

ftdholbrook:
…f upstream changes to TenantConfiguration.localization

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGit.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGit.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.admin.multitenant.git;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.io.FileUtils;
 import org.eclipse.jgit.api.Git;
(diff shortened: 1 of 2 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGit.java
```diff
@@ -95,16 +97,34 @@ public synchronized void saveTenantConfiguration(TenantConfiguration tenantConfi
                 new UsernamePasswordCredentialsProvider(tenantConfigurationPersistenceProperties.getGitUsername(),
                         tenantConfigurationPersistenceProperties.getGitPassword());
 
+        makeTenantDirectory(tenant);
         writeApplicationTenantConfiguration(applicationTenantConfiguration, tenant);
 
+        //TODO: Mock localization until TenantConfiguration changes
+        // writeLocalization(tenantConfiguration.getLocalization());
+        logger.warn("Using mock localization data");
```
ftdholbrook: We can hold off this PR until the upstream change is in place and update it, or merge it now and make the change at a later time.

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/TenantConfigurationPersistenceServiceGitIT.java
```diff
@@ -100,7 +98,7 @@ public void testSaveFile() throws Exception {
 
         ArchivePropertiesTenant archivePropertiesTenant = new ArchivePropertiesTenant();
         archivePropertiesTenant.setPathPrefix("/CA");
-        archivePropertiesTenant.setUriRoot( "s3://ca-archive");
+        archivePropertiesTenant.setUriRoot("s3://ca-archive");
```
malaffoon: Is this really being used, i.e. is the IT really writing to S3? Does it need to?

ftdholbrook: no, it is just so there is some data to move around, this test isn't actually exercising anything but the git operations

PR #1940, opened 2019-05-28
revisit where grants are being filtered

malaffoon:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/auth/DefaultAuthorizationService.java
```diff
@@ -50,12 +51,26 @@ public DefaultAuthorizationService(
     @Override
     public Set<Grant> getGrants(final String[] encodedGrants) {
         final ImmutableSet.Builder<Grant> grants = ImmutableSet.builder();
+        String stateId = null;
         for (final String encodedGrant : encodedGrants) {
             try {
                 final Grant grant = Grant.fromString(encodedGrant);
```
malaffoon: Making two changes here: filter based on known roles (that was being done in the SAML service before) and reduce the set to only those grants for the first state encountered.

I moved the logic here because it felt right. Imagine a situation where the grants weren't coming in via SAML, we'd want this same functionality.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -67,10 +67,9 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
         }
 
         final String stateId = grants.stream()
-                .filter(g -> g.getRole() != null && g.getStateId() != null)
-                .filter(g -> ReportingRoles.contains(g.getRole()))
-                .findFirst()
                 .map(Grant::getStateId)
+                .filter(Objects::nonNull)
+                .findFirst()
                 .orElse(NO_TENANT_ID);
```
malaffoon: I'm not sure this is the right thing to do if no tenant/state is found. Should we be throwing? Or creating a user with no permissions or ...?

ftdholbrook: I was wondering the same thing.  At some point "NO_TENANT_ID" becomes an error condition.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -50,10 +48,16 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
             logger.info("User's login (mail:{}) does not match their name id (NameID:{})", username, credential.getNameID().getValue());
         }
 
+        // the user must have reporting grants with a state so we can determine tenant id
         final Set<Grant> grants = authorizationService.getGrants(encodedGrants);
+        final String stateId = grants.stream()
+                .map(Grant::getStateId)
+                .filter(StringUtils::isNotBlank)
+                .findFirst()
+                .orElse(null);
```
malaffoon: Ok, i've changed this to create a user with no permissions if there isn't a state in the grants. @ftdholbrook ?

Also, i discovered that the state id may come back either null or an empty string when it isn't there.

PR #1946, opened 2019-05-29
Fixing bug with reset functionality. Some minor refactoring

burzum619:


webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
```diff
@@ -89,6 +71,27 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
     return false;
   }
 
+  private setPropertyValue(override: ConfigurationProperty, newVal: string) {
+    const group = override.formControlName.split(/\.(.+)/)[0];
+    const key = override.formControlName.split(/\.(.+)/)[1];
```
esotrope: this kind of parsing is a bit of a red flag for me. usually means we are packing data into strings when it should be represented as an object. what are we trying to accomplish here?

burzum619: We need to assign unique formControlNames to each input field in order to associate them with a `ConfigurationProperty`. Normally we could just use the key itself, but thats not necessarily unique on its own (for example, each datasource has the same properties).

burzum619: We join the group and key separated by a `.` inside of the logic that creates the tree and its treenodes (see `property-override-tree-table.component.ts` line ~143-155

esotrope: can we just expand on config property interface to have this data already parsed out?
```typescript
interface ConfigurationProperty {
  key: string;
  group: string;
  compositeId: string; /* A composite of group + key used as the form control name */
  value: string;
}
```

burzum619: `formControlName` is already a property of `ConfigurationProperty`. Are you suggesting changing the name to `compositeId` and then creating a getter that will return `{group}.{key}`?

esotrope: i was just thinking instead of parsing the `group` and `key` from the `formControlName` using `String.split()` why not already store it on the `ConfigurationProperty` object instances under the field names `group` and `key`

PR #1947, opened 2019-05-29
inject tenant id in group processing message

malaffoon:
There are more changes here than you'd expect. That's mostly because i thought i had to inject the bearer token header and that resulted in some cleanup.


admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupsSource.java
```diff
@@ -30,22 +32,16 @@ public DefaultGroupsSource(final Source source) {
...
+    public void send(final byte[] body, final String contentType, final Long importId, final User user) {
         logger.info("sending GROUPS data for {}", importId);
 
         final RdwMessageHeaderAccessor accessor = wrap(null)
                 .setReceivedNow()
                 .setContent("GROUPS")
                 .setContentType(contentType)
-                .setImportId(importId);
+                .setImportId(importId)
+                .setTenantId(user.getTenantId());
```
malaffoon: The change. This is needed because the group processor expects it to be in the message. The ingest and reporting sides of the house use different techniques to get the tenant out of a message.

admin-service/src/test/java/org/opentestsystem/rdw/admin/ApplicationIT.java
```diff
@@ -7,12 +7,12 @@
 import org.springframework.test.context.junit4.SpringRunner;
 
 @RunWith(SpringRunner.class)
-@ActiveProfiles("adminTest")
+@ActiveProfiles("test")
```
malaffoon: this was inconsistent with the other projects, not sure why

admin-service/src/main/java/org/opentestsystem/rdw/admin/security/WebSecurityConfiguration.java
```diff
@@ -1,15 +0,0 @@
-package org.opentestsystem.rdw.admin.security;
```
malaffoon: I really couldn't see a reason for this. Perhaps it is left-over.

PR #1948, opened 2019-05-29
fixes sample script loading issue

esotrope:
image

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineScriptService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineScriptService.java
@@ -1,6 +1,5 @@
 package org.opentestsystem.rdw.admin.service.impl;
 
-import com.google.common.io.Resources;
 import org.apache.commons.io.IOUtils;
 import org.opentestsystem.rdw.admin.model.Pipeline;
 import org.opentestsystem.rdw.admin.model.PipelineScript;
@@ -13,8 +12,6 @@ import org.springframework.stereotype.Service;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URL;
-import java.nio.charset.Charset;
 import java.util.Optional;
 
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/admin/ingest-pipeline/model/pipeline.ts
```diff
@@ -89,21 +89,6 @@ export interface PipelineScript extends Script {
...
-
-  /**
-   * The script publish time
-   */
-  publishedOn?: Date;
-
-  /**
-   * The user who published the script
-   */
-  publishedBy?: string;
```
esotrope: unrelated cleanup

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineScriptService.java
```diff
@@ -38,7 +35,7 @@ public PipelineScript get(final int pipelineId, final int id) {
 
         // if there is a saved script return it
         final PipelineScript savedScript = repository.findById(id);
-        if (savedScript != null) {
+        if (savedScript != null && savedScript.getPipelineId() == pipelineId) {
```
esotrope: unrelated "security"

webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
```diff
@@ -166,7 +176,10 @@ export class PipelineComponent implements ComponentCanDeactivate, OnDestroy {
         ]) => {
           const pipeline = {
             ...basePipeline,
-            script,
+            script: script || {
+              id: -1,
+              pipelineId: basePipeline.id
+            },
```
esotrope: wonky but gets the job done by putting a placeholder script in that results in a sample script lookup

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineScriptService.java
burzum619: why no `===`?

esotrope: ha! if only java let me - i would do it to troll people

PR #1950, opened 2019-05-30
fixes test

esotrope:


admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineServiceTest.java
```diff
@@ -45,7 +49,8 @@ public void before() {
                 pipelineTestRepository,
                 publishedPipelineRepository,
                 pipelineFactory,
-                pipelineOutputComparisionService
+                pipelineOutputComparisionService,
+                scriptSource
```
esotrope: of course the placeholder test made with good intentions just got in my way ; j

PR #1954, opened 2019-05-31
Fixing bug where last character in tenant key was sometimes not capitalized

burzum619:


webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
```diff
@@ -82,7 +82,7 @@ export class NewTenantConfigurationComponent implements OnInit, AfterViewInit {
     );
     const newTenant = {
       ...this.tenantForm.value,
-      code: this.tenantForm.get('key').value,
+      code: this.tenantForm.get('key').value.toUpperCase(),
```
esotrope: is this just for display purposes? or do we actual take all their lowercase characters and uppercase them before storing it in the DB?

burzum619: not necessarily in the DB, but in the tenant configuration itself. Eventually this stuff cascades into the tenant config .yml files.

esotrope: yeah but why? is that a requirement? if we enforce all caps we should make the UI do it too so when they type it turns their characters into all caps. i think it would be confusing to allow them to put in say `ca` and then it shows up as `CA` everywhere in the config and UI no?

burzum619: It is a requirement, and the UI is converting key inputs for lower case characters to uppercase on keyup. This bugfix is for an edgecase that occurs when a user enters a key using lower case letters (which are converted to upper on the UI as keys are pressed), but for some reason when the user submits the form, the last letter in the key is not convert to uppercase (although it appears as uppercase in the UI). This resulted in bug RP-423.

burzum619: At the moment I am not masking lower case characters, I'm simply converting them using the `onkeyup` event on the input field.

esotrope: gotcha - thanks for clarifying. just making sure we don't do something to trip people up.

PR #1955, opened 2019-05-31
add a tenantKeyGenerator bean

malaffoon:


common/src/main/java/org/opentestsystem/rdw/reporting/common/multitenant/ReportingTenantMessageRequestResolver.java
```diff
@@ -24,7 +24,7 @@ public String resolveTenantId(final Message<?> message) {
         try {
             final User user = messageSecurityService.getUser(message);
             tenantId = user.getTenantId();
-            logger.info("Resolving user {} to tenant {} for message.", user.getUsername(), tenantId);
+            logger.debug("Resolving user {} to tenant {} for message.", user.getUsername(), tenantId);
```
malaffoon: This is going to be a too chatty since every report generated involves multiple messages

PR #1956, opened 2019-06-01
Feature/rp366 add tenant lookup

vgonzalez9:
Pulled in all of the changes from the the PR #1920
Created a new Service to manage retrieving the default and tenant specific properties from each service.
It currently grabs the the first instance of the scopedTarget. properties returned from the rdw services.
TenantService will call the service when the getAll() endpoint is called.
Still using in Memory to save configuration.
To run locally, make sure that the tenantProperties are configured in the services and webapp
Removed unused files


admin-service/src/main/java/org/opentestsystem/rdw/admin/client/RdwService.java
```diff
@@ -0,0 +1,16 @@
+package org.opentestsystem.rdw.admin.client;
+
+/**
+ * Contains the RDW Service's Url to obtain the service's current properties
+ */
+public class RdwService {
```
esotrope: if we're just wrapping one field here can we unwrap?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClient.java
```diff
@@ -0,0 +1,302 @@
...
+                .getServices()
+                .values()
+                .stream()
+                .map(RdwService::getUrl)
+                .collect(Collectors.toList());
+        List<ResponseEntity<String>> responses = serviceUrls.stream()
+                .map(url -> restTemplate.getForEntity(url + "/configprops", String.class))
+                .collect(Collectors.toList());
+
+        return responses;
```
esotrope: may as well one-line the above code since it chains.

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClient.java
```diff
@@ -0,0 +1,302 @@
...
+    public TenantConfigurationLookupClient(TenantConfigurationLookupProperties tenantConfigurationLookupProperties) {
+        this.tenantConfigurationLookupProperties = tenantConfigurationLookupProperties;
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    }
+
+    /**
+     * @return a List of Response Entities for each tenant aware service
+     */
+    public List<ResponseEntity<String>> getRawConfigurations() {
+        RestTemplate restTemplate = new RestTemplate();
```
esotrope: i think we want to autowire this in the constructor since we _may_ want to take advantage of shared settings on it. other than that we should at least reuse the instance for all calls - i don't think it is stateful right?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClient.java
```diff
@@ -0,0 +1,302 @@
...
+    public Optional<Tenants> getTenantSetProperties(List<ResponseEntity<String>> responseEntities) {
+        for (ResponseEntity<String> response : responseEntities) {
+            Optional<Tenants> settings = parseTenantSetProperties(response);
+            return settings;
+        }
+        return Optional.empty();
+    }
+
+
+    private Optional<ReportingSystemSettings> parseReportingSystemSettings(ResponseEntity<String> entity) {
```
esotrope: do we need this if we do `restTemplate.getForEntity(path, ClassWithReportingSystemSettingsField.class)`?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupProperties.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.admin.client;
+
+import java.util.Map;
+
+/**
+ * This contains the RDW service name and their corresponding endpoints
+ */
+public class TenantConfigurationLookupProperties {
```
esotrope: same comment about wrapping. seems like we just need `Map<String, String> servicePathByServiceId`

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/Tenants.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Used to retrieve TenantProperties
+ */
+public class Tenants {
+    private Map<String, Tenant> tenants = new HashMap<>();
+    @JsonProperty(value = "status")
```
esotrope: looks safe to rm

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/ApplicationTenantConfiguration.java
```diff
@@ -20,8 +21,8 @@
     private Map<String, DataSourceElementsTenant> dataSources = new LinkedHashMap<>();
     @JsonProperty("archive")
     private ArchivePropertiesTenant archiveProperties;
-
-    //private AggregateReportSettings aggregator;
+    @JsonProperty("aggregator")
```
esotrope: first time i've heard this referred to as aggregator - usually we call it olap or aggregate-service

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfigurationStubs.java
```diff
@@ -215,6 +221,41 @@ private static TenantConfiguration getTenantConfiguration(TenantConfiguration te
...
 
+    public static  AggregateReportingPropertiesTenant getAggregrateReportingProperties() {
+        AggregateReportingPropertiesTenant aggregateReportingProperties = new AggregateReportingPropertiesTenant();
+        Set<String> assessmentTypes = new HashSet<>();
+        assessmentTypes.add("ica");
+        assessmentTypes.add("sum");
+        assessmentTypes.add("iab");
+        aggregateReportingProperties.setAssessmentTypes(assessmentTypes);
+        Set<String> assessments = new HashSet<>();
+        assessmentTypes.add("sum");
```
esotrope: guava has some nice convenience methods for one-lining things like this
```
newHashSet("ica", "sum", "iab")
```

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfigurationStubs.java
```diff
@@ -215,6 +221,41 @@ private static TenantConfiguration getTenantConfiguration(TenantConfiguration te
...
+        aggregateReportingPropertiesRoot.setStateAggregateAssessmentTypes(aggregationProperties.getStateAggregateAssessmentTypes());
+        aggregateReportingPropertiesRoot.setStatewideUserAssessmentTypes(aggregationProperties.getStatewideUserAssessmentTypes());
+        Map<String, AggregateReportingPropertiesTenant> aggregatePropertiesMap = new HashMap<>();
+        aggregatePropertiesMap.put("CA", getCAAggregateReportingProperties());
+        aggregateReportingPropertiesRoot.setTenants((HashMap<String, AggregateReportingPropertiesTenant>) aggregatePropertiesMap);
+        return Optional.of(aggregateReportingPropertiesRoot);
+    }
+
     public static String getLocalization() {
         return "{" +
```
esotrope: i'd prefer we store this in a file. and use
```java
IOUtils.toString(getClass().getClassLoader().getResourceAsString("pathToFileFromResourcesDir"), UTF_8)
```

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/ActuatorClientRdwDefaultPropertiesService.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * The service retrieve the configuration properties from the rdw services, for default and current tenants
+ */
+public interface ActuatorClientRdwDefaultPropertiesService {
+
+    /*
+     * @return the ResponseEntities from all of the rdw services
+     */
+    List<ResponseEntity<String>> getAllResponseEntities() ;
```
esotrope: does the consumer use all of these methods or can some just be internal?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultActuatorClientRdwService.java
```diff
@@ -0,0 +1,243 @@
...
+    private TenantConfigurationLookupClient tenantConfigurationLookupClient;
+
+    private ObjectMapper mapper;
+
+    @Autowired
+    DefaultActuatorClientRdwService(TenantConfigurationLookupClient tenantConfigurationLookupClient,
+                                    ObjectMapper objectMapper) {
+        this.tenantConfigurationLookupClient = tenantConfigurationLookupClient;
+        this.mapper = objectMapper.copy();
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
esotrope: 👍

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultActuatorClientRdwService.java
```diff
@@ -0,0 +1,243 @@
...
+            logger.debug("testing response to get for writableDataSource = {}", writableDataSource.get());
+            try {
+                DataSourceElementsTenant writableDataSourceElementsTenant = mapper.readValue(mapper.writeValueAsString(writableDataSource.get()), DataSourceElementsTenant.class);
+                return Optional.of(writableDataSourceElementsTenant);
+            } catch (Exception e) {
+                logger.error("Error parsing writableDataSource from json: {}", writableDataSource);
+            }
+        }
+        return Optional.empty();
+    }
```
esotrope: looks like this method body could be extracted and reused seems really similar to the other methods

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantService.java
```diff
@@ -44,41 +64,227 @@ private void addMockData() {
...
+            }
+            // Add each tenantConfiguration's dataSourceWarehouse overrides
+            Map<String, DataSourceElementsTenant> warehouseTenants = warehouseDataSetProperties.get().getTenants();
+            warehouseTenants.forEach((k,d) -> {
+                if(tenantConfigurationMap.containsKey(k)) {
+                    logger.info("Add tenant {}'s Reporting Warehouse DataSource overrides", k);
+                    tenantConfigurationMap.get(k).getApplicationTenantConfiguration().getDataSources().put("warehouse_rw", d);
+                } // todo tbd on handling if service's warehouse dataSets is not in the overall tenant list
+            });
+        }
```
esotrope: same comment here. seems like we need a for-loop here

admin-service/src/test/resources/aggregateReportingProperties.json
```diff
@@ -0,0 +1,19 @@
+{
```
esotrope: 👍

admin-service/src/test/resources/application-tenant_client_docker.yml
```diff
@@ -0,0 +1,14 @@
+logging:
+  level:
+    org.opentestsystem.rdw.admin: debug
```
esotrope: did you mean to check this in?

admin-service/src/test/resources/application-tenant_client_docker.yml
```diff
@@ -0,0 +1,14 @@
...
+tenant-configuration-lookup-test:
+  services:
+    admin-service:
+      url: http://localhost:8208
+    aggregate-service:
+      url: http://localhost:8308
+    report-processor:
+      url: http://localhost:8086
+    reporting-service:
+      url: http://localhost:8108
```
esotrope: what is this file used for and why is docker in the name? little confused on this one

admin-service/src/test/resources/reportingServiceWritableDataSourceElementsProperties.json
```diff
@@ -0,0 +1,48 @@
+{
```
esotrope: like that you captured these in json files 👍

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/ActuatorClientRdwDefaultPropertiesService.java
```diff
@@ -0,0 +1,141 @@
...
+     * @param responseEntities from the services configProps
+     * @return ReportingSystemSettings from the ResponseEntities with default and tenant information
+     */
+    Optional<ReportingSystemSettings> getReportingSystemSettings(List<ResponseEntity<String>> responseEntities);
+    /**
+     *
+     * @param reportingSystemSettings the settings returned from the service
+     * @return ReportingSystemPropertiesImpl of the default reporting system settings
+     */
+    Optional<ReportingSystemPropertiesImpl> getReportingSystemSettingsDefault(Optional<ReportingSystemSettings> reportingSystemSettings);
```
malaffoon: This is a lot of methods. It implies that each one will make a call out to a service and i don't think we want to be making a bunch of separate calls: we want to fetch all the config props in one go and then filter them, right?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/ActuatorClientRdwDefaultPropertiesService.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * @param warehouseDataSource from the service calls
+     * @return Warehouse DataSourceElementsTenant from the DataSourceElementsProperties
+     */
+    Optional<DataSourceElementsTenant> getWarehouseDataSourceElementsTenantDefault( Optional<DataSourceElementsProperties> warehouseDataSource);
+    /**
+     * @param responseEntities from the service calls
+     * @return Warehouse DataSourceElementsTenant from the responseEntities - contains tenant info
+     */
+    public Optional<DataSourceElementsProperties> getWarehouseDataSource(List<ResponseEntity<String>> responseEntities);
```
malaffoon: ~public~

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/ActuatorClientRdwDefaultPropertiesService.java
malaffoon: Okay, read a bit more ...

We should separate the concerns better:
1. There is a provider that hits the services and retrieves their configprops. This should be isolated because we are going to want to multithread that to fetch the results quickly. With short timeouts, etc. It should be returning a mostly unprocessed thing.
2. It feels like most of the rest of this is parsing configprops. Personally, i think that could be utility methods. But i could see it being a service that you hand the configprops to (like all these additional methods you have). One way you could do that is to have a wrapper class that gets the configprops in the constructor. Like a message header accessor.

It just seems like we could simplify things.

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClient.java
```diff
@@ -0,0 +1,302 @@
...
+                });
+
+        return responseEntityMap;
+    }
+
+    /**
+     * @param responseEntityMap Map of services and their Responses
+     * @return Optional ReportingSystemSettings for the admin-service
+     */
+    public Optional<ReportingSystemSettings> getAdminReportingSystemSetting(Map<String, ResponseEntity<String>> responseEntityMap) {
```
malaffoon: See my comment below: there is too much stuff in here. The client should do just client things: pull results back from the other apps.

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/TenantConfigurationLookupClient.java
```diff
@@ -0,0 +1,302 @@
...
+                .map(url -> restTemplate.getForEntity(url + "/configprops", String.class))
+                .collect(Collectors.toList());
+
+        return responses;
+    }
+
+    /**
+     * @return a Map of services, and their ResponseEntities for each tenant Aware service
+     */
+    public Map<String, ResponseEntity<String>> getRawConfigurationsByService() {
```
malaffoon: I don't think this should be returning ResponseEntity's. Imagine that the implementation of this service is to go read configurations from a shared redis cache: there would be no ResponseEntity involved. So the interface for this should use some other construct. Either a plain ol' string or a DOM of some sort.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultRdwServiceConfigPropsAccessor.java
```diff
@@ -0,0 +1,292 @@
...
+    /**
+     * each wanted propertySetting must have a 'properties' field
+     * @param propertySetting to search for Looking in each of the configProps Responses
+     * @return if found, the first found jsonNode of the propertySetting
+     */
+    private Optional<JsonNode> findPropertySetting(String propertySetting) {
+        // stop looking once setting is found
+        for(String response : configProps.getConfigProps() ) {
+            try {
+                JsonNode rootNode = mapper.readTree(response);
```
malaffoon: You should probably cache the json tree(s)

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultRdwServiceConfigPropsAccessor.java
```diff
@@ -0,0 +1,292 @@
...
+
+    // REPORTING SYSTEM SETTINGS
+    @Override
+    public Optional<ReportingSystemPropertiesImpl> getDefaultReportingSystemSettings() {
+        Optional<ReportingSystemSettings> reportingSystemSettings = parseReportingSystemSettings();
+        if(reportingSystemSettings.isPresent()) {
+            logger.debug("Response to get Default ReportingSystemProperties = {}", reportingSystemSettings.get());
+            try {
+                String node = mapper.writeValueAsString(reportingSystemSettings.get());
+                ReportingSystemPropertiesImpl reporting = mapper.readValue(node, ReportingSystemPropertiesImpl.class);
```
malaffoon: Perhaps add a comment in the javadocs about this general pattern of writing then reading the nodes? Or is it obvious enough and i'm just tired?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultRdwServiceConfigPropsAccessor.java
```diff
@@ -0,0 +1,292 @@
...
+        try {
+            final Optional<JsonNode> propertySetting = findPropertySetting("scopedTarget.aggregateReportingPropertiesRoot");
+            if(propertySetting.isPresent()) {
+                return Optional.of(mapper.treeToValue(propertySetting.get(), AggregateReportingPropertiesRoot.class));
+            }
+        } catch (Exception e) {
+            logger.warn("AggregateReportingPropertiesRoot not found parsing configProps");
+        }
+        return Optional.empty();
+    }
```
malaffoon: I haven't looked closely but it still feels like there is some code repetition we could get reduce.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
```diff
@@ -91,11 +105,47 @@ public void delete(final String key) {
...
 
+    private SandboxConfigurationPackage getCurrentDefaultSandboxPackage() {
+        defaultTenantConfigPropsService.servicesPropertiesRefresh();
+
+        SandboxConfigurationPackage sandboxConfigurationPackage = new SandboxConfigurationPackage();
+
+        ApplicationSandboxConfiguration applicationSandboxConfiguration = defaultTenantConfigPropsService.getDefaultApplicationSandboxConfiguration();
+
+        // TODO remove mock
+        if(applicationSandboxConfiguration.getReporting() == null) {
```
malaffoon: nit: i'm starting to be able to tell who coded a module because the style is inconsistent; not going to lose too much sleep over it but please make whitespace consistent (i.e. `if (application...`) and use `final` whenever possible.

admin-service/src/main/resources/application.yml
```diff
@@ -117,3 +117,14 @@ reporting:
...
+tenant-configuration-lookup:
+  services:
+    admin-service:
+      url: http://localhost:8208
+    aggregate-service:
+      url: http://localhost:8308
+    report-processor:
+      url: http://localhost:8086
+    reporting-service:
+      url: http://localhost:8108
```
ftdholbrook: I'm not sure it makes sense to make the defaults match what will only work with docker.  It seems like these should be what the production names and ports are.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineOutputComparisonService.java
```diff
@@ -12,7 +12,7 @@
 class DefaultPipelineOutputComparisonService implements PipelineOutputComparisonService {
 
     private static final Map<String, Comparer<String>> comparersByInputType = ImmutableMap.<String, Comparer<String>>builder()
-            .put("xml", new XmlComparer())
+         //   .put("xml", new XmlComparer())
```
esotrope: what's this about?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultRdwServiceConfigPropsAccessor.java
```diff
@@ -0,0 +1,292 @@
...
+    @Override
+    public Map<String, DataSourceElementsTenant> getTenantsWarehouseDataSourceElements() {
+        Optional<DataSourceElementsProperties> warehouseDataSourceElementsProperties = parseDataSourceConfigProps("scopedTarget.warehouseDataSourceElementsProperties");
+        if(warehouseDataSourceElementsProperties.isPresent()) {
+            return warehouseDataSourceElementsProperties.get().getTenants();
+        }
+        return ImmutableMap.of();
```
esotrope: could make this slightly more concise for fun

```java
return parseDataSourceConfigProps("scopedTarget.warehouseDataSourceElementsProperties")
    .map(x -> x.getTenants())
    .orElse(ImmutableMap.of());
```

admin-service/src/main/resources/application.yml
ftdholbrook: OK if that's the way the app is generally set up it makes sense to continue with that convention.

PR #1958, opened 2019-06-01
make db status tenant-aware

malaffoon:


admin-service/src/test/java/org/opentestsystem/rdw/admin/status/DatabaseStatusIndicatorTest.java
```diff
@@ -47,17 +54,21 @@ public void itShouldBeReadDatabaseLevel() {
         assertThat(statusIndicator.doLevelCheck(3)).isTrue();
     }
 
-//    @Test
-//    public void itShouldReturnIdealStatus() {
-//        final Status status = statusIndicator.status(3);
-//        assertThat(status.getStatusRating()).isEqualTo(Rating.Ideal.value());
-//        assertThat((List<Status>)status.getDetails().get("databaseOperations")).isNotEmpty();
-//    }
+    @Ignore("TODO - broken, the mock return isn't working properly, can't figure out why")
```
malaffoon: I spent an hour trying to figure this out. There is something in the environment that is messing this up. The test is *exactly* the same as in other modules, but the jdbcOperations mock is just not working. It is getting invoked but deep down in mock-land it is saying that List<String> can't be returned for toString(). No idea what that means.

PR #1960, opened 2019-06-03
Sandbox Login - UI work + Security changes to allow access

burzum619:
This PR includes the basic frontend for sandbox login including a (currently stubbed) service for fetching general sandbox/role data. Additionally, there are some changes that were necessarily for publicly exposing a few APIs, and for returning/accepting anonymous application users.

Most of the APIs that required public exposure were related to translations/localization. In particular the following APIs are called by our frontend translation service:

`/api/reporting-service/subjects/` 
`/api/translations/`
`/api/settings/`

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/Users.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/Users.java
@@ -4,6 +4,7 @@ import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 
 public final class Users {
+    private static final String ANONYMOUS_USER = "anonymousUser";
 
     private Users() {
     }
@@ -12,6 +13,13 @@ public final class Users {
         final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
         if (authentication == null) {
             throw new IllegalStateException("user not authenticated");
+        }
+        if (ANONYMOUS_USER.equals(authentication.getPrincipal())) {
+            return User.builderExt()
+                    .id(ANONYMOUS_USER)
+                    .username(ANONYMOUS_USER)
+                    .password(ANONYMOUS_USER)
+                    .build();
         }
         if (!(authentication.getPrincipal() instanceof User)) {
             throw new ClassCastException("authentication principal must be of type " + User.class.getName());
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/Users.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/security/Users.java
@@ -14,6 +14,7 @@ public final class Users {
         if (authentication == null) {
             throw new IllegalStateException("user not authenticated");
         }
+        // TODO: By the time we get here, we should already haven an "anonymous" user object in our security context
         if (ANONYMOUS_USER.equals(authentication.getPrincipal())) {
             return User.builderExt()
                     .id(ANONYMOUS_USER)
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
@@ -510,6 +510,11 @@ public abstract class SamlWebSecurityConfiguration extends WebSecurityConfigurer
 
         http
                 .authorizeRequests()
+                .antMatchers("/sandbox-login").anonymous()
+                .antMatchers("/api/translations/**").permitAll()
+                .antMatchers("/api/user").permitAll()
+                .antMatchers("/api/settings").permitAll()
+                .antMatchers("/api/reporting-service/subjects").permitAll()
                 .antMatchers("/**").authenticated()
                 .anyRequest().authenticated();
         http
--- a/webapp/src/main/webapp/src/app/app.component.html
+++ b/webapp/src/main/webapp/src/app/app.component.html
@@ -44,10 +44,7 @@
             [ngClass]="{ show: navbarOpen }"
           >
             <ul class="nav navbar-nav navbar-right">
-              <li
-                class="navbar-link nav-item"
-                *ngIf="user && user.firstName && user.lastName"
-              >
+              <li class="navbar-link nav-item" *ngIf="!isAnonymousUser()">
                 <a
                   href="{{
                     applicationSettings ? applicationSettings.userGuideUrl : '#'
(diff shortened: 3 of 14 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -510,6 +510,11 @@ protected void configure(final HttpSecurity http) throws Exception {
 
         http
                 .authorizeRequests()
+                .antMatchers("/sandbox-login").anonymous()
+                .antMatchers("/api/translations/**").permitAll()
+                .antMatchers("/api/user").permitAll()
+                .antMatchers("/api/settings").permitAll()
+                .antMatchers("/api/reporting-service/subjects").permitAll()
```
burzum619: We could potentially get around having to expose these APIs publicly by doing away with localization for the sandbox login screen (I'm not sure if localization is a requirement for this particular feature anyway). Basically we would just hard-code the labels in the template.

webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -44,7 +44,7 @@
             [ngClass]="{ show: navbarOpen }"
           >
             <ul class="nav navbar-nav navbar-right">
-              <li class="navbar-link nav-item">
+              <li class="navbar-link nav-item" *ngIf="!isAnonymousUser()">
```
burzum619: We do not want to show these particular nav items in the header if the user isn't actually logged in yet, especially the "Welcome <firstname> <lastname>" dropdown.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -431,11 +431,21 @@ export const routes: Routes = [
         pathMatch: 'full',
         component: ErrorComponent
       }
-    ]
+    ].map(route => {
+      return {
+        ...route,
+        canActivate: [RoutingAuthorizationCanActivate]
+      };
+    })
```
burzum619: Thanks to @esotrope for the suggestion - here we basically want only the `children` of the home page to enforce authorization - at the top level, where `access-denied` and `sandbox-login` pages live, we do not need to authorize/authenticate.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
esotrope: i like the idea of the i18n being public actually. technically only test results and user-entered data should be protected anyway i think 👍

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -85,6 +85,12 @@ export class AppComponent {
     this.navbarOpen = !this.navbarOpen;
   }
 
+  isAnonymousUser(): boolean {
+    return (
+      !this.user || (this.user.firstName == null && this.user.lastName == null)
```
esotrope: a finishing touch might be to put a flag on our user context object for this to avoid duplicating this logic anywhere

webapp/src/main/webapp/src/app/app.routes.ts
esotrope: 👍 cool glad this worked

webapp/src/main/webapp/src/app/sandbox/sandbox-login.component.html
```diff
@@ -0,0 +1,53 @@
+<div class="row" style="display: flex;">
```
esotrope: nit: inline style

burzum619: good catch, meant to move that out to the less file. Thanks.

webapp/src/main/webapp/src/app/sandbox/sandbox-login.component.ts
```diff
@@ -0,0 +1,49 @@
...
+      username: [null, Validators.required],
+      role: [{ value: null, disabled: true }, Validators.required]
+    });
+
+    this.service.getAll().subscribe(sandboxes => (this.sandboxes = sandboxes));
+  }
+
+  sandboxSelected(selectedSandboxKey: string): void {
+    const roleControl = this.form.get('role');
+    if (!selectedSandboxKey || selectedSandboxKey === 'null') {
```
esotrope: is the idea that they can't select a role until they select a sandbox? do we have different roles per sandbox? if not i would think it is ok to let them select a role before a sandbox if they want

burzum619: @malaffoon  might have more context - but my understanding is that the roles are going to be configured per-sandbox. It's not just a generic list of `District, School Admin, Teacher` roles. Check the mock data I am returning in the service for more context.

burzum619: also, just realized I can remove the `selectedSandboxKey === 'null'` - I was seeing some strange behavior with `null` being converted to `"null"` - but I've since fixed that issue.

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/Users.java
```diff
@@ -13,6 +14,13 @@ public static User getCurrentUser() {
         if (authentication == null) {
             throw new IllegalStateException("user not authenticated");
         }
+        if (ANONYMOUS_USER.equals(authentication.getPrincipal())) {
+            return User.builderExt()
+                    .id(ANONYMOUS_USER)
+                    .username(ANONYMOUS_USER)
+                    .password(ANONYMOUS_USER)
+                    .build();
+        }
```
esotrope: ideally i think we'll want the "anonymous" use object already in the security context by here - can you leave a todo to that effect?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
malaffoon: I think we need to talk about this:
 * I don't think it is okay to allow settings and subjects to be accessed by all. If nothing else, they are tenant-specific. Unless you have a tenant, you shouldn't have any settings or subjects. I guess you could have default settings but that doesn't make sense since they'll be mostly wrong.
* Translations are also tenant-specific. As long as the UI re-fetches translations once credentials are established, i guess exposing them is okay.

It feels like we need to partition the anonymous translations and settings, returning only a subset that make sense. 

I'm okay with this, for now. But we need to revisit it before delivery.

webapp/src/main/webapp/src/app/app.component.ts
malaffoon: Yeah, not thrilled with checking names as a decision factor for an anonymous user. A flag would be better.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
esotrope: great points let's discuss for sure

PR #1961, opened 2019-06-03
Defect/434 null results exception not wrapped

esotrope:


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/PipelineTestResult.java
```diff
@@ -2,8 +2,10 @@
 
 import org.opentestsystem.rdw.script.ScriptError;
 
+import java.util.Collections;
```
burzum619: unused import maybe?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineService.java
```diff
@@ -210,15 +210,16 @@ private PipelineTestRun run(final PipelineTest test, final List<PipelineScript>
...
-                    error
-            )))
-                    .build();
+            return run.result(PipelineTestResult.failedWithScriptError(
+                    new ScriptError(
+                            exception.getMessage(),
+                            exception.getLineNumber()
+                    )
+            )).build();
+        } catch (final RuntimeException exception) {
```
esotrope: @gregcharles should it be the pipeline's job to wrap these as `ScriptRuntimeExceptions` or should i be catching them here? wasn't sure so just did this.

admin-service/src/main/resources/pipeline-script-samples/csv.groovy
```diff
@@ -1 +1,7 @@
-// csv sample script goes here
+// enables csv utilities
+enable 'csv'
+
+// add your transformations here
+
+// outputs transformed csv
+outputCsv
```
esotrope: better csv sample - thanks @gregcharles

webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
```diff
@@ -51,7 +51,7 @@ const defaultCompileDebounceTime = 2000;
 function compilationErrorToMessage(value: ScriptError): Message {
   return {
     type: <MessageType>'error',
-    row: value.row != null ? value.row - 1 : undefined,
+    row: value.row != null ? value.row - 1 : 0,
```
esotrope: default to show errors without rows on the first row

webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipelines/pipelines.component.html
```diff
@@ -1,9 +1,14 @@
...
+    <ng-container *ngFor="let pipeline of _pipelines">
+      <pipeline-card
+        [pipeline]="pipeline"
+        [routerLink]="[pipeline.id]"
+      ></pipeline-card>
+    </ng-container>
+  </div>
+</ng-container>
+<ng-template #loading>
+  <spinner></spinner>
```
esotrope: added loading spinner so the page isn't just blank for a second

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/PipelineTestResult.java
esotrope: sure is : j

PR #1962, opened 2019-06-03
Adding sandbox login redirect based on query parameters

burzum619:


webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -444,7 +444,7 @@ export const routes: Routes = [
     component: AccessDeniedComponent
   },
   {
-    path: 'sandbox-login',
+    path: 'sandbox',
```
esotrope: i actually think sandbox-login is most appropriate as you are not viewing a sandbox on that page

webapp/src/main/webapp/src/app/sandbox/sandbox-login.component.ts
```diff
@@ -43,7 +49,25 @@ export class SandboxLoginComponent implements OnInit {
...
+      this.form.value.username,
+      this.form.value.role
+    );
+  }
+
+  private redirect(queryParams: any): void {
+    // We only want to redirect if all three required params are defined, otherwise stay on the login page
+    if (queryParams.key && queryParams.username && queryParams.role) {
+      this.login(queryParams.key, queryParams.username, queryParams.role);
+    }
```
esotrope: we should probably patch the form with the data too so they could partially fill out with a url maybe

webapp/src/main/webapp/src/app/sandbox/sandbox-login.component.ts
```diff
@@ -43,7 +49,25 @@ export class SandboxLoginComponent implements OnInit {
...
+  }
+
+  private login(key: string, username: string, role: string) {
+    this.service.login(key, username, role).subscribe(
```
esotrope: i'd pass the form value here instead of each param individually

webapp/src/main/webapp/src/app/sandbox/sandbox-login.component.ts
```diff
@@ -14,16 +16,36 @@ export class SandboxLoginComponent implements OnInit {
...
+        this.form.get('sandboxKey').setValue(queryParams.key);
+      }
+      if (queryParams.username) {
+        this.form.get('username').setValue(queryParams.username);
+      }
+      if (queryParams.role) {
+        const roleControl = this.form.get('role');
+        roleControl.enable();
+        roleControl.setValue(queryParams.role);
+      }
```
esotrope: i want to say you can just replace the above code with `this.form.patchValue(queryParams)` and it will even invoke the change handler for the tenant key. Then you can check form.valid before logging in instead of calling out each key too maybe

burzum619: good suggestion!

PR #1965, opened 2019-06-04
Removing the need to call /subjects and /settings for anonymous users

burzum619:
Removing the need to call `/settings` and `/subjects` endpoints when an anonymous user is detected.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/user/UserContext.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/user/UserContext.java
@@ -34,6 +34,10 @@ class UserContext {
         return permissions;
     }
 
+    public boolean isAnonymous() {
+        return firstName == null && lastName == null && permissions.isEmpty();
+    }
+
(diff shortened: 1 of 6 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/user/UserContext.java
```diff
@@ -34,6 +34,10 @@ public String getLastName() {
         return permissions;
     }
 
+    public boolean isAnonymous() {
+        return firstName == null && lastName == null && permissions.isEmpty();
```
esotrope: i think we should just have a boolean flag and just set it when we are creating an anon

webapp/src/main/webapp/src/app/user/user.service.ts
```diff
@@ -21,7 +22,8 @@ export class UserService {
           <User>{
```
esotrope: can you replace `<User>{}` with `({...})` for me while you're in there?

PR #1966, opened 2019-06-05
reverts route chagnes to support prod build

esotrope:


webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -431,12 +432,7 @@ export const routes: Routes = [
         pathMatch: 'full',
         component: ErrorComponent
       }
-    ].map(route => {
-      return {
-        ...route,
-        canActivate: [RoutingAuthorizationCanActivate]
-      };
-    })
```
esotrope: i even tried exporting this lambda as a named method to be compiler friendly but no go - sad

PR #1968, opened 2019-06-05
Adding permission checks to sandbox/tenant admin pages

burzum619:


webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/sandbox-details.component.html
```diff
@@ -17,6 +17,7 @@ <h2>
     <!-- Actions -->
     <div class="mb-sm pull-right">
       <button
+        *hasPermission="'TENANT_WRITE'"
```
malaffoon: why are there single quotes in the double quotes?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-details.component.html
```diff
@@ -27,6 +28,7 @@ <h2>
       </button>
       <p-menu #menu [popup]="true" [model]="menuItems"></p-menu>
       <button
+        *hasPermission="'TENANT_WRITE'"
```
malaffoon: This is just me being ignorant but why are we checking the permission for some of these and checking the readonly flag for others?

burzum619: The `*hasPermission` directive can not be negated directly, nor is there any sort of "else" mechanism. In this case, we just want to display the element if the permission is found. In the other cases, want to display a different element (a read-only span) if the permission is not found

burzum619: With that said, `*hasPermission = 'TENANT_WRITE'` in this case is equal to `!readonly`, so we could use the `readonly` flag exclusively if you prefer.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/sandbox-details.component.html
burzum619: If single quotes are not included, the `*hasPermission` directive will treat `TENANT_WRITE` as a variable rather than a string literal.

PR #1970, opened 2019-06-05
Task/security cleanup

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -25,12 +40,25 @@
 @ConditionalOnProperty(name = "app.test-mode", havingValue = "false", matchIfMissing = true)
 public class WebSecurityConfiguration extends SamlWebSecurityConfiguration {
 
+    @Autowired
+    private Set<AbstractEndpoint> actuatorEndpoints;
```
esotrope: moved more app-specific config out of the `SamlWebSecurityConfiguration` into here

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -40,7 +68,7 @@ protected void configure(final AuthenticationManagerBuilder auth) {
     @Override
     public AuthenticationSuccessHandler authenticationSuccessHandler() {
         final SavedRequestAwareAuthenticationSuccessHandler handler = new SavedRequestAwareAuthenticationSuccessHandler();
-        handler.setDefaultTargetUrl("/landing");
+        handler.setDefaultTargetUrl("/");
```
esotrope: not sure why we had this configured this way but it still works because it never actually is invoked?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -51,15 +79,52 @@ public AuthenticationFailureHandler authenticationFailureHandler() {
...
+                Stream.of(
+                        "/",
+                        "/landing",
+                        "/error",
+                        "/error/**",
+                        "/api/user"
+                ),
+                actuatorEndpoints.stream()
+                        .map(endpoint -> "/" + endpoint.getId() + "/**")
+        ).forEach(path -> authorizeRequests.antMatchers(path).permitAll());
```
esotrope: before `/landing` was not actually exposed but you can still get there because we have a controller that redirects you on logout

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -51,15 +79,52 @@ public AuthenticationFailureHandler authenticationFailureHandler() {
...
+
+        // protect all other paths and enforce authentication on them
+        http.authorizeRequests()
+                .anyRequest().authenticated();
+
+        http.logout()
+                .logoutSuccessUrl("/landing");
```
esotrope: we may want to add a delegating strategy here to logout sandbox users back to the sandbox login page and saml users back to the normal landing page

webapp/src/main/webapp/src/app/error/access-denied/access-denied.component.ts
```diff
@@ -6,7 +6,7 @@ import { ApplicationSettingsService } from '../../app-settings.service';
   templateUrl: './access-denied.component.html'
 })
 export class AccessDeniedComponent implements OnInit {
-  loading: boolean = true;
+  loading: boolean = false;
```
esotrope: @malaffoon just let me know if/when i have permission to axe the access-denied-url

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -51,15 +79,52 @@ public AuthenticationFailureHandler authenticationFailureHandler() {
...
 
+
+    @Override
+    public void configure(final WebSecurity web) {
+        web.ignoring().antMatchers("/*.ico", "/assets/public/**");
+    }
+
     @Override
     protected void configure(final HttpSecurity http) throws Exception {
+        super.configure(http);
```
esotrope: calling super first just seems right when extending something

burzum619: doesn't the compiler force you to?

esotrope: i think that only applies in constructors

esotrope:
any merge objections?
@malaffoon i plan on adding the UUID thing in a followup.

PR #1973, opened 2019-06-06
fixes issue with target reports not showing subject options

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/query-forms/base-aggregate-query-form.component.ts
```diff
@@ -108,7 +109,7 @@ export abstract class BaseAggregateQueryFormComponent
     const { options } = route.snapshot.data;
     this.aggregateReportOptions = options;
     this.originalOptions = optionMapper.map(this.aggregateReportOptions);
-    this.filteredOptions = { ...this.originalOptions };
+    this.filteredOptions = cloneDeep(this.originalOptions);
```
esotrope: wasn't the cause of the bug but could have been a future bug. i gotchu future devs ; p

webapp/src/main/webapp/src/app/aggregate-report/query-forms/target-report-form.component.ts
```diff
@@ -104,7 +104,7 @@ export class TargetReportFormComponent extends BaseAggregateQueryFormComponent {
 
   initialize(): void {
     this.settings.reportType = 'Target';
-    this.hasTargetEnabledSubjects = this.filteredOptions.subjects.some(
+    this.hasTargetEnabledSubjects = this.originalOptions.subjects.some(
```
esotrope: i think it took me an hour or more to track just this down with the debugger and understand the problem smh

PR #1974, opened 2019-06-06
RP-445 Create SandboxLogin service - initially with mock data

vgonzalez9:
Created a webapp sandboxService with the endpoint to return a collection of Sandbox objects that contain each sandboxes Roles.
Included TenantProperties - as is used to retrieve the list of sandboxes (once the common Tenant obj is updated)
Updated with PR changes.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
@@ -1,0 +1,109 @@
+package org.opentestsystem.rdw.reporting.sandbox;
+
+import org.opentestsystem.rdw.multitenant.TenantProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.HashSet;
+import java.util.Set;
+
+@Service
+public class DefaultSandboxService implements SandboxService {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultSandboxService.class);
+
+    private TenantProperties tenantProperties;
+
+    @Autowired
+    public DefaultSandboxService(TenantProperties tenantProperties) {
+        logger.info("Found {} tenants.", tenantProperties.getTenants().size());
+        this.tenantProperties = tenantProperties;
+    }
+
+    @Override
+    public Set<Sandbox> getSandboxes() {
+        return getMockSandboxSet();
+    }
+
+    /**
+     * @return temp collection of sandboxes that is required for sandbox login
+     */
+    public static Set<Sandbox> getMockSandboxSet() {
+        Set<Role> caRoles = new HashSet<>();
+        Role caRole_thg4 = Role.builder()
+                .id("teacher_harborside_g4")
+                .label("Teacher - Harborside - Grade 4")
+                .build();
+        caRoles.add(caRole_thg4);
+        Role caRole_thg5 = Role.builder()
+                .id("teacher_harborside_g5")
+                .label("Teacher - Harborside - Grade 5")
+                .build();
+        caRoles.add(caRole_thg5);
+        Role caRole_thg6 = Role.builder()
+                .id("teacher_harborside_g5")
+                .label("Teacher - Harborside - Grade 6")
+                .build();
+        caRoles.add(caRole_thg6);
+        Role caRole_da_sw = Role.builder()
+                .id("district_admin_sweetwater")
+                .label("District Administrator - Sweetwater")
+                .build();
+        caRoles.add(caRole_da_sw);
+        Role caRole_da_h = Role.builder()
+                .id("district_admin_harborside")
+                .label("District Administrator - Harborside")
+                .build();
+        caRoles.add(caRole_da_h);
+
+        Set<Sandbox> sandboxes = new HashSet<>();
+        Sandbox sandboxCa = Sandbox.builder()
+                .key("ca_19091")
+                .label("California Sandbox")
+                .roles(caRoles)
+                .build();
+        sandboxes.add(sandboxCa);
+
+
+        Set<Role> miRoles = new HashSet<>();
+        Role miRole_tg6 = Role.builder()
+                .id("teacher_rosebank_g6")
+                .label("Teacher - Rosebank - Grade 6")
+                .build();
+        miRoles.add(miRole_tg6);
+        Role miRole_tg5 = Role.builder()
+                .id("teacher_rosebank_g7")
+                .label("Teacher - Rosebank - Grade 7")
+                .build();
+        miRoles.add(miRole_tg5);
+        Role miRole_tg8 = Role.builder()
+                .id("teacher_rosebank_g8")
+                .label("Teacher - Rosebank - Grade 8")
+                .build();
+        miRoles.add(miRole_tg8);
+        Role miRole_da_d = Role.builder()
+                .id("district_admin_detroit")
+                .label("District Administrator - Detroit")
+                .build();
+        miRoles.add(miRole_da_d);
+        Role miRole_da_l = Role.builder()
+                .id("district_admin_landsing")
+                .label("District Administrator - Landsing")
+                .build();
+        miRoles.add(miRole_da_l);
+        Role miRole_da_gr = Role.builder()
+                .id("school_admin_grand_rapids")
+                .label("District Administrator -  Grand Rapids")
+                .build();
+        miRoles.add(miRole_da_gr);
+        Sandbox sandboxMi = Sandbox.builder()
+                .key("mi_42582")
+                .label("Michigan Sandbox")
+                .roles(miRoles)
+                .build();
+        sandboxes.add(sandboxMi);
+
+        return sandboxes;
+    }
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
@@ -13,7 +13,7 @@ import java.util.Set;
 public class DefaultSandboxService implements SandboxService {
     private static final Logger logger = LoggerFactory.getLogger(DefaultSandboxService.class);
 
-    private TenantProperties tenantProperties;
+    private final TenantProperties tenantProperties;
 
     @Autowired
     public DefaultSandboxService(TenantProperties tenantProperties) {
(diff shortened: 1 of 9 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Set;
+
+@RestController
+@RequestMapping("/api/sandbox-login")
```
ftdholbrook: I know no one came up with a good name for this.  Can we call it `sandbox-metadata` or something?

vgonzalez9: we could just have `/sandbox `

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
```diff
@@ -0,0 +1,109 @@
...
+                .label("District Administrator - Sweetwater")
+                .build();
+        caRoles.add(caRole_da_sw);
+        Role caRole_da_h = Role.builder()
+                .id("district_admin_harborside")
+                .label("District Administrator - Harborside")
+                .build();
+        caRoles.add(caRole_da_h);
+
+        Set<Sandbox> sandboxes = new HashSet<>();
```
ftdholbrook: I know we are just stubbing this out, but can we go ahead and grab the actually defined sandboxes?  If non are defined can we define a few in rdw_config?

vgonzalez9: I don't have the git save working for sandboxes - so they are still just in memory
I was going to update once that is in place.

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
ftdholbrook: it happens to be used by the sandbox login page, but it has nothing directly o do with actually logging in.  I would prefer to avoid the word login in the api.

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/Sandbox.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+        public Builder roles(final Set<Role> roles) {
+            this.roles = roles;
+            return this;
+        }
+        Sandbox build() {
+            Sandbox sandbox = new Sandbox();
+            sandbox.label = label;
+            sandbox.key = key;
+            sandbox.setRoles(roles);
```
esotrope: should be `sandbox.roles = safeImmutableOf(roles);`

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/Sandbox.java
```diff
@@ -0,0 +1,65 @@
...
+        this.label = label;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public void setKey(final String key) {
+        this.key = key;
+    }
```
esotrope: we should remove all the setters on these models so they are immutable

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/Sandbox.java
```diff
@@ -0,0 +1,65 @@
...
+        public Builder label(final String label) {
+            this.label = label;
+            return this;
+        }
+        public Builder key(final String key) {
+            this.key = key;
+            return this;
+        }
+        public Builder roles(final Set<Role> roles) {
+            this.roles = roles;
```
esotrope: should make an immutable copy at "set-time" `this.roles = roles != null ? copyOf(roles) : of()

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
esotrope: `/api/sandboxes` would be my vote and would fit in with the existing pattern

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private SandboxService service;
+
+    public SandboxController(SandboxService service) {
+        this.service = service;
+    }
+
+    @GetMapping
+    @ResponseStatus(value = HttpStatus.OK)
+    @ResponseBody
```
esotrope: since it is a `@RestController` the below annotations are inferred defaults and don't need to be specified:
```
@ResponseStatus(value = HttpStatus.OK)
@ResponseBody
```

webapp/src/test/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxTest.java
```diff
@@ -0,0 +1,88 @@
...
+        Optional<Role> caRole = getRoleFromKey(caSandbox.get().getRoles(), "teacher_harborside_g4");
+        assertThat(caRole.isPresent()).isTrue();
+        assertThat(caRole.get().getId()).isEqualTo(expectedCaRole.get().getId());
+        assertThat(caRole.get().getLabel()).isEqualTo(expectedCaRole.get().getLabel());
+
+        Optional<Sandbox> miSandbox = getSandboxFromKey(sandboxes, "mi_42582");
+        assertThat(miSandbox.isPresent()).isTrue();
+        assertThat(miSandbox.get().getKey()).isEqualTo(expectedMiSandbox.get().getKey());
+        assertThat(miSandbox.get().getLabel()).isEqualTo(expectedMiSandbox.get().getLabel());
+        assertThat(miSandbox.get().getRoles().size()).isEqualTo(expectedMiSandbox.get().getRoles().size());
```
esotrope: don't forget about `isEqualComparingFieldByFieldRecursively()`

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
vgonzalez9: /api/sandboxes  it is.

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
```diff
@@ -0,0 +1,109 @@
...
+import org.springframework.stereotype.Service;
+
+import java.util.HashSet;
+import java.util.Set;
+
+@Service
+public class DefaultSandboxService implements SandboxService {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultSandboxService.class);
+
+    private TenantProperties tenantProperties;
```
malaffoon: final

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/Role.java
```diff
@@ -0,0 +1,46 @@
+package org.opentestsystem.rdw.reporting.sandbox;
+
+public class Role {
```
malaffoon: javadocs please, especially for something so generic  ;)

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/Sandbox.java
malaffoon: 👍

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Set;
+
+@RestController
+@RequestMapping("/api/sandbox-login")
+public class SandboxController {
+
+    private SandboxService service;
```
malaffoon: final

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public SandboxController(SandboxService service) {
```
malaffoon: i know it's not technically necessary but can you put `@Autowired` on this (and, yes, i see there are other controllers that deserve the same treatment).

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
esotrope: is there a language that makes things `final` by default? and variable only when specified?

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
```diff
@@ -0,0 +1,29 @@
...
+        this.service = service;
+    }
+
+    @GetMapping
+    @ResponseStatus(value = HttpStatus.OK)
+    @ResponseBody
+    Set<Sandbox> getSandboxes() {
```
malaffoon: public

webapp/src/test/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void setup() {
+        tenantProperties = mock(TenantProperties.class);
+        service = new DefaultSandboxService(tenantProperties);
+    }
+
+    @Test
+    public void itShouldGetSandboxes() {
+        Set<Sandbox> expectedSandboxSet = DefaultSandboxService.getMockSandboxSet();
+        Optional<Sandbox> expectedCaSandbox = getSandboxFromKey(expectedSandboxSet, "ca_19091");
+        assertThat(expectedCaSandbox.isPresent()).isTrue();
```
malaffoon: i'm not convinced these asserts are necessary - they are just checking that the test data is set up right?

webapp/src/test/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.opentestsystem.rdw.multitenant.TenantProperties;
+
+import java.util.Optional;
+import java.util.Set;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import static org.mockito.Mockito.mock;
+
+
+public class DefaultSandboxTest {
```
malaffoon: nit: `DefaultSandboxServiceTest`

webapp/src/test/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxTest.java
malaffoon: Also, i wouldn't work too hard asserting that the service is returning the mock data. Once there is a real repository in there, then the assertions will be more effective. Really the only thing you have to make sure with the service is that it is calling the (yet to be created) repository under the covers.

webapp/src/test/java/org/opentestsystem/rdw/reporting/sandbox/SandboxControllerTestIT.java
```diff
@@ -0,0 +1,43 @@
...
+    private SandboxService service;
+
+    @Test
+    public void itShouldReturnSandboxData() throws  Exception{
+        Set<Sandbox> sandboxes = DefaultSandboxService.getMockSandboxSet();
+        when(service.getSandboxes()).thenReturn(sandboxes);
+
+        mvc.perform(get("/api/sandbox-login")
+                .contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk());
```
malaffoon: Might want to spot-check a few things in the response payload.

webapp/src/test/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxTest.java
vgonzalez9: yep. can remove

PR #1976, opened 2019-06-06
Feature/446 dynamic logout success url

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/DefaultLogoutSuccessHandler.java
```diff
@@ -0,0 +1,47 @@
...
+        // gracefully handles logout when the user is not authenticated
+        // either because they never logged in or their session expired
+        if (authentication == null || !(authentication.getPrincipal() instanceof User)) {
+            response.sendRedirect(samlLogoutSuccessUrl);
+            return;
+        }
+
+        final User user = (User) authentication.getPrincipal();
+        response.sendRedirect(
+                user.getTenant().isSandbox()
```
esotrope: user object now carries a bit more tenant metadata so we can easily find out if they are in a sandbox or not.
further down the road i think we would directly store a logout URL in the user though? but that raises the question of where do we send a user when their session expires - we will not know their tenant... do we send them to a session expiry page?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -30,6 +32,10 @@
 import java.util.Set;
 import java.util.stream.Stream;
 
+/**
+ * TODO SAML logout success URL and access denied URL should be tenant-specific with local fallbacks
+ * to do this we will have to infer the tenant from the tenancy chain even for non-reporting roles
+ */
```
esotrope: @malaffoon added a todo for what it's worth

webapp/src/main/webapp/src/app/shared/security/routing-authorization.can-activate.ts
```diff
@@ -32,7 +32,9 @@ export class RoutingAuthorizationCanActivate implements CanActivate {
       map(permissions => {
         const hasPermission: boolean = (permissions || []).length > 0;
         if (!hasPermission) {
-          this.router.navigate([this.accessDeniedRoute]);
+          this.router.navigate([this.accessDeniedRoute], {
+            skipLocationChange: true
+          });
```
esotrope: now when the UI routes you to access denied it does not change the address bar - always found that super annoying because you cant just hit refresh

webapp/src/main/webapp/src/app/error/access-denied/access-denied.component.ts
```diff
@@ -1,23 +1,8 @@
...
-  ngOnInit() {
-    this.loading = true;
-    this.service.getSettings().subscribe(settings => {
-      if (settings.accessDeniedUrl != null) {
-        window.location.href = settings.accessDeniedUrl;
-      } else {
-        this.loading = false;
-      }
-    });
-  }
```
esotrope: access denied page will now display and not error out

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlWebSecurityConfiguration.java
```diff
@@ -357,14 +350,14 @@ public SecurityContextLogoutHandler logoutHandler() {
...
-        return new SAMLLogoutProcessingFilter(successLogoutHandler(), logoutHandler());
+        return new SAMLLogoutProcessingFilter(logoutSuccessHandler(), logoutHandler());
     }
 
     // Overrides default logout processing filter with the one processing SAML
     // messages
     @Bean
     public SAMLLogoutFilter samlLogoutFilter() {
-        return new SAMLLogoutFilter(successLogoutHandler(),
+        return new SAMLLogoutFilter(logoutSuccessHandler(),
```
esotrope: now i remember why the saml config is annoying and cant be imported. it needs to refer to app-level configured components so it would be circular to import it. i guess the best we could do is move declarations like this to the app-level config and keep things that are non-circular in the saml config file..

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -65,6 +71,15 @@ protected void configure(final AuthenticationManagerBuilder auth) {
         auth.authenticationProvider(sandboxAuthenticationProvider);
     }
 
+    @Override
+    @Bean
+    public LogoutSuccessHandler logoutSuccessHandler() {
+        return new DefaultLogoutSuccessHandler(
+                "/landing",
```
esotrope: logout success no longer goes to "/" (which redirects you to the IDP login) but instead sends you the the "/landing" page (login page)

common/src/main/java/org/opentestsystem/rdw/reporting/common/security/UserTenant.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Represents some tenant metadata of interest for security operations
+ *
+ * TODO possibly store more tenant/sandbox metadata in here?
+ */
+public class UserTenant {
+
+    private String id;
+    private boolean sandbox;
```
malaffoon: i think it would be good to default this to `true`? If we accidentally get a production tenant id and default this to `false` won't that risk allowing access when we shouldn't?

PR #1980, opened 2019-06-08
RP-455 - Add Validation to Tenant creation

vgonzalez9:
Closing this PR - These changes have been pulled into the following PR
https://github.com/SmarterApp/RDW_Reporting/pull/1986
Added validation to check if databases for each dataSource exists
Added validation to check if the archive prefix exists
Added those validations to the tenant creation
fix the TenantConfigPropsService tests to verify the methods being called
updated stub/mocks class for testing
A deeper system check with currently running/configured database and & s3 it needed


admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantConfigPropsService.java
```diff
@@ -51,6 +51,51 @@ public void servicesPropertiesRefresh() {
...
+
+    /**
+     *
+     * @param dataSourceDatabase the database to search for
+     * @param tenantsReadOnlyDataSourceElements Map of dataSources to search
+     * @return true if the dataSource database is found in the list of current tenants
+     */
+    private boolean databaseExists(final String dataSourceDatabase, final Map<String, DataSourceElementsTenant> tenantsReadOnlyDataSourceElements) {
+        for(Map.Entry<String, DataSourceElementsTenant> dataSource : tenantsReadOnlyDataSourceElements.entrySet()) {
+            if(dataSource.getValue().getUrlParts().getDatabase().equals(dataSourceDatabase)) {
```
malaffoon: I don't think this logic is right for OLAP datasources. The `database` isn't likely to be configured differently for different tenants: the `username` is what distinguishes the schema being used.

Yes, the `database` *can* be changed per tenant, but that isn't the likely scenario. This is qa configuration to help with an example ... the general configuration looks like this (just a snippet):
```
datasources:
  olap_ro:
    url-server: rdw-qa.cibkulpjrgtr.us-west-2.redshift.amazonaws.com:5439
    url-db: qa
  reporting_ro:
    url-server: rdw-aurora-qa-cluster.cluster-cugsexobhx8t.us-west-2.rds.amazonaws.com:3306
```
The tenant override looks like this (just a snippet):
```
datasources:
  olap_ro:
    tenants:
      TS:
        username: awsqa_ts
        password: '{cipher}533659f965b5a496f92952f5bef5d843b70e868b3e687acf73ad657c4d3a453f'
  reporting_ro:
    tenants:
      TS:
        urlParts:
          database: ts_reporting
        username: sbac
        password: '{cipher}533659f965b5a496f92952f5bef5d843b70e868b3e687acf73ad657c4d3a453f'
```

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantService.java
```diff
@@ -139,21 +139,63 @@ private String getCurrentTenantKey() {
...
+     * Validates the tenant does not exist, thr
+     * the archive prefix does not exist
+     * the database for each dataSource does not exist
      * @param tenantConfiguration the tenant to validate does not exist
      */
     private void validateNewTenant(final TenantConfiguration tenantConfiguration) {
+        verifyNewTenantExists(tenantConfiguration.getTenant());
+        verifyTenantArchive(tenantConfiguration);
+        verifyTenantDataSources(tenantConfiguration);
+    }
```
malaffoon: So this checks conflicts with other configured tenants. That's good.

@ftdholbrook and i chatted about this briefly and i may be misrepresenting the final approach but ... we also need to check conflicts with actual resources out there. It may be that a tenant has been removed but the underlying resources are still out there. If a new tenant is then created using the same schema, S3 prefix, etc. what are we going to do? Automatically overwriting isn't acceptable. Reusing the existing resources seems wrong as well. So the creation should fail. Where will that be handled?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantConfigPropsService.java
ftdholbrook: so for the MySQL/Aurora data-sources we can create different users, but they are still going to need different databases.  The behavior is different on the Redshift/postgres data sources.   We can discuss further and add some better rules.  @vgonzalez9 is just implementing what I asked her to.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantService.java
ftdholbrook: Yes this is a first pass, the deeper check was blocked by the admin data-source setup I just merged.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantConfigPropsService.java
vgonzalez9: So for the olap dataSource we can add different logic.

ftdholbrook:
I created this ticket https://jira.fairwaytech.com/browse/RP-459 to capture the follow up work needed.  I'd like to merge this @malaffoon rather than keep this branch open and create a new one for the followup work.

vgonzalez9:
Closing this PR as its changes are now part of the Closing this PR - These changes have been pulled into the following PR
#1986

PR #1981, opened 2019-06-08
sandbox dataset load

ftdholbrook:
- depends on a upstream change in RDW_Schma version will need to be bumped before merging


admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcDatasetImport.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param datasetName    The name of the dataset
+     * @param importDatabase The name of the database being imported from.
+     * @return List of each item in the manifest (table_name.txt)
+     */
+    @VisibleForTesting
+    List<String> manifest(String datasetName, String importDatabase) {
+        try {
+            final String bucketName = archivePropertiesRoot.getUriRoot().substring("s3://".length());
+            logger.debug("bucketName: {}", bucketName);
+            final String key = "sandbox-datasets/" + datasetName + "/" + importDatabase + "/manifest.txt";
```
malaffoon: a "location strategy" or helper method would make this consistent since it is used in more than one place

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcDatasetImport.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Value("${sql.datasetImport.tableEmptyProbe}")
+    private String tableEmptyProbe;
+
+    @Autowired
+    public JdbcDatasetImport(@Qualifier("adminJdbcTemplate") final NamedParameterJdbcTemplate adminJdbcTemplate,
+                             final ArchivePropertiesRoot archivePropertiesRoot) {
+        this.adminJdbcTemplate = adminJdbcTemplate;
+        this.archivePropertiesRoot = archivePropertiesRoot;
+        this.amazonS3 = AmazonS3ClientBuilder.standard()
```
malaffoon: Guess it isn't worth it with the time we have available but it would be nice to use the archive abstraction so we could test this locally without S3. Maybe a techdebt JIRA.

admin-service/src/main/resources/application.sql.yml
```diff
@@ -425,4 +425,22 @@ sql:
...
-      WHERE id = :id
\ No newline at end of file
+      WHERE id = :id
+
+  datasetImport:
+      auroraS3LoadData: >-
+        LOAD DATA FROM S3 :s3_uri
+        INTO TABLE $[databaseName].$[tableName]
+
+      useDatabase: >-
```
malaffoon: Not sure if it is difficult or not but just having generic `pre`/`post` sql would be nice. That way, if there is some goofy thing we have to do in an environment, we'd have a place to stick it. But there is the tech issue of executing multiple SQL statements with a single JDBC call.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcDatasetImport.java
ftdholbrook: That's my plan, I think just a straight up 2nd version set for the docker profile that uses mysql and local file system will do the trick.

admin-service/src/main/resources/application.sql.yml
ftdholbrook: I'm not sure how I could structure that.  Let me think on it.

PR #1983, opened 2019-06-11
adds maxlength restriction to pipeline test input/output

esotrope:
image

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -1579,14 +1579,14 @@
         "label": "Test Input",
         "error": {
           "required": "*",
-          "maxLength": "(Please use a smaller test input)"
+          "maxLength": "(Please reduce length to {{requiredLength}} characters or less)"
         }
       },
       "output": {
         "label": "Expected Output",
         "error": {
           "required": "*",
-          "maxLength": "(Please use a smaller test output)"
+          "maxLength": "(Please reduce length to {{requiredLength}} characters or less)"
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1578,13 +1578,15 @@
...
+          "required": "*",
+          "maxLength": "(Please use a smaller test input)"
         }
       },
       "output": {
         "label": "Expected Output",
         "error": {
-          "required": "*"
+          "required": "*",
+          "maxLength": "(Please use a smaller test output)"
```
esotrope: this text alright?
maybe "Please reduce the test input size" is better. let me know what you guys think

gregcharles: Maybe explicitly reference the acceptable size? It's likely the users would just paste in a real file, even if the admin guide tells them not to do that. Then they'd have to iterate through reduced sizes until one works. (You could have a comment over TextMaximumLength that warns devs to update the error messages if they change the value.)

esotrope: yeah i was thinking it isn't so helpful since the number is so big (65535) but maybe it is?

gregcharles: If it's not done by parameter, then it'd be just as easy to say 64 kilobytes in the messages.

esotrope: this look ok?
image

PR #1985, opened 2019-06-11
Fixing bug with null config property values

burzum619:
Snuck in a quick regression bug fix for making config properties *not* read only for the "create new tenants" screens

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.html
+++ b/webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.html
@@ -92,6 +92,7 @@
           <property-override-tree-table
             [form]="tenantForm"
             [configurationProperties]="configurationProperties"
+            [readonly]="false"
             propertiesArrayName="configurationProperties"
           >
           </property-override-tree-table>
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.html
```diff
@@ -105,6 +106,7 @@ <h2>{{ 'tenant-config.new-tenant.header' | translate }}</h2>
           <property-override-table
             [form]="tenantForm"
             [configurationProperties]="localizationOverrides"
+            [readonly]="false"
```
esotrope: does it default true?

burzum619: yep. I could flip it if you think that makes more sense?

esotrope: nope - more secure by default is the right move 👍

burzum619: Cool, thats what I figured.

PR #1987, opened 2019-06-12
Tenant configuration defaults based on key

burzum619:
Updating datasource/archive defaults based on tenant key

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
@@ -96,6 +96,31 @@ export class NewTenantConfigurationComponent implements OnInit, AfterViewInit {
     );
   }
 
+  updateConfigProperties() {
+    const updatedProperties = { ...this.configurationProperties };
+    const key = this.tenantForm.get('key').value.toUpperCase();
+    const defaultDataBaseName = `reporting_${key}`;
+    const archivePathPrefixProperty = <ConfigurationProperty>(
+      updatedProperties.archive.find(property => property.key === 'pathPrefix')
+    );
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
```diff
@@ -96,6 +96,31 @@ export class NewTenantConfigurationComponent implements OnInit, AfterViewInit {
     );
   }
 
+  updateConfigProperties() {
+    const updatedProperties = { ...this.configurationProperties };
```
esotrope: deep clone isn't needed because it is flat i'm guessing

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
```diff
@@ -96,6 +96,31 @@ export class NewTenantConfigurationComponent implements OnInit, AfterViewInit {
...
+    const key = this.tenantForm.get('key').value.toUpperCase();
+    const defaultDataBaseName = `reporting_${key}`;
+    const archivePathPrefixProperty = <ConfigurationProperty>(
```
burzum619: In retrospect, it would be  have been a better idea to store a map of `key` -> `ConfigurationProperty` instead of just an array of config properties, so that we didn't have to iterate over each property. However refactoring that might now might be more trouble than it's worth since we're dealing with fairly small lists.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
burzum619: unfortunately its not flat :/ Datasources have an additional layer

PR #1988, opened 2019-06-12
tenant database admin

ftdholbrook:
RP-460 Expand tenant administration functionality


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
```diff
@@ -0,0 +1,131 @@
+package org.opentestsystem.rdw.admin.multitenant.database;
```
ftdholbrook: This has become kind of large.  It may be worth splitting this up between reporting and OLAP in the future, but for no I'm leaving it as one class.

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdminIT.java
```diff
@@ -0,0 +1,260 @@
...
+    public void createOlapSchemaShouldCreateASchema() {
+        jdbcMultitenantDatabaseAdmin.createOlapSchema(targetOlapSchema);
+        boolean isWarehouseCreateTest = jdbcMultitenantDatabaseAdmin.isOlapSchemaExists(targetOlapSchema);
+        assertThat(isWarehouseCreateTest).isTrue();
+    }
+
+    @Test
+    @Ignore
+    //TODO: pg_catalog.pg_user useconfig is type text[] can't find a way to get value form result set looks fine queried by hand
+    public void modifyUserToUseOlapGrants() throws Exception {
```
ftdholbrook: I am struggling dealing with the return type of text[], any result mapping I've tried results in an error.  I will revisit this test later.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
```diff
@@ -0,0 +1,131 @@
...
+
+public interface MultitenantDatabaseAdmin {
+
+    /**
+     * Create Aurora/Mysql Database
+     *
+     * @param targetDatabase database name to create
+     */
+    void createDatabase(String targetDatabase);
```
malaffoon: Is there a distinction between warehouse and reporting?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
```diff
@@ -0,0 +1,131 @@
...
+     */
+    boolean isDatabaseUserExists(String username);
+
+    /**
+     * Aurora/Mysql GRANT ALL on Aurora/Mysql database to user
+     *
+     * @param username       the user to grant
+     * @param targetDatabase database to grant to user
+     */
+    void grantUserToDatabase(String username, String targetDatabase);
```
malaffoon: Can bury all the user-account details into a single method. Do we really need the client of the service to be able to create a user that doesn't have all the required grants? So basically we'd have createDatabase(database) and createUser(database, username, password, readonly).

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
```diff
@@ -0,0 +1,131 @@
...
+     */
+    void grantAllOlapUserToSchemaTables(String schema, String username);
+
+    /**
+     * Set schema search path for Redshift/PostgreSQL users
+     *
+     * @param schema   schema to add to search path
+     * @param username user apply search path to
+     */
+    void searchPathOlapUserToSchema(String schema, String username);
```
malaffoon: ditto about a single user method.

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdminIT.java
```diff
@@ -0,0 +1,260 @@
...
+        List<Array> useconfigResults = adminOlapJdbcTemplate.getJdbcOperations()
+                .query(
+                        "SELECT useconfig " +
+                                "FROM pg_catalog.pg_user " +
+                                "WHERE usename='" + createDatabaseUserUsername + "'",
+                        (row, index) -> row.getArray("useconfig"));
+
+        assertThat(useconfigResults.size()).isEqualTo(1);
+        //TODO: this is always null
+        String[] results = (String[]) useconfigResults.get(0).getArray();
```
malaffoon: Very likely you have to go through a SQL `Array` to get there.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
ftdholbrook: Seems like a lot of added complexity to me.  I'm really focused on solving this particular set of problems right now.  I think we can revisit a more general / flexible solution at a later time.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
ftdholbrook: the distinction is between aurrora/mysql and PostgreSQL/redshift

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
ftdholbrook: I am working on a higher level service for tenant/sandbox creation that orchestrates these lower level calls.

malaffoon: We keep saying that ("we can revisit") but i feel really strongly about this one. SB is notorious about not letting software upgrades go out while allowing configuration changes. We *are* going to get burned by hard-coding this. If we want to punt, let's put in a JIRA and i'll curate them all once we're through the crunch.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantDatabaseAdmin.java
malaffoon: Ok. Not clear to me how the datasource is being switched between reporting and warehouse.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * Repository for loading datasets for new Tenants.  This particular implementation
+ * will only work with AWS Aurora and AWS S3.
+ */
+@Repository
+public class JdbcMultitenantDatabaseAdmin implements MultitenantDatabaseAdmin {
+
+    private static final Logger logger = LoggerFactory.getLogger(JdbcMultitenantDatabaseAdmin.class);
+
+    private final String delimiter = "/";
```
burzum619: nit: this can be static can it not?

PR #1989, opened 2019-06-12
Feature/463 map sandbox roles to reporting permissions

esotrope:
*Adds writing groups to sandbox user login process

still writing tests for this but i figured i would request some early eyes


reporting-service/src/test/java/org/opentestsystem/rdw/reporting/group/UserGroupControllerIT.java
```diff
@@ -123,4 +123,4 @@ public void deleteGroupShouldCallService() throws Exception {
     }
 
 
-}
\ No newline at end of file
+}
```
esotrope: got some artifacts from experimentally moving files i guess

webapp/src/main/java/org/opentestsystem/rdw/reporting/WritableDataSourceConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+import javax.sql.DataSource;
+
+@Configuration
+@EnableTransactionManagement
+public class WritableDataSourceConfiguration {
```
esotrope: copy pasted from reporting-service

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationController.java
```diff
@@ -26,6 +26,7 @@
 
     @RequestMapping("/{locale}")
     public ObjectNode getTranslations(@PathVariable final Locale locale,
+                                      // TODO revisit need for this - i don't think this is actually used
```
esotrope: does anyone know if this is used? i could only find this used in a test case yet we have a bit of code in place to make this possible - found it kind of weird

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxUserDetailsService.java
```diff
@@ -41,39 +45,43 @@ public UserDetails loadUserDetails(final SandboxAuthenticationToken token) throw
...
 
+        TenantContextHolder.clear();
+
         return User.builderExt()
-                // TODO use requested & valid sandbox key
-                //.tenantId(token.getSandboxKey())
                 .tenant(
                         UserTenant.builder()
-                                .id("CA")
+                                .id(token.getSandboxKey())
```
ftdholbrook: really sandbox should have an id and tenant key.. in practice I believe they will be exactly the same, we may want to circle back on that.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxAuthenticationProvider.java
```diff
@@ -23,15 +22,7 @@ public SandboxAuthenticationProvider(
...
-
-        final String role = ((SandboxAuthenticationToken) authentication).getRole();
-        if (sandboxRoleRepository.findById(role) == null) {
-            // The request was tampered with or someone's saved sandbox login link is no longer valid
-            throw new AuthenticationServiceException(String.format(
-                    "sandbox role \"%s\" not found",
-                    role
-            ));
-        }
+        // Security checks are performed by the user details service
```
esotrope: found that this was being checked only after the user was created and we already check it during user creation

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxUserDetailsService.java
```diff
@@ -41,39 +45,43 @@ public UserDetails loadUserDetails(final SandboxAuthenticationToken token) throw
         // TODO allow way to log back in again with same key through extra param?
         final String usernameAndId = UUID.randomUUID().toString();
 
+        TenantContextHolder.setTenantId(token.getSandboxKey());
```
esotrope: this not being placed here is probably a bug in the deployed code

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxUserDetailsService.java
esotrope: this not being the selected tenant in the UI is also a bug currently deployed

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationController.java
ftdholbrook: If it is used it has to be in the UI somewhere.  I add info level "AUDIT: This {endpoint} active" messages, and then you can ping logstash or whatever  to see if it is in use.

webapp/src/main/resources/application.sql.yml
```diff
@@ -2,6 +2,23 @@ sql:
   status:
     readTest: >-
       SELECT code FROM asmt_type
+
+  user-student-group:
+    create:
+      INSERT IGNORE INTO user_student_group
```
esotrope: currently no records are inserted at login because there are no groups matching the provided group_name

webapp/src/main/resources/application.yml
```diff
@@ -123,6 +123,37 @@ datasources:
     removeAbandonedTimeout: 60
     logAbandoned: false
 
+  reporting_rw:
```
esotrope: also copy pasted from reporting-service

webapp/src/main/resources/application.sql.yml
ftdholbrook: Should we add a #TODO: here?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxUserDetailsService.java
esotrope: ah good point. because we need the tenant ID to find the datasource not the key huh

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationController.java
esotrope: when i do a project search for `include-ui` it comes up with only this controller method so i think it is dead code

webapp/src/main/resources/application.sql.yml
esotrope: hm. i think the todo is really just for mark to make sure the groups are in the sandbox data set though. everything else here should align once that is in place

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxUserDetailsService.java
```diff
@@ -26,54 +29,66 @@
...
+        this.sandboxRepository = sandboxRepository;
         this.sandboxRoleRepository = sandboxRoleRepository;
+        this.groupRepository = groupRepository;
     }
 
     @Override
     public UserDetails loadUserDetails(final SandboxAuthenticationToken token) throws UsernameNotFoundException {
 
+        final Tenant sandbox = sandboxRepository.findByKey(token.getSandboxKey());
+        TenantContextHolder.setTenantId(sandbox.getId());
```
esotrope: now setting tenant ID to the sandbox ID instead of the key

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/TenantPropertiesSandboxRepository.java
```diff
@@ -0,0 +1,35 @@
...
+import org.opentestsystem.rdw.multitenant.Tenant;
+import org.opentestsystem.rdw.multitenant.TenantProperties;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+@Repository
+class TenantPropertiesSandboxRepository implements SandboxRepository {
```
esotrope: its thin but i figured this should have its own repo layer

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/TenantPropertiesSandboxRepository.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    public List<Tenant> findAll() {
+        return tenantProperties.getTenants().values().stream()
+                .filter(Tenant::isSandbox)
+                .collect(toList());
+    }
+
+    @Override
+    public Tenant findByKey(final String key) {
+        return tenantProperties.getTenants().values().stream()
```
esotrope: @ftdholbrook just curious - what is the reason for storing the tenants in a map and not a list?

ftdholbrook: It is so we can split the tenant configuration across multiple [tenantKey]/application.yml files.

esotrope: without overwriting another i guess? still fizzy for me but looks like you have it figured out -thanks : j

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationController.java
malaffoon: We use this to generate the translation package to send off to the translation agency.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxUserDetailsService.java
```diff
@@ -26,54 +29,66 @@
...
-            // TODO insert group memberships
+            // This needs to match data generator output to correctly associate the assigned group
+            final String conventionalGroupName = String.format(
+                    "%s (%s)",
+                    role.getOrganization().getName(),
+                    role.getGradeCode()
+            );
+            groupRepository.create(
+                    usernameAndId,
+                    conventionalGroupName
```
malaffoon: I'll have to see the rest but this is a tad misleading, since it isn't creating a group, only associating the user with a group.

webapp/src/main/resources/application.yml
malaffoon: I won't complain too much since you've done the work but do we really need separate _ro and _rw datasources? They are logically completely intertwined in this sandbox stuff.

webapp/src/main/java/org/opentestsystem/rdw/reporting/i18n/TranslationController.java
esotrope: Ah. now it makes sense. i feel like this deserves its own endpoint/service with comments to this effect.

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxUserDetailsService.java
esotrope: if i want to name it to the actual table it would be `UserStudentGroupRepository` but we don't follow that convention for the other group related components

webapp/src/main/resources/application.yml
esotrope: yeah i should have clarified this. when i remove the `_ro` config nothing starts up for reasons unknown so i left it to keep things moving : J

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
```diff
@@ -45,14 +39,19 @@ public DefaultSandboxService(final TenantProperties tenantProperties,
                                 .roles(rolesBySandboxKey.get(tenant.getKey()))
                                 .build()
                 )
-                .collect(Collectors.toSet());
+                .sorted(
+                        Comparator
+                                .comparing(Sandbox::getLabel)
+                                .thenComparing(Sandbox::getKey)
+                )
```
esotrope: added sorting of sandboxes by name. normally this would be by entity ID but since we don't have that i went with this

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxController.java
```diff
@@ -20,7 +20,7 @@
     }
 
     @GetMapping
-    public Set<Sandbox> getSandboxes() {
+    public List<Sandbox> getSandboxes() {
```
esotrope: should be ordered

webapp/src/main/java/org/opentestsystem/rdw/reporting/status/DatabaseStatusIndicator.java
```diff
@@ -28,11 +30,11 @@
     @Value("${sql.status.readTest}")
     private String sqlReadTest;
 
-    private final JdbcTemplate jdbcTemplate;
+    private final NamedParameterJdbcTemplate jdbcTemplate;
     private final TenantProperties tenantProperties;
 
     @Autowired
-    DatabaseStatusIndicator(final JdbcTemplate jdbcTemplate,
+    DatabaseStatusIndicator(final NamedParameterJdbcTemplate jdbcTemplate,
```
esotrope: couldn't start the app without changing this - @ftdholbrook said he experienced the same thing. still an ongoing mystery for me

webapp/src/test/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxServiceTest.java
```diff
@@ -1,196 +1,85 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
-import static org.opentestsystem.rdw.reporting.sandbox.SandboxRoleType.SchoolAdministrator;
-import static org.opentestsystem.rdw.reporting.sandbox.SandboxRoleType.Teacher;
 import static org.opentestsystem.rdw.reporting.sandbox.SandboxRoleType.DistrictAdministrator;
+import static org.opentestsystem.rdw.reporting.sandbox.SandboxRoleType.Teacher;
 
-
+@RunWith(MockitoJUnitRunner.class)
```
esotrope: simplified these tests while still providing the same coverage

PR #1990, opened 2019-06-13
removes excessive security whitelist on routes config

esotrope:


webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -48,16 +48,6 @@ import { SandboxLoginComponent } from './sandbox/sandbox-login.component';
...
-  data: {
-    permissions: [
-      'CUSTOM_AGGREGATE_READ',
-      'GROUP_WRITE',
-      'INDIVIDUAL_PII_READ',
-      'INSTRUCTIONAL_RESOURCE_WRITE',
-      'EMBARGO_WRITE'
-    ]
-  },
-  canActivate: [AuthorizationCanActivate],
```
esotrope: this whitelist was missing the new permissions. however we don't even need this whitelist because each route enforces it's own security already.

PR #1991, opened 2019-06-13
fixes language filter

esotrope:


webapp/src/main/webapp/src/app/assessments/filters/adv-filters/adv-filters.component.ts
```diff
@@ -70,14 +68,8 @@ export class AdvFiltersComponent {
     );
   }
 
-  optionsChanged(values: { value: any }[]): void {
-    this.filterBy.languageCodes = values.map(({ value }) => ({
```
esotrope: not sure why this was stored as a map and then queried like an array... i just went with array since it is a set. no real need for O(1) lookup

webapp/src/main/webapp/src/app/assessments/filters/exam-filters/exam-filter.service.ts
```diff
@@ -284,10 +283,7 @@ export class ExamFilterService {
   }
 
   private filterByLanguageCode(exam: Exam, filterValue: any): boolean {
-    return (
-      (exam.student && !filterValue.length) ||
```
esotrope: there is no need for these checks here or anywhere in the other filters...

PR #1993, opened 2019-06-13
RP-461 Tenant Administration - DataSource password handling API

vgonzalez9:
Part one of the changes needed, this will unblock UI dev
New service to expose decrypt API endpoint, this handles and expected encrypted string values created from the configServer's envrypt function/endpoint.  It will also handle removing the {cipher} prefix from the encrpyted string if it exists.
Created ConfigServerClient that will call the config server's enpoint for encrypt and decrypt, monitor and to retrieve the unmasked passwords.
the client pulls inthe uri from the spring.cloud.config.uri endpoint
TODO complete the work for the monitor and unmasked passwords in a TenantConfiguration

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/client/ConfigServerClient.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/client/ConfigServerClient.java
@@ -1,36 +1,37 @@
 package org.opentestsystem.rdw.admin.client;
 
-import org.checkerframework.checker.nullness.Opt;
-
-import java.util.Optional;
-
 /**
  * Client to the system's config Server
  */
 public interface ConfigServerClient {
     /**
-     *  Initiate a refresh of system properties by posting to the /monitor endpoint
-     *  atm refresh all reporting applications
+     * Initiate a refresh of system properties by posting to the /monitor endpoint
+     * atm refresh all reporting applications
+     *
+     * @throws IllegalArgumentException not able to call /monitor endpoint
      */
-    void postMonitor();
+    void postMonitor() throws IllegalArgumentException;
 
     /**
-     * @param service to search through for the dataSource's password
+     * @param service    to search through for the dataSource's password
      * @param dataSource of the password - need to be the property
      * @return the actual encrypted password of the dataSource for the service
+     * @throws IllegalArgumentException not able to to call service endpoint
      */
-    Optional<String> getActualServicePassword(String service, String dataSource);
+    String getActualServicePassword(String service, String dataSource) throws IllegalArgumentException;
 
     /**
      * @param password to encrypt
      * @return the encrypted password
+     * @throws IllegalArgumentException not able to call /encrypt endpoint
      */
-    Optional<String> encryptedPassword(String password);
+    String encryptPassword(String password) throws IllegalArgumentException;
 
     /**
-     * @param password encrypted
+     * @param password password to decrypt
      * @return the decrypted password
+     * @throws IllegalArgumentException not able to call /decrypt endpoint
      */
-    Optional<String> decryptedPassword(String password);
+    String decryptPassword(String password) throws IllegalArgumentException;
 
 }
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/client/ConfigServerClient.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/client/ConfigServerClient.java
@@ -1,0 +1,36 @@
+package org.opentestsystem.rdw.admin.client;
+
+import org.checkerframework.checker.nullness.Opt;
+
+import java.util.Optional;
+
+/**
+ * Client to the system's config Server
+ */
+public interface ConfigServerClient {
+    /**
+     *  Initiate a refresh of system properties by posting to the /monitor endpoint
+     *  atm refresh all reporting applications
+     */
+    void postMonitor();
+
+    /**
+     * @param service to search through for the dataSource's password
+     * @param dataSource of the password - need to be the property
+     * @return the actual encrypted password of the dataSource for the service
+     */
+    Optional<String> getActualServicePassword(String service, String dataSource);
+
+    /**
+     * @param password to encrypt
+     * @return the encrypted password
+     */
+    Optional<String> encryptedPassword(String password);
+
+    /**
+     * @param password encrypted
+     * @return the decrypted password
+     */
+    Optional<String> decryptedPassword(String password);
+
+}
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
@@ -13,12 +13,12 @@ import org.springframework.web.client.RestTemplate;
 import java.util.Optional;
 
 /**
- * Client to retrieve the config server's uri to call the encrypt decrypt and monitor endpoints
+ * Client to retrieve the config server's uri to call the encrypt,decrypt, monitor (refresh) endpoints
+ * And call a service to retrieve the datasource's unmasked password
  */
 public class DefaultConfigServerClient implements ConfigServerClient {
     private static final Logger logger = LoggerFactory.getLogger(DefaultConfigServerClient.class);
 
-    //@Value("${spring.cloud.config.uri}")
     private String configServerUri;
     private RestTemplate restTemplate = new RestTemplate();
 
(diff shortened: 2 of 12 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/client/ConfigServerClient.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param dataSource of the password - need to be the property
+     * @return the actual encrypted password of the dataSource for the service
+     */
+    Optional<String> getActualServicePassword(String service, String dataSource);
+
+    /**
+     * @param password to encrypt
+     * @return the encrypted password
+     */
+    Optional<String> encryptedPassword(String password);
```
malaffoon: nit: verb form would make more sense, i.e. `encryptPassword` ?

nit: why is this returning an optional? When would it not encrypt?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/ConfigServerClient.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * @param password encrypted
+     * @return the decrypted password
+     */
+    Optional<String> decryptedPassword(String password);
```
malaffoon: same questions, although i can see a decrypt failure ... but i'd think that would throw an IAE?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.util.Optional;
+
+/**
+ * Client to retrieve the config server's uri to call the encrypt decrypt and monitor endpoints
+ */
+public class DefaultConfigServerClient implements ConfigServerClient {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultConfigServerClient.class);
+
+    //@Value("${spring.cloud.config.uri}")
```
malaffoon: delete?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public void postMonitor() {
+        try {
+            // todo fix this
+            HttpHeaders headers = getHeaders();
+            HttpEntity<String> request = new HttpEntity<>("path=*", headers);
+            ResponseEntity<String> responseEntity = restTemplate.postForEntity(configServerUri + "/monitor", request, String.class);
+            logger.debug("Successfully called {}/encrypt returned {}.", configServerUri, responseEntity.getStatusCode());
```
malaffoon: message doesn't match, `/encrypt` != `/monitor`

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
```diff
@@ -0,0 +1,95 @@
...
+        } catch (Exception e) {
+            logger.debug("Cannot contact {}/monitor at this time.", configServerUri);
```
malaffoon: include exception in log?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/ConfigServerClient.java
vgonzalez9: Your are correct, it should always encrypt as string - the errors I saw were bash related. 
I'll update to encryptPassword - or so you thing a more generic encrypt is better?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    @Override
+    public Optional<String> getActualServicePassword(final String service, final String dataSource) {
+        // will need to know which service is being called - or I go get all of the passwords
+        //todo work on this call
+        try {
+            ResponseEntity<String> responseEntity = restTemplate.getForEntity(configServerUri + service +"/awsuat", String.class);
+            logger.debug("Successfully called {}/tbd returned {}.", configServerUri, responseEntity.getStatusCode());
```
malaffoon: nit: log messages `tbd` can be replaced

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
vgonzalez9: yes

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
vgonzalez9: will fix

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantConfigService.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * TenantConfigService to expose TenantConfigService features to the UI
+ */
+public interface TenantConfigService {
+    /**
+     * give the ui the ability to decrypt a password so it can be displayed to a Tenant Admin user
+     * @param encrypted password to decrypt
+     * @return the decrypted password
+     */
+    Optional<String> decrypt(String encrypted);
```
malaffoon: The UI doesn't need to encrypt too? If the user types in a new password, we are just going to leave it plain text until they submit?

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
```diff
@@ -0,0 +1,95 @@
...
+            return Optional.of(responseEntity.getBody());
+        } catch (Exception e) {
+            logger.debug("Cannot contact {}/tbd at this time.", configServerUri);
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<String> encryptedPassword(final String encryptedPassword) {
```
malaffoon: nit: param name should not be `encryptedPassword`, probably just `password`

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantConfigController.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import org.opentestsystem.rdw.admin.service.impl.DefaultTenantConfigService;
+
+@RestController
+@RequestMapping("/config")
+public class TenantConfigController {
+    DefaultTenantConfigService service;
```
malaffoon: nit: `private final`

admin-service/src/main/java/org/opentestsystem/rdw/admin/client/DefaultConfigServerClient.java
vgonzalez9: actually that is correct since the call returns the password unencrypted.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantConfigController.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import org.opentestsystem.rdw.admin.service.impl.DefaultTenantConfigService;
+
+@RestController
+@RequestMapping("/config")
+public class TenantConfigController {
```
malaffoon: nit: ~public~ (controllers can be package private scope)

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/TenantConfigServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void setUp() {
+        defaultConfigServerClient = mock(DefaultConfigServerClient.class);
+        service = new DefaultTenantConfigService(defaultConfigServerClient);
+    }
+
+    @Test
+    public void
+    itShouldDecryptSuccessfully() {
+        when(defaultConfigServerClient.decryptedPassword(any())).thenReturn(Optional.of("decryptedResults"));
+        Optional<String> decrypted = service.decrypt("{cypher}BlahBlah");
```
malaffoon: need a more robust test i think since `{cypher}` is a typo

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/TenantConfigServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+        assertThat(decrypted).isPresent();
+        assertThat(decrypted.get()).isEqualTo("decryptedResults");
+        verify(defaultConfigServerClient).decryptedPassword(any());
+    }
+
+    @Test
+    public void
+    itShouldDecryptSuccessfullyWithNoCipher() {
+        when(defaultConfigServerClient.decryptedPassword(any())).thenReturn(Optional.of("decryptedResults"));
```
malaffoon: ditto: make this `when` more specific

admin-service/src/test/resources/application-tenant_client_docker.yml
```diff
@@ -11,4 +11,9 @@ tenant-configuration-lookup-test:
     report-processor:
       url: http://localhost:8086
     reporting-service:
-      url: http://localhost:8108
\ No newline at end of file
+      url: http://localhost:8108
+
+spring:
```
malaffoon: Not checking this in i assume?

All the services already have this define in the bootstrap.yml. To override for local testing, recommend test configuration in IJ or something.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantConfigService.java
vgonzalez9: So the password should always be returned to the UI as encrypted.  The decrypt endpoint is to allow the Admin to peak at the password.  
and if the Admin changes it, the thought was the https was good enough security, had some discussions with Ernie & Dave on this. If it's, I am ok to add an encrypt endpoint for the UI.

admin-service/src/test/resources/application-tenant_client_docker.yml
vgonzalez9: Right, will not check in.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/TenantConfigService.java
malaffoon: no, i'm fine with that, just making sure the discussion was had

ftdholbrook: Yeah we talked about it and that seemed like the most reasonable approach.

PR #1994, opened 2019-06-14
Adding support for datasource passwords

burzum619:
- Adding the ability to decrypt encrypted passwords via UI
- Generating "random" passwords for datasource when creating new tenants


webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
```diff
@@ -73,6 +89,19 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
     return false;
   }
 
+  calculatePadding(override: ConfigurationProperty): string {
```
burzum619: This was added to prevent text in input fields from being covered by reset/decrypt icons. Not the prettiest solution, but it works and is an improvement IMO.

esotrope: i'd be tempted just to style it with 50px always just to remove the complexity

webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -383,3 +383,9 @@ export function flattenJsonObject(ob: any): any {
   }
   return toReturn;
 }
+
+export function generateRandomPassword(): string {
+  return Math.random()
```
esotrope: @malaffoon @ftdholbrook what is our threat model for this feature? `Math.random()` is not considered "random enough" when creating passwords. [here](https://gist.github.com/joepie91/7105003c3b26e65efcea63f3db82dfba) is some basic info on this.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
burzum619: I was going honestly about to go with that originally... but the dynamic padding is pretty and i dont think really has a noticable performance impact. Is there any other reasom besides simplicity to remove this?

burzum619: Because many input fields actually do have values that take up the full width of the input filed, it looks really awkward when you dont have any icons on the right side. Its a little ugly IMO

webapp/src/main/webapp/src/app/shared/support/support.ts
burzum619: Im very open to better suggestions, but ultimately  these are just password "suggestions". I'm happy to utilize a better pattern, but I'd like to keep the password generation on the front end if possible.

malaffoon: Personally, i wouldn't have this at all. They (SmarterBalanced) have a password generation policy. We probably do need a popup that requires whatever redshift requires for that one.

PR #1998, opened 2019-06-16
Tenant admin handler

ftdholbrook:
this iteration only supports sandbox and tenant creation


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/TenantAdministrationRequestHandlerDefault.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.support.MessageBuilder;
+
+import java.util.HashMap;
+
+import static org.opentestsystem.rdw.admin.multitenant.TenantAdministrationProcessor.SandboxCreationRequest;
+import static org.opentestsystem.rdw.admin.multitenant.TenantAdministrationProcessor.TenantCreationRequest;
+
+public class TenantAdministrationRequestHandlerDefault implements TenantAdministrationRequestHandler {
```
malaffoon: nit: i noticed this with the TenantConfigurationPersistenceServiceGit -- putting the implementation at the end of the name instead of the beginning. I can see an argument for it (it'll sort with the interface) but the convention in this project has been to put it at the beginning, i.e. DefaultTenantAdministrationRequestHandler.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/TenantAdministrationRequestHandlerDefault.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(TenantAdministrationRequestHandlerDefault.class);
+
+    private final MessageSecurityService messageSecurityService;
+    private final TenantAdministrationProcessor tenantAdministrationProcessor;
+    private final TenantDatabaseCreationService tenantDatabaseCreationService;
+    private final TenantConfigurationPersistenceService tenantConfigurationPersistenceService;
+    private final ConfigServerClient configServerClient;
+
+    @Autowired
```
malaffoon: nit: this isn't actually autowired since we have our own @Bean. I'm surprised that IJ didn't warn you about having Autowired on a non-Component.

PR #2000, opened 2019-06-17
Better error handling for tenants/sandboxes

burzum619:
Resolves RP-422 & RP-436
- Added ErrorHandlers/interceptors for a `DuplicateTenant` custom exception as well as the generic error thrown by jackson when it is unable to deserialize a configuration property
- For now, errors are returned directly from the backend - no localization for these particular error messages at the moment
- Added some additional UI tweaks requested by Mark
- *Important Before Merge* - Still needs test coverage for the exception handler before this PR can be merged.


webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-details.component.ts
```diff
@@ -101,11 +101,15 @@ export class TenantConfigurationDetailsComponent implements OnInit, OnChanges {
             existing.code === updatedTenant.code ? updatedTenant : existing
           )
         );
+        this.editMode = false;
       },
       error =>
-        this.notificationService.error({ id: 'tenant-config.errors.update' })
+        error.json().message
+          ? this.notificationService.error({ id: error.json().message })
```
burzum619: This is not ideal, but for the sake of time its what I went with. I couldn't figure out a way to pass in a "pre-translated" message into the notification service, it normally expects an i18n id. In this case, this works because the fallback is simply printing the "id" which in this case is the actual message.

admin-service/src/main/java/org/opentestsystem/rdw/admin/exception/DuplicateTenantException.java
```diff
@@ -0,0 +1,14 @@
+package org.opentestsystem.rdw.admin.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+public class DuplicateTenantException extends RuntimeException {
```
esotrope: did we make a jira for client side async validation on this?

burzum619: not yet (at least I did not)

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/sandbox-details.component.ts
```diff
@@ -144,25 +144,19 @@ export class SandboxConfigurationDetailsComponent implements OnInit, OnChanges {
             existing.code === updatedSandbox.code ? updatedSandbox : existing
           )
         );
+        this.editMode = false;
```
esotrope: i'm curious about what this means. this is if you save we change the view state out of edit mode?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-details.component.ts
esotrope: i think i've had to do the same one time before as well.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-sandbox.component.ts
```diff
@@ -114,7 +114,11 @@ export class NewSandboxConfigurationComponent implements OnInit, AfterViewInit {
         this.router.navigate(['sandboxes']);
       },
       error =>
-        this.notificationService.error({ id: 'sandbox-config.errors.create' })
+        error.json().message
+          ? this.notificationService.error({ id: error.json().message })
+          : this.notificationService.error({
+              id: 'sandbox-config.errors.create'
+            })
```
esotrope: looks like we could have some duplicate code extraction here

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/sandbox-details.component.ts
burzum619: Correct, but only the save call was successful. Otherwise, we don't want to kick the user out of edit mode - we want to leave it open and present an error so they can possibly correct it.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantHttpMessageExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+     * This matches all words wrapped in quotes
+     */
+    private static final Pattern QUOTES_REGEX = Pattern.compile("([\"'])(?:(?=(\\\\?))\\2.)*?\\1");
+
+    @Override
+    public ResponseEntity<Object> handleHttpMessageNotReadable(final HttpMessageNotReadableException ex,
+                                                               final HttpHeaders headers,
+                                                               final HttpStatus status,
+                                                               final WebRequest request) {
+        final Matcher matcher = QUOTES_REGEX.matcher(ex.getMessage());
```
malaffoon: Why do all this? If we have an exception for this situation, simply add an extra user-message field to the exception and set it when throwing the exception. Then this becomes a simple handler that extracts the user-message field.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TenantHttpMessageExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+        return ResponseEntity.badRequest().body(
+                new ApiError("An error occurred while processing the request. Invalid input: " + varWithError)
+        );
+    }
+
+    @ExceptionHandler(DuplicateTenantException.class)
+    public ResponseEntity<Object> handleDuplicateTenantException(final DuplicateTenantException ex,
+                                                                 final WebRequest request) {
+        return ResponseEntity.badRequest().body(new ApiError(ex.getMessage()));
```
malaffoon: please add `logger.info(ex.getMessage())`

PR #2003, opened 2019-06-17
update dataset loading

ftdholbrook:
- truncate before loading each table
- fix integration tests by reverting back to @Repository
- some code consistency fixes

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
@@ -8,27 +8,32 @@ import com.amazonaws.services.s3.model.S3Object;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableMap;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
 import org.opentestsystem.rdw.admin.multitenant.database.MultitenantDatabaseAdmin;
 import org.opentestsystem.rdw.archive.ArchivePropertiesRoot;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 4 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
```diff
@@ -155,7 +167,10 @@ public boolean isDatabaseUserExists(String username, RdwDatabaseType databaseTyp
 
     @Override
     public void grantUserToDatabase(String username, String targetDatabase, RdwDatabaseType databaseType) {
-        resolveTemplate(databaseType).update(amendSqlWithPlaceholder(targetDatabase, null, grantUserToDatabase),
+        Map<String, String> replacements = ImmutableMap.<String, String>builder()
+                .put("databaseName", targetDatabase)
+                .build();
```
esotrope: could this be a parameter source? if not you could just do `ImmutableMap.of("databaseName", targetDatabase)` to shorten this

ftdholbrook: I started with that, and was getting type errors (It was inferring <Sting,Object> for some reason.  I gave up and went with this slightly more verbose form.

esotrope: interesting, i usually only have the reverse issue - when i use `ImmutableMap.builder()` i have to pass the pass types like you have it

admin-service/src/main/resources/application.sql.yml
```diff
@@ -428,9 +428,12 @@ sql:
...
 
   multitenantDatabaseAdmin:
+    truncateAuroraTable: >-
+      TRUNCATE TABLE `$[databaseName]`.`$[tableName]`
+
     auroraS3LoadData: >-
-      LOAD DATA FROM S3 :s3_uri
-      INTO TABLE $[databaseName].$[tableName]
+      LOAD DATA FROM S3 '$[s3Uri]'
+      INTO TABLE `$[databaseName]`.`$[tableName]`
```
vgonzalez9: why was the ' needed?

ftdholbrook: not needed I added when trying to debug a table not found error, it is allows for some characters and names that would not otherwise be allowed.

PR #2007, opened 2019-06-18
RP-438: change LEP label to "English Learner"

gregcharles:
Label only change, but was in six places in four files, and I'm not sure where they all appear in the UI. It was updated in everywhere I looked though!


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.mapper.ts
```diff
@@ -155,7 +155,7 @@ export class AggregateReportOptionsMapper {
         limitedEnglishProficiencies: options.studentFilters.limitedEnglishProficiencies.map(
           optionMapper(
             value => translate(`common.strict-boolean.${value}`),
-            value => `Limited English Proficiency: ${value}`
+            value => `English Learner: ${value}`
```
esotrope: i would leave these "value => ..." ones the same - these are generating the codes for analytic tracking.

webapp/src/main/webapp/src/app/aggregate-report/subgroup/subgroup-filters.component.html
```diff
@@ -86,7 +86,7 @@
           analyticsEvent="FilterClick"
           analyticsCategory="AggregateQueryBuilder"
           [allOptionAnalyticsProperties]="{
-            label: 'Limited English Proficiency: All'
+            label: 'English Learner: All'
```
malaffoon: Gail informed me this should be "English Learner Status" ... sigh.

gregcharles: Gah! Oh, well.

PR #2011, opened 2019-06-19
possible fix for     RP-484 admin service is spewing stack traces due…

ftdholbrook:
… to ?database problems?

the errors are not coming from any RDW code but from org.springframework.boot.actuate.endpoint.ConfigurationPropertiesReportEndpoint

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/Application.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/Application.java
@@ -2,6 +2,9 @@ package org.opentestsystem.rdw.admin;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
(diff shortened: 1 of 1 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/Application.java
```diff
@@ -10,7 +13,10 @@
 @SpringBootApplication(exclude = {
         ElastiCacheAutoConfiguration.class,
         ContextRegionProviderAutoConfiguration.class,
-        ContextStackAutoConfiguration.class
+        ContextStackAutoConfiguration.class,
+        DataSourceAutoConfiguration.class,
+        DataSourceTransactionManagerAutoConfiguration.class,
+        HibernateJpaAutoConfiguration.class
```
ftdholbrook: based on https://stackoverflow.com/a/49276219 I'm not sure this will help, but I don't believe it can hurt.  I'm unable to reproduce this locally in docker.

PR #2013, opened 2019-06-19
RP-493 Handle In Process State for tenant Creation

ftdholbrook:


admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/GuavaCacheTenantAdministrationStatusServiceTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test;
+import org.opentestsystem.rdw.admin.model.TenantConfiguration;
+
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class GuavaCacheTenantAdministrationStatusServiceTest {
+
+    private GuavaCacheTenantAdministrationStatusService statusService;
```
ftdholbrook: I looked at spring cache, and it didn't have the map style semantics I was looking for.  This implementation uses guava cache, the idea being that adding another implementation that uses spring cloud Redis (similar semantics) for shared state across services.  This is really more of an expiring shared store than a cache.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/TenantAdministrationStatusService.java
```diff
@@ -0,0 +1,18 @@
+package org.opentestsystem.rdw.admin.multitenant;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.opentestsystem.rdw.admin.model.TenantConfiguration;
+
+import java.util.Optional;
+
+public interface TenantAdministrationStatusService {
```
esotrope: nit: javadocs - beat you @malaffoon

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/DefaultTenantAdministrationRequestHandler.java
```diff
@@ -75,10 +98,25 @@ public void createSandbox(TenantConfiguration tenantConfiguration) {
...
-        logger.debug("Processing SandboxCreationRequest");
-        tenantDatabaseCreationService.createSandboxDatabases(tenantConfiguration);
-        tenantConfigurationPersistenceService.saveTenantConfiguration(tenantConfiguration);
-        configServerClient.postMonitor();
+        final String tenantKey = tenantConfiguration.getTenant().getKey();
+        try {
+            tenantAdministrationStatusService.putStatus(tenantKey, STARTED, tenantConfiguration);
+            logger.debug("Processing SandboxCreationRequest");
+
+            tenantAdministrationStatusService.updateStatus(tenantKey, CREATING_DATABASES);
```
esotrope: is the UI able to read this stuff?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/DefaultTenantAdministrationRequestHandler.java
```diff
@@ -56,10 +64,25 @@ public void createTenant(TenantConfiguration tenantConfiguration) {
...
+
+            tenantAdministrationStatusService.updateStatus(tenantKey, PERSISTING_CONFIGURATION);
+            tenantConfigurationPersistenceService.saveTenantConfiguration(tenantConfiguration);
+
+            tenantAdministrationStatusService.updateStatus(tenantKey, NOTIFYING_CONFIG_SERVER);
+            configServerClient.postMonitor();
+        } catch (Exception e) {
+            //catch all / rethrow to update status
+            tenantAdministrationStatusService.updateStatus(tenantKey, FAILED);
+            logger.error("Error creating tenant", e);
```
malaffoon: i would suggest this is a "warning" not an error. I reserve error for things that cause a service to stop working. Yes, the tenant isn't going to be set up properly but the admin service itself is still okay.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/TenantAdministrationConfiguration.java
```diff
@@ -17,6 +18,12 @@
 @Configuration
 public class TenantAdministrationConfiguration {
 
+    @Value("${tenant.tenantAdministration.statusService.maximumSize:1000}")
+    Integer statusServiceMaximumSize;
```
malaffoon: private

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/TenantAdministrationConfiguration.java
```diff
@@ -17,6 +18,12 @@
...
+
+    @Value("${tenant.tenantAdministration.statusService.expireAfterDays:7}")
+    Integer statusServiceExpireAfterDays;
```
malaffoon: Hmm, 1000 for 7 days ... those seem like big numbers. To make sure i understand this: once a tenant is stood up, the cache isn't going to be used for it at all, right? The system will get the tenant info from the configuration properties?  Just seems like we'll never have more than a few tenants actively being stood up. And it take more than a couple hours, we have problems?

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/GuavaCacheTenantAdministrationStatusServiceTest.java
malaffoon: Personally i love guava cache and i've always been disappointed it doesn't align well with spring cache

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/TenantAdministrationConfiguration.java
ftdholbrook: yep, once we see the real config on the listing page we can remove it.  The scenario I was thinking about was someone creates one Friday night.  They leave, it fails, and they come back Monday morning to check... I would hope it would still tell them it failed on Monday.  Regardless those defaults are a trivial amount of data, I'm happy to lower them.

ftdholbrook: 4 days and 100 sandboxes sound reasonable @malaffoon

malaffoon: Sure.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/DefaultTenantAdministrationRequestHandler.java
ftdholbrook: Not directly, but the get all calls will use this to construct a response for after a tenant / sandbox is created but not yet saved.

PR #2016, opened 2019-06-19
changes roles for sandbbox users

esotrope:


common/src/main/java/org/opentestsystem/rdw/reporting/common/security/ReportingRole.java
```diff
@@ -6,26 +6,26 @@
...
-    public static final String GeneralRole = "GENERAL";
-    public static final String PiiRole = "PII";
-    public static final String PiiGroupRole = "PII_GROUP";
-    public static final String CustomAggregateReporterRole = "CUSTOM_AGGREGATE_REPORTER";
-    public static final String EmbargoAdminRole = "EMBARGO_ADMIN";
-    public static final String GroupAdminRole = "GROUP_ADMIN";
-    public static final String InstructionalResourceAdminRole = "Instructional Resource Admin";
-    public static final String PipelineAdminRole = "PIPELINE_ADMIN";
-    public static final String TenantAdminRole = "TENANT_ADMIN";
+    public static final String General = "GENERAL";
```
esotrope: removed "Role" suffix from naming since all of these are now in the "ReportingRole" class

PR #2019, opened 2019-06-20
Feature/rp 494 dataset configuration reporting

vgonzalez9:
RP-495 RP-494 Dataset Configuration
Remove Sandbox mock data.
Update sandbox service to call new SandboxDataSets with the preconfigured datasets.
move the datasets to the reporting common.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
@@ -4,20 +4,20 @@ import org.opentestsystem.rdw.admin.model.ApplicationSandboxConfiguration;
 import org.opentestsystem.rdw.admin.model.ApplicationSandboxConfigurations;
 import org.opentestsystem.rdw.admin.model.SandboxConfiguration;
 import org.opentestsystem.rdw.admin.model.SandboxConfigurationPackage;
-import org.opentestsystem.rdw.admin.model.SandboxConfigurationStubs;
 import org.opentestsystem.rdw.admin.model.TenantConfigurationStubs;
 import org.opentestsystem.rdw.admin.repository.impl.InMemorySandboxConfigurationRepository;
 import org.opentestsystem.rdw.admin.service.SandboxService;
 import org.opentestsystem.rdw.multitenant.Tenant;
 import org.opentestsystem.rdw.multitenant.TenantProperties;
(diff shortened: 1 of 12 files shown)


admin-service/src/main/resources/application.yml
```diff
@@ -207,3 +207,10 @@ tenant-configuration-lookup:
       url: http://localhost:8086
     reporting-service:
       url: http://localhost:8108
+
+sandbox-datasets:
```
malaffoon: This shouldn't be checked in as a non-test embedded application.yml; you realize it gets shipped to california staging tomorrow, right? We need to document it as a required configuration change and remove it from the code.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
```diff
@@ -152,11 +132,15 @@ private void validateNewSandbox(final SandboxConfiguration sandboxConfiguration)
...
      */
     private String getCurrentTenantKey() {
-        // todo replace else with something else
-        return reportingTenantIdResolver.getTenantId().orElse(SandboxConfigurationStubs.getSandboxCAInfo().getTenantKey());
+        if(reportingTenantIdResolver.getTenantId().isPresent()) {
+            return reportingTenantIdResolver.getTenantId().get();
+        }
+        String errorMsg = "A Tenant must be configured to create a Sandbox.";
+        logger.error(errorMsg);
+        throw new IllegalArgumentException(errorMsg);
```
esotrope: isn't log & throw an anti-pattern? why not just throw?

common/src/test/java/org/opentestsystem/rdw/reporting/common/multitenant/SandboxDataSetsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+// TODO RP-394 Move some integration test (IT) to End to End tests (ETE) and configure to run in CI
+//this test only runs against a local docker instance
+// set this property to run this test (or temporarily comment out this annotation)
+// java -Dtenant-client.docker-it-enable=true
```
esotrope: i'm curious about how this works with all the `@Ingore()` do we want to use `@Condintional` for this prop?

PR #2023, opened 2019-06-20
adds http no-cache settings to request headers

esotrope:


webapp/src/main/webapp/src/app/shared/data/data-request-options.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Adds no-cache headers to all async http requests
+ */
+@Injectable()
+export class DataRequestOptions extends BaseRequestOptions {
+  headers = new Headers({
+    'Cache-Control': 'no-cache',
+    Pragma: 'no-cache',
+    Expires: 'Sat, 01 Jan 2000 00:00:00 GMT',
+    'If-Modified-Since': '0'
+  });
```
esotrope: for safe measure and a stronger guarantee of the client not caching server api resources.

webapp/src/main/webapp/src/app/shared/data/no-cache-http-interceptor.ts
```diff
@@ -0,0 +1,27 @@
...
+  HttpHandler,
+  HttpInterceptor,
+  HttpRequest
+} from '@angular/common/http';
+import { Observable } from 'rxjs';
+
+/**
+ * Adds no-cache headers to all async http requests
+ */
+export class NoCacheHttpInterceptor implements HttpInterceptor {
```
esotrope: not currently used but would be if we upgrade to use the new HttpClient module of angular

PR #2024, opened 2019-06-20
fixes error message for student search form

esotrope:
image


webapp/src/main/webapp/src/app/student/student.component.html
```diff
@@ -16,14 +24,16 @@
...
           type="submit"
           class="btn btn-primary"
-          [disabled]="searchForm.invalid"
+          [disabled]="formGroup.invalid"
         >
           {{ 'common.action.search' | translate }}
         </button>
       </div>
     </div>
+    <span class="help-block" role="alert">
```
esotrope: `role="alert"` should give us the accessibility behavior we want

PR #2026, opened 2019-06-20
fixes longitundinal year range computation

esotrope:


webapp/src/main/webapp/src/app/aggregate-report/support.ts
```diff
@@ -13,19 +13,31 @@ import { isNullOrEmpty } from '../shared/support/support';
...
+
+/**
+ * @param endYear The year to start the list with
+ * @param gradeCodes The consecutive grade codes covered by the report (there cannot be gaps)
  */
 export function computeEffectiveYears(
   endYear: number,
   gradeCodes: string[]
 ): number[] {
-  // This is not foolproof and assumes each grade code can be parsed into a numeric value
```
esotrope: planned obsolescence

webapp/src/main/webapp/src/app/aggregate-report/support.spec.ts
```diff
@@ -16,17 +16,4 @@ describe('computeEffectiveYears', () => {
       98
     ]);
   });
-
-  it('should support grade code gaps', () => {
```
esotrope: this method was only used for longitudinal report queries which do not support gaps in years & grades

webapp/src/main/webapp/src/app/aggregate-report/support.ts
malaffoon: Dang, this will break as soon as we deploy because ELPAC has a grade = 'KG' assessment.

esotrope: well this should fix it

PR #2027, opened 2019-06-20
fixes filtering of total category for writing trait export

esotrope:


webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -366,15 +366,21 @@ export class CsvExportService {
     let maxPoints: number = 0;
 
     exportRequest.assessmentItems.forEach((item, i) => {
-      exportRequest.summaries[i].rows.forEach(summary => {
-        compositeRows.push({
-          assessmentItem: item,
-          writingTraitAggregate: summary
-        });
+      exportRequest.summaries[i].rows
+        .filter(
```
esotrope: i guess we dont want to see this category for summatives : j

malaffoon: That is correct, the display is different for summative assessments.

PR #2031, opened 2019-06-22
Allows for running a significant subset of sandbox database creation locally

ftdholbrook:
RP-516

- setup locally with minimal changes to the typical configuration
- archive.uri-root should be set to a file (usually "file://tmp")
- the local archive root needs to be populated with datasets that match the configured datasets (copy from QA S3 s3://rdw-qa-archive/sandbox-datasets/ to /tmp/sandbox-datasets)
- no olap schemas or users are set up in local mode
  - if sandbox aggregate report testing is needed a schema and user in redshift will need to be manually provisioned and configured


vgonzalez9:
Is this only for Sandbox creation?

ftdholbrook:
> Is this only for Sandbox creation?

No, also general tenant creation, but most of the problems running locally were around dataset loading form S3 which is only used by sandbox creation.

PR #2033, opened 2019-06-24
moves user service and modeling to security module for changes to come

esotrope:


webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -103,7 +103,7 @@
                       <form
                         name="logoutForm"
                         method="post"
-                        action="/saml/logout"
+                        action="{{ user.logoutUrl }}"
```
esotrope: logout target is now dynamic

webapp/src/main/webapp/src/app/error/access-denied/access-denied.component.html
```diff
@@ -1,14 +1,11 @@
...
 <div class="well">
   <p>{{ 'access-denied.message' | translate }}</p>
-  <a
-    class="btn btn-primary"
-    href="javascript:void(0)"
-    onclick="window.document.logout.submit()"
-    >{{ 'access-denied.logout' | translate }}</a
-  >
-  <form name="logout" method="post" action="/saml/logout" class="hidden"></form>
+  <form name="logout" method="post" action="{{ logoutUrl$ | async }}">
```
esotrope: access denied logout target is now dynamic

PR #2035, opened 2019-06-24
moves routes into public and protected routes to prevent infinite loo…

esotrope:
…ping problem caused by protecting the route that we route to when you are not authenticated...


webapp/src/main/webapp/src/app/admin/admin.routes.ts
```diff
@@ -0,0 +1,23 @@
+import { aggregateReportRoutes } from '../aggregate-report/aggregate-report.routes';
+import { embargoRoutes } from './embargo/embargo.routes';
+import { groupRoutes as groupAdminRoutes } from './groups/groups.routes';
+import { ingestPipelineRoutes } from './ingest-pipeline/ingest-pipeline.routes';
+import { instructionalResourceRoutes } from './instructional-resource/instructional-resource.routes';
+import { organizationExportRoutes } from '../organization-export/organization-export.routes';
+import { tenantRoutes } from './sandbox-tenants/tenant.routes';
+import { Route } from '@angular/router';
+
+export const adminRoutes: Route[] = [
```
esotrope: modularized routes so that each module presents the routes for that module

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -1,231 +1,60 @@
...
-          path: '',
-          pathMatch: 'prefix',
-          component: EmbargoComponent,
-          resolve: { embargoesByOrganizationType: EmbargoResolve }
-        }
-      ]
+/**
+ * Routes accessible without permissions
+ */
+export const publicRoutes: Route[] = [
```
esotrope: the functional change of this PR was pulling these routes out from under the root protected route into unprotected sibling routes

webapp/src/main/webapp/src/app/user-group/user-group.module.ts
```diff
@@ -31,7 +28,7 @@ import { ExamModule } from '../exam/exam.module';
     StudentSearchModule,
     TableModule
   ],
-  exports: [UserGroupComponent, UserGroupsComponent],
-  providers: [UserGroupService, UserGroupOptionsService, UserGroupResolve]
```
esotrope: cleaned up unnecessary exports and providers wherever it was obvious

PR #2037, opened 2019-06-24
Defect/502 session expiry

esotrope:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -139,7 +139,8 @@ protected void configure(final HttpSecurity http) throws Exception {
                         "/**/*.ico",
                         "/**/*.woff2*",
                         "/**/*.woff*",
-                        "/**/*.ttf*"
+                        "/**/*.ttf*",
+                        "/**/*.map"
```
esotrope: maybe this will fix downloading the src map?

webapp/src/main/webapp/src/app/shared/security/service/security.service.ts
```diff
@@ -0,0 +1,99 @@
...
+      .subscribe(({ sessionRefresh }) => {
+        this.resolveResource(sessionRefresh);
+      });
+  }
+
+  checkHasOneOrMorePermission(): Observable<boolean> {
+    return this.hasOneOrMorePermission().pipe(
+      map(has => {
+        if (!has) {
+          this.denyAccess();
```
esotrope: image

webapp/src/main/webapp/src/app/shared/security/service/security.service.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  constructor(
+    private settingService: SecuritySettingService,
+    private router: Router,
+    windowRefService: WindowRefService
+  ) {
+    this.window = windowRefService.nativeWindow;
+  }
+
+  expireSession(): void {
```
esotrope: image

webapp/src/main/webapp/src/app/admin/embargo/embargo.routes.ts
```diff
@@ -11,7 +11,7 @@ export const embargoRoutes: Route[] = [
       breadcrumb: { translate: 'embargo.title' },
       permissions: ['EMBARGO_WRITE']
     },
-    canActivate: [AuthorizationCanActivate],
```
esotrope: did a bit of code shuffling, renaming and deleting now unused security related components.
now sandbox users can refresh their session by clicking "OK" on the session expired page and restore their data.

webapp/src/main/webapp/src/app/shared/security/authentication.service.ts
```diff
@@ -1,100 +0,0 @@
...
-  /**
-   * Call this on authentication failure.
-   * This will navigate the user to the session-timeout route.
-   */
-  navigateToAuthenticationExpiredRoute(): void {
-    let url = this.window.location.href;
-
-    // In the case the the root URL is reserved for public pages.
-    // Store the prefered default authentication route instead of the root URL.
-    if (
```
esotrope: all of this code and the tests for this code were because we required users to be authorized on the /session-expired route instead of making it public smh. this code avoided the infinite redirect loop that would cause...

webapp/src/main/webapp/src/app/app-security-setting.service.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  getSettings(): Observable<SecuritySettings> {
+    return forkJoin(
+      this.settingsService.getSettings(),
+      this.userService.getUser()
+    ).pipe(
+      map(([settings, user]) => {
+        const accessDenied: Resource =
+          settings.accessDeniedUrl != null
+            ? { url: settings.accessDeniedUrl }
```
malaffoon: reading configured value again?   👍

webapp/src/main/webapp/src/app/shared/security/service/security.service.ts
malaffoon: So, if i'm a sandbox user clicking OK will just take me right back in, correct? And if i'm not a sandbox user, i'll have to go through SSO? Do we know which is which when we put this up?

esotrope: atm it is black-box but we can tailor this message to sandbox users if we want

malaffoon: Yeah, i think a better message for sandbox users would be something like "Your session has timed out. Please click OK to continue working."

PR #2038, opened 2019-06-25
Defect/502 session expiry

esotrope:


webapp/src/main/webapp/src/app/app-settings.ts
```diff
@@ -46,19 +46,11 @@ export interface TargetReportSettings {
...
-    ? value.substring(RedirectPrefix.length)
-    : undefined;
-}
-
 export function toApplicationSettings(
   serverSettings: any
 ): ApplicationSettings {
   return {
-    accessDeniedUrl: parseAccessDeniedUrl(serverSettings.accessDeniedUrl),
+    accessDeniedUrl: serverSettings.accessDeniedUrl,
```
esotrope: allows the access denied URL to actually work

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -773,9 +773,12 @@
       "user-dropdown": "Welcome {{firstName}} {{lastName}}"
     },
     "session-expired": {
-      "message": "Your session has expired. Please click OK to log in again.",
-      "ok": "OK",
-      "title": "Session Expired"
+      "title": "Session Expired",
+      "message": {
+        "reload": "Your session has timed out. Please click OK to login again",
+        "refresh": "Your session has timed out. Please click OK to continue working."
```
esotrope: now there is a context specific message depending on whether the action refreshes session or prompts login again

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/ViewController.java
```diff
@@ -8,48 +8,53 @@
...
     }
 
-    @RequestMapping(value = {"/"})
-    public String index(@AuthenticationPrincipal final User user) {
-        if (user == null) {
-            return this.landingPageUrl;
+    @RequestMapping("/access-denied")
+    public String accessDenied() {
+        if (!isBlank(this.accessDeniedUrl)) {
+            return this.accessDeniedUrl;
```
esotrope: these changes make it such that we no longer need the static `access-denied.html` and `error.html` we can rely on the newly "public" angular views

PR #2040, opened 2019-06-25
RP-520 integrate sandbox getAll calls with TenantAdministrationStatus

vgonzalez9:
Integrated TenantServices's getAll with TenantAdministrationStatus
Integrated SandboxServices's getAll with TenantAdministrationStatus
Added new list ability to the TenantAdministrationStatus to get the tenant/sandboxes being created.
Updated & Added new test cases.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/SandboxConfigurationStubs.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/SandboxConfigurationStubs.java
@@ -1,6 +1,7 @@
 package org.opentestsystem.rdw.admin.model;
 
 
+import java.time.Instant;
 import java.util.HashSet;
 import java.util.Set;
 import org.opentestsystem.rdw.admin.multitenant.DataSet;
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfigurationStubs.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfigurationStubs.java
@@ -179,6 +179,7 @@ public class TenantConfigurationStubs {
         tenant.setKey("CA");
         tenant.setName("California");
         tenant.setDescription("This is a CA description");
+        tenant.setSandbox(false);
         return tenant;
     }
 
@@ -188,6 +189,7 @@ public class TenantConfigurationStubs {
         tenant.setKey("NV");
         tenant.setName("Nevada");
         tenant.setDescription("This is a Nevada description");
+        tenant.setSandbox(false);
         return tenant;
     }
 
@@ -456,9 +458,9 @@ public class TenantConfigurationStubs {
         aggregateReportingPropertiesRoot.setAssessmentTypes(aggregationProperties.getAssessmentTypes());
         aggregateReportingPropertiesRoot.setStateAggregateAssessmentTypes(aggregationProperties.getStateAggregateAssessmentTypes());
         aggregateReportingPropertiesRoot.setStatewideUserAssessmentTypes(aggregationProperties.getStatewideUserAssessmentTypes());
-        Map<String, AggregateReportingPropertiesTenant> aggregatePropertiesMap = new HashMap<>();
+        HashMap<String, AggregateReportingPropertiesTenant> aggregatePropertiesMap = new HashMap<>();
         aggregatePropertiesMap.put("CA", getCAAggregateReportingProperties());
-        aggregateReportingPropertiesRoot.setTenants((HashMap<String, AggregateReportingPropertiesTenant>) aggregatePropertiesMap);
+        aggregateReportingPropertiesRoot.setTenants(aggregatePropertiesMap);
         return Optional.of(aggregateReportingPropertiesRoot);
     }
 
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/GuavaCacheTenantAdministrationStatusService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/GuavaCacheTenantAdministrationStatusService.java
@@ -4,8 +4,11 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
+
 import org.opentestsystem.rdw.admin.model.TenantConfiguration;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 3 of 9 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/GuavaCacheTenantAdministrationStatusService.java
```diff
@@ -21,6 +24,15 @@ public GuavaCacheTenantAdministrationStatusService(int maximumSize, int expireAf
...
     }
 
+    @Override
+    public Map<String, TenantAdministrationStatus> getTenants() {
+        Map<String, TenantAdministrationStatus> statusMap = new HashMap<>();
+        statusCache.asMap().forEach((k, v) -> {
+            statusMap.put(k, v.getLeft());
+        });
+        return statusMap;
+    }
```
ftdholbrook: I don't think this is the right abstraction here.  I'd be OK with just a `statusCache.asMap()` pass through, In the `feature/sandbox-create-api-integration` branch I implemented this:
```
    public Set<String> allTenantKeys() {
        return statusCache.asMap().keySet();
    }
```

so maybe  something like

``` 
    public Set<...> allTenantsMap() {
        return statusCache.asMap();
    }
```

then anything downstream can work with the full set of data however it wants.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
```diff
@@ -33,22 +42,35 @@
...
         this.sandboxDatasets = sandboxDatasets;
+        this.tenantAdministrationStatusService = tenantAdministrationStatusService;
     }
 
     @Override
     public ApplicationSandboxConfigurations getAll() {
         ApplicationSandboxConfigurations applicationSandboxConfigurations = new ApplicationSandboxConfigurations();
+        Map<String, SandboxConfiguration> sandboxConfigurationMap = getSandboxConfigurationFromTenantProperties();
+        //todo remove once calling real create
+        if (inMemorySandboxConfigurationRepository.getAll().isEmpty()) {
```
ftdholbrook: Instead of depending on InMemorySandboxConfigurationRepository we could just depend on the interface, then when we implement the repository, there is no code change.  Please use the higher level abstraction.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/SandboxConfigurationStubs.java
```diff
@@ -1,6 +1,7 @@
 package org.opentestsystem.rdw.admin.model;
 
 
+import java.time.Instant;
```
malaffoon: this used?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TenantConfigurationStubs.java
```diff
@@ -179,6 +179,7 @@ public static Tenant getTenantCA() {
         tenant.setKey("CA");
         tenant.setName("California");
         tenant.setDescription("This is a CA description");
+        tenant.setSandbox(false);
```
malaffoon: please tell me Tenant defaults this to false and this is just you being explicit

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
```diff
@@ -93,6 +115,112 @@ public void resetDataSet(String key) {
...
+            //sandbox.setTenantKey();
+            sandboxConfiguration.setSandbox(sandbox);
+            sandboxConfiguration.setLocalization(tenantConfiguration.getLocalization());
+            ApplicationSandboxConfiguration applicationSandboxConfiguration = new ApplicationSandboxConfiguration();
+            applicationSandboxConfiguration.setReporting(tenantConfiguration.getApplicationTenantConfiguration().getReporting());
+            applicationSandboxConfiguration.setAggregateReportingPropertiesTenant(tenantConfiguration.getApplicationTenantConfiguration().getAggregateReportingPropertiesTenant());
+            sandboxConfiguration.setApplicationSandboxConfiguration(applicationSandboxConfiguration);
+            return sandboxConfiguration;
+        }
+    };
```
ftdholbrook: I think these two conversion methods show that there is not actual difference between a Sandbox and a Tenant?  Tenants have all the data needed for a sandbox?  Can we remove all this and just use Tenant?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/GuavaCacheTenantAdministrationStatusService.java
malaffoon: One comment about that: statusCache.asMap() returns a map that affects the underlying cache, i don't like that. And the returned value doesn't have to be concurrent, so how about `return newHashSet(statusCache.asMap().keySet())`?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultSandboxService.java
```diff
@@ -93,6 +115,112 @@ public void resetDataSet(String key) {
         inMemorySandboxConfigurationRepository.reset(key);
     }
 
+    /**
+     * Create a SandboxConfiguration from a TenantConfiguration
+     */
+    public Function<TenantConfiguration, SandboxConfiguration> tenantToSandbox = new Function<TenantConfiguration, SandboxConfiguration>() {
+        @Override
+        public SandboxConfiguration apply(final TenantConfiguration tenantConfiguration) {
+            SandboxConfiguration sandboxConfiguration = new SandboxConfiguration();
```
malaffoon: buillder with copy would make this much more readable

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantService.java
```diff
@@ -240,30 +244,70 @@ private boolean isTenantFound(final String key) {
...
-            applicationTenantConfiguration.setDataSources(tenantDataSources);
-
-            tenantConfiguration.setApplicationTenantConfiguration(applicationTenantConfiguration);
-            tenantConfigurationMap.put(key, tenantConfiguration);
+
+            // only want tenants
+            if (!t.isSandbox()) {
+                TenantConfiguration tenantConfiguration = new TenantConfiguration();
+
+                Tenant tenant = new Tenant();
```
malaffoon: please use a builder for this stuff

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultTenantService.java
```diff
@@ -240,30 +244,70 @@ private boolean isTenantFound(final String key) {
...
+                tenantConfiguration.setTenant(tenant);
+
+                // see the any of the currents exist in the createTenantMap
+                if (createTenantMap.containsKey(tenant.getKey())) {
+                    // if it exists need to remove that tenant from the statusService
+                    tenantAdministrationStatusService.remove(tenant.getKey());
+                    createTenantMap.remove(tenant.getKey());
+                }
+
+                // each tenant has it's own applicationTenantConfiguration
```
malaffoon: nit: s/it's/its

(ask Will, this is my favorite PR feedback comment)

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/GuavaCacheTenantAdministrationStatusService.java
ftdholbrook: Or even an immutable map

ftdholbrook:
This needs some further discussion, pulling this PR for now.

PR #2042, opened 2019-06-25
Task/498 sandbox role sort order

esotrope:
image


webapp/src/main/resources/application.sql.yml
```diff
@@ -54,7 +54,7 @@ sql:
       (
         SELECT
           'DistrictAdministrator' as type,
-          1 as type_order,
+          3 as type_order,
```
esotrope: swapped the order of the unions completely forgetting that i was sorting on this instead x D

PR #2046, opened 2019-06-25
Defect/524 cannot view app with client level role

esotrope:
You can now log in with only client level reporting roles

image

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/authority/DefaultAuthorityService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/authority/DefaultAuthorityService.java
@@ -2,6 +2,7 @@ package org.opentestsystem.rdw.reporting.security.authority;
 
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
+import org.apache.commons.lang.StringUtils;
 import org.opentestsystem.rdw.multitenant.TenantProperties;
 import org.opentestsystem.rdw.reporting.common.security.ReportingRole;
 import org.opentestsystem.rdw.reporting.sandbox.SandboxRole;
@@ -19,15 +20,21 @@ import org.springframework.stereotype.Service;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static com.google.common.collect.Maps.newHashMap;
 import static com.google.common.collect.Sets.newHashSet;
 import static java.util.Collections.emptySet;
+import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
+import static org.apache.commons.lang.StringUtils.isBlank;
 import static org.opentestsystem.rdw.reporting.common.security.ReportingRole.ReportingRoles;
 import static org.opentestsystem.rdw.reporting.security.authority.Authorities.toAuthority;
+import static org.opentestsystem.rdw.utils.Grant.Level.CLIENT;
 import static org.opentestsystem.rdw.utils.Grant.Level.DISTRICT;
 import static org.opentestsystem.rdw.utils.Grant.Level.DISTRICT_GROUP;
 import static org.opentestsystem.rdw.utils.Grant.Level.INSTITUTION;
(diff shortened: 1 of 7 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/authority/DefaultAuthorityService.java
```diff
@@ -83,34 +90,39 @@
...
-                authorities.add(authority);
-            } catch (final IllegalArgumentException exception) {
-                // in a well-behaved and properly configured ecosystem this should never happen
-                // so log a WARN level message so the log may be monitored
-                logger.warn("invalid grant: {}", encodedGrant);
-            }
-        }
-        return authorities;
+
+        final List<Authority> authorities = Arrays.stream(encodedGrants)
```
esotrope: @malaffoon i made the logic here much more readable - it passes tests and seems logical to me but can you double check i didn't miss anything here?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/authority/DefaultAuthorityService.java
```diff
@@ -155,6 +167,8 @@
                         .computeIfAbsent(permissionId, (key) -> PermissionScope.builder());
 
                 switch (authority.getLevel()) {
+                    case CLIENT:
+                    case STATE_GROUP:
```
esotrope: these now map to statewide permission which they really already should have

webapp/src/main/webapp/src/app/shared/i18n/rdw-translate-loader.ts
```diff
@@ -52,20 +52,8 @@ export class RdwTranslateLoader implements TranslateLoader {
...
-      map(([clientTranslations, serverTranslations, subjects]) => {
-        const asmtTranslations = this.createSubjectTranslations(
-          subjects,
-          serverTranslations
-        );
-        return flattenJsonObject(
-          merge(clientTranslations, asmtTranslations, serverTranslations)
-        );
-      })
+    return this.getUserTranslations(languageCode).pipe(
```
esotrope: dedup

webapp/src/main/webapp/src/app/shared/i18n/rdw-translate-loader.ts
```diff
@@ -122,16 +114,16 @@ export class RdwTranslateLoader implements TranslateLoader {
   private getUserTranslations(languageCode: string) {
     return forkJoin(
       this.getClientTranslations(languageCode),
-      this.getServerTranslations(languageCode),
-      this.subjectService.getSubjectCodes()
+      this.getServerTranslations(languageCode).pipe(catchError(() => of({}))),
+      this.subjectService.getSubjectCodes().pipe(catchError(() => of([])))
```
esotrope: need this for when no tenant is set

webapp/src/test/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsServiceTest.java
```diff
@@ -89,21 +90,6 @@ public void itShouldCreateAUser() {
                 );
     }
 
-    @Test
-    public void itShouldCreateAnEmptyUserIfNoGrantsWithState() {
```
esotrope: no longer a valid assertion

webapp/src/test/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsServiceTest.java
```diff
@@ -113,7 +99,6 @@ public void itShouldCreateAnEmptyUserIfNoGrants() {
         assertThat(user.getUsername()).isEqualTo("alice@example.com");
         assertThat(user.getPermissionsById()).isEmpty();
         assertThat(user.getAuthorities()).isEmpty();
-        assertThat(user.isEnabled()).isFalse();
```
esotrope: doesn't really make sense to model user accounts as  "disabled" when they just don't have permissions in this system

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/authority/DefaultAuthorityService.java
malaffoon: LGTM

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/authority/DefaultAuthorityService.java
malaffoon: Hmmm. Good for now but this does raise the issue of what it means to grant, say, PII at the client level. But i guess that is more of an SSO/ART concern than RDW's problem.

esotrope: hm good point. we could validate that the role can only be an admin role?

webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlUserDetailsService.java
```diff
@@ -54,56 +52,54 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
...
 
+        final User.Builder user = User.builderExt()
+                .id(id)
+                .firstName(firstName)
+                .lastName(lastName)
+                .email(email)
+                .username(username)
+                .tenant(
+                        UserTenant.builder()
+                                .id(stateId)
```
malaffoon: having a tenant in here with a null stateId is ok?

esotrope: maybe not the most logical but it will have no side affects i'm aware of so long as ART permissions are assigned right

ftdholbrook: FWIW I'm fine with it there, it is just more user metadata.

PR #2048, opened 2019-06-26
RD-509: Improperly displaying LEP filters when LEP disabled.

gregcharles:
I can't fully test this until it's deployed to QA. I did at least test that it doesn't blow up.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
@@ -392,6 +392,7 @@ export class AggregateReportSummary {
         });
       }
       if (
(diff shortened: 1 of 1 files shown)


gregcharles:
No comments for two days. I'm going to go ahead and merge this. It's small, and won't be hard to undo if it doesn't work.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -392,6 +392,7 @@ export class AggregateReportSummary {
         });
       }
       if (
+        options.dimensionTypes.includes('LEP') &&
```
esotrope: this won't be exactly correct as i think this is for display advanced filters which != dimension types -> i'll take a peek and see if i cant reproduce locally

PR #2053, opened 2019-06-27
RP-538: improve XML compare by normalizing expected output

gregcharles:
The XML parser is specifically set to preserve the formatting of the original XML document, but it can change whitespacing inside of elements (between attributes or before the closing ">" or "/>". I'm normalizing the expected output when tests are saved by running it through a "no-op" pipeline script, which still does the parsing and reformatting. This keeps the relatively stupid Ace editor diff tool from highlighting differences based on whitespace changes.


admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineService.java
```diff
@@ -120,6 +123,45 @@ public Pipeline updatePipeline(final Pipeline pipeline) {
...
+                                    pipeline.getCode()
+                            )));
+            toScriptCompiler(
+                    scriptSource.getScript(PipelineScriptType.Base, pipeline.getCode())
+                            .map(PipelineScriptDefinition::getBody)
+                            .orElseThrow(() -> new IllegalStateException(String.format(
+                                    "The \"%s\" pipeline has no base script configured",
+                                    pipeline.getCode()
+                            )))
+            ).compile(scriptBody);
```
esotrope: i think i'm following this but want to make sure. we are taking the saved expected output and running it through a no-op script to normalize the output before comparison? there is no way to just use the underlying `outputXml` code directly to avoid involving a pipeline in the process?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineTestService.java
```diff
@@ -28,17 +32,21 @@ public PipelineTest get(final int id) {
...
                 test.copy()
+                        .output(normalizedOutput)
                         .updatedBy(username)
                         .build()
         );
     }
 
     @Override
     public PipelineTest update(final String username, final PipelineTest test) {
+        final String normalizedOutput = pipelineService.getNormalizedOutput(test.getOutput(), test.getPipelineId());
```
esotrope: does normalizing before storing have any side effect on the view when a user would refresh after saving?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/PipelineService.java
```diff
@@ -65,6 +65,17 @@
...
+    /**
+     * Gets a normalized version of the expected output. This adjusts whitespace and other formatting
+     * to match with changes the pipeline script would make. These are formatting type changes, separate
+     * from any data transformations the script makes.
+     *
+     * @param expectedOutput the expected output as entered by the user
+     * @param pipelineId the pipeline ID which is the type of the pipeline
+     * @return normalized version of the expected output
+     */
+    String getNormalizedOutput(String expectedOutput, Integer pipelineId);
```
malaffoon: nit: maybe param name should be just `output`? The fact that it is the expected output is a use-case, not really what this method is about. Okay, i'll go crawl back into my corner  ;)

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineService.java
malaffoon: Poor Greg, he had to have this conversation with me yesterday.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineService.java
```diff
@@ -120,6 +123,45 @@ public Pipeline updatePipeline(final Pipeline pipeline) {
         return emptyList();
     }
 
+    @Override
+    public String getNormalizedOutput(final String expected, final Integer pipelineId) {
+        final Pipeline pipeline = repository.findById(pipelineId);
+        try {
+            final String scriptBody =
```
malaffoon: nit: might be worth a comment in the javadocs or somewhere that this method is not optimized for performance: it should only be used in "user time".

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineService.java
```diff
@@ -120,6 +123,45 @@ public Pipeline updatePipeline(final Pipeline pipeline) {
...
+                    scriptSource.getScript(PipelineScriptType.Normalize, pipeline.getCode())
+                            .map(PipelineScriptDefinition::getBody)
+                            .orElseThrow(() -> new NoSuchElementException(String.format(
```
malaffoon: If this method is always called, for a pipeline test create/update then it not existing doesn't feel exceptional, and throwing an exception and logging at INFO level doesn't seem appropriate. At the very least, please reduce the logging to DEBUG.

If i'm misinterpreting the flow, let me know.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineService.java
gregcharles: The only flows that deal with the problem are this one or adjusting the expected output on load to display in the case of a failed test. This one means what is saved could be adjusted slightly from what the user asked to be saved. The other would mean that what showed up in the test's expected output could be slightly different from what showed up in the side-by-side comparison. Neither is perfect, but I preferred the first one.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultPipelineTestService.java
gregcharles: The view automatically updates after the test is saved, and the adjustments to whitespace are displayed then. The user doesn't have to explicitly refresh. I intend to add a note to the admin guide that calls out the possibility of these slight adjustments, so at least the user would be aware that could happen.

gregcharles:
Incorporated suggestions before merge.

PR #2054, opened 2019-06-27
makes writing trait score component more robust

esotrope:
Refactored this component to avoid possible race conditions causing export to fail.

image


webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
```diff
@@ -9,227 +13,293 @@ import { WritingTraitScoreSummary } from '../../../model/writing-trait-score-sum
...
-    WritingTraitScoreSummary,
-    Column[]
-  > = new Map();
+  private assessment$: Subject<Assessment> = new BehaviorSubject(undefined);
+  private exams$: Subject<Exam[]> = new BehaviorSubject(undefined);
+  private assessmentProvider$: Subject<
+    AssessmentProvider
+  > = new BehaviorSubject(undefined);
+  private destroyed$: Subject<void> = new Subject();
+  private _hasDataToExport: boolean;
```
esotrope: entire component is now "reactive" which is nice as it clearly shows the dependency chain for values computed from one or more inputs

PR #2056, opened 2019-06-28
Feature/rp245 update copyright year

bgsavage:


webapp/src/main/webapp/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "node --max_old_space_size=2048 node_modules/@angular/cli/bin/ng build --outputHashing=all",
     "build-prod": "node --max_old_space_size=2048 node_modules/@angular/cli/bin/ng build --prod",
-    "start": "ng serve --proxy-config proxy.conf.json",
+    "start": "ng serve --proxy-config proxy.conf.json -o",
```
bgsavage: Unrelated, but neat trick I recently discovered which opens the browser for you.

esotrope: nice

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -757,7 +757,7 @@
       "home-sr": "Home",
       "window-title": "Smarter Balanced | Reporting System"
     },
-    "footer": "© 2018 The Regents of the University of California",
+    "footer": "© 2019 The Regents of the University of California",
```
esotrope: @malaffoon i can't remember - did we want this to be a dynamic value or no?

esotrope: Also makes me wonder if SBAC wants tenants to be able to overwrite this text through config

bgsavage: I was wondering if this should be `new Date().getFullYear()` or not.  I know it certainly complicates things with Admin now being able to edit these values.

malaffoon: meh, i don't think it needs to be dynamic; this will make them ask for work every year  ;)

PR #2058, opened 2019-06-28
RP-252: If the ISR report is empty, it should display as "No Results"

gregcharles:
This one is tricky. The printed report is created in multiple asynchronous steps. The first pulls a list of students for the requested report, and if the list is size 0, it marks the status as "No Results" and stops processing. That much was working already.

However, for the scenario Nohemi calls out in the JIRA, there are 7 potential students. These are divided into chunks (just one in this case) and the chunks are re-queued. The next phase of processing handles each chunk by loading the relevant reports for each student based on the query parameters. In Nohemi's scenario, that meant loading just summative reports, and none of the 7 students had any to load. The processing continues regardless to produce a blank PDF file for the chunk. The file is not empty, but it does produce a blank page, or actually two blank pages. This file is then saved to a temp location by the archive service.

The final step assembles all the temporarily archived PDFs into a single master report, and then marks the report as complete. It is unaware that it has created a blank report.

I think the best way to fix this would be to move the report loading up into the fetch student step, and filter out any students that have no loaded reports. That seemed to me a pretty big structural change, which I wanted to avoid, so I looked for a workaround. 

I can't  just abort during the "chunk to PDF" phase, because in the general case it could happen that one chunk would be emtpy, but others have reports.  What I'm doing instead is suppressing the creation of the partial PDFs for chunks that would just create blank pages. Then, in the assemble phase, I prevent it from erroring on missing chunk PDF files, and if there are no PDFs at all to assemble, I mark the report as "No Results". This feels a little klugey, but it was relatively straightforward to implement, and does work. The JUnits were a little harder since I had to mock conditions that would create a non-empty report. All-in-all, I think it was much easier than restructuring the report processor.

### DIFF
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/GeneratePdf.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/GeneratePdf.java
@@ -61,6 +61,12 @@ public class GeneratePdf {
             final List<Supplier<InputStream>> chunkPdfs = reportGenerationTemporaryStorage.getChunks(
                     report.getId(),
                     parseInt(report.getMetadata().get(FetchStudents.TotalChunk)));
+
+            if (chunkPdfs.isEmpty()) {
+                setNoContent(report);
+                return;
+            }
+
             final MergedReport mergedReport = pdfMerger.mergePdfs(report, chunkPdfs);
 
             logger.debug("Merged UserReport: {}", message.getReportId());
@@ -88,5 +94,12 @@ public class GeneratePdf {
         } catch (final IOException e) {
             throw new IllegalStateException("Unable to collate report: " + message.getReportId(), e);
         }
+    }
+
+    private void setNoContent(final UserReport report) {
+        userReportService.update(UserReport.builder()
+                .copy(report)
+                .status(ReportStatus.NO_RESULTS)
+                .build());
     }
 }
--- a/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportGenerationTemporaryStorage.java
+++ b/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ArchiveBackedReportGenerationTemporaryStorage.java
@@ -35,8 +35,13 @@ public class ArchiveBackedReportGenerationTemporaryStorage implements ReportGene
         final List<Supplier<InputStream>> chunkContents = new ArrayList<>();
         for (int i = 0; i < chunkCount; i++) {
             final int chunkIdx = i;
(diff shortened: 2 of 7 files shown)


malaffoon:
Nice description, thanks.

Could there be a problem with the fetching of the students in the early step? Why is it returning any students if none of the students have test results? Could that step be (easily) adjusted to not return any students?

malaffoon:
It seems like adjusting the sql.student.findByExamQueryParams* queries should be changed to join subject_asmt_type and filter out those that don't have printed_report enabled. Assuming we know it is only ISR's that use those queries.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream/GeneratePdf.java
```diff
@@ -61,6 +61,12 @@ public void collateReportPdf(final ReportRequestMessage message) {
             final List<Supplier<InputStream>> chunkPdfs = reportGenerationTemporaryStorage.getChunks(
                     report.getId(),
                     parseInt(report.getMetadata().get(FetchStudents.TotalChunk)));
+
+            if (chunkPdfs.isEmpty()) {
+                setNoContent(report);
+                return;
+            }
```
malaffoon: Although i think we should fix the student SQL, i can't object to this change. It'll take care of lingering gotchas.

gregcharles: Thanks for that. The queries are complex to the point that it took me a long time of staring at it and running experiments to be convinced that "printed_reports=1" was enough to filter the no-result students, in particular because the query for Summative/ICA reports is very different from the one for IAB.

I'm not sure what ISR's are, but I assume somehow that means the printed reports that we're creating here. The modified sql in application.sql.yml is only used in StudentRepository.findStudentsByExamQueryParams(), and that is only called by FetchStudents.fetchAndPublishStudents(). That's a stream listener on the channel "ReportFetchStudentsIn", which base on my reading of the config, and some experimentation, is only invoked for Printed Reports. 

I left my original code in place since you said you thought part of it could take care of lingering gotchas, and that part wouldn't be effective without the other parts. Let me know though.

malaffoon: LGTM. ISR = Individual Student Report.

gregcharles: Aha, I figured that's what it stood for, but got confused because these are often for multiple students. It just occurred to me that it's ISR vs Aggregate, so many students, but taken one-by-one. I'm slow, but usually come around eventually.

PR #2060, opened 2019-06-29
Defect/545 disable export buttons under embargo

esotrope:


webapp/src/main/webapp/src/app/assessments/assessments.component.ts
```diff
@@ -402,10 +403,11 @@ export class AssessmentsComponent implements OnChanges {
 
   private updateFilterOptions() {
     this.exportDisabled =
-      this.embargoed &&
+      this.embargo != null &&
+      this.embargo.enabled &&
       this.selectedAssessments.every(
         ({ type, schoolYear }) =>
-          type === 'sum' && schoolYear === this.currentSchoolYear
+          type === 'sum' && schoolYear === this.embargo.schoolYear
```
esotrope: technically the same thing but this is more clear i think

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -288,17 +288,27 @@
...
-                [disabled]="
-                  !currentExportResults ||
-                  !currentExportResults.hasDataToExport()
-                "
-                (click)="currentExportResults.exportToCsv()"
-              >
-                <i class="fa fa-cloud-download"></i>
-                {{ 'common.export' | translate }}
-              </button>
+              <ng-container *ngIf="exportButtonDisabled; else exportEnabled">
```
esotrope: now export button correctly disables for teacher target reports and writing trait scores

webapp/src/main/webapp/src/app/shared/embargo/reporting-embargo.service.ts
```diff
@@ -17,50 +17,34 @@ export class ReportingEmbargoService {
...
-      )
-    );
+    return this.getEmbargo().pipe(map(({ enabled }) => enabled));
   }
 
   /**
    * Gets the current embargo settings
    */
   getEmbargo(): Observable<Embargo> {
-    return this.userHasEmbargoReadAccess().pipe(
```
esotrope: yikes was this a bug - it was only embargoing things in csv export and disabling export buttons if the user had EMBARGO_READ permission - not anymore. This check is only really needed for displaying the embargo alert for admin at the top of the page so i moved the logic there.
@malaffoon do we possibly want to show everyone that banner? or really just admin?

webapp/src/main/webapp/src/app/student/results/student-results.component.ts
```diff
@@ -97,8 +97,14 @@ export class StudentResultsComponent implements OnInit {
           this.minimumItemDataYear = settings.minItemDataYear;
         });
 
-        this.embargoService.isEmbargoed().subscribe(embargoed => {
-          this.exportDisabled = embargoed;
+        this.embargoService.getEmbargo().subscribe(embargo => {
```
esotrope: improved student test history top level export button to be more "exam" aware. now it only disables if the only exams on the page are summative and embargoed

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -288,17 +288,27 @@
...
+              </ng-container>
+              <ng-template #exportEnabled>
+                <button
+                  class="btn btn-default btn-xs"
+                  (click)="currentExportResults.exportToCsv()"
+                >
+                  <i class="fa fa-cloud-download"></i>
+                  {{ 'common.export' | translate }}
+                </button>
+              </ng-template>
```
esotrope: image

webapp/src/main/webapp/src/app/shared/embargo/reporting-embargo.service.ts
malaffoon: If it is the banner i'm thinking of then only the admin should see it. All other users don't need to see it because, when embargo is on, it is as if those test results don't exist. 

To clarify: admins are allowed to see the summatives during embargo but not allowed to export them. All other users just don't see the summatives during embargo. They have no knowledge of embargo.

esotrope: ahh ok. makes sense. either way this should work fine then as the results wont reach the view for the unworthy ; j

PR #2061, opened 2019-06-29
change sandbox teacher roles to be one per school

malaffoon:
SB really didn't like having a teacher per grade. They want a teacher per school instead.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxRole.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxRole.java
@@ -13,7 +13,6 @@ public class SandboxRole {
     private String id;
     private SandboxRoleType type;
     private Organization organization;
-    private String gradeCode;
 
     public String getId() {
         return id;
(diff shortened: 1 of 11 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/SandboxRole.java
```diff
@@ -13,7 +13,6 @@
     private String id;
     private SandboxRoleType type;
     private Organization organization;
-    private String gradeCode;
```
malaffoon: With this change, we can completely eliminate grade-code

webapp/src/main/resources/application.sql.yml
```diff
@@ -3,40 +3,30 @@ sql:
...
-        student_group_id
-      )
-      SELECT
-        :user_login,
-        id
+      INSERT IGNORE INTO user_student_group ( user_login, student_group_id )
+      SELECT :user_login, id
       FROM student_group
-      WHERE name LIKE CONCAT(:organization_name, ' (%,', :grade_code, ')')
+      WHERE name LIKE CONCAT(:organization_name, '%')
```
malaffoon: matches all groups that start with the school name

esotrope: i like it. nice simplification

PR #2062, opened 2019-07-01
Feature/rp 371 add ui mask to sensitive fields

bgsavage:
The goal for this PR is just to mask sensitive fields, but I did get a little carried away with the unrelateds as I noticed things and was testing.  I'll call each one out.


webapp/src/main/webapp/angular.json
```diff
@@ -142,27 +142,6 @@
...
-              {
-                "glob": "*",
-                "input": "node_modules/@sbac/sbac-ui-kit/dist/images",
-                "output": "/assets/image"
-              },
-              {
-                "glob": "*",
-                "input": "node_modules/@sbac/sbac-ui-kit/dist/images/icon",
-                "output": "/"
-              }
```
bgsavage: I noticed the unit tests weren't able to run.  I first came here and saw that the test build wasn't in sync with the prod build.  Since the styles and assets weren't kept in sync anyways, just thought it was better to just remove them.  The unit tests shouldn't be using the styles and assets anyways, and would just slow them down.

webapp/src/main/webapp/package.json
```diff
@@ -72,18 +72,17 @@
     "husky": "^1.3.1",
     "jasmine-core": "~3.3.0",
     "jasmine-spec-reporter": "~4.2.1",
-    "karma": "~3.1.3",
-    "karma-cli": "~2.0.0",
```
bgsavage: The real reason the unit tests weren't running was because of an outdated karma version.  I upgrade us to v4.

webapp/src/main/webapp/package.json
```diff
@@ -72,18 +72,17 @@
...
+    "karma": "^4.1.0",
+    "karma-chrome-launcher": "^2.2.0",
     "karma-coverage-istanbul-reporter": "^2.0.4",
     "karma-jasmine": "~2.0.1",
     "karma-jasmine-html-reporter": "^1.4.0",
-    "karma-phantomjs-launcher": "^1.0.0",
```
bgsavage: An unrelated to this unrelated is that phantomjs is no longer supported.  So i replaced phantomjs with chromium headless browser to run unit tests.

webapp/src/main/webapp/karma.conf.js
```diff
@@ -1,33 +1,38 @@
 // Karma configuration file, see link for more information
 // https://karma-runner.github.io/1.0/config/configuration-file.html
 
-module.exports = function (config) {
+const puppeteer = require("puppeteer");
+process.env.CHROME_BIN = puppeteer.executablePath();
```
bgsavage: Puppeteer is required to set this.

webapp/src/main/webapp/karma.conf.js
```diff
@@ -1,33 +1,38 @@
...
+    reporters:
+      config.angularCli && config.angularCli.codeCoverage
+        ? ["progress", "karma-remap-istanbul", "teamcity"]
+        : ["progress"],
     port: 9876,
     colors: true,
     logLevel: config.LOG_INFO,
     autoWatch: true,
-    browsers: ['PhantomJS'],
+    browsers: ["ChromeHeadless"],
```
bgsavage: Replace PhantomJs as the unit test browser of choice with ChromeHeadless.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.html
```diff
@@ -95,12 +95,16 @@
           <td>
             <input
               #overrideInput
-              type="text"
+              [type]="
+                rowNode.node.data.secure && !rowNode.node.data.showSecure
+                  ? 'password'
+                  : 'text'
+              "
```
bgsavage: Ok, here is the real code for this PR> . When the data type is secure and want to show secure, then set the type to password.  This makes the input show the dots instead of clear text.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.html
```diff
@@ -95,12 +95,16 @@
...
+              data-lpignore="true"
```
bgsavage: This tells lastpass to not add any iconography and try to generate passwords.  It messes with the ui icons in the input.  These passwords should already be known anyways.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.html
```diff
@@ -95,12 +95,16 @@
...
               [value]="rowNode.node.data.value"
               [formControlName]="rowNode.node.data.formControlName"
               (change)="updateOverride(rowNode.node.data)"
               class="property-input"
-              [style.padding-right]="calculatePadding(rowNode.node.data)"
```
bgsavage: Oops i need to bring this back, was testing what this does.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.html
```diff
@@ -114,6 +119,20 @@
...
+                  'fa-eye': !rowNode.node.data.showSecure
+                }"
+                *ngIf="rowNode.node.data.secure"
+                (click)="
+                  rowNode.node.data.showSecure = !rowNode.node.data.showSecure
+                "
+                placement="left"
+                container="body"
+                triggers="mouseenter:mouseleave"
+              ></i>
```
bgsavage: Adds clickable icon to hide/unhide the password.

Screen Shot 2019-07-01 at 9 37 46 AM

Screen Shot 2019-07-01 at 9 37 59 AM

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
```diff
@@ -32,6 +32,13 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
   @Input()
   readonly = true;
 
+  // Should these be data driven?
+  readonly secureFields = ['password', 's3SecretKey'];
+
+  readonly requiredFields = ['password'];
+
+  readonly encryptedFields = ['password'];
```
bgsavage: Added these arrays which will drive some of the fields below.  Not sure if we want the API to return this data at some point or not?  It's kind of configurations of configurations, which may be too much.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-sandbox/tenant-sandbox.component.ts
```diff
@@ -70,6 +74,7 @@ export class TenantSandboxComponent implements OnInit, OnChanges {
   onOpenButtonClick(): void {
     this.formGroupOriginalValues = cloneDeep(this.formGroup.value);
     this.open = true;
+    this.cdRef.detectChanges();
```
bgsavage: This resolves a console error on page load.  Consider this a band-aid fix.   Angular is complaining that the model has changed (specifically, the form validity) after the view has been checked.  To get around it for now, we have to explicitly call detectChanges.  But this might be better fixed with how this form is loaded.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
```diff
@@ -103,25 +103,22 @@ export class NewTenantConfigurationComponent implements OnInit, AfterViewInit {
     const updatedProperties = { ...this.configurationProperties };
     const key = this.tenantForm.get('key').value.toUpperCase();
     const defaultDataBaseName = `reporting_${key}`;
-    const archivePathPrefixProperty = <ConfigurationProperty>(
-      updatedProperties.archive.find(property => property.key === 'pathPrefix')
-    );
-    archivePathPrefixProperty.value = key;
-    archivePathPrefixProperty.originalValue = key;
```
bgsavage: These locally scoped variables aren't use anywhere below and are throwing null reference errors in the console.  Seems to be dead code I can just remove.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
```diff
@@ -103,25 +103,22 @@ export class NewTenantConfigurationComponent implements OnInit, AfterViewInit {
...
-        const dbProperty = <ConfigurationProperty>(
-          updatedProperties.datasources[dataSourceKey].find(
-            property => property.key === 'urlParts.database'
-          )
-        );
-        dbProperty.value = defaultDataBaseName;
-        dbProperty.originalValue = defaultDataBaseName;
-      }
-    );
+    if (this.configurationProperties) {
```
bgsavage: Added this null check, was getting a console error here.  This also seems like a band-aid fix, I'm not sure exactly yet what the expected behavior is when the default configuration properties from the server is null.

webapp/src/main/webapp/src/index.html
```diff
@@ -1,32 +1,57 @@
...
+    <link
+      rel="icon"
+      type="image/png"
+      sizes="192x192"
+      href="/android-icon-192x192.png"
+    />
+    <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png" />
+    <link rel="icon" type="image/png" sizes="96x96" href="/favicon-96x96.png" />
+    <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
+    <!-- <link rel="manifest" href="/manifest.json"> -->
```
bgsavage: This file is mostly formatting from VS code.  The real change is here, removing manifest.json.  Just commented because I'm not sure if we are using one?  Either way, it seems to be throwing 404 as it's not there anyways?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
```diff
@@ -32,6 +32,13 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
   @Input()
   readonly = true;
 
+  // Should these be data driven?
```
esotrope: yes ; J
As stated we'll want to move toward storing metadata on each of these fields in the future

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
```diff
@@ -173,17 +180,18 @@ export class PropertyOverrideTreeTableComponent implements OnInit {
         group.value &&
         typeof group.value === 'string' &&
         group.value.startsWith('{cipher}') &&
-        group.key === 'password';
+        this.encryptedFields.some(x => x === group.key);
 
       childrenNodes.push({
-        data: {
+        data: <ConfigurationProperty>{
```
esotrope: why is the cast required here?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/tenant-sandbox/tenant-sandbox.component.ts
esotrope: 👍

webapp/src/main/webapp/src/app/admin/sandbox-tenants/component/property-override-tree-table.component.ts
bgsavage: So the IDE "find Usages" will find the usages here.  (Otherwise it wasn't).

Also nice to enforce the data model here should it somehow come out of synch.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/model/configuration-property.ts
```diff
@@ -5,6 +5,8 @@ export class ConfigurationProperty {
   group?: string;
   formControlName: string;
   encrypted?: boolean;
+  secure?: boolean;
+  showSecure?: boolean;
   required?: boolean;
```
esotrope: beyond the scope of this PR but it would be nice if we separated the view-specific fields from the "metadata provided" fields from the actual data provided by the backend

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.html
```diff
@@ -16,7 +16,7 @@ <h1 class="h2" heading>
             #tenantKeyInput
             onkeyup="this.value = this.value.toUpperCase();"
             type="text"
-            class="form-control"
+            class="form-control text-uppercase"
```
esotrope: i know it is a tad futile but in all the new work i've been introducing i've been trying to pull away from bootstrap util classes in favor of custom css in a component `less` stylesheet. Also, do we need the `onkeyup="this.value = this.value.toUpperCase();"` if we style it uppercase and just convert to uppercase when sending to the backend?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
esotrope: sweet

webapp/src/main/webapp/src/index.html
esotrope: what is the manifest.json for do you know?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.html
bgsavage: Ah, yeah this didn't have a component `less` stylesheet so wasn't sure what the established pattern is.  Yes, I'd much prefer this in a component stylesheet as well. 

Also, I agree the onkeyup data changing is a bit odd, but I didn't want to change it just yet, just wanted to improve the ux for now.  And honestly, wouldn't this be a concern of the API prior to saving to the data store?

webapp/src/main/webapp/src/app/admin/sandbox-tenants/model/configuration-property.ts
bgsavage: Nice to have :)

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.html
esotrope: yeah it would be best that the API enforces the rule.

webapp/src/main/webapp/src/app/admin/sandbox-tenants/pages/new-tenant.component.ts
vgonzalez9: There should be archive properties. they just may not be configured for local.dev.

bgsavage: hm, might have been unfinished code then?

PR #2064, opened 2019-07-01
make json resource translation provider tenant-aware

malaffoon:
The JsonResourceTranslationProvider used to take a string that was the resource location; set during bean creation. Changed things so it resolves the location dynamically, so it uses any tenant-overridden configuration based on the session.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemProperties.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemProperties.java
@@ -139,7 +139,7 @@ public interface ReportingSystemProperties {
      */
     List<String> getEffectiveReportLanguages();
 
-    public static class StateProperties {
+    class StateProperties {
 
         private String name;
         private String code;
@@ -176,7 +176,7 @@ public interface ReportingSystemProperties {
 
     }
 
-    public static class TargetReportProperties {
+    class TargetReportProperties {
 
         private Double insufficientDataCutoff;
         private Integer minNumberOfStudents;
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonResourceTranslationProvider.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonResourceTranslationProvider.java
@@ -16,6 +16,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
+import org.opentestsystem.rdw.multitenant.TenantContextHolder;
+import org.opentestsystem.rdw.reporting.common.multitenant.StaticTenantIdUtil;
 
 /**
  * This translation provider accepts a location accessible via {@link ResourceLoader}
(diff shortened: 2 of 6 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/ReportingSystemProperties.java
```diff
@@ -176,7 +176,7 @@ public String toString() {
 
     }
 
-    public static class TargetReportProperties {
```
malaffoon: unrelated, fix warnings

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/MessageFormattingTest.java
```diff
@@ -62,7 +64,7 @@ public static void beforeClass() {
      * Unformatted messages do not require escaping single apostrophes.
      * This test asserts after processing of messages, there is no instance of consecutive apostrophes in the message.
      *
-     * @throws Exception
+     * @throws Exception it's a test
```
malaffoon: IJ was complaining about the non-documented @throws

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/TranslationConfiguration.java
```diff
@@ -48,4 +44,37 @@ public MessageSource messageSource(final TranslationProvider translationProvider
         messageSource.setUseCodeAsDefaultMessage(true);
         return messageSource;
     }
+
```
malaffoon: The same impl is used for the embedded resources and remove resources. So i need an abstraction around getting the location.

common/src/test/java/org/opentestsystem/rdw/reporting/common/i18n/MessageFormattingTest.java
esotrope: doh

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonResourceTranslationProvider.java
```diff
@@ -42,8 +44,13 @@ public JsonResourceTranslationProvider(final String translationLocation,
         if (locale == null) {
             return Collections.emptyMap();
         }
+        final String location = settings.getTranslationLocation();
```
malaffoon: This is the main change: get the translation location (through the tenant-aware report properties resolver) based on the tenant in the current session.

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonResourceTranslationProvider.java
```diff
@@ -26,13 +28,13 @@
...
-    private final String translationLocation;
+    private final TranslationProperties settings;
     private final ResourceLoader resourceLoader;
     private final JavaPropsMapper propertiesMapper;
 
-    public JsonResourceTranslationProvider(final String translationLocation,
+    public JsonResourceTranslationProvider(final TranslationProperties settings,
                                            final ResourceLoader resourceLoader) {
-        this.translationLocation = translationLocation;
+        this.settings = settings;
```
esotrope: this can be set at construction time? it is a scoped bean i take it?

malaffoon:
@ftdholbrook , the problem i was having locally is that the resource loader was being too smart finding things. So, instead of putting the en.json file in a /i18n/ subfolder i just put it in the tenant folder:
```
/tenant-CA
  application.yml
  en.json
```
where application.yml has this in it:
```
reporting:
  tenants:
    CA:
      translation-location: "binary-${spring.cloud.config.uri}/*/*/master/tenant-CA/"
```

common/src/main/java/org/opentestsystem/rdw/reporting/common/i18n/JsonResourceTranslationProvider.java
malaffoon: Yes, the settings are a proxied class so they will be refreshed behind the scenes. At least that's what @ftdholbrook promised me  ;)

PR #2065, opened 2019-07-01
proper xml is now required for saving exam pipeline tests

esotrope:


webapp/src/main/webapp/src/app/admin/ingest-pipeline/ingest-pipeline.support.ts
```diff
@@ -44,3 +47,37 @@ export function getOrAppendAsyncScript(
...
+ * Identifies and returns all the XML syntax errors for the provided XML string value
+ *
+ * @param value The xml string to check for errors
+ */
+export function getXmlParserErrors(value: string): string[] {
+  const normalizedValue = (value || '').trim();
+  const dom = domParser.parseFromString(normalizedValue, 'text/xml');
+  return Array.from(
+    dom.documentElement.getElementsByTagName('parsererror')
+  ).map(element => element.textContent);
```
esotrope: didn't realize there was such a browser native way of doing xml validation. very convenient

webapp/src/main/webapp/src/app/admin/ingest-pipeline/model/pipelines.ts
```diff
@@ -1,18 +1,47 @@
...
-function isValidPipelineTextText(value: string): boolean {
-  return !isBlank(value) && value.length <= maximumTextLength;
+function getInputTypeValidators(inputType: string): ValidatorFn[] {
+  if (inputType === 'xml') {
+    return [xml];
+  }
+  return [];
 }
 
-export function isValidPipelineTest(value: PipelineTest): boolean {
```
esotrope: now this method literally makes a form group to perform validation.
if i did the component setup again i would be passing around wrappers with a handle on a form group so i wouldn't need to create these on the fly.
At least now the validation is using the same code as the form itself so there will be no chance of inconsistency

webapp/src/main/webapp/src/app/admin/ingest-pipeline/model/pipelines.ts
```diff
@@ -1,18 +1,47 @@
...
+}
+
+export function pipelineTestFormGroup(inputType: string): FormGroup {
+  const inputOutputValidators: ValidatorFn[] = getInputOutputValidators(
+    inputType
   );
+  return new FormGroup({
+    name: new FormControl(''),
+    input: new FormControl('', inputOutputValidators),
+    output: new FormControl('', inputOutputValidators)
```
bgsavage: Reactive forms eh?

esotrope: yeeah boii ; p

PR #2067, opened 2019-07-02
fixes left nav on aggregate report form. removes filter for target re…

esotrope:
…ports on result page

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
@@ -77,10 +77,10 @@ export class ScrollNavComponent {
     }
   }
 
-  @HostListener('window:scroll', [])
+  // Disabled because the item highlighting logic is very inaccurate and may cause more confusion than aid
+  //@HostListener('window:scroll', [])
   onWindowScroll(): void {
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -27,31 +34,18 @@ export class ScrollNavComponent {
...
-        : items.map(item => {
-            //Saturate element reference from id
-            if (
-              Utils.isNullOrUndefined(item.scrollTo) &&
-              !Utils.isNullOrUndefined(item.id)
-            ) {
-              item = Object.assign({}, item, {
-                scrollTo: document.getElementById(item.id)
-              });
-            }
```
esotrope: fix was to not aggressively find and set the document element as it may not have loaded yet

webapp/src/main/webapp/src/app/shared/nav/scroll-nav.component.ts
```diff
@@ -70,35 +64,36 @@ export class ScrollNavComponent {
   @Input()
   set activeItem(item: ScrollNavItem) {
     this._activeItem = item;
-    if (item.scrollTo) {
-      item.scrollTo.scrollIntoView();
+    const scrollTo = document.getElementById(item.id);
```
esotrope: query dom on demand

PR #2069, opened 2019-07-02
add verbiage for longitudinal school year info pop-up

malaffoon:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -55,12 +55,10 @@
     "preview": "This table does not contain actual data and is for preview purposes only.",
     "subheading": "Select the attributes to include in the report. Then click \"Create Report.\"",
     "field": {
-      "report-type-label": "Report Type",
-      "report-type-info": "TODO",
```
malaffoon: i didn't see these being used anywhere

PR #2070, opened 2019-07-02
prevents subject assessments with no standard cutoff from grouping ac…

esotrope:
…hievement levels

image


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report.component.ts
```diff
@@ -400,8 +400,10 @@ export class AggregateReportComponent implements OnInit, OnDestroy {
             valueDisplayType:
               query.valueDisplayType || ValueDisplayTypes.Percent,
             performanceLevelDisplayType:
-              query.achievementLevelDisplayType ||
-              PerformanceLevelDisplayTypes.Separate
+              subjectDefinition.overallScore.standardCutoff != null
+                ? query.achievementLevelDisplayType ||
+                  PerformanceLevelDisplayTypes.Separate
+                : undefined
```
esotrope: we were already correctly hiding the control but because we preserved the display options from the sample table on the query screen it allowed you to group the performance levels that way

PR #2071, opened 2019-07-02
RP-475: multiple database statements as a single named operation

gregcharles:
Proof of concept for multiple database statements under a single name. I put in the four grants already being tested in the test method JdbcMultitenantDatabaseAdminIT.grantsShouldBeAppliedToUser()

Note: yaml array syntax not handled correctly by Springs @Value tag, but comma-separated strings are fine, and can be on multiple lines in the yml file if they have the ">-" prefix.


admin-service/src/main/resources/application.sql.yml
```diff
@@ -485,6 +485,12 @@ sql:
     grantUserFile: >-
       GRANT FILE ON *.* TO :username @'%'
 
+    grantAllBasicPermissions: >-
+      GRANT ALL PRIVILEGES ON $[databaseName].* TO :username @'%',
+      GRANT SELECT ON mysql.proc TO :username @'%',
+      GRANT LOAD FROM S3 ON *.* TO :username @'%',
+      GRANT SELECT INTO S3 ON *.* to :username @'%',
+
```
ftdholbrook: I like it, is the trialing comma required?  Or was this to test out blank line handling?

gregcharles: The second one. It's not required.

malaffoon: I admit i was expecting something a bit more yaml-y, e.g.
```
  grantAllBasicPermissions:
    - GRANT ALL PRIVILEGES ON $[databaseName].* TO :username @'%'
    - GRANT SELECT ON mysql.proc TO :username @'%'
    - GRANT LOAD FROM S3 ON *.* TO :username @'%'
    - GRANT SELECT INTO S3 ON *.* to :username @'%'
```
I saw your note that this wouldn't work with @Value, but i think it would be better to do it this way and switch from @Value to a config properties class. Pretty sure something like this would work:
```
@ConfigurationProperties(prefix = "sql.multitenantDatabaseAdmin")
public class AdminSqlConfiguration {
    private List<String> grantAllBasicPermissions;
    ...
}
```
Of course, that configuration class would probably have all the necessary SQL collections.

gregcharles: OK, I rewrote it with @ConfigurationProperties. Pro: config file can use the idiomatic YAML array syntax. Cons: configuration properties (as opposed to @Value) needs public setters for each of the simple properties and a public getter for the array property. Also the class needs an annotation in the Application class. It took me a fair while to work out why the properties weren't getting bound at first.

admin-service/src/main/java/org/opentestsystem/rdw/admin/Application.java
```diff
@@ -19,6 +22,7 @@
         HibernateJpaAutoConfiguration.class
 })
 @PropertySource(value = "classpath:build-info.properties", ignoreResourceNotFound = true)
+@EnableConfigurationProperties(JdbcMultitenantDatabaseAdmin.class)
```
malaffoon: The better way to do this is to just Import the configuration class. Please look at ApplicationConfiguration.java and put it there.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
```diff
@@ -51,75 +54,32 @@
...
-    @Value("${sql.multitenantDatabaseAdmin.grantAllOlapUserToSchema}")
     private String grantAllOlapUserToSchema;
-
-    @Value("${sql.multitenantDatabaseAdmin.grantAllOlapUserToSchemaTables}")
     private String grantAllOlapUserToSchemaTables;
-
-    @Value("${sql.multitenantDatabaseAdmin.searchPathOlapUserToSchema}")
     private String searchPathOlapUserToSchema;
 
+    private List<String> grantAllBasicPermissions = new ArrayList<>();
```
malaffoon: I think it would be better to introduce a separate class that has the properties, e.g. MultitenantAdminConfiguration. Then include it in here as class variable passed in the constructor.

If you do that and the previous comment, the class won't need setters.

admin-service/src/main/resources/application.sql.yml
malaffoon: See my comments: separate the config properties from the implementation that uses them and you won't need any special handling. There are examples of this in the code already.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
```diff
@@ -176,6 +136,22 @@ public boolean isDatabaseUserExists(String username, RdwDatabaseType databaseTyp
                 Boolean.class);
     }
 
+    @Override
+    public void grantAllBasicPermissions(String username, String targetDatabase, RdwDatabaseType databaseType) {
+        Map<String, String> replacements =
+                ImmutableMap.<String, String>builder().put("databaseName", targetDatabase).build();
+
+        // Filter out blank strings. (May occur with trailing comma in SQL config file)
```
malaffoon: not sure if you need the filtering any more (the comment in parens can be removed)

admin-service/src/main/resources/application.sql.yml
gregcharles: I still need the setters though, right? Online docs and my own experiments are pretty clear about this being a requirement for @ConfigurationProperties.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
gregcharles: +1

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdmin.java
gregcharles: See above comment. Spring docs say @ConfigurationProperties won't bind without setters, and that was my finding as well even after these changes.

admin-service/src/main/java/org/opentestsystem/rdw/admin/Application.java
gregcharles: This was more of a matter of the tests not working unless I did it this way. (Actually, it took me some time to realize that the app itself was working right, and it was just the integration test that was loading the configuration properties class with all null values.) However, I finally found an extra bit of configuration that mollified it.

admin-service/src/main/resources/application.sql.yml
malaffoon: Oh, right. But the getter/setters are on the dumb POJO which is "normal".

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/MultitenantAdminConfiguration.java
```diff
@@ -0,0 +1,227 @@
...
+
+    public String getGrantUserOnSelectProc() {
+        return grantUserOnSelectProc;
+    }
+
+    public void setGrantUserOnSelectProc(final String grantUserOnSelectProc) {
+        this.grantUserOnSelectProc = grantUserOnSelectProc;
+    }
+
+    public String getGrantUserLoadFromS3() {
```
malaffoon: I'm assuming that a subsequent checkin will remove the now obsolete methods? That this review is just proofing out the concept?

ftdholbrook: Can we hold of on removing those methods?  If they get removed right now it will mean some dependencies will have to change that I'm pretty sure will cause a merge conflict right now.  I'd rather leave them until we merge back to the mainline and then remove them.  I'm happy to make a ticket to track.

PR #2072, opened 2019-07-02
Feature/555 add tenant name to application header

esotrope:
what it will look like until the tenant API is in

image

What it will look like with a tenant name

image

dropdown looks a bit better now

image


webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -5,134 +5,24 @@
     (click)="scrollToMainContent()"
     >{{ 'home.skip-link' | translate }}</a
   >
-  <div *ngIf="(user$ | async) as user" class="container">
```
esotrope: moved to component

PR #2075, opened 2019-07-03
improves iab card ordering

esotrope:
only a prob when the labels are bad so i don't think this is affecting anyone

image

### DIFF
--- a/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
+++ b/webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
@@ -11,8 +11,8 @@ import {
   AssessmentCardEvent,
   GroupCard
 } from './group-assessment-card.component';
-import { first, map, mergeMap } from 'rxjs/operators';
-import { byString } from '@kourge/ordering/comparator';
+import { map, mergeMap } from 'rxjs/operators';
+import { byString, Comparator, join } from '@kourge/ordering/comparator';
 import { ordering } from '@kourge/ordering';
 import { UserGroupService } from '../../user-group/user-group.service';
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -184,18 +189,15 @@ export class GroupDashboardComponent implements OnInit {
...
-          }
-      );
+      .map(measuredAssessment => ({
+        group: this.group,
+        measuredAssessment: measuredAssessment,
+        performanceLevels: this._subjectDefinitions.find(
+          definition =>
+            definition.subject === measuredAssessment.assessment.subject &&
+            definition.assessmentType === measuredAssessment.assessment.type
+        ).overallScore.levels
```
esotrope: moved to new prop path

webapp/src/main/webapp/src/app/dashboard/group-dashboard/group-dashboard.component.ts
```diff
@@ -232,9 +234,9 @@ export class GroupDashboardComponent implements OnInit {
   }
 
   updateMeasuredAssessments(assessments: MeasuredAssessment[]): void {
-    this.measuredAssessments = assessments.sort(
-      ordering(byString).on<MeasuredAssessment>(x => x.assessment.label).compare
-    );
+    this.measuredAssessments = assessments
+      .slice()
+      .sort(measuredAssessmentComparator);
```
esotrope: now sorts on grade then label

esotrope: not it is not sort on grade sequence however : /

PR #2076, opened 2019-07-03
removes unwanted test and updates package lock

esotrope:


webapp/src/main/webapp/src/app/shared/component/navigation/navigation.component.spec.ts
```diff
@@ -1,24 +0,0 @@
...
-
-  beforeEach(() => {
-    fixture = TestBed.createComponent(NavigationComponent);
-    component = fixture.componentInstance;
-    fixture.detectChanges();
-  });
-
-  it('should create', () => {
-    expect(component).toBeTruthy();
-  });
```
bgsavage: Tests can't fail if they don't exist 🤔

esotrope: i'm opposed to these tests. only introduced it because i happened to use `ng g component`

bgsavage: I like them as they'll catch html / runtime errors that the compiler misses.  But I also get that they can be annoying to maintain esp if it's a component being actively developed.

esotrope: i have never found them helpful. That may just be because my workflow is to actually play with the UI and make sure everything renders though. With the velocity at which the UI changes i find them to be a time drain that has little to no payoff </rant>

PR #2079, opened 2019-07-03
Defect/479 accessibility

esotrope:
new tab focus highlighting and hover contrast

image

image

image

image

image

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
@@ -8,17 +8,17 @@
       [label]="assessmentExam.assessment.label"
       [gradeCode]="assessmentExam.assessment.grade"
     ></assessment-label>
-
-    <sb-radio-group
+    <sb-radio-button-group
       *ngIf="scoreTypeOptions != null && scoreTypeOptions.length > 1"
+      id="score-type"
+      name="score-type"
       [options]="scoreTypeOptions"
       [(ngModel)]="scoreType"
-      id="score-type"
-      buttonGroupStyles="btn-group-sm"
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
```diff
@@ -8,17 +8,17 @@
       [label]="assessmentExam.assessment.label"
       [gradeCode]="assessmentExam.assessment.grade"
     ></assessment-label>
-
-    <sb-radio-group
+    <sb-radio-button-group
```
esotrope: the newer component highlights correctly when tabbed to

webapp/src/main/webapp/src/app/exam/component/show-results-divider/show-results-divider.component.html
```diff
@@ -1,5 +1,6 @@
 <hr />
 <a
+  href="javascript:void(0)"
```
esotrope: cant receive focus without this

esotrope:
@malaffoon something to note is that advanced filters are not tab accessible or at least not easily tab accessible. I don't want to touch that unless i have the license for a full redesign of it though

malaffoon:
Sounds good, thanks for the info.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.html
malaffoon: should we be using this everywhere, did you do a global search on this component?

esotrope: i'll run a search and see what i find

PR #2083, opened 2019-07-05
Lazy loaded admin modules - first stab

esotrope:
The 0.js is proof the ingest pipeline module was lazy loaded

image


admin-service/src/main/resources/application.yml
```diff
@@ -206,4 +206,14 @@ tenant-configuration-lookup:
     reporting-service: http://localhost:8108
 
 
-
+tenantProperties:
```
esotrope: got a little sloppy with git add i guess

webapp/src/main/webapp/src/app/admin/embargo/embargo.module.ts
```diff
@@ -8,17 +7,18 @@ import { EmbargoTable } from './embargo-table.component';
...
 import { HttpClientModule } from '@angular/common/http';
 import { EmbargoResolve } from './embargo.resolve';
-import { CommonModule } from '../../shared/common.module';
+import { ReportingCommonModule } from '../../shared/reporting-common.module';
 import { TableModule } from 'primeng/table';
+import { CommonModule } from '@angular/common';
 
 @NgModule({
   imports: [
-    BrowserModule,
```
esotrope: had to remove every instance of `Browser` and `BrowserAnimationModule` from every module in the app to avoid errors raised for loading them multiple times (something that only surfaced when lazy modules were introduced)

webapp/src/main/webapp/src/app/admin/embargo/embargo.module.ts
```diff
@@ -8,17 +7,18 @@ import { EmbargoTable } from './embargo-table.component';
...
-    ButtonsModule.forRoot(),
     CommonModule,
+    ButtonsModule,
+    ReportingCommonModule,
     FormsModule,
     HttpClientModule,
-    ModalModule.forRoot(),
```
esotrope: forRoot() should only be called in the app.module

webapp/src/main/webapp/src/app/admin/ingest-pipeline/ingest-pipeline.module.ts
```diff
@@ -23,19 +22,22 @@ import { PipelineTestFormComponent } from './component/pipeline-test-form/pipeli
...
     FormsModule,
     ReactiveFormsModule,
     ButtonsModule,
-    BsDropdownModule.forRoot(),
+    BsDropdownModule,
     ModalModule,
-    TooltipModule.forRoot()
+    TooltipModule,
+    TranslateModule.forChild(translateModuleConfiguration),
+    RouterModule.forChild(ingestPipelineRoutes)
```
esotrope: how the child module is configured

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-options.service.ts
```diff
@@ -1,6 +1,10 @@
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
-import { AggregateReportOptions } from './aggregate-report-options';
+import {
+  AggregateReportOptions,
+  Claim,
+  Subject
+} from './aggregate-report-options';
```
esotrope: also to get things working i had to fix all the circular dependencies in the app.
ended up using a static analysis tool called `madge` to do this

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -62,12 +62,6 @@ interface ResultsView {
   display: boolean;
 }
 
-export interface ExportResults {
-  exportToCsv(): void;
-
-  hasDataToExport(): boolean;
-}
```
esotrope: all these code changes outside the bounds of `module.ts` are to fix circular ts file imports

webapp/src/main/webapp/tsconfig.json
```diff
@@ -9,14 +9,9 @@
...
-      "node_modules/@types"
-    ],
-    "lib": [
-      "es2017",
-      "dom"
-    ],
-    "module": "es2015",
+    "typeRoots": ["node_modules/@types"],
+    "lib": ["es2017", "dom"],
+    "module": "esNext",
```
esotrope: we needed this setting to take advantage of lazy loaded modules

webapp/src/main/webapp/src/app/admin/embargo/embargo.module.ts
malaffoon: Some day I will need you to explain why it was the way it was (i.e. why import if it's not being used), and why the framework doesn't automagically deal with it.

malaffoon:
Looks fine; can you wait before merging? I may have one more bug fix i need to get into the stable build.

esotrope:
@malaffoon yup i'll hold. need to fix a couple tests for this branch still anyway. also i'd like to do a before and after of performance for the initial load time after all the admin stuff is lazy-loaded so i might keep this branch open until i have everything loaded lazily.

webapp/src/main/webapp/src/app/admin/embargo/embargo.module.ts
esotrope: `BrowserModule` imports `CommonModule` and what we actually wanted was `CommonModule` i think but we incorrectly established an early pattern of importing the `BrowserModule` which is only supposed to be imported once at the root of your app. Some imports in this framework have side-effects you could say so order of import matters and some modules complain when imported multiple times. and as we see now this complaint only surfaced when we had multiple app "chunks" due to lazy loading

malaffoon: Thanks for the explanation.

webapp/src/main/webapp/src/app/admin/admin.routes.ts
```diff
@@ -1,23 +1,79 @@
...
+
+export const sandboxesLoadChildren = () =>
+  import('app/admin/sandbox-tenants/sandbox.module').then(
+    module => module.SandboxModule
+  );
+
+export const tenantsLoadChildren = () =>
+  import('app/admin/sandbox-tenants/tenant.module').then(
+    module => module.TenantModule
+  );
```
esotrope: all admin modules are now lazy loaded. i would also like to make the other modules lazy loaded but i think that will be in a followup PR

PR #2085, opened 2019-07-09
Defect/479 button contrast

esotrope:
Tried different things but what ended up working the best was just making all the buttons grey when disabled which actually kind of makes sense anyway.

https://webaim.org/resources/contrastchecker/?fcolor=555555&bcolor=F8F8F8

image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
@@ -76,6 +76,13 @@ function createItem<T>(type: ItemType, value: T, changed = false): Item<T> {
   };
 }
 
+function isTestLoadedAndInvalid(
+  test: PipelineTest,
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/admin/ingest-pipeline/page/pipeline/pipeline.component.ts
```diff
@@ -570,9 +577,12 @@ export class PipelineComponent implements ComponentCanDeactivate, OnDestroy {
     // dont run
     const hasInvalidTests =
       this.selectedItem.type === 'Test'
-        ? !isValidPipelineTest(this.selectedItem.value, this.pipeline.inputType)
-        : tests.some(
-            ({ value }) => !isValidPipelineTest(value, this.pipeline.inputType)
+        ? isTestLoadedAndInvalid(
```
esotrope: footnote - we don't allow the storage of invalid tests atm so not loaded is synonymous with valid

PR #2086, opened 2019-07-09
improves some back to foregrouned color ratios

esotrope:
subtle grey text darkening to improve the color ratios - still not compliant though. basically the text needs to be very dark grey on white to be compliant which doesn't allow for any accenting with color (only size) so i think this may be a broader app style problem than any one page

before 

image

after

image

### DIFF
--- a/webapp/src/main/webapp/src/app/shared/form/sb-button-typeahead.component.ts
+++ b/webapp/src/main/webapp/src/app/shared/form/sb-button-typeahead.component.ts
@@ -38,13 +38,14 @@ import { Utils } from '../support/support';
           (click)="onButtonClickInternal()"
           [disabled]="buttonDisabledInternal"
         >
-          <span [ngClass]="{ gray: buttonDisabledInternal }"
+          <span [ngClass]="{ 'pointer-events-none': buttonDisabledInternal }"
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/shared/form/sb-button-typeahead.component.ts
```diff
@@ -38,13 +38,14 @@ import { Utils } from '../support/support';
           (click)="onButtonClickInternal()"
           [disabled]="buttonDisabledInternal"
         >
-          <span [ngClass]="{ gray: buttonDisabledInternal }"
```
esotrope: this was compounding the disabled gray color making the button non-compliant when disabled

webapp/src/main/webapp/src/primeng-overrides.less
```diff
@@ -13,10 +13,14 @@ body {
   button {
     font-size: @font-size-base;
   }
-}
 
-body .ui-table-scrollable-header.ui-widget-header {
-  border: none;
+  .dropdown-header {
+    color: @gray-dark;
+  }
```
esotrope: slightly darkens dropdown headers

PR #2087, opened 2019-07-09
adds keyboard trap fix

esotrope:
image

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-item/pipeline-item.component.html
+++ b/webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-item/pipeline-item.component.html
@@ -1,5 +1,6 @@
 <a
   #element
+  href="javascript:void(0)"
   class="item"
   [ngClass]="{
     active: _active
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
@@ -127,7 +127,7 @@ function createOrderingByColumnField(
     claimCode:
       reportType === 'Target'
         ? createOrganizationalClaimOrdering(subjectDefinition.subject, preview)
-        : ordering(ranking(subjectDefinition.scorableClaims)).on(
+        : ordering(ranking(subjectDefinition.claimScore.codes)).on(
             row => row.claimCode
           )
   };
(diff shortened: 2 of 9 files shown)


webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/pipeline-item/pipeline-item.component.html
```diff
@@ -1,5 +1,6 @@
 <a
   #element
+  href="javascript:void(0)"
```
esotrope: the "files" on the lhs are now tab accessible.

malaffoon:
So, if i'm editing and i hit Escape, where am i? Sorry, not clear to me.

esotrope:
@malaffoon 

image

malaffoon:
So then i can hit tab to go ... ? Just want to make sure the keyboard-only navigation works reasonably well.

esotrope:
@malaffoon tab will then go to the next form field or the browser address bar if it is the last thing on the page

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table.component.ts
```diff
@@ -127,7 +127,7 @@ function createOrderingByColumnField(
     claimCode:
       reportType === 'Target'
         ? createOrganizationalClaimOrdering(subjectDefinition.subject, preview)
-        : ordering(ranking(subjectDefinition.scorableClaims)).on(
+        : ordering(ranking(subjectDefinition.claimScore.codes)).on(
```
esotrope: unrelated moving away from deprecated fields

webapp/src/main/webapp/src/styles.less
```diff
@@ -2194,91 +2194,103 @@ fieldset.marginLeft50 {
 
 .rdw-scrollable-table {
   table,
-  tr td {
+  table > tbody > tr > td {
```
esotrope: aggregate report tables are now only is one `<table>` element and have the same scrolling behavior

webapp/src/main/webapp/src/styles.less
```diff
@@ -2194,91 +2194,103 @@ fieldset.marginLeft50 {
...
-        text-overflow: ellipsis;
-        text-align: left;
-        overflow: hidden;
+  table {
+    > thead {
+      > tr {
+        > th {
+          white-space: normal !important;
+          &.aggregate-report-organization {
+            width: @agg-organization-width;
```
esotrope: we already fixed the width of all the columns so we didn't need the javascript solution that the lib gave us.

PR #2089, opened 2019-07-10
RP-557: apply actuator config changes to other services

gregcharles:


aggregate-service/src/main/resources/application.yml
```diff
@@ -9,6 +9,14 @@ management:
     db:
       enabled: false
 
+# configure actuator endpoints to unmask "key" properties so tenant keys won't be masked. However, we still
```
malaffoon: Really makes me want to change what we called that field: it's not sensitive and having to do this for all the services is just ... annoying.

PR #2091, opened 2019-07-11
Defect/467 button highlight

esotrope:


webapp/src/main/webapp/src/app/admin/admin.routes.ts
```diff
@@ -51,9 +51,6 @@ export const adminRoutes: Route[] = [
         data: {
           breadcrumb: { translate: 'aggregate-reports.heading' },
           permissions: ['CUSTOM_AGGREGATE_READ']
-        },
-        resolve: {
-          options: AggregateReportOptionsResolve
```
esotrope: pushed down into the lazy module as this component isn't defined in the context until the module is loaded

PR #2094, opened 2019-07-11
RP-475: integrate with main code and delete obsolete methods

gregcharles:
Especially look at deduplication logic in DefaultTenantDatabaseCreationService.createDatabaseAndUsers()

The main difference is the compound statement is run once for each distinct username, schema, database type, whereas before most of the grants were run once for each distinct username and database type only, and only the grantUserToDatabase was run for each distinct username, schema, and database type.  

I _think_ this is OK,. Also, the test of the deduplication logic still passes.  The other option would be to remove that last grant from the compound statement.


admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
```diff
@@ -92,6 +92,13 @@ public TenantProperties tenantProperties() {
         return new TenantProperties();
     }
 
+    @Bean
+    @ConfigurationProperties("sql.multitenantDatabaseAdmin")
+    @RefreshScope
```
ftdholbrook: Is the idea that we can change these without rebooting the instance?  In general we have only been adding "RefreshScope" to tenant aware configuration properties.  The sql is admin level that operates on tenants, but is not tenant aware.

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdminIT.java
```diff
@@ -57,9 +58,10 @@
 @RunWith(SpringRunner.class)
 @IfProfileValue(name = "archive.uri-root", value = "s3://rdw-ci-archive")
 @RepositoryIT
-@ContextConfiguration(classes = {
-        JdbcMultitenantDatabaseAdmin.class
-})
+@ContextConfiguration(
+        classes = JdbcMultitenantDatabaseAdmin.class,
+        initializers = ConfigFileApplicationContextInitializer.class
```
ftdholbrook: I'm not familiar with ConfigFileApplicationContextInitializer what is it bringing to the table?

admin-service/src/main/java/org/opentestsystem/rdw/admin/ApplicationConfiguration.java
gregcharles: I'll take that out then. I didn't see why we would need it, but the other config prop beans were using it, so I was trying to stay consistent.

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/database/impl/JdbcMultitenantDatabaseAdminIT.java
gregcharles: It has something to do with loading the config properties bean through the ApplicationConfiguration with an @Bean annotated method rather than @EnableConfigurationProperties in the Application class. That resulted in the properties not being bound in JdbcMultitenantDatabaseAdminIT, i.e., all the db strings were null. I experimented with TenantProperties and found that wasn't bound either, so it wasn't unique to my case. I wasn't familiar with ConfigFileApplicationContextInitializer, but I found it suggested in a comment to a non-accepted answer to a question referenced by a question I found during a SO search. So, not the first thing I tried, and probably not even the tenth, but it's the one that worked. Hooray?

PR #2097, opened 2019-07-12
Task/fix build

esotrope:


webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/code-editor/code-editor.component.html
```diff
@@ -1,10 +1,16 @@
...
 ></div>
 
 <!-- addresses keyboard trap accessibility issue -->
-<a #escapePoint href="javascript:void(0)" class="escape-prompt">
+<a
+  #escapePoint
+  id="editorLabel"
+  href="javascript:void(0)"
+  class="escape-prompt"
+>
```
esotrope: gets read once when tabbing to the editor and again when escaping. if anyone has a creative solution for this let me know. all i can think of is to use an empty anchor tag next to this text instead

webapp/src/main/webapp/src/app/admin/embargo/embargo.module.ts
```diff
@@ -22,7 +22,7 @@ import { embargoRoutes } from './embargo.routes';
     ReportingCommonModule,
     FormsModule,
     HttpClientModule,
-    ModalModule,
+    ModalModule.forRoot(), // this is needed in lazy modules for some reason
```
esotrope: so this module does not provide a `forChild()` method and also modals are not working in lazy-loaded modules without `forRoot()` being called. i think this is likely a bug in the `ngx-bootstrap` lib

webapp/src/main/webapp/src/app/admin/ingest-pipeline/component/code-editor/code-editor.component.html
```diff
@@ -1,10 +1,16 @@
 <div
   #editor
   (keyup.escape)="escapePoint.focus()"
-  attr.aria-label="{{ 'code-editor.keyboard-trap-prompt' | translate }}"
+  role="editing"
+  aria-labelledby="editorLabel"
```
malaffoon: And this builds for you with `ng build --prod`?

esotrope: sure did

PR #2099, opened 2019-07-15
RP-586

malaffoon:


webapp/src/main/webapp/src/app/assessments/results/target-statistics-calculator.ts
```diff
@@ -261,7 +261,7 @@ export class TargetStatisticsCalculator {
 
     return this.mapTargetScoreDeltaToReportingLevel(
       average(scored),
-      standardErrorOfMean(scores)
+      standardErrorOfMean(scored)
```
malaffoon: `scores` is the unfiltered list that may contain null values; `scored` is the filtered list.

malaffoon:
I need to get this into qa so i'll merge and deal with feedback later

PR #2101, opened 2019-07-16
Feature/pub sub configprops

ftdholbrook:
RP-489 alternative to polling services directly for configprops using spring cloud stream pub sub
also fixes RP-484 admin service is spewing stack traces due to ?database problems?


admin-service/src/main/resources/application.yml
```diff
@@ -134,8 +141,8 @@ spring:
     validationQueryTimeout: 10000
     driverClassName: com.mysql.jdbc.Driver
     initialize: false
-    initialSize: 4
-    maxActive: 10
+    initialSize: 0
+    maxActive: 1
```
esotrope: i'm assuming these are something we override in the prod env but we should def make sure ; j

admin-service/src/main/resources/application.yml
```diff
@@ -187,6 +194,14 @@ spring:
     validationQuery: SELECT 1
     validationQueryTimeout: 30000
     driverClassName: com.amazon.redshift.jdbc42.Driver
+    initialize: false
```
ftdholbrook: this is the primary fix for the actuator endpoint emiting errors, as configprops reflected on this configuration, it was causing a database connection initialization.

admin-service/src/main/resources/application.yml
ftdholbrook: No, these admin datasources are seldom used, and seldom used by more than one user at once.  No point in chewing up resources.

esotrope: is that the metrics we have? i would think group import and some other admin services have a bunch of users all at once at certain times of the year - but i would also hope the appropriate env sets these props accordingly

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configuiration/TenantAdministrationConfiguration.java
```diff
@@ -138,4 +133,20 @@ public CacheManager configpropsCacheManager(){
...
+    @Bean
+    public ConfigpropsUpdateSubscriber configpropsUpdateSubscriber(ConfigpropsPublicationStorage configpropsPublicationStorage) {
+        return new ConfigpropsUpdateSubscriber(configpropsPublicationStorage);
+    }
+
+    @Bean
+    @ConfigurationProperties("sql.multitenantDatabaseAdmin")
+    public MultitenantAdminConfiguration adminSqlConfiguration() {
+        return new MultitenantAdminConfiguration();
+    }
```
ftdholbrook: Not related, moved this here from the main ApplicationConfiguration class

admin-service/src/main/resources/application.yml
ftdholbrook: These admin datasources in admin-service are only used for creating schemas, running flyway migrations, and other tasks created to tenant / sandbox creation.  They have no other purpose.

esotrope: oh ok - good to know. thanks for clarifying : j

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/ConfigpropsPublicationStorage.java
```diff
@@ -0,0 +1,9 @@
+package org.opentestsystem.rdw.admin.multitenant.configprops;
+
+import java.util.Map;
+
+public interface ConfigpropsPublicationStorage {
```
malaffoon: nit: javadocs

reporting-service/src/main/resources/application.yml
```diff
@@ -40,6 +40,12 @@ spring:
       enabled: true
       gzipped: true
 
+  cloud:
+    stream:
+      bindings:
+        ConfigpropsUpdateOutput:
+          destination: ConfigpropsUpdateExchange
```
malaffoon: Without the default won't this fail to publish if the consumer hasn't set up yet?

ftdholbrook: No, in pub-sub mode (publisher bound to "@Out" and subscriber bound to "@In") the publisher side only binds to and exchange topic, not a queue.  The "@In" side binds to a queue and the topic feeds the queue.  (I read the spring docks like 50 times and did a bunch of experiments to see what was creating what in rabbit)  Thus the naming convention I used for destination ConfigpropsUpdateExchange, I thought it was exactly the opposite.

PR #2105, opened 2019-07-17
makes school name and group name filters wider

esotrope:
image

image

image

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
@@ -137,6 +137,7 @@ public class WebSecurityConfiguration extends SamlWebSecurityConfiguration {
                         "/**/*.jpg",
                         "/**/*.jpeg",
                         "/**/*.ico",
+                        "/**/*.svg",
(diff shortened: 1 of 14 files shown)


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/WebSecurityConfiguration.java
```diff
@@ -137,6 +137,7 @@ protected void configure(final HttpSecurity http) throws Exception {
                         "/**/*.jpg",
                         "/**/*.jpeg",
                         "/**/*.ico",
+                        "/**/*.svg",
```
esotrope: should fix 404s nohemi was seeing

webapp/src/main/java/org/opentestsystem/rdw/reporting/user/UserContext.java
```diff
@@ -10,6 +10,7 @@
 
     private static final UserTenant DefaultUserTenant = UserTenant.builder()
             .logoutUrl("/logout")
+            .sandbox(false)
```
esotrope: anon users were seeing "Reporting System Sandbox" on the login page as well as access denied/session timeout pages which could be misleading

PR #2106, opened 2019-07-17
Task/merge sandbox and tenant controllers

esotrope:
not done yet but i wanted to get buy-in on these endpoints before i adjust the UI to use them.


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/TenantService.java
```diff
@@ -20,24 +20,27 @@
      *
      * @return The Default information and a list of all existing tenants for this Instance
      */
+    @PreAuthorize("hasAuthority('PERM_TENANT_READ')")
```
esotrope: added security to the service and todoned that todo

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/TenantService.java
```diff
@@ -46,20 +49,19 @@
...
     /**
      * @return a TenantConfiguration containing the default properties of the current tenant
      */
+    @PreAuthorize("hasAuthority('PERM_TENANT_READ')")
     TenantConfiguration getTenantDefaults();
 
-    /**
-     * @return a TenantConfiguration containing the default properties of the current tenant
-     */
-    TenantConfiguration getParentTenantDefaults(String key);
```
esotrope: there was no use of this in the UI

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantService.java
```diff
@@ -68,21 +68,28 @@
     }
 
     @Override
-    public void createTenant(TenantConfiguration tenantConfiguration) {
+    public TenantConfiguration createTenant(final TenantConfiguration tenantConfiguration) {
```
esotrope: this is out of convention - looks like right now the UI ignores what update/create return though - @ftdholbrook any reason we shouldn't return the result of update/create?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantService.java
```diff
@@ -68,21 +68,28 @@
...
+        // update the translationLocalization with specific
+        tenantConfiguration.getReporting().setTranslationLocation(
+                getTenantLocalization(
+                        tenantConfiguration.getLocalization(),
+                        tenantConfiguration.getReporting(),
+                        tenantConfiguration.getTenant().getKey()
+                )
+        );
+        tenantAdministrationRequestHandler.createTenant(tenantConfiguration);
+        return tenantConfiguration;
```
ftdholbrook: a language that evaluates if statements as expressions makes this style if block an anti-pattern.  I know it is common in Java, but I don't care for it.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantService.java
ftdholbrook: There is no result it is passed down stream to a message. This should be void

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantService.java
```diff
@@ -94,7 +101,7 @@ public void deleteTenant(final String tenantKey) {
     }
 
     @Override
-    public TenantConfiguration updateTenant(TenantConfiguration tenantConfiguration) {
+    public TenantConfiguration updateTenant(final TenantConfiguration tenantConfiguration) {
```
ftdholbrook: this should return void

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/web/TenantController.java
```diff
@@ -3,58 +3,60 @@
...
+    TenantConfiguration get(@PathVariable("key") final String key) {
+        return service.findActiveTenantConfigurationByKey(key)
+                .orElseThrow(NoSuchElementException::new);
     }
 
     @PostMapping
     @ResponseStatus(value = HttpStatus.CREATED)
-    void create(@RequestBody final TenantConfiguration tenantConfiguration) {
-        tenantService.createTenant(tenantConfiguration);
+    TenantConfiguration post(@RequestBody final TenantConfiguration configuration) {
```
ftdholbrook: this should be void

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/web/TenantController.java
```diff
@@ -3,58 +3,60 @@
...
+        return service.createTenant(configuration);
     }
 
     @PutMapping
     @ResponseStatus(value = HttpStatus.OK)
-    void put(@RequestBody final TenantConfiguration tenantConfiguration) {
-        tenantService.updateTenant(tenantConfiguration);
+    TenantConfiguration put(@RequestBody final TenantConfiguration configuration) {
```
ftdholbrook: dido: void

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/web/TenantController.java
```diff
@@ -3,58 +3,60 @@
...
-    public TenantController(TenantService tenantService) {
-        this.tenantService = tenantService;
+    TenantController(final TenantService service) {
+        this.service = service;
     }
 
     @GetMapping
-    public Set<TenantConfiguration> get() {
-        return tenantService.getAll(TenantType.TENANT);
+    Set<TenantConfiguration> get(@RequestParam final TenantType type) {
```
esotrope: allows the client to specify if they want just sandboxes or just tenants or all - which isn't a case

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/web/TenantController.java
```diff
@@ -3,58 +3,60 @@
...
     @GetMapping("/{key}")
-    TenantConfiguration getByKey(@PathVariable("key") final String key) {
-        final Optional<TenantConfiguration> activeTenantConfigurationByKey = tenantService.findActiveTenantConfigurationByKey(key);
-        if (activeTenantConfigurationByKey.isPresent()) {
-            return activeTenantConfigurationByKey.get();
-        }
-        throw new NoSuchElementException();
+    TenantConfiguration get(@PathVariable("key") final String key) {
+        return service.findActiveTenantConfigurationByKey(key)
+                .orElseThrow(NoSuchElementException::new);
```
ftdholbrook: nice much cleaner forgot about ::new

PR #2109, opened 2019-07-18
Feature/sandbox delete

ftdholbrook:
Sandbox Delete RP-477


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/TenantAdministrationProcessor.java
```diff
@@ -15,6 +15,8 @@
     String TenantCreationResponse = "tenant-TenantCreationResponse";
     String SandboxCreationRequest = "tenant-SandboxCreationRequest";
     String SandboxCreationResponse = "tenant-SandboxCreationResponse";
+    String SandboxDeleteRequest = "tenant-SandboxCreationRequest";
+    String SandboxDeleteResponse = "tenant-SandboxCreationResponse";
```
malaffoon: intentionally the same channel?

ftdholbrook: No, typo.  Good catch thank you!  Not actually using the responses yet, was planning on using them for the stats service, but am just calling that directly.  Should probably just pull them.

PR #2110, opened 2019-07-19
RP-603: two implementations RFC

gregcharles:
This is really an RFC, not a PR. The aggregate service has a CacheConfiguration class with one (empty) method that has a @CacheEvict annotation with a list of the cache names to clear.  One approach would be to use that. We'd need to create something similar in the other services that don't already have a CacheConfiguration. The other approach is to wire in the CacheManager and clear all the caches it manages.  I'm showing both approaches here.

Pros of CacheManager approach: we don't need to supply the cache names up front. All caches are cleared. The aggregate service keeps its cache names together in a constants class, but other services aren't so well organized. rdw-reporting-common  defines "translations" and "subjectDefinitions" caches, which means some cache names aren't contained directly in the service's module, but this method would catch those too. If new cache names were added for a new repo, the implementor wouldn't need to remember to go update this method. Also, we can probably  put this into common-multitenant and use it everywhere.

Pros of CacheConfiguration approach: we can define the caches to be cleared, just in case some shouldn't be cleared for this event. It seems, in a way, less brutal. However, we said we wanted this method to be brutal.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/ConfigurationChangeListener.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param environmentChangeEvent event payload
+     */
+//    @EventListener
+    public void environmentChangeEventClearCaches(final EnvironmentChangeEvent environmentChangeEvent) {
+        logger.warn("EnvironmentChangeEvent - {}", environmentChangeEvent);
+        try {
+            cacheManager.getCacheNames().forEach(name -> {
+                logger.warn("clearing " + name);
+                cacheManager.getCache(name).clear();
+            });
```
ftdholbrook: I prefer this approach.  Future cache users won't have to keep it updated.  I don't think it is a big deal to lose the cache on a config update.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/ApplicationConfiguration.java
```diff
@@ -2,6 +2,23 @@
...
+import org.springframework.cloud.context.config.annotation.RefreshScope;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.context.MessageSource;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.support.ReloadableResourceBundleMessageSource;
+import org.springframework.context.support.ResourceBundleMessageSource;
+import org.springframework.core.io.ResourceLoader;
+import org.springframework.integration.config.GlobalChannelInterceptor;
```
malaffoon: That's a lot of new imports ... are they just left over from various experiments?

malaffoon: Oh nvm, this is an RFC not a PR

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/ConfigurationChangeListener.java
malaffoon: I agree. Not that fond of the CacheConfiguration thing. Especially in the other services as you outlined.

PR #2112, opened 2019-07-19
prunes en.json

esotrope:


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1185,52 +1193,146 @@
...
+        "head": "Delete <strong>{{label}}</strong>?",
+        "body": "Are you sure you wish to delete the sandbox <strong>{{label}}</strong>?"
+      }
+    }
+  },
+
+  "tenants": {
+    "heading": {
+      "SANDBOX": "Sandboxes",
+      "TENANT": "Tenants"
```
esotrope: any objection to changing the verbiage here?

image

esotrope: reads better to me

image

ftdholbrook: looks better to me

esotrope: 👍

PR #2114, opened 2019-07-19
validation cleanups and improvements

ftdholbrook:
- check for required sandbox fields
- filter out failed sandboxes on creation exists check

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
@@ -1,6 +1,9 @@
 package org.opentestsystem.rdw.admin.multitenant.service.impl;
 
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.opentestsystem.rdw.admin.multitenant.TenantType;
+import org.opentestsystem.rdw.admin.multitenant.administration.TenantAdministrationStatus;
 import org.opentestsystem.rdw.admin.multitenant.model.TenantConfiguration;
 import org.opentestsystem.rdw.admin.multitenant.service.TenantConfigurationViewService;
 import org.opentestsystem.rdw.admin.multitenant.service.TenantValidationService;
@@ -8,7 +11,6 @@ import org.opentestsystem.rdw.admin.service.impl.RDWDataSourceNames;
 import org.opentestsystem.rdw.archive.ArchivePropertiesTenant;
 import org.opentestsystem.rdw.multitenant.Tenant;
 import org.opentestsystem.rdw.multitenant.datasource.DataSourceElementsTenant;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -57,16 +60,36 @@ public void sandboxUpdateValidation(TenantConfiguration tenantConfiguration) {
...
-        if (knownTenantConfigurations.containsKey(tenant.getKey())) {
+        //ignore failed for the purpose of existence detection
+        Map<String, TenantConfiguration> filteredConfigurations = knownTenantConfigurations.entrySet().stream()
+                .filter(entry -> entry.getValue().getAdministrationStatus().getTenantAdministrationStatus() != TenantAdministrationStatus.FAILED)
+                .map(entry -> ImmutablePair.of(entry.getKey(), entry.getValue()))
+                .collect(Collectors.toMap(ImmutablePair::getLeft, ImmutablePair::getRight));
+        if (filteredConfigurations.containsKey(tenant.getKey())) {
             String errorMsg = String.format("%s: %s, key: %s exists already. Use a unique key (or UPDATE the existing %s)",
                     tenantType, tenant.getName(), tenant.getKey(), tenantType);
-            logger.error(errorMsg);
```
esotrope: 👍 👍

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -57,16 +60,36 @@ public void sandboxUpdateValidation(TenantConfiguration tenantConfiguration) {
         verifyTenantExistsForUpdate(tenantConfiguration.getTenant(), knownTenantConfigurations);
     }
 
+    @VisibleForTesting
+    void verifySandboxRequiredFields(TenantConfiguration tenantConfiguration) {
```
esotrope: i do feel like we should just test the overall behavior of the method that is supposed to enforce this validation. otherwise this should be an expose utility method not on this service

ftdholbrook: How is that any different than package private, the utility class would sit in the same package and also be package private, nothing but this class should call this method, and nothing should call this except via the public service interface... that is just a bunch of extra noise.

ftdholbrook: Also if you don't test each of these in isolation, the public method that call them will explode in complexity.  This is the best way.  Really :-)

esotrope: interesting perspective. let the philosophical war begin! 
i'm merely coming at it from the perspective of tests need only cover public interfaces. but i'm willing to be convinced otherwise and don't feel strong enough about it to stop you : J

ftdholbrook: Philosophically is it ok to have a class that is a collaborator only be package-public?  (if not then why does that scope exist).  If that class only has a single collaborator would it make sense to merge the two classes (probably).  The end result would be this, package public collaborators  in-line.   The visible for testing annotation is a hint in the clearest way possible that these are collaborator methods.

PR #2115, opened 2019-07-19
network requests are now shared

esotrope:
before

image

after

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.html
+++ b/webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.html
@@ -77,7 +77,9 @@
                 class="fa fa-undo fa-md"
                 *ngIf="overrides.value !== overrides.originalValue && !readonly"
                 (click)="resetClicked(overrides)"
-                popover="{{ 'tenant-config.reset-info' | translate }}"
+                popover="{{
+                  'tenant-configuration-override-form.popover.reset' | translate
+                }}"
                 placement="left"
                 container="body"
                 triggers="mouseenter:mouseleave"
--- a/webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.html
+++ b/webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.html
@@ -1,22 +1,32 @@
-<ng-container *ngIf="(initialized$ | async); else loading">
-  <app-tenant-form
-    class="sandbox-config"
-    [value]="tenant$ | async"
-    [mode]="mode$ | async"
-    [writable]="writable$ | async"
-    [tenants]="tenants$ | async"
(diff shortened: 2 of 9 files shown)


webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.html
```diff
@@ -77,7 +77,9 @@
                 class="fa fa-undo fa-md"
                 *ngIf="overrides.value !== overrides.originalValue && !readonly"
                 (click)="resetClicked(overrides)"
-                popover="{{ 'tenant-config.reset-info' | translate }}"
+                popover="{{
+                  'tenant-configuration-override-form.popover.reset' | translate
```
esotrope: misc message cleanup that i missed

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.html
```diff
@@ -1,22 +1,32 @@
...
-    "
-    (create)="onCreate($event)"
-    (update)="onUpdate($event)"
-    (delete)="onDelete($event)"
-  ></app-tenant-form>
-</ng-container>
-<ng-template #loading>
-  <spinner></spinner>
-</ng-template>
+<!-- TODO investigate why this spinner display logic isn't working -->
```
esotrope: the loading spinner logic did not work in conjunction with the `share()` operator - have no clue why yet...

ftdholbrook:
This is cool I noticed it was spamming some of the same calls

webapp/src/main/webapp/src/app/admin/tenant/service/tenant.service.ts
```diff
@@ -51,15 +55,15 @@ export class TenantService {
...
       .get(DataSetsRoute)
       .pipe(catchError(ResponseUtils.throwError));
   }
 
   getAll(type: TenantType): Observable<TenantConfiguration[]> {
     const tenants$ = this.dataService.get(ResourceRoute, { params: { type } });
-    const defaults$ = this.dataService.get(DefaultsRoute);
-    const dataSets$ = this.dataService.get(DataSetsRoute);
+    const defaults$ = this.cachingDataService.get(DefaultsRoute);
+    const dataSets$ = this.cachingDataService.get(DataSetsRoute);
```
esotrope: @ftdholbrook can system defaults and data sets be cached until a browser tab is refreshed? i assume so but let me know if that is wrong

ftdholbrook: defaults changing would probably require restarting the web server.  Datasets same thing, any way to tie the cache to the session?

esotrope: yeah we technically can however we don't do client-side session caching for users atm - if we do session level caching we do it on the backend atm

PR #2116, opened 2019-07-19
validation inverted logic check for null bug

ftdholbrook:
tests don't do you any good if you test for the wrong thing... oops my bad.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
@@ -62,15 +62,15 @@ public class DefaultTenantValidationService implements TenantValidationService {
 
     @VisibleForTesting
     void verifySandboxRequiredFields(TenantConfiguration tenantConfiguration) {
-        if(tenantConfiguration.getTenant() == null) {
+        if (tenantConfiguration.getTenant() == null) {
             String errorMsg = "Required field 'tenant' missing";
             throw new IllegalArgumentException(errorMsg);
         }
-        if(!tenantConfiguration.getTenant().isSandbox()) {
+        if (!tenantConfiguration.getTenant().isSandbox()) {
             String errorMsg = "Required field 'tenant.sandbox' must be 'true'";
             throw new IllegalArgumentException(errorMsg);
         }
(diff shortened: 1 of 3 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -62,15 +62,15 @@ public void sandboxUpdateValidation(TenantConfiguration tenantConfiguration) {
...
             String errorMsg = "Required field 'tenant' missing";
             throw new IllegalArgumentException(errorMsg);
         }
-        if(!tenantConfiguration.getTenant().isSandbox()) {
+        if (!tenantConfiguration.getTenant().isSandbox()) {
             String errorMsg = "Required field 'tenant.sandbox' must be 'true'";
             throw new IllegalArgumentException(errorMsg);
         }
-        if(tenantConfiguration.getTenant().getSandboxDataset() != null) {
+        if (tenantConfiguration.getTenant().getSandboxDataset() == null) {
```
ftdholbrook: this is the bug

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationServiceTest.java
```diff
@@ -92,7 +92,7 @@ public void itShouldVerifySandboxRequiredSandboxDataset() {
         tenantValidationService.verifySandboxRequiredFields(TenantConfiguration.builder()
                 .tenant(Tenant.builder()
                         .sandbox(true)
-                        .sandboxDataset("TEST")
+                        .sandboxDataset(null)
```
ftdholbrook: and this the faulty test

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationServiceTest.java
```diff
@@ -294,6 +294,7 @@ private TenantConfiguration generateTenantConfiguration(String key, boolean isSa
                         .key(key)
                         .name(key + "-name")
                         .sandbox(isSandbox)
+                        .sandboxDataset("test-dataset")
```
ftdholbrook: making it correct caused the end to end test to fail, fixed it also

PR #2120, opened 2019-07-19
Feature/436 async tenant key validation

esotrope:
image


webapp/src/main/webapp/src/app/admin/tenant/mapper/tenant.mapper.ts
```diff
@@ -98,7 +98,7 @@ export function toTenantApiModel(tenant: TenantConfiguration): any {
 
   return {
     tenant: {
-      key,
+      key: key != null ? key.toUpperCase() : key,
```
esotrope: maybe the backend handles this now but figured i should do it here for good measure

PR #2122, opened 2019-07-19
RP-601: Clear caches on environment change event

gregcharles:
Real PR this time. I moved the listener to RDW_Common, and imported it into all services using either @Cacheable or @TenantCacheable.


ftdholbrook:
Looks good to me

PR #2124, opened 2019-07-20
Feature/598 tenant form

esotrope:
made usernames show up with the required filter so you could do matching when you get the username/password related error (which still doesn't work - need to change the way the form is initialized)

image


ftdholbrook:
I like it

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -320,6 +316,10 @@ export class TenantFormComponent implements OnChanges {
   }
 
   onKeyChange(code: string): void {
```
esotrope: only applies to tenants - think this also may be disconnected atm too

PR #2130, opened 2019-07-22
Feature/update tenant

ftdholbrook:
2nd try for this PR
- Get active tenant now pulls directly against the config server


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/TenantAdministrationRequestHandler.java
```diff
@@ -5,5 +5,7 @@
 public interface TenantAdministrationRequestHandler {
     void createTenant(TenantConfiguration tenantConfiguration);
     void createSandbox(TenantConfiguration tenantConfiguration);
+    void updateTenant(TenantConfiguration tenantConfiguration);
+    void updateSandbox(TenantConfiguration tenantConfiguration);
```
esotrope: doesn't tenant config have a sandbox flag? curious why this can't be just `update(TenantConf ..)`

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configserver/DefaultConfigServerClient.java
```diff
@@ -86,6 +62,24 @@ public String decryptPassword(final String password) {
         return configPassword(getConfigServerUriEndpoint("/decrypt"), password);
     }
 
+    public Optional<ApplicationTenantConfigurationPersistence> findApplicationTenantConfigurationPersistence(String tenantKey) {
+        final String tenantConfigEndpoint = configServerUri + "/*/*/master/tenant-" + tenantKey + "/application.yml";
+
+        try {
+            final ResponseEntity<String> responseEntity = restTemplate.getForEntity(tenantConfigEndpoint, String.class);
+            logger.debug("Successfully called {} returned {}.", configServerUri, responseEntity.getStatusCode());
+            return Optional.of(ymlMapper.readValue(responseEntity.getBody(), ApplicationTenantConfigurationPersistence.class));
```
esotrope: cool

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configserver/DefaultConfigServerClient.java
```diff
@@ -86,6 +62,24 @@ public String decryptPassword(final String password) {
         return configPassword(getConfigServerUriEndpoint("/decrypt"), password);
     }
 
+    public Optional<ApplicationTenantConfigurationPersistence> findApplicationTenantConfigurationPersistence(String tenantKey) {
+        final String tenantConfigEndpoint = configServerUri + "/*/*/master/tenant-" + tenantKey + "/application.yml";
```
ftdholbrook: this fixes a host of problems with update including thrashing on the passwords, encrypting and decrypting them.  It does introduce a few complications.  Jackson is more particular about naming consistence for properties.  Unable to call this and have a mix of kebab case and camel case.  I coded this to expect camel case.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/model/ApplicationTenantConfigurationPersistence.java
```diff
@@ -46,21 +44,22 @@
 
     //TODO: we will need to expand to accommodate other roots, this is the current set in ApplicationTenantConfiguration
     private Map<String, Map<String, Tenant>> tenantProperties;
-    private Map<String, Map<String, ReportingSystemProperties>> reporting;
-    private Map<String, Map<String, Map<String, DataSourceElements>>> datasources;
-    private Map<String, Map<String, ArchiveProperties>> archive;
+    private Map<String, Map<String, ReportingSystemPropertiesImpl>> reporting;
+    private Map<String, Map<String, Map<String, DataSourceElementsTenant>>> datasources;
+    private Map<String, Map<String, ArchivePropertiesTenant>> archive;
```
ftdholbrook: switch to concrete types to allow Jackson to deserialize

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/TenantAdministrationRequestHandler.java
ftdholbrook: So I made a split between each here at the handler layer because the flows start to get pretty different.  For update that happens to not be the case, but for create and delete it is.  Decided to keep the pattern for contextual consistency.

PR #2131, opened 2019-07-22
adds info popups to tenant form

esotrope:
image

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -1241,6 +1241,7 @@
     "field": {
       "key": {
         "label": "Key",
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1241,6 +1241,7 @@
     "field": {
       "key": {
         "label": "Key",
+        "info": "This is the internal identifier for the tenant, used in the configuration to uniquely identify the tenant. For consistency, it may be the same as the ID.",
```
esotrope: @malaffoon slightly tweaked your suggested messages and added one for tenant + data set - can you give me a grammar/clarity check here

PR #2133, opened 2019-07-22
RP-609, don't let user session interfere with sandbox role repository

malaffoon:
There is no existing test for this and it is just urgent enough that i'd like to get the fix in ASAP.


webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
```diff
@@ -50,12 +52,22 @@ public DefaultSandboxService(final SandboxRepository sandboxRepository,
...
+
         sandboxes.forEach(tenant -> {
             TenantContextHolder.setTenantId(tenant.getId());
             final List<SandboxRole> roles = sandboxRoleRepository.findAll();
             TenantContextHolder.clear();
             rolesBySandboxKey.put(tenant.getKey(), roles);
         });
+
+        // ... and restore the session context authentication
+        SecurityContextHolder.getContext().setAuthentication(authentication);
```
esotrope: sweet glad this worked

PR #2135, opened 2019-07-23
surround message handling code with try catch block

ftdholbrook:


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
```diff
@@ -57,7 +56,7 @@ public DefaultConfigpropsService(final ConfigpropsConsolidationService configpro
         //special case not using condenseConfigurations for RP-548 workaround
         HashSet<ReportingSystemSettings> set = Sets.newHashSet(reportingSystemSettings.values());
         if (set.size() > 1) {
-            logger.warn("More than one ReportingSystemSettings resolved count[{}] {}", set.size(), set);
+            logger.info("More than one ReportingSystemSettings resolved count[{}] {}", set.size(), set);
```
ftdholbrook: unrelated - this is important info, but occurs to often, lowering to INFO

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultTenantAdministrationRequestHandler.java
```diff
@@ -100,23 +100,28 @@ private void sendAdministrationMessage(TenantConfiguration tenantConfiguration,
...
-                updateTenantAsync(payload.getTenantConfiguration());
-                break;
-            case UPDATE_SANDBOX:
-                updateSandboxAsync(payload.getTenantConfiguration());
-                break;
+        try {
+            TenantAdministrationMessage payload = message.getPayload();
+            switch (payload.getTenantAdministrationRequestType()) {
+                case CREATE_TENANT:
+                    createTenantAsync(payload.getTenantConfiguration());
```
malaffoon: nit: might get the tenant configuration once outside switch since every branch uses it

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
malaffoon: Is there additional information that might indicate what to do about it? Or is it truly just informational?

ftdholbrook: The way to fix it is to make sure that tenant configurable data is uniform across all the services.  This message indicates they are not.  The way to fix is to update configurations. I can add more explanatory text, but his message hits multiple times whenever you get default tenant information.

PR #2136, opened 2019-07-23
RP-612 TenantAdmin - add id & key not null to validation

vgonzalez9:
re-add the validation on the tenant/sandbox keys & id are not null or empty.
added tests.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/RdwDatabaseType.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/RdwDatabaseType.java
@@ -6,7 +6,7 @@ public enum RdwDatabaseType {
     REPORTING("reporting_r", false),
     WAREHOUSE("warehouse_r", false);
 
-    final String datasourceNameHint;
+    public final String datasourceNameHint;
     final boolean isOlap;
 
     RdwDatabaseType(String datasourceNameHint, boolean isOlap) {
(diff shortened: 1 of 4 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -84,6 +84,12 @@ void verifySandboxRequiredFields(TenantConfiguration tenantConfiguration) {
      */
     private void verifyNotExists(final Tenant tenant, final Map<String, TenantConfiguration> knownTenantConfigurations) {
         String tenantType = tenant.isSandbox() ? "Sandox" : "Tenant";
+        if (tenant.getKey() == null || tenant.getKey().isEmpty() || tenant.getId() == null || tenant.getId().isEmpty()) {
```
ftdholbrook: I think this method is doing a bit too much.  maybe create a separate method for required field checks?  Something similar to verifySandboxRequiredFields

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/database/impl/RdwDatabaseType.java
```diff
@@ -6,7 +6,7 @@
     REPORTING("reporting_r", false),
     WAREHOUSE("warehouse_r", false);
 
-    final String datasourceNameHint;
+    public final String datasourceNameHint;
```
esotrope: should use a getter no? this makes it mutable

ftdholbrook: it is an enum, and final, can't be mutated

esotrope: ah good point. seeing the public just pushed my frazzled mind over the edge x J

PR #2138, opened 2019-07-23
moves effective report languages logic to ui

esotrope:


webapp/src/main/webapp/src/app/app-settings.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { toReportLanguages } from './app-settings';
+
+describe('toReportLanguages', () => {
```
esotrope: tests moved here

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AbstractReportingSystemProperties.java
```diff
@@ -1,26 +0,0 @@
-package org.opentestsystem.rdw.reporting.common.configuration;
-
-
-import com.fasterxml.jackson.annotation.JsonFormat;
-
-import java.util.List;
-
-import static com.google.common.collect.Lists.newArrayList;
-
-public abstract class AbstractReportingSystemProperties implements ReportingSystemProperties {
```
esotrope: this could go bye bye so that is always nice

PR #2140, opened 2019-07-23
fixes tests

esotrope:


reporting-service/src/test/resources/application-test.yml
```diff
@@ -13,10 +13,6 @@ reporting:
   state:
     code: CA
     name: California
-  student-fields:
-    economicDisadvantage: disabled
-    ethnicity: admin
-    gender: enabled
```
esotrope: these were getting clobbered by the defaults in application.yml...

PR #2145, opened 2019-07-24
makes toServerTenant null safe

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/model/tenants.ts
```diff
@@ -174,17 +174,17 @@ export function toServerTenant(tenant: TenantConfiguration): any {
     localizations: localization
   } = tenant;
 
-  return {
+  return valued({
     tenant: {
-      key: (key != null ? key.toUpperCase() : key).trim(),
-      id: id.trim(),
+      key: key != null ? key.toUpperCase().trim() : null,
+      id: id != null ? id.trim() : null,
```
esotrope: key and id are null when creating a sandbox and sending it to the server

PR #2146, opened 2019-07-24
remove incorrect validation checks

ftdholbrook:


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -73,8 +73,6 @@ void verifySandboxRequiredFields(TenantConfiguration tenantConfiguration) {
             String errorMsg = "Required field 'tenant' missing";
             throw new IllegalArgumentException(errorMsg);
         }
-        verifyTenantKeyAndIdFields(tenantConfiguration.getTenant());
-        verifyReportingRequiredFields(tenantConfiguration.getReporting());
```
ftdholbrook: these work for tenants, not sandboxes

PR #2153, opened 2019-07-25
Feature/persist aggregate reporting settings

ftdholbrook:
- Switch from Snake YAML to Jackson for serialization when writing to git
- add AggregateReportingPropertiesTenant to git persistence model

note: this change will cause explicit null fields to appear in the git /tenant-<KEY>/applciation.yml files, a additional set of changes to fix that is forthcoming.


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitTenantConfigurationPersistenceService.java
```diff
@@ -47,6 +35,8 @@ public GitTenantConfigurationPersistenceService(final TenantConfigurationPersist
                                                     final ConfigServerClient configServerClient) {
         this.tenantConfigurationPersistenceProperties = tenantConfigurationPersistenceProperties;
         this.configServerClient = configServerClient;
+        this.ymlMapper = new ObjectMapper(new YAMLFactory().disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER));
+        this.ymlMapper.findAndRegisterModules();
```
ftdholbrook: attempted to make this spring managed (along with matching object mapper used by config server client), but it caused spring to attempt to use this yml mapper as the default mapper.  Using in-line.

PR #2154, opened 2019-07-25
restores readonly display in update mode on tenants

esotrope:
image


webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.html
```diff
@@ -7,7 +7,6 @@
       #table
       [value]="_rows"
       dataKey="key"
-      [globalFilterFields]="['key', 'value']"
```
esotrope: dont need these settings anymore

PR #2156, opened 2019-07-25
Feature/field metadata

esotrope:
fixes boolean comparison


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.html
```diff
@@ -298,26 +298,26 @@ <h1 class="h2" heading>
...
             </label>
 
             <label>
               <input
-                id="configurationModified"
-                name="configurationModified"
-                formControlName="modified"
+                id="configurationRequired"
+                name="configurationRequired"
+                formControlName="required"
```
esotrope: moved required filter to the rightmost slot

PR #2160, opened 2019-07-25
adds saving/creating state to buttons when saving since slow

esotrope:
image

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -1405,8 +1405,8 @@
       "error": "An unknown error occurred when attempting to delete the sandbox",
       "success": "Sandbox was successfully deleted",
       "modal": {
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1404,8 +1405,8 @@
       "error": "An unknown error occurred when attempting to delete the sandbox",
       "success": "Sandbox was successfully deleted",
       "modal": {
-        "head": "Delete <strong>{{label}}</strong>?",
-        "body": "Are you sure you wish to delete the sandbox <strong>{{label}}</strong>?"
+        "head": "Delete <strong>{{label}} ({{code}})</strong>?",
+        "body": "Are you sure you wish to delete the sandbox <strong>{{label}} ({{code}})</strong>?"
```
esotrope: image

PR #2161, opened 2019-07-25
adds spinner to status for in-progress tenants

esotrope:
Not a real state but you get the picture

image


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-link/tenant-link.component.html
```diff
@@ -30,6 +36,7 @@ <h2 class="h3">
     </div>
   </div>
   <div class="body">
+    <!-- TODO add created date -->
```
esotrope: would also be nice to get the creation date time here but didn't see it in the current payload so i left a todo

PR #2163, opened 2019-07-26
Reduce payload of getAll tenant configuration call to only contain Tenant and Status

ftdholbrook:
This is WIP for the broader re-factor of the status service interactions.  It provides some decent quality of life improvements to the list view tenant and sandbox pages that I think it is worth merging now.

- Have TenantAdministrationStatusContext hold only Tenant instead of full tenant configuration
- Only populate tenant and status on tenant configuration

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantConfigurationViewService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantConfigurationViewService.java
@@ -33,6 +33,8 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.opentestsystem.rdw.admin.multitenant.administration.TenantAdministrationStatus.ACTIVE;
+
 /**
  * Classes to return tenant and sandbox properties from the RDW Actuator endpoint calls
  * This is the service the tenant & sandbox service use to retrieve default and tenant specific properties & settings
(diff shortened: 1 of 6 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantConfigurationViewService.java
```diff
@@ -153,58 +155,6 @@ public boolean exists(final String key) {
...
-
-        //we aren't storing this recreating based off convention
-        if (isSandbox) {
-            //sandbox only
-            tenantConfigurationBuilder.parentTenantKey(parentTenantKey(tenantKey));
-        }
-
-        return Optional.of(tenantConfigurationBuilder.build());
-    }
-
```
ftdholbrook: unrelated dead code removal

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantConfigurationViewService.java
```diff
@@ -226,11 +176,15 @@ private String getLanguageCode(final List<String> uiLanguages) {
...
-                .keySet()
+                .entrySet()
                 .stream()
-                .map(tenantKey -> ImmutablePair.of(tenantKey, findActiveTenantConfiguration(tenantKey)))
-                .filter(pair -> pair.getRight().isPresent())
-                .collect(Collectors.toMap(ImmutablePair::getLeft, pair -> pair.getRight().get()));
+                .map(entry -> ImmutablePair.of(entry.getKey(), TenantConfiguration.builder()
+                        .tenant(entry.getValue())
+                        .administrationStatus(TenantAdministrationStatusContext.builder()
+                                .tenantAdministrationStatus(ACTIVE) //TODO: temporary - next move to activeTenantsByType
```
ftdholbrook: this is WIP future version get all will be full driven from status service

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantConfigurationViewService.java
```diff
@@ -226,11 +176,15 @@ private String getLanguageCode(final List<String> uiLanguages) {
...
+                                .build())
+                        .build()))
+                .collect(Collectors.toMap(ImmutablePair::getLeft, ImmutablePair::getRight));
```
malaffoon: This feels like this could be simpler:
```
    .stream()
    .collect(Collectors.toMap(e -> e.getKey(), e -> TenantConfiguration.builder()...));
```

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/web/TenantControllerIT.java
```diff
@@ -2,14 +2,12 @@
```
malaffoon: 👍

PR #2164, opened 2019-07-26
RP-602: initial implementation

gregcharles:
I think this will work now. I found the TenantConfigurationViewService.defaultTenantConfiguration(), which says it returns, "Non-tenant specific default configuration instance wide," so I'm using that to get a non-tenant specific archive service, and using that to delete the s3 archive for the sandbox.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
@@ -1,40 +1,38 @@
 package org.opentestsystem.rdw.admin.multitenant.administration.impl;
 
 
-import com.google.common.annotations.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.concurrent.ConcurrentHashMap;
 import org.opentestsystem.rdw.admin.multitenant.administration.SandboxArchiveRemovalService;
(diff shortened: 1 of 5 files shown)


malaffoon:
Is it appropriate to use the tenant-aware archive service? Perhaps, like database deletion, we should be using a non-tenant s3 service?

gregcharles:
> Is it appropriate to use the tenant-aware archive service? Perhaps, like database deletion, we should be using a non-tenant s3 service?

Aha, yes, that would be much better. As I stepped through the deletion process, I saw that's in fact what I am getting, but only because no tenant key is resolved, and I still don't understand why that is or if it will always be the case. If there's a direct way to get a non-tenant s3 service, that would be the way to go. I'll see if I can figure that out based on the db case.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public DefaultSandboxArchiveRemovalService(final TenantConfigurationViewService tenantConfigurationViewService) {
+        this.tenantConfigurationViewService = tenantConfigurationViewService;
+
+    }
+
+    @Override
+    public void remove(final String tenantKey) {
+        final ArchiveProperties rootArchiveProperties =
+                tenantConfigurationViewService.defaultTenantConfiguration().getArchiveProperties();
```
ftdholbrook: this wouldn't work for full tenants because they can point to different buckets, but sandboxes are always against the instance configured root.  Looks good to  me.

PR #2166, opened 2019-07-26
RP-564 Write tests to verify TenantConfiguration Validation parentTen…

vgonzalez9:
…at and Reporting

finsh adding the sandbox validations - must be in sync with UI before merging

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
@@ -21,6 +21,9 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.stream.Collectors;
 
+/**
+ * Validation for tenantConfigurations created from the UI were only override data is expected
+ */
 @Service
 public class DefaultTenantValidationService implements TenantValidationService {
 
@@ -48,6 +51,8 @@ public class DefaultTenantValidationService implements TenantValidationService {
         verifySandboxRequiredFields(tenantConfiguration);
         final Map<String, TenantConfiguration> knownTenantConfigurations = tenantConfigurationViewService.allTenantConfigurations(TenantType.SANDBOX);
         verifyNotExists(tenantConfiguration.getTenant(), knownTenantConfigurations);
+        verifyParentTenantExists(tenantConfiguration.getParentTenantKey());
+        verifyReportingRequiredFields(tenantConfiguration.getReporting());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -48,6 +51,8 @@ public void sandboxCreationValidation(TenantConfiguration tenantConfiguration) {
         verifySandboxRequiredFields(tenantConfiguration);
         final Map<String, TenantConfiguration> knownTenantConfigurations = tenantConfigurationViewService.allTenantConfigurations(TenantType.SANDBOX);
         verifyNotExists(tenantConfiguration.getTenant(), knownTenantConfigurations);
+        verifyParentTenantExists(tenantConfiguration.getParentTenantKey());
+        verifyReportingRequiredFields(tenantConfiguration.getReporting());
```
malaffoon: Should this call go in `verifySandboxRequiredFields`?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -48,6 +51,8 @@ public void sandboxCreationValidation(TenantConfiguration tenantConfiguration) {
         verifySandboxRequiredFields(tenantConfiguration);
         final Map<String, TenantConfiguration> knownTenantConfigurations = tenantConfigurationViewService.allTenantConfigurations(TenantType.SANDBOX);
```
malaffoon: Might not be worth the work, but getting all the tenants twice (once for sandboxes, then once for tenants) seems wasteful.

vgonzalez9: actually we can just retrieve that tenant specifically to determine if it exists, will update.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantValidationService.java
```diff
@@ -64,6 +67,14 @@ public void sandboxUpdateValidation(TenantConfiguration tenantConfiguration) {
         verifyTenantTypeMatchesExisting(tenantConfiguration.getTenant(), knownTenantConfigurations);
     }
 
+    private void verifyParentTenantExists(final String parentTenantkey) {
+        if (!tenantConfigurationViewService.findActiveTenantConfiguration(parentTenantkey).isPresent()) {
+            String errorMsg = String.format("%s key does not exist. Please choose a existing tenant to create your sandbox.",
```
esotrope: nit: can we spell out the abbr. hurts my fragile brain to unpack words. actually my pref would be that we just inline the string in the `throw` statement since it isn't reused

PR #2170, opened 2019-07-30
switch from immutable sets to mutable lists

malaffoon:
The basic problem is that Spring's SnakeYAML doesn't do Sets and doesn't do immutable. So Immutable Sets are right out.

You can create custom resolvers or some such to deal with it but it is surprisingly convoluted. I was hoping i could convince everybody that this is okay. Even though i'm changing the implementation because of technical difficulties, something i advise against generally.

@ftdholbrook, given all your poking around with SnakeYAML, if you know a way to get the deserializer to call a custom setter, that would probably be a more elegant solution. I know how to do it for Jackson YAML, XML, and JSON but it didn't seem to work for Spring's SnakeYAML.

### DIFF
--- a/admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/OjbectMapperTest.java
+++ b/admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/OjbectMapperTest.java
@@ -1,10 +1,8 @@
 package org.opentestsystem.rdw.admin.multitenant.git;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
-import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
+import com.google.common.collect.ImmutableSet;
 import org.junit.Before;
 import org.junit.Test;
 import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
@@ -21,6 +19,7 @@ import java.util.Properties;
 import org.opentestsystem.rdw.admin.multitenant.model.ApplicationTenantConfigurationPersistence;
 import org.opentestsystem.rdw.admin.multitenant.model.TenantConfiguration;
 import org.opentestsystem.rdw.multitenant.Tenant;
+import org.opentestsystem.rdw.reporting.common.configuration.AggregateReportingPropertiesTenant;
 import org.opentestsystem.rdw.reporting.common.configuration.ReportingSystemPropertiesImpl;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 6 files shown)


admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/OjbectMapperTest.java
```diff
@@ -68,15 +63,40 @@ public void itShouldWriteReportingProperties() throws JsonProcessingException, B
...
+                "aggregateReporting:\n" +
+                "  tenants:\n" +
+                "    TS_S002:\n" +
+                "      assessmentTypes:\n" +
+                "        - \"sum\"\n" +
+                "        - \"iab\"\n" +
+                "        - \"sum\"";
+
+        final ApplicationTenantConfigurationPersistence out =
+                buildFromYaml(new ApplicationTenantConfigurationPersistence(), yaml);
```
malaffoon: Before this change, this line would fail just like the error in QA

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/OjbectMapperTest.java
```diff
@@ -68,15 +63,40 @@ public void itShouldWriteReportingProperties() throws JsonProcessingException, B
...
+        assertThat(out.getAggregateReporting().get("tenants").get("TS_S002").getAssessmentTypes())
+                .containsExactlyInAnyOrder("sum", "iab", "sum");
```
malaffoon: note this: when reading via SnakeYAML this does not dedupe.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AggregateReportingProperties.java
```diff
@@ -13,27 +12,28 @@
...
      */
-    Set<String> getAssessmentTypes();
+    List<String> getAssessmentTypes();
 
     /**
-     * The assessment types a statewide aggregate reporter can select for aggregate reports
+     * @return The assessment types a statewide aggregate reporter can select for aggregate reports
      */
-    Set<String> getStatewideUserAssessmentTypes();
+    List<String> getStatewideUserAssessmentTypes();
```
esotrope: yikes no sets? that is sad

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/OjbectMapperTest.java
```diff
@@ -68,15 +63,40 @@ public void itShouldWriteReportingProperties() throws JsonProcessingException, B
...
+
     private TenantConfiguration tenantConfiguration() {
         final ReportingSystemPropertiesImpl reporting = new ReportingSystemPropertiesImpl();
         reporting.setIrisVendorId("IRIS");
 
-        // TODO - add other things
+        final AggregateReportingPropertiesTenant aggregate = new AggregateReportingPropertiesTenant();
+        aggregate.setAssessmentTypes(ImmutableSet.of("sum", "iab"));
+
+        // TODO - add more data to test stuff
```
esotrope: idea here that we test de/serialization of ever field eventually?

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AggregateReportingProperties.java
ftdholbrook: We can't change how spring does binding. :-|

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/OjbectMapperTest.java
malaffoon: Yeah. I did just enough to test for the bugs so i could get this checked in.

common/src/main/java/org/opentestsystem/rdw/reporting/common/configuration/AggregateReportingProperties.java
malaffoon: Yeah, technically this is Spring not SnakeYAML - it is when it binds the values from the yaml parsing to the object: there's an if/else block that checks for Array, List, Map ... and that's it.

PR #2171, opened 2019-07-30
Feature/status service fix

ftdholbrook:
Fix for RP-605
- created new service "TenantAdministrationStatusSyncService" that keeps the status service in sync based of periodic pulls from the config props service
- DefaultTenantConfigurationViewService#allTenantConfigurations now driven entirely from status service


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/TenantAdministrationStatus.java
```diff
@@ -1,10 +1,24 @@
...
+    DELETE_PUBLISHING_CONFIGURATION_CHANGES,
+    DELETE_CLEARING_DATABASE_RESOURCES,
+    DELETE_CLEARING_S3_RESOURCES,
+    DELETE_CLEAR_CACHES,
+    DELETE_FAILED,
+    UPDATE_STARTED,
+    UPDATE_PERSISTING_CONFIGURATION,
+    UPDATE_PUBLISHING_CONFIGURATION_CHANGES,
+    UPDATE_CLEAR_CACHES,
+    UPDATE_FAILED,
```
esotrope: we'll need to make ui messages for each of these. shall i open a PR against this branch?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultTenantAdministrationStatusSyncService.java
```diff
@@ -0,0 +1,125 @@
...
+                                                        Long updateDelayMinutes) {
+        this.statusService = statusService;
+        this.configpropsService = configpropsService;
+        this.configServerClient = configServerClient;
+        this.createDelayMinutes = createDelayMinutes;
+        this.updateDelayMinutes = updateDelayMinutes;
+    }
+
+    @Scheduled(initialDelayString = "${app.multitenant.statusSync.initialDelay.:30000}",
+            fixedDelayString = "${app.multitenant.statusSync.fixedDelay.:30000}")
```
esotrope: this also means the UI should poll at approx this rate or a little slower then right?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/TenantAdministrationStatus.java
ftdholbrook: Yes @esotrope please update the UI.  Unless it is a huge change, just check directly into this branch and we will do it all in the same PR.

esotrope: sounds good - should be small

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultTenantAdministrationStatusSyncService.java
ftdholbrook: I don't think there is much utility in polling faster that 1/2 that speed (15 seconds).  It may be the boundaries align such that you effectively were polling once a minute (59.9 seconds) if you went all the way up to 30 seconds.

malaffoon: Are those trailing dots correct, i've never seen that before?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configuiration/TenantAdministrationConfiguration.java
```diff
@@ -42,15 +45,15 @@
...
-    @Value("${tenant.tenantAdministration.statusService.expireAfterDays:4}")
-    private Integer statusServiceExpireAfterDays;
-
     @Value("${spring.cloud.config.uri}")
     private String configServerUri;
 
+    @Value("${app.multitenant.statusSync.createDelayMinutes:5}")
+    private Long createDelayMinutes;
+
+    @Value("${app.multitenant.statusSync.updateDelayMinutes:1}")
```
malaffoon: Seems inconsistent to have these configured as minutes while other `statusSync` values are millis. Probably not worth changing, i'm just curious why you chose to do these in minutes.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configuiration/TenantAdministrationConfiguration.java
```diff
@@ -42,15 +45,15 @@
...
-    @Value("${tenant.tenantAdministration.statusService.maximumSize:100}")
-    private Integer statusServiceMaximumSize;
-
-    @Value("${tenant.tenantAdministration.statusService.expireAfterDays:4}")
-    private Integer statusServiceExpireAfterDays;
-
     @Value("${spring.cloud.config.uri}")
     private String configServerUri;
 
+    @Value("${app.multitenant.statusSync.createDelayMinutes:5}")
```
malaffoon: You have these as members of the Configuration class, that seems ... different. Why are they not just @Value params in the @Bean method?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultTenantAdministrationStatusSyncService.java
ftdholbrook: typo from bulk search replace, I'll fix

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configuiration/TenantAdministrationConfiguration.java
ftdholbrook: minutes are the scale of the pause, so saving them in minutes.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configuiration/TenantAdministrationConfiguration.java
ftdholbrook: No reason, both ways work equally well.  I'm happy to swap them.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -911,11 +911,25 @@
     },
     "subject-select-label": "Subject",
     "tenant-status": {
-      "STARTED": "Started",
-      "CREATING_DATABASES": "Creating Databases",
-      "PERSISTING_CONFIGURATION": "Persisting Configurations",
-      "NOTIFYING_CONFIG_SERVER": "Notifying Configuration Server",
-      "FAILED": "Failed",
+      "CREATE_STARTED": "Creating",
```
esotrope: @malaffoon @ftdholbrook please double check these status messages and make sure they make sense

webapp/src/main/webapp/src/app/admin/tenant/pages/tenants/tenants.component.ts
```diff
@@ -13,11 +13,11 @@ import { filter, flatMap, map, takeUntil } from 'rxjs/operators';
...
 import { tap } from 'rxjs/internal/operators/tap';
 import { ordering } from '@kourge/ordering';
-import { byString, join } from '@kourge/ordering/comparator';
-import { completedTenantStatuses } from '../../model/tenant-statuses';
+import { byString } from '@kourge/ordering/comparator';
+import { completed } from '../../model/tenant-statuses';
 import { TenantModalService } from '../../service/tenant-modal.service';
 
-const pollingInterval = 1000;
+const pollingInterval = 1000 * 15; // backend polls at 30s
```
esotrope: now polling at 15s in UI (there is also logic to make sure the previous request finishes before the next is fired so it will not exactly fire every 15s)

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -911,11 +911,25 @@
...
+      "DELETE_PUBLISHING_CONFIGURATION_CHANGES": "Publishing Configuration Changes",
+      "DELETE_CLEARING_DATABASE_RESOURCES": "Deleting Database Resources",
+      "DELETE_CLEARING_S3_RESOURCES": "Deleting Archive Resources",
+      "DELETE_CLEAR_CACHES": "Clearing Caches",
+      "DELETE_FAILED": "Deletion Failed",
+      "UPDATE_STARTED": "Updating",
+      "UPDATE_PERSISTING_CONFIGURATION": "Saving Configurations",
+      "UPDATE_PUBLISHING_CONFIGURATION_CHANGES": "Publishing Configuration Changes",
+      "UPDATE_CLEAR_CACHES": "Clearing Caches",
+      "UPDATE_FAILED": "Update Failed",
```
ftdholbrook: @esotrope LGTM

PR #2172, opened 2019-07-30
fixes lowercasing during form submission

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/model/tenants.ts
```diff
@@ -18,7 +18,9 @@ function lowercase(object: { [key: string]: any }): { [key: string]: any } {
     object,
     (result: any, value: any, key: string) => {
       result[key] =
-        fieldConfiguration(key) && value != null && typeof value === 'string'
+        fieldConfiguration(key).lowercase &&
```
esotrope: doh - before i was lowercasing all string values sent to the backend...
thanks for the find @vgonzalez9

PR #2173, opened 2019-07-30
Defect/626 state input

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/component/property-override-tree-table/property-override-tree-table.component.ts
```diff
@@ -157,6 +157,18 @@ export class PropertyOverrideTreeTableComponent
...
+  readonlyValue(property: ConfigurationProperty): any {
+    const value = this.formGroup.value[property.key];
+    if (value != null) {
+      return value;
+    }
+    const defaultValue = this.defaults[property.key];
+    if (defaultValue != null) {
+      return defaultValue;
+    }
+    return '';
```
esotrope: unrelated fix for rendering readonly values. @ftdholbrook i'm curious if we set password for tenant will in come back as **** in the readonly view? my guess is it will but i wasn't able to confirm locally

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.html
```diff
@@ -275,6 +225,56 @@ <h1 class="h2" heading>
       </div>
     </ng-container>
 
+    <div
```
esotrope: moved tenant and data set form fields above the name field since the default name will change when you change the base tenant

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -251,11 +252,10 @@ function setDefaultDatabaseNameAndUsername(
     }, [])
     // for every datasource name apply defaults
     .forEach(source => {
-      const usernameControl =
-        formGroup.controls[`datasources.${source}.username`];
-      if (usernameControl.pristine) {
-        usernameControl.patchValue(defaultUsername);
-      }
+      patch(
```
esotrope: when you change tenant now the sandbox label and config props will correctly change to match that tenants overrides

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -54,8 +60,14 @@ export class TenantComponent implements OnDestroy {
...
     private translationLoader: RdwTranslateLoader
   ) {
-    const sandboxDataSets$ = this.service.getSandboxDataSets().pipe(share());
-    const tenants$ = this.service.getAll('TENANT').pipe(share());
+    const sandboxDataSets$ = this.service.getSandboxDataSets().pipe(
+      map(values => values.slice().sort(byLabel)),
+      share()
+    );
+    const tenants$ = this.service.getAll('TENANT').pipe(
+      map(values => values.slice().sort(byLabel)),
```
esotrope: tenants and data sets are now sorted by label on the sandbox form

webapp/src/main/webapp/src/app/shared/support/support.ts
```diff
@@ -553,7 +553,18 @@ export function rightDifference(a: any, b: any): any {
  */
 export function valued(value: any): any {
   return Object.entries(value).reduce((valued, [key, value]) => {
-    if (value != null && !(typeof value === 'string' && value.length === 0)) {
+    if (
+      // null or undefined
+      value != null &&
+      // empty strings
+      !(typeof value === 'string' && value.length === 0) &&
+      // empty maps/objects
```
esotrope: the fix for the state being blank. - backend returns `{}` instead of `null` for the state when there was no override i guess

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-tree-table/property-override-tree-table.component.ts
ftdholbrook: Defaults (sourced from configprops) are masked with ****, if they come from the config server (get tenant for edit/view) they are encrypted.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1535,6 +1535,7 @@
     },
     "no-results": "There are no configuration properties matching your search",
     "boolean": {
+      "undefined": "False",
```
esotrope: needed for switching tenants

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-tree-table/property-override-tree-table.component.ts
esotrope: 👍

PR #2175, opened 2019-07-31
adds datasource consistency validations in UI

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.validators.spec.ts
```diff
@@ -0,0 +1,125 @@
+import {
+  oneDatabasePerDataSource,
+  onePasswordPerUser,
+  uniqueDatabasePerInstance
+} from './tenant-form.validators';
+import { FormControl, FormGroup } from '@angular/forms';
+
+describe('onePasswordPerUser', () => {
```
esotrope: i know it works because tests : J

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.validators.ts
```diff
@@ -59,3 +59,83 @@ export function onePasswordPerUser(
...
+      if (names == null) {
+        namesBySource[source] = [name];
+      } else if (!names.includes(name)) {
+        names.push(name);
+      }
+    }
+  ).filter(([, names]) => names.length > 1);
+}
+
+function duplicateNamesAndSources(
```
esotrope: this data crunching is surprisingly similar between validation methods i was surprised

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1488,7 +1488,9 @@
       "configurations": {
         "label": "Configurations",
         "error": {
-          "onePasswordPerUser": "Please provide only one password for a each unique data source username provided below."
+          "onePasswordPerUser": "Please provide only one password for a each unique data source username provided below.",
+          "uniqueDatabasePerInstance": "Please provide unique database and schemaSearchPath for each unique datasource.",
+          "oneDatabasePerDataSource": "Please provide only one database or schemaSearchPath for each unique datasource."
```
esotrope: explaining these to someone who wants to deviate from the defaults we give them is going to be fun : j

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.validators.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+describe('uniqueDatabasePerInstance', () => {
+  it('should return null when form is valid', () => {
+    expect(
+      uniqueDatabasePerInstance(
+        new FormGroup({
+          'datasources.reporting_ro.urlParts.database': new FormControl('a'),
+          'datasources.reporting_rw.urlParts.database': new FormControl('a'),
+          'datasources.warehouse_ro.urlParts.database': new FormControl('b'),
+          'datasources.warehouse_rw.urlParts.database': new FormControl('b'),
+          'datasources.olap_ro.schemaSearchPath': new FormControl('c'),
```
ftdholbrook: the more likely scenario here is that olap_ro/olap_rw share the same database name as reporting_ro/reporting_rw.  That is valid, olap falls into it's own bucket of uniqueness.   I don't think I see a case for that.

PR #2176, opened 2019-07-31
safely skip getting roles for sandboxes that are broken

malaffoon:


webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
```diff
@@ -39,8 +44,8 @@ public DefaultSandboxService(final SandboxRepository sandboxRepository,
                 .comparing(sandbox -> Optional.ofNullable(sandbox.getLabel()).orElse(""));
 
         return sandboxes.stream()
-                .map(tenant ->
-                        Sandbox.builder()
+                .filter(tenant -> rolesBySandboxKey.containsKey(tenant.getKey()))
```
malaffoon: If there are no roles for a sandbox it seems pointless to return it; however, i defer to @ftdholbrook or @esotrope if they feel we can/should return the sandbox with no roles.

ftdholbrook: I think that is correct.   Really we should get the sandbox list from the admin service, not the way we are doing it here.  we could decorate it here with the roles.

ftdholbrook: ... I guess the only down side is when you are waiting on a migration, that's the only sign we have right now that is what is going on.

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
```diff
@@ -58,12 +63,16 @@ public DefaultSandboxService(final SandboxRepository sandboxRepository,
...
-            final List<SandboxRole> roles = sandboxRoleRepository.findAll();
-            TenantContextHolder.clear();
-            rolesBySandboxKey.put(tenant.getKey(), roles);
-        });
+        for (final Tenant sandbox : sandboxes) {
+            TenantContextHolder.setTenantId(sandbox.getId());
+            try {
+                rolesBySandboxKey.put(sandbox.getKey(), sandboxRoleRepository.findAll());
+            } catch (final Exception e) {
+                logger.warn("Failed to retrieve sandbox roles for {}: {}", sandbox.toString(), e.getMessage());
```
esotrope: does this happen after filtering?

PR #2185, opened 2019-08-02
adds cancel button. fixes issue with form values changing on submit. …

esotrope:
Bugfix: Form values no longer change when submitting the form*

Added cancel button

image

Added mask when submitting

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
@@ -416,6 +416,14 @@ export class TenantFormComponent implements OnChanges, OnDestroy {
   }
 
   ngOnChanges(changes: SimpleChanges): void {
+    const { submitting } = changes;
+
+    // do not re-initialize when submitting state changes
+    // this also shows that the form controls should lie outside of this form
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -416,6 +416,14 @@ export class TenantFormComponent implements OnChanges, OnDestroy {
   }
 
   ngOnChanges(changes: SimpleChanges): void {
```
esotrope: trouble with using this for convenience is that not all params changes should trigger a re-configuration of the page

webapp/src/main/webapp/src/app/shared/loading/spinner.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="loader">
   <div class="inner">
     <i class="fa fa-spinner fa-pulse fa-fw" [ngClass]="iconStyles"></i
-    ><span *ngIf="text" class="text">{{ text }}</span>
+    ><span class="text">{{ text }}<ng-content></ng-content></span>
```
esotrope: made this more flexible by supporting content transclusion

PR #2186, opened 2019-08-02
RP-650: fix sandbox delete failure

gregcharles:
decrypt password before using it

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
@@ -3,10 +3,13 @@ package org.opentestsystem.rdw.admin.multitenant.administration.impl;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.BeanUtils;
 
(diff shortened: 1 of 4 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
```diff
@@ -3,33 +3,48 @@
...
 
     @Override
     public void remove(final String tenantKey) {
         final ArchiveProperties rootArchiveProperties =
                 tenantConfigurationViewService.defaultTenantConfiguration().getArchiveProperties();
 
-        final ArchiveService rootArchiveService = new ArchiveServiceDecorator(rootArchiveProperties);
+        // Make a copy of the root properties and decrypt the key
+        final ArchivePropertiesRoot copiedArchiveProperties = new ArchivePropertiesRoot();
+        BeanUtils.copyProperties(rootArchiveProperties, copiedArchiveProperties);
```
esotrope: this feels strange but i'm not sure a better solution besides exposing a `copy()` method which is also strange for a service to have..

malaffoon: If we had a proper builder ...

PR #2189, opened 2019-08-03
RP-650: fix issues with delete location and error handling

gregcharles:
This should be the last one. I tested the delete("") in sample code and it was successful.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
@@ -36,14 +36,13 @@ public class DefaultSandboxArchiveRemovalService implements SandboxArchiveRemova
             // For safety, verify resolved archived properties contain a valid sandbox prefix
             final String prefix = archivePropertiesResolver.getPathPrefix();
             if (StringUtils.isBlank(prefix) || !prefix.toLowerCase().contains("_s")) {
-                throw new IllegalStateException(
-                        "Failed to resolve sandbox-specific archive service. Found prefix: " + prefix);
+                throw new IllegalStateException("Invalid path prefix: " + prefix);
             }
 
             // Delete all objects contained in the sandbox specific archive.
             archiveService.delete("");
         } catch (final Exception e) {
-            logger.info("Failed to remove archives for sandbox id {}: {}", tenantKey, e.getMessage());
+            logger.info("Failed to remove archives for sandbox with tenant key {}: {}", tenantKey, e.getMessage());
         } finally {
             TenantContextHolder.clear();
         }
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
@@ -37,13 +37,13 @@ public class DefaultSandboxArchiveRemovalService implements SandboxArchiveRemova
             final String prefix = archivePropertiesResolver.getPathPrefix();
             if (StringUtils.isBlank(prefix) || !prefix.toLowerCase().contains("_s")) {
                 throw new IllegalStateException(
-                        "Cannot resolve sandbox-specific archive service. Found prefix: " + prefix);
+                        "Failed to resolve sandbox-specific archive service. Found prefix: " + prefix);
             }
 
             // Delete all objects contained in the sandbox specific archive.
             archiveService.delete("");
         } catch (final Exception e) {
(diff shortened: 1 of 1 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
```diff
@@ -31,15 +31,21 @@ public DefaultSandboxArchiveRemovalService(ArchivePropertiesResolver archiveProp
...
+                        "Cannot resolve sandbox-specific archive service. Found prefix: " + prefix);
+            }
 
-            TenantContextHolder.setTenantId(tenant.getId());
-            archiveService.delete(archivePropertiesResolver.getPathPrefix());
-            TenantContextHolder.clear();
+            // Delete all objects contained in the sandbox specific archive.
+            archiveService.delete("");
         } catch (final Exception e) {
             logger.info("Cannot remove archives for sandbox id {}: {}", tenantKey, e.getMessage());
```
malaffoon: nit: perhaps "Failed to remove ..."

For me "cannot" means it isn't going to try, where "failed to" implies it tried and failed. Not gonna lose sleep over it though.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
```diff
@@ -31,15 +31,21 @@ public DefaultSandboxArchiveRemovalService(ArchivePropertiesResolver archiveProp
...
             final Tenant tenant = tenantProperties.getTenants().get(tenantKey);
+            TenantContextHolder.setTenantId(tenant.getId());
 
-            // TODO: remove log once delete is proven
-            logger.info("s3 delete: " + StringUtils.abbreviate(archivePropertiesResolver.getS3SecretKey(), 11));
+            // For safety, verify resolved archived properties contain a valid sandbox prefix
+            final String prefix = archivePropertiesResolver.getPathPrefix();
+            if (StringUtils.isBlank(prefix) || !prefix.toLowerCase().contains("_s")) {
+                throw new IllegalStateException(
+                        "Cannot resolve sandbox-specific archive service. Found prefix: " + prefix);
```
malaffoon: nit: what is this message saying, exactly? It looks to me like it has the archive service but it isn't the right one? Or it isn't a sandbox one with a real prefix? Perhaps "Cannot remove resources: archive service for " + tenant.getId() + " has an invalid path prefix " + prefix

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
```diff
@@ -37,13 +37,13 @@ public void remove(final String tenantKey) {
             final String prefix = archivePropertiesResolver.getPathPrefix();
             if (StringUtils.isBlank(prefix) || !prefix.toLowerCase().contains("_s")) {
                 throw new IllegalStateException(
-                        "Cannot resolve sandbox-specific archive service. Found prefix: " + prefix);
+                        "Failed to resolve sandbox-specific archive service. Found prefix: " + prefix);
```
malaffoon: Thanks for the changes, i think tenant id or key would be useful in the message.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
gregcharles: This probably can't happen. I was mostly concerned with a blank coming through, because I think that would cause everything in the bucket to be wiped. A non-blank, but incorrect prefix is even more unlikely, but I'll change the error message anyway.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
gregcharles: OK, good tip! I'll change the wording.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/impl/DefaultSandboxArchiveRemovalService.java
gregcharles: We actually catch and log everything so an s3 failure doesn't end up causing other parts of the delete to fail. This would normally happen if the location didn't exist in s3 or there was some problem with the @#$%& s3 secret key.

So:

logger.info("Failed to remove archives for sandbox with tenant key {}: {}", tenantKey, e.getMessage());

and I changed that message to be, "Invalid path prefix: " + prefix.

PR #2192, opened 2019-08-05
fixes some issues with value comparisons

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/model/fields.ts
```diff
@@ -35,9 +31,19 @@ const inputTypeByPropertyDataType: { [key: string]: InputType } = <
 };
 
 const identity = value => value;
+
 // TODO this should really be based on key as sometimes arrays are sets and sometimes lists
 const normalizeByInputType: { [key: string]: (value: any) => any } = {
-  input: value => emptyToNull(value),
+  // null, blank to null, number to string
```
esotrope: before `"2016"` was not considered equal to `2016`

webapp/src/main/webapp/src/app/admin/tenant/model/fields.ts
```diff
@@ -121,13 +127,9 @@ export function isModified(
   value: any,
   originalValue: any
 ): boolean {
-  if (value == null) {
-    return false;
-  }
-  if (typeof value === 'string' && isBlank(value)) {
-    return false;
-  }
-  if (isArray(value) && value.length === 0) {
```
esotrope: this was causing the issue - i'm not even sure what the spirit of this logic was at this point - made sure to add comments for the spirit of the other not so straightforward logic here

PR #2195, opened 2019-08-05
adds ordering of languages and restricts allowed languages

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/model/field-configurations.ts
```diff
@@ -42,13 +46,20 @@ const stateOptions = () =>
...
-  languageCodes.map(value => ({
-    value,
-    label: translateService.instant(`common.language.${value}`)
-  }));
+  ['en']
+    .map(value => ({
+      value,
+      label: translateService.instant(`common.language.${value}`)
+    }))
+    .sort(byLabel);
```
esotrope: fixes ordering

webapp/src/main/webapp/src/app/admin/tenant/model/field-configurations.ts
```diff
@@ -42,13 +46,20 @@ const stateOptions = () =>
...
 
-const primitiveSetEquals = (a, b) =>
-  isEqual(a != null ? a.slice().sort() : [], b != null ? b.slice().sort() : []);
+const reportLanguageOptions = translateService =>
+  ['en', 'es']
```
esotrope: i recall there being some question about this - do we actually support Spanish in UAT for printed reports?

malaffoon: UAT does have a `es.json` file in the configuration so, yeah, i guess that counts.

PR #2197, opened 2019-08-06
Defect/666 tenant states

esotrope:
we'll need to add this to the config repo admin-service app yml before this goes in to avoid breakage:

```yaml
art-client:
  url: 
  oauth2:
    access-token-uri:
    client-id:
    client-secret:
    username: 
    password: 
```


admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/ArtClient.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestClientException;
+
+import java.util.List;
+
+/**
+ * ART client providing access to organization data.<br/>
+ */
+@Component
+public class ArtClient {
```
esotrope: more or less copy-pasted from ingest's task-service

admin-service/src/test/java/org/opentestsystem/rdw/admin/service/impl/ArtBackedStateServiceIT.java
```diff
@@ -0,0 +1,34 @@
...
+        ObjectMapper.class
+})
+public class ArtBackedStateServiceIT {
+
+    @Autowired
+    ArtBackedStateService service;
+
+    @Ignore("This test can be run to confirm ART connection")
+    @Test
+    public void itShouldConnect() {
```
esotrope: i suppose i should add more tests than this ignored one : j

webapp/src/main/webapp/src/app/admin/tenant/model/data/state.ts
```diff
@@ -1,9 +1,4 @@
-export interface State {
-  name: string;
-  abbreviation: string;
-}
-
-export const states: State[] = [
+export const states: any[] = [
```
esotrope: this can be removed now

admin-service/src/main/resources/application.yml
```diff
@@ -218,6 +218,15 @@ archive:
...
   #s3-sse:
 
+art-client:
+  url: https://art-deployment.sbtds.org/rest/
+  oauth2:
+    access-token-uri: https://sso-deployment.sbtds.org:443/auth/oauth2/access_token?realm=/sbac
+    client-id: sbacdw
+    client-secret: ${rdw.reporting.art-client.oauth2.client-secret:}
+    username: prime.user@example.com
+    password: ${rdw.reporting.art-client.oauth2.password:}
```
esotrope: @gregcharles @malaffoon @ftdholbrook  devs are going to have to add some new props to their env for this connection to work locally:
```
RDW_REPORTING_ART_CLIENT_OAUTH2_CLIENT_SECRET=...
RDW_REPORTING_ART_CLIENT_OAUTH2_PASSWORD=...
```

ftdholbrook: @esotrope Can you update the Readme that has those env settings with that?

esotrope: sure can

PR #2198, opened 2019-08-07
Defect/503 null writing purpose

esotrope:
Still some testing to do but hoping to get some early eyes here.

**Webapp**

1. Now shows tab in student responses view.
image

2. Now shows "Not Specified" instead of blank or null
image

**Printed Reports**

Still working on some testing and screenshots for this


common/src/main/resources/i18n/en.json
```diff
@@ -367,27 +367,28 @@
         "Organization": "Organization / Purpose",
         "Unknown": {
           "Conventions": {
-            "0": "Unknown.Conventions.0",
-            "1": "Unknown.Conventions.1",
-            "2": "Unknown.Conventions.2"
+            "0": "",
+            "1": "",
+            "2": ""
```
esotrope: was thinking we shouldn't bother showing any description for these states but maybe @malaffoon you can think of something applicable?

webapp/src/main/webapp/src/app/assessments/items/item-tab.component.ts
```diff
@@ -118,6 +118,13 @@ export class ItemTabComponent implements AfterViewInit {
     this.trackAnalyticsEvent('ItemTabSelection', tabName);
   }
 
+  hasWritingType(item: AssessmentItem): boolean {
+    return (
+      item.performanceTaskWritingType != null ||
```
esotrope: really the below line is the true "source of truth" for this data but this attribute shows us how to categorize it which is why we rely on it in most cases for presence checks

webapp/src/main/webapp/src/app/assessments/model/student-responses-item.model.ts
```diff
@@ -0,0 +1,16 @@
+import { AssessmentItem } from './assessment-item.model';
+import { WritingTraitScores } from './writing-trait-scores.model';
+
+/**
+ * This model wrapper class contains both an AssessmentItem and
+ * calculated correctness value.
+ *
+ * TODO and make this a view model for item tab because that is what it is scoped to
+ */
+export interface StudentResponsesAssessmentItem {
```
esotrope: was a class and was in another package that could not be shared with the writing trait score view

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.html
```diff
@@ -142,6 +147,7 @@
         <td [attr.colspan]="columns.length">
           <item-tab
             [item]="itemView.item"
+            [responsesAssessmentItem]="itemView.responsesAssessmentItem"
```
esotrope: allows for writing trait score tab to display under items in Writing Trait Scores exam result view. This was not enabled before. I kept it disabled because i noticed the data was not complete when i enabled it.

image

common/src/main/resources/i18n/en.json
malaffoon: I'm sure SB has some text you could put in there. Have to ask Greg or Ray to get that though.

common/src/main/resources/i18n/en.json
```diff
@@ -367,27 +367,28 @@
...
+            "1": "",
+            "2": "",
+            "3": "",
+            "4": ""
           }
         },
         "outof": "of {0}",
-        "purpose": "Writing Purpose: {0}"
+        "purpose": "Writing Purpose: {0}",
+        "purpose-not-specified": "Writing Purpose: Not Specified"
```
esotrope: image

esotrope: image

PR #2203, opened 2019-08-08
fix warnings, remove unused component

malaffoon:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/sandbox/SandboxAuthenticationProvider.java
```diff
@@ -9,13 +8,9 @@
...
 
     private final SandboxUserDetailsService userDetailsService;
-    private final SandboxRoleRepository sandboxRoleRepository;
 
-    public SandboxAuthenticationProvider(
-            final SandboxUserDetailsService userDetailsService,
-            final SandboxRoleRepository sandboxRoleRepository) {
+    public SandboxAuthenticationProvider(final SandboxUserDetailsService userDetailsService) {
         this.userDetailsService = userDetailsService;
-        this.sandboxRoleRepository = sandboxRoleRepository;
```
malaffoon: Landon, am i missing something, this doesn't appear to be used?

Seems like it got moved up to the service instead of being used here.

esotrope: honestly i have no recollection so if it is unused cut it. maybe it was put in there because i thought we needed to do role assertions? not sure : j

PR #2204, opened 2019-08-08
adds exam ingest validation tenant admin config section

esotrope:
image


webapp/src/main/webapp/src/app/admin/tenant/model/data/configurable-data-elements.ts
```diff
@@ -0,0 +1,28 @@
+export const configurableDataElements = [
+  'FirstName',
```
esotrope: should this order be respected in the list or should these be sorted alphabetically?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1527,7 +1527,8 @@
       "datasources": "Data Sources",
       "task": "Task",
       "archive": "Archive",
-      "aggregate": "Aggregate"
+      "aggregate": "Aggregate",
+      "validation": "Exam Ingest Validation"
```
esotrope: @malaffoon does this name make sense?

webapp/src/main/webapp/src/app/admin/tenant/model/data/configurable-data-elements.ts
malaffoon: Probably alphabetical. There's no "right" order and alphabetical makes it easier to find things.

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: Sure, although you could probably make a bit more generic, just "Ingest Validation". A couple of the validations are for test packages.

PR #2205, opened 2019-08-08
makes the field hidden for sandbox forms

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/model/field-configurations.ts
```diff
@@ -227,6 +227,7 @@ export const fieldConfigurationsByKey: { [key: string]: FieldConfiguration } = {
   }, {}),
   'validation.requiredDataElements': {
     dataType: 'enumeration-list',
-    options: examProcessorRequiredDataElements
+    options: examProcessorRequiredDataElements,
+    hidden: sandboxHidden
```
esotrope: shouldn't see it on the sandbox form - almost forgot

PR #2212, opened 2019-08-09
RP-473 - task service multi-tenant aware 

ftdholbrook:
Reporting admin-service api changes to support task service


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/model/ApplicationTenantConfigurationPersistence.java
```diff
@@ -48,17 +51,43 @@
  *         - FirstName
  *         - LastOrSurname
  *         - Birthdate
+ * taskUpdateOrganizationsArtClient:
+ *   tenants:
+ *     CAX:
```
esotrope: not saying we should make a change but it does feel weird that we index props by tenant under the prop instead of at a root level

ftdholbrook: mostly to be consisted with the other tenant aware configs.

PR #2213, opened 2019-08-09
Feature/task multi tenant

esotrope:
reconciliation report

image

art client + import client

image

unrelated data source prune yw @malaffoon 

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
@@ -51,15 +51,16 @@ import {
 import { TranslateService } from '@ngx-translate/core';
 import { fieldRequired, isModified } from '../../model/fields';
 import { notBlank } from '../../../../shared/validator/custom-validators';
-import { byString } from '@kourge/ordering/comparator';
+import { byString, Comparator, join } from '@kourge/ordering/comparator';
 import { TenantService } from '../../service/tenant.service';
 import { FieldConfigurationContext } from '../../model/field';
 import { State } from '../../model/state';
+import { ordering } from '@kourge/ordering';
 
 export type FormMode = 'create' | 'update';
 export type FormState = 'creating' | 'saving' | 'deleting';
 const keyboardDebounceInMilliseconds = 300;
-const updateModeWritableConfigurationsPattern = /^(aggregate|reporting)\..+$/;
+const updateModeWritableConfigurationsPattern = /^(aggregate|reporting|artClient|importServiceClient|sendReconciliationReport)\..+$/;
 
 export function tenantFormGroup(
   mode: FormMode,
@@ -155,16 +156,17 @@ interface PropertySearch {
  * @param defaults The default values of the form
  * @param submitted$ Observable that produces a signal when the form is submitted (used for configurations)
  * @param keyTransform Transform placed on keys before matching them with the search word (used for configurations)
+ * @param comparator The result comparison method to use for sorting
  */
 function toPropertiesProvider(
   searchFormGroup: FormGroup,
   formGroup: FormGroup,
   defaults: { [key: string]: any },
   submitted$: Observable<boolean> = of(false),
-  keyTransform: (key: string) => string = value => value
+  keyTransform: (key: string) => string = value => value,
+  comparator: Comparator<string> = byString
 ): Observable<PropertySearch> {
   const { controls, value } = searchFormGroup;
-  //const entries = Object.entries(defaults);
   return combineLatest(
     controls.search.valueChanges.pipe(
       startWith(value.search),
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -51,15 +51,16 @@ import {
...
 import { TenantService } from '../../service/tenant.service';
 import { FieldConfigurationContext } from '../../model/field';
 import { State } from '../../model/state';
+import { ordering } from '@kourge/ordering';
 
 export type FormMode = 'create' | 'update';
 export type FormState = 'creating' | 'saving' | 'deleting';
 const keyboardDebounceInMilliseconds = 300;
-const updateModeWritableConfigurationsPattern = /^(aggregate|reporting)\..+$/;
+const updateModeWritableConfigurationsPattern = /^(aggregate|reporting|artClient|importServiceClient|sendReconciliationReport)\..+$/;
```
esotrope: allows configuration of these things after tenant creation (note these are hidden for sandbox forms)

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -155,16 +156,17 @@ interface PropertySearch {
...
+ * @param comparator The result comparison method to use for sorting
  */
 function toPropertiesProvider(
   searchFormGroup: FormGroup,
   formGroup: FormGroup,
   defaults: { [key: string]: any },
   submitted$: Observable<boolean> = of(false),
-  keyTransform: (key: string) => string = value => value
+  keyTransform: (key: string) => string = value => value,
+  comparator: Comparator<string> = byString
```
esotrope: allows correct sorting of config prop groups

image

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -249,7 +251,8 @@ function stateToTenant(
   };
 }
 
-// lookup values
+// dont fill the send report archives
+const archivePathPrefixPattern = /^archive\.pathPrefix$/;
```
esotrope: only update archive.pathPrefix - the others dont really need/want a default per @malaffoon

webapp/src/main/webapp/src/app/admin/tenant/model/field-configurations.ts
```diff
@@ -90,85 +72,163 @@ const examProcessorRequiredDataElements = () =>
...
+    [`${basePath}.password`]: {
       dataType: 'password',
       required: true,
       hidden: sandboxHidden
     }
-  })),
+  };
+}
+
+function oauth2(basePath: string): { [key: string]: FieldConfiguration } {
```
esotrope: extracted reusable config factories for oauth2, archive and data sources : j

webapp/src/main/webapp/src/app/admin/tenant/model/field-configurations.ts
```diff
@@ -90,85 +72,163 @@ const examProcessorRequiredDataElements = () =>
...
+    }
+  };
+}
+
+function archives(
+  basePath: string,
+  count: number
+): { [key: string]: FieldConfiguration } {
+  let archives = {};
+  for (let i = 0; i < count; i++) {
```
esotrope: just realized we may need special handling for archives.0 and archives.1 right now the UI will always send both and in this order. do we want to get in the business of removing empty ones before sending?

ftdholbrook: I don't think sending empty ones makes much sense.

esotrope: k just patched to filter empty ones out

webapp/src/main/webapp/src/app/admin/tenant/model/field-configurations.ts
```diff
@@ -90,85 +72,163 @@ const examProcessorRequiredDataElements = () =>
...
+    [`${basePath}.s3AccessKey`]: {
+      dataType: 'string',
+      hidden: sandboxHidden
+    },
+    [`${basePath}.s3RegionStatic`]: {
+      dataType: 'string',
+      hidden: sandboxHidden
+    },
+    [`${basePath}.s3SecretKey`]: {
+      dataType: 'password',
```
esotrope: @malaffoon does it make sense to impose our password constraints on this - feels like an issue now that i look at this again

malaffoon: I don't think so. We're putting constraints on, not to encourage a good policy, but to stop them from creating something that won't work. Since S3 secret keys are generated by AWS, i don't think we should be trying to second guess it.

PR #2214, opened 2019-08-12
RP-674: archive tenant configuration on tenant delete

gregcharles:
Backend component of Tenant Deletion, which will only archive the tenant configuration in gitlab. The other resources (databases, archives, etc.) will be left in place.

I did a sort of end-to-end test of this, even though I can't create a non-sandbox tenant locally, by manually creating a configuration in my forked version of the config repo and then sending a manual request to the tenant controller. It renamed my tentant-XX folder to archive-XX and checked in the changes as expected. However, the tenant list in the UI still showed tenant-XX until I completely restarted. Logging out and in again didn't work. Neither did stopping the config-server container and restarting. I'm assuming there's a cache not being cleared somewhere, but I'm not sure if that's a missing call, or just to be expected when running in local development.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitTenantConfigurationPersistenceService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitTenantConfigurationPersistenceService.java
@@ -128,6 +128,7 @@ public class GitTenantConfigurationPersistenceService implements TenantConfigura
         }
     }
 
+    @Override
     public void deleteSandbox(final String tenantKey) {
         if (tenantKey == null || !tenantKey.contains("_S")) {
             throw new TenantConfigurationPersistenceException("Only sandboxes can be deleted, invalid tenant key " + tenantKey);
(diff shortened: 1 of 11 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitTenantConfigurationPersistenceService.java
```diff
@@ -167,6 +168,52 @@ public void deleteSandbox(final String tenantKey) {
...
+            throw new TenantConfigurationPersistenceException("Only regular tenants can be deleted by this method, invalid tenant key " + tenantKey);
+        }
+
+        forceLocalRepositoryToCleanState();
+
+        final UsernamePasswordCredentialsProvider credentialsProvider = credentialsProvider();
+
+        try {
+            final File tenantDirectory = tenantDirectory(tenantKey);
+            final File archiveDirectory = archiveDirectory(tenantKey);
```
malaffoon: What if they delete the same tenant multiple times? I don't think we necessarily have to worry about this for this initial check-in but perhaps create a JIRA if it will be problematic.

gregcharles: Right, Dave and I discussed this issue last week. The users are supposed to manually delete the resources after deleting the tenant through the UI. That includes the databases, s3 files, and, I think, the archived configuration too. If they don't, and then create a the same tenant again, then there would probably be various conflicts. Are you mainly worried about the archived configuration though?

gregcharles: Implemented collision avoidance for archived configurations. If archive-<tenantkey> exists, it will try to use archive-1-<tenantkey>, then archive-2-<tenantkey> all the way to archive-9-<tenantkey>. If all these are taken, the delete will fail.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1428,7 +1428,7 @@
         "head": "Delete <strong>{{label}} ({{code}})</strong>?",
         "body": {
           "SANDBOX": "Are you sure you want to delete sandbox <strong>{{label}} ({{code}})</strong>?",
-          "TENANT": "Are you sure you want to delete tenant <strong>{{label}} ({{code}})</strong>?"
+          "TENANT": "<strong>WARNING!!! This is a tenant, not a sandbox!</strong><br>Are you really sure you want to delete tenant <strong>{{label}} ({{code}})</strong>?"
```
esotrope: Perfection : J

PR #2216, opened 2019-08-12
adds edit icon on hover of tenant in tenant list

esotrope:
edit-icon-hover

Can't tell on the GIF but the item also highlights on hover as it always did


malaffoon:
I assume "tenant" here refers to both tenants and sandboxes?

PR #2218, opened 2019-08-12
add Jackson DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES to false

ftdholbrook:
At least partial fix for RP-699

json now being serialized (tested locally with CA)

http://localhost:8080/api/admin-service/tenants/CA

```
  "artClient": {
    "oauth2": {
      "grantType": "password",
      "clientAuthenticationScheme": "header",
      "tokenName": "access_token",
      "password": "password",
      "authenticationScheme": "header",
      "clientOnly": false,
      "authenticationRequired": false,
      "scoped": false
    }
  },
  "importServiceClient": {
    "oauth2": {
      "grantType": "password",
      "clientAuthenticationScheme": "header",
      "tokenName": "access_token",
      "password": "password",
      "authenticationScheme": "header",
      "clientOnly": false,
      "authenticationRequired": false,
      "scoped": false
    }
  },
```

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configserver/DefaultConfigServerClient.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configserver/DefaultConfigServerClient.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.admin.multitenant.configserver;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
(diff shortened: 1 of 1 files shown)


ftdholbrook:
> Can we use a mixin or something to get rid of the fields we don't want them to mess with?

Unfortunately the class that is causing the problem is owned by spring `org.springframework.security.oauth2.client.token.grant.password.ResourceOwnerPasswordResourceDetails` The other alaternative would be to make a mirror class and copy it at construction time.  This seemed more expedient.

PR #2219, opened 2019-08-12
prevents form submission on enter key

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.html
```diff
@@ -2,6 +2,7 @@
   *ngIf="formGroup"
   [formGroup]="formGroup"
   (ngSubmit)="onSubmit()"
+  (keydown.enter)="$event.preventDefault()"
```
esotrope: couldn't have asked for this to be easier

PR #2221, opened 2019-08-13
RP-694: warn on duplicate tenant IDs

gregcharles:


webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -83,6 +84,11 @@ export class TenantComponent implements OnDestroy {
         .exists((value || '').toUpperCase())
         .pipe(map(exists => !exists));
 
+    this.tenantIdAvailable = (value: string) =>
+      this.service
+        .existsId((value || '').toUpperCase())
```
esotrope: @ftdholbrook do we force uppercase on IDs too? i cant remember. if we allow mixed case i think we would do an exact case match

ftdholbrook: I don't think we are forcing anything on ID's but I can double check.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/TenantAdministrationStatusService.java
```diff
@@ -67,6 +67,14 @@
      */
     Optional<TenantAdministrationStatusContext> findStatus(String tenantKey);
 
+    /**
+     * Checks if an tenant with the given ID already exists.
+     *
+     * @param id the tenant ID
+     * @return true if the tenant already exists, false if not
+     */
+    boolean checkId(String id);
```
malaffoon: This seems out of place? Perhaps i'll see further on but this has nothing to do with status does it?

malaffoon: Hmm, as part of the status service should this be more like `Optional<...> findStatusById(String tenantId)`?

malaffoon:
Does it bother anybody else that it requires enhancing three services to do this?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/administration/TenantAdministrationStatusService.java
gregcharles: I'd be more inclined to change the other method. The status is never actually used, except to determine its presence.

gregcharles:
> Does it bother anybody else that it requires enhancing three services to do this?

Ha, it bothered me! :) Less now that the work is done though.

gregcharles:
Updated status service to return status by ID instead of a boolean.

PR #2222, opened 2019-08-13
RP-695 Enhance serialization persistence POJO

ftdholbrook:
- validation and task elements to exclude tenant map when not present

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/model/ApplicationTenantConfigurationPersistence.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/model/ApplicationTenantConfigurationPersistence.java
@@ -84,10 +84,10 @@ public class ApplicationTenantConfigurationPersistence {
     private Map<String, Map<String, AggregateReportingPropertiesTenant>> aggregateReporting;
     private Map<String, Map<String, Map<String, DataSourceElementsTenant>>> datasources;
     private Map<String, Map<String, ArchivePropertiesTenant>> archive;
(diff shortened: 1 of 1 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/model/ApplicationTenantConfigurationPersistence.java
```diff
@@ -120,25 +120,25 @@ public ApplicationTenantConfigurationPersistence(TenantConfiguration tenantConfi
...
-        taskUpdateOrganizationsImportServiceClient = ImmutableMap.of("tenants",
-                ImmutableMap.of(tenant.getKey(),
-                        Optional.ofNullable(tenantConfiguration.getImportServiceClient())
-                                .orElse(new ImportServiceClientPropertiesTenant())));
-
-        taskSendReconciliationReport = ImmutableMap.of("tenants",
-                ImmutableMap.of(tenant.getKey(),
-                        Optional.ofNullable(tenantConfiguration.getSendReconciliationReport())
-                                .orElse(new SendReconciliationReportPropertiesTenant())));
+        Optional.ofNullable(tenantConfiguration.getValidation()).ifPresent(validationPropertiesTenant -> {
```
esotrope: if it's not a nullable already and we aren't using .map or something fancy i feel like just doing a traditional if check would do

ftdholbrook: purely a matter of style... I think this reads slightly better.  The "correct" solution is only available in Java 9 or later ifPresentOrElse()

PR #2224, opened 2019-08-13
Defect/687 keep alive

esotrope:


webapp/src/main/webapp/src/app/shared/security/service/not-authenticated.http-interceptor.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(private router: Router) {}
+
+  intercept(
+    request: HttpRequest<any>,
+    next: HttpHandler
+  ): Observable<HttpEvent<any>> {
+    return next.handle(request).pipe(
+      catchError(error => {
+        if (error.status === 401) {
+          this.router.navigateByUrl('/session-expired');
```
esotrope: deleting my cookie and clicking the page routed me successfully to the session expiry page : j

webapp/src/main/webapp/src/app/shared/security/service/authenticated-http.service.ts
```diff
@@ -11,6 +11,9 @@ import { Observable, of, throwError as _throw } from 'rxjs';
 import { catchError } from 'rxjs/operators';
 import { SecurityService } from './security.service';
 
+/**
+ * @deprecated use {@link HttpClient} and {@link NotAuthenticatedHttpInterceptor}
```
esotrope: one outstanding tech debt issue of the webapp is we use the HttpModule which has been deprecated since angular v5 x J

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -34,6 +36,9 @@ import { ordering } from '@kourge/ordering';
 import { byString } from '@kourge/ordering/comparator';
 import { State } from '../../model/state';
 import { StateOptionsService } from '../../service/state-options.service';
+import { KeepAliveService } from '../../../../shared/security/service/keep-alive.service';
+
+const keepAliveDebounce = 1000 * 10; // 10 seconds
```
esotrope: is this too much? too little? i'm not really sure what a good value for this is

webapp/src/main/java/org/opentestsystem/rdw/reporting/web/KeepAliveController.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Ping this controller to extend the active user's session
+ */
+@RestController
+@RequestMapping("/api/keep-alive")
+class KeepAliveController {
```
esotrope: important note here is that this is an authentication-required endpoint. that way it will return a 401 when the user pings it with an expired session

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -160,6 +166,21 @@ export class TenantComponent implements OnDestroy {
     );
   }
 
+  ngOnInit(): void {
+    combineLatest(
```
esotrope: maybe `concat()` is what i want here. need to read up on `rxjs` operators again

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
malaffoon: Don't we know the session ttl is 15 minutes; wouldn't 60s be more than enough? To Dave's point it doesn't matter a lot since there will typically be only one such user logged in at a time.

esotrope: keep in mind that a setting of 60s will only ping the server 60s after the interaction is made. so if they are into minute 14 the keep alive ping will fail.

esotrope: maybe what we are trying to do here  is a "throttle" actually

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -34,6 +36,9 @@ import { ordering } from '@kourge/ordering';
 import { byString } from '@kourge/ordering/comparator';
 import { State } from '../../model/state';
 import { StateOptionsService } from '../../service/state-options.service';
+import { KeepAliveService } from '../../../../shared/security/service/keep-alive.service';
+
+const keepAliveThrottleTime = 1000 * 60; // 1 minute
```
esotrope: @malaffoon changed to a throttle of 1 min - this means that the first click/keydown will ping the server immediately and it will prevent server pings until a minute has transpired - sound right?

PR #2225, opened 2019-08-13
allow score and stderr to be zero (vs. null)

malaffoon:


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
```diff
@@ -9,22 +9,22 @@
...
+    private Integer studentCount;
 
     public Integer getAvgScaleScore() {
-        return (avgScaleScore != null && avgScaleScore == 0) ? null : avgScaleScore;
+        return avgScaleScore;
     }
 
     public Integer getAvgStdErr() {
-        return (avgStdErr != null && avgStdErr == 0) ? null : avgStdErr;
+        return avgStdErr;
```
malaffoon: Don't conflate null and 0. But now we have to make sure any place we set it we know whether it should be null or zero. There are only three places that seem to set the values ...

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/impl/AbstractReportRepository.java
```diff
@@ -183,8 +183,8 @@ protected void validate(final Q query) {
 
     protected Measures measures(final ResultSet row, final String asmtTypeCode, final String prefix) throws SQLException {
         final Measures.Builder builder = Measures.builder()
-                .avgScaleScore(row.getInt(prefix + "score"))
-                .avgStdErr(row.getInt(prefix + "std_err"))
+                .avgScaleScore(getNullable(row, row.getInt(prefix + "score")))
+                .avgStdErr(getNullable(row, row.getInt(prefix + "std_err")))
```
malaffoon: recall that `row.getInt` returns a 0 if the value is null ... we can't have that

esotrope: nice find!

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/Measures.java
esotrope: the UI the only thing that consumes these values? or does the backend process them more anywhere?

malaffoon: I don't see it being processed anywhere else ...

PR #2226, opened 2019-08-14
fixes oauth2 config prop mappings

esotrope:
Still working with @gregcharles to test this but i think this is the fix.


webapp/src/main/webapp/src/app/admin/tenant/model/tenants.ts
```diff
@@ -161,8 +169,11 @@ export function toConfigurations(
       : {
           aggregate,
           archive,
+          artClient,
           datasources,
-          reporting
+          importServiceClient,
+          reporting,
+          sendReconciliationReport
```
esotrope: the tenant edit form might want these x J

Do these have instance level defaults we want to suppress by chance? if so, that will require another logic block changing as well.

PR #2227, opened 2019-08-14
RP-701 correctly set translation location on tenant create

ftdholbrook:


admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantServiceTest.java
```diff
@@ -184,343 +176,4 @@ public void itShouldGenerateAValidKey() {
         assertThat(nvKey).isEqualTo("CC_S001");
     }
 
-}
-
-//@Ignore
```
ftdholbrook: unrelated, left over cruft from the sandbox / tenant consolation effort

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantServiceTest.java
```diff
@@ -130,20 +136,6 @@ public void getSandboxDataSetsShouldGetSandboxDatasets() {
         assertThat(expectedOne.size()).isEqualTo(1);
     }
 
-
```
ftdholbrook: we no longer have defaults resolved like this removing

PR #2229, opened 2019-08-14
Defect/700 oauth2 fields

esotrope:
This is not a perfect solution but it is comparable to the other multi-field consistency validation solutions that are in place.

image

The problem i have with this is we don't get the red highlighting and required star next to the fields - only a message at the top of the table.

In order to improve this the form logic will have to be changed considerably to make the form validators form group-aware which would introduce more form-frameworkish code around angular forms

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -1498,6 +1498,7 @@
       "configurations": {
         "label": "Configurations",
         "error": {
+          "oauth2s": "Please enter both clientSecret and password for all OAuth2 forms OR leave both blank.",
           "onePasswordPerUser": "Please provide only one password for each unique datasource username.",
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1498,6 +1498,7 @@
       "configurations": {
         "label": "Configurations",
         "error": {
+          "oauth2s": "Please enter both clientSecret and password for all OAuth2 forms OR leave both blank.",
```
esotrope: @malaffoon @ftdholbrook  thinking this logic is incorrect. do any of the fields have instance level defaults or are they all "override required"? let me know so i can update accordingly

ftdholbrook: I actually think those two fields are the only ones we should show, based off of @malaffoon comments on RP-709.  And they should require override.

esotrope:
fixed in other branch

PR #2231, opened 2019-08-15
fix for RP-709

ftdholbrook:
- Jackson mixin for ResourceOwnerPasswordResourceDetails

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitTenantConfigurationPersistenceService.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitTenantConfigurationPersistenceService.java
@@ -13,8 +13,10 @@ import org.opentestsystem.rdw.admin.multitenant.configserver.ConfigServerClient;
 import org.opentestsystem.rdw.admin.multitenant.model.ApplicationTenantConfigurationPersistence;
 import org.opentestsystem.rdw.admin.multitenant.model.TenantConfiguration;
 import org.opentestsystem.rdw.multitenant.Tenant;
+import org.opentestsystem.rdw.multitenant.task.ResourceOwnerPasswordResourceDetailsMixin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.security.oauth2.client.token.grant.password.ResourceOwnerPasswordResourceDetails;
 
 import java.io.File;
 import java.io.IOException;
@@ -46,6 +48,7 @@ public class GitTenantConfigurationPersistenceService implements TenantConfigura
         final ObjectMapper mapper = new ObjectMapper(new YAMLFactory().disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER));
         mapper.setSerializationInclusion(JsonInclude.Include.NON_EMPTY);
         mapper.findAndRegisterModules();
+        mapper.addMixIn(ResourceOwnerPasswordResourceDetails.class, ResourceOwnerPasswordResourceDetailsMixin.class);
         return mapper;
     }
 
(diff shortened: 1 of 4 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/model/ApplicationTenantConfigurationPersistence.java
```diff
@@ -84,10 +85,10 @@
...
     private Map<String, Map<String, Map<String, DataSourceElementsTenant>>> datasources;
     private Map<String, Map<String, ArchivePropertiesTenant>> archive;
-    private Map<String, Map<String, ExamProcessorValidationPropertiesTenant>> validation = ImmutableMap.of();
-    private Map<String, Map<String, ArtClientPropertiesTenant>> taskUpdateOrganizationsArtClient = ImmutableMap.of();
-    private Map<String, Map<String, ImportServiceClientPropertiesTenant>> taskUpdateOrganizationsImportServiceClient = ImmutableMap.of();
-    private Map<String, Map<String, SendReconciliationReportPropertiesTenant>> taskSendReconciliationReport = ImmutableMap.of();
+    private Map<String, Map<String, ExamProcessorValidationPropertiesTenant>> validation = new HashMap<>();
+    private Map<String, Map<String, ArtClientPropertiesTenant>> taskUpdateOrganizationsArtClient = new HashMap<>();
+    private Map<String, Map<String, ImportServiceClientPropertiesTenant>> taskUpdateOrganizationsImportServiceClient = new HashMap<>();
+    private Map<String, Map<String, SendReconciliationReportPropertiesTenant>> taskSendReconciliationReport = new HashMap<>();
```
ftdholbrook: For reasons not clear to me, after adding the mixins in, the spring relaxed binding code started failing on these.  After trial and error replacing these with mutable hash maps fixed the problem.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/git/GitTenantConfigurationPersistenceService.java
```diff
@@ -13,8 +13,10 @@
 import org.opentestsystem.rdw.admin.multitenant.model.ApplicationTenantConfigurationPersistence;
 import org.opentestsystem.rdw.admin.multitenant.model.TenantConfiguration;
 import org.opentestsystem.rdw.multitenant.Tenant;
+import org.opentestsystem.rdw.multitenant.task.ResourceOwnerPasswordResourceDetailsMixin;
```
gregcharles: Is this a new class? I can't find it. Does the common version need to be bumped? (ResourceOwnerPasswordResourceDetailsMixin)

ftdholbrook: https://github.com/SmarterApp/RDW_Common/pull/159

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/model/ApplicationTenantConfigurationPersistence.java
malaffoon: I was wondering about that during the initial checkin ... i've never been able to use immutable maps with spring property beans.

admin-service/src/test/java/org/opentestsystem/rdw/admin/multitenant/git/OjbectMapperTest.java
```diff
@@ -46,10 +48,16 @@ public void bigTest() throws JsonProcessingException, BindException {
         final String output = ymlMapper.writeValueAsString(config);
         System.out.println(output);
```
malaffoon: should probably suppress this

PR #2232, opened 2019-08-15
698 oauth2 validation

esotrope:
Added oauth2 validation (reconciliation report validations to come)

oauth-validation

This is another ticket but i added the caution sign which shows all the errors of that form field in a popup when hovered

oauth-validation-2


webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.html
```diff
@@ -30,26 +30,33 @@
             modified: modified(property)
           }"
         >
-          <td [title]="property.key">
-            {{ property.key }}
+          <td [title]="property.configuration.name">
```
esotrope: matching the localization table rendering up with the configuration table a bit

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.ts
```diff
@@ -7,40 +7,14 @@ import {
 import {
   ControlContainer,
   ControlValueAccessor,
-  FormControl,
   FormGroup,
   NG_VALUE_ACCESSOR
 } from '@angular/forms';
-import { ConfigurationProperty, Property } from '../../model/property';
```
esotrope: configuration and localization tables now have the same factory method so i extracted it

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-tree-table/property-override-tree-table.component.html
```diff
@@ -176,6 +158,30 @@
                         </ng-container>
                       </ng-container>
 
+                      <ng-template #errorsPopover>
```
esotrope: the error popup

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-tree-table/property-override-tree-table.component.ts
```diff
@@ -141,43 +117,75 @@ export class PropertyOverrideTreeTableComponent
...
+    return formFieldModified(
+      property.configuration.dataType.inputType,
+      this.formGroup.value[property.configuration.name],
       property.originalValue
     );
   }
 
-  showPasswordToggle(property: ConfigurationProperty): boolean {
-    const value = this.formGroup.value[property.key];
+  showPasswordToggle(property: Property): boolean {
```
esotrope: config property and property are now really the same thing so they share the same model

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-tree-table/property-override-tree-table.component.ts
```diff
@@ -141,43 +117,75 @@ export class PropertyOverrideTreeTableComponent
...
-    const defaultValue = this.defaults[property.key];
+    const defaultValue = this.defaults[name]; // property.originalValue?
     if (defaultValue != null) {
       return defaultValue;
     }
+    // TODO mask values?
     return '';
   }
 
+  getOptions(property: Property): Observable<Option<any>[]> {
```
esotrope: get form field options lazily when the form is rendered - this simplified form initialization logic considerably so you didn't need to resolve all of this up front and pass around a bunch of params everywhere

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -1,219 +1,45 @@
...
 import { ordering } from '@kourge/ordering';
+import { Property } from '../../model/property';
+import { configurationFormFields } from '../../model/configuration-forms';
+import { stringDataType } from '../../model/data-types';
+import { propertyForm } from '../../model/property-forms';
+import { propertiesProvider, tenantForm } from './tenant-forms';
 
 export type FormMode = 'create' | 'update';
 export type FormState = 'creating' | 'saving' | 'deleting';
-const keyboardDebounceInMilliseconds = 300;
```
esotrope: below was moved to tenant-forms.ts

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -469,11 +297,30 @@ export class TenantFormComponent implements OnChanges, OnDestroy {
       states != null &&
       (value.type !== 'SANDBOX' || (tenants != null && dataSets != null))
     ) {
-      this.formGroup = tenantFormGroup(
+      this.configurationProperties = configurationFormFields(
```
esotrope: pulled this logic up top to avoid duplicating the effort downstream

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -486,63 +333,43 @@ export class TenantFormComponent implements OnChanges, OnDestroy {
...
         this.formGroup.controls.configurations as FormGroup,
-        configurationDefaults,
+        this.configurationProperties,
         this.submitted$,
         key => (/^\w+\.(.*)$/.exec(key) || ['', ''])[1], // only match last segment of key
         configurationPropertyComparator
       ).pipe(
         takeUntil(this.destroyed$),
         map(({ results, hasSearch, invalid }) => {
-          const properties = results.map(key => {
```
esotrope: now there is a form factory that tells you what is readonly/visible based on tenant type and form mode (edit/create) instead of doing it here with a regexp

webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-forms.ts
```diff
@@ -0,0 +1,183 @@
...
+              configuration: {
+                name,
+                dataType: { inputType }
+              }
+            } = property;
+            const caseInsensitiveSearch = search.toLowerCase();
+            const key = keyTransform(name);
+            const formValue = formGroup.getRawValue()[name];
+
+            // support search of input's property placeholders
```
esotrope: there was a regression here when we added the placeholder logic - you could no longer search the field values - this is fixed with this though

webapp/src/main/webapp/src/app/admin/tenant/model/data-types.ts
```diff
@@ -0,0 +1,106 @@
...
+} from './option-providers';
+import {
+  urlFragment,
+  uri,
+  archiveUri,
+  databasePassword,
+  notPostgresReservedWord
+} from './form/validators';
+
+export const defaultDataType: DataType = {
```
esotrope: all the data type metadata in a much clearer form

webapp/src/main/webapp/src/app/admin/tenant/model/configuration-forms.ts
```diff
@@ -0,0 +1,344 @@
...
+  ...taskServiceReconciliationReportFieldConfigurations()
+];
+
+export function configurationsFormFieldConfiguration(
+  name: string
+): FormFieldConfiguration {
+  return configurations.find(configuration => configuration.name === name);
+}
+
+export function configurationsFormFieldConfigurations(
```
esotrope: the config form factory - this organization put in one place what the form should look like based on tenant type and form mode

webapp/src/main/webapp/src/app/admin/tenant/model/form/data-type.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Observable } from 'rxjs';
+import { Option } from './option';
+import { InputType } from './input-type';
+import { ValidatorFn } from '@angular/forms';
+import { ConstraintType } from './constraint-type';
+
+/**
+ * Represents a data type and how it should be treated
+ */
+export interface DataType<T = any> {
```
esotrope: new files are from moving everything out of fields.ts into dedicated files so things are easier to find

webapp/src/main/webapp/src/app/admin/tenant/model/form/validators.ts
```diff
@@ -0,0 +1,201 @@
...
+): ValidatorFn {
+  return function(control: AbstractControl): ValidationErrors {
+    const value = control.value;
+    const formGroup = control.parent as FormGroup;
+    if (formGroup == null) {
+      return null;
+    }
+
+    // only do this for the first two controls in the chain
+    if (recursionCount++ === 0) {
```
esotrope: the silliness of the oauth2 validator - we need to tell the form to update other fields and when those update they are revalidated which update this control - this prevents the stack from overflowing : j

webapp/src/main/webapp/src/app/admin/tenant/model/tenants.ts
```diff
@@ -193,7 +195,7 @@ export function toConfigurations(
 export function toDefaultConfigurations(defaults: any, type: TenantType): any {
   return omitByKey(toConfigurations(defaults, type), key =>
     // blank out any defaults for these fields
-    /^(aggregate\.tenant|datasources\.\w+\.(username|password|urlParts\.database|schemaSearchPath))$/.test(
+    /^(aggregate\.tenant|datasources\.\w+\.(username|password|urlParts\.database|schemaSearchPath)|\w+\.oauth2\.password)$/.test(
```
esotrope: suppress instance level configured oauth2.password

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1549,8 +1549,16 @@
...
+      "reset": "Click this button to reset the value back to its original default value"
+    },
+    "error": {
+      "archiveUri": "Archive URI must start with s3:// or file://",
+      "databasePassword": "The password must be between 8 to 64 characters in length, and must contain at least one uppercase letter, one lower case letter, and one number.",
+      "required": "This field is required",
+      "requiredOAuth2Field": "Please enter both oauth2 username and password or leave both blank",
+      "reservedWord": "{{reservedWord}} is a reserved word and cannot be used",
+      "uri": "Please enter a valid URL",
+      "urlFragment": "Please enter a valid URL fragment"
```
esotrope: these error messages ok?

webapp/src/main/webapp/src/app/admin/tenant/model/form/validators.ts
malaffoon: and this is where i don't understand scoping of variable; i assume you tested this lots of times and it work as expected

webapp/src/main/webapp/src/assets/i18n/en.json
malaffoon: Where do we have urlFragments?

webapp/src/main/webapp/src/app/admin/tenant/model/form/validators.ts
esotrope: the variable scope is global - because the recursion is guaranteed it is guaranteed to be reset to zero after every call to this method. kind of rediculous i know. The only "better" way i could think of doing this was also stateful and relied on removing the validator from the other form field - updating its validity and then adding it back.

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: interpretive guide and user guide url. This validation is very permissive and says that the value must either start with a slash or start with a scheme

PR #2234, opened 2019-08-15
adds filtering of non-active tenants from sandbox tenant list. adds e…

esotrope:
…rror handling for missing parent tenant

image

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/tenant/model/data-types.ts
+++ b/webapp/src/main/webapp/src/app/admin/tenant/model/data-types.ts
@@ -15,10 +15,6 @@ import {
   notPostgresReservedWord
 } from './form/validators';
 
-export const defaultDataType: DataType = {
-  inputType: 'input'
-};
-
 export const stringDataType: DataType = {
   inputType: 'input'
 };
--- a/webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
+++ b/webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
@@ -40,6 +40,9 @@ import { KeepAliveService } from '../../../../shared/security/service/keep-alive
 
 const keepAliveThrottleTime = 1000 * 60; // 1 minute
 
+const validTenantStatus = status =>
+  ['ACTIVE', 'UPDATE_FAILED'].includes(status);
+
 const byLabel = ordering(byString).on<TenantConfiguration | DataSet>(
   ({ label }) => label
 ).compare;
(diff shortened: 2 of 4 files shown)


webapp/src/main/webapp/src/app/admin/tenant/model/data-types.ts
```diff
@@ -15,10 +15,6 @@ import {
   notPostgresReservedWord
 } from './form/validators';
 
-export const defaultDataType: DataType = {
```
esotrope: unused

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -40,6 +40,9 @@ import { KeepAliveService } from '../../../../shared/security/service/keep-alive
 
 const keepAliveThrottleTime = 1000 * 60; // 1 minute
 
+const validTenantStatus = status =>
+  ['ACTIVE', 'UPDATE_FAILED'].includes(status);
```
esotrope: this felt right to me as these are the status values a tenant can have when you select one to edit it but let me know what you think makes sense for this too @ftdholbrook

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1486,14 +1486,16 @@
...
+        "missing": "No longer available"
       },
       "dataSet": {
         "label": "Data Set",
         "info": "The sample data that the sandbox will be initialized with",
         "error": {
           "required": "Please select a data set to initialize the sandbox with"
-        }
+        },
+        "missing": "No longer available"
```
esotrope: figured i'd get two birds with this one

webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -83,7 +86,12 @@ export class TenantComponent implements OnInit, OnDestroy {
       share()
     );
     const tenants$ = this.service.getAll('TENANT').pipe(
-      map(values => values.slice().sort(byLabel)),
+      map(values =>
+        values
+          .slice()
+          .filter(({ status }) => validTenantStatus(status))
```
esotrope: were you going to put the filter here @malaffoon ?

malaffoon: yeah but i just did `status => status == 'ACTIVE'` so it's a good thing you're doing this  ;)

malaffoon: also, it wasn't perfectly clear to me if this would just affect the list of tenants in the sandbox config screen

esotrope: yeah. not exactly clear huh. you have to deduce from we are on `tenant.comonent.ts` meaning we are displaying one tenant (so we are viewing/editing a tenant) and we don't show the tenants in the tenant form - only the sandbox form.

PR #2235, opened 2019-08-15
fixes session expiry redirect and state option ordering

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/model/option-providers.ts
```diff
@@ -42,10 +42,12 @@ export const reportLanguageOptions = ({ translateService }) =>
...
-      states.map(value => ({
-        label: value.name,
-        value
-      }))
+      states
+        .map(value => ({
+          label: value.name,
+          value
+        }))
+        .sort(byLabel)
```
esotrope: knocked this out while i was in there

webapp/src/main/webapp/src/app/shared/security/service/not-authenticated.http-interceptor.ts
```diff
@@ -25,7 +25,7 @@ export class NotAuthenticatedHttpInterceptor implements HttpInterceptor {
     return next.handle(request).pipe(
       catchError(error => {
         if (error.status === 401) {
-          this.router.navigateByUrl('/session-expired');
+          this.service.expireSession();
```
esotrope: already had a service to do this right - just forgot - easy peasy

PR #2236, opened 2019-08-15
fixes iab printed report not-scored message condition

esotrope:


report-processor/src/main/resources/templates/iab-body.html
```diff
@@ -65,7 +65,7 @@
                         </div>
                         <div class="row relative-container">
                             <div class="col-xs-9">
-                                <div th:if="${exam.scaleScore != null and exam.scaleScore.standardError != null}" class="scale-score-graph individual"
```
esotrope: not sure how this ever made sense

esotrope: also not sure how i didn't run across this in my app-wide search for "standardError"

malaffoon: i think i added it after all the other changes; it probably made sense at the time but i can't recall why

PR #2237, opened 2019-08-15
fixes regression with switching parent tenants on sandbox form

esotrope:
caught a would-be regression _wewps_


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -413,11 +405,15 @@ export class TenantFormComponent implements OnChanges, OnDestroy {
     // TODO improve by only writing over pristine form fields
     this.formGroup.patchValue({
       configurations: propertyForm(
-        this.configurationProperties,
+        toConfigurationProperties(
```
esotrope: actually looking at this again i think the existing code works fine as the defaults are no longer tenant defaults but instance level defaults - this change makes the code less stateful and more functional though i guess _shrug_

PR #2238, opened 2019-08-15
RP-713 remove California from instance wide defaults

ftdholbrook:


webapp/src/main/resources/application.yml
```diff
@@ -36,9 +36,6 @@ reporting:
   landing-page-url: forward:/landing.html
   min-item-data-year: 2016
   school-year: 2019
-  state:
-    code: CA
-    name: California
```
esotrope: nothing needs to be added to `-test.yml` after these are removed?

ftdholbrook:
Smoke tested by creating a new tenant locally, and ran IT tests

webapp/src/main/resources/application.yml
```diff
@@ -36,9 +36,6 @@ reporting:
   landing-page-url: forward:/landing.html
   min-item-data-year: 2016
   school-year: 2019
```
esotrope: looking at these again - does it really make sense to have defaults for these:
`min-item-data-year`
`school-year`
?

ftdholbrook: I wondered about that too, but if we remove them, then they have to become required for tenants correct?

webapp/src/main/resources/application.yml
ftdholbrook: looks to me like they are already there in all the cases I could find.

webapp/src/main/resources/application.yml
esotrope: sure would - but i think rightfully so - the 2016 is an SBAC item specific thing and school-year _should_ default to the current year someone is making the tenant for

webapp/src/main/resources/application.yml
esotrope: kew

webapp/src/main/resources/application.yml
ftdholbrook: Want to open a ticket for that @esotrope?

esotrope: sure - first i want @malaffoon 's buy in though - thoughts Mark?

malaffoon: min-item-data-year should be left alone; yeah, it is specific to the migrated legacy data for SBAC CA deployment but having an "old" value as the default is very safe.

school-year should be managed explicitly. I suppose we could remove the default and make them set it by tenant.

malaffoon:
Btw @ftdholbrook , i'm waiting for the follow up change to RDW_Ingest?

ftdholbrook:
> Btw @ftdholbrook , i'm waiting for the follow up change to RDW_Ingest?

kind of didn't realize there were any.  I'll go take a look.

PR #2239, opened 2019-08-16
Feature/676 tenant create errors

esotrope:
image

I didn't stick the stacktrace in there yet because i couldn't really think of a nice way to do it. Something would need to be a button that you can click to bring you to a stack trace view or modal but just making the label that button felt non-intuitive.
What do you guys think? should the label open a page with the stack trace?


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantConfigurationViewService.java
```diff
@@ -146,6 +146,9 @@ public TenantConfiguration defaultTenantConfiguration() {
                                 .tenant(entry.getValue().getTenant())
                                 .administrationStatus(TenantAdministrationStatusContext.builder()
                                         .tenantAdministrationStatus(entry.getValue().getTenantAdministrationStatus())
+                                        .message(entry.getValue().getMessage())
+                                        .stackTrace(entry.getValue().getStackTrace())
+                                        .updated(entry.getValue().getUpdated())
```
esotrope: error info and updated timestamp are now passed to the UI (though i didn't add the timestamp to the view yet)

PR #2240, opened 2019-08-16
RP-715 fix cipher prefix based on feedback from QA

ftdholbrook:


ftdholbrook:
> cool - any screenshot of the values coming into the ui?

no, very similar to what comes back for data sources

PR #2242, opened 2019-08-16
fixes task service art client credentials validation

esotrope:
image

ignore the other missing defaults - the defaults are spoofed by my local atm

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/tenant/model/configuration-forms.ts
+++ b/webapp/src/main/webapp/src/app/admin/tenant/model/configuration-forms.ts
@@ -74,7 +74,10 @@ function postgresDataSource(basePath: string): FormFieldConfiguration[] {
   ];
 }
 
-function oauth2(basePath: string): FormFieldConfiguration[] {
+function oauth2(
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/admin/tenant/model/configuration-forms.ts
```diff
@@ -95,17 +98,26 @@ function oauth2(basePath: string): FormFieldConfiguration[] {
     {
       name: `${basePath}.username`,
       dataType: stringDataType,
-      validators: [
-        // TODO updating this needs to dirty the other fields
-        requiredIfOthersPresent([`${basePath}.password`], 'requiredOAuth2Field')
-      ]
+      validators: hasDefaultCredentials
+        ? []
```
esotrope: dont add validation when we expect defaults

esotrope: actually it would be nice if our form validators were somehow "default value" aware and this logic could be implied... but whatever this does the trick for now

PR #2247, opened 2019-08-19
fixes sending of school year to backend + adds tests

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/model/tenants.ts
```diff
@@ -251,6 +255,12 @@ export function toServerTenant(tenant: TenantConfiguration): any {
 }
 
 export function toServerConfigurations(configurations: any): any {
+  // Account for default school year
+  // set default school year to current year
+  if (configurations['reporting.schoolYear'] == null) {
+    configurations['reporting.schoolYear'] = defaultSchoolYear();
+  }
```
esotrope: our diffing logic made it so it didn't actually send the default school year - oops x l

PR #2248, opened 2019-08-20
improves archive validation

esotrope:
I'll share a vid of the behavior on the slack chan


webapp/src/main/webapp/src/app/admin/tenant/model/configuration-forms.ts
```diff
@@ -122,15 +122,11 @@ function oauth2(
...
-function archive(
-  basePath: string,
-  pathPrefixRequired: boolean
-): FormFieldConfiguration[] {
+function archive(basePath: string): FormFieldConfiguration[] {
   return [
     {
       name: `${basePath}.pathPrefix`,
-      dataType: stringDataType,
-      required: pathPrefixRequired
```
esotrope: wasn't required actually

webapp/src/main/webapp/src/app/admin/tenant/model/form/validators.ts
```diff
@@ -199,3 +200,70 @@ export function requiredIfOthersPresent(
...
+
+export function archives(formGroup: FormGroup): ValidationErrors {
+  // for each archive config
+  Object.keys(formGroup.controls)
+    .filter(key => archivePattern.test(key))
+    .map(key => archivePattern.exec(key)[1])
+    .forEach(prefix => {
+      // if s3 field is filled OR uriRoot is s3 scheme
+      // then the rest of the required s3 fields are required
+      // also the uriRoot field is required to start with s3://
```
esotrope: this is the logic going on here so if this is right we should be good

webapp/src/main/webapp/src/app/admin/tenant/model/tenants.ts
```diff
@@ -215,6 +215,7 @@ export function toDefaultConfigurations(defaults: any, type: TenantType): any {
   configurations['reporting.schoolYear'] = defaultSchoolYear();
 
   delete configurations['aggregate.tenant'];
+  delete configurations['archive.s3RegionStatic']; // TODO delete instance level application.yml entry for this
```
esotrope: suppress instance level default for this so i can require it

PR #2249, opened 2019-08-20
RP-548 fix  fix consolidated configuration inconsistencies

ftdholbrook:


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
```diff
@@ -202,8 +210,27 @@ public DefaultConfigpropsService(final ConfigpropsConsolidationService configpro
      */
     private <T> Optional<T> condenseConfigurations(final Map<String, T> configurations, final String configTypeKey) {
         HashSet<T> configurationSet = Sets.newHashSet(configurations.values());
-        //one or none are expected
+        //rules for those with more than one configuration expected
         if (configurationSet.size() > 1) {
```
ftdholbrook: this is the core of the change, the rules are somewhat of an educated guess on my part, please comment on RP-548 if they need some change.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
```diff
@@ -53,24 +66,25 @@ public DefaultConfigpropsService(final ConfigpropsConsolidationService configpro
 
     @Override
     public Optional<ReportingSystemSettings> defaultReportingSystemProperties() {
+        final String configTypeKey = "reporting" ;
```
esotrope: was this meant to be extracted?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
```diff
@@ -202,8 +210,27 @@ public DefaultConfigpropsService(final ConfigpropsConsolidationService configpro
...
+        //rules for those with more than one configuration expected
         if (configurationSet.size() > 1) {
+            switch (configTypeKey) {
+                case ArchiveConfigTypeKey:
+                    if (configurations.containsKey(RDW_REPORTING_REPORT_PROCESSOR.getName())) {
+                        return Optional.of(configurations.get(RDW_REPORTING_REPORT_PROCESSOR.getName()));
+                    }
+                    break;
+                case DatasourcesReportingRoConfigTypeKey:
+                case DatasourcesReportingRwConfigTypeKey:
```
malaffoon: reporting_rw is more of an ingest thing. I don't think reporting-service even uses rw access.

esotrope: what about for user created reports and groups?

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
ftdholbrook: reporting is kind of a special case, so I didn't but I guess I can to be consistent.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
ftdholbrook: @esotrope  this is picking an "exemplar" if there is more than one config for some reason, so it is basically "fall back to this one if we can't consolidate it down to one". So on some level, one is as good as another as long as it is correct.

esotrope: i'll trust you on this one : j

PR #2250, opened 2019-08-20
729 tenant form performance

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-forms.ts
```diff
@@ -154,7 +154,12 @@ export function propertiesProvider(
             } = property;
             const caseInsensitiveSearch = search.toLowerCase();
             const key = keyTransform(name);
-            const formValue = formGroup.getRawValue()[name];
```
esotrope: the culprit - i guess this method somehow can cause infinite loops : J
thanks Angular

PR #2252, opened 2019-08-20
more robust solution for scrolling to first invalid form field

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.ts
```diff
@@ -442,20 +445,66 @@ export class TenantFormComponent implements OnChanges, OnDestroy {
...
+
+    const isConfigurationProperty = !topLevelConfigurationFormFieldNames.includes(
+      firstInvalidFieldName
+    );
+
+    // open configurations form and select first invalid form field
+    if (isConfigurationProperty) {
+      // clear searches
+      this.configurationControlsFormGroup.patchValue({
+        search: '',
```
esotrope: if the user has a config prop search filter the logic couldn't find the invalid field on the page.
this change clears the search filter, waits till the elements render and then focuses the first invalid form field.

PR #2259, opened 2019-08-23
fixes redirect after save of tenant

esotrope:


webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -258,11 +254,7 @@ export class TenantComponent implements OnInit, OnDestroy {
           })
         )
         .subscribe(
-          () => {
-            this.router.navigate(['..'], {
```
esotrope: so in the route config we have
```
{
  path: 'tenants',
  children: [{
    path: ':key',
    children: [{
      path: '',
      component: OurPage
    }]
  }]
}
```
 `/tenants/:key` resolves to the 3rd route in. but `..` from that route just goes to `/tenants/:key` i think this is probably an angular bug. who would say `..` from the innermost route here and expect this behavior... : /

PR #2268, opened 2019-08-29
enables stats page

esotrope:
image

### DIFF
--- a/admin-service/src/main/resources/application.sql.yml
+++ b/admin-service/src/main/resources/application.sql.yml
@@ -536,32 +536,32 @@ sql:
   metrics:
     findAll: >-
       SELECT 'Subjects' AS category,
-          sy.year,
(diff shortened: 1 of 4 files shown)


admin-service/src/main/resources/application.sql.yml
```diff
@@ -536,32 +536,32 @@ sql:
...
           JOIN asmt a ON a.id = e.asmt_id
           JOIN school_year sy ON sy.year = a.school_year
-      GROUP BY category, sy.year, code
+      GROUP BY category, school_year, subject_code
       UNION ALL
       SELECT 'Schools' AS category,
-          sy.year,
-          'Schools' AS code,
+          sy.year as school_year,
+          NULL AS subject_code,
```
esotrope: main change here was to select null for these

PR #2269, opened 2019-08-30
749 tenant audit fields

esotrope:


admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantService.java
```diff
@@ -147,53 +146,6 @@ public void deleteTenant(final String tenantKey) {
         }
     }
 
-    private TenantConfiguration generateUpdateTenantConfiguration(TenantConfiguration requestTenantConfiguration) {
```
esotrope: just moved it

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantService.java
```diff
@@ -95,39 +96,37 @@ public boolean existsId(final String id) {
...
 
     @Override
-    public void createTenant(final TenantConfiguration tenantConfiguration) {
+    public void createTenant(final String username, final TenantConfiguration tenantConfiguration) {
         final String tenantKey = tenantConfiguration.getTenant().getKey();
         if (tenantConfiguration.getTenant().isSandbox()) {
-            logger.debug("Creating a new sandbox with key {}", tenantKey);
-            TenantConfiguration sandbox = generateSandboxTenantConfigurationForCreate(tenantConfiguration);
+            logger.debug("Creating sandbox for tenant ({})", tenantKey);
+            tenantValidationService.sandboxCreationValidation(tenantConfiguration);
```
esotrope: pulled validation out to here from the next line's method

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/service/impl/DefaultTenantService.java
```diff
@@ -278,6 +234,75 @@ TenantConfiguration generateSandboxTenantConfigurationForCreate(final TenantConf
...
+    TenantConfiguration tenantConfigurationForCreate(final String username, final TenantConfiguration configuration) {
+
+        // Not using immutable here
+        if (isLocalizationSupplied(configuration.getLocalization())) {
+            configuration.getReporting().setTranslationLocation(
+                    generateTranslationLocation(configuration.getTenant().getKey())
+            );
+        }
+
+        final String created = DateTimeFormatter.ISO_INSTANT.format(Instant.now());
```
esotrope: not sure if this is exactly fits with our time scheme - i think in the DB we are using UTC but i'll maybe defer to @malaffoon. What i do know is that on the UI `new Date(isoString)` will localize the time so _maybe_ this is ok but i didn't look into it very much

PR #2271, opened 2019-08-30
fixes school input alignment

esotrope:


webapp/src/main/webapp/src/app/school-grade/results/school-results.component.html
```diff
@@ -55,16 +55,20 @@ <h1 heading class="h2">
             [options]="schoolOptions"
             (schoolChange)="onSchoolChange($event)"
           ></school-typeahead>
+
+          <!-- typeahead search result messages and progress spinner -->
           <div>
```
esotrope: when there are more than 10 schools assigned to the user the school input switches to a typeahead instead of a select box - in the case of a typeahead we show this lovely div that pushes things around : J

webapp/src/main/webapp/src/app/assessments/assessments.component.less
```diff
@@ -2,5 +2,4 @@
   display: grid;
   grid-template-columns: 1fr auto;
   gap: 15px;
-  align-items: center;
```
esotrope: the fix

PR #2276, opened 2019-09-11
RP-777: fix router link for tenant/sandbox edit cancel

gregcharles:


webapp/src/main/webapp/src/app/admin/tenant/component/tenant-form/tenant-form.component.html
```diff
@@ -74,7 +74,10 @@ <h1 class="h2" heading>
           </div>
         </li>
         <li>
-          <a class="btn btn-default" [routerLink]="['..']">
+          <a
+            class="btn btn-default"
+            [routerLink]="[mode === 'create' ? '..' : '../..']"
+          >
```
gregcharles: This feels hacky, but for some reason in edit mode, setting router link to [..] just routed back to the edit page, i.e., was exactly the same as setting it to [.]   In create mode, that doesn't happen.

esotrope: it has to do with the way we configured routing - its wack i know

PR #2277, opened 2019-09-15
RP-763: allow setting session cookie max-age.

gregcharles:


webapp/src/main/java/org/opentestsystem/rdw/reporting/ApplicationConfiguration.java
```diff
@@ -121,9 +121,21 @@ public ConfigurationChangeCacheClearer configurationChangeListener(final CacheMa
...
+            @Value("${server.servlet.session.cookie.max-age:#{null}}") final Integer maxAge2) {
+
+        // Trick to make server.session.cookie.max-age apply to a Spring-managed session.
+        // e.g., when spring.session.store-type=redis. (This property will change name in
+        // SpringBoot 2.0, so accepting both properties is to future-proof this method.)
         DefaultCookieSerializer serializer = new DefaultCookieSerializer();
-        serializer.setCookieMaxAge(7200);
+
+        if (maxAge != null) {
+            serializer.setCookieMaxAge(maxAge);
```
esotrope: interesting so does this let them control the session times?

gregcharles: There's already: spring.session.timeout for that. I don't know why they'd use a different property for spring session timeouts vs. regular session timeouts, but they do. I also don't know why they didn't just create spring.session properties for the cookie configuration. It would have saved me some agony.

PR #2278, opened 2019-09-17
RP-726 Show summary metrics about each tenant in UI

ftdholbrook:
- switch to olap query
- add assessment type
- add assessment type column in UI

Last migration time element still isn't included.  That feature will be a separate PR.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/DataSourceConfiguration.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/DataSourceConfiguration.java
@@ -219,4 +219,55 @@ public class DataSourceConfiguration {
     public PlatformTransactionManager warehouseTxManager(@Qualifier("warehouseDatasource") final DataSource dataSource) {
         return new DataSourceTransactionManager(dataSource);
     }
+
+    /**
+     * DataSourceElements for OLAP
+     *
+     * @return OLAP DataSourceElements
+     */
+    @Bean(name = "olapDataSourceElementsProperties")
+    @ConfigurationProperties("datasources.olap_ro")
+    @RefreshScope
+    public DataSourceElementsProperties olapDataSourceElementsProperties() {
+        return new DataSourceElementsProperties();
+    }
+
+    /**
+     * olap DataSourceElementsResolver
+     *
+     * @param tenantKeyResolver            id resolver
+     * @param dataSourceElementsProperties config properties
+     * @return DataSourceElementsResolver
+     */
+    @Bean(name = "olapDataSourceElementsResolver")
+    @RefreshScope
+    public DataSourceElementsResolver dataSourceElementsResolver(TenantKeyResolver tenantKeyResolver,
+                                                                 @Qualifier("olapDataSourceElementsProperties") DataSourceElementsProperties dataSourceElementsProperties) {
+        return new DataSourceElementsResolver(tenantKeyResolver, dataSourceElementsProperties);
+    }
+
+    /**
+     * Data source for the olap reporting data mart with no transaction manager.
+     *
+     * @param dataSourceElementsResolver olap DataSourceElementsResolver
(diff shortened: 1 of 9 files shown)


ftdholbrook:
Screen Shot 2019-09-17 at 9 45 03 AM

admin-service/src/main/java/org/opentestsystem/rdw/admin/configuration/DataSourceConfiguration.java
```diff
@@ -219,4 +219,55 @@ public NamedParameterJdbcTemplate warehouseJdbcTemplate(@Qualifier("warehouseDat
...
+     */
+    @Bean(name = "olapDataSourceElementsResolver")
+    @RefreshScope
+    public DataSourceElementsResolver dataSourceElementsResolver(TenantKeyResolver tenantKeyResolver,
+                                                                 @Qualifier("olapDataSourceElementsProperties") DataSourceElementsProperties dataSourceElementsProperties) {
+        return new DataSourceElementsResolver(tenantKeyResolver, dataSourceElementsProperties);
+    }
+
+    /**
+     * Data source for the olap reporting data mart with no transaction manager.
```
malaffoon: comment seems a bit misleading, unless i'm missing something: nothing to do with a transaction manager is it?

ftdholbrook: copy and paste from aggregate-service

admin-service/src/main/resources/application.sql.yml
```diff
@@ -535,33 +535,75 @@ sql:
...
+              FROM exam
+              GROUP BY school_year, asmt_id ) e
+        JOIN asmt a
+          ON a.id = e.asmt_id
+        JOIN asmt_type t
+          ON a.type_id = t.id
+        JOIN subject s
+          ON s.id = a.subject_id
+        GROUP BY  e.school_year, s.code, t.code
+      UNION
```
malaffoon: seems like these should still be `UNION ALL` or does redshift not support that?

ftdholbrook: It does, I saw no difference in the results.

admin-service/src/main/resources/application.sql.yml
```diff
@@ -535,33 +535,75 @@ sql:
...
+        FROM (
+          SELECT DISTINCT school_id,
+                 school_year
+                 from iab_exam
+          UNION
+          SELECT DISTINCT school_id,
+                 school_year from exam) s
+        GROUP BY s.school_year, subject_code, asmt_type
+      ) as m
+      GROUP BY m.category, m.school_year, m.subject_code, m.asmt_type
```
malaffoon: Is this `GROUP BY` actually doing anything? Each union'd select is distinct wrt to this grouping so it seems like it isn't necessary. But i'm too lazy to actually play with the SQL and figure it out myself  ;)

ftdholbrook: yes, there are some cross results because of the exam table split

admin-service/src/main/resources/application.sql.yml
ftdholbrook: actually, i switched to inner queries, so maybe can pull it

malaffoon: It's a performance thing: `UNION` does a distinct. Because there is no overlap it probably isn't noticeable but ... best practice and all.

admin-service/src/main/resources/application.sql.yml
ftdholbrook: actually, i switched to inner queries (schools and students), so maybe can pull it

ftdholbrook: modifed

admin-service/src/main/resources/application.sql.yml
ftdholbrook: modified

PR #2279, opened 2019-09-17
RP-726 Show summary metrics about each tenant in UI

ftdholbrook:
- change order of precedence in StatidTenantIdUtil.tenantId() to allow tenant context to override the context for an already logged in user
- audited code and made sure all places where TenantContextHolder.setTenantId() is used a matching TenantContextHolder.clear() is used in a finally block

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/reporting/common/multitenant/StaticTenantIdUtil.java
+++ b/common/src/main/java/org/opentestsystem/rdw/reporting/common/multitenant/StaticTenantIdUtil.java
@@ -19,7 +19,7 @@ public class StaticTenantIdUtil {
      * @return tenantId Option
      */
     public static Optional<String> tenantId() {
-        return Stream.of(tenantIdFromSecurityContext(), tenantIdFromTenantContext())
+        return Stream.of(tenantIdFromTenantContext(), tenantIdFromSecurityContext())
                 .filter(Optional::isPresent)
                 .map(Optional::get)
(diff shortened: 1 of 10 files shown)


common/src/main/java/org/opentestsystem/rdw/reporting/common/multitenant/StaticTenantIdUtil.java
```diff
@@ -19,7 +19,7 @@
      * @return tenantId Option
      */
     public static Optional<String> tenantId() {
-        return Stream.of(tenantIdFromSecurityContext(), tenantIdFromTenantContext())
```
ftdholbrook: This is the resolution order change required

webapp/src/main/java/org/opentestsystem/rdw/reporting/sandbox/DefaultSandboxService.java
```diff
@@ -64,8 +64,8 @@ public DefaultSandboxService(final SandboxRepository sandboxRepository,
         SecurityContextHolder.getContext().setAuthentication(null);
 
         for (final Tenant sandbox : sandboxes) {
-            TenantContextHolder.setTenantId(sandbox.getId());
             try {
+                TenantContextHolder.setTenantId(sandbox.getId());
```
malaffoon: not sure this change was needed but it doesn't hurt anything ...

PR #2283, opened 2019-10-08
RP-826: enhancement to allow internal commas in tenancy chain

gregcharles:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/authority/DefaultAuthorityService.java
```diff
@@ -88,7 +89,7 @@
 
         final List<Authority> authorities = Arrays.stream(encodedGrants)
                 // each encodedGrant could actually be a list of comma-separated grants
-                .map(g -> g.split(",")).flatMap(Arrays::stream)
+                .map(TenancyChain::splitGrants).flatMap(Arrays::stream)
```
ftdholbrook: Nice moving the responsibility out to the TenancyChain object

PR #2284, opened 2019-10-18
id and name unique for assessment results score button toggle

ftdholbrook:


ftdholbrook:
Attempted only making ID unique, but that didn't solve the problem.  Looking at other components in the code base it looked like making the name unique also mattered, so I followed that pattern.  This seems to have resolved the problem.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -363,6 +363,8 @@ export class AssessmentResultsComponent implements OnInit {
   scoreTypeOptions: Option[] = [];
   showInstructionalResources: boolean;
   embargo: Embargo;
+  scoreTypeId: String = 'score-type-' + Utils.newGuid();
+  scoreTypeName: String = 'score-type-' + Utils.newGuid();
```
malaffoon: Don't know why but this feels inelegant. The only suggestion i'd have is to use something from `this._assessmentExam.assessment`. That said, i am completely out of my league as to what "elegant" even is for Angular.

esotrope: x J
My only comment would be that `id` can be the same as `name` uuid and doesn't need any prefix like `score-type-` because it is a uuid : J

```
uuid = uuid();
[id]="uuid"
[name]="uuid"
```

PR #2287, opened 2019-11-12
More Okta updates

gregcharles:


webapp/src/main/java/org/opentestsystem/rdw/reporting/security/saml/SamlSettings.java
```diff
@@ -100,6 +102,19 @@ public String getEntityBaseUrl() {
         return loadBalance.getBaseUrl();
     }
 
+    /**
+     * @return maximum time in seconds that system allows users to single sign-on since their initial authentication
+     * with the IDP. Setting this to a very high value can avoid getting hard-to-resolve "access denied"
+     * errors in some situations.
+     */
+    public Integer getMaxAuthenticationAge() {
```
malaffoon: nit: what will a null value do? Will it fall back to some default, have no max at all, or ...?

webapp/src/main/resources/application.yml
```diff
@@ -190,3 +192,18 @@ zuul:
...
+spring:
+  profiles: okta
+
+server:
+  session:
+    cookie:
+      max-age: 3600
+
+reporting:
+  logout-redirect-url: "https://smarterbalanced.oktapreview.com"
```
malaffoon: !!! did you mean to check this in? The embedded yml should have absolutely no production (or pre-production) configuration in it. 

Also, it violates the whole 12-factor app thing to have profiles in the embedded yml. I sort of see an argument for having defaults for the `okta` configuration but not if they are environment-specific values.

PR #2292, opened 2020-01-27
RDW-11: fix UI accessibility issues called out by T-Base.

gregcharles:


gregcharles:
T-Base report is attached to JIRA: [https://smarterbalanced.atlassian.net/browse/RDW-11](url)

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.html
```diff
@@ -59,6 +59,8 @@
                     *ngIf="modified(property)"
                     href="javascript:void(0)"
                     (click)="onResetButtonClick(property)"
+                    (keyup.space)="onResetButtonClick(property)"
+                    (keydown.space)="$event.preventDefault()"
```
esotrope: better than this would be to turn the <a> tag into a <button> so that we get the browser-default space key behavior when this element has tab-focus

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -77,10 +77,10 @@ export class AppComponent implements OnInit {
...
 
-  scrollToMainContent() {
-    setTimeout(() => {
-      document.getElementById('maincontent').scrollIntoView();
-    }, 0);
+  getMainContentLink(url) {
+    // Strips any existing anchor off url and adds main content anchor in its place.
+    let n = url.indexOf('#');
+    return (n < 0 ? url : url.substring(0, n)) + '#maincontent';
   }
```
esotrope: right idea! - could be more concise though
```typescript
get mainContentHref(): string {
  return this.router.url.replace(/(#.*)/, '') + '#maincontent';
}
```

webapp/src/main/webapp/src/app/shared/form/sb-button-group.ts
```diff
@@ -157,11 +157,7 @@ const ControllerByInputType: { [inputType: string]: InputController } = {
       >
         <input
           type="checkbox"
-          [attr.checked]="
-            isAllOption
-              ? stateInternal.selectedAllOption
-              : stateInternal.selectedOptions.has(option)
-          "
+          [attr.checked]="stateInternal.selectedOptions.has(option)"
```
esotrope: was this a bug? or the actual expected behavior for another use-case?
Make sure other places that use this component aren't adversely affected by this change.

Same comment for the other change in this file

gregcharles: No, I noticed that too when I looked at the PR file diffs. I can't remember making that change, but I've restored it and making sure it still works.

webapp/src/main/webapp/src/app/admin/tenant/component/property-override-table/property-override-table.component.html
gregcharles: Into a ... what? A button maybe? Would that have styling issues?

esotrope: i think my message got xss-stripped.
Yeah `<a>` -> `<button>`

esotrope: And then yes - you'll probably have to mess with the styles.
first give the button a special class `button-icon-only` or something
```
.button-icon-only {
  border: none;
  background: none;
}
```

PR #2293, opened 2020-02-14
fix translation provider cache

ftdholbrook:
- json/git provider and jdbc provider are cached separately with individual cache management
- expanded @RefreshScope to handle edge-case were location of git/json file changes

### DIFF
--- a/webapp/src/main/resources/application.yml
+++ b/webapp/src/main/resources/application.yml
@@ -1,3 +1,9 @@
+app:
+  cache:
+    translation:
+      # every 2 mins (2 * 60 * 1000)
+      refresh-delay: "120000"
+
 # actuator end-points should be served from a different port than the API/web content
 management:
   port: 8008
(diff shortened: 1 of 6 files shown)


webapp/src/main/resources/application.yml
```diff
@@ -1,3 +1,9 @@
+app:
+  cache:
+    translation:
+      # every 2 mins (2 * 60 * 1000)
+      refresh-delay: "120000"
```
malaffoon: Is it only subject loading that affects this? Every 2m seems like a lot for something that happens so rarely.

ftdholbrook: My thinking was this was a relatively inexpensive operation that has the potential to be called a lot.  In that case a relatively short lived cache  should work fine.  It will save most of the trashing against the database, but will reflect database changes relatively quickly.

PR #2294, opened 2020-05-08
refresh gradle

malaffoon:
Update gradle to 6.0. Fix resulting fall-out, and try to make things consistent with other projects. Update spring-boot from 1.5.18 to 1.5.21.


README.md
```diff
@@ -17,70 +24,18 @@ npm install @angular/cli
 ```
 
 #### MySQL
-MySQL is required for building (integration tests) and running this application. To better match production, MySQL
```
malaffoon, line 20: references RDW_Schema which has the full instructions in one place.

admin-service/src/main/java/org/opentestsystem/rdw/admin/Application.java
```diff
@@ -18,7 +18,7 @@
         DataSourceTransactionManagerAutoConfiguration.class,
         HibernateJpaAutoConfiguration.class
 })
-@PropertySource(value = "classpath:build-info.properties", ignoreResourceNotFound = true)
+@PropertySource(value = "classpath:META-INF/build-info.properties", ignoreResourceNotFound = true)
```
malaffoon, line 21: I switched to using SpringBoot buildInfo instead of building it ourselves; default location changed.

admin-service/src/main/java/org/opentestsystem/rdw/admin/multitenant/configprops/DefaultConfigpropsService.java
```diff
@@ -8,7 +8,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Sets;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
```
malaffoon, line 12: clean up the inconsistent use of the older commons-lang

aggregate-service/build.gradle
```diff
@@ -1,43 +1,41 @@
...
-    compile 'org.springframework.cloud:spring-cloud-starter-aws'
-    compile 'org.springframework.cloud:spring-cloud-starter-bus-amqp'    
-    compile 'org.springframework.boot:spring-boot-starter-jdbc'
-    compile 'org.springframework.boot:spring-boot-starter-security'
-    compile 'org.springframework.cloud:spring-cloud-starter-config'
-    compile 'org.springframework.cloud:spring-cloud-starter-stream-rabbit'
-    compile 'org.springframework.integration:spring-integration-security'
+    implementation 'com.amazon.redshift:redshift-jdbc42-no-awssdk:1.2.8.1005'
+    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
+    implementation 'com.healthmarketscience.sqlbuilder:sqlbuilder:2.1.7'  // this shouldn't be necessary, why??
```
malaffoon, line 17: @agorina any idea why the transitive dependency isn't sufficient for this? The change is due to the upgrade to gradle but i couldn't figure out why i'd need this here. No biggie, just asking in case it triggers an old memory.

build.gradle
```diff
@@ -50,101 +44,158 @@ subprojects {
         mavenLocal()
         mavenCentral()
         maven { url 'http://redshift-maven-repository.s3-website-us-east-1.amazonaws.com/release' }
+        maven { url 'https://artifactory.smarterbalanced.org/artifactory/libs-releases' }
```
malaffoon, line 47: Instead of pulling in jfrog, just list the public repository for common libs

webapp/src/test/java/org/opentestsystem/rdw/reporting/i18n/TranslationProviderBackedTranslationJsonServiceTest.java
```diff
@@ -52,7 +52,7 @@ public void itShouldReturnTranslationsAsJson() {
     public void itShouldIncludeUITranslations() {
         final ObjectNode node = service.getLocaleMessages(Locale.US, true);
         assertThat(node).hasSize(4);
-
-        assertThat(node.fieldNames()).containsOnly("report", "child_A", "child_B", "child_C");
+        assertThat(node.fieldNames()).hasNext()
+            .toIterable().containsOnly("report", "child_A", "child_B", "child_C");
```
malaffoon, line 56: In all the RDW code and tests, this is the only assertion involving an Iterator<> of results. Changed in this version of assertj-core.

webapp/build.gradle
```diff
@@ -1,92 +1,109 @@
...
+ is hooked into the build at the 'processResources' stage.
+
+ Poke around angular.json and the file structure to fully grok this: the angular
+ 'webapp' folder is under 'src/main' and it builds into 'webapp/dist'. Those
+ resources need to be under 'src/main/resources' for the app build so, after
+ doing the front-end build, they are moved.
+
+ This may be a bit confusing (case-in-point: /webapp/src/main/webapp/src) but i
+ believe the idea is the front-end "sub project" should exist and build blissfully
+ unaware that it is going to be delivered in a spring boot app as static resources.
```
malaffoon, line 54: Mostly for me after (re)discovering how it works. Seems needlessly complicated but i don't want to mess up the development workflow of the angular devs so i left it mostly the same.

webapp/build.gradle
```diff
@@ -1,92 +1,109 @@
...
 }
 
-//Hopefully this allows us to hot add resources
-bootRun {
-    addResources = true
-    if (project.hasProperty('jvmArgs')) {
-        jvmArgs project.jvmArgs.split('\\s+')
-    }
+npm_install {
+    workingDir = file('src/main/webapp')
```
malaffoon, line 64: I really don't understand how this was *ever* working before. The default behavior of `npm_install` is to run at the project root directory. Meh, whatever.

webapp/build.gradle
agorina: Do not know, maybe this help: https://docs.gradle.org/6.0/userguide/java_library_plugin.html ?

aggregate-service/build.gradle
agorina: I honestly do not know. Maybe this is helpfull: https://docs.gradle.org/6.0/userguide/java_library_plugin.html ?

malaffoon: Thanks @agorina . I banged my head against gradle java dependencies for a few hours and then decided to give up and just repeat dependencies. Maybe i'll buy @wtritch a beer and see if he has any ideas.

wtritch: I wonder if we're going a little crazy with `implementation` and aren't selectively using `api` when a library's API depends upon/extends another library's API?

agorina: +1, that was my thought too, but I assumed that Mark has tried it.

malaffoon: I tried. The "api" configuration doesn't appear to do anything useful. Obviously, i must be doing something wrong (thus my comment in `developer` chat).

PR #2296, opened 2020-05-13
Fixes for node build / sbac-ui-kit

gregcharles:
Upped sbac-ui-kit to version 0.0.30 and removed .npmrc, which pointed at old report. I also matched "node" configuration in build.gradle with what TIMS has. However, we still have the problem of the "dist" folder of sbac-ui-kit not being brought in and causing the "ng build" to fail. On TIMS, the dist folder is also missing on new builds, though it was in my cached version. The TIMS "ng build" does _not_  fail despite the missing sbac-ui-kit/dist folder.

### DIFF
--- a/webapp/build.gradle
+++ b/webapp/build.gradle
@@ -60,7 +60,6 @@ dependencies {
 node {
     version = '12.11.1'
     npmVersion = '6.11.3'
-    distBaseUrl = 'https://nodejs.org/dist'
     download = true
 }
 
--- a/webapp/build.gradle
+++ b/webapp/build.gradle
@@ -58,8 +58,9 @@ dependencies {
  ***************************************************************************************/
 
 node {
-    version = '8.9.1'
-    npmVersion = '5.5.1'
+    version = '12.11.1'
+    npmVersion = '6.11.3'
+    distBaseUrl = 'https://nodejs.org/dist'
     download = true
 }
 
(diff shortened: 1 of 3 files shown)


webapp/build.gradle
```diff
@@ -58,8 +58,9 @@ dependencies {
  ***************************************************************************************/
 
 node {
-    version = '8.9.1'
-    npmVersion = '5.5.1'
+    version = '12.11.1'
+    npmVersion = '6.11.3'
```
malaffoon, line 62: Are we comfortable with these upgrades? Seems fine and good but i just don't know what all the effects might be.

webapp/build.gradle
```diff
@@ -58,8 +58,9 @@ dependencies {
...
+    distBaseUrl = 'https://nodejs.org/dist'
```
malaffoon, line 63: That is the default setting and shouldn't be necessary.

webapp/build.gradle
gregcharles: I don't know either. I think it makes sense to do now given that we're early in dev and will be able to spot problems and roll back if necessary before uat. I'll ask Landon for an opinion though.

webapp/build.gradle
gregcharles: Ah, OK. I was hoping that would be the solution to the missing dist directory, but it wasn't. If it's redundant, I'll just take it out.

webapp/build.gradle
gregcharles: Thumbs up from Landon!

PR #2303, opened 2020-06-01
add AltScores to ReportOptions in aggregate service

malaffoon:
This makes the changes to only the aggregate-service (with a little collateral damage). Nothing has been done in the webapp itself.

@gregcharles, this should result in the ReportOptions have the list of known alt-scores in the system.


aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/AltScore.java
```diff
@@ -0,0 +1,18 @@
+package org.opentestsystem.rdw.olap.model;
```
malaffoon, line 1: For this purpose, AltScore and Claim are identical in content. I chose to make them separate things (with separate services, caches, etc.) to keep things a bit more specific.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/model/ScorableEntity.java
```diff
@@ -0,0 +1,50 @@
+package org.opentestsystem.rdw.olap.model;
```
malaffoon, line 1: Factor out the common behavior of Claim and AltScore.

aggregate-service/src/main/java/org/opentestsystem/rdw/olap/repository/ClaimRepository.java
```diff
@@ -1,6 +1,6 @@
-package org.opentestsystem.rdw.reporting.common.repository;
```
malaffoon, line 1: I'm not quite sure why Claim and ClaimRepository were in common, they are only used in the aggregate-service. I moved them; we can always put them back if they do end up being used in more than one place.

aggregate-service/src/main/resources/application.sql.yml
```diff
@@ -41,13 +48,12 @@ sql:
       SELECT id, code FROM asmt_type ORDER BY id
 
   claim:
-    aggregate:
```
malaffoon, line 44: this extra level was to differentiate from the non-aggregate implementation ... which wasn't used anywhere.

build.gradle
```diff
@@ -123,7 +123,7 @@ subprojects {
         implementation 'org.slf4j:slf4j-api'
 
         // handy for creating metadata with @ConfigurationProperties
-        compileOnly 'org.springframework.boot:spring-boot-configuration-processor'
+        implementation 'org.springframework.boot:spring-boot-configuration-processor'
```
malaffoon, line 126: This shouldn't be necessary, but it helped get rid of some red in IntelliJ.

common/src/main/java/org/opentestsystem/rdw/reporting/common/util/CollectionUtils.java
```diff
@@ -113,6 +113,14 @@
...
     }
 
+    /**
+     * @param value collection to check
+     * @return true if collection is null or empty
+     */
+    public static boolean safeIsEmpty(final Collection<?> value) {
+        return value == null || value.isEmpty();
+    }
+
```
gregcharles, line 123: Why did you need this? It seems the same as isNullOrEmpty() ... except for declaring the parameter to be final, and Collection<?> in place of Collection<T>. So it's functionally the same, but different in documentation maybe? Could you just change isNullOrEmpty() to use a final parameter? That wouldn't break anything, or would it?

malaffoon: I removed isNullOrEmpty because we don't use the amazon utilities anywhere else (and i really don't like using it since it is redundant with a few other, more common libraries). I was going to pull in apache commons collection4, but just did this instead. I could pull in that library if you'd rather.

gregcharles: Ah, my mistake. I was looking at isNullOrEmpty() in CollectionUtils and missed that it wasn't _our_ CollectionUtils. I vaguely prefer using Apache since we do in some other places, but it's one more thing to import. I wish Java would just add this. (Of course by Java 14, maybe they have.)

PR #2305, opened 2020-06-02
Backend infrastructure for submitting an Alt Score report query

gregcharles:
Infrastructure for submitting a query and return results. Database support is not included and will need to be stubbed out in this stage.


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/web/converter/AbstractAggregateReportCsvHandler.java
```diff
@@ -92,6 +92,7 @@ public boolean accept(final ReportContentResource resource, final MediaType requ
                         query.getType() == ReportQueryType.CustomAggregate ||
                                 query.getType() == ReportQueryType.Longitudinal ||
                                 query.getType() == ReportQueryType.Claim ||
+                                query.getType() == ReportQueryType.AltScore ||
```
malaffoon, line 95: Presumably we need `AltScoreAggregateReportCsvHandler`?

gregcharles: Yes, and probably sooner than later, i.e., maybe in this stage. I haven't needed it yet though because I'm still getting the report to create and display correctly.

PR #2307, opened 2020-06-03
Initial implementation of Alt Score report

gregcharles:
Report configuration and preview working.  Basic generation of the report works using data faked in the back end. Some report management functions still not working correctly ... e.g., deleting and exporting.


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AltScore.java
```diff
@@ -0,0 +1,44 @@
+package org.opentestsystem.rdw.reporting.common.model;
```
malaffoon, line 1: There is one of these in aggregate-service. Is this one here because it needs to be shared between aggregate-service and report-processor, or does the fake service live in the wrong place?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/FakedAltScoreExamReportService.java
```diff
@@ -0,0 +1,248 @@
+package org.opentestsystem.rdw.reporting.processor.service;
```
malaffoon, line 1: (see note above)

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AltScore.java
gregcharles: Yes, actually all the imports are already updated, so this one is a straight-up orphan. I'm not sure what happened there, but I'll delete it.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.ts
```diff
@@ -148,6 +148,26 @@ export interface AggregateReportFormSettings {
...
+
+    /**
+     * The school years to be covered on the report
+     */
+    schoolYears: number[];
+
+    /**
+     * The alt score codes
+     */
+    altScoreCodesBySubject: AltScore[];
```
malaffoon, line 168: i know you're just copying the claim stuff but is this really "BySubject"? It's not even a map/dict.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/FakedAltScoreExamReportService.java
gregcharles: This is supposed to be something like a decorator on DefaultExamReportService, processing the AltScore queries and passing the others through. Once the Alt Score service  is implemented, we should be able to just toss this away without any impact to the front end code.

malaffoon: What i'm confused about is there is this service that backs the report-processor/reports endpoint but there is also an aggregate-service/aggregate endpoint. I initially thought that the aggregate-service endpoint was the only one to worry about.

webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-form-settings.ts
gregcharles: Yeah, it's not a great name. It does contain a subject, so that's something. It is how I was able to find an appropriate subject for the preview report, but if it were a map, it would have been an easier fix. I applied the same change over to claims, although with the current data set, there was no scenario for the claim preview to go bad like the alt score one did.

PR #2308, opened 2020-06-03
cleanup warnings, TODOs, add some comments

malaffoon:


common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -13,11 +13,20 @@
...
+ * concrete classes override the behavior as necessary to enforce their specific rules:<ul>
+ *     <li>subjectCodes: TargetReport:1, Longitudinal:N, Claim:N(map), AltScore:N(map), Custom:N</li>
+ *     <li>schoolYears: TargetReport:1, Longitudinal:1, Claim:N, AltScore:N, Custom:N</li>
+ * </ul>
+ * </p>
  */
 public abstract class AbstractAggregateReportQuery extends AbstractReportQuery implements AggregateQuery {
-
-    private Set<String> subjectCodes;
+    private ImmutableSet<Integer> schoolYears;
```
malaffoon, line 28: I moved this down into the base class. I don't particularly think it belongs here but that makes it consistent with subjectCodes.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -47,14 +56,31 @@ public boolean isFilteredSubgroupQuery() {
...
+    public Set<Integer> getSchoolYears() {
+        return safeImmutableOf(schoolYears);
+    }
+
+    @Override
     public Set<String> getSubjectCodes() {
         return safeImmutableOf(subjectCodes);
     }
 
+    @Override
```
malaffoon, line 69: It's considered good java practice to annotate overrides.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -119,11 +137,12 @@ public StudentFilters getStudentFilters() {
         return safeImmutableOf(subgroups);
     }
 
-    public abstract <B extends Builder> B copy();
+    public abstract <B extends Builder<? extends AbstractAggregateReportQuery, B>> B copy();
```
malaffoon, line 140: fixes a warning

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -182,16 +203,21 @@ protected B copy(final AbstractAggregateReportQuery reportQuery) {
                     .valueDisplayType(reportQuery.valueDisplayType)
                     .achievementLevelDisplayType(reportQuery.achievementLevelDisplayType)
                     .columnOrder(reportQuery.columnOrder)
-                    .dimensionTypes(reportQuery.dimensionTypes == null ? null : newHashSet(reportQuery.dimensionTypes)) // TODO redundant?
+                    .dimensionTypes(reportQuery.dimensionTypes)
                     .studentFilters(reportQuery.studentFilters)
-                    .subgroups(newHashMap(reportQuery.getSubgroups()));
+                    .subgroups(reportQuery.subgroups);
```
malaffoon, line 208: I *think* i know why these were the way they were (the collections need to be mutable after the copy) but it shouldn't be necessary. If this breaks things (all tests pass right now), then the real fix is to change *all* the copied collections to be mutable copies.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AltScoreAggregateReportQuery.java
```diff
@@ -11,27 +10,21 @@
...
-    @Override
-    public Set<Integer> getSchoolYears() {
-        return safeImmutableOf(schoolYears);
-    }
-
     public ImmutableMap<String, Set<String>> getAltScoreCodesBySubject() {
         return altScoreCodesBySubject == null ? ImmutableMap.of() : altScoreCodesBySubject;
     }
 
-    // TODO do we need this?
```
malaffoon, line 34: These are the TODOs that were bugging me. Btw, the answer is: yes.

common/src/main/java/org/opentestsystem/rdw/reporting/common/model/AbstractAggregateReportQuery.java
```diff
@@ -47,14 +56,31 @@ public boolean isFilteredSubgroupQuery() {
         return !getSubgroups().isEmpty();
     }
 
+    @Override
+    public Set<Integer> getSchoolYears() {
+        return safeImmutableOf(schoolYears);
+    }
+
```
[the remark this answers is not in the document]
malaffoon: Yup, i went through the same thought process, discovering that that helper method just returns the collection if it is already immutable.

PR #2311, opened 2020-06-04
Implement report management for Alt Scores reports.

gregcharles:
Fills in some places where AltScore handling was missing for report management  and customization.

### DIFF
--- a/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
+++ b/webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
@@ -97,7 +97,10 @@ export class AggregateReportTableExportService {
             );
           }
         );
-    } else if (options.reportType === 'Claim') {
+    } else if (
+      options.reportType === 'Claim' ||
+      options.reportType === 'AltScore'
+    ) {
       this.addPerformanceLevelColumns(builder, options);
     } else {
       builder.withColumn(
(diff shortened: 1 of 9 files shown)


webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
```diff
@@ -97,7 +97,10 @@ export class AggregateReportTableExportService {
             );
           }
         );
-    } else if (options.reportType === 'Claim') {
+    } else if (
+      options.reportType === 'Claim' ||
+      options.reportType === 'AltScore'
+    ) {
       this.addPerformanceLevelColumns(builder, options);
```
malaffoon, line 104: It might already be covered but i'm using this as a place to comment: the AltScore report will have scale_score and performance level columns like the overall score report.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
```diff
@@ -198,6 +201,19 @@ export class AggregateReportTableExportService {
       );
     }
 
+    if ('altScore' === column) {
+      return builder.withColumn(
+        this.translateService.instant(
+          'aggregate-report-table.columns.alt-score'
```
malaffoon, line 207: This is for the alt score code column, right?

malaffoon:
As long as you're checking something in, can you bump the common dependency to 14 to get the xsd color changes? Thanks.

webapp/src/main/webapp/src/app/aggregate-report/results/aggregate-report-table-export.service.ts
gregcharles: Yes, there are separate rows for each of the alt score codes, similar to how there's a row for each claim code in claim reports.

gregcharles:
Also: 
UI Kit to 0.0.33
RDW Common to 2.4.0-14

PR #2312, opened 2020-06-05
use latest RDW_Common to remove spring-boot-starter-data-jpa dependency

malaffoon:
I'm waiting for a complete local rebuild before committing, but it does seem to fix the problem.


build.gradle
```diff
@@ -123,7 +123,7 @@ subprojects {
         implementation 'org.slf4j:slf4j-api'
 
         // handy for creating metadata with @ConfigurationProperties
-        implementation 'org.springframework.boot:spring-boot-configuration-processor'
+        compileOnly 'org.springframework.boot:spring-boot-configuration-processor'
```
malaffoon, line 126: Unrelated, just reverting a change from last week that isn't really helping the way i thought it would.

PR #2315, opened 2020-06-08
Remove stubbed out estimated rows for AltScore report

gregcharles:


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report.service.ts
```diff
@@ -41,11 +41,6 @@ export class AggregateReportService {
    * @param query the report parameters
    */
   getEstimatedRowCount(query: AggregateReportQueryType): Observable<number> {
-    // TODO: (rdw6) remove this stub when alt-score backend implemented
-    if (query.type === 'AltScore') {
-      return of(10);
-    }
-
```
malaffoon, line 48: Oh, that's why i couldn't get that number to change.

PR #2317, opened 2020-06-09
RDW-6, req-4: remove white override of standard foregrounds

gregcharles:


gregcharles:
Already discussed with Collin

PR #2320, opened 2020-06-12
Relax component-selector linting rule. Update info placeholder text.

gregcharles:
two small fixes


webapp/src/main/webapp/tslint.json
```diff
@@ -1,33 +1,19 @@
 {
```
malaffoon, line 1: This looks like just reformatting, was anything changed? If so, why?

PR #2322, opened 2020-06-15
Apm 6466 test result visibility management ui

vgonzalez9:
Requirement 1 branch. 
Contains new admin feature for test results management, for UI functionality and styles.


webapp/src/main/webapp/src/app/admin/admin.routes.ts
```diff
@@ -7,6 +7,9 @@ export const aggregateReportBreadcrumb = ({ translateService }) =>
 export const embargoBreadcrumb = ({ translateService }) =>
   translateService.instant('embargo.title');
 
+export const testResultsBreadcrumb = ({ translateService }) =>
```
malaffoon, line 10: I'm probably just getting stuck on this but ... I really don't like shortening this as "testResults". Can we use "testVisibility" or "resultsVisibility" or something instead? Test results is a general term that is used all over the place to refer to the results of taking a test.

webapp/src/main/webapp/src/app/admin/test-results/service/test-results.service.ts
```diff
@@ -0,0 +1,186 @@
...
+    this.testResultFilters = this.getTestResultFilterDefaults();
+  }
+
+  // receive filter's options
+  getTestResults(testResultFilters: TestResultFilters): TestResult[] {
+    // todo replace with real data
+    let testResults = this.mockTestResults.getMockTestResults();
+
+    // todo apply filters to results.
+    console.log(
```
malaffoon, line 38: Can/should we remove console output at this point? At least comment it out?

webapp/src/main/webapp/src/app/admin/test-results/service/test-results.service.ts
```diff
@@ -0,0 +1,186 @@
...
+  private applyTestResultsFilter(
+    testResult: TestResult,
+    testResultFilters: TestResultFilters
+  ) {
+    // may need to adjust order
+    if (
+      testResultFilters.status != this.statusDefault &&
+      testResultFilters.status != testResult.status
+    )
+      return false;
```
malaffoon, line 129: Two questions/comments:
1. For me this formatting is hard to read. If you disagree, feel free to leave it as is.
2. Not having parens, and having the return on a newline will lead to errors and should be avoided. For example, let's say i'm debugging this and add a log line ...
```
    if ( 
        testResult...
    )
        console.log("got here");
        return false;
```

oops

webapp/src/main/webapp/src/app/admin/test-results/test-results.component.html
```diff
@@ -0,0 +1,265 @@
+<page-heading>
+  <h1 class="h2" heading>{{ 'test-results.title' | translate }}</h1>
+  <ul class="list-unstyled list-inline" controls *ngIf="showAudit">
+    <li>
+      <a
+        class="btn btn-default"
+        href="/assets/testResultsAudit202006040.csv"
+        download="/assets/testResultsAudit202006040.csv"
```
malaffoon, line 8: Shouldn't this be provided by a service or ts code? Seems like swapping out mock data for real shouldn't change the static html content at all.

webapp/src/main/webapp/src/app/admin/test-results/test-results.component.ts
```diff
@@ -0,0 +1,175 @@
...
+import { TestResultsService } from './service/test-results.service';
+import { TestResultFilters } from './model/test-result-filters';
+import { TestResultsChangeStatusModal } from './test-results-change-status.modal';
+
+class Column {
+  id: string; // en.json name
+  field: string; // TestResult field
+  sortable: boolean;
+
+  constructor({ id, field = '', sortable = true }) {
```
malaffoon, line 17: This is me being a ts newb: the default value for field, can/should that be `undefined`?

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1987,5 +1991,66 @@
     "head": "Deactivate pipeline",
     "body": "Are you sure you want to deactivate the {{pipelineName}} pipeline? By deactivating this script, no user transformations will be applied in the {{pipelineName}} pipeline.",
     "accept": "Deactivate"
+  },
+  "test-results": {
```
malaffoon, line 1995: Yeah, i'm doubling down on my request: i find "test-results" to be very confusing

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1987,5 +1991,66 @@
...
+    "state-header": "State",
+    "status-tooltip": "Status Tooltip information here",
+    "title": "Manage Test Results"
+  },
+  "test-results-subject-options": {
+    "ela": "ELA",
+    "elpac": "ELPAC",
+    "math": "Math",
+    "science": "Science"
+  },
```
malaffoon, line 2028: This data is dynamic and needs to come from a service. It shouldn't be in here, even when mocking.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1987,5 +1991,66 @@
...
+  "test-results-report-type-options": {
+    "individual": "Individual",
+    "aggregate": "Aggregate"
+  },
+  "test-results-status-options": {
```
malaffoon, line 2033: Can/should each of these sections be under the main "test-results" grouping? This really is a question, i'm not sure what the paradigm is. To show by example ...
```
{
  ...
  "test-results": {
    "all-text": "All",
    ...
    "status-options": {
      "loading": "Loading",
      ...
    }
  }
}
```
The difference in the code is minimal, `test-results.status-options.loading` vs. `test-results-status-options.loading`, but i think, in the en.json it groups things nicely.

webapp/src/main/webapp/package.json
```diff
@@ -53,12 +54,11 @@
...
     "tslib": "^1.9.0",
-    "webpack-dev-server": "^3.1.14",
+    "webpack-dev-server": "^3.11.0",
     "zone.js": "^0.8.14"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^0.13.1",
-    "@angular/cli": "7.1.3",
+    "@angular/cli": "^7.1.3",
     "@angular/compiler-cli": "^7.2.9",
```
gregcharles, line 62: What is driving these version changes? Should we clear them with Landon? In particular, I think ^7.1.3 means that the build could automatically update to any Angular less than 8.0.0, and it seems to me that at least could cause issues. I'm not sure about the others.

webapp/src/main/webapp/src/app/admin/test-results/service/test-results.service.ts
gregcharles: I don't know the typescript convention either, but personally I'd prefer braces even for one-line if-blocks, and I'd put the if expression on two lines instead of four. If typescript had `value in (item1, item2)` syntax, I'd be tempted to condense it down to one line. You could do `[item1, item2].includes(value)`, but that doesn't read as well.

webapp/src/main/webapp/src/app/admin/test-results/service/test-results.service.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    return true;
+  }
+
+  isDistrictAdmin(): boolean {
+    return this.districtAdmin;
+  }
+
```
gregcharles, line 157: I think the typescript  convention (or possibly Angular, I'm not sure of the boundaries yet) is just to use the field directly and a getter is implied. If you need extra logic, you can write a special getter method like:  
`get districtAdmin(): boolean {
    return someOtherCheck() && this._districtAdmin;
}`
Note: get is a separate word, and the field has to be distinguished in some way from the method. This would be invoked in the html template as `districtAdmin` (no parentheses, no "get").

webapp/src/main/webapp/src/app/admin/test-results/test-results.component.html
vgonzalez9: agreed, Ray wanted to basics there for UI the first part. 
The plan is to go back have that created by a service.

webapp/src/main/webapp/src/app/admin/test-results/test-results.component.ts
vgonzalez9: Well I followed an existing example. 
But I think it set to an empty string since the field is being used as an index in the template.

webapp/src/main/webapp/src/assets/i18n/en.json
vgonzalez9: Makes sense, will change.

webapp/src/main/webapp/package.json
vgonzalez9: this should not have been changed.

webapp/src/main/webapp/src/assets/i18n/en.json
vgonzalez9: those 'option' groups should have been removed as they are not being used.

webapp/src/main/webapp/src/assets/i18n/en.json
vgonzalez9: Yep they will be removed.

webapp/src/main/webapp/src/app/admin/test-results/service/test-results.service.ts
vgonzalez9: agreed on the parens being needed. 
Using multiple lines I was following some examples existing in the code. And there the value in convention, but I not sure it helps in this case. 
There is also some auto-formatting that is happening, where a single line is moved to multiple And I'm not sure where that is set.

webapp/src/main/webapp/package.json
vgonzalez9: How do you roll this file back?

webapp/src/main/webapp/src/app/admin/test-results-availability/service/test-results-availability.service.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  private applyTestResultsFilter(
+    testResult: TestResultAvailability,
+    testResultFilters: TestResultAvailabilityFilters
+  ) {
+    // may need to adjust order
+    if (this.filterStatus(testResultFilters.status, testResult.status)) {
+      return false;
+    }
+    if (
```
vgonzalez9, line 86: I'm not sure why this keeps getting reformatted - or why it only reformats some of these.
But these all initially looked like the the filterStatus if().

webapp/src/main/webapp/src/app/admin/test-results-availability/service/test-results-availability.service.ts
```diff
@@ -0,0 +1,200 @@
...
+    filteredSchoolYear: number,
+    schoolYear: number
+  ): boolean {
+    return (
+      filteredSchoolYear != this.schoolYearsDefault &&
+      filteredSchoolYear != schoolYear
+    );
+  }
+
+  isDistrictAdmin(): boolean {
```
vgonzalez9, line 135: And depending on were the actual check is down I left these as functions.

vgonzalez9:
so if we are ok with Availbility we can leave it but it can be changed to Visibility.

PR #2323, opened 2020-06-16
Handle Alt Score report review

gregcharles:
Alt Score report was missing grades and other info from the report review section of the configuration screen.


webapp/src/main/webapp/src/app/aggregate-report/aggregate-report-summary.component.ts
```diff
@@ -175,128 +184,108 @@ export class AggregateReportSummary {
...
+      case 'Claim':
+      case 'AltScore':
+        const reportSettings = getReportSettings(effectiveReportType);
+
+        assessmentAttributes = [
+          {
+            label: translate('aggregate-report-form.field.subjects-label'),
+            values: orAll(options.subjects, settings.subjects, subject =>
+              translate(`subject.${subject.code}.name`)
             )
```
gregcharles, line 210: This looks like bigger changes than it is. It's mostly just adding AltScore handling, but it also changes the string of if-else if's to a switch, and collapses the nearly identical handling of CustomAggregate, Claim, and AltScore into one.

PR #2335, opened 2020-06-23
BigFix_tempFiles

vgonzalez9:
Updated Audit file to match req1 listing
added place holder text in downloadable temp files.


vgonzalez9:
Updated with Marks suggestions on title/button change
replaced the downloadable pdf reference-template to be a real html file, still named reference-template

PR #2341, opened 2020-06-24
Req1: Fix default sort order using all fields but results_count

vgonzalez9:
Change mock district names to match existing test data.
Fixed the default sort order to include all of the fields (expect the results count)

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/test-results-availability/service/test-results-availability.service.ts
+++ b/webapp/src/main/webapp/src/app/admin/test-results-availability/service/test-results-availability.service.ts
@@ -34,8 +34,16 @@ export class TestResultsAvailabilityService implements OnInit {
     testResultFilters: TestResultAvailabilityFilters
   ): TestResultAvailability[] {
     // TODO: replace with real data
+
+    // Default Sort order SchoolYear(D),District(A),Subject(A),ReportType(A),Status(A)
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/src/app/admin/test-results-availability/service/test-results-availability.service.ts
```diff
@@ -34,8 +34,16 @@ export class TestResultsAvailabilityService implements OnInit {
...
     let testResults = this.mockTestResults.getMockTestResults().sort((a, b) => {
-      return b.schoolYear - a.schoolYear;
+      return (
+        b.schoolYear - a.schoolYear ||
+        a.district.localeCompare(b.district) ||
+        a.subject.localeCompare(b.subject) ||
+        a.reportType.localeCompare(b.reportType) ||
+        a.status.localeCompare(b.status)
+      );
     });
```
gregcharles, line 47: Interesting. So this will return a  number -- negative, positive, or zero -- and not just a boolean? It's pithy, but weird.

vgonzalez9: The localeCompare() method returns a number indicating whether the string comes before, after or is equal as the compareString in sort order.

PR #2349, opened 2020-06-25
Add-Check-For-Sandbox-User

vgonzalez9:
Used the UserService to retrieve the sandboxUser to use in the ISR Template and Test Results Availability features.
Updated and cleaned up old methods and flags.

### DIFF
--- a/webapp/src/main/webapp/src/app/admin/isr-template/isr-template-delete.modal.ts
+++ b/webapp/src/main/webapp/src/app/admin/isr-template/isr-template-delete.modal.ts
@@ -44,16 +44,14 @@ export class IsrTemplateDeleteModal implements OnInit {
   }
 
   delete() {
-    console.log(this.isrTemplate.templateName + ' Sandbox?' + this.sandboxUser);
     if (this.sandboxUser) {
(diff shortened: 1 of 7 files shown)


webapp/src/main/webapp/src/app/admin/isr-template/isr-template-delete.modal.ts
```diff
@@ -43,12 +44,8 @@ export class IsrTemplateDeleteModal implements OnInit {
...
   delete() {
-    console.log(
-      this.isrTemplate.templateName +
-        ' Sandbox?' +
-        this.isrTemplateService.sandbox
-    );
-    if (this.isrTemplateService.sandbox) {
+    console.log(this.isrTemplate.templateName + ' Sandbox?' + this.sandboxUser);
+    if (this.sandboxUser) {
       console.log('IS SANDBOX');
```
malaffoon, line 49: Should probably get rid of log messages now that it's working?

vgonzalez9: yes I should have removed that message

PR #2352, opened 2020-07-01
RDW6: fix bug with sandbox creation.

gregcharles:
The problem is actually with RDW_Schema, so I'm not sure this is the right place to fix it. The RDW_Schema code is:

`public class SchemaMigration {
    private Flyway flyway;

    public SchemaMigration(DataSource dataSource, String schemaName, SchemaMigration.Migration migration) {
        this.flyway = Flyway.configure().table("schema_version").locations(new String[]{migration.directory + "/sql"}).placeholders(Collections.singletonMap("schemaName", schemaName)).schemas(new String[]{schemaName}).dataSource(dataSource).load();
    }
...
`

RDW_Reporting was bringing in flyway 3.2.1, based on Spring Boot dependencies, but 5.2.4 is needed for the Flyway.configure() method. Should the RDW_Schema be changed to ensure that its internal calls use the version of flyway that it expects, even when RDW_Schema is used as a dependency to another project? 

This fix does force RDW_Reporting to use 5.2.4 instead of 3.2.1.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,6 +57,7 @@ subprojects {
         rootProject.ext['jackson.version'] = '2.9.9'
         rootProject.ext['mockito.version'] = '2.28.2'
         rootProject.ext['mysql.version'] = '5.1.46'  // 5.1.47 causes problems
+        rootProject.ext['flyway.version'] = '5.2.4'
         rootProject.ext['thymeleaf.version'] = '3.0.6.RELEASE'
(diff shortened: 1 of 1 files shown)


malaffoon:
Fix looks good. 

I slightly disagree with the problem statement: i think the issue is that, for whatever reason, upgrading gradle changed the dependency resolution: before it was getting the flyway version from RDW_Schema, but after it was getting it from Spring dependencies. I don't think there is anything we can do to the RDW_Schema project to change that behavior so this fix is the right thing to do.

PR #2360, opened 2020-07-13
Display dates now in this format: Jul 2, 2020, 2:30:07 PM

vgonzalez9:
Updated date format for config status message


webapp/src/main/webapp/src/app/admin/isr-template/service/isr-template.service.ts
```diff
@@ -16,7 +16,7 @@ export class IsrTemplateService {
   ) {}
 
   formatAsLocalDate(date: Date): string {
-    return this.datePipe.transform(date, 'yyyy-MM-dd hh:mm');
+    return this.datePipe.transform(date, 'MMM d, y, hh:mm:ss a');
```
malaffoon, line 19: American users. Sigh.

PR #2368, opened 2020-07-15
UAT-8-R1-UpdateSuccessMessage

vgonzalez9:
broke up current message so a manual message could be put together for the alert.
returning the new message


webapp/src/main/webapp/src/app/admin/test-results-availability/test-results-availability-change-status.modal.ts
```diff
@@ -60,18 +61,31 @@ export class TestResultsAvailabilityChangeStatusModal implements OnInit {
...
 
   private triggerEventStatus(
-    successfulChange: boolean,
-    unableToChange: boolean
+    unableToChange: boolean,
+    successChangeMsgOptions: string
   ) {
     this.changeStatusEvent.emit({
-      data: successfulChange,
+      // return the selected filters and status of the successful change
```
vgonzalez9, line 87: should indicate it's in the message - but probably should remove comment

PR #2373, opened 2020-07-15
UAT-62 (Req3)

greigcolling:
Adding a note above the Writing Trait Scores and updating the en.json file to include new text.

### DIFF
--- a/webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.html
+++ b/webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.html
@@ -278,7 +278,7 @@
     </ng-template>
   </p-table>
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.html
```diff
@@ -267,7 +278,7 @@
     </ng-template>
   </p-table>
 
-  <div *ngFor="let traitScoreSummary of traitScoreSummaries" class="mt-sm">
+  <div *ngFor="let traitScoreSummary of traitScoreSummaries" class="">
```
vgonzalez9, line 281: was class suppose to be blank?

greigcolling: It is not, thank you for catching that. ✅  fixed

PR #2376, opened 2020-07-16
UAT-8-R1 Fix Success Message

vgonzalez9:
Updated the the Successful Change message to clarify the status being changed.
Also fixed the error & success alert messages sometimes showing at the same time.

### DIFF
--- a/webapp/src/main/webapp/src/assets/i18n/en.json
+++ b/webapp/src/main/webapp/src/assets/i18n/en.json
@@ -2052,8 +2052,9 @@
     "school-year": "School Year:",
     "set-status-text": "Set status to:",
     "subjects": "Subjects:",
-    "successful-change-1": "Successfully submitted request to change the selected test results availability. ",
-    "successful-change-2": " It can take up to 24 hours to see these changes. ",
+    "successful-change-1": "Successfully submitted request to change the selected test results availability to ",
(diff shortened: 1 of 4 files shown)


vgonzalez9:
Also updated the message to pull all of the filter headers into the message

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -2052,8 +2052,9 @@
     "school-year": "School Year:",
     "set-status-text": "Set status to:",
     "subjects": "Subjects:",
-    "successful-change-1": "Successfully submitted request to change the selected test results availability. ",
-    "successful-change-2": " It can take up to 24 hours to see these changes. ",
+    "successful-change-1": "Successfully submitted request to change the selected test results availability to ",
+    "successful-change-2": " status.  (",
+    "successful-change-3": " It can take up to 24 hours to see these changes.",
     "state": "State:"
```
gregcharles, line 2058: Probably a nitpick, but I don't really like the -1, -2, -3.  Is there something more descriptive you can use? Also, successful-change-2 has an opening parenthesis, but where is closing one? What you might want to do is have:
`"successful-change-status": " status. ({{status-details}})"`
and then pass a string with the details into the translate call.

PR #2387, opened 2020-07-24
Patch for canned data for summative writing trait scores

gregcharles:


webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -132,37 +132,49 @@ export class ExamStatisticsCalculator {
         summaries.push(new WritingTraitScoreSummary(writingTraits));
       }
 
-      const itemsWithTraitScores = assessmentItem.scores.filter(
```
malaffoon, line 135: Sorry, not obvious to me from the context ... is this just for summatives? It looks like it applies to any?

malaffoon: Hmm, i see this method is only called for summatives. Nevermind.

gregcharles: Yes, that's right. I needed to take it out so the process didn't short-circuit and report no data before it could get to the canned data code.

PR #2395, opened 2020-07-30
First pass at displaying summative trait score data

malaffoon:
There are still a couple TODO's laying about. I'll work on them next.
Also, the Export functionality doesn't work right, so i need to address that. It wasn't called out in the requirements and is actually a bit tricky with the multiple purposes.


webapp/src/main/webapp/src/app/assessments/model/export-writing-trait-request.model.ts
```diff
@@ -1,12 +1,12 @@
 import { AssessmentItem } from './assessment-item.model';
-import { WritingTraitScoreSummary } from './writing-trait-score-summary.model';
+import { TraitScoreSummary } from './trait-score-summary.model';
```
malaffoon, line 2: As part of this, i renamed things. If "Writing" appears in the name, it refers to the legacy WER trait scores. Otherwise, it is a generic trait score that could, in theory, be used for non-ELA assessments/items.

webapp/src/main/webapp/src/app/assessments/model/trait-category-aggregate.model.ts
```diff
@@ -0,0 +1,21 @@
+import { TraitCategoryInfo } from './trait-category-info.model';
```
malaffoon, line 1: rename, not new

webapp/src/main/webapp/src/app/assessments/model/trait-category-info.model.ts
```diff
@@ -0,0 +1,23 @@
+/**
```
malaffoon, line 1: rename, not new

webapp/src/main/webapp/src/app/assessments/model/trait-score-summary.model.ts
```diff
@@ -0,0 +1,83 @@
+import { TraitCategoryAggregate } from './trait-category-aggregate.model';
```
malaffoon, line 1: rename, not new. But significantly changed.

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -6,9 +6,13 @@ import {
 } from '../model/exam-statistics.model';
```
malaffoon, line 6: Lots of changes in here, sorry.
Some of it is just cleaning up warnings (const vs. let, triple equals, etc.)

webapp/src/main/webapp/src/app/assessments/results/exam-statistics-calculator.ts
```diff
@@ -21,13 +25,18 @@ export class ExamStatisticsCalculator {
     exams: Exam[],
     numberOfLevels: number
   ): ClaimStatistics[] {
-    let stats = [];
+    const stats = [];
 
-    if (exams == null || exams.length == 0 || exams[0].claimScaleScores == null)
+    if (
```
malaffoon, line 30: gah, why does it insist on doing this reformatting?

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
```diff
@@ -50,8 +48,7 @@ class Column {
     headerInfo = false,
     styleClass = '',
     sortable = true,
-    index = -1,
-    points = -1
```
malaffoon, line 54: For the life of me, i couldn't find where this (points) was being used.

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
```diff
@@ -212,14 +199,11 @@ export class WritingTraitScoresComponent
...
     this.hasWritingTraitItems$ = this.items$.pipe(
       map(items => items.length > 0)
     );
 
-    // this.hasWritingTraitItems$.subscribe(val => console.log(val));
-
+    // TODO - where is this used? if only for interims, document that
     this.writingTraitType$ = this.items$.pipe(
       map(items =>
         items.length > 0 ? items[0].performanceTaskWritingType : undefined
```
gregcharles, line 209: I can't see that it's used anywhere. I think we could assume it's obsolete and see if deleting breaks anything we care about.

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
```diff
@@ -41,7 +40,6 @@ class Column {
 
   // Writing trait item column properties
   index?: number;
-  points?: number;
```
gregcharles, line 43: I'm sure there's a reason for removing points here. Is it obsolete or was it never necessary in the first place?

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
```diff
@@ -247,46 +231,37 @@ export class WritingTraitScoresComponent
       shareReplay(1)
     );
 
-    // this.itemViews$.subscribe(val => console.log(val));
-
+    // traitScoreSummaries is an array of maps of {purpose -> TraitScoreSummary}.
+    // For summatives, this is derived from exam-level trait scores.
+    // For interims it comes from item-level data.
+    //
+    // TODO - Why is it an array? Darn good question. I think it shouldn't be.
```
gregcharles, line 238: I agree. I'm also sort of vague on what it being observable does ... are we live loading in the background? In any case, I think we could make it observable of a single object rather than an array, or we could look at the size and log a big warning if it's more than one, and from then on just work with the first element.

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.html
```diff
@@ -105,20 +103,22 @@
...
+              {{ item.percents[column.index] | number: '1.0-0' }}%
+            </span>
           </div>
         </td>
       </tr>
     </ng-template>
 
+    <!-- TODO - how is this used??? -->
     <ng-template pTemplate="emptymessage" let-columns>
       <tr>
```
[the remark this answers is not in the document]
malaffoon: Yeah, what baffles me is how it is invoked. Is the pTemplate attribute a special value? I don't see `emptymessage` anywhere else in the source code. Is it that let-columns is empty? That doesn't seem right, either.

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
malaffoon: Yeah, see my comment below. It isn't being used anywhere that i can see. And, logically, it doesn't make sense. This is the column definition, so which points value (i.e. which row) would be in there? I think it is left over from something,

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
malaffoon: Hmm, maybe i'll make it a single value, extracting the first item if there is more than one. And, yes, log a warning if there is more than one.

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.html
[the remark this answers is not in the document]
gregcharles: It's special to PrimeNG widgets. In the table component they use various values as like hooks for table customizations, eg., emptymessage, caption, footer, etc. You can see the source by digging into node_modules, or at https://github.com/primefaces/primeng/blob/master/src/app/components/table/table.ts (but that will be a newer version unless you can find the older branch.)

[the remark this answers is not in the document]
malaffoon: Ah. Thanks for the explanation, i'll remove my accusatory TODO ;)

PR #2396, opened 2020-07-31
cleanup CSV export of trait summary

malaffoon:


webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -651,26 +651,6 @@ export class CsvBuilder {
     return this;
   }
 
-  withPerformanceTaskWritingType(
```
malaffoon, line 654: i combined the purpose column with the other trait columns (since the required arguments are similar)

webapp/src/main/webapp/src/app/csv-export/csv-builder.service.ts
```diff
@@ -787,7 +802,7 @@ export class CsvBuilder {
       ),
       item =>
         this.numberPipe.transform(
-          getWritingTraitAggregate(item).average,
+          getTraitCategoryAggregate(item).average,
```
malaffoon, line 805: There is nothing "writing"-specific about this trait score summary export so i removed that from a few variable names.

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -50,7 +48,7 @@ function getScoreCodes<T>(
       if (assessment != null && !isNullOrEmpty(getScoreCodes(assessment))) {
         const { subject: subjectCode, type: assessmentTypeCode } = assessment;
         const score = scoreCodes.find(
-          score => score.subjectCode === subjectCode
+          scoreType => scoreType.subjectCode === subjectCode
```
malaffoon, line 51: unrelated warning fixes

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -368,32 +365,22 @@ export class CsvExportService {
 
     exportRequest.assessmentItems.forEach((item, i) => {
       exportRequest.summaries[i].forEach((summary, purpose) => {
-        summary.rows
-          .filter(
-            row =>
-              exportRequest.assessment.type !== 'sum' ||
-              row.trait.type !== 'total'
-          )
```
malaffoon, line 376: The summaries are now type-specific so the summative summaries just don't have a 'total' row, so no need to filter it.

webapp/src/main/webapp/src/app/school-grade/results/school-assessment-export.service.ts
```diff
@@ -24,7 +24,7 @@ export class SchoolAssessmentExportService implements AssessmentExporter {
   ) {}
 
   exportItemsToCsv(exportRequest: ExportItemsRequest) {
-    let filename: string = this.getFilename(exportRequest);
+    const filename: string = this.getFilename(exportRequest);
```
malaffoon, line 27: unrelated warning fixes

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
gregcharles: +1   I hate all the warnings if only because they obfuscate real problems. I've been mostly taking the approach of making sure my new code is clean and cleaning up the general areas around new or changed code, but leaving everything as is. I assume your comment here pertains to the extraneous imports. Cleaning those up is safe. Changing == to === is safe _most_ of the time. I did hit one case where I applied some autocorrections over the whole file, instead of just the line I was looking at, and it actually broke functionality. Just something to keep in mind.

PR #2397, opened 2020-08-04
RDW6, req.1: Use real database values to populate Test Results screen

gregcharles:
Query for test results probably not performant on production-scale data. I'm researching ways to optimize it.


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsReportType.java
```diff
@@ -0,0 +1,18 @@
+package org.opentestsystem.rdw.admin.model;
+
+/**
+ * The report types for filtering Test Result Availability: Aggregate and Individual.
+ */
+public enum TestResultsReportType {
+    Aggregate,
+    Individual;
+
+    public static TestResultsReportType caseInsensitiveValue(final String value) {
```
malaffoon, line 10: I see that you're copying an existing naming pattern but this really should be `caseInsensitiveValueOf`

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
```diff
@@ -0,0 +1,43 @@
...
+    Reviewing(1),
+    Released(2);
+
+    int statusCode;
+
+    TestResultsStatus(int statusCode) {
+        this.statusCode = statusCode;
+    }
+
+    public static TestResultsStatus byStatusCode(Integer code) {
```
malaffoon, line 17: Wouldn't this be `valueOf`? I guess it doesn't matter much but that gives us three methods to get the enum from another value: `valueOf`, `byStatusCode`, `caseInsensitiveValue`. Not seeing much of a naming consistency.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
```diff
@@ -36,6 +38,11 @@
      */
     Embargo findOneBySchoolYearAndDistrictId(int schoolYear, long districtId);
 
+    List<TestResultsAvailability> findTestResultsAvailability(
+        Collection<Integer> includedStatuses,
+        Collection<Integer> filteredStatuses,
+        Collection<Long> districtIds);
+
```
malaffoon, line 45: Judgement call: perhaps a query object?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -147,4 +174,17 @@ private MapSqlParameterSource toSource(final Embargo embargo, final String usern
             .schoolYear(rs.getInt("school_year"))
             .readOnly(false)
             .build() : null;
+
+    private TestResultsAvailability mapToTestResults(ResultSet row) throws SQLException {
```
malaffoon, line 178: nit: param could be `final`

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -147,4 +174,17 @@ private MapSqlParameterSource toSource(final Embargo embargo, final String usern
...
+        return TestResultsAvailability.builder()
+            .districtId(row.getLong("district_id"))
+            .districtName(row.getString("district_name"))
+            .schoolYear(row.getInt("school_year"))
+            .subjectId(row.getLong("subject_id"))
+            .subjectCode(row.getString("subject_code"))
+            .status(TestResultsStatus.byStatusCode(getNullable(row, row.getInt("status"))))
```
malaffoon, line 185: status can be null? i'll read more ...

admin-service/src/main/java/org/opentestsystem/rdw/admin/security/AdminPermission.java
```diff
@@ -11,5 +11,11 @@
     public static final String TenantRead = "TENANT_READ";
     public static final String TenantWrite = "TENANT_WRITE";
 
+    public static final String TestDataLoadingRead = "TEST_DATA_LOADING_READ";
+    public static final String TestDataLoadingWrite = "TEST_DATA_LOADING_WRITE";
+    public static final String TestDataReviewingRead = "TEST_DATA_REVIEWING_READ";
+    public static final String TestDataReviewingWrite = "TEST_DATA_REVIEWING_WRITE";
```
malaffoon, line 17: I know the requirements use this verbiage but having "test" in these names is really confusing, and will get more and more confusing as time passes. Not sure i have a good suggestion. Perhaps just `DataLoadingRead` or `ExamDataLoadingRead` or `ExamLoadingRead`.

malaffoon: Hmm. Nevermind. I guess.

That's what the client called out. I hate the verbiage (and think we should push back on it) but the constants should match the verbiage.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -118,6 +125,62 @@ public void save(final User user, final Embargo embargo) {
...
+            user.getPermissionsById().get(EmbargoWrite) != null) {
+
+            loadingPermission = new Permission(TestDataLoadingWrite,
+                PermissionScope.builder().build());
+            reviewingPermission = new Permission(TestDataReviewingWrite,
+                PermissionScope.builder().statewide(true).build());
+        }
+
+        if (loadingPermission == null && reviewingPermission == null) {
+            throw new NoSuchElementException("Not authorized to change test results availability");
```
malaffoon, line 147: i bet you waffled on this exception: one could argue that this is InvalidStateException since a request should not make it to this service if the user doesn't have some embargo permission; or should it be NotAuthorized. Or, do you just let it through and return an empty collection.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -118,6 +125,62 @@ public void save(final User user, final Embargo embargo) {
                 .collect(Collectors.toList());
     }
 
+    @Override
```
malaffoon, line 128: Perhaps some javadoc (or just code comments) that explains what's going on?

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TestResultsController.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Collection;
+
+/**
+ * End-points for administering {@link TestResultsAvailability}s.
+ */
+@RestController
+@RequestMapping("/testResults")
+public class TestResultsController {
```
malaffoon, line 19: I would be fine with leaving all this in an EmbargoController. Having `Test` in the name of any class is confusing.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsAvailability.java
```diff
@@ -0,0 +1,142 @@
+package org.opentestsystem.rdw.admin.model;
+
+/**
+ * This class encapsulates the current test results availability state for a single
+ * district, school year, subject, and report type.
+ */
+public class TestResultsAvailability {
```
malaffoon, line 7: `TestResultsAvailability` or `TestResultAvailability`?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
malaffoon: I read more, it can't be null, right?

admin-service/src/main/resources/application.sql.yml
```diff
@@ -76,6 +76,34 @@ sql:
...
+        LEFT JOIN district_embargo de ON
+          asmt.school_year = de.school_year AND
+          sch.district_id = de.district_id AND
+          sub.id = de.subject_id
+        WHERE asmt.type_id = 3 AND exam.deleted = 0
+          AND IF(ert.name = 'Aggregate', coalesce(de.aggregate,0), coalesce(de.individual,0)) IN (:included_statuses)
+          AND (
+            IF(ert.name = 'Aggregate', coalesce(de.aggregate,0), coalesce(de.individual,0)) NOT IN (:filtered_statuses)
+              OR dist.id IN (:district_ids)
+           )
```
malaffoon, line 104: A conditional WHERE will not be performant (optimizer won't be able to use an index). I guess we're probably going to be doing something completely different so it doesn't matter.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TestResultsController.java
```diff
@@ -0,0 +1,36 @@
...
+    TestResultsController(final EmbargoService service) {
+        this.service = service;
+    }
+
+    /**
+     * @param user user
+     * @return collection of embargo settings for which the user has permission
+     */
+    @GetMapping
+    public Collection<TestResultsAvailability> get(@AuthenticationPrincipal final User user) {
```
malaffoon, line 33: In the UI code, you have a comment about the default order from the query. That implies that this should return an ordered collection, probably a `List`. Then, of course, the service should be changed to return a `List` (and the javadocs updated to indicate what the default sort order is).

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsReportType.java
gregcharles: Changed

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
gregcharles: Changed to valueOf(Integer). SO search showed a variety of naming possibilities for the case, but this one seems to have the most support.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
gregcharles: No, the query "coalesces" nulls to 0. I was being overly defensive.

admin-service/src/main/java/org/opentestsystem/rdw/admin/security/AdminPermission.java
gregcharles: Yes, that's why I did it. I don't like the underscores either, but I went with what the req said.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
gregcharles: I did, but other exceptional cases throw NoSuchElement even if that's not the best description of the problem. It does end up in the REST calls error message, so I thought it might be to prevent information leaking, e.g., somebody trying to hack into the service wouldn't know there was an auth problem.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
gregcharles: Added Javadoc to interface and some extra comments here.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TestResultsController.java
gregcharles: Another one I waffled on. I don't like TestResultsAvailability as a name. It's too long, especially in Angular naming conventions, which already lead to long names. Mostly I kept everything in embargo classes, but I created a new controller so I could have a different path. I could move it into EmbargoController as I start to delete the old implementations there.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsAvailability.java
gregcharles: Hmm, good point. The front end _mostly_ uses TestResultsAvailability, but the TypeScript model class is TestResultAvailability. I guess I'll match this to the model name, and think about what makes sense for the other components.

admin-service/src/main/resources/application.sql.yml
gregcharles: Well, not completely different. That is, I can see pre-aggregating the exam counts, but that still doesn't solve this problem. This problem being that the district_embargo sensibly has individual and aggregate status as two fields on the same row, but this query potentially has to filter out one of those statuses and not the other. If this approach fails in performance, we may have to redo the district_embargo table to split the rows and just have a single status field. This would have big impacts on data migration and the auditing process though.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/TestResultsController.java
gregcharles: Oh, yeah. I hadn't intended to guarantee a sort order, but when I saw that comment on the UI, I thought for a while about how to make a multi-field sort in javascript, got a headache, and then just threw it into the query and let the db worry about it. I didn't think about the documentation impacts to the server classes though. I'll either update it in the next rev, or really move the sort up into the frontend code.

PR #2401, opened 2020-08-10
get tenant-specific localizations when loading a tenant

malaffoon:


webapp/src/main/webapp/src/app/admin/tenant/pages/tenant/tenant.component.ts
```diff
@@ -126,13 +126,24 @@ export class TenantComponent implements OnInit, OnDestroy {
...
     );
 
-    this.localizationDefaults$ = this.translationLoader
-      .getTranslation(this.languageStore.currentLanguage)
-      .pipe(
-        map(defaults => flatten(defaults)),
-        publishReplay(),
-        refCount()
-      );
+    // If editing an existing tenant then load the localizations and details
```
malaffoon, line 129: If there is a tenant in context, use it to get the tenant-specific localizations. Otherwise, do what it was doing before: fetch the default localizations for the current tenant. (I did try to make a temporary local variable for the `id` but it wasn't working quite right so a couple lines of duplication.)

I do have a question: what is the lifecycle of an Angular component. I *think* this is being instantiated for every tenant/sandbox being viewed (either edited or created) but i'm not sure of that. If it is reused for different tenants, then this fix is not okay.

gregcharles: I _think_ this kind of work should be in ngOnInit() rather than the constructor. That way you'd definitely rerun it every time the page (or route, whatever) was reloaded. You could put a breakpoint in the constructor in the browser dev tools though, and see if there's a new instance for new tenant view. I don't know enough to say that's impossible.

malaffoon: To be honest, i was afraid to move the code around too much. Since the existing code was fetching the tenant details here, i figured this was safe. However, how the data is being used elsewhere may be different.

gregcharles: Yeah, that makes sense to me too. My preference for ngOnInit is based on my very shallow understanding of the lifecycle and generally what I've seen in the other components. However, I assume Landon wrote this code, so what's there must be right and you are doing basically the same sort of thing.

PR #2403, opened 2020-08-12
RDW6, req.1: wire in the filter options based on permissions

gregcharles:


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultAvailabilityOptions.java
```diff
@@ -0,0 +1,98 @@
...
+            viewAudit = options.viewAudit;
+            districtAdmin = options.districtAdmin;
+            districts = options.districts;
+
+            return this;
+        }
+
+        public Builder statuses(final Collection<TestResultsStatus> statuses) {
+            this.statuses = new ArrayList<>(statuses);
+            Collections.sort(this.statuses);
```
malaffoon, line 79: minor logic nit: i think the sorting should be done in build. Why? If you add another convenience method that add a single status, then each method would have to sort.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultAvailabilityOptions.java
```diff
@@ -0,0 +1,98 @@
...
+            return this;
+        }
+
+        public Builder districtAdmin(boolean districtAdmin) {
+            this.districtAdmin = districtAdmin;
+            return this;
+        }
+
+        public Builder districts(final Collection<Organization> districts) {
+            this.districts = districts;
```
malaffoon, line 94: nit: should be making a copy of the collection, not using it directly. I know we have mixed patterns for this: 
1. Use the passed collection
2. Create a collection in the builder
3. Create a copy of the collection in the object getter
4. ...
But i think creating a collection in the builder is the safest and most versatile.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -185,6 +177,27 @@ public void save(final User user, final Embargo embargo) {
         return repository.findTestResultsAvailability(includeStatuses, filterStatuses, districtIds);
     }
 
+    @Override
+    public TestResultAvailabilityOptions getFilterOptions(final User user) {
+        final Permission loadingPermission = user.getPermissionsById().get(TestDataLoadingWrite);
+        final Permission reviewingPermission = user.getPermissionsById().get(TestDataReviewingWrite);
+
+        final boolean districtAdmin = reviewingPermission != null && !reviewingPermission.getScope().isStatewide();
```
malaffoon, line 185: I'm not sure exactly how this flag will be interpreted by the UI but ... if i have state reviewing permission AND a district reviewing permission, this will set me as a districtAdmin. What will the UI show?

malaffoon: Okay, i read more. Flag as is is probably good.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultAvailabilityOptions.java
gregcharles: OK, good. I appreciate the explanation.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultAvailabilityOptions.java
gregcharles: Copied both statuses and districts into immutable collections on build.

PR #2404, opened 2020-08-12
ELA summative assessments shouldn't display trait scores unless trait…

malaffoon:
… reports are enabled; add docs about remote debugging containers

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -279,6 +279,45 @@ tenantProperties:
       name: Test Tenant
 ```
 
+#### Remote Debugging (from IDE)
+There may be times when it is necessary (or just easier) to remote debug instead of running from the IDE.
+
+If you are running docker locally, this is straightforward:
+1. Modify the docker-compose.yml file to enable the jvm debug client in the application of interest.
+There are examples in the checked in file, but you'll be adding an environment variable and a port
+mapping; example snippet:
+```yaml
+    ports:
+      - 5007:5007
+    environment:
+      - JAVA_OPTS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5007
+```
+2. Configure your IDE to remote debug to localhost on the specified port, `5007` in this example.
+
+If you want to remote debug a kubernetes instance, there is an extra step.
+1. Modify the k8s deployment configuration and add an environment variable and container port;
+example snippet:
+```yaml
+kind: Deployment
+spec:
+  template:
+    spec:
+      containers:
+        env:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -279,6 +279,45 @@ tenantProperties:
       name: Test Tenant
 ```
 
+#### Remote Debugging (from IDE)
```
malaffoon, line 282: I spent a while getting this working so i thought i'd capture the steps.

webapp/src/main/webapp/src/app/assessments/results/assessment-results.component.ts
```diff
@@ -248,16 +248,17 @@ export class AssessmentResultsComponent implements OnInit {
...
-      (this._assessmentExam.assessment.type === 'sum' &&
-        this._assessmentExam.assessment.traitReportEnabled)
-    );
+    if (this._assessmentExam.assessment.type === 'sum') {
+      return this._assessmentExam.assessment.traitReportEnabled;
+    } else if (this._assessmentExam.assessment.subject === 'ELA') {
+      return this._assessmentExam.assessment.hasWerItem;
+    } else {
+      return false;
+    }
```
malaffoon, line 261: The way the logic was before, a Summative ELA assessment would show trait scores even if trait reports were disabled for that assessment. The new logic is more deliberate: *all* summatives use the trait report flag, *then* the hard-coded ELA check.

CONTRIBUTING.md
```diff
@@ -279,6 +279,45 @@ tenantProperties:
...
+```
+2. Use `kubectl` to port forward the jvm debug port for the pod of interest. This needs a dedicated terminal
+(or you need to background the process). 
+```bash
+kubectl get po | grep myapp
+myapp-deployment-6d9f7fb7b-qmd5k     1/1     Running   0          17h
+
+kubectl port-forward myapp-deployment-6d9f7fb7b-qmd5k 5007:5007
+``` 
+3. Configure your IDE to remote debug to localhost on the specified port, `5007` in this example.
```
gregcharles, line 321: Interesting, so we could attach to a QA pod and debug against that with these instructions? Yes, that could be very useful! 

- Maybe there should be a disclaimer about issues with debugging against shared servers?
- "... there is an extra step," to me implies to take the above instructions and do this one more thing. That could be just me though. I'd probably put it as, "If you want to remote debug a kubernetes instance, the configuration is just slightly more complicated."

malaffoon: Good suggestions, i'll add those comments.

PR #2405, opened 2020-08-18
RDW6, req1 -- Embargo audit log

gregcharles:
I have in mind that all the TestResults code will change to Embargo, and the TestResultsController will just go away. That's why I put the audit log request into the Embargo controller.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoAuditRecord.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoAuditRecord.java
@@ -3,8 +3,7 @@ package org.opentestsystem.rdw.admin.model;
 import java.util.Date;
 
 /**
- * This class encapsulates the current test results availability state for a single
- * district, school year, subject, and report type.
+ * This class encapsulates an embargo audit record by school year, district, and subject.
  */
 public class EmbargoAuditRecord {
     private int schoolYear;
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoAuditRecord.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoAuditRecord.java
@@ -1,0 +1,180 @@
+package org.opentestsystem.rdw.admin.model;
+
+import java.util.Date;
+
+/**
+ * This class encapsulates the current test results availability state for a single
+ * district, school year, subject, and report type.
+ */
+public class EmbargoAuditRecord {
+    private int schoolYear;
+    private String district;
+    private String subject;
+    private TestResultsStatus individualStatus;
+    private TestResultsStatus previousIndividualStatus;
+    private TestResultsStatus aggregateStatus;
+    private TestResultsStatus previousAggregateStatus;
+    private String user;
+    private Date updated;
+    private String reportType;
+
+    public int getSchoolYear() {
+        return schoolYear;
+    }
+
+    public String getDistrict() {
+        return district;
+    }
+
+    public String getSubject() {
+        return subject;
+    }
+
+    public TestResultsStatus getIndividualStatus() {
+        return individualStatus;
+    }
+
+    public TestResultsStatus getPreviousIndividualStatus() {
+        return previousIndividualStatus;
+    }
+
+    public TestResultsStatus getAggregateStatus() {
+        return aggregateStatus;
+    }
+
+    public TestResultsStatus getPreviousAggregateStatus() {
+        return previousAggregateStatus;
+    }
+
+    public String getUser() {
+        return user;
+    }
+
+    public Date getUpdated() {
+        return updated;
+    }
+
+    public String getReportType() {
+        return reportType;
+    }
+
+    public TestResultsStatus getStatus() {
+        if ("Individual".equalsIgnoreCase(getReportType())) {
+           return getIndividualStatus();
+        }
+        if ("Aggregate".equalsIgnoreCase(getReportType())) {
+            return getAggregateStatus();
+        }
+
+        return null;
+    }
+
+    /**
+     * @return new builder with this instance's values
+     */
+    public Builder copy() {
+        return builder().copy(this);
+    }
+
+    /**
+     * @return new builder
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private int schoolYear;
+        private String district;
+        private String subject;
+        private TestResultsStatus individualStatus;
+        private TestResultsStatus previousIndividualStatus;
+        private TestResultsStatus aggregateStatus;
+        private TestResultsStatus previousAggregateStatus;
+        private String user;
+        private Date updated;
+        private String reportType;
+
+        public EmbargoAuditRecord build() {
+            final EmbargoAuditRecord auditRecord = new EmbargoAuditRecord();
+
+            auditRecord.schoolYear = schoolYear;
+            auditRecord.district = district;
+            auditRecord.subject = subject;
+            auditRecord.individualStatus = individualStatus;
+            auditRecord.previousIndividualStatus = previousIndividualStatus;
+            auditRecord.aggregateStatus = aggregateStatus;
+            auditRecord.previousAggregateStatus = previousAggregateStatus;
+            auditRecord.user = user;
+            auditRecord.updated = updated;
+            auditRecord.reportType = reportType;
+
+            return auditRecord;
+        }
+
+        public Builder copy(final EmbargoAuditRecord auditRecord) {
+            schoolYear = auditRecord.schoolYear;
+            district = auditRecord.district;
+            subject = auditRecord.subject;
+            individualStatus = auditRecord.individualStatus;
+            previousIndividualStatus = auditRecord.previousIndividualStatus;
+            aggregateStatus = auditRecord.aggregateStatus;
+            previousAggregateStatus = auditRecord.previousAggregateStatus;
+            user = auditRecord.user;
+            updated = auditRecord.updated;
+            reportType = auditRecord.reportType;
+
+            return this;
+        }
+
+        public Builder district(final String district) {
+            this.district = district;
+            return this;
+        }
+
+        public Builder schoolYear(final Integer schoolYear) {
+            this.schoolYear = schoolYear;
+            return this;
+        }
+
+        public Builder subject(final String subject) {
+            this.subject = subject;
+            return this;
+        }
+
+        public Builder individualStatus(final TestResultsStatus status) {
+            this.individualStatus = status;
+            return this;
+        }
+
+        public Builder previousIndividualStatus(final TestResultsStatus status) {
+            this.previousIndividualStatus = status;
+            return this;
+        }
+
+        public Builder aggregateStatus(final TestResultsStatus status) {
+            this.aggregateStatus = status;
+            return this;
+        }
+
+        public Builder previousAggregateStatus(final TestResultsStatus status) {
+            this.previousAggregateStatus = status;
+            return this;
+        }
+
+        public Builder user(final String user) {
+            this.user = user;
+            return this;
+        }
+
+        public Builder updated(final Date updated) {
+            this.updated = updated;
+            return this;
+        }
+
+        public Builder reportType(final String reportType) {
+            this.reportType = reportType;
+            return this;
+        }
+    }
+}
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
@@ -15,11 +15,6 @@ public enum TestResultsStatus {
     }
 
     public static TestResultsStatus valueOf(Integer code) {
-        // Default for non-existing record is Loading
-        if (code == null) {
-            return Loading;
-        }
-
         for (final TestResultsStatus status : values()) {
             if (status.getStatusCode() == code) {
                 return status;
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
@@ -15,8 +15,9 @@ public enum TestResultsStatus {
     }
 
     public static TestResultsStatus valueOf(Integer code) {
+        // Default for non-existing record is Loading
         if (code == null) {
-            return null;
+            return Loading;
         }
 
         for (final TestResultsStatus status : values()) {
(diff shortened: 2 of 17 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoAuditRecord.java
```diff
@@ -0,0 +1,180 @@
+package org.opentestsystem.rdw.admin.model;
+
+import java.util.Date;
+
+/**
+ * This class encapsulates the current test results availability state for a single
+ * district, school year, subject, and report type.
+ */
```
malaffoon, line 8: I'm sure i'll see later on, but for my edification, the only purpose for this class is to read from the database and write to the CSV file, right?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
```diff
@@ -15,8 +15,9 @@
     }
 
     public static TestResultsStatus valueOf(Integer code) {
+        // Default for non-existing record is Loading
```
malaffoon, line 18: Feels like business logic that doesn't belong in the enum class itself. If you leave it like this, should the caseInsensitiveValueOf have the same logic?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
```diff
@@ -43,6 +45,14 @@
         Collection<Integer> filteredStatuses,
         Collection<Long> districtIds);
 
+    /**
+     * Retrieve an audit log of embargo updates.
+     *
+     * @param startDate start date for the retrieved audit log.
+     * @return date ordered list of embargo update records
```
malaffoon, line 52: ascending or descending?

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoAuditRecord.java
gregcharles: Hmm, yes. I'll fix that comment.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
```diff
@@ -43,6 +45,14 @@
...
+     */
+    List<EmbargoAuditRecord> findEmbargoAuditRecords(Date startDate);
```
malaffoon, line 54: I'm mostly okay with the way it is but i feel obliged to suggest that the audit repository could be a separate interface than the rest of the embargo repository functionality.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultsStatus.java
gregcharles: Yeah, I didn't like it there either. It's easiest and arguably cleanest though, so I thought I'd throw it out there. How about I move it to the service?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
gregcharles: I vacillated, but decided on descending. I'll fix the comment unless you think it should be ascending.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -177,6 +179,32 @@ public void save(final User user, final Embargo embargo) {
...
+        final Permission loadingPermission = user.getPermissionsById().get(TestDataLoadingWrite);
+        if (loadingPermission == null) {
+            throw new NoSuchElementException("Audit log not available");
+        }
+
+        return splitByReportType(repository.findEmbargoAuditRecords(startDate));
+    }
+
+    // Splits record into rows for Aggregate and Individual results. Filters out any that haven't changed status.
+    private List<EmbargoAuditRecord> splitByReportType(List<EmbargoAuditRecord> embargoAuditRecords) {
```
malaffoon, line 194: nit: param could be final

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -177,6 +179,32 @@ public void save(final User user, final Embargo embargo) {
...
+        // Check permission
+        final Permission loadingPermission = user.getPermissionsById().get(TestDataLoadingWrite);
+        if (loadingPermission == null) {
+            throw new NoSuchElementException("Audit log not available");
+        }
+
+        return splitByReportType(repository.findEmbargoAuditRecords(startDate));
+    }
+
+    // Splits record into rows for Aggregate and Individual results. Filters out any that haven't changed status.
```
malaffoon, line 193: We aren't really going to be putting entries in the table that don't have any status change, are we?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
gregcharles: I vacillated on that too, especially since if we didn't hand-roll the jdbc, it would more naturally be two repos. Hmm, it's more injection to the service, but I guess I'll split it.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
```diff
@@ -40,6 +50,26 @@
...
+     * @param fileName the name for the downloaded file
+     */
+    @GetMapping(path = "/audit")
+    public void getAudit(
+        @AuthenticationPrincipal final User user,
+        final HttpServletResponse response,
+        @Value("${embargo.audit-log-days:7}") int auditLogDays,
+        @Value("${embargo.audit-log-file-name:}") String fileName)
+    {
+        final Date startDate = Date.from(Instant.now().plus(-auditLogDays, DAYS));
```
malaffoon, line 68: isn't there a `minus` method?

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
```diff
@@ -40,6 +50,26 @@
...
+     * @param response the response to attach the file to
+     * @param auditLogDays number of days to include in the log file
+     * @param fileName the name for the downloaded file
+     */
+    @GetMapping(path = "/audit")
+    public void getAudit(
+        @AuthenticationPrincipal final User user,
+        final HttpServletResponse response,
+        @Value("${embargo.audit-log-days:7}") int auditLogDays,
+        @Value("${embargo.audit-log-file-name:}") String fileName)
```
malaffoon, line 66: I would suggest that these be class member variables, not arguments (since they are derived from properties and won't change from call to call). Are we going to allow for a `days` argument, or are they stuck always getting just the last 7 days worth?

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/converter/EmbargoAuditRecordsCsvHelper.java
```diff
@@ -0,0 +1,115 @@
...
+        final String auditLogFileName,
+        final List<EmbargoAuditRecord> auditRecords,
+        final HttpServletResponse response)  {
+
+        try {
+            setHeadersResponse(auditLogFileName, response);
+            printCsv(auditRecords, response.getOutputStream());
+        } catch (IOException e) {
+            logger.warn("Problem exporting report CSV", e);
+            throw new RuntimeException(e);
```
malaffoon, line 51: log-and-throw anti-pattern

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
gregcharles: Only because we have two statuses in the same row ... if there's a status update for Individual results, we still have a row showing the new Individual status and the unchanged Aggregate status.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/converter/EmbargoAuditRecordsCsvHelper.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    private static void setHeadersResponse(String filename, HttpServletResponse response) {
```
malaffoon, line 55: nit: params could be final

admin-service/src/main/resources/application.sql.yml
```diff
@@ -104,6 +104,23 @@ sql:
            )
       GROUP BY exam.school_year, dist.id, sub.id, ert.name
 
+    findEmbargoAuditRecords: >-
+      SELECT
+        ade.school_year AS school_year,
+        d.name AS district,
```
malaffoon, line 110: Just noticed this: did the requirements specify just the district name? I ask because there are districts with duplicate names and, usually, they want both the name and the natural id.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
gregcharles: Shoot, there is. I specifically looked for that and still somehow missed it.

admin-service/src/main/resources/application.yml
```diff
@@ -20,6 +20,12 @@ endpoints:
 flyway:
   enabled: false
 
+embargo:
+  # Number of days to include in audit log of embargo changes
+  audit-log-days: 365
```
malaffoon, line 25: It's odd to have the default here not match the default in the code.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
gregcharles: The reqs says Smarter will tell us how far back the audit log should go, but not that it's configurable through the UI.

webapp/src/main/webapp/src/app/admin/test-results-availability/test-results-availability.component.ts
```diff
@@ -221,6 +220,8 @@ export class TestResultsAvailabilityComponent
       this.toReportTypeKey
     );
 
+    this.schoolYearOptions$.subscribe(opt => console.log(opt));
```
malaffoon, line 223: remove?

webapp/src/main/webapp/src/app/admin/test-results-availability/test-results-availability.component.ts
```diff
@@ -296,19 +297,20 @@ export class TestResultsAvailabilityComponent
...
-      `.csv`;
-    this.testResultsService.getTemplateFile().subscribe(
-      (download: Download) => {
-        saveAs(download.content, auditFilename);
-      },
-      (error: Error) => {
-        console.error(error);
-      }
-    );
+    // const now = this.testResultsService.formatAsLocalDate(new Date());
```
malaffoon, line 300: remove lava code or WIP?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
malaffoon, line 193: nit: suggest "proper" javadocs with an explanation that we'll understand in 7 months :)

admin-service/src/main/resources/application.sql.yml
gregcharles: Hmm, the reqs just say district. Vicky just had the district name in the mocked-up results. Should I add natural ID or wait and see what they say?

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
malaffoon: No preference, just like it documented.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
malaffoon: Again, you could go halfway: separate interfaces but one implementation.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
malaffoon: Right. But at least one of the two will always be changed.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/EmbargoController.java
malaffoon: Cool. So, yeah, just change to class member variables.

admin-service/src/main/resources/application.sql.yml
malaffoon: Just wait then. They signed off.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/converter/EmbargoAuditRecordsCsvHelper.java
gregcharles: Copied verbatim from AbstractAggregateReportCsvHandler and I can't really think of way to improve it. Log but don't throw? Throw but don't log?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
gregcharles: Right, I think so. In theory, it would be possible to insert a district embargo record with Loading status for both report types, which would be no status change for either one. I don't think we'll do that through the app though.

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/converter/EmbargoAuditRecordsCsvHelper.java
malaffoon: I'd have to dig a bit but, in general, if you are throwing the exception without adding more value, then you can let the higher level method deal with it. So, "throw but don't log". Of course, it shouldn't be changed such that nothing is logged, so have to look around.

PR #2406, opened 2020-08-18
add trait sorting so display and CSV export is consistent

malaffoon:


webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
```diff
@@ -364,19 +364,26 @@ export class CsvExportService {
...
+                purpose: purpose,
+                traitCategoryAggregate: row
+              });
+
+              if (row.trait.maxPoints > maxPoints) {
+                maxPoints = row.trait.maxPoints;
+              }
+            });
         });
-      });
```
gregcharles, line 379: Were we going to sort the categories by name and max points?

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
```diff
@@ -322,7 +322,7 @@ export class WritingTraitScoresComponent
   }
 
   purposes(traitScoreSummary: Map<string, any>) {
-    return Array.from(traitScoreSummary.keys());
+    return Array.from(traitScoreSummary.keys()).sort();
```
gregcharles, line 325: This sorts alphabetically by the purpose code, right? That should be OK.

webapp/src/main/webapp/src/app/csv-export/csv-export.service.ts
malaffoon: Hmm, i do recall discussing that. For now, i'm going to leave it just by name and see what they say.

webapp/src/main/webapp/src/app/assessments/results/view/writing-trait-scores/writing-trait-scores.component.ts
malaffoon: Yup. Hope they're okay with that. I could certainly imagine a situation where the codes don't "match" up to the names and it gets weird.

PR #2407, opened 2020-08-20
SRS-171: ELAS Status display update.

junyuchen1994:


common/src/main/java/org/opentestsystem/rdw/reporting/common/service/impl/DefaultUserStudentFieldService.java
```diff
@@ -31,10 +31,11 @@
             final SearchFieldPermissionLevel permissionLevel = reportingSystemProperties.getStudentFields().get(field);
 
             // Add the filter if
-            // 1. There is no entry in the reportingSystemProperties map (the default permission level is "everyone")
+            // 1. There is no entry in the reportingSystemProperties map
+            // (the default permission level is "everyone" except for ELAS)
             // 2. The permission level is explicitly set to enabled
             // 3. The permission level is set to admin and the user has individual PII read permission
-            if (permissionLevel == null
+            if (permissionLevel == null && !field.equals(StudentFieldType.EnglishLanguageAcquisitionStatus)
```
malaffoon, line 38: What is the requirement? This is hardcoding something making it no longer configurable. I'd really like to understand what they are asking for that couldn't be handled by a configuration setting?

malaffoon: To put it another way: why did they not just set the ELAS field to Disabled?

junyuchen1994: final List<StudentFieldType> fields = new ArrayList<>();
        for (final StudentFieldType field : StudentFieldType.values()) {
            final SearchFieldPermissionLevel permissionLevel = reportingSystemProperties.getStudentFields().get(field);

            // Add the filter if
            // 1. There is no entry in the reportingSystemProperties map (the default permission level is "everyone")
            // 2. The permission level is explicitly set to enabled
            // 3. The permission level is set to admin and the user has individual PII read permission
            if (permissionLevel == null&&!field.equals(StudentFieldType.EnglishLanguageAcquisitionStatus)
            || permissionLevel == SearchFieldPermissionLevel.Enabled
            || (
                permissionLevel == SearchFieldPermissionLevel.Admin
                && user.getPermissionsById().containsKey(ReportingPermission.IndividualPiiRead)
            )) {
                fields.add(field);
            }
        }
        return fields;
    }

I guess we should make it more flexible for the user. This is not hard coding , it's still configurable.  there are three checks for the studentFilter field,  if there is no entry defined in the config server,  it returns null , and therefore gets added to the student Filter response because they are enabled by default , but not for the ELAS , UI then uses the response to determine whether to display the section or not. they override the studentFilter  in the tenant level application.yaml file, and only specify some attributes. Second and third check  can be used to check for different Setting for the field in config server , you can still set  ELAS to Enabled or Admin , it will get added to the response,  you can also set it to Disabled , which will bypass the if statement , the the UI will be display the field.

malaffoon: I call it hard-coding because one of the fields (ELAS) is being hard-coded to behave differently than the others. 

The intent is that the embedded defaults provide a default value for all student fields. If that isn't working properly, we should figure out why because it may affect other fields, or future configuration changes. In lieu of that, i can only object to this change and move on.

PR #2409, opened 2020-08-24
RDW6,req1 -- embargo server-side pagination

gregcharles:
WIP -- more work to do on retrieving filters independently of  embargo data.


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoQuery.java
```diff
@@ -0,0 +1,186 @@
...
+    private Set<Long> districtIds;
+    private Set<TestResultsStatus> statuses;
+
+    // Optional sort field and flag for descending
+    private String sortField;
+    private boolean descending = false;
+
+    // Page size and offset for paging.
+    private Integer pageSize;
+    private Integer rowOffset;
```
malaffoon, line 27: Did you look into `Pageable`. Spring has a set of classes for doing pagination (and slicing) in a standard way. All the controllers, etc. have default handling of them.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/EmbargoRepository.java
```diff
@@ -40,10 +41,14 @@
...
-        Collection<Integer> filteredStatuses,
-        Collection<Long> districtIds);
+    /**
+     * Find a page of test result availability records with the given filters, sort order
+     * and page definitions.
+     *
+     * @param query query object with filters, sort order, page size and offset.
+     * @return sorted list of one page of TestResultAvailability records.
+     */
+    List<TestResultAvailability> findTestResultsAvailability(EmbargoQuery query);
```
malaffoon, line 51: Per comment above this would be either `Slice<TestResultAvailability>` or `Page<TestResultAvailability>`

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -125,22 +139,65 @@ public Embargo findOneBySchoolYearAndDistrictId(final int schoolYear, final long
...
+                .addValue("page_size", query.getPageSize())
+                .addValue("offset", query.getRowOffset()),
             (row, index) -> mapToTestResults(row)
         );
     }
 
+    private String getOrderBy(final String sortField, final boolean descending) {
+        // Change hyphens to underscores to match aliases in sql
+        final String normalizedSortField = StringUtils.isBlank(sortField) ? null :
+            sortField.replace('-', '_');
```
malaffoon, line 178: Seems brittle. Perhaps just have a mapping of the names. Is the sortField string just something the UI knows about? Why not have underscores then? Sorry, just not seeing what the whole picture.

admin-service/src/main/resources/application.sql.yml
```diff
@@ -77,32 +77,42 @@ sql:
...
+              d.id                       AS district_id,
+              d.name                     AS district_name,
+              s.id                       AS subject_id,
+              s.code                     AS subject_code,
+              ert.name                   AS report_type,
+              coalesce(de.individual, 0) AS individual,
+              coalesce(de.aggregate, 0)  AS aggregate,
+              coalesce(ec.count, 0)      AS exam_count
+      FROM school_year y
+            JOIN district d ON TRUE
```
malaffoon, line 90: Do you need the `ON TRUE` or are you just being explicit?

admin-service/src/main/resources/application.sql.yml
```diff
@@ -77,32 +77,42 @@ sql:
...
+            LEFT JOIN exam_count ec ON
+                  y.year = ec.school_year and
+                  d.id = ec.district_id and
+                  s.id = ec.subject_id
+      WHERE ((ert.name = 'Individual' AND coalesce(individual, 0) IN (:statuses)) OR
+              (ert.name = 'Aggregate' AND coalesce(aggregate, 0) IN (:statuses)))
+              AND (:filter_by_year = 0 OR y.year = :year)
+              AND (:filter_by_district = 0 OR d.id IN (:district_ids))
+              AND (:filter_by_subject = 0 OR s.id = :subject_id)
+              AND (:filter_by_report_type = 0 OR ert.name = :report_type)
```
malaffoon, line 106: nit: i would suggest using null params instead of having a flag for every one. Fewer params to pass in and consistent with other SQL. For example:
`    AND (:year IS NULL OR y.year = :year)`

malaffoon: Oh, unless, of course, there is additional logic going into the setting of the flags.

admin-service/src/main/resources/application.sql.yml
```diff
@@ -77,32 +77,42 @@ sql:
...
+            LEFT JOIN district_embargo de ON
+                  y.year = de.school_year and
+                  d.id = de.district_id and
+                  s.id = de.subject_id
+            LEFT JOIN exam_count ec ON
+                  y.year = ec.school_year and
+                  d.id = ec.district_id and
+                  s.id = ec.subject_id
+      WHERE ((ert.name = 'Individual' AND coalesce(individual, 0) IN (:statuses)) OR
+              (ert.name = 'Aggregate' AND coalesce(aggregate, 0) IN (:statuses)))
```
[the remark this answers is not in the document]
malaffoon: nit/question: did you find you had to coalesce for this to work? That is, if individual is NULL (presumably because no entry in table), did it falsely match/not-match the statuses? Hmm, i must be tired because i'm having trouble thinking this one through.

webapp/src/main/webapp/src/app/admin/test-results-availability/service/test-results-availability.service.ts
```diff
@@ -107,11 +108,13 @@ export class TestResultsAvailabilityService implements OnInit {
...
   // receive test results and apply filter's options
   getTestResults(
-    testResultFilters: TestResultAvailabilityFilters
+    testResultFilters: TestResultAvailabilityFilters,
+    query: EmbargoQueryType
   ): Observable<TestResultAvailability[]> {
     // Default Sort order from query:
     // SchoolYear(D),District (by name) (A),Subject (by ID) (A), ReportType(A), Status(A)
-    return this.dataService.get(`${ResourceContext}`).pipe(
+    console.log('Post to data service');
```
malaffoon, line 116: nit: remove

webapp/src/main/webapp/src/app/admin/test-results-availability/test-results-availability.component.ts
```diff
@@ -58,6 +56,15 @@ export class TestResultsAvailabilityComponent
 
   private grabFocusToAlert = false;
 
+  // Paging data
+  totalRowCount = 1000;
```
malaffoon, line 60: nit: this seems inappropriate

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoQuery.java
gregcharles: Hmm, no, I haven't used it before and didn't think of it. Giving it a cursory look now, it seems like it would have been somewhat useful, but not a whole lot. I'd also worry it would be hard to fit our special use cases, especially around status, into its framework.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -125,22 +139,65 @@ public Embargo findOneBySchoolYearAndDistrictId(final int schoolYear, final long
...
 
+    public int countDistricts() {
+        return template.query("select count(*) from district", new MapSqlParameterSource(), rs -> rs.next() ? rs.getInt(1) : 0);
+    }
+
+    public List<String> findAllDistricts() {
+        return template.query("select id, name from district order by id",
+            new MapSqlParameterSource(),
+            (row, index) -> "" + (index + 1) + ":" + row.getInt("id") + ":" + row.getString("name"));
+    }
```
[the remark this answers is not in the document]
gregcharles: Yes, these were obsolete. Now that I'm re-implementing the filters, I removed them already and intended make a new repo. I wanted a total count, but I really just need a way to know there are more than n districts, where n is probably configurable.  I could probably force OrganizationRepo to give me that info if just by setting the limit to (n+1) and seeing if I got that many districts back. I will still need a way to filter district by partial names though, and I don't see how to get that with a new query.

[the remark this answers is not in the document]
gregcharles: Also, how are the permissions configured? If I pass in the new permissions for TestDataLoading/ReviewingWrite to the Org Repo, will they work right now, or are there new settings somewhere needed to make the new permissions authorized to access the appropriate districts?

admin-service/src/main/resources/application.sql.yml
gregcharles: Hmm, weird. Existing queries use 1=1, and I thought TRUE was slightly less cryptic. I was 92% sure I tried it with no "ON" at all though and it didn't work, so I just thought it was a quirk of MySQL, like protecting us from doing outer joins unless we proved we were sure. However, I just tried again and it's fine doing outer joins with no "ON TRUE", at least from IntelliJ's sql client. I'll try it again in integration.

admin-service/src/main/resources/application.sql.yml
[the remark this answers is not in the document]
gregcharles: Yeah, that's true. I'll add the de.

Re: coalesce, I did experiment with various scenarios to avoid the coalesce, but couldn't make them work. If we assumed there are only two cases -- permissions to all statuses or permissions to just Reviewing and Released, I could simplify. Maybe that's a better way to go.

admin-service/src/main/resources/application.sql.yml
gregcharles: I waffled on that and finally decided I liked the readability of this way better, but only by a small amount. I could go either way.

webapp/src/main/webapp/src/app/admin/test-results-availability/test-results-availability.component.ts
gregcharles: Yes, I'm going to need another repo call to get the total rows based on the current filter settings. I just set a high value to enable the paging controls in the table.

admin-service/src/main/resources/application.sql.yml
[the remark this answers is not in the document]
gregcharles: Actually, there's a third case of filtering by any one of the statuses, including Loading. Still there might be a way to simplify even so.

admin-service/src/main/java/org/opentestsystem/rdw/admin/model/EmbargoQuery.java
malaffoon: This will probably fall in the category of "i wish i had stuck to my guns". Those are the classes to be used when doing paging in a Spring app. They provide consistency and predictability. 

However, this is not my project or product so i'm not going to die on this (or any RDW) hill.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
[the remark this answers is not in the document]
malaffoon: Good question, have to look and see. I suspect it is based on PII being granted for the district.

gregcharles:
I've made some of the recommended changes to the sql query. Other recommendations will be incorporated into the next PR (for the filter functionality). I'll look into refactoring to add Spring's paging framework once other features are complete.

PR #2410, opened 2020-08-26
WIP: add template resolution for thymeleaf

malaffoon:
Trying to keep the PR's small: this adds template fragment resolution. That is, when generating reports, use configured ISR templates to render the body of the report.


common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/ReportTemplateRepository.java
```diff
@@ -0,0 +1,60 @@
+package org.opentestsystem.rdw.reporting.common.repository;
```
malaffoon, line 1: This is in common because it will be used by admin (when configuring templates) and report-processor (when rendering reports).

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ArchiveBackedReportTemplateRepository.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.OutputStream;
+import java.util.Optional;
+import java.util.StringJoiner;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * A ReportTemplateRepository backed by an archive service.
+ */
+public class ArchiveBackedReportTemplateRepository implements ReportTemplateRepository {
```
malaffoon, line 18: Because it is in common, i prefer to not have prototype annotations on it. The app that needs it must create an appropriate bean.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/thymeleaf/ReportTemplateResolver.java
```diff
@@ -0,0 +1,133 @@
...
+ * matching template available, it will return the default template. NOTE: there are
+ * two default templates: it will look in the repository for a template named
+ * "default.html" and there is an embedded default.html as the final fallback.
+ *
+ * As a safety net for upgrades, the resolver will fallback to legacy report templates
+ * if configured templates are not found for them. For the subjects "Math" and "ELA"
+ * the appropriate embedded template, "iab-body.html", "ica-body.html", or "sum-body.html"
+ * will be returned instead of "default.html".
+ */
+public class ReportTemplateResolver extends AbstractConfigurableTemplateResolver {
```
malaffoon, line 35: Thymeleaf and Spring combine to make this pretty easy: all TemplateResolvers found in the classpath will be loaded and given a chance to find and load fragments.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/thymeleaf/ReportTemplateResolver.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private Optional<String> getDefaultContents() {
+        Optional<String> contents = repository.getAsString(DefaultTemplateName);
+        if (!contents.isPresent()) {
+            contents = getTemplatesResourceAsString(DefaultTemplateName);
+        }
+        return contents;
+    }
+
+    /**
```
malaffoon, line 98: I chose to do this for ease of migration and development. The upgrade instructions will still have separate, external templates for the SmarterBalanced subjects (there will be six of them) that will need to be either manually installed or configured in the UI.

webapp/src/main/resources/application.yml
```diff
@@ -4,6 +4,15 @@ app:
       # every 2 mins (2 * 60 * 1000)
       refresh-delay: "120000"
 
+cloud:
```
malaffoon, line 7: Annoyance. Because reporting common now depends on archive, and webapp depends on reporting common, the automatic AWS region detection got pulled in. These settings just disable that auto detection. Which we don't want anywhere, but can't seem to get rid of.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ArchiveBackedReportTemplateRepository.java
```diff
@@ -0,0 +1,64 @@
...
+    final static String BasePath = "TEMPLATES/";
+
+    private final ArchiveService archiveService;
+    private final LocationStrategy locationStrategy;
+
+    public ArchiveBackedReportTemplateRepository(final ArchiveService archiveService) {
+        this.archiveService = archiveService;
+        this.locationStrategy = id -> new StringJoiner("/", BasePath, "").add(id).toString();
+    }
+
```
gregcharles, line 28: So, just so I can find these later, we expect them to be in /tmp/TEMPLATES locally?

Also regarding logistics, these will disappear for us on restart. Will that mess up our local testing?

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/thymeleaf/ReportTemplateResolver.java
gregcharles: If they fat finger something and end up with Moth_sum.html, does that just not match anything? What about math_sum.html? Would that get kicked too? It's probably moot anyway, since it will be impossible for them to make a mistake like that if they use UI, right?

webapp/src/main/resources/application.yml
gregcharles: Yeah, I vaguely remember that being a problem somewhere. It is annoying.

common/src/main/java/org/opentestsystem/rdw/reporting/common/repository/impl/ArchiveBackedReportTemplateRepository.java
malaffoon: I believe that is correct: by default, local dev stores the archive under /tmp.

For local testing, you'll have to drop the files in, either manually or using the UI. That's kind of why i put that legacy template handling in: so that stuff will basically work the way it used to even if you don't have any templates configured.

report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/thymeleaf/ReportTemplateResolver.java
malaffoon: Correct. If they use the UI it will work properly. And if they manually drop in files, they have to live with any mistakes they make.

PR #2411, opened 2020-08-26
RDW6, req1 -- retrieve filter options in separate  REST call

gregcharles:
Test results availability data in now paged, so the filter options can't be derived from the results, so must be retrieved up front. Also, more careful consumption of the server data, since multiple subscriptions and pipes through async were causing the framework to rerun queries unnecessarily, significantly affecting performance.

### DIFF
--- a/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultAvailabilityOptions.java
+++ b/admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultAvailabilityOptions.java
@@ -1,9 +1,11 @@
 package org.opentestsystem.rdw.admin.model;
 
 import com.google.common.collect.ImmutableList;
+import org.opentestsystem.rdw.reporting.common.model.CodedEntity;
 import org.opentestsystem.rdw.reporting.common.model.Organization;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -17,7 +19,11 @@ public class TestResultAvailabilityOptions {
     private List<TestResultsStatus> statuses;     // All the statuses available to the user, sorted by enum order.
     private boolean viewAudit;                    // True if user has permission to view audits.
     private boolean districtAdmin;                // True for district administrator
-    private Collection<Organization> districts;   // Collection of districts for district admins.
+    private Collection<Organization> districts;   // Collection of districts (complete for district admin, possibly size limited for state users.
+    private int districtCount;                    // Total count of districts, in case it is too many to retrieve all at once.
+    private Collection<CodedEntity> subjects;     // Collection of districts
+    private Collection<Integer> schoolYears;      // Collection of active school years in the system.
+    private Collection<TestResultsReportType> reportTypes;
 
     public List<TestResultsStatus> getStatuses() {
         return statuses;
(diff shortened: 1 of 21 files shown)


admin-service/src/main/java/org/opentestsystem/rdw/admin/model/TestResultAvailabilityOptions.java
```diff
@@ -17,7 +19,11 @@
     private List<TestResultsStatus> statuses;     // All the statuses available to the user, sorted by enum order.
     private boolean viewAudit;                    // True if user has permission to view audits.
     private boolean districtAdmin;                // True for district administrator
-    private Collection<Organization> districts;   // Collection of districts for district admins.
+    private Collection<Organization> districts;   // Collection of districts (complete for district admin, possibly size limited for state users.
+    private int districtCount;                    // Total count of districts, in case it is too many to retrieve all at once.
+    private Collection<CodedEntity> subjects;     // Collection of districts
```
malaffoon, line 24: comment seems wrong

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/SchoolYearRepository.java
```diff
@@ -0,0 +1,14 @@
+package org.opentestsystem.rdw.admin.repository;
```
malaffoon, line 1: could we put this in reporting common instead of duplicating; there are other repositories there to copy from

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -139,39 +139,51 @@ public Embargo findOneBySchoolYearAndDistrictId(final int schoolYear, final long
...
+    @Override
+    public int countByQuery(final EmbargoQuery query) {
+        // Remove any page limits to ensure we get the proper count.
+        final EmbargoQuery noLimitsQuery = query.copy().rowOffset(0).pageSize(Integer.MAX_VALUE).build();
+
+        // Convert existing SQL to count rows.
+        final String sql = String.format("select count(*) from (%s) as results",
+            addOrdering(findTestResultsAvailability, " "));
+
+        System.out.println(sql);
```
malaffoon, line 163: eep!

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/SchoolYearRepository.java
gregcharles: Makes sense. It's a different schema, but I see where the OrganizationRepository manages to overcome that, so I'll just follow that example.

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
gregcharles: ugh, I may be working too hard

gregcharles:
Hmm, SchoolYearRepository is actually in three services now ... reporting-service, aggregate-service, and now admin-service. I'm going to make it a tech debt ticket to consolidate all three. SubjectRepository is in _four_ services, so I'll make a tech debt ticket for that as well.

PR #2416, opened 2020-08-31
RDW6-req1 implement embargo save.

gregcharles:


admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
```diff
@@ -239,13 +249,63 @@ public void update(final Embargo embargo, final String username) {
...
+    public int update(final EmbargoUpdate embargo, final String username) {
+        int updated = 0;
+
+        final boolean filterByReportType = embargo.getReportType() != null;
+        // New records may need to be created if transitioning from Loading
+        if (embargo.getStatus() == Loading) {
+            int count = template.update(createEmbargo, toCreateSource(embargo, username));
+
+            // If not limited to a single report type, count the records affected as double
+            updated += (filterByReportType ? count : 2 * count);
```
malaffoon, line 263: If transitioning from Loading, there will be both inserts and updates?

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -132,6 +136,58 @@ public void save(final User user, final Embargo embargo) {
...
+            ImportContent.EMBARGO,
+            "admin embargo modification",
+            digest,
+            "embargo update");
+
+        try {
+            int updateCount = repository.update(reconciledUpdate, user.getUsername());
+
+            if (updateCount == 0) {
+                final String message = "No test results availability updates needed for given update query. Ignoring.";
```
malaffoon, line 164: nit: i wouldn't put that final `Ignoring` in the message

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
```diff
@@ -280,10 +375,10 @@ private EmbargoQuery reconcileQueryWithUserPermissions(EmbargoQuery query, User
     private List<EmbargoAuditRecord> splitByReportType(final List<EmbargoAuditRecord> embargoAuditRecords) {
         final List<EmbargoAuditRecord> split = new ArrayList<>(2 * embargoAuditRecords.size());
         embargoAuditRecords.forEach(r -> {
-            if (!statusMatch(r.getIndividualStatus(), r.getPreviousIndividualStatus())) {
+            if (statusChanged(r.getIndividualStatus(), r.getPreviousIndividualStatus())) {
```
malaffoon, line 378: 👍

webapp/src/main/webapp/src/app/admin/test-results-availability/test-results-availability.component.ts
```diff
@@ -312,6 +316,12 @@ export class TestResultsAvailabilityComponent implements OnInit, DoCheck {
       this.successfulChange = true;
       this.alertFailure = null;
       this.grabFocusToAlert = true;
+      this.testResultAvailabilityFilters.status = updatedStatus;
+
+      // This needs a slight delay to work properly
```
malaffoon, line 321: why is that? perhaps further clarification in the comment

admin-service/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcEmbargoRepository.java
gregcharles: There could be inserts and updates, say if they updated one district to reviewing, then later updated the whole state.  For the count,  I don't know if I'll use it for anything except  for logging when there were 0 changes, because the current embargo save does that. So, the exact count probably doesn't mater, but it seemed reasonable to keep the counts consistent at one per year/district/subject/type change.

admin-service/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultEmbargoService.java
gregcharles: 👍

webapp/src/main/webapp/src/app/admin/test-results-availability/test-results-availability.component.ts
gregcharles: Yeah, I don't know. Without the delay, it just refreshes with no rows or sometimes one row showing. I have a short list of things to ask Landon about, so I'll add this one.

PR #2417, opened 2020-09-01
add end-point to get reference template

malaffoon:


admin-service/build.gradle
```diff
@@ -45,6 +45,7 @@ dependencies {
     implementation 'org.springframework:spring-context-support'
     implementation 'org.xmlunit:xmlunit-core'
 
+    compileOnly 'javax.servlet:javax.servlet-api'
```
malaffoon, line 48: IntelliJ is a lot happier when it has the source, even if it isn't needed for the build.

docker-compose.yml
```diff
@@ -81,7 +81,7 @@ services:
     ports:
       - 8280:8080
       - 8208:8008
-#      - 5005:5005 #debug
+      - 5008:5008 #debug
```
malaffoon, line 84: It is convenient to have the services running with the jvm debugging enabled

webapp/src/main/webapp/src/app/admin/isr-template/service/isr-template.service.ts
```diff
@@ -108,6 +108,7 @@ export class IsrTemplateService {
 
   public getTemplateFile(): Observable<any> {
     // TODO: Replace with call to real file location
+    // TODO - hmm, need the real reference template in a common lib i guess
```
malaffoon, line 111: the point of this PR; obviously it still needs to be wired to the front-end

docker-compose.yml
```diff
@@ -81,7 +81,7 @@ services:
...
     volumes:
```
gregcharles, line 85: Probably not meant for checkin?

webapp/src/main/webapp/src/app/admin/isr-template/service/isr-template.service.ts
```diff
@@ -16,7 +16,7 @@ export class IsrTemplateService {
   ) {}
 
   formatAsLocalDate(date: Date): string {
-    return this.datePipe.transform(date, 'MMM d, y, hh:mm:ss a');
+    return this.datePipe.transform(date, 'MMM d, y');
```
malaffoon, line 19: The requirements don't have the time in the mockup

docker-compose.yml
malaffoon: I left it there on purpose: docker-compose is only used by developers and it doesn't hurt to have the jvm debugging enabled

admin-service/src/main/java/org/opentestsystem/rdw/admin/web/ReportTemplateController.java
```diff
@@ -43,15 +41,52 @@ public ReportTemplateController(final ReportTemplateService service) {
...
+     *
+     * @param response HTTP response to write file contents to
+     * @param subjectCode subject code, e.g. "Math"
+     * @param assessmentType assessment type, e.g. "sum"
+     */
+    @GetMapping(path = "{subjectCode}/{assessmentType}", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
+    public void getTemplate(final HttpServletResponse response,
+                            @PathVariable final String subjectCode,
+                            @PathVariable final String assessmentType) {
+        try {
```
gregcharles, line 81: I see you changed the interface for this call as well. Just curious about that ... is it a generally better way to do it, or did the reference template endpoint need to do it this way and you're trying to keep the two endpoints consistent?

PR #2427, opened 2020-09-11
RDW6, req5: stopgap fix for cached ISR Templates to allow testing.

gregcharles:


report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/service/ThymeleafTemplateProcessor.java
```diff
@@ -20,7 +21,7 @@
     private String encoding = DEFAULT_ENCODING;
 
     @Autowired
-    ThymeleafTemplateProcessor(final ITemplateEngine engine) {
+    ThymeleafTemplateProcessor(final ITemplateEngine engine, final ReportTemplateRepository templateRepository) {
```
malaffoon, line 24: what is this for?

gregcharles: Are you still connected? :) It was an experiment that I'm taking out now. However, there is a problem with templates being cached that I'm trying to solve now. Basically, you upload a new template, but it's resolved from cache without ever going to s3. I thought maybe I could get the latest change timestamp, and clear caches as appropriate. That doesn't work for deletes though. For now, I'm just going to clear the cache every time, and then work back to a real solution.

malaffoon: One advantage (?) of an open source project. I'm still getting notifications.

Yeah, that caching worried me a bit. It would be great to flush the cache whenever a change to a template is made but they are separate services, so that involves eventing, yada, yada. I think a better approach would be to make the cache short, maybe 5 minutes. That would make the generation of a large report faster but give a reasonable response to changes. Of course, that could lead to the format of a really large report changing mid-stream but that would be a corner case.

PR #2547, opened 2021-07-15
SRS-197: Create permissions in RDW instead of Permissions service

gregcharles:


xuping-xu:
Overall the fixes are ok. 

One comment on IT test: The itShouldFindAllRolesAndPermissions test in JdbcPermissionRepositoryIT.java simply checks whether map is not empty. However, this does not establish that all roles and permissions are obtained. Maybe consider checking one specific role's permission (you may need to add such role and permission first -- maybe in RDW Schema?).

PR #2600, opened 2022-11-01
Fix accessibility issue with background color

mfish0005:


xuping-xu:
Looks good to me. We can merge it to develop once I set up TeamCity build access to GitHub RDW private repos.

